PR #56, opened 2025-07-07
Product knowledge for Cursor

moxious:
I am building a knowledge base about this project that I found useful to pour into Cursor.  Sometimes I'm iteratively building "context" that we can use to generate other artifacts.   

It is not important to merge this PR, we can leave it open - I may contribute more.  This is all in a separate directory so no merge conflicts, it'll just be here when we want it.


moxious:
Ref: https://github.com/grafana/grafana-recommender/pull/16

Jayclifford345 (approved):

PR #95, opened 2025-08-04
Refactor interactive hook, break up functionality, greatly increase unit testing

moxious:
Goal is to refactor interactive hook and get it to the point where it's got solid unit test coverage to keep the interactive portions of the code base as safe as is reasonable.

This branch introduces no new functionality, just reorganization and unit tests

Major things:

One giant unified hook is broken into many pieces:

* No "magic wait times" (setTimeout with some random number).  Instead we're aggressively reusing waitForReactUpdates() to let the DOM settle)
* NavigationManager (handles scrolling to elements, making sure they're visible, docking the nav bar)
* StateManager (handles unified error logging, ensuring that elements are always set to "complete" or "error" when executing)
* DOM utils (miscellaneous functions that don't have React dependencies which can be gotten out of the interactive hook and tested separately)
* Action handlers (focus, formfill, button, navigate).  This abstracts each individual interactive action that the app can take into its own discrete class, with good unit testing.  This in turn makes the hook pretty simple.


Jayclifford345:
Tested all interactive tutorials on branch and works great!

PR #108, opened 2025-08-13
Add post verify steps to interactive elements

tomglenn:
Adds a `data-verify` attribute to allow for post-step verification checks. These checks run after the "Do it" button is pressed and should be used to verify the action was successful. Uses the same checks as `data-requirements` under the hood.

### DIFF
--- a/src/utils/action-handlers/navigate-handler.ts
+++ b/src/utils/action-handlers/navigate-handler.ts
@@ -3,7 +3,10 @@ import { InteractiveElementData } from '../../types/interactive.types';
 import { locationService } from '@grafana/runtime';
 
 export class NavigateHandler {
-  constructor(private stateManager: InteractiveStateManager, private waitForReactUpdates: () => Promise<void>) {}
+  constructor(
+    private stateManager: InteractiveStateManager,
+    private waitForReactUpdates: () => Promise<void>
+  ) {}
 
   async execute(data: InteractiveElementData, navigate: boolean): Promise<void> {
     this.stateManager.setState(data, 'running');
--- a/src/utils/docs-retrieval/components/interactive/interactive-step.tsx
+++ b/src/utils/docs-retrieval/components/interactive/interactive-step.tsx
@@ -285,10 +285,10 @@ export const InteractiveStep = forwardRef<{ executeStep: () => Promise<boolean> 
                 {checker.isChecking
                   ? 'Checking...'
                   : isShowRunning
-                  ? 'Showing...'
-                  : !checker.isEnabled
-                  ? 'Requirements not met'
-                  : 'Show me'}
+                    ? 'Showing...'
+                    : !checker.isEnabled
+                      ? 'Requirements not met'
+                      : 'Show me'}
               </Button>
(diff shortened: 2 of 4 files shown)


src/utils/requirements-checker.utils.ts
```diff
@@ -137,6 +137,74 @@ export async function checkRequirements(
...
+    return {
+      requirements: verifyString || '',
+      pass: true,
+      error: []
+    };
+  }
+
+  const checks: string[] = verifyString.split(',').map(check => check.trim());
+
+  async function performVerify(check: string): Promise<CheckResultError> {
```
moxious, line 161: just a general comment -- the body of this function is basically a big switch statement on different kinds of checks.

I'm 99% sure that this exact function is duplicated somewhere else for requirements checking.  Can you see if that is the case, and then reuse the other version rather than re-implementing?   Main reason is that when we introduce a new check, this is another place to copy the code to, to make sure that whatever new checks are implemented are provided for everywhere in the code base.

src/utils/requirements-checker.utils.ts
```diff
@@ -36,105 +36,130 @@ export interface RequirementsCheckOptions {
...
+
+interface CheckContext {
+  targetAction?: string;
+  refTarget?: string;
+}
+
+function shouldIncludeInMode(_check: string, _mode: CheckMode): boolean {
+  // Allow all checks in both modes. Authors can choose appropriate tokens for pre vs post.
+  return true;
+}
```
jdbaldry, line 49: Maybe a little early to abstract this predicate given that it always returns true?

jdbaldry (commented):
Looks like there's a few instances of passing around strings and trimming to check for specific values. It feels like we can use TypeScript to define these as types like enums so we can avoid a lot of the defensive coding like  `(postVerify && postVerify.trim() !== '')`

Let me know if any of my comments are unclear. I'm happy to dig in a bit more

Jayclifford345:
Hey @tomglenn would love to get this in when you have a chance

src/utils/docs-retrieval/components/interactive/interactive-section.tsx
```diff
@@ -146,7 +148,8 @@ export function InteractiveSection({
   const { 
     executeInteractiveAction, 
     startSectionBlocking, 
-    stopSectionBlocking 
+    stopSectionBlocking,
+    checkPostconditionsFromString,
```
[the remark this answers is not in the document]
tomglenn: It does make sense, but post conditions use the same check logic as pre-step requirements. Both of these can take arbitrary "parameters" which need to be parsed.

e.g. `has-datasource:prometheus` or `on-page:/dashboards`

Jayclifford345:
Yo my dude! just FYI I moved the first-dashboard example out of the bundled and into the other repo a while back. Happy to add it back as a bundled example or just remove altogether. The reason i mention it is due to the fact we specify the example but do not declare it in `index.json` for it to show up. I shall add it manaully for testing but just wanted to give you an FYI

PR #116, opened 2025-08-22
Pass through target action and ref target to requirements checker appropriately

tomglenn:


moxious (commented):
LGTM, though, general comment that the number of things that can occur on an interactive element is getting big (requirements, objectives, hints, targetAction, refTarget, etc) and rather than passing them as keyword args, where appropriate it might be good to factor out a separate type like InteractiveStepType that can capture some of the what's optional/what's not, default value stuff.

PR #118, opened 2025-08-25
Adds delayed retries for checking reftargets in requirements

tomglenn:


Jayclifford345:
LGTM ~ would you mind fixing lint issues with Pretty and we can get merged :)

Issue #132, opened 2025-08-27
Repeated refreshes of recommender during pageloads


moxious:
https://ops.grafana-ops.net/a/grafana-metricsdrilldown-app

In the drilldown section -- as the page progressively loads and the various drilldown animations come in, the Pathfinder window gets spammed with updates.  It seems like there's something in the update loop that isn't waiting for the page to settle

the key thing to look for is if your visualizations are already loaded you won't see this, but if they're progressively loading in drilldown, you'll see spam refreshes in the recommender window

PR #133, opened 2025-08-28
fix(context): removed duplicate debouncing systems

Jayclifford345:
This PR solves #132 

This PR introduces a central timeout manager which supports either debouncing or standard timeouts. This allows other components such as the context engine to leverage these features in a standarized method. This primarily to prevent services from being called multiple times in quick session due to events being triggered in grafana


jdbaldry (commented):
The new `TimeoutManager` class says that [`setTimeout` isn't debounced](https://github.com/grafana/docs-plugin/pull/133/files#diff-e48a7a613e8f4ac10dcaaa60af9394ed1df517fcb5f8e3f20609bc3c05bb0d06R54-R61) but we're using it in all the situations where we expect debouncing. Is that correct?

Jayclifford345:
Hey @jdbaldry, so we are using the timeout manager as expected
```
  // All context changes that trigger a refresh should share this debounce to prevent repeated API calls and changes to the UI.
  const debouncedRefresh = useCallback(
    (delay?: number) => {
      timeoutManager.setDebounced('context-refresh', fetchContextData, delay, 'contextRefresh');
```

This is primarily to solve the context update issue, where we call a context reload on each dom change which caused the spam. I have updated the documentation in the setDebounced to be more distinctive from the setTimeout

jdbaldry:
> Hey @jdbaldry, so we are using the timeout manager as expected
> 
> ```
>   // All context changes that trigger a refresh should share this debounce to prevent repeated API calls and changes to the UI.
>   const debouncedRefresh = useCallback(
>     (delay?: number) => {
>       timeoutManager.setDebounced('context-refresh', fetchContextData, delay, 'contextRefresh');
> ```
> 
> This is primarily to solve the context update issue, where we call a context reload on each dom change which caused the spam. I have updated the documentation in the setDebounced to be more distinctive from the setTimeout

Sorry, you're right, I misread things!

PR #152, opened 2025-09-03
chore: Fix repo to support auto update workflows

Jayclifford345:
The plugins team recently added translation support to the core plugin scaffolding rather than needing an extended web confg. I also moved our docker compose changes to the top level docker compose so we no longer have our settings overwritten.


jdbaldry (approved):
Nice! Had some of this patch in my own branch during development, glad it's making it into `main`

PR #153, opened 2025-09-03
fix(context): hotfix deduplicate datasources

Jayclifford345:
Sorry built from #152 

Hotfix for the database type. Wraps it set to eliminate duplicate database type entries in the list.


jdbaldry (approved):
LGTM I've updated the base branch because it is based on the workflow changes and this cleans up the diff.

PR #159, opened 2025-09-05
feat(config): Add custom docs and feature flag

Jayclifford345:
This PR adds a new tab to the configuration page called custom docs. This will lay the ground work for allowing customers to eventually add their own docs to the plugin. For now it acts as a development tool so we can add test repos with interactive tutorials and bypass adding them to the recommender.

I have added this feature under a feature flag which is enabled by default when you deploy locally using npm run server. Otherwise, the feature will be blocked by default. 

To test: 
1. Head to: https://docsplugin2.grafana-dev.net/plugins/grafana-grafanadocsplugin-app?page=custom-docs-config
2. Open developer console
3. type: window.features.enable('custom_docs');
4. Refresh the page, but do not close the tab, as this will remove your session.

Set up a custom docs repo. My recommendation for testing is just to use: https://github.com/grafana/interactive-tutorials

Screenshot 2025-09-05 at 16 01 28

### DIFF
--- a/src/components/docs-panel/docs-panel.tsx
+++ b/src/components/docs-panel/docs-panel.tsx
@@ -646,13 +646,17 @@ function CombinedPanelRenderer({ model }: SceneComponentProps<CombinedLearningJo
 
   // Scroll tracking
   useEffect(() => {
-    if (activeTab && activeTab.content && contentRef.current) {
-      const cleanup = setupScrollTracking(contentRef.current, activeTab, isRecommendationsTab);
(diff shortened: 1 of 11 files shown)


jdbaldry:
Thinking about this implementation a bit and stepping back.

I don't believe we think this is how custom user provided docs will work so this is really just to provide alternative recommendations in the recommender.

Would it be easier to just configure the recommender URL so you can just run a recommender with whatever configuration you like and have it recommend documentation in the way it would in production?

PR #161, opened 2025-09-08
fix(interaction): simplified model detection logic

Jayclifford345:
Removed extensive DOM selectors, which caused the full-screen block to work for some modals but not all. To test this PR you will need to run the Prometheus interactive  tutorial:
1. Open Pathfinder on the Grafana homepage
2. Select Prometheus & Grafana 101
3. Scroll down to `do section` and select
4. Watch screen blocker (you will see a flashing border around areas you should not be able to click)
5. You will need to complete the second section to see the full-screen blocker at work. Select do section

You will see a full-screen blocker appear for initial datasource selection in dashboard creation, and when the dashboard is being saved at the end. If you wish to understand how this PR fixes the issue try the same exerise as above on main branch you will see when the dashboard saves at the end the full screen block fails to activate.

### DIFF
--- a/src/components/docs-panel/docs-panel.tsx
+++ b/src/components/docs-panel/docs-panel.tsx
@@ -646,13 +646,17 @@ function CombinedPanelRenderer({ model }: SceneComponentProps<CombinedLearningJo
 
   // Scroll tracking
   useEffect(() => {
-    if (activeTab && activeTab.content && contentRef.current) {
-      const cleanup = setupScrollTracking(contentRef.current, activeTab, isRecommendationsTab);
+    // Only set up scroll tracking for actual content tabs (not recommendations)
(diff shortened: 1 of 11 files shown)


jdbaldry:
OK, I think I get you :) Let me run this back to you and see if I've got it right.

- There are two mechanism for blocking user actions.
- One is called integrated and blocks everything on the left-side (is that everything except our sidebar or just the navigation menu?)
- The other is called full screen and blocks the entire screen.

- We always want to block the full screen when there are modals open. Even if there's a navigation drawer open.
- The removed logic doesn't use the full screen block when there's a drawer of some kind but that code wasn't correct when it was originally committed.

- Removing the code fixes that and ensures our desired blocking when modals are on screen.

PR #240, opened 2025-10-20
Fix allowed recommender domains

jdbaldry:


Jayclifford345:
will merge and fix test in another PR

PR #251, opened 2025-10-21
fix: resolve Windows CRLF line ending issue in entrypoint.sh

tonypowa:
## Problem

Dear team. I created this PR with help from Cursor. Totally understand if it's not a priority - just wanted to contribute back after hitting this locally.

Windows users might encounter `exec /entrypoint.sh: no such file or directory` when running `npm run server`.

Please test on macOS and Linux to ensure the fix doesn't introduce any regressions for non-Windows users.
## Root Cause

Git's default `core.autocrlf=true` on Windows checks out files with CRLF line endings. The Linux container cannot execute shell scripts with CRLF, causing the entrypoint to fail with the cryptic error above.

## Solution

This PR implements a two-layer fix:

1. **Immediate fix (Dockerfile):** Strip carriage returns during Docker build using `sed`
2. **Prevention (.gitattributes):** Enforce LF line endings for shell scripts across all platforms

## Impact

- ✅ Fixes the issue for Windows users
- ✅ Backward compatible (no impact on Linux/Mac users)
- ✅ Prevents future occurrences via `.gitattributes`

## Testing

Tested on Windows 10 with Docker Desktop - Grafana now starts successfully.

### DIFF
--- a/.config/Dockerfile
+++ b/.config/Dockerfile
@@ -50,5 +50,5 @@ RUN sed -i 's|</body>|<script src="http://localhost:35729/livereload.js"></scrip
 
 
 COPY entrypoint.sh /entrypoint.sh
-RUN chmod +x /entrypoint.sh
+RUN sed -i 's/\r$//' /entrypoint.sh && chmod +x /entrypoint.sh
 ENTRYPOINT ["/entrypoint.sh"]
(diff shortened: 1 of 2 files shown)


tomglenn (approved):
LGTM, can confirm Docker still working on Mac OS. Thanks!

Jayclifford345:
Hey team, this awesome thank you! I shall accept this one for now but we will need to check to see if the plugin upgrader auto overwrites this as these files are generated by them.

PR #365, opened 2025-11-27
Add noop interactive action

jdbaldry:
Perform no interactive action for the step in an interactive sequence.

This creates an instructional step that displays content without Show me or Do it buttons. Useful for manual steps that users must complete themselves, or for providing context between automated steps.

It means you could have steps that go.
1. ONE
1. TWO
1. THREE (non interactive)
1. FOUR
1. FIVE

Instead of steps that go like:

1. ONE
1. TWO

Three

1. FOUR
1. FIVE


Jayclifford345:
aaah really interesting @jdbaldry, does this work with the current sequential manager? basically disregards to the step to unlock the step after?

Jayclifford345 (approved):
Tested and LGTM!

PR #394, opened 2025-12-02
Fix a bunch of title case based upon the new instructions 

jdbaldry:
There's probably a lot more.


robbymilo (approved):
My brain thanks you :)

PR #473, opened 2026-01-19
Replace Markdown parsers and renderers with appropriate packages

jdbaldry:
Grafana already has a mechanism for rendering (and therefore parsing Markdown). `renderMarkdown` function from `@grafana/data` uses the battle-tested Marked JS library. This PR uses that to render Markdown blocks.

The other instance of Markdown parsing is in the Tiptap rich text editor. Tiptap has experimental support for Markdown and it also uses Marked under the hood. I've configured it to be as similar as I can to the `renderMarkdown` function from `@grafana/data`.

You can experiment with the renderer and see what is and isn't supported using the following Markdown:

````

These are the elements outlined in John Gruber’s original design document. All Markdown applications support these elements.

**bold text**

*italicized text*

> blockquote

1. First item
2. Second item
3. Third item

- First item
- Second item
- Third item

`code`

---

[Markdown Guide](https://www.markdownguide.org)

![alt text](https://www.markdownguide.org/assets/images/tux.png)

These elements extend the basic syntax by adding additional features. Not all Markdown applications support these elements.

| Syntax | Description |
| ----------- | ----------- |
| Header | Title |
| Paragraph | Text |

```
{
  "firstName": "John",
  "lastName": "Smith",
  "age": 25
}
```

Here's a sentence with a footnote. [^1]

[^1]: This is the footnote.

term
: definition

~~The world is flat.~~

- [x] Write the press release
- [ ] Update the website
- [ ] Contact the media

That is so funny! :joy:

(See also [Copying and Pasting Emoji](https://www.markdownguide.org/extended-syntax/#copying-and-pasting-emoji))

I need to highlight these ==very important words==.

H~2~O

X^2^
````

### DIFF
--- a/src/docs-retrieval/components/interactive/interactive-quiz.tsx
+++ b/src/docs-retrieval/components/interactive/interactive-quiz.tsx
@@ -5,7 +5,6 @@ import { GrafanaTheme2 } from '@grafana/data';
 
 import { useStepChecker } from '../../../requirements-manager';
 import { reportAppInteraction, UserInteraction } from '../../../lib/analytics';
-import { ParsedElement } from '../../content.types';
 import { testIds } from '../../../components/testIds';
 
 // ============ Types ============


tomglenn (approved):
LGTM!

PR #482, opened 2026-01-21
feat(dev-tools): update dom selector logic

Jayclifford345:
## Summary

Enhances the selector generator to make selectors more stable.

**Key changes:**

- **Centralized configuration** - Extract all magic numbers and repeated arrays into `SELECTOR_CONFIG` and `STABILITY_SCORES` constants for easier tuning
- **New selector strategies** - Add placeholder and title attribute strategies for better input/icon button selection
- **Sibling-based context** - New `findSiblingContext()` function handles form inputs that follow labels
- **Narrower testid preference** - Refactored parent context search to prefer closer testid ancestors (two-pass algorithm)
- **Selector caching** - WeakMap-based cache with 5-second TTL to avoid regenerating selectors for the same element
- **Quality metrics** - `getSelectorInfo()` now returns `stabilityScore` (0-100) and `warnings[]` for fragile selectors
- **State attribute exclusion** - Exclude dynamic attributes (`data-state`, `data-focus`, `data-hover`, etc.) from compound selectors
- **Helper functions** - Added `tryUniqueSelector()`, `tryUniqueSelectorWithFallback()`, and `getTestIdAttr()` to reduce code duplication

### DIFF
--- a/src/lib/dom/selector-generator.ts
+++ b/src/lib/dom/selector-generator.ts
@@ -16,7 +16,7 @@
  */
 
 import { findButtonByText } from './dom-utils';
-import { querySelectorAllEnhanced } from './enhanced-selector';
+import { querySelectorAllEnhanced, querySelectorAllEnhancedVisible } from './enhanced-selector';
 
 // ============================================================================
 // Configuration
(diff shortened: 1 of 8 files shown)


src/lib/dom/selector-generator.ts
```diff
@@ -18,6 +18,128 @@
...
+ */
+function setCachedSelector(element: HTMLElement, selector: string): void {
+  selectorCache.set(element, { selector, timestamp: Date.now() });
+}
+
+/**
+ * Clear the selector cache. Useful for testing.
+ * Note: WeakMap doesn't have a clear() method, so we create a new one.
+ */
+export function clearSelectorCache(): void {
```
moxious: dead code

PR #535, opened 2026-02-05
Tell agents to to run `test:ci` not `test` script

jdbaldry:
The `test` script runs continuously with `--watch` and therefore always hits a command execution timeout for agents.


moxious (approved):
yeah I've run into that too, thanks for fixing

PR #569, opened 2026-02-11
Render content.json first if available for all docs

jdbaldry:
Let's us start to make use of the blocks anywhere in the docs and importantly unlocks the interactivity in learning paths (formerly known as learning journeys) at https://grafana.com/docs/learning-journeys/.


tomglenn (approved):
lgtm!

PR #574, opened 2026-02-13
feat: add page redirect after deep link sidebar launch

moxious:
## Problem

Deep links to Pathfinder guides (e.g., `/a/grafana-pathfinder-app?doc=bundled:first-dashboard`) must route through the plugin page so the plugin can intercept the request, open the sidebar, and launch the correct guide. But the plugin page itself has no meaningful content — once the sidebar opens, the center console stays blank. The user lands on what is effectively a dead page, with the guide running in the sidebar but nothing useful in the main view.

In short: the deep link _has_ to go to the plugin page for the plugin to hijack the click and open the sidebar, but once that's done, the user is staring at an empty center console with no content. Since the user isn't going "anywhere in particular," they should land somewhere useful — like the home screen or the page the guide is designed for.

## Solution

Add a `page` query parameter to the deep link system so that after the sidebar launches a guide, the center console automatically navigates to a relevant Grafana page.

**Example:** `?doc=bundled:first-dashboard&page=/explore` opens the sidebar with the "Create your first dashboard" guide while navigating to the Explore page.

When no explicit `page` param is provided, the system falls back to the bundled guide's default target page (from `index.json`), or `/` (home) as a final default.

### How it works

1. Plugin reads both `doc` and `page` query parameters during `init()`
2. The `page` value is validated through `validateRedirectPath()` — a new security utility
3. Once the sidebar fires `auto-launch-complete`, `locationService.replace()` navigates to the validated target
4. Using `replace()` instead of `push()` prevents back-button loops (user won't navigate back to the plugin page and re-trigger the deep link)
5. Invalid `doc` parameters now redirect to `/` and open the sidebar with Recommendations, instead of leaving the user on a blank plugin page

### Security: redirect path validation (`page` parameter)

The `page` parameter is user-supplied input that controls a navigation target. `validateRedirectPath()` enforces strict validation:

| Vector | Mitigation |
|---|---|
| Open redirect (`https://evil.com`) | Rejected: doesn't start with `/` |
| Protocol-relative (`//evil.com`) | Rejected: starts with `//` |
| Path traversal (`/../etc/passwd`) | URL API normalizes to safe local path (`/etc/passwd` — a harmless client-side route) |
| Encoded traversal (`/%2e%2e/`) | URL API decodes and normalizes before check |
| Query injection (`/path?admin=true`) | Only `pathname` used; query silently dropped |
| Fragment injection (`/path#x`) | Only `pathname` used; fragment silently dropped |
| `javascript:` / `data:` URIs | Rejected: don't start with `/` |
| Internal path abuse (`/admin/...`) | Same-origin paths respect Grafana's existing auth — accepted by design |

The `page` parameter is **only processed when `doc` is also present**, preventing the plugin page from becoming a general-purpose open redirector.

### Security: guide entry validation (`doc` parameter)

The existing `findDocPage()` function validates `doc` values through strict allowlists:

- **Bundled IDs** (`bundled:first-dashboard`): exact match lookup in `index.json` — no dynamic loading
- **Interactive learning URLs**: hostname must be in `ALLOWED_INTERACTIVE_LEARNING_HOSTNAMES` (exact match)
- **Grafana docs URLs**: hostname must be in `ALLOWED_GRAFANA_DOCS_HOSTNAMES` (exact match), path must contain valid docs prefixes
- **Invalid doc + valid page**: redirects to `/` regardless of `page` value — no amplification

### Combined attack scenarios

- **`page` without `doc`**: `redirectTarget` is `null`; no redirect occurs
- **Invalid `doc` with valid `page`**: redirects to `/`, ignores `page`
- **Timing**: redirect fires in `auto-launch-complete` callback; if sidebar fails to mount, event never fires, no redirect happens

## Test plan

- [x] 14 new unit tests for `validateRedirectPath` covering valid paths and attack vectors
- [x] All 1613 existing tests pass
- [x] Zero linter errors
- [ ] Manual: verify sidebar persists after `locationService.replace()` navigation
- [ ] Manual: verify back button does not re-trigger deep link
- [ ] Manual: `?doc=bundled:first-dashboard` alone redirects to guide's default target
- [ ] Manual: `?doc=bundled:first-dashboard&page=/explore` redirects to `/explore`
- [ ] Manual: invalid `doc` value redirects to `/` with sidebar open
- [ ] Manual: `?page=/explore` alone (no `doc`) is ignored entirely


moxious:
Tested and verified scenarios:

## Exercise Redirect Options

http://localhost:3000/a/grafana-pathfinder-app?doc=bundled:welcome-to-grafana

(Opens that guide, redirects to / on the instance)

http://localhost:3000/a/grafana-pathfinder-app?doc=bundled:welcome-to-grafana&page=/alerting

(Opens that guide, goes to /alerting)

http://localhost:3000/a/grafana-pathfinder-app?doc=bundled:welcome-to-grafana&page=https://grafana.com/

(Opens that guide, goes to home page, because invalid URL)

## Exercise doc options

http://localhost:3000/a/grafana-pathfinder-app?doc=bundled:nonexistent-guide

(Opens at home screen, plugin window open, recommendations pane showing because non-existant guide)

http://localhost:3000/a/grafana-pathfinder-app?doc=bundled:../../../etc/passwd

(Hostile guide location, same result as above)

http://localhost:3000/a/grafana-pathfinder-app?doc=https://interactive-learning.grafana.net/guides/first-dashboard

(Valid guide, works)

## Other Patterns that fail security checks

* ?doc=https://evil.com/docs/grafana/latest/ — wrong domain
* ?doc=https://evil.com/bundled:first-dashboard — doesn't start with bundled:
* ?doc=javascript:alert(1) — not a recognized format
* ?doc=data:text/html,<script>alert(1)</script> — data URI
* ?doc= — empty string, returns null at the guard
* ?doc= — whitespace only, returns null at the trim check
* ?doc=/admin/settings — starts with / but not /docs/ or /tutorials/
* ?doc=https://github.com/grafana/grafana/blob/main/README.md — not an allowed domain
* ?doc=//evil.com/docs/grafana/latest/ — doesn't match any case pattern

Jayclifford345 (approved):

PR #790, opened 2026-04-27
fix: restore sidebar tabs after toggle off/on

### LINKED ISSUE #782: "Pop-Out" sidebar bug, disappearing/dismissing guide, tab state
Reproduction steps:

1. Go to play.grafana.org
2. Click first available recommendation (any, doesn't matter)
3. Guide opens correctly
4. Toggle sidebar off using ?
5. Toggle sidebar back on using ? again.   Originally opened guide is no longer visible

Original discovery path (this path produces same results)

1. Go to this docs page: https://grafana.com/docs/grafana-cloud/visualizations/
2. Click "Read this in Grafana Cloud" and follow the prompts to get to your correct stack
3. For me, the end target URL was: `/?doc=%2Fdocs%2Fgrafana-cloud%2Fvisualizations%2F` which issues a redirect which eventually lands at `/a/grafana-setupguide-app/home`
4. Click the ? help menu again.  Sidebar toggles closed, makes sense.
7. Click the ? help menu again.  Sidebar toggles open (as expected) _but the guide is now dismissed_ and we're back to recommendations. What I expected: no amount of clicking the ? menu should dismiss any active tab.
8. Further -- if you then close that browser tab and come back to your instance via another means (or via step 1 a second time) -- you'll get 2 open tabs.  

There seems to be a bug in the way that the state of tabs is synchronized; when coming in via `/?doc=whatever` something about state isn't being set correctly.  State is going into local storage (as evidenced by step 8, it's still there) but toggle closed/toggle open isn't displaying the state correctly when coming in specifically through this path.  There may be a broader issue with

moxious:
## Summary

- **Bug:** Toggling the sidebar off and on via the help icon dismissed the active guide, showing "recommendations" instead of the user's open tabs (#782)
- **Root cause:** `_hasRestoredTabs` was a static (class-level) flag on `CombinedLearningJourneyPanel` that persisted across instances. When the sidebar remounted and created a new panel via useMemo, `restoreTabsAsync()` bailed out because the flag was still true from the previous mount.
- **Fix:** Changed `_hasRestoredTabs` from static to instance-level. Each new panel instance starts with the guard unset, so sidebar reopen restores tabs correctly. React StrictMode double-mount protection is preserved (both effect calls operate on the same cached useMemo instance). Removed the now-unnecessary `resetTabRestorationGuard()` static method and its 3 call sites.

## Test plan

- [x] 4 new unit tests in `docs-panel.tab-restore-guard.test.ts`:
  - First instance restores tabs normally
  - Same instance blocks double-restore (StrictMode protection)
  - New instance (sidebar toggle) restores independently
  - Without fix, new instance is stuck on "recommendations"
- [x] All 2409 existing Jest tests pass (123 suites)
- [x] TypeScript, ESLint, and Prettier checks pass
- [x] Manual: open a guide, toggle sidebar off via help icon, toggle back on — guide should still be visible
- [x] Manual: open a guide, pop-out to floating panel, dock back to sidebar — guide should persist

Fixes #782

### DIFF
--- a/docs/developer/bugfix-patterns.md
+++ b/docs/developer/bugfix-patterns.md
@@ -1,0 +1,7 @@
+# Bugfix patterns
+
+Patterns from resolved robustness issues. Loaded during `/bugfix` Phase 2 to inform classification and root cause hypotheses.
+
+| Issue | Category  | Root cause pattern                                                                              | Files                                                      | Date       |
+| ----- | --------- | ----------------------------------------------------------------------------------------------- | ---------------------------------------------------------- | ---------- |
+| #782  | State bug | Static class flag survives across instances — use instance-level guards for per-lifecycle state | docs-panel.tsx, ContextPanel.tsx, FloatingPanelManager.tsx | 2026-04-27 |
--- a/src/components/App/ContextPanel.tsx
+++ b/src/components/App/ContextPanel.tsx
@@ -27,7 +27,6 @@ export default function MemoizedContextPanel() {
   useEffect(() => {
     if (mode === 'floating') {
       panelModeManager.restoreSidebarTabSnapshot();
-      CombinedLearningJourneyPanel.resetTabRestorationGuard();
       panelModeManager.setMode('sidebar');
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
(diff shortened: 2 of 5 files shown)


moxious:
@cursoragent review

Jayclifford345 (approved):

PR #793, opened 2026-04-27
fix: prevent double skip buttons in guided blocks (#786)

### LINKED ISSUE #786: Double skip button options
I was reviewing a LP milestone in Pathfinder and noticed that there was no option to skip the step, but also I _could_ still skip it and move on to the next step (the option to `fix this` was the only option). So I decided to add `"skippable": true` to the JSON

Image

Then, after making the commit, the `skip` option showed up in two different places (both buttons work) 

Image

This happened in https://github.com/grafana/interactive-tutorials/pull/101

moxious:
## Summary

- When a guided block has both `block.skippable: true` (block-level) and a step with `isSkippable: true` (step-level), two skip buttons appeared simultaneously: one in the React step card and one in the floating DOM overlay comment box.
- Root cause: `executeStep` called `setIsExecuting(true)` but entered the execution loop immediately, before React committed the state update that hides the idle-state skip button. The guided handler could then synchronously append its DOM overlay (with its own skip button) while the idle skip button was still rendered.
- Fix: add `await waitForReactUpdates()` between `setIsExecuting(true)` and the execution loop in `interactive-guided.tsx`, ensuring React flushes first.

## Test plan

- [ ] New test in `interactive-guided.test.tsx` verifies `waitForReactUpdates` is called **before** `executeGuidedStep` (order assertion)
- [ ] Test verifies the idle-state skip button is gone once execution begins
- [ ] Test verifies only one skip button is rendered in idle state when both `skippable` levels are set
- [ ] Full test suite: 123 suites / 2408 tests — all passing

Fixes #786

---

> [!NOTE]
> **Low Risk**
> Low risk: adds a single awaited React-flush before guided step execution and a focused regression test; behavior change is limited to guided-step start timing and UI rendering order.
> 
> **Overview**
> Prevents a race where starting a guided block could briefly render **two** “Skip” buttons (idle card + guided overlay) when both block-level `skippable` and step-level `isSkippable` are enabled.
> 
> `InteractiveGuided.executeStep` now `await`s `waitForReactUpdates()` immediately after setting `isExecuting`, ensuring the idle UI unmounts before any guided DOM overlay is created.
> 
> Adds a regression test (`interactive-guided.test.tsx`) asserting the idle skip button disappears once execution begins and verifying call order (`waitForReactUpdates` before `executeGuidedStep`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ee7f61b162bb2d81f895b37c8663ed176335f4f0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


moxious:
@cursoragent review

Jayclifford345 (approved):

PR #796, opened 2026-04-27
Fix nested nav reveal for guided steps

### LINKED ISSUE #781: Nested nav items aren't consistently revealed
Given this first interactive step:

```
        {
          "type": "interactive",
          "action": "highlight",
          "reftarget": "a[data-testid='data-testid Nav menu item'][href='/alerting']",
          "content": "Click **Alerting** in the main menu.",
          "requirements": ["navmenu-open"]
        },
```

This is referring to a nav item that's nested underneath of alerting.  When the Nav is visible but that section isn't expanded, the action fails.  We'll need some general way of expanding nav to ensure that the right items are visible so they have a chance to work.

moxious:
## Summary
- Expands parent navigation sections before guided steps resolve Grafana nav item selectors.
- Adds a regression test for collapsed nested nav targets in guided mode.
- Records the #781 bugfix pattern alongside the latest mainline entries.

## Test plan
- `npx jest src/interactive-engine/action-handlers/guided-handler.test.ts --no-coverage`
- Previously run: `npm run test:ci` (passed on rerun)
- Previously run: `npm run test:go`
- Note: `npm run check` could not complete locally because `golangci-lint` is not installed on PATH.

Fixes #781

---

> [!NOTE]
> **Medium Risk**
> Adds pre-resolution navigation expansion for guided steps based on selector parsing, which can change guided-step behavior and trigger navigation side effects if selectors are mis-detected.
> 
> **Overview**
> Guided mode now attempts to **expand a parent navigation section before locating a step target**, when the `refTarget` matches a Grafana nav item selector (extracting the `href` via `resolveSelector` + regex and calling `navigationManager.expandParentNavigationSection`).
> 
> Adds a regression test covering a collapsed nested nav link scenario, asserting the parent section is expanded and the newly revealed link can be highlighted and interacted with.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b5c77f303fe915aef402379d62f95c47d0d3d1f1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


moxious:
Scope note for reviewers:

This fix is intentionally limited to nav items where the parent section can be inferred from the target URL path. For example, a target like `/alerting/list` implies a parent nav section at `/alerting`, so the guided handler can ask the navigation manager to expand that parent before resolving the target selector.

It will not necessarily work automatically for submenu structures where the visible parent/child relationship does not match URL ancestry, or where submenus use entirely different URL paths. This PR also does not attempt to solve arbitrarily nested nav items; it handles the current one-parent expansion case and leaves deeper/more irregular navigation structures out of scope.

Issue #745, opened 2026-04-08
Plugin error on going to "next" module in a learning path


moxious:
This state was triggered by clicking "Next" on a learning path accessed from "My Learning" home page.  Mechanism unknown, needs attempt to replicate or investigation.

<img width="1505" height="727" alt="Image" src="https://github.com/user-attachments/assets/db32922e-1a54-4baf-8bcb-fa27b23cd056" />

I tried to replicate and was unable; suggest looking into those error logs when we can; it looks like HTTP errors that are inelegantly handled somehow/somewhere.

moxious:
## Research investigation by AI agent

### TL;DR

Bug status: **PARTIALLY FIXED by #798** — the root navigation error (#744) that shared this code area is fixed, but whether the specific *plugin error modal* from this screenshot is gone is **CANNOT DETERMINE with certainty** (issue was not reproducible at time of filing; mechanism differs from #744).

---

### Files involved

| File | Role |
|---|---|
| `src/learning-paths/fetch-path-guides.ts` | Fetches `index.json` for URL-based paths and builds per-guide metadata |
| `src/learning-paths/learning-paths.hook.ts` | `useLearningPaths` — `getPathGuides`, `getGuideUrlForPath`, `resolveGuideMetadata` |
| `src/learning-paths/useNextLearningAction.ts` | `computeNextAction` — picks the guideUrl for the UserProfileBar "Next" button |
| `src/components/LearningPaths/MyLearningTab.tsx` (lines 295-349) | `handleOpenGuide` — dispatches the URL when user clicks "Continue" on My Learning |
| `src/components/LearningPaths/LearningPathsPanel.tsx` | `handleContinue` — **dead code** (component is not imported or mounted anywhere) |
| `src/components/docs-panel/docs-panel.tsx` (lines 241-383) | `openLearningJourney` + `loadTabContent` — creates the tab and fetches content |
| `src/docs-retrieval/content-fetcher.ts` (lines 111-175, 270-383) | HTTP fetch pipeline; HTTP errors are caught and returned as structured results, not thrown |
| `src/components/docs-panel/components/MyLearningErrorBoundary.tsx` | Error boundary wrapping `MyLearningTab` in HomePanel (catches render crashes within MyLearningTab) |
| `src/components/App/App.tsx` (lines 27-62) | `PluginErrorBoundary` — the outermost catch-all that renders the screenshot's error UI |

---

### Root cause hypothesis

[...]

PR #798, opened 2026-04-27
fix: open the next module on My Learning "Continue" for URL-based paths (#744)

moxious:
## Summary

Fixes #744. On the My Learning page, clicking "Continue" on a partially-completed URL-based learning path opened the path's base URL (= first module) instead of the per-guide URL the card already labelled "Next:".

**Root cause:** `fetchPathGuides` parsed each item's `relpermalink` only to derive a slug, then discarded the absolute URL. `GuideMetadataEntry.url` was never populated for URL-based paths, so the three navigation call sites (`MyLearningTab.handleOpenGuide`, `LearningPathsPanel.handleContinue`, `computeNextAction` in `useNextLearningAction.ts`) all fell back to `parentPath.url` / `targetPath.url`, which is the path base / first module.

**Fix (minimal, fallback-preserving):**

- `fetch-path-guides.ts` now stores the absolute per-guide URL (`origin + relpermalink`) on each metadata entry, mirroring `fetchLearningJourneyMetadataFromJson` in `docs-retrieval/content-fetcher.ts`.
- `learning-paths.hook.ts`:
  - `resolveGuideMetadata` accepts an optional `pathId` and scopes dynamic metadata to that path so two paths sharing a guide slug don't bleed URLs/titles across each other.
  - `getPathGuides` populates the new `PathGuide.url` from path-scoped metadata.
  - New `getGuideUrlForPath(guideId, pathId)` helper exposed on the hook so handlers receiving `(guideId, pathId)` can resolve the correct URL without a lookup loop.
- `useNextLearningAction.computeNextAction` prefers `currentGuide.url` over `targetPath.url` before falling back to bundled lookup.
- `MyLearningTab.handleOpenGuide` and `LearningPathsPanel.handleContinue` use `getGuideUrlForPath`, falling back to `parentPath.url` only when dynamic data has not loaded yet.

The fallback chain (resolved per-guide URL → path base URL → static metadata → bundled) preserves existing behaviour for static or partially-loaded paths.

**Architectural note:** changes stay within tier 2 (`learning-paths`, `types`) and tier 4 (`components`). No new cross-tier imports.

**Related:** #745 ("Plugin error on going to next module") may live in the same area but emphasises HTTP/API failures; likely a separate bug — worth re-checking after this lands.

## Commits

Two-commit structure required by `/bugfix` (test → fix), plus a docs entry:

- `test:` adds 4 failing tests (`fetch-path-guides.test.ts` × 2, `useNextLearningAction.test.ts` × 3, with one of those covering path-scoping for slug collisions across paths via the existing `currentGuide.url` channel).
- `fix:` the implementation described above.
- `docs:` records the bugfix pattern in `docs/developer/bugfix-patterns.md`.

## Test plan

- [x] `npx jest src/learning-paths/fetch-path-guides.test.ts src/learning-paths/useNextLearningAction.test.ts` — 4 new tests fail before fix, all 23 pass after.
- [x] `npm run typecheck` — clean.
- [x] `npm run lint` — clean.
- [x] `npm run prettier-test` — clean.
- [x] `npm run test:ci` — 2,439 tests pass, 18 skipped, 0 failed (no regressions).
- [ ] Manual verification on a Grafana dev instance: complete the first module of a URL-based learning path (e.g. Linux Server Integration), reload, click "Continue" on My Learning — should open module 2 (`/install-alloy/`), not the path base URL.


moxious:
@cursoragent review

Jayclifford345 (approved):

PR #802, opened 2026-04-28
feat: online package recommendations for OSS recommender-disabled mode

Jayclifford345:
## Summary

When the online recommender is disabled (the default for OSS Pathfinder), users were limited to the 7 bundled guides shipped with the plugin. This PR adds a paired-down discovery path so OSS users with internet access can see CDN-authored guides — without enabling the full recommender — and can navigate them with the same rich rendering they get when the recommender is on.

### Discovery
- New Go endpoint `GET /package-recommendations` proxies the public CDN's `repository.json`, host-allowlisted to `interactive-learning.grafana.net` (and dev/ops siblings). Process-lifetime cache: refreshes at most once per 6h, sticky-503 on failure (no upstream retry within the window).
- The handler also fetches each package's `manifest.json` in parallel (concurrency 8, 5s per fetch, 256KB cap each) and inlines it into the response. Per-manifest failures are skipped silently — the entry stays discoverable but loses rich rendering.
- Untargeted entries stay in the response so milestone / recommends / suggests IDs from learning paths can be resolved by ID. The frontend's matcher drops them from the recommendation list.

### Frontend
- New client at `src/lib/package-recommendations-client.ts` calls the endpoint **only when** `!isRecommenderEnabled(pluginConfig)` AND `navigator.onLine !== false` AND no prior session failure. A `window 'online'` listener clears the sticky-failure flag so transient offline states recover, but the client never proactively retries — air-gapped installs make at most one attempt per session.
- `ContextService.fetchRecommendations` extends the recommender-disabled branch only, merging online matches into the bundled list. Filtering reuses the existing low-weight `matchesUrlPrefix` and `matchesPlatform` helpers (`urlPrefix`, `urlPrefixIn`, `targetPlatform`, plus `and`/`or`).
- Match expressions referencing predicates outside the supported set (`urlRegex`, `datasource`, `cohort`, `tag`, ...) are now **rejected at the entry level** — the base matchers' fall-through-to-true behaviour was surfacing entries like `assistant-self-hosted` (`urlRegex: "^/?$"`) on every page on OSS.
- Recommendations emit `type: 'package'` with a normalized manifest, plus `contentUrl`/`manifestUrl` built with proper slash handling so trailing-slash paths don't produce `//content.json`.

### Resolution & rendering
- New `OnlineCdnPackageResolver` registered as the third tier of the composite resolver when the online recommender is disabled. Without it, `resolvePackageMilestones` couldn't resolve bare IDs from CDN learning paths and dropped every step as "unresolvable", leaving paths with no milestone list and no next-step navigation.
- The resolver looks up bare IDs against the cached package index and returns CDN URLs + the inlined manifest, so `metadata-only` resolution avoids extra round-trips.

### Robustness
- Upstream fetch is now wrapped with `context.WithoutCancel` so a user closing the panel mid-fetch no longer poisons the 6-hour cache with a "context canceled" error. (Caught by Cursor Bugbot.)

The feature is auto-disabled when the recommender is enabled (the branch isn't taken) and when the browser is offline.

## Verification

- `npm run check` (typecheck, lint, prettier, lint:go, test:go, test:ci): green; full suite is **2,460 frontend tests** + Go suite.
- New unit tests:
  - `src/lib/package-recommendations-client.test.ts` — 8 cases (offline, sticky, online-reset, dedupe, ...)
  - `src/context-engine/context.service.online-packages.test.ts` — 9 cases (manifest normalization, type=path routing, urlRegex rejection, double-slash URL building, ...)
  - `src/package-engine/online-cdn-resolver.test.ts` — 3 cases (id lookup, not-found, slash handling)
  - `pkg/plugin/package_recommendations_test.go` — 11 cases (200/503/405, sticky-on-failure, manifest enrichment, slash normalization, request-cancellation detachment, ...)
- E2E: `tests/online-package-recommendations.spec.ts` (3 cases) verified against `npm run server` — endpoint serves 341 real CDN packages with inlined manifests, rejects non-GET with 405, docs panel triggers the call when T&C is unaccepted.
- Manual end-to-end: confirmed that path-typed packages now render the rich learning-journey card with milestone list and next-step navigation when the recommender is off.

## Out of scope (explicit non-goals)

- No new plugin setting — CDN URL is hardcoded against the existing `ALLOWED_INTERACTIVE_LEARNING_HOSTNAMES` allowlist.
- No port of recommender's `MatchExpr` beyond URL-prefix + platform.
- No featured/sponsored content, no scoring.
- No persistent (cross-session) cache.

## Test plan

- [ ] Verify on a fresh OSS Grafana install (`acceptedTermsAndConditions = false`) that navigating to `/dashboards`, `/connections`, `/explore`, `/alerting` shows CDN-sourced guide titles in the help panel.
- [ ] Verify enabling T&C suppresses the network call (no hit to `/package-recommendations` in the network tab).
- [ ] Verify with DevTools "Offline" mode that zero calls go to the endpoint.
- [ ] Verify after blocking `interactive-learning.grafana.net` in DevTools that exactly one call is made, then no further calls for the session, and bundled-only results are shown.
- [ ] Open a path-typed package (e.g. `Visualize GitHub data in Grafana Cloud`) and confirm the milestone list renders with next-step arrows.
- [ ] Confirm the assistant-self-hosted card no longer appears outside `/connections` or `/plugins/grafana-assistant-app`.

---

> [!NOTE]
> **Medium Risk**
> Adds a new backend resource endpoint that fetches/caches data from an external CDN and wires it into recommendation and package-resolution flows; correctness and safety depend on allowlisting, matching rules, and caching behavior under network failures.
> 
> **Overview**
> **OSS recommender-disabled mode can now surface online (CDN) guides/learning paths** in addition to the bundled set.
> 
> Adds a new backend resource `GET /package-recommendations` that proxies the public `repository.json` from `interactive-learning.grafana*.net`, enforces an HTTPS hostname allowlist, caches results for 6 hours with single-flight refresh + sticky failure behavior, and opportunistically inlines each package’s `manifest.json` via bounded parallel fetches.
> 
> On the frontend, introduces `fetchOnlinePackageRecommendations` (offline-aware, session-cached, sticky-unavailable) and updates `ContextService` to merge matching online packages into the recommender-disabled recommendations list, including URL normalization, manifest allowlist/normalization, and fail-closed targeting checks for unsupported predicates or unconstrained matches. Package resolution is extended with an `OnlineCdnPackageResolver` tier (used only when recommender is disabled) so CDN learning-journey dependencies resolve by ID. Docs are updated with a generated **Terms and conditions** page plus a sync script/check added to `npm run check`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 982fa899342fd45d1cc48e675352ab01f06c783a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/sources/_index.md
+++ b/docs/sources/_index.md
@@ -27,6 +27,10 @@ cards:
       description: Learn about Interactive learning administrators features.
       href: administrators-reference/
       height: 24
+    - title: Terms and conditions
+      description: Data usage notice for context-aware recommendations.
+      href: terms-and-conditions/
+      height: 24
     - title: Upgrade notes
       description: Important information about upgrading, including breaking changes.
       href: upgrade-notes/
--- a/docs/sources/administrators-reference/_index.md
+++ b/docs/sources/administrators-reference/_index.md
@@ -33,7 +33,7 @@ When disabled, Interactive learning only displays bundled examples and static re
 
 When you enable the recommender service, Interactive learning sends contextual data to Grafana's hosted recommendation service.
 The service uses this data to provide personalized documentation recommendations.
-For more information about data collection and privacy, refer to the data usage details in the configuration page and the [Architecture](/docs/grafana/latest/pathfinder/architecture/) documentation.
+For more information about data collection and privacy, refer to the [Terms and conditions](../terms-and-conditions/) page and the [Architecture](/docs/grafana/latest/pathfinder/architecture/) documentation.
 
 {{< admonition type="note" >}}
 The recommender service is enabled by default for Grafana Cloud instances.
--- a/docs/sources/terms-and-conditions/_index.md
+++ b/docs/sources/terms-and-conditions/_index.md
@@ -1,0 +1,62 @@
+---
+title: Terms and conditions
+menuTitle: Terms and conditions
+description: Data usage notice for Interactive learning's context-aware recommendations.
+weight: 50
+---
+
+<!--
+DO NOT EDIT BY HAND.
+This page is generated from src/components/AppConfig/terms-content.ts
+by scripts/sync-terms-and-conditions.js. Update the source file and run
+`npm run docs:sync-terms` to refresh this page.
+-->
+
+# Terms and conditions
+
+**Version:** 1.1.0
+
+This page is the data usage notice that Interactive learning shows in the plugin configuration when an administrator enables context-aware recommendations. It is reproduced here so it is reviewable outside Grafana.
+
+## Context-aware recommendations
+
+When enabled, Interactive learning sends contextual data from your Grafana instance to the recommendation service which returns personalized documentation recommendations.
+
+### Data collection and usage
+
+When you enable the context-aware recommendations, Interactive learning collects the following information:
+
+- **Current page path and URL** - to identify which Grafana feature you're using.
+- **A list of the types of installed data sources** - to recommend relevant data source documentation.
+- **Dashboard information** - including dashboard titles, tags, and folder information when you're viewing dashboards. Interactive learning processes this information locally and doesn't send it to the recommendation service
+- **Visualization types** - when creating or editing panels.
+- **User role** - your organizational role, such as Admin, Editor, or Viewer.
+- **Grafana instance type** - whether you are using Grafana Cloud, Grafana Enterprise, or open source Grafana.
+- **User identifier and email** - for Grafana Cloud, a hashed user identifier and email address used for personalization. For open source Grafana, only a generic identifier ('oss-user') and email ('oss-user@example.com') are used. All user data is hashed using SHA-256 before transmission for privacy protection.
+
+### How Grafana uses your data
+
+- **Personalized recommendations** - to provide documentation and learning paths that are contextually relevant
+- **Service improvement** - to enhance the quality and accuracy of recommendations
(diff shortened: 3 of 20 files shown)


moxious:
@cursoragent review PR; use your normal rules.  Extra feedback I would like on this PR includes: 

* an analysis against the architectural tests included in the repo (tier 1, tier 2, etc) and an analysis of how the code introduced impacts that architectural gating. 
* Comments on coupling to the pathfinder package design & repo implementation strategy; is coupling appropriate?  What change, if introduced to the package design *wouldn't* flow through due to the design of this code and could silently break OSS releases in the future?  Are there appropriate tests in place which capture these code paths and prevent such breakage as we evolve the app?
* Check for filtering of potential package recommendations for platform: OSS.  the intent of this PR is to extend more interactive content to OSS without the recommender, but many of the things that would be delivered by the package repo are targetPlatform: cloud, and those should not be delivered to OSS because they target features OSS won't have.
* Check the exact wording of user agreement for enabling the recommender.  The reason the recommender is opt-in is so users have control over their data. if they're fetching a package repo endpoint, does the language we disclose to users comport with reasonable expectations in that scenario?  Why or why not?

PR #806, opened 2026-04-28
fix: disable plugin config button in recommender banner for non-admins

### LINKED ISSUE #626: Insufficient permissions yields strange behavior on plugin settings
This menu in the plugin that navigates to plugin settings:   (this is key because it's how users get to the point where they can enable "dev mode" and access the block editor)

Image

**This fails if the user doesn't have permissions to configure plugins**.  There are a few instances where users that are "read only".  Trying to go to the settings page results in a weird redirect that the plugin can't control.  The redirect results from the user trying to access a page they don't have permissions on.

You'll know this is the case if you go to this part of Grafana and don't see "Plugins" listed:

Image

This ticket is about:

1. Determining what roles this applies to
2. Check inside of the plugin what roles the user has
3. If they aren't eligible for plugin settings, we need to just disable the menu item, perhaps with a tooltip "You don't have enough permissions to access this action"

It won't change the outcome for the user, but it will be far less confusing.

moxious:
## Summary

- `EnableRecommenderBanner` showed a 'Go to plugin configuration' button to all users regardless of role, causing a confusing Grafana redirect for Viewer/Editor users who lack the required plugin settings permission
- Added the same `isGrafanaAdmin || orgRole === 'Admin'` permission guard already used in `TabBarActions` — consistent with Grafana's own backend enforcement
- Button is now `disabled` with a tooltip for unauthorized users; Admins and Grafana Admins are unaffected

Fixes #626

## Test plan

- [ ] 7 new unit tests in `EnableRecommenderBanner.test.tsx` — Admin/Grafana Admin: button enabled + navigates; Editor/Viewer: button disabled + no navigation
- [ ] Manually verify: log in as a Viewer or Editor user and confirm the banner button is disabled with the tooltip 'You don\'t have permission to access plugin settings'
- [ ] Manually verify: log in as an Org Admin and confirm the button still navigates to the recommendations config page as expected


moxious:
@cursoragent review pr

PR #810, opened 2026-04-29
feat(recovery): prompted implied 0th step for guide launches

Jayclifford345:
## Summary

Phase 1 of the auto-recovery design ([`docs/design/AUTORECOVERY_DESIGN.md`](docs/design/AUTORECOVERY_DESIGN.md)) — the prompted "implied 0th step". When a guide is launched from a non-aligned surface (home page, `?doc=` URL param, command palette, link interception, etc.) and its declared `startingLocation` differs from the user's current path, render a one-click navigation prompt above the guide content. While the prompt is up, step 1's requirement checker is paused so it can't race the user's redirect decision with its own "Fix this" or "waiting for element" states.

### What's in the PR

**Recovery primitives** (`src/recovery/`):
- `evaluateAlignment` — pure decision: aligned / no-startingLocation / source-skipped / mismatch
- `pathMatchesStartingLocation` — segment-prefix match with trailing-slash normalization (avoids the `/connections` matching `/connections-new` substring trap)
- `resolveStartingLocation` — manifest first, falls back to `bundled-interactives/index.json` `url[0]` for unmigrated guides; supports both `bundled:<id>` (legacy) and `bundled:<id>/content.json` (package format)
- `isAlignedByConstruction` — source classifier with explicit aligned/needs-check sets

**Tab + model wiring** (`src/components/docs-panel/docs-panel.tsx`):
- `pendingAlignment` field on `LearningJourneyTab`
- `_recordAutoLaunchSource` / `_consumeAutoLaunchSource` (consume-once) to thread the launch source from the listener through to `loadDocsTabContent` without changing the public open signatures
- `confirmAlignment(tabId)` — fires telemetry, navigates via `locationService.push`, clears state
- `dismissAlignment(tabId)` — clears state without navigating
- `reevaluateAlignment(tabId, currentPath)` — reactive re-check on location changes (see below)

**UI** (`src/components/docs-panel/components/AlignmentPrompt.tsx` + `interactive-section.tsx`):
- `<AlignmentPrompt>` renders as a Grafana `<Alert>` banner above `<ContentRenderer>` with Confirm / Continue here buttons (sentence case throughout)
- Inline section hint: every interactive section renders a "📍 Steps are paused. This guide starts on `/xxx` — use the prompt above to navigate." banner while a prompt is up — visible even when the user has scrolled past the top banner

**Step-checker pause via context** (`src/global-state/alignment-pending-context.ts`):
- New Tier 1 React context exposing `{ isPending, startingLocation }`
- `useStepChecker` reads `isPending` and gates `isEligibleForChecking` — single change captures all 7 step-checker callers (interactive-step, code-block-step, interactive-guided, terminal-step, interactive-section, interactive-quiz, interactive-multi-step) without per-component plumbing
- `useAlignmentStartingLocation` exposes the path so sections can render the inline hint

**Reactive re-prompt**: both panels subscribe to `locationService.getHistory().listen` and call `reevaluateAlignment` for the active tab whenever location changes — gated by `hasInteractiveProgress`. Once any step completes, the user is treated as "in" the guide and we stop second-guessing their location. Re-prompts use `launch_source: 'location_change'` for analytics distinction.

**Source propagation**: `source` field added to all 6 guide-launch dispatchers (HomePanel, link interception, queued links, content links, grot guide block, MCP launch).

**Floating panel parity**: same prompt rendering, context provider, and reactive re-prompt subscription.

**Telemetry** (`src/lib/analytics.ts`): three new `UserInteraction` events — `AlignmentPromptShown` / `Confirmed` / `Dismissed` with `guide_url`, `guide_title`, `launch_source`, `current_path`, `starting_location`, plus `latency_ms` on Confirmed/Dismissed.

### Source classification (v1)

| | Sources |
|---|---|
| **Aligned by construction (skip prompt)** | `recommender`, `browser_restore`, `mcp_launch`, `navigate-action`, `grot_guide_block`, `experiment_treatment`, `experiment_treatment_navigation`, `auto_open`, `floating_panel_dock` |
| **Needs alignment check** | `home_page`, `url_param`, `command_palette*`, `external_suggestion`, `link_interception`, `queued_link`, `content_link`, `location_change` |

Unknown sources default to "needs check" (safe — at worst a redundant prompt the user dismisses).

### Bug fixes bundled in this PR

- Drain the leaked launch source on the learning-journey path (LJ never consumes `_pendingLaunchSource`, so without the explicit drain it would contaminate the next docs open)
- Tag restored tabs as `browser_restore` (otherwise initializeRestoredActiveTab would treat the restore as a fresh launch and prompt against the user's already-active guide)
- Tighten `pathMatchesStartingLocation` to segment-prefix matching with trailing-slash normalization (the original substring rule false-positived `/connections-new` against `/connections`)
- Strip `/content.json` suffix before bundled-index lookup (package-format URLs were silently missing their fallback)
- Remove dead `navmenu-open` fallback in the navigation fix-handler (now that requirements thread through `dispatchFix` cleanly)

## Test plan

- [ ] **Home page** → click any tile while on `/explore` → prompt appears above the guide content with the starting path; section hint visible inside step 1; step 1's "Fix this" does NOT fire → click Navigate → location changes → step 1 mounts cleanly
- [ ] **Home page** → click tile → Continue here → step 1 mounts, on-page Fix this fires for guides that need it (parity with today)
- [ ] **`?doc=`** → visit `/a/grafana-pathfinder-app/?doc=bundled:loki-grafana-101` from `/explore` → prompt appears
- [ ] **Command palette** → cmd-K → "Open Interactive learning" → click a tile → prompt appears
- [ ] **MCP** → `sidebarState.openWithGuide('loki-grafana-101')` from `/dashboards` → NO prompt
- [ ] **Recommender** → visit `/connections`, click recommendation → NO prompt (bypasses event entirely)
- [ ] **Browser restore** → dock the panel, refresh → NO prompt for restored tabs
- [ ] **Already aligned** → from `/connections`, launch any `/connections` guide → NO prompt (path matches)
- [ ] **Reactive re-prompt** → launch a guide from a misaligned surface, dismiss the prompt, then navigate via Grafana nav to a different page → banner reappears with `launch_source: 'location_change'`
- [ ] **Reactive clears** → navigate from a misaligned path back into the starting location → banner disappears silently
- [ ] **Re-prompt gate** → complete step 1, then navigate away → banner does NOT reappear (user is "in" the guide)
- [ ] **Section hint** → with prompt up, scroll past the top banner → inline hint visible inside the first interactive section explaining why steps are paused
- [ ] **Floating panel** → switch to floating mode, repeat home page launch → prompt + section hint render in the floating panel; reactive re-prompt works there too
- [ ] **Remote URL without manifest** → no prompt (resolver returns null, existing on-page Fix this is the safety net)
- [ ] `npm run check` passes (typecheck + lint + prettier + go lint + go test + frontend tests)

## Test coverage added

| Test file | What it covers |
|---|---|
| `src/recovery/alignment-evaluator.test.ts` | Pure evaluator: aligned (exact + segment prefix), source-skipped, no-startingLocation, unknown-source default, substring false-positive regression, trailing slash normalization, root special-case |
| `src/recovery/launch-sources.test.ts` | Source classifier coverage for known sources + unknown default |
| `src/recovery/starting-location.test.ts` | Manifest, bundled (array + string), missing-bundled, no-fallback, defensive non-string, **package-format URL suffix stripping** |
| `src/components/docs-panel/components/AlignmentPrompt.test.tsx` | Renders, callbacks, testIds |
| `src/components/docs-panel/docs-panel.alignment.test.ts` | Model: pendingAlignment set/cleared per source/manifest, restored tabs tagged `browser_restore`, source-leak drain, confirm/dismiss/reevaluate methods |
| `src/requirements-manager/step-checker.hook.test.ts` | New `AlignmentPendingContext gate` describe block: paused = no checkRequirements; default = normal behavior |

## Out of scope (follow-up PRs)

- Checkpoint capture/storage (next PR — Phase 1 groundwork for divergence monitor and modal recovery)
- Continuous URL/UI divergence monitoring beyond no-progress reactive re-prompt (Phase 3)
- Modal recovery (Phase 2)
- Async manifest fetch for remote-without-manifest URLs
- Learning-path next-guide transitions (no `endingLocation` field exists yet)
- Silent (Level 0) variant — Level 1 only for now; revisit silent as a per-guide opt-in once telemetry shows confirm rates

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes docs-panel/floating-panel navigation flow and step-checker eligibility gating, which could affect guide execution and tab restore/reload behavior across multiple launch surfaces.
> 
> **Overview**
> Adds a prompted *“implied 0th step”* alignment flow: when launching a guide whose `startingLocation` doesn’t match the current route (and the launch source isn’t *aligned-by-construction*), the panel stores `pendingAlignment` on the tab and renders a new `AlignmentPrompt` banner with Navigate/Continue actions.
> 
> Introduces recovery primitives in `src/recovery/` to resolve `startingLocation`, classify/validate `LaunchSource`, and evaluate path alignment, then threads the source through guide opens via new `openDocsPage`/`openLearningJourney` options plus a consume-once `_recordAutoLaunchSource` fallback for internal reload/restore paths.
> 
> Adds reactive alignment re-evaluation (new `useAlignmentReevaluation`) gated by “no interactive progress”, a global `AlignmentPendingContext` that pauses `useStepChecker` while alignment is pending, and an inline “steps are paused” hint in interactive sections; includes floating-panel parity, source propagation updates to all auto-open dispatchers, and new telemetry events for shown/confirmed/dismissed prompts with expanded test coverage.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5a4563a1ee825bd1f6168fc92b041487db2b53e7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Jayclifford345:
@cursoragent review

moxious:
@cursoragent review.   Relative to the original design, let's think about this pr in terms of state machine design.  There are two state machines: the main area of the application including nav, and the right hand pathfinder sidebar. This means that there are at least four combinations. One is when the two state machines match. Two is when the state machine of the main app is ahead of the Pathfinder sidebar. Three is when the Pathfinder sidebar is ahead of the main nav. And four is when they are both completely out of sync. There may be other combinations that you can think of. I want you to use those as a lens to evaluate this PR and ask systematically how the PR addresses all of those different state machine combinations. There is a risk that as we introduce this auto recovery design that the code will become very complex and so I'm also looking for a principal engineer review about to what extent this code can be simplified and still retain the same functionality. The last lens I would like you to use to evaluate this is that in future phases of the auto recovery design, what primitives that we introduce here can readily be reused, versus in future phases, how much net new code do we need to keep developing?

moxious:
@cursoragent review PR.  Use your included PR review skill and go hard on this one.

This is foundational work for an important next step in the overall robustness of pathfinder. For a moment I want you to adopt a devil's advocate position and criticize the code about how it will cause breakages in the future. Based on the strength of your criticisms, I will be able to see certain things about the code. Make a case that the author of the PR should have done things differently, that the code would be more maintainable if it was done in another way, perhaps that too much complexity is introduced, or look for situations where subtle bugs can creep into other subsystems of Pathfinder due to the introduction of this code. In addition to your standard review, I want an adversarial review and a list of all the reasons where this will go wrong.

PR #846, opened 2026-05-06
feat: full-screen mode + real PR-tester learning journeys

Jayclifford345:
## Summary

Adds a third presentation surface — **full-screen mode** — alongside the existing sidebar and floating panel, and overhauls the PR tester so it opens path/journey packages as **real learning journeys** (cover page + milestone toolbar + Alt + ←/→ navigation) instead of fabricating one giant guide of section blocks.

```mermaid
flowchart LR
    Sidebar -- pop out --> Floating
    Sidebar -- full screen --> Fullscreen
    Floating -- dock --> Sidebar
    Floating -- full screen --> Fullscreen
    Fullscreen -- exit --> Sidebar
    Fullscreen -- pop out --> Floating
```

## Full-screen mode

- New `?panelMode=fullscreen` route (`/a/<plugin>/fullscreen`) with its own `CombinedLearningJourneyPanel` instance, driven by a new scene page (`fullScreenPage`) and `FullScreenLayout` / `FullScreenPanel` components.
- `panelMode` extended with `'fullscreen'`; sidebar/floating/fullscreen hand off via `setPendingGuide` and a snapshot/restore flow on `tabStorage`.
- **`PendingGuide` now carries `packageInfo`** so synthetic packages whose URL is not a recognised package URL (the PR-tester case) survive every transition with their milestone toolbar intact.
- `BlockEditor` toolbar takes an explicit `surface` prop so it stays in sync with the actual host instead of trusting the (sometimes stale) global `panelModeManager`.
- Module init handles `?panelMode=fullscreen` and `?type=` so deep links / share URLs rehydrate the right surface and tab kind.

## PR tester: real package journeys

- Detects `path` / `journey` `manifest.json` files in the PR, builds a real `PackageOpenInfo` (manifest + pre-resolved milestones whose URLs point at PR raw files), and opens it through the same pipeline the recommender uses for production paths.
- Milestones are resolved via each child manifest's `id` field, **not** the directory name, so authors can use prefix-numbered directories (e.g. `01-where-we-are/`) while the manifest declares the canonical package id (`pathfinder-roadmap-where-we-are`).
- `github-api.ts` broadened to fetch `manifest.json` alongside `content.json`; new `fetchPrManifest` helper parses with the loose `ManifestJsonObjectSchema`.
- New pure helpers `indexContentByPackageId` and `buildPathPackageInfo` with full unit coverage including the real-world prefix case.
- Removed the old `bundled:pr-tests/` synthetic-guide path in `fetchBundledInteractive` and the drag-to-reorder UI / state / storage in `PrTester.tsx`.
- `PrTesterProps` simplified: `onOpenLearningJourney` dropped; `onOpenDocsPage` now accepts `packageInfo` so the devtools surface forwards it to `model.openDocsPage`.

## Cover-page rendering fix

`injectJourneyExtrasIntoJsonGuide` previously collapsed every cover-page markdown block through `simpleMarkdownToHtml`, which only handles headings/lists/links and silently dropped `**bold**` / `*italic*`. Now markdown blocks are preserved as-is so the downstream `parseMarkdownToElements` renderer handles full markdown; the cover-page extras (Ready to Begin button + bottom navigation) are appended as a single trailing HTML block.

## Other changes

- New `package-info-from-url.ts` derives a manifest for deep-link package URLs that arrive without recommender context.
- Analytics + launch-source taxonomy gain `fullscreen_handoff`, `floating_panel_dock`, etc. so cross-surface flows are attributable.
- Tests added for `panel-mode` `packageInfo` round-trip, manifest detection in github-api, and the new package-id index helper.

## Test plan

- [x] `npm run check` — typecheck + lint + prettier + lint:go + test:go + test:ci (158 suites / 2992 tests)
- [ ] Manual: pop out a journey from the sidebar to floating; verify the milestone toolbar persists.
- [ ] Manual: switch a journey from sidebar → fullscreen → floating → sidebar; verify the milestone toolbar and current milestone persist on every hop.
- [ ] Manual (PR tester): paste https://github.com/grafana/interactive-tutorials/pull/305 (a path package PR with prefixed child directories), pick "Learning path", verify all 8 milestones resolve, open as a real journey with cover, milestone toolbar, and Alt+arrow navigation.
- [ ] Manual: refresh on `/fullscreen?doc=…&type=learning-journey` and confirm the journey rehydrates with the milestone toolbar.
- [ ] Manual: cover page with `**bold**` / `*italic*` markdown renders correctly.

---

> [!NOTE]
> **Medium Risk**
> Adds a new full-screen navigation surface and changes cross-surface handoff/tab restoration logic, which can affect routing, persistence, and milestone navigation across sidebar/floating/fullscreen modes.
> 
> **Overview**
> Adds a new **full-screen** presentation surface by registering `fullScreenPage` in the `SceneApp`, extending `panelModeManager` with `fullscreen`, and wiring a new `pathfinder-request-full-screen` flow that hands off the active tab (including `packageInfo`) and navigates to the `/fullscreen` route while preventing sidebar/fullscreen double-mount drift via `FullScreenModeNotice` and updated tab-restore guards.
> 
> Refactors shared UI/behavior: extracts `LearningJourneyMilestoneToolbar` and `PanelModeActionButtons` so sidebar/fullscreen milestone nav + actions stay consistent, updates the block editor header to be sticky/collapsible and adds a Full screen CTA.
> 
> Overhauls **PR Tester** to treat `manifest.json` as first-class: GitHub PR fetching now returns `content` + `manifest` entries, preloads/parses manifests with abort-safe timeouts, builds real path/journey `PackageOpenInfo` via new `pr-path-package` helpers, and opens paths through `onOpenDocsPage(..., packageInfo)` (removing the synthetic drag-to-reorder learning-path guide). Also updates floating-panel link sharing/handling to preserve guide type in share URLs and enable full-screen switching.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bc7af13df1b3cf33ef4ecdc4514c669d5a4ba218. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tomglenn (changes requested):
Looks good on the whole. Getting a few issues though:

- Entering full screen mode and then leaving it again with the "Return to sidebar" button always puts users at the My Learning page for pathfinder. It should remember where they were previously.
- Clicking Do It on an interactive step inside a learning journey does the action, pops pathfinder back into the sidebar, but puts you back at page one of the learning journey, and doesn't always mark the interactive step as actually completed (i assume some race condition on the checker)
- Clicking Resume section in a partially completed section while in fullscreen will pop it back into the sidebar, but none of the interactive steps will mark themselves as complete despite them actually being actioned in sequence. More concerning here is that the steps still show as not available yet (complete previous step first) but are still being actioned by the automated process.
- I see a "4/7" label in the top right of the fullscreen mode and I'm not sure what this is. I thought it was a visual of how many steps are complete, but completing another or resetting one doesn't update that label? This looks to be broken if i switch to popout mode too but I can't tell where the bug is originating without a deeper dive.
- The copy link button has a tooltip that says "copy workshop link", needs reworded as workshop is misleading.
- We should probably make the top bar sticky so that users don't lose a way to "get back" to sidebar mode when they're halfway through a guide.

moxious:
On merge, we should close this other PR which I spiked a while ago.  It's so similar, there's definitely no case for keeping both, and this one is further along so the other one can die:  https://github.com/grafana/grafana-pathfinder-app/pull/779

moxious:
@cursoragent review PR -- in addition to the concerns-based review of this, I want you to use sub-agents to look into the following issues: 

1. identify if there's a 'deep link' option.  I.e. I need the ability to link to main-screen content, e.g. something like `/a/grafana-pathfinder-app/fullscreen?doc=whatever`.  Take a look at the URL params that are possible on /a/grafana-pathfinder-app that permit "deep linking to specific guides" and ensure that they are maintained/refactored into some kind of a common module so that we can treat all paths similarly and not need to re-implement different URL routing params per-viewtype.
2. General code reuse between the new custom `fullscreen` route and other routes which render documentation.  Duplication of logic/code paths is not desirable; ideally we would want to have some "core logic" about how to intercept and render a guide and then reuse it in different contexts to prevent differences from accumulating.
3. In the diagram on this PR, the one step I question is "Fullscreen -> Pop out".  This may be excessively complex, do you think we need it?  Thing is, because full screen content can't interact with the main UI without stepping on itself (e.g. a nav action would dismiss the full screen content) -- my thinking has been that full screen content would only be for non-interactive content.  In which case "popping out" full screen content would not be necessary.  Additionally, full screen is good for space-rich stuff like wide videos or diagrams that don't fit well in a sidebar, and so "popping out" that kind of content would just make it look worse.  Might we simplify things by not permitting the "pop out" of full screen content?  Why should or shouldn't we remove that one linkage?
4. Conduct a pre-mortem on this PR.  Let's assume the PR fails and introduces bugs and UX problems that users don't like.  When this PR fails in a few months time and we end up backing the code out, provide an analysis of what went wrong.  That way we can think this through very carefully and mitigate conceptual issues.

tomglenn (approved):
LGTM!

PR #858, opened 2026-05-08
fix(block-editor): remove duplicate title in preview mode (#850)

### LINKED ISSUE #850: Duplicated titles in Block editor preview mode
Specifically in block editor preview (not viewing a completed guide, just preview) there is a title duplication that occurs.

To replicate, create a new guide titled "Hello World" and preview it.  You should see an outer preview frame that announces "Hello world" and underneath of it the actual previewed guide, which begins with the same rendered title.

To make matters worse, some guide authors start their guides with a markdown heading:

```
# Hello world
```

Which effectively makes a "Triple redundant heading" in preview mode.  

The scope of this ticket is to ferret out the differences between how guides appear in preview mode vs. normal render mode, and look for source of duplicated titles.  Don't worry about the inner markdown heading; that's an artifact of what's actually in the guide rather than how it's rendered.

moxious:
Fixes #850.

## Summary
- Preview rendered the guide title twice: once as the editable title input in `BlockEditorHeader`, and again as the `<h1>` emitted by `ContentRenderer` for native JSON content. Production rendering only shows the in-content `<h1>`, so the editor's title input was redundant chrome that diverged from real-product appearance.
- Hide the title input in `BlockEditorHeader` when `viewMode === 'preview'`, so the rendered guide is the only title visible — matching how the guide actually appears in product.
- Move the **Reset guide** button from inside `BlockPreview`'s content area into the header actions cluster (gated on preview mode + live progress) so the rendered guide stays free of editor chrome.
- Extract progress/reset state into a shared `useGuidePreviewProgress` hook; the header and `BlockPreview` stay in sync via the existing `interactive-progress-saved` / `interactive-progress-cleared` events.
- `BlockEditorContent` passes `hideResetButton` to the full-guide preview. Inline pinned block previews from `BlockList` still render their own inline reset button.
- `showTitle` is preserved — `BlockList` still uses it to suppress the guide title when previewing a single non-section block out of context.

## What was deliberately left for follow-up
The original analysis surfaced several other structural differences between preview and production rendering. This PR is a first pass on the title duplication only. Follow-ups worth considering:
- Drop the `previewContent` padding/margin overrides so spacing matches `journeyContentHtml`.
- Restructure the preview container so warnings/reset chrome live outside the scroll region (mirroring `inner-docs-content` in `docs-panel.tsx`).
- Provide an `AlignmentPendingContext.Provider` in `BlockPreview` so location-aware interactives behave the same way as in production.

## Test plan
- [ ] Create a guide titled "Hello World" and switch to preview — expect a single `<h1>` from the rendered content; no editable title input above it.
- [ ] In preview, run an interactive step ("Show me" / "Do it") and confirm the **Reset guide** button appears in the header actions cluster.
- [ ] Click **Reset guide** in the header — confirm progress clears, ContentRenderer remounts, and the button disappears.
- [ ] Switch back to **Edit** — confirm the title input reappears and is editable.
- [ ] Inline pinned block previews in edit mode (eye icon on a block) still show their own reset button when applicable.
- [ ] `npm run check` passes.


moxious:
@cursoragent review

Jayclifford345 (dismissed):

Jayclifford345:
LGTM!

Jayclifford345 (approved):

PR #871, opened 2026-05-12
fix(sections): number all content blocks in a section (#841)

### LINKED ISSUE #841: Number all steps inside of sections irrespective of interactive/non-interactive
PROBLEM:  At present, Pathfinder applies numbered list formatting to each subsequent block within a section, but it has an undesirable quirk: if a block is non-interactive (e.g. type markdown) then the numbered styling is not rendered.  This creates two negative effects: indentation looks too far to the left (because numbered lists are naturally indented) and numbering is broken.  On the website, a set of steps renders like this:

1. Step 1 Interactive
2. Step 2 Non-interactive
3. Step 3 Interactive

And inside of the UI, it renders like this:

```
       1. Interactive
   Non-interactive
       2. Interactive
```

INTENT:  The scope of this ticket is to ensure that all blocks irrespective of type are sequentially numbered.  This is true **only within sections**.  Numbered list formatting is not to be applied outside of pathfinder sections, to avoid front-matter explanatory text being rendered as a numbered list, which is **not** wanted.

At present, some authors are using `no-op` "interactive" blocks as a work-around.  This is undesirable, and fools pathfinder into thinking the step is interactive (hence rendering a numbered item) but creates other undesirable effects.  Authors are encouraged not to use "no op workarounds", which after this ticket is closed, will no longer be necessary in any case.

moxious:
## Summary

Closes #841.

Sections now number every content block sequentially, regardless of whether it's interactive. Previously, only `.interactive-step / .interactive-multistep / .interactive-guided` participated in the section's CSS counter, so a markdown block between two interactive steps broke the `1. 2. 3.` sequence and rendered flush-left. Authors had been using no-op interactive blocks as a workaround.

## What changed

- **`interactive-section.tsx`** -- children of the section's `<ol>` are wrapped in an `<li>`, with `data-numbered="true"` set on content blocks. The wrapping logic is exposed as `wrapSectionChildrenForNumbering` and `shouldNumberSectionChild` for test access.
- **`interactive.styles.ts`** -- CSS counter selector changed from a class allowlist to `& > li[data-numbered="true"]`. The `::before` step number and `paddingLeft` now live on the wrapper `<li>`.
- **`docs-retrieval/index.ts`** -- re-export `ImageRenderer`, `VideoRenderer`, `YouTubeVideoRenderer` from the barrel (consistent with the existing `CodeBlock` re-export, no new barrel-bypass entries).
- **`section-numbering.test.tsx`** (new, 20 cases) -- covers the denylist, numbered types (including markdown and code samples), and the wrapper's DOM output.

## Numbering rules

| Block kind                                                                                                                          | Numbered? |
| ----------------------------------------------------------------------------------------------------------------------------------- | --------- |
| markdown, code-block, interactive, multistep, guided, quiz, input, terminal, terminal-connect, html, assistant, grot-guide, raw HTML | **Yes**   |
| image, video, youtube-video, conditional                                                                                            | No        |

## Notes

- Component-identity lookup in `shouldNumberSectionChild` happens at call time, not module-init, to avoid a circular-import cycle through the docs-retrieval barrel (which re-exports `content-renderer.tsx`, which imports back into `components/interactive-tutorial/`).
- `li.interactive` dead CSS in `interactive.styles.ts:280-287` is **not** touched here -- it belongs to a larger legacy `getInteractiveSequenceStyles` block (the `<span data-targetaction="sequence">` HTML format whose parser was removed in `html-parser.ts:552-555`). That cleanup deserves its own focused PR.
- After this lands, authors no longer need the no-op interactive workaround. A follow-up could migrate bundled guides off the pattern, but that's author-owned content and out of scope here.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run prettier-test`
- [x] `npm run test:ci` (3300 passed; one unrelated SIGSEGV worker crash in `html-parser.test.ts` that passes on retry)
- [ ] **Visual smoke test in `npm run server`**: open a guide with a mixed-block section (markdown + interactive + code-block) and confirm sequential `1. 2. 3.` numbering with consistent indentation. (Author to verify before un-drafting -- no local Grafana available in this session.)
- [ ] Confirm an existing guide with image/video inside a section still renders the media inline without a number, and the surrounding steps keep their sequence.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/components/interactive-tutorial/interactive-section.tsx
+++ b/src/components/interactive-tutorial/interactive-section.tsx
@@ -50,13 +50,28 @@ export function shouldNumberSectionChild(child: React.ReactNode): boolean {
  * Wrap each section child in an <li>, marking content blocks with
  * data-numbered="true" so CSS can apply sequential numbering. Media and
  * wrapper blocks (image/video/conditional) sit in the list without a number.
+ *
+ * data-step="true"  → React component (InteractiveStep, quiz, terminal, etc.)
+ *                     These carry their own CSS margin-top that naturally aligns
+ *                     the card with the ::before number at top: theme.spacing(2).
+ *                     The <li> needs no extra padding.
+ * data-step="false" → Plain HTML content (markdown <p>, headings, etc.)
+ *                     No built-in top margin, so the <li> gets paddingTop via CSS
+ *                     to push the content start down to match the number position.
  */
 export function wrapSectionChildrenForNumbering(children: React.ReactNode): React.ReactNode {
   return React.Children.map(children, (child, index) => {
     const numbered = shouldNumberSectionChild(child);
     const childKey = React.isValidElement(child) && child.key != null ? child.key : `section-child-${index}`;
+    // React components (interactive steps, quiz, terminal…) have typeof type === 'function'.
+    // Plain HTML elements (p, div, h2…) have typeof type === 'string'.
+    const isStep = React.isValidElement(child) && typeof child.type !== 'string';
     return (
-      <li key={childKey} data-numbered={numbered ? 'true' : undefined}>
+      <li
+        key={childKey}
+        data-numbered={numbered ? 'true' : undefined}
+        data-step={numbered ? String(isStep) : undefined}
+      >
         {child}
       </li>
     );
--- a/src/components/interactive-tutorial/interactive-section.tsx
+++ b/src/components/interactive-tutorial/interactive-section.tsx
@@ -12,6 +12,8 @@ import { InteractiveQuiz, resetQuizCounter } from './interactive-quiz';
 import { TerminalStep, resetTerminalStepCounter } from './terminal-step';
 import { TerminalConnectStep, resetTerminalConnectStepCounter } from './terminal-connect-step';
 import { CodeBlockStep, resetCodeBlockStepCounter } from './code-block-step';
+import { InteractiveConditional } from './interactive-conditional';
+import { ImageRenderer, VideoRenderer, YouTubeVideoRenderer } from '../../docs-retrieval';
 import { reportAppInteraction, UserInteraction, getSourceDocument, calculateStepCompletion } from '../../lib/analytics';
 import { interactiveStepStorage, sectionCollapseStorage, interactiveCompletionStorage } from '../../lib/user-storage';
 import { INTERACTIVE_CONFIG, getInteractiveConfig } from '../../constants/interactive-config';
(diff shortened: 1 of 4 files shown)


moxious:
## Alignment fix — post-merge notes

After the initial commit landed, a visual regression was discovered during manual preview testing: **the number for interactive step rows was aligned correctly, but the number for markdown/plain-HTML rows appeared ~16 px too low** (below the text instead of beside it).

### Root cause

The `::before` number is absolute-positioned at `top: theme.spacing(2)` (16 px). That value was chosen to match the interactive step card's own `margin-top: 16px` — the card's top edge naturally lands at 16 px, so the number lands at exactly the same height. ✓

Plain HTML content (`<p>`, `<div>`, etc.) has no built-in top margin. Its text starts at `top: 0` within the `<li>`, while the number sits at `top: 16 px`. Result: number rendered 16 px below the text. ✗

### Why it was tricky

Two earlier approaches made it worse before arriving at the final fix:

1. **Flexbox rewrite** — converted the `<li>` to `display: flex` so the `::before` and content sat as siblings. The interactive step's `margin-top` then became a flex-item margin rather than a block margin, pushing the card down inside the flex row while the number stayed at the top. Displaced the step visually.

2. **`paddingTop` + `marginTop: 0` override** — added `paddingTop: theme.spacing(2)` to all numbered `<li>` items and tried to suppress the interactive step's margin with `& > .interactive-step { marginTop: 0 }`. The specificity calculation looked fine on paper (nested selector should win), but the step ended up double-spaced anyway — the card was pushed 32 px down while the number stayed at 16 px. The user reported it as "displaced down and to the right."

The key insight: **the interactive step's `margin-top` is load-bearing for alignment**. Any change that interfered with it (flexbox, margin suppression) broke the step rows. The markdown rows needed fixing without touching the step layout at all.

### Final fix (two files)

**`interactive-section.tsx` — `wrapSectionChildrenForNumbering`**

Added a `data-step` attribute to each numbered `<li>`:
- `data-step="true"` → child is a React component (interactive step, quiz, terminal, etc.). `typeof child.type !== 'string'` distinguishes these from plain HTML.
- `data-step="false"` → child is a plain HTML element (`<p>`, `<div>`, `<h2>`, …).

**`interactive.styles.ts`**

Added one targeted rule:

```css
& > li[data-numbered="true"][data-step="false"] {
  padding-top: theme.spacing(2);
}
```

This pushes plain content down 16 px to match where the `::before` number sits — without touching the interactive step layout at all. The step rows continue to use their own `margin-top` for alignment, exactly as before.

All 20 section-numbering tests pass; new assertions added for the `data-step` attribute values.

moxious:
## Horizontal alignment fix — `768c1ae`

Picking up where the previous comment left off: after the vertical alignment fix landed (`paddingTop` on `data-step="false"` rows), a second misalignment remained on the **X axis**. Markdown text rendered ~18 px to the left of the interactive step's title text in the same numbered list — visible in the screenshot as "1. I'm markdown" sitting flush-left while "2. I'm interactive" sat noticeably further right.

### Runtime evidence (DOM measurements)

I instrumented `interactive-section.tsx` with a `useEffect` that posted computed styles + bounding-box deltas for the first three numbered `<li>` elements to a debug log. The before/after comparison:

| Row              | Before `padding-left` | Before `title.x` | After `padding-left` | After `title.x` |
| ---------------- | --------------------- | ---------------- | -------------------- | --------------- |
| li[0] markdown   | 32 px                 | 680              | **50 px**            | **698** ✓       |
| li[1] interactive| 32 px                 | 698              | 32 px (unchanged)    | 698 ✓           |
| li[2] markdown   | 32 px                 | 680              | **50 px**            | **698** ✓       |

### Root cause

Both `<li>` types had identical `padding-left: 32px`, but `.interactive-step` wraps its title in a card that adds **18 px of internal chrome on the left**:
- `padding: theme.spacing(2)` → 16 px on every side, including left
- `border: 2px solid transparent` → 2 px on every side, including left

So a step's title text sits at `li.x + 32 + 2 + 16 = li.x + 50`, while a markdown `<p>` (no chrome) sat at `li.x + 32`. Δ = 18 px.

### Fix

Mirror the prior agent's Y-axis pattern. Add `paddingLeft` to `li[data-numbered="true"][data-step="false"]` equal to the step card's internal chrome, so markdown text starts at the same x as the step's title:

```ts
'& > li[data-numbered="true"][data-step="false"]': {
  paddingTop: theme.spacing(2),
  paddingLeft: `calc(${theme.spacing(4)} + ${theme.spacing(2)} + 2px)`,
},
```

The interactive step's load-bearing margin / padding / border are **untouched**, satisfying the constraint documented in the prior comment (*"the interactive step's margin-top is load-bearing for alignment"*). The same selector-attribute pattern (`data-step="false"`) now handles both axes — symmetric and easy to reason about.

### Why not change the step's chrome instead?

I considered three other approaches before settling on this one:

1. **Remove the step's left padding/border** — breaks the card's visual chrome. Executing/completed states have a colored left border that would disappear.
2. **Negative margin-left on the step** — extends the card 18 px leftward into the number column; the digit `2.` would visually sit on the card.
3. **Scope the chrome change to only `<li>`-nested steps** — same problem as #1, just narrower.

Pushing markdown rightward is the only option that preserves both the card's visual integrity and the number's positioning. It does widen the gap between the number and markdown text from 8 px to 26 px — but that gap now matches the gap between the number and the step's title text (which is 26 px because the card chrome eats 18 px of it), so the section reads consistently.

### Tests

- `section-numbering.test.tsx` extended in the previous commit with assertions on the `data-step` attribute values; all pass.
- All 127 tests in `src/components/interactive-tutorial/` pass.
- `npm run typecheck`, `npm run lint`, `npm run prettier-test` all green.

Co-authored-by: Claude Opus 4.7 (1M context)

moxious:
@cursoragent review PR

Jayclifford345 (approved):

PR #872, opened 2026-05-13
fix(section): gate completion on acknowledgement for trailing passive content (#842)

### LINKED ISSUE #842: Add manual "Mark section Complete" behavior to interactive sections.
PROBLEM:  It looks weird if we have sections where there are 3 interactive steps, followed by non-interactive text, because the section auto-completes when the interactive steps are completed, whether or not the user got a chance to read the non-interactive stuff after it.

INTENT:  We want to create a situation where users must acknowledge having read non-interactive material; this means that 'section complete' styling and logic does still apply to non-interactive sections.

## Considered & Rejected alternatives

(From discussion with @tacole02 )

* We could remove the concept of completion for non-interactive sections. This isn't acceptable because it means that "impossible to complete" sections can be in guides, between other sections that might have been completed.  (Looks weird)
* We could ban the use of non-interactive blocks at the end of the sections to avoid the problem from arising.  This isn't acceptable because it constrains writers and prevents carry-over of material from website
* We could live with the current behavior as-is (auto-completed section as soon as last interactive step is completed).  This was rejected because this is considered a UX bug by docs writers.

## Design 

We want to create a "Mark Section as Complete" button, _replace_ the Reset Section button in certain circumstances.

This mark as complete button is only to appear and be an option in one scenario:

* All interactive steps in the section have been completed
* There are remaining steps that are non-interactive
* In the scenario where there is any existing interactive step that has not yet been executed, the "Mark as complete" button will not appear.
* If a section is composed 100% of non-interactive steps, then the **only** option the user will ever be offered on that section is the "mark as complete" button

Implied consequences of this design, good & bad:
* Users will be blocked and unable to go on to the next section until they click "Mark section as complete" (indicating they've read the non-interactive text).  This is a likely consequence since section 1 must be complete before section 2 is eligible to execute.  Users won't be able to "skip" clicking that button
* All sections will be subject to the same "styled as complete" logic when they're done

Authors are advised to avoid no-op blocks as a work-around.  No-op blocks still count as interactive, and so to keep 3 no-op blocks at the end of a section would effective undo the logic of this change and work against the system design.

moxious:
## Summary

Fixes #842. Interactive sections that ended with non-interactive content (markdown, image, video, html, or `noop` steps) auto-completed the moment the last interactive step finished — hiding trailing content behind auto-collapse before the user could read it. All-passive sections were treated as complete on first render.

This PR introduces an explicit "Mark section as complete" acknowledgement gate, and documents the four-site tracked-step-types registry the fix relies on so future component additions don't silently regress this behaviour.

### Behaviour

A section now requires the user to click **Mark section as complete** when, in document order:

- there is any passive child positioned **after** the last interactive child (the `interactive, interactive, markdown, markdown` case + the original trailing-noop case), **or**
- the section is composed entirely of passive content (the button replaces "Do Section" outright).

Mid-section passive content (`markdown, interactive, interactive` or `interactive, passive, interactive`) does **not** trigger the gate — behaviour is unchanged.

Objectives-based completion (`data-objectives`) still bypasses the gate; the button never appears on objectives-completed sections.

### Implementation

- **Pure analyzer** `analyzeAcknowledgement(kinds[])` in `step-section-utils.ts`, plus `classifySectionChild` in a new `section-child-classifier.ts` that distinguishes the 7 tracked interactive step components from any other rendered child.
- **Persistence** new `sectionAcknowledgementStorage` in `user-storage.ts`. `get()` returns `boolean | null` so absence is distinguishable from explicit `false` — used by the migration path.
- **Migration safety** sections that were already "complete" under the pre-fix rules (all real steps in `completedSteps`) auto-acknowledge on first read so existing finished work doesn't spontaneously become incomplete after upgrade.
- **Reset** `handleResetSection` clears acknowledgement state + storage. `clearAllForContent` and `clearAll` sweep the new prefix.
- **Button** new primary-action button in the existing action slot. Replaces Do / Reset only while acknowledgement is pending. New testId `markSectionCompleteButton`.

### Documentation — four-site registry awareness

The fix relies on `INTERACTIVE_STEP_COMPONENT_TYPES` in `section-child-classifier.ts`. That set is part of a four-site registry that drifts independently (two parse-string sites in `content-renderer.tsx`, two component-identity sites). Forgetting site 4 would mean the new step type is misclassified as passive and the acknowledgement gate triggers spuriously.

- **New rule** `.cursor/rules/tracked-step-types.mdc` — glob-triggered on the natural entry points (`src/components/interactive-tutorial/*.tsx`, `src/docs-retrieval/content-renderer.tsx`, `src/docs-retrieval/json-parser.ts`). Lists all four sites with per-site failure modes and a 10-step checklist.
- **Inline cross-references** — uniform `⚠ TRACKED STEP TYPE REGISTRY — site N of 4` JSDoc blocks at each registry site point to all the others and to the rule.
- **AGENTS.md row** — adds the new rule to the on-demand context table.

## Open questions for review

1. **`CodeBlockStep` classification.** Currently classified as `interactive` in `INTERACTIVE_STEP_COMPONENT_TYPES`. If you'd rather treat "look at this code" as passive content, it's a one-line move.
2. **Catch-all for unknown React elements.** Anything outside the tracked set is classified `passive`. Intentional — covers markdown/image/video/html. The cost: a future interactive step type that isn't added to `INTERACTIVE_STEP_COMPONENT_TYPES` will be misclassified as passive. The new documentation is the mitigation.
3. **All-passive sections require ack on every visit.** Today my code persists ack, so a section is only marked complete once. If you'd prefer auto-acknowledge after some scroll/dwell heuristic (no click required), that's a different design — flag it in review.

## Test plan

- [x] `npm run typecheck` — clean
- [x] `npm run lint` — clean
- [x] `npm run prettier-test` — clean
- [x] `npm run test:ci` — 3327 tests pass across 180 suites (18 skipped, pre-existing)
- [x] `npm run test:go` — passes
- [x] New `analyzeAcknowledgement` covered by 10 unit cases in `step-section-utils.test.ts`
- [x] New `classifySectionChild` covered by 24 unit cases in `section-child-classifier.test.tsx`
- [ ] Manual sanity check in `npm run server`:
  - [ ] Section with trailing markdown shows the Mark button only after interactives done
  - [ ] All-passive section shows only the Mark button
  - [ ] Sections with interior-only passive content behave unchanged
  - [ ] Upgrade migration: a section with `completedSteps` from before the fix stays complete on first load
  - [ ] Reset Section clears acknowledgement and re-opens the gate

## Risk

Low-medium. The change is gated on a derived `needsAcknowledgement` flag — sections that don't match the new pattern are completely unaffected. The migration default-to-acknowledged on first read covers the upgrade path. Failure mode would be a stuck-incomplete section, which is recoverable by clicking the new button. No data loss possible.


moxious:
This PR was closed because it was superceded by another PR which took a different approach.  This PR introduced more issues than it solved, and the conflicts weren't worth resolving.

PR #877, opened 2026-05-13
fix(section): Mark Section Complete gate + state-machine refactor (#842)

### LINKED ISSUE #842: Add manual "Mark section Complete" behavior to interactive sections.

moxious:
## Summary

Fixes #842. Interactive sections that end with non-interactive content auto-completed the moment the last interactive step finished — hiding trailing markdown / images / video behind auto-collapse before the user could read them. All-passive sections were treated as complete on first render.

This PR introduces an explicit **"Mark section as complete"** acknowledgement gate, and rebuilds the section's state machine on a discriminated-union `useReducer` foundation so the entire class of state-asymmetry bugs that killed PR #872 becomes structurally impossible.

## User-visible behavior

The gate fires when, in document order:

- **(a)** the section has any passive child positioned *strictly after* the last interactive step (e.g. `[interactive, markdown]`, `[interactive, interactive, image]`), **or**
- **(b)** the section is composed entirely of passive content (e.g. `[markdown, image]`).

Mid-section passive content sandwiched between interactives does **not** trigger the gate — the user can still act on the trailing interactive. Objectives-completed sections bypass the gate entirely (`doneVia: 'objectives'`). `noop` `InteractiveStep`s count as interactive, per the issue text.

## State machine

```
init ──Do Section / step complete──▶ partial ──last step done──▶ awaiting-ack  (if gate applies)
                                                              └─▶ done(no-gate-needed)
awaiting-ack ──Mark────────────────────────────────────────────▶ done(ack)
awaiting-ack ──step Redo──▶ partial    (ack cleared in lockstep — Bug 1 fix)
done(any) ──step Redo──▶ partial       (ack cleared in lockstep — Bug 1 fix)
done(any) ──Reset Section──▶ init      (storage + ack + event all cleared — Bug 2 fix)
done(any) ──auto-collapse──▶ done-collapsed   (unchanged)
*         ──Objectives met──▶ done(objectives)   (gate bypassed)
```

The `acknowledged` flag lives inside the reducer state — every transition that mutates `completed` mutates `acknowledged` in the same atomic step. There is no representable state where `completed.size === 0 && acknowledged === true`, and the reducer refuses to fire `ACKNOWLEDGE` when `completed.size === 0`. Bug 1 (Redo bypasses the gate, which killed the previous PR) is structurally impossible.

## Why this is structured as 8 commits

Following the [High-Risk Refactor Guidelines](https://github.com/grafana/grafana-pathfinder-app/wiki/High-Risk-Refactor-Guidelines). The wiki's protocol matches the active patterns here: **Pattern G** (async state machine decomposition) + **Pattern J** (contract-surface extraction for storage + testIds + events) + **Pattern I** (module-level registry — `INTERACTIVE_STEP_COMPONENT_TYPES`). Phase ordering is by risk, each phase is an atomic commit with its own pre-test → change → post-test loop, and structural moves are separated from behavior changes:

| # | Commit | What it does | Behavior change? |
|---|---|---|---|
| 0 | `test(section): add baseline characterization tripwires` | Pin current behavior in disposable tests + harness | None |
| 1 | `feat(section): extract section-child-classifier + analyzeAcknowledgement` | Pure utility extraction, no consumers | None |
| 2 | `feat(storage): add sectionAcknowledgementStorage + four-site registry doc` | New storage namespace + .cursor rule, no consumers | None |
| 3 | `fix(section): preview sandbox + reset event + persist-outside-updater` | Three orthogonal bug fixes in the existing structure | Bug 2, Bug 3, Bug 5 |
| 4 | `refactor(section): replace state machine with useReducer` | Structural reducer rewrite, gate inert | None |
| 5 | `feat(section): add Mark Section Complete acknowledgement gate` | Flip the gate on + migration + Mark button | The new feature |
| 6 | `fix(block-editor): preserve ContentRenderer identity when reset button toggles` | Defensive sibling-keying for TEST 2 | Preventive |
| 7 | `test(section): remove disposable baseline tests` | Close-out — disposable baselines removed | None |

## Decisions made along the way

These are the load-bearing design calls. Each is followed by the rationale and the alternative we deliberately turned down.

### 1. Discriminated-union `useReducer` over parallel `useState` slots

The previous attempt (PR #872) added `isAcknowledged` as a parallel boolean alongside `completedSteps`. Every mutator that touched one had to remember to touch the other, and **four** confirmed bugs flowed from that asymmetry — most importantly Bug 1 (Redo bypassed the gate). The original implementation was flagged as "buggy and didn't take into account how the new button affects the state machine of transitions"; the explicit ask was a structural fix rather than a patch.

We collapsed `{ completedSteps, currentStepIndex, isAcknowledged }` into one reducer-owned `{ completed, cursor, acknowledged }` shape and wrote the reducer so transitions atomically maintain the cross-field invariant. **Alternative rejected:** keep the parallel-boolean shape and audit every mutator manually (PR #872's approach). Rejected because every future addition would need the same audit.

### 2. `noop` `InteractiveStep`s are classified as **interactive**, not passive

The issue text says: *"No-op blocks still count as interactive, and so to keep 3 no-op blocks at the end of a section would effectively undo the logic of this change and work against the system design."*

PR #872 classified `noop` as **passive**. We deliberately diverge — noop counts as interactive per the literal issue text. The cost: an author can place trailing noop blocks to bypass the gate. The issue acknowledges this is a known weakness and warns authors against it rather than blocking it at the classifier level.

### 3. Hard preview-mode sandbox for **all** section-scoped storage

The previous code sandboxed `sectionCollapseStorage` in preview but persisted `interactiveStepStorage` and (in PR #872) `sectionAcknowledgementStorage`. This caused stale preview state to leak into the next preview session — and is the strongest hypothesis for the TEST 2 reset-on-Do-Section symptom that killed PR #872's investigation.

We made the sandbox total: in preview mode, **all three** storage namespaces are write- and read-no-op. **Alternative rejected:** session-scoped sandbox (sessionStorage). Rejected because no session-storage layer exists in this plugin and adding one is out of scope.

### 4. Mount-restore is true mount-only via `didRestoreRef`

The previous mount-restore effect had `[sectionId, stepComponents]` in its dep array. Because `stepComponents` recomputes on every parent re-render (`children` prop reference changes), the effect re-fired routinely. With `didRestoreRef`, it fires exactly once per `InteractiveSection` instance — remount of the React tree resets the ref via fresh React state, but a stable instance with churning children does not re-restore.

This kills both reported sub-bugs (Bug 4(a) intent mismatch, Bug 4(b) runtime race) without changing any dep arrays elsewhere.

### 5. Migration policy: auto-acknowledge pre-#842 completed sections on first read

A user with a section they completed under the old rules has `acknowledged === null` in storage (no key). On first mount after upgrade, `restoreFromStorage` detects this and writes `acknowledged = true` once.

**Alternative rejected:** re-prompt — make every prior-completed section ask for acknowledgement again. Rejected because it creates regression-feel: "I already finished this; why is it asking me again?"

### 6. Symmetric `interactive-progress-cleared` event from `handleResetSection`

`useGuidePreviewProgress` listens for both `interactive-progress-saved` (already dispatched) and `interactive-progress-cleared` (only dispatched from `BlockPreview.reset()`, never from the per-section Reset button). So a single-section guide with a per-section reset left the "Reset guide" button stuck visible.

**Alternative rejected:** make `persistCompletedSteps` symmetric (fire `interactive-progress-saved` even on empty→non-empty). Rejected because mid-edit consumers might double-fire on every persist call; the targeted dispatch in `handleResetSection` is the smaller change.

### 7. All-passive sections use a synthetic completion marker

The reducer's `ACKNOWLEDGE` action refuses to fire when `completed.size === 0` (it's the inverse half of the Bug 1 invariant). All-passive sections (zero interactive steps) never accumulate real entries in `completed`, so `handleMarkSectionComplete` injects a marker `${sectionId}::ack-marker` before dispatching `ACKNOWLEDGE`. The marker is filtered out of `deriveSectionState`'s all-passive branch and never visible to the user. Downside: it adds 1 to `interactiveStepStorage.countAllCompleted` for the guide. Cosmetic only.

**Alternative rejected:** weaken the reducer's invariant. Rejected — invariant is what kills Bug 1's class.

### 8. Defensive sibling-keying in `BlockPreview` for TEST 2

The handoff doc's TEST 2 symptom — `Do Section` in block-editor preview occasionally reverts to INIT — was unreproducible in unit tests. The leading hypothesis is React's child-array reconciliation around `<ContentRenderer key={resetKey}>` and its conditional siblings. We added explicit keys to every sibling in `BlockPreview`'s outer container and used a composite key on `ContentRenderer` itself.

**Status:** unverified at the runtime level. The fix is cheap, structural, and well-targeted; if TEST 2 still reproduces after this branch, the root cause lies deeper and the diagnostic anchor from `.cursor/local/SECTION_STATE_MACHINE.md` stays valid.

### 9. Four-site registry doc (`.cursor/rules/tracked-step-types.mdc`)

There are now **four** places in the codebase that need to know "what counts as an interactive step type" — two parse-time (`INTERACTIVE_STEP_TYPES`, `SECTION_TRACKED_STEP_TYPES` in `content-renderer.tsx`) and two runtime (`stepComponents` useMemo branches in `interactive-section.tsx`, `INTERACTIVE_STEP_COMPONENT_TYPES` in `section-child-classifier.ts`). They drift independently because they operate on different value spaces (strings vs component identity). Ported the rule from PR #872 with the noop-is-interactive correction.

## Tests

```
main baseline:    185 suites / 3384 tests / 24.2s
this PR:          189 suites / 3478 tests / 24.0s
```

**Net: +4 suites, +94 tests, ~same runtime.** All baseline + new tests green at every phase checkpoint.

New permanent test files:

- `src/components/interactive-tutorial/section-state.test.ts` — 30 cases, exhaustive reducer + derivation coverage
- `src/components/interactive-tutorial/section-child-classifier.test.tsx` — 18 cases pinning the gate's input layer
- `src/components/interactive-tutorial/interactive-section.state-machine.test.tsx` — 10 end-to-end integration scenarios incl. **Bug 1 regression**
- `src/components/block-editor/hooks/useGuidePreviewProgress.test.ts` — 8 cases pinning the listener side of the BlockPreview ↔ InteractiveSection event contract
- `src/test-utils/interactive-section-harness.tsx` — shared mock surface for the above

Extensions to existing suites: `analyzeAcknowledgement` (+11), `sectionAcknowledgementStorage` (+10).

## Test plan (manual — please reproduce in `npm run server`)

Paste each variant into the block editor → preview and exercise the section. ☐ = unchecked while in draft, please tick on review.

- [x] `[interactive]` — no gate; behaves as before
- [x] `[interactive, markdown]` — gate fires after step; Mark click → DONE
- [x] `[interactive, interactive, markdown]` — gate fires after both; Mark → DONE
- [x] `[interactive, markdown, interactive]` — no gate; mid-section passive does not trigger
- [x] `[markdown, markdown]` — all-passive; only Mark button visible; Mark → DONE
- [x] `[interactive, noop]` — no gate (noop counts as interactive)
- [x] `[interactive]` with `data-objectives` met → DONE, gate bypassed
- [x] Reset section in any DONE state → returns to INIT; gate (if applicable) re-arms
- [x] Auto-collapse fires only after Mark click (non-preview), never in `awaiting-ack`
- [x] **TEST 2** reproduction: paste the JSON from `.cursor/local/SECTION_STATE_MACHINE.md` §1, click `Do Section`, observe whether the section silently resets. If it reproduces, the Phase 6 defensive fix did not catch the root cause — flag for follow-up.
- [x] Migration smoke: install over a Grafana with prior interactive-step progress in localStorage. Existing completed sections stay complete; new gate appears for fresh work.

## What this PR does NOT do

- Does **not** address Bug 6 from the handoff doc (final-sweep blind persist in `handleDoSection`) — flagged for follow-up; currently stable in practice and orthogonal to the gate.
- Does **not** rename existing testIds or storage keys (per refactor-guidelines Pattern J: "Do not rename or clean up contract surfaces during a structural refactor").
- Does **not** add feature flags or kill switches — overhead exceeds risk for this component scope.
- Does **not** redesign `ContentRenderer` or `BlockPreview` reconciliation beyond the minimal sibling-keying fix in Phase 6.

## Risk

Medium. The reducer rewrite touches `interactive-section.tsx` (1900+ lines, load-bearing). The mitigation strategy is the per-phase testing sandwich: each commit is independently verifiable and revertable. The structural commit (#4) ships with the gate inert so externally observable behavior is preserved across it — the actual behavior change is isolated to commit #5. Pre-#842 completed sections auto-acknowledge once on upgrade so no work is lost. No data migrations.

## Open questions for review

1. **`CodeBlockStep` classification.** Currently in the interactive bucket (it has a "paste this in" affordance). If you prefer treating it as passive, it's a one-line move in `section-child-classifier.ts`.
2. **All-passive section persistence.** Today the ack persists, so an all-passive section is only marked complete once. Alternative: auto-acknowledge after a scroll / dwell heuristic with no click. Different design — flag if you want it.
3. **TEST 2.** I could not reproduce it in unit tests on a fresh branch. The Phase 6 fix is defensive. Please attempt reproduction in the block-editor preview and let me know whether the symptom is gone or still present.


moxious:
@cursoragent review PR.

I would consider this branch to be a high-risk refactor because it cleaned up and re-implemented the state machine in a critical part of the application. I believe that this refactor was necessary because in order to add the targeted feature, there was no easy way to do that without creating a much bigger, more unmaintainable mess.                          
                                                         
I do intend to eventually completely refactor the interactive section module which has gotten too large and is now too unwieldy. I expect it to be a source of maintainability problems, but first I just want to add this feature and make sure that the state machine is cleaned up. So I would like for you to evaluate this branch on several different criteria.

1. Does it accomplish the ticket that the PR links to?  Is it sufficient to add the feature?
2. Does the code increase test coverage and hence safety for future evolution of the interactive-section module?
3. Does the code run major risks of regressions that aren't in the test plan identified in the PR?

Apply all of the other standard review criteria from the Cursor PR review instructions found via AGENTS.md

tomglenn:
Running some manual testing, but in the meantime Claude caught some things worth addressing:

---

  **Please fix before merge**

  1. `handleMarkSectionComplete` all-passive branch clobbers existing completions —
  `persistCompletedSteps(new Set([markerId]))` should be `new Set([...completedSteps, markerId])` to mirror
  the reducer's add semantics. Safe today (all-passive ⇒ empty set), but a footgun if a guide is edited to
  drop its interactive steps.
  2. Drop `acknowledged: false` from the type or define when it's set — leaving an unused state value is
  exactly the asymmetry the reducer was built to prevent.

  **Worth addressing**

  3. Phase 6's sibling-keying in `BlockPreview` is defensive without a confirmed repro. Cheap and
  well-targeted, but if TEST 2 reproduces post-merge the root cause is elsewhere and this is dead weight.
  Consider rendering conditional siblings into a fixed-position wrapper so `ContentRenderer`'s slot index
  can't shift regardless of reconciler behavior.
  4. `handleResetSection` storage clears (lines 1488-1490) aren't guarded by `isPreviewMode`. No bug — the
  clears are no-ops in preview because nothing was persisted — but it's three async calls of wasted work and
   a comment-or-guard would document the intent.
  5. `pathfinder-step-progress` effect fires a window event + cached storage scan on every `completedSteps`
  change. Cached so not catastrophic, but worth tracking as step counts grow.
  6. `getResumeInfo()` is invoked twice per render in the catch-all `Button`'s `title` and children IIFEs
  (lines 2079, 2093). Compute once at the top of the JSX.
  7. Pre-existing `console.warn('[Section] …')` scroll debug logging (lines 333, 341-352, 365-373, 404)
  rides along on the touched file. Not introduced here, but ripe for cleanup while we're in the area.

  **Nits**

  - `allInteractiveStepsCompleted` is computed in both `deriveSectionState` and
  `interactive-section.tsx:711-712`. Not strictly redundant (component-level version gates
  `useStepChecker`), but the filter logic will rot if one branch changes.
  - `handleResetSection` `useCallback` deps omit transitively-used helpers. Likely fine (module-level
  singletons), but the dep set should match what's actually closed over.
  - Three unchecked manual scenarios (`[markdown, markdown]`, `[interactive, noop]`, objectives-met) each
  exercise a load-bearing branch — worth ticking before merge.

tomglenn (dismissed):
Manual testing confirms working as intended. Approving as I don't want to block getting this in, but a once over of the Claude review is probably worth doing before merging. :)

moxious:
Thanks for the thorough review @tomglenn. I've landed five commits addressing items #1, #2, #4, #6, and #7:

- `71957d2a` fix(section): preserve prior completions when marking all-passive section — accumulate via `new Set([...completedSteps, markerId])` instead of clobbering
- `114c9b0d` refactor(section): narrow acknowledged to `true | null` — dropped the speculative `false` branch from `SectionState`, the `RESTORE` variant, `restoreFromStorage`, and `sectionAcknowledgementStorage.get()`'s return type
- `cc652765` refactor(section): skip storage clears in preview-mode reset — wrapped the three `.clear()` calls in `if (!isPreviewMode)`; left the `interactive-progress-cleared` event dispatch unguarded so `useGuidePreviewProgress` still clears the "Reset guide" button in preview (Decision 6)
- `b2a10d27` refactor(section): compute resumeInfo once per render — hoisted above the catch-all Button so both IIFEs share one result
- `beb0f7c8` chore(section): drop pre-existing `[Section]` scroll debug logs

Deferred, with reasoning:

**#3 — BlockPreview fixed-position wrapper.** The Phase 6 sibling-keying is cheap defensive code with no confirmed repro of TEST 2. Rewriting `BlockPreview` to render conditional siblings into a fixed-position wrapper is a real refactor; doing it without a repro risks introducing the very bug we're guarding against. If TEST 2 fires post-merge I'll take your suggestion as the next step.

**#5 — `pathfinder-step-progress` event cost.** Agreed worth tracking. The fix would be debouncing the dispatch and/or coalescing the cached storage scan; both interact with the progress-chip update semantics that drive `FullScreenLayout` and `FloatingPanel`. Out of scope for a review-cleanup PR — filing as a follow-up.

**Nit — `allInteractiveStepsCompleted` duplication.** Investigated: the local computation at line 713 feeds `useStepChecker` at line 722, whose output (`isCompletedByObjectives`) feeds the `derived` `useMemo` at line 731. Replacing the local with `derived.allInteractiveStepsCompleted` reverses the dependency direction and breaks hook ordering. You flagged it as "not strictly redundant" — confirmed: it's structural. Leaving as-is rather than adding a comment for noise.

**Nit — `handleResetSection` deps.** Skipped per usual policy of matching ESLint exhaustive-deps output rather than enumerating module-level singletons.

**Manual scenarios.** Will run the three unchecked PR test-plan scenarios (`[markdown, markdown]`, `[interactive, noop]`, objectives-met) in `npm run server` and tick the boxes before merge.

tomglenn (dismissed):

Jayclifford345 (approved):

PR #878, opened 2026-05-14
fix(block-editor): preserve new guide when leaving JSON mode

Jayclifford345:
## Summary

- Fix a stale-closure bug in `useGuideHistory.setState` that caused the block editor to revert to the **old** guide when pasting a fresh JSON guide and switching to **Preview**.
- Same defect also silently corrupted `handleLoadGuideFromBackend` and the persistence onLoad path whenever local state was dirty when the load fired — both now resolved by the same change.

## What changed

`useGuideHistory` now tracks the latest state in a synchronously-updated `latestStateRef`, and `setState` / `undo` / `redo` source `prev` from that ref instead of from closure-captured `state`. Subsequent batched `setState` calls in a single event handler now correctly compose instead of overwriting each other.

- `src/components/block-editor/hooks/useGuideHistory.ts` — the fix.
- `src/components/block-editor/hooks/useGuideHistory.test.ts` — unit regression test.
- `src/components/block-editor/hooks/useJsonModeHandlers.test.ts` — new file, integration test wiring the real editor through the user-facing paste-then-preview flow.

## Why

`handleExitJsonMode` runs three setState calls back-to-back in one event:

```
editor.loadGuide(newGuide);          // object form  -> queues new guide
editor.updateGuideMetadata({});      // function form -> uses STALE closure 'state' (old guide)
editor.setViewMode(targetMode);      // function form -> STALE again
```

Because the wrapped `setState` evaluated function-form updaters against the closure `state` and passed the result as a literal to React's `setStateInternal`, calls 2 and 3 received the old guide as `prev` and silently overwrote the new guide queued by call 1. End result: preview rendered the old guide.

The `latestStateRef` approach makes the wrapper behave the way React's native `setState((prev) => ...)` does for batched updates.

## Test plan

- [x] New unit test `'batched setState calls compose, not overwrite'` failed before the fix, passes after.
- [x] New integration test `'switching from JSON to preview applies the pasted guide'` failed before the fix, passes after.
- [x] Full block-editor suite (404 tests) green.
- [x] `npm run check` green: 186 suites, 3386 tests, 0 failures, plus typecheck / lint / prettier / Go.
- [ ] Manual: open the block editor with a guide loaded -> JSON view -> replace JSON with a different guide -> Preview -> renders the new guide.

## Risk and reversibility

Low risk, fully reversible. The change is contained to `useGuideHistory` internals; the public hook API (`setState` / `undo` / `redo` / `canUndo` / `canRedo` / `resetHistory`) is unchanged. No call sites needed modification — the fix is purely additive correctness for batched updates. Reverting is a single-file rollback.

---

> [!NOTE]
> **Medium Risk**
> Touches the core `useGuideHistory` state wrapper used by the editor, so regressions could affect undo/redo and any batched `setState` flows. The change is localized and covered by new unit + integration regression tests.
> 
> **Overview**
> Fixes a stale-closure issue in `useGuideHistory` where function-form updates evaluated against a closure-captured `state` could overwrite earlier queued updates in the same event (e.g., JSON paste → switch to Preview reverting to the previous guide).
> 
> `useGuideHistory` now keeps a synchronously-updated `latestStateRef` and uses it as the source of truth for `setState`, `undo`, and `redo` (including history entry bookkeeping). Adds a unit regression test for batched composition behavior and a new integration test covering the JSON-mode paste-then-preview flow via `useBlockEditor` + `useJsonModeHandlers`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 83921f3dba5b3c72b89653b813890fa98bcc73c7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/components/block-editor/hooks/useGuideHistory.test.ts
+++ b/src/components/block-editor/hooks/useGuideHistory.test.ts
@@ -210,4 +210,37 @@ describe('useGuideHistory', () => {
 
     expect(result.current.canUndo).toBe(false);
   });
+
+  it('batched setState calls compose, not overwrite (regression: stale-closure prev)', () => {
+    // Reproduces the JSON-paste -> preview bug. When `handleExitJsonMode`
+    // calls `loadGuide(newGuide)` (object form) followed by
+    // `updateGuideMetadata({})` and `setViewMode('preview')` (function
+    // form) inside the same event, every function-form `prev` must reflect
+    // the latest queued state, not a stale closure capture. Otherwise the
+    // second call's `{...prev, ...}` overwrites the new guide with the old
+    // one and the user sees the old guide in preview.
+    const { result } = renderHook(() => useGuideHistory(makeState({ guide: { id: 'old', title: 'Old' } })));
+
+    act(() => {
+      // Step 1: replace whole state (object form) — analogous to loadGuide.
+      result.current.setState(
+        {
+          guide: { id: 'new', title: 'New' },
+          blocks: [],
+          viewMode: 'edit',
+          isDirty: false,
+        },
+        { skipHistory: true }
+      );
+      // Step 2: function-form mutation — analogous to updateGuideMetadata({}).
+      result.current.setState((prev) => ({ ...prev, isDirty: true }));
+      // Step 3: function-form mutation — analogous to setViewMode('preview').
+      result.current.setState((prev) => ({ ...prev, viewMode: 'preview' }), { skipHistory: true });
+    });
+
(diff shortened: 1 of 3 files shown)


tomglenn (dismissed):
LGTM. Pre-emptively approving for once the cursor bugbot issue is resolved. Thanks!

tomglenn (approved):

PR #888, opened 2026-05-15
refactor(mcp): migrate Go MCP tools to TS (MH4) + retire Go MCP (MH5)

### LINKED ISSUE #883: MCP HTTP transport: SIGINT/SIGTERM hangs because server.close() does not close idle keep-alive connections
## Summary

`pathfinder-cli mcp --transport http` does not shut down cleanly on `SIGINT` / `SIGTERM`. The signal handler runs, but the process hangs on `await handle.close()` and never reaches `process.exit(0)`. In local dev you have to `kill -9`; in Cloud Run this means scale-down and rolling deploys wait the full container grace period (then SIGKILL) on every replica termination.

## Repro (~30 seconds)

```bash
node -r ts-node/register/transpile-only src/cli/index.ts mcp \
  --transport http --port 8765 --host 127.0.0.1 &
SERVER_PID=$!

# Open one HTTP request to seed a keep-alive connection.
curl -sS http://127.0.0.1:8765/healthz >/dev/null

# Ask it to shut down.
kill -INT "$SERVER_PID"

# Process is still alive 25+ seconds later.
sleep 25 && kill -0 "$SERVER_PID" && echo "still hung"
```

Observed: `still hung`. Expected: process exits within ~1s of the signal.

## Root cause

`runHttp` in `src/cli/mcp/transports/http.ts` returns a handle whose `close()` only calls `server.close(...)`:

```265:265:src/cli/mcp/transports/http.ts
    close: () => new Promise<void>((resolve, reject) => server.close((err) => (err ? reject(err) : resolve()))),
```

`http.Server#close()` stops accepting **new** connections and waits for **all existing** connections to finish — but it does not proactively close idle keep-alive connections. Any client (curl, MCP client, even a probe that hit `/healthz` once) leaves an idle keep-alive socket behind, and that socket holds shutdown open until either the peer disconnects or `server.keepAliveTimeout` fires (5s in our config). In practice the timer also doesn't reliably fire under `server.close()` because the server has stopped scheduling new I/O on those sockets, so the process can hang well past the 5s budget.

The signal handler in `src/cli/mcp/index.ts` then awaits this and never calls `process.exit(0)`:

```47:55:src/cli/mcp/index.ts
    const shutdown = async (): Promise<void> => {
      try {
        await handle.close();
      } finally {
        process.exit(0);
      }
    };
    process.on('SIGINT', () => void shutdown());
    process.on('SIGTERM', () => void shutdown());
```

## Suggested fix

Standard Node 18+ graceful-shutdown pattern. Inside `handle.close()`:

1. `server.close(cb)` — stop accepting new.
2. `server.closeIdleConnections()` — drop idle keep-alives immediately.
3. After a deadline (e.g., 5s), `server.closeAllConnections()` — force-close lingering active connections so we exit before the orchestrator's grace-period SIGKILL.

Sketch:

```ts
close: () =>
  new Promise<void>((resolve, reject) => {
    let done = false;
    const finish = (err?: Error): void => {
      if (done) {
        return;
      }
      done = true;
      err ? reject(err) : resolve();
    };
    server.close((err) => finish(err ?? undefined));
    server.closeIdleConnections();
    const force = setTimeout(() => server.closeAllConnections(), 5_000);
    force.unref();
  }),
```

Also worth considering at the same time:

- Make the SIGINT / SIGTERM handler in `src/cli/mcp/index.ts` idempotent (a second signal currently kicks off a second `shutdown()` that races against the first) and add a hard fallback timer so a wedged shutdown still exits non-zero rather than hanging forever.
- Add a regression test along the same subprocess pattern as `src/cli/__tests__/validate-stdin.test.ts`: spawn the server, hit `/healthz` to seed a keep-alive, send SIGTERM, assert the process exits within (say) 2s with code 0.

## Why it matters

- **Cloud Run scale-down + rolling deploys.** Cloud Run sends SIGTERM, waits the configured `--timeout-seconds` grace, then SIGKILLs. With this bug, every replica burns the full grace period on shutdown, slowing rollouts and amplifying tail latency for any in-flight client during scale events.
- **Local dev annoyance.** `Ctrl+C` doesn't work; contributors learn to reflexively `kill -9`, which masks future regressions.
- **Pre-existing.** This predates #881 (which only renamed the entrypoint) and the `parseAsync` fix on `chore/scrub-pathfinder-mcp` — both are unrelated. Filing here so the MCP owner is aware before the next deployment.

## Out of scope

- The `runHttp` `'error'` event also isn't wired on `server.listen`, so `EADDRINUSE` hangs forever instead of rejecting the returned promise. Adjacent but separate; happy to file a second issue if useful.

moxious:
## Also fixes

Closes #883

`pathfinder-cli mcp --transport http` hung on SIGINT/SIGTERM because `server.close()` does not terminate idle keep-alive connections. Fix landed in commit `23b0de53`:

- `handle.close()` now calls `server.closeIdleConnections()` immediately (drops idle sockets without waiting for the peer) and schedules `server.closeAllConnections()` after 5 s as a force-close backstop.
- The SIGINT/SIGTERM handler is now idempotent (second signal is a no-op) and has a 10 s hard-exit fallback so the process never waits for a SIGKILL.
- Regression test added: seeds an idle keep-alive connection, asserts `handle.close()` resolves in under 2 s.

---

## Summary

This branch retires the Go MCP server entirely and consolidates all MCP tools into the centrally hosted TypeScript server under `src/cli/mcp/`. Net change: ~3,300 lines of Go (+ embedded JSON) gone, ~1,000 lines of TS added, two HTTP routes return 404, the build pipeline loses a copy-static prelude.

Three tracked phases land in one branch, in order:

- **P4 follow-up (`1c3d1624`)** — `clientGuidance.grafanaAppPlatform.steps` in `finalize.ts` now names `pathfinder_manage_guide_drafts` and `pathfinder_publish_guide` directly, matching the Assistant repo's PR #6457 web-surface tools.
- **MH4 — Migrate Go MCP runtime tools to TS** ([phase doc](docs/design/phases/mcp-hardening-4-go-mcp-migration.md)). The five stateless tools (`list_guides`, `get_guide`, `get_guide_schema`, `validate_guide_json`, `create_guide_template`) move to two new tool groups in `src/cli/mcp/tools/`:
  - `pathfinder_get_schema` wraps `src/cli/commands/schema.ts` — generated from the canonical Zod schemas in `src/types/`, no hand-maintained schema map.
  - `pathfinder_create_guide_template` returns a schema-valid starter guide; round-tripped through `runValidate` for clean-by-construction guarantees.
  - The remaining three tools (`list_guides`, `get_guide`, `validate_guide_json`) compose directly onto the existing CLI runners.
  - `8697495d` adds `$ref` for reused subschemas in `pathfinder_get_schema` so the exported JSON Schema stays compact.
- **MH5 — Retire the Go MCP entirely** ([phase doc](docs/design/phases/mcp-hardening-5-retire-go-mcp.md)). With MH4 done and Grafana Assistant's web-surface tools (Assistant repo PR #6457) replacing the per-instance back-channel for the "open the published guide" leg, `launch_guide` and the pending-launch queue have no live consumer. MH5 drops them along with the file itself:
  - **Phase A** — delete the frontend polling hook (`src/hooks/usePendingGuideLaunch.ts`) and its three mount sites (`ContextPanel`, `FullScreenPanel`, `FloatingPanelManager`).
  - **Phase B** — destructive Go cut: delete `pkg/plugin/mcp.go`, `mcp_test.go`, `static.go`, `pkg/plugin/static/guides/*.json`. Unregister `/mcp`, `/mcp/pending-launch`, `/mcp/pending-launch/clear` in `resources.go`. Drop `scripts/copy-static.js` and all its build-chain references.
  - **Phase C** — doc sweep across `AGENTS.md`, `MCP_SERVER.md`, `HOSTED-AUTHORING-MCP.md`, `AI-AUTHORING-IMPLEMENTATION.md`, and the MH4 phase doc.

## Why

The architecture pivoted to a single centrally hosted TS MCP on Cloud Run plus Grafana Assistant's web-surface tools for the per-instance handover (P4 status note, 2026-05-15). The per-tenant Go MCP is unused on the new path. Keeping a server alive purely to host one tool that nothing reaches is dead-code defense at a measurable maintenance cost.

## What changed

- **Go backend** — `pkg/plugin/mcp.go`, `mcp_test.go`, `static.go`, `static/guides/` removed. Three `/mcp*` routes unregistered in `resources.go`. New `helpers_test.go` holds the relocated `newTestApp` helper.
- **Frontend** — `src/hooks/usePendingGuideLaunch.ts` + test removed; barrel re-export, three call sites, two test mock entries, one stale comment cleaned up.
- **Build** — `scripts/copy-static.js` deleted; `copy-static` npm script removed; `build:backend*` variants no longer prefix with it; `.husky/pre-commit` no longer runs it.
- **TS MCP** — two new tool groups in `src/cli/mcp/tools/`. `finalize.ts` `clientGuidance` updated to name the Assistant web-surface tools.
- **Docs** — MH4 and MH5 phase docs added; `AGENTS.md`, `MCP_SERVER.md`, `HOSTED-AUTHORING-MCP.md`, `AI-AUTHORING-IMPLEMENTATION.md` swept of Go-MCP references.
- **HTTP transport shutdown** — graceful SIGINT/SIGTERM fix (closes #883); regression test added.

## Test plan

- [x] `npm run check` exit 0 (typecheck + lint + prettier + docs:sync + lint:go + test:go + test:ci).
- [x] `mage test`, `mage build:linux`, `mage build:linuxARM64` green.
- [x] Live curl against `npm run server`: `/resources/mcp` and `/resources/mcp/pending-launch` both return **404**.
- [x] Grep across `pkg/`, `scripts/`, `src/` for `launch_guide`, `pending-launch`, `guidesFS`, `repositoryJSON`, `copy-static` — zero live hits.
- [ ] Real-instance smoke against the Cloud Run TS MCP — reviewer's call whether to re-run before merge.

## Risk and blast radius

- The retired Go MCP routes were undocumented and labeled "non-production" in their lifetime; no documented consumer. Blast radius expected to be zero.
- `src/global-state/sidebar.ts::openWithGuide` is now unreferenced (its only caller was the deleted polling hook). The MH5 plan explicitly instructed to keep it. Worth a separate cleanup pass; not load-bearing for this PR.

## Audit trail

Every per-task commit references the slice and phase in its message. The two phase docs carry per-task commit SHAs in their checkboxes for fine-grained traceability.


moxious:
@cursoragent review PR

for quality and completeness.  I would like you to characterize what was added and what was  
  removed.  Please note that the removal of the Go MCP server is an architectural decision tied to factors in                         
  @docs/design/AI-AUTHORING-IMPLEMENTATION.md that deal with how we will integrate our MCP server together with Grafana Assistant.    
                                                                                                                                    
  I would like for you to do the following:                                                                                           
                                                                                                                                      
  1. Evaluate the appropriateness of the architectural decisions                                                                      
  2. Evaluate the quality of the code on the branch                                                                                   
  3. Examine whether there will be any blast radius or harm to other parts of the app (you may assume that the present Go MCP server  
  is unused, so the fact that it is being removed is not of concern, however other build artifacts and other things adjacent to it    
  might still be)                                                                                                                     
  4. Detect if there is any drift in the documentation and the implementation.  Check and see if any docs should have been written
  but weren't.  By "documentation" I mean to exlclude the plan & phase documentation, and I mean mostly developer documentation

Jayclifford345 (approved):

PR #895, opened 2026-05-17
fix: hide unsafe guided multistep type switches

### LINKED ISSUE #804: Block editor: multistep ↔ guided type switch via dropdown silently loses step description data
**Split off from #619 — Audit "switch type" in the Block Editor**

## Bug description

When converting between `multistep` and `guided` block types using the **TypeSwitchDropdown** in the block editor, step description text is silently lost. The converted block is structurally valid (passes schema validation and shows no error), but step descriptions render as empty.

## Root cause

The block editor exposes two code paths for `multistep ↔ guided` conversion:

1. **TypeSwitchDropdown** → calls `handleSwitchBlockType` → `convertBlockType` (generic engine in `block-conversion.ts`) — copies the `steps` array as-is but does **not** remap `tooltip` → `description`. `multistep` steps use `tooltip`; `guided` steps use `description`.

2. **Dedicated convert button** (exchange-alt icon in the multistep form) → calls `handleConvertType` — correctly maps `tooltip` ↔ `description`.

The dropdown path produces a block that Zod accepts without error, so there is no visible failure — only silent data corruption. The guided renderer ignores `tooltip` and shows blank step text.

## Steps to reproduce

1. Open the block editor
2. Create a `multistep` block and add at least one step with tooltip text
3. Use the **TypeSwitchDropdown** (not the dedicated convert button) to switch type to `guided`
4. Observe that the resulting guided block's step descriptions are empty

The reverse direction (`guided → multistep` via dropdown) has the same problem.

## Files involved

- `src/components/block-editor/hooks/useBlockConversionHandlers.ts` — `handleSwitchBlockType` (generic path) vs `handleConvertType` (field-aware path)
- `src/components/block-editor/forms/TypeSwitchDropdown.tsx` — exposes the problematic generic path for these types
- `src/components/block-editor/forms/MultistepBlockForm.tsx` — contains the dedicated convert button that works correctly
- `src/components/block-editor/utils/block-conversion.ts` — generic `convertBlockType` engine

## Proposed fix

Exclude the counterpart type (`guided` when source is `multistep`, and vice versa) from the results of `getAvailableConversions()`, so the only available path for converting between these two types is the dedicated `handleConvertType` handler that correctly remaps the fields.

Alternatively, intercept the dropdown selection for these specific types and redirect to `handleConvertType`.

## Architectural tier

**Tier 4 (components)** — all affected files are in `src/components/block-editor/`. No architecture violations involved.

wyf027:
## Summary
- Exclude guided from the generic multistep type-switch options and exclude multistep from generic guided options.
- Keep the dedicated field-aware conversion path as the only UI path for multistep ↔ guided conversions, preventing silent tooltip/description data loss.
- Add regression coverage for both dropdown option lists.

## Test plan
- `./node_modules/.bin/jest src/components/block-editor/utils/block-conversion.test.ts --runInBand --forceExit --coverage=false`
- `npm run typecheck`
- `./node_modules/.bin/eslint --no-cache src/components/block-editor/utils/block-conversion.ts src/components/block-editor/utils/block-conversion.test.ts`
- `npm run build` (passes with existing webpack warnings from @grafana/scenes/table panel exports and asset sizes)
- `git diff --check`

Fixes #804


moxious:
@cla-assistant check

moxious:
This PR looks good to me but it seems that our CLA integration didn't fire here, and need to follow up and verify that leno23 has signed the CLA before we can merge, will investigate further.

moxious:
@leno23 please take a look at the CLA comment - signing this is required before we can merge the PR.  Many apologies that it took a bit of time to get this integration back on the road working again, we appreciate the code contributions!

moxious:
There's nothing wrong with this PR, but we're unable to merge it due to the CLA block. I'm closing this for author non-response, but with branch updates & CLA signature we would be willing to consider again.  Thanks for the contribution.

Issue #824, opened 2026-04-30
useContentReset: surface user-facing error toast on reset failure


Jayclifford345:
`src/components/docs-panel/hooks/useContentReset.ts` catches errors during guide-reset, logs them with `console.error`, and re-throws — there's a `// TODO: Show error toast to user` in the catch block.

Wire the catch block up to the existing notification/toast API so the user sees a real error message when the reset fails. Look around the codebase for how other components surface user-facing errors.

> Interview task — the body is intentionally light. Ask clarifying questions or expand the scope as you go.

moxious:
## Agent analysis (leave-behind for next contributor)

**Status: Confirmed.** Line 85 of `useContentReset.ts` has a `// TODO: Show error toast to user` comment in the catch block that only calls `console.error` and re-throws. The user never sees a notification when reset fails.

**Difficulty: Easy**

**Fix:** Add `getAppEvents().publish({ type: 'alert-error', payload: ['Failed to reset guide', errorMessage] })` in the catch block. `@grafana/runtime` is already a project dependency; `BlockEditor.tsx` has an identical `notify()` helper to copy from. ~8 lines, 1 file, no risk to existing callers (re-throw is preserved).

PR #899, opened 2026-05-19
fix(docs-panel): surface toast when guide reset fails

leslie-heinzen:
Completes #824 

## Summary

Replace the TODO in useContentReset's catch block with an alert-error toast published via getAppEvents. Adds the resetGuideErrorTitle / resetGuideErrorMessage en-US strings and extends the hook test to assert the toast on both error paths (plus a happy-path negative).

https://github.com/user-attachments/assets/5c3a144f-ac3a-4ee3-bd84-07d8055fdb82

## Test Plan
- [ ] Test a known broken guide, or short-circuit [useContentReset](https://github.com/grafana/grafana-pathfinder-app/blob/fix/add-toast-notification-for-reset-guide-error-824/src/components/docs-panel/hooks/useContentReset.ts#L40-L45) with an error like `throw new Error("Ooops");` to force catch block execution. Verify that the toast notification is displayed when "Reset guide" is clicked.
- [ ] Test a working guide / remove the short-circuit `Error`. Verify that the happy path does not display display an error notification.

### DIFF
--- a/src/components/docs-panel/hooks/useContentReset.test.ts
+++ b/src/components/docs-panel/hooks/useContentReset.test.ts
@@ -12,6 +12,17 @@ import type { LearningJourneyTab } from '../../../types/content-panel.types';
 // Mock dependencies
 jest.mock('../../../lib/analytics');
 jest.mock('../../../lib/user-storage');
+jest.mock('@grafana/runtime', () => {
+  const mockPublish = jest.fn();
+  return {
+    getAppEvents: jest.fn(() => ({ publish: mockPublish })),
+    __mockPublish: mockPublish,
+  };
+});
+
+jest.mock('@grafana/i18n', () => ({
+  t: jest.fn((_key: string, fallback: string) => fallback),
+}));
 
 const mockReportAppInteraction = reportAppInteraction as jest.MockedFunction<typeof reportAppInteraction>;
 const mockEnrichWithStepContext = enrichWithStepContext as jest.MockedFunction<typeof enrichWithStepContext>;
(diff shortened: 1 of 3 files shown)


leslie-heinzen:
@cursoragent review

src/components/docs-panel/hooks/useContentReset.test.ts
```diff
@@ -12,6 +12,17 @@ import type { LearningJourneyTab } from '../../../types/content-panel.types';
...
+  const mockPublish = jest.fn();
+  return {
+    getAppEvents: jest.fn(() => ({ publish: mockPublish })),
+    __mockPublish: mockPublish,
+  };
+});
+
+jest.mock('@grafana/i18n', () => ({
+  t: jest.fn((_key: string, fallback: string) => fallback),
+}));
```
leslie-heinzen, line 25: I found this pattern in a few other tests for these lib dependencies.

src/components/docs-panel/hooks/useContentReset.ts
```diff
@@ -82,7 +84,16 @@ export function useContentReset({ model }: UseContentResetOptions) {
...
+        getAppEvents().publish({
+          type: 'alert-error',
+          payload: [
+            t('docsPanel.resetGuideErrorTitle', 'Reset failed'),
+            t(
+              'docsPanel.resetGuideErrorMessage',
+              "Couldn't reset guide progress. Please try again or reload the page."
+            ),
+          ],
+        });
```
leslie-heinzen, line 96: The block editor contains a little util `notify` function that is nice because it gets used multiple times in that component: https://github.com/grafana/grafana-pathfinder-app/blob/fix/add-toast-notification-for-reset-guide-error-824/src/components/block-editor/BlockEditor.tsx#L73-L76

I don't think this single call warrants abstracting it out to a shared function, but if we end up needing alerts in other places I think we should.

Jayclifford345 (approved):
LGTM!

PR #900, opened 2026-05-19
fix(first-dashboard): repair guide selectors and reactive conditional rendering

Jayclifford345:
## Summary

- Repairs the **OSS and Cloud** `first-dashboard` guides so the dashboard-creation flow works end-to-end without page refreshes.
- Fixes the underlying engine bug where `InteractiveConditional` did not re-render branches when `exists-reftarget` flipped after an action (e.g. opening the visualization picker).

### Guide content (`src/bundled-interactives/first-dashboard*/content.json`)

- Replaced the `conditional` wrapper around "All visualizations" with an explicit step using the canonical `data-testid Tab Visualizations` selector. Bar chart only lives under that tab, so the previous conditional silently skipped the step on first render and forced a refresh.
- Switched brittle raw selectors to Grafana's canonical e2e ids:
  - `grafana:components.PanelEditor.toggleVizPicker`
  - `grafana:components.PanelEditor.OptionsPane.fieldInput:Title`
- Added missing `tooltip` and `exists-reftarget` requirement on the Bar chart step.
- Save flow: the final "Save" in the drawer now targets `button[data-testid="data-testid Save dashboard drawer button"]` via `highlight` instead of the text-based `button` action.
- Sentence-case names per the writers' toolkit: "Walking adventure", "Jack and Jill walk altitude".
- Cloud guide brought up to parity with the OSS guide for the dashboard section.

### Engine (`src/components/interactive-tutorial/interactive-conditional.tsx`)

- Listen to `interactive-action-completed` on `document` (highlight/button actions dispatch there before the section's `interactive-step-completed`).
- Add a debounced `MutationObserver` when conditions include `exists-reftarget`, so branches re-render the moment a portal-injected element (e.g. picker tabs) appears.
- Keep the current branch mounted during re-evaluation — the "Checking conditions…" spinner now only appears on the initial check, not on every action.

### Tests

- New `interactive-conditional.test.tsx` covers:
  - `whenTrue` branch appears after `interactive-action-completed` without remount.
  - `whenTrue` branch appears after `interactive-step-completed` for `exists-reftarget` conditions.
  - No loading spinner takeover on re-evaluation of an already-resolved branch.
  - Stable `data-testid` + `data-passed` after evaluation.

## Why

Multiple users hit the same dead end at "Change the visualization to Bar chart": the picker opened on the Suggestions tab, Bar chart wasn't visible, and the conditional that should have inserted an "All visualizations" step required a page refresh to evaluate. The combination of selector fragility, late conditional evaluation, and a spinner overlay on every re-check was the failure mode.

## Test plan

- [x] `npx tsc -p tsconfig.json --noEmit` — clean
- [x] `npx eslint` on changed files — clean
- [x] `npx prettier --check` on changed files — clean
- [x] Targeted Jest run (`interactive-conditional|bundled-repository|parity|package-content|first-dashboard|build-repository|loader`) — 109/109 passing
- [x] `pathfinder-cli validate` on both guide JSONs — both valid
- [x] Manual run-through of `first-dashboard` against a local Grafana container (Explore → labels → Add to dashboard → Edit panel → All visualizations → Bar chart → title → unit → Save)
- [ ] Same manual run-through for `first-dashboard-cloud` against a Grafana Cloud instance

## Risk and reversibility

- Selector changes are scoped to two bundled guides and are testable in-product.
- Engine change adds two new event/observer subscriptions but only activates the MutationObserver when conditions include `exists-reftarget`; otherwise the behaviour is unchanged from a steady-state perspective. The branch is now kept mounted during re-evaluation, which is a UX improvement but worth a quick eyeball for any consumer that relied on the spinner state.
- Fully revertible — no schema, storage, or backend changes.

### DIFF
--- a/src/bundled-interactives/first-dashboard-cloud/content.json
+++ b/src/bundled-interactives/first-dashboard-cloud/content.json
@@ -186,41 +186,33 @@
         {
           "type": "interactive",
           "action": "highlight",
-          "reftarget": "button[data-testid=\"data-testid toggle-viz-picker\"]",
-          "content": "Let's change the visualisation type",
-          "tooltip": "Let's change the visualisation type",
+          "reftarget": "grafana:components.PanelEditor.toggleVizPicker",
+          "content": "Open the visualization picker to change the panel type.",
+          "tooltip": "Choosing the right visualization is key to telling your data's story. Time series graphs show trends over time, stat panels display single values, tables show raw data, and bar charts compare values side by side.",
           "requirements": ["exists-reftarget"]
         },
         {
-          "type": "conditional",
-          "conditions": ["exists-reftarget"],
-          "description": "Check whether we need to toggle to All visualisations",
-          "reftarget": "button[data-testid='data-testid Tab All visualizations']",
-          "whenTrue": [
-            {
-              "type": "interactive",
-              "action": "highlight",
-              "reftarget": "button[data-testid='data-testid Tab All visualizations']",
-              "content": "Let's select all visualisations.",
-              "tooltip": "Let's select all visualisations.",
-              "requirements": ["exists-reftarget"],
-              "skippable": true
-            }
-          ],
-          "whenFalse": []
+          "type": "interactive",
+          "action": "highlight",
+          "reftarget": "button[data-testid=\"data-testid Tab Visualizations\"]",
+          "content": "Select **All visualizations** to browse every panel type.",
+          "tooltip": "Bar chart is not listed under suggested visualizations. Open **All visualizations** to find it.",
+          "requirements": ["exists-reftarget"]
         },
         {
           "type": "interactive",
           "action": "highlight",
           "reftarget": "div[data-testid=\"data-testid Plugin visualization item Bar chart\"]",
-          "content": "Choosing the right visualization is key to telling your data's story. Time series graphs show trends over time, stat panels display single values, tables show raw data, and bar charts compare values side by side. For comparing Jack and Jill's final altitudes, a bar chart makes the difference instantly clear - you can see at a glance who climbed higher!"
+          "content": "Change the visualization to **Bar chart** to compare altitudes.",
+          "tooltip": "For comparing Jack and Jill's final altitudes, a bar chart makes the difference instantly clear - you can see at a glance who climbed higher!",
+          "requirements": ["exists-reftarget"]
         },
         {
           "type": "interactive",
           "action": "formfill",
-          "reftarget": "input[data-testid=\"data-testid Panel editor option pane field input Title\"]",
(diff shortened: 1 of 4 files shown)


Jayclifford345:
## Validation update

Followed the [validate-and-harden plan](.cursor/plans/validate-conditional-fix_f92b467d.plan.md) to confirm the conditional fix actually solves the original failure and to tighten the latent issues identified.

### Reproduction confirmed

Added an integration-style test that drives the real `InteractiveConditional` with real DOM mutations (no synthetic events):

- Initially: target element absent -> `whenTrue` branch is hidden.
- Append `button[data-testid="data-testid Tab Visualizations"]` to `document.body` (mimics the picker animation injecting the tab).
- Assertion: `whenTrue` branch appears within ~250 ms via the MutationObserver path.
- Reverse direction also verified: remove the element -> branch unmounts.

Result: fix validated. The original first-dashboard failure (need-to-refresh-to-see-the-step) does not reproduce against the patched component.

### Hardening landed in commit `d132d59`

- **Stable `conditionsKey`** via `useMemo` on `conditions.join(',')`. The parent passes a fresh array each render from JSON parsing, so without memoization the MutationObserver effect was tearing down and re-attaching on every parent render.
- **`evaluateRef` pattern** decouples the long-lived observer + event listeners from `evaluateConditions` callback identity churn (the test confirms re-renders no longer re-subscribe).
- **`runIdRef` race guard** inside `evaluateConditions`: concurrent re-evals (MutationObserver burst + step-completed + action-completed) could land out of order; older promises' results are now dropped. New test confirms.
- **`interactive-action-completed`** is now listened for on both `window` AND `document` (`challenge-block` fires on `window`, `interactive-state-manager` fires on `document`).
- **`reevalTimerRef`** cancels pending re-eval setTimeouts on unmount and when a new schedule supersedes one.
- **MutationObserver scope reduced** to `childList` + `subtree`. Attribute filter dropped: panel re-renders flip data-testid/class constantly, but element-existence transitions are childList events.

### Test status

- 8/8 conditional tests pass (4 originals + 4 new: DOM-injection reproduction, removal reproduction, observer-stability, race guard).
- 213/218 broader sweep tests pass (5 are todos).
- `tsc --noEmit`, ESLint, Prettier clean.

The engine fix stays in this PR. The patched component is now demonstrably correct against the original failure mode and resilient to render churn.

leslie-heinzen:
@Jayclifford345 I gave it a run through on a local instance all looks good. How do I test your changes in this PR on a Cloud instance?

Jayclifford345:
Amazing if you checkout: https://github.com/grafana/plugin-graft, this will allow you to render my changes onto a cloud instance to try I will send you an instance via VM

tomglenn (approved):

PR #903, opened 2026-05-20
Agent hardening Phase A: stop lying to the agent

moxious:
Design doc: [`docs/design/AGENT_HARDENING.md`](../blob/agent-hardening/phase-a/docs/design/AGENT_HARDENING.md)

This is all of Phase A (A1–A5) from the design doc — "fix the lies" prose corrections plus one new test. No product behavior change.

- **A1** — Reconciled tier-model prose in `AGENTS.md` and `systemPatterns.mdc` to the 5-tier (0–4) `TIER_MAP` in `src/validation/import-graph.ts` (was 4-tier, with engines/UI/support all numbered wrong).
- **A2** — New `Tier documentation sync` test in `architecture.test.ts` parses both docs and asserts equality with `TIER_MAP` so A1 cannot regress.
- **A3** — Fixed 10 of 23 quoted section references in `prevent-doc-drift` skill that pointed at non-existent headings; renumbered tier-detection heuristics.
- **A4** — F1–F4 IDs in `pr-review.md` were rotated relative to canonical `frontend-security.mdc`; re-aligned. Added one-line pointer in `CONCERNS.md`.
- **A5** — Removed misleading "auto-triggered by globs" / "always-apply" claims from `CONTEXT_INDEX.md`, `AGENTS.md`, `pr-review.md`, `prevent-doc-drift` (Cursor honors `globs:`; Claude Code does not).

## Test plan

- [x] `npm run check` clean (typecheck + lint + prettier + lint:go + test:go + test:ci; 3809 tests pass)
- [x] Architecture ratchet unchanged (`vertical=4 lateral=9 barrel=0`)
- [x] New tier-doc-sync test passes against post-A1 docs and fails on injected divergence (permanent negative cases in the suite)


docs/design/CONCERNS.md
```diff
@@ -31,6 +31,10 @@ Signal counting: a matching changed file path = 1 signal; a high-value semantic
 
 Flag unmapped clusters, disclose reduced confidence, and suggest new concerns when: a directory with multiple changed files maps weakly to no concern; repeated high-value symbols appear with no concern trigger; most meaningful hunks are covered only by always-on concerns. Coverage-gap detection is a disclosure mechanism, not a gate — do not suppress reviewers to avoid the gap.
 
+## Canonical rule sources
+
+F1–F6 security rules are defined in `.cursor/rules/frontend-security.mdc`. Other files (`.cursor/rules/pr-review.md`, `.cursor/skills/secure/SKILL.md`) reference these IDs without redefining them — if the summaries diverge, `frontend-security.mdc` wins.
```
leslie-heinzen: Stale ref, needs to be updated or removed?

```suggestion
F1–F6 security rules are defined in `.cursor/rules/frontend-security.mdc`. Other files (`docs/design/PR_REVIEW.md`, `.cursor/skills/secure/SKILL.md`) reference these IDs without redefining them — if the summaries diverge, `frontend-security.mdc` wins.
```

docs/developer/CONTEXT_INDEX.md
```diff
@@ -1,6 +1,8 @@
 # On-demand context index
 
-Load these files **only when working in the relevant domain**. Many are auto-triggered by glob patterns in `.cursor/rules/*.mdc` frontmatter; the rest are loaded by name.
+Load these files **only when working in the relevant domain**.
+
+**Cross-tool behavior — important.** In **Cursor**, many `.cursor/rules/*.mdc` files auto-load via `globs:` / `alwaysApply:` frontmatter. In **Claude Code**, that frontmatter is inert — `.mdc` files are discoverable but not auto-loaded; load them by name when working in the relevant domain (typically because another file like `pr-review.md` or a skill cites them). The "Auto-triggered by globs" column in the tables below documents the Cursor behavior for cross-tool reference; treat it as a hint about a rule's intended scope, not as a load-on-demand guarantee in Claude Code.
```
leslie-heinzen:
```suggestion
**Cross-tool behavior — important.** In **Cursor**, many `.cursor/rules/*.mdc` files auto-load via `globs:` / `alwaysApply:` frontmatter. In **Claude Code**, that frontmatter is inert — `.mdc` files are discoverable but not auto-loaded; load them by name when working in the relevant domain (typically because another file like `docs/design/PR_REVIEW.md` or a skill cites them). The "Auto-triggered by globs" column in the tables below documents the Cursor behavior for cross-tool reference; treat it as a hint about a rule's intended scope, not as a load-on-demand guarantee in Claude Code.
```

docs/developer/CONTEXT_INDEX.md
leslie-heinzen, line 101: Should the new `docs/design/AGENT_HARDENING.md` file get a row in this table?

leslie-heinzen (changes requested):
Couple of small file reference things, otherwise LGTM.

Kind of a side note/rant:

I've been thinking about this stuff after making a few code changes where I did bump into an architecture constraint violation. FWIW, my experience with the test failing was a positive one. The agent was also able to deal with the test failure in a reasonable way via a dynamic import. 

Kind of depends on the situation, but sometimes a good fix require a refactor, where violating code is extracted out and up into a higher tier, for example. I can't find that explicitly mentioned anywhere in the documentation to "encourage" the agents to think that way, or if it's something most models are smart enough to do on their own. On the other hand there are like...a lot of ways to solve this kind of thing, and maybe it's worse in this case to be too prescriptive, and best to get some human eyeballs on the arch constraint failure first. WDYT?

docs/developer/CONTEXT_INDEX.md
moxious: No, I don't think so -- because Agent hardening is a design doc that describes various phases; after the phases are executed, most of the context in that file will be useless.  And we don't really have to document the agent context system itself, because basically AGENTS.md acts as an index to it.

moxious:
@leslie-heinzen: 

> I've been thinking about this stuff after making a few code changes where I did bump into an architecture constraint violation. FWIW, my experience with the test failing was a positive one. The agent was also able to deal with the test failure in a reasonable way via a dynamic import.
> 
> Kind of depends on the situation, but sometimes a good fix require a refactor, where violating code is extracted out and up into a higher tier, for example. I can't find that explicitly mentioned anywhere in the documentation to "encourage" the agents to think that way, or if it's something most models are smart enough to do on their own. On the other hand there are like...a lot of ways to solve this kind of thing, and maybe it's worse in this case to be too prescriptive, and best to get some human eyeballs on the arch constraint failure first. WDYT?

I agree, have seen the same thing. I do think we need this in the repo, but this isn't (imo) the right place to put it.  Went down the rabbit hole, did some research and I found that the reason why agents are so shy on refactors comes down to this: 

- Agents optimize for the prompted task, not for coherence
- Agents can't see what they don't fetch into context (the thing they're supposed to discover/know to refactor they probably didn't read, e.g. duplicate copies of the same code elsewhere)
- Agents have no incentive to refactor what's already working - off the path of what they're asked
- Agents have no cross-session memory of the architecture

The 2nd one is really hard to get around, even in the planning process.  I'm working on a `/techdebt` skill that knows how to identify refactor opportunities, paired with common tech debt patterns known to be issues with agents.  I'm not sure we can build "opportunistically consider refactors" into _writing new code_ because we'd have to tell the agents to search potentially ALL the code to know what's in scope that could need refactoring.  But by contrast, you can look at existing modules that are bounded, and see refactoring opps within them.  Even **that** misses refactor opportunities that cross component boundaries, there - probably no substitute for just architectural understanding.

leslie-heinzen (approved):

PR #904, opened 2026-05-20
feat(deep-linking): centralize deep-linking with SPA nav listener

leslie-heinzen:
## Summary

Previously the ?doc= / ?panelMode= / ?kiosk_session= deep-link flow lived inline in plugin.init in module.tsx, so it only fired once per Grafana session - at the moment the plugin module was loaded (i.e. when Grafana opens). This means pathfinder can be deep-linked externally, but not from another page in grafana.

Extract the consumption side of the deep-link contract into a new module utils/pathfinder-deep-link-handler.ts:
  - handlePathfinderDeepLink is idempotent and runnable any number of times. Dedupes via lastProcessedSearch so listeners can fire it freely.
  - installDeepLinkNavListener registers a single navigation listener using the fallback-ladder pattern from experiment-orchestrator.ts and highlighted-guide-orchestrator.ts: grafana:location-changed always, history.listen when the runtime exposes it, and popstate only as a defensive escape hatch when locationService.getHistory() is unavailable.

plugin.init now calls the handler for the initial load case and installs the listener so subsequent SPA navigations carrying ?doc= re-run the same flow. The kiosk and experiment auto-open branches keep their existing !docsParam gating via a pre-handler snapshot of the search string.

### Example:

The setup guide app devs [are attempting to nav to pathfinder](https://raintank-corp.slack.com/archives/C08VB4WC64R/p1779208488792399) from a link in one of its onboarding pages with the url `/a/grafana-pathfinder-app?doc=%2Fdocs%2Flearning-paths%2Fsend-traces-alloy%2F`, which took the user to the `grafana-pathfinder-app` "My Learning" page, but the querystring param was ignored. With this PR's change, it opens the pathfinder sidebar and requested doc.

https://github.com/user-attachments/assets/01cd391c-b35c-4536-8742-6d6ef3b24ef3

### DIFF
--- a/src/module.tsx
+++ b/src/module.tsx
@@ -1,5 +1,4 @@
 import { AppPlugin, AppPluginMeta, type AppRootProps, PluginExtensionPoints, usePluginContext } from '@grafana/data';
-import { locationService } from '@grafana/runtime';
 import React, { lazy, Suspense, useEffect, useMemo } from 'react';
 import { LoadingPlaceholder } from '@grafana/ui';
 import { reportAppInteraction, UserInteraction } from './lib/analytics';
@@ -13,8 +12,8 @@ import { linkInterceptionState } from './global-state/link-interception';
 import { sidebarState } from 'global-state/sidebar';
 import { panelModeManager } from './global-state/panel-mode';
 import { suggestionState } from './global-state/suggestion';
-import { validateRedirectPath } from './security/url-validator';
-import { parsePathfinderDeepLink, stripPathfinderParams } from './utils/pathfinder-search-params';
+import { handlePathfinderDeepLink, installDeepLinkNavListener } from './utils/pathfinder-deep-link-handler';
+import { parsePathfinderDeepLink } from './utils/pathfinder-search-params';
 
 // Buffer pathfinder-suggest events that arrive before async init completes.
 // Registered synchronously (before any await) so events from faster-loading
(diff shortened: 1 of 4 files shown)


src/utils/pathfinder-deep-link-handler.test.ts
```diff
@@ -0,0 +1,284 @@
...
+  ...overrides,
+});
+
+const setSearch = (search: string) => {
+  // jsdom doesn't reflect `replaceState` calls into `window.location.search`
+  // reliably in all versions — set the URL via `history.replaceState` so the
+  // handler's `window.location.search` read matches what the test intends.
+  const url = new URL(window.location.href);
+  url.search = search;
+  window.history.replaceState({}, '', url.toString());
```
leslie-heinzen, line 87: Honestly, I don't think this is necessary so I might change this. This was an Opus 4.7 invention.

src/utils/pathfinder-deep-link-handler.ts
```diff
@@ -0,0 +1,223 @@
...
+}
+
+/** Process Pathfinder deep-link params from the current URL. Idempotent. */
+export function handlePathfinderDeepLink(deps: DeepLinkHandlerDeps): boolean {
+  const search = window.location.search;
+  if (!hasAnyPathfinderParam(search)) {
+    return false;
+  }
+
+  if (search === lastProcessedSearch) {
```
moxious, line 55: The dedup is for back-to-back firings from the multi-transport listener (grafana:location-changed + history.listen both firing on the same nav). That's good.
                                                                                                                                    
But the handler stores the pre-strip search. After one successful run with `?doc=api:foo, lastProcessedSearch === '?doc=api:foo'` and the URL is stripped to ''. The user then navigates somewhere else, comes back, and clicks a `?doc=api:foo` link again, `window.location.search === '?doc=api:foo'`, matches lastProcessedSearch, deduped. Same guide can't be re-deep-linked.

The test proves dedup of synchronous re-fires; it doesn't cover "nav away, nav back with same search".

Suggest clear lastProcessedSearch once the URL is stripped, add a test it('re-runs the handler when the user nav-aways and returns to the same ?doc=')

src/module.tsx
```diff
@@ -141,172 +141,22 @@ plugin.init = function (meta: AppPluginMeta<DocsPluginConfig>) {
...
+  handlePathfinderDeepLink(deepLinkDeps);
+  // Re-runs on SPA navigations; plugin.init fires only once per session.
+  installDeepLinkNavListener(deepLinkDeps);
 
-  if (docsParam && !shouldMountSidebar(pathfinderEnabled, mainVariant, after24hVariant)) {
-    const url = new URL(window.location.href);
-    stripPathfinderParams(url);
-    window.history.replaceState({}, '', url.toString());
-
-    import('./components/ControlGroupDocPopup').then(({ showControlGroupDocPopup }) => {
```
moxious, line 200: Change introduced a static import instead of a lazy import.  This drags a bunch of extra code into sessions, 99% of which won't need it.  This came from an agent tier-violation followup which is fair, but fix doesn't have to be a static import

Can pass a lazy loader:

```
  interface DeepLinkHandlerDeps {              
    ...                                                                                                                             
    loadControlGroupDocPopup: () => Promise<{ showControlGroupDocPopup: (source: string) => void }>;                              
  }                                                                                                 
  loadControlGroupDocPopup: () => import('./components/ControlGroupDocPopup')
```

src/utils/pathfinder-deep-link-handler.ts
```diff
@@ -0,0 +1,223 @@
...
+    window.addEventListener('popstate', handler);
+  }
+}
+
+function hasAnyPathfinderParam(search: string): boolean {
+  if (!search) {
+    return false;
+  }
+  return (
+    search.indexOf('doc=') !== -1 || search.indexOf('panelMode=') !== -1 || search.indexOf('kiosk_session=') !== -1
```
moxious: This will match URLs like ?my_doc=foo or ?panelMode= from another plugin's links. Low practical risk, easy to make correct:       

```                                                                                                                                    
  const params = new URLSearchParams(search);                                                                                       
  return params.has('doc') || params.has('panelMode') || params.has('kiosk_session');                                             
```
                                                                                                                                    
Reuse `PATHFINDER_PARAMS` (even better) from pathfinder-search-params.ts if you want to keep one source of truth.

moxious (changes requested):
Validating that this is a good PR I'd like to get in.  Giving you some comments on how to harden/improve, but I have actually tripped over what this is fixing so glad to have it. 

One other nit that bugs me a lot about this area of deep linking in our code.

This PR keeps the existing behavior when findDocPage returns null for an unknown, user gets a console.warn and silent fall-through to the recommendations tab. From a user's perspective that's indistinguishable from "the link did nothing".  This has bit me repeatedly.
                                                                                                                                    
Either in this PR or if you prefer as a follow-up, we should fire a Grafana toast in three failure modes:                                                        
  1. findDocPage returns null (unknown prefix / format).
  2. fetchBackendInteractive returns 404 (api:<id> with no matching CRD).                                                           
  3. fetchBackendInteractive returns other non-2xx.

leslie-heinzen:
Review feedback implemented here: https://github.com/grafana/grafana-pathfinder-app/compare/5bf9bb2bdd7b9946bb352b50f5ee6ec37bfeaee8..6a75ae0c70cb1e633708dd282036d44ad18f0849

leslie-heinzen:
@moxious Appreciate the review -- I'll take a look at the `findDocPage` bad url/err handling issue you've raised in a separate PR.

MattIPv4 (commented):
General thought here, why not rely on react-router or Grafana's locationService here, instead of listening manually to window events etc?

leslie-heinzen:
> General thought here, why not rely on react-router or Grafana's locationService here, instead of listening manually to window events etc?

Thanks for asking, Matt. We use `locationService` when [setting up the listener](https://github.com/grafana/grafana-pathfinder-app/pull/904/changes/780f2198d41f02708cfc3e0fdb9c9fc262a97a25#diff-1fcdfb371e43ab1c10cafadbb6a009d926bdbdf6ac472d4ba2308b12fadaeae5R182-R190). The listener registered with it would be the hottest code path. The catch block `popstate` listener is just in case locationService fails for some reason. Not sure that's likely at all, so arguably this is hyper-defensive.

You prompted me to dig a little more into the grafana/grafana code, and a couple of observations:

* The `grafana:location-changed` custom event listener was here before (and in some of the other plugin code), but I think we'd be good without it at all? Possibly a legacy event we don't need to listen for?
* There's an [observable for the location](https://github.com/grafana/grafana/blob/main/packages/grafana-runtime/src/services/LocationService.tsx#L46-L62) we could subscribe to instead of `locationService.getHistory().listen()`. Same behavior, slightly more idiomatic API.

moxious (approved):
LGTM, review comments addressed

PR #906, opened 2026-05-21
fix(experiments): fire feature-flag exposure on override + document /packages URL form

Jayclifford345:
## Summary

Two related fixes that make the local-override workflow documented in [`docs/developer/EXPERIMENT_TESTING.md`](docs/developer/EXPERIMENT_TESTING.md) actually work end-to-end for the highlighted-guide experiment:

1. **`pathfinder_feature_flag_evaluated` now fires when an override is set.** The exposure hook only ran on real OpenFeature client evaluations; overrides short-circuited the readers before the client was called, so the event never fired and the `clearExposures()` / `showExposures()` debug helpers were no-ops for any override-driven QA / demo run. The filter + dedup + report logic is extracted into a shared `reportFeatureFlagExposure` helper that both the hook and the override branches call — prod and dev paths share identical filtering and the same `(hostname, flag, variant)` dedup state.
2. **Document the required `guideId` URL form for interactive-learning guides.** Found while debugging: a `guideId` like `https://interactive-learning.grafana.net/guides/<slug>/` makes the docs panel fall through to a plain `fetchContent(url)` render — single page, no milestones — because `isPackageContentUrl` only matches `/packages/<slug>/content.json`. Doc now explains the requirement up front, the broken control-arm example URL is fixed, and the symptom is mapped in "Common gotchas".

## What changed

- `src/utils/openfeature-tracking.ts` — export `reportFeatureFlagExposure(flagKey, value)`; `TrackingHook.after` becomes a thin delegate.
- `src/utils/openfeature.ts` — call the helper from the override branches in `getExperimentConfig` and `getHighlightedGuideConfig`.
- `src/utils/openfeature-tracking.test.ts` — new `describe('reportFeatureFlagExposure')` block (8 cases) covering once-per-tuple dedup, variant reassignment re-fire, excluded / boolean / unknown-flag skips.
- `src/utils/openfeature.test.ts` — 5 new cases under `flag overrides` asserting the wiring fires from `getExperimentConfig` / `getHighlightedGuideConfig` and not from invalid-override fall-through.
- `docs/developer/EXPERIMENT_TESTING.md` — new "\`guideId\` URL form" section + control-arm URL fix + "only one page renders" gotcha entry.

## Why

Both bugs surfaced during a single debug session:

- A QA user tried to verify analytics for a manually overridden arm — no event fired, hence the analytics-verification workflow in the doc was misleading.
- The same session swapped the control-arm guide to a `/guides/<slug>/` URL (a plausible URL form, also documented historically), and only the first page of the journey rendered — because the package gate is URL-shape-based, not type-based.

The `docType` operator override already wins over `findDocPage`'s URL inference, but `openLearningJourney`'s manifest-loading redirect at [docs-panel.tsx:305](src/components/docs-panel/docs-panel.tsx) is gated by `isPackageContentUrl`, so the URL form still matters. Rather than loosen that gate, the doc fix makes the requirement explicit so future testers don't hit the same trap.

## Test plan

- [x] `npm run test:ci -- --testPathPatterns=openfeature` — 54/54 pass (8 new in `openfeature-tracking.test.ts`, 5 new in `openfeature.test.ts`).
- [x] Manual: set override with `/packages/grafana-irm-configuration-lj/content.json`, reload, confirm:
  - `pathfinder_feature_flag_evaluated` fires in the analytics tap on first hit.
  - `__pathfinderExperiment.showExposures()` lists the marker after the fire.
  - `__pathfinderExperiment.clearExposures()` + reload re-fires the event.
  - Sidebar opens, learning-journey tab auto-launches with milestone toolbar populated from `manifest.json`.
- [x] Manual: doc walkthrough — the new URL-form section + updated control example produce a working learning journey end-to-end.

## Risk and reversibility

Low risk. The code change adds a single new call site in each of two override branches; the helper is the same filtering / dedup logic that already ran on the client path, with full unit-test coverage for the new entry point. Doc changes are additive plus one URL-string fix. Easy to revert.

### DIFF
--- a/docs/developer/EXPERIMENT_TESTING.md
+++ b/docs/developer/EXPERIMENT_TESTING.md
@@ -134,7 +134,7 @@ This is the form `openLearningJourney` / `openDocsPage` recognise as a package U
 
 ```js
 __pathfinderExperiment.setOverride('pathfinder.highlighted-guide-experiment', {
-  variant: 'control',
+  variant: 'treatment',
   pages: ['/a/grafana-irm-app*'],
   guideId: 'https://interactive-learning.grafana.net/packages/irm-configuration/content.json',
   docType: 'interactive',
@@ -150,7 +150,7 @@ Forces the Featured card type to `learning-journey` so the click-through opens w
 
 ```js
 __pathfinderExperiment.setOverride('pathfinder.highlighted-guide-experiment', {
-  variant: 'treatment',
+  variant: 'control',
   pages: ['/a/grafana-irm-app*'],
   guideId: 'https://interactive-learning.grafana.net/packages/grafana-irm-configuration-lj/content.json',
   docType: 'learning-journey',
--- a/docs/developer/EXPERIMENT_TESTING.md
+++ b/docs/developer/EXPERIMENT_TESTING.md
@@ -120,11 +120,21 @@ location.reload();
 
 A/B test for guide content. Both `control` and `treatment` keep Pathfinder visible, auto-open the sidebar, **and auto-launch the configured `guideId` as a tab** on matched pages — using the same `auto-launch-tutorial` seam as the `?doc=` deep link. The user stays on the page they were on; no navigation happens. The Featured-slot injection still runs in parallel so the card is also visible in the recommendations tab. Use this to compare two candidate guides on the same page.
 
+### `guideId` URL form (read this before configuring an interactive-learning guide)
+
+For guides hosted on `interactive-learning.grafana.net`, **always use the package content URL**:
+
+```
+https://interactive-learning.grafana.net/packages/<slug>/content.json
+```
+
+This is the form `openLearningJourney` / `openDocsPage` recognise as a package URL (`isPackageContentUrl` in [src/docs-retrieval/package-info-from-url.ts](../../src/docs-retrieval/package-info-from-url.ts)) — that's the gate that triggers the sibling `manifest.json` fetch which populates milestones, the toolbar, and the multi-page navigation. The platform also serves the same guide under `https://interactive-learning.grafana.net/guides/<slug>/`, but that public web-page form bypasses the package gate and falls through to a plain `fetchContent(url)` call — the guide opens as a single page with no milestones. If you only see one page when you expected a learning journey, this is almost always the URL form. See "Common gotchas" below.
+
 ### Treatment — interactive package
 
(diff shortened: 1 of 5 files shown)


moxious:
fix the docs heading/variant mismatch. Everything else is non-blocking.

  [blocking] [docs] Variant values contradict section headings in EXPERIMENT_TESTING.md                                             
  docs/developer/EXPERIMENT_TESTING.md:133-161 — the second commit swapped the variant field in both highlighted-guide examples, so:
  - ### Treatment — interactive package now has variant: 'control'                                                                  
  - ### Control — learning journey now has variant: 'treatment'                                                                     
                                                                                                                                    
  Either the headings are stale or the variant values are wrong; one of them needs to flip back. This file is the primary reference 
  for QA overrides, so the mismatch will actively mislead.                                                                          
                                                         
  [suggestion] Override path fires exposure on every call; helper handles dedup                                                     
  src/utils/openfeature.ts:490, 552 — getExperimentConfig / getHighlightedGuideConfig are plain getters that may be called from   
  React render paths. Calling reportFeatureFlagExposure synchronously inside them is safe today because the helper dedups via the   
  in-memory Set + localStorage marker, but the console.warn immediately above will still fire on every call. Not introduced by this
  PR, but worth knowing — and the test should pass through repeated override calls to the exposure helper codifies the every-call   
  behavior, which is fine as long as the dedup contract is preserved.                                                             
                                                         
  [nit] JsonValue cast hides a structural assumption                                                                                
  src/utils/openfeature.ts:490, 552 — config as unknown as JsonValue and validated as unknown as JsonValue work because both objects
   are plain JSON shapes today. A short comment ("plain JSON object — safe cast") or narrowing reportFeatureFlagExposure's signature
   to accept a structural type ({ variant?: string } & object) would make the contract explicit and remove the double-cast.       
  Non-blocking.                                                                                                                     
                                                                                                                                  
  [suggestion] Test naming                                                                                                          
  src/utils/openfeature-tracking.test.ts:184 — freshReportFn() uses jest.resetModules() + require() to get a fresh module so the
  in-memory Set resets between persistence tests. This is correct but subtle; the inline comment in freshHook covers the rationale —
   consider sharing a one-line note above freshReportFn too.

moxious (changes requested):
See comment

moxious (approved):

PR #907, opened 2026-05-21
feat(snippets): reusable snippet references in interactive guides

tomglenn:
image

## What this does

Lets a guide reuse a common set of steps instead of copy-pasting them. An author drops in a **snippet reference** by name, and when the guide loads Pathfinder pulls that snippet's content from the CDN and slots it into place. Improve a snippet once upstream and every guide that uses it picks up the change — no edits to the guides themselves.

In the block editor there's a new "Reusable" group in the block palette with a picker to insert a reference. If a snippet can't be loaded, the guide still renders with a small placeholder where it would have gone, rather than breaking.

**Scope for this first version:** guides consume snippets that are authored upstream (in `grafana/interactive-tutorials`), snippets can't contain other snippets, and guides always get the latest version (no pinning). Authoring snippets and richer editor previews are deliberately left for follow-ups.

## Why this is larger than the 500-line guideline

A snippet reference is only useful when the whole path works end to end, so this lands as one coherent unit: the new reference type, the engine that fetches and inserts snippets, the wiring that resolves references when a guide loads, and the editor UI to add one. Splitting it would ship half-features that can't be exercised on their own — an editor that inserts references nothing can resolve, or an engine with no way to author against it.

It's also test-heavy by design: roughly **40% of the diff (~510 of ~1,300 lines) is tests**. The remaining feature code is genuinely new surface, not padding, and there's no smaller slice of "snippets work" worth reviewing in isolation.

## Testing

- **Automated:** schema validation, resolver caching/dedupe and failure handling, the CDN fetch paths, and the insert-and-splice logic (top level, nested sections, conditional branches, and failure placeholders).
- **Manual:** inserting a snippet reference in the block editor and confirming it renders from the live catalog — in progress.

### DIFF
--- a/src/components/content-renderer/content-renderer.tsx
+++ b/src/components/content-renderer/content-renderer.tsx
@@ -1,7 +1,7 @@
 import React, { useRef, useEffect, useLayoutEffect, useMemo, useState, useCallback } from 'react';
 import { css } from '@emotion/css';
 import { GrafanaTheme2 } from '@grafana/data';
-import { TabsBar, Tab, TabContent, Badge, Tooltip } from '@grafana/ui';
+import { TabsBar, Tab, TabContent, Badge, Tooltip, LoadingPlaceholder } from '@grafana/ui';
 
 import { RawContent, ContentParseResult } from '../../types/content.types';
 import {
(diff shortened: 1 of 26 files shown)


leslie-heinzen:
@tomglenn I will try to get you a review for this by EOD (for me) today 👀

src/components/content-renderer/content-renderer.tsx
```diff
@@ -551,14 +553,45 @@ function ContentProcessor({ html, contentType, baseUrl, onReady, responses }: Co
...
+        return;
+      }
+      if (!guideHasSnippetRefs(parsedGuide)) {
+        return;
+      }
+      const resolved = await inlineSnippetRefsInGuide(parsedGuide);
+      if (cancelled) {
+        return;
+      }
+      setParseResult(parseJsonGuide(resolved, baseUrl));
```
leslie-heinzen: Not a change request, just probing for opinion. Feel free to ignore for the sake of the review:

So, `cancelled` here _does_ prevent us from setting state on an unmounted component, so it [satisfies the anti-pattern described here](https://github.com/grafana/grafana-pathfinder-app/blob/af9c3f37310b98c43c1e6a892b971b8500a71987/.cursor/rules/react-antipatterns.mdc#L136-L162), but it doesn't cancel the inflight request, which personally I like where possible (and is what is happening in the linked example via AbortController/AbortSignal).

That said, `cancelled` is consistent with what I'm seeing elsewhere in the app: just prevent the state update on unmounted component, then let the network request resolve and ignore it. 

I do see some AbortController/AbortSignals sprinkled around, but it's not a consistent approach.

And just to be clear I think the existing pattern is fine. I would be more concerned if we had a UI that unmounted/remounted _a lot_, and we risked hammering a server with network requests for no reason. I don't think that's a problem here.

Curious if you have any thoughts.

src/components/content-renderer/content-renderer.tsx
```diff
@@ -551,14 +553,45 @@ function ContentProcessor({ html, contentType, baseUrl, onReady, responses }: Co
...
     if (isJsonGuideContent(html)) {
       return parseJsonGuide(html, baseUrl);
     }
     return parseHTMLToComponents(html, baseUrl);
   }, [html, baseUrl]);
+  const [parseResult, setParseResult] = useState<ContentParseResult>(initialParseResult);
+
+  useEffect(() => {
+    setParseResult(initialParseResult);
+  }, [initialParseResult]);
```
leslie-heinzen: Very minor: an agent suggests this breaks the [derived state anti-pattern](https://github.com/grafana/grafana-pathfinder-app/blob/af9c3f37310b98c43c1e6a892b971b8500a71987/.cursor/rules/react-antipatterns.mdc#L644-L660) (R14).

TBH I wouldn't have caught it but I get why it was flagged: "When html / baseUrl change, initialParseResult is recomputed, but useState's initializer doesn't re-run, so the first render after a prop change uses the old parseResult; the effect then schedules a second render with the new value."

src/components/content-renderer/content-renderer.tsx
```diff
@@ -551,14 +553,45 @@ function ContentProcessor({ html, contentType, baseUrl, onReady, responses }: Co
...
+      } catch {
+        return;
+      }
+      if (!guideHasSnippetRefs(parsedGuide)) {
+        return;
+      }
+      const resolved = await inlineSnippetRefsInGuide(parsedGuide);
+      if (cancelled) {
+        return;
+      }
```
leslie-heinzen: Little weird transient state thing:

If the guide is ONLY snippet refs, we will display the "no elements" error splash until they load because of the `parsedContent.elements.length === 0` block that produces warnings. 

Again, think this is minor (what are the odds the guide is just snippets?), but wanted to raise, so here's a little vid (sorry I'm in a coworking space and forgot to turn off the mic). 

https://github.com/user-attachments/assets/dfd81cc3-e4e5-4a18-bbab-68ca79f9687c

leslie-heinzen (dismissed):
Looking good -- couple of super minor things; changes aren't critical for either at this point imo, so I'm going to approve but would appreciate your 👀 on them.

One comment about handling inflight network requests in unmounted components. Just curious for your thoughts on that.

src/components/content-renderer/content-renderer.tsx
tomglenn: A good observation, we should probably be aborting the request as well, but I think this is a larger issue outside of the scope of this PR as there's currently no way to thread these through. Worth raising a separate issue to address though.

src/components/content-renderer/content-renderer.tsx
tomglenn: Yep this is fair, addressed.

src/components/content-renderer/content-renderer.tsx
tomglenn: Good catch, this should be resolved now,

leslie-heinzen (approved):

PR #885, opened 2026-05-15
refactor(interactive-section): Tier A+B extractions following high-risk refactor guidelines

moxious:
## Summary

Refactor of `src/components/interactive-tutorial/interactive-section.tsx` (2118 → 1418 lines, **-33%**) following the [High-Risk Refactor Guidelines](https://github.com/grafana/grafana-pathfinder-app/wiki/High-Risk-Refactor-Guidelines). Tier A + Tier B complete; Tier C deliberately not attempted (gate STOP, see below).

The commit log is structured to be the source of truth — each commit declares its `Phase`, `Step`, `Pattern`, and `Risk`, and explains what it preserves and why. `git log --grep "Phase: Tier A"` (or `Tier B`, or `Pattern:`) reconstructs the reasoning path.

## What landed

- **Phase 0** — 4 tripwire test files added before any code moves (registry, contracts, symmetry, runner) covering Patterns I, J, E, and G.
- **Tier A** — 4 atomic extractions: module-level registry → `section-registry.ts` (Pattern I), numbering helpers → `section-numbering.tsx`, step-type schema table → `step-type-registry.ts` (Pattern E), and collapse of the two parallel `switch-on-component-type` chains to one table-driven loop.
- **Tier B** — 5 hooks: `useSectionAutoCollapse`, `useSectionScroll` (Pattern F), `useDocumentStepProgress` (Pattern J), `useSectionRequirements` (Pattern F + timing contract), and the largest one, `useSectionPersistence` (Pattern J, all 4 storage namespaces + `interactive-progress-saved` dispatch + mount-only restore).

## What we did NOT do — and why

Tier C (extract the 430-line `handleDoSection` async orchestrator into `section-runner.ts`) is **not** attempted in this PR. The plan defined 4 gate criteria for proceeding; one failed:

- **Criterion 2** (each of the 7 runner-tripwire cases caught a real discrepancy during A/B): **not met**. Five of the seven cases remain `it.todo` because the shared test harness doesn't yet expose the seams needed for them (per-test override of `useInteractiveElements`, deterministic mid-loop suspension of `executeInteractiveAction`, sequenced returns from `checkRequirementsFromData`). Promoting the todos to real `it` cases requires test infrastructure work that's its own project. Per the Guidelines (Principle 7 + Pattern G stop conditions): extracting an async state machine without comprehensive interleaving coverage is the exact failure mode the tripwire system exists to prevent.

[...]

PR #909, opened 2026-05-21
refactor(interactive): collapse state ownership, restore tier hygiene

Jayclifford345:
## Summary

Whole-pipeline cleanup of the interactive-guide rendering layer. The before-state had several places that each knew "is this step done?", four legacy event names, four mental translations for the same field, two parallel tab loaders, three architecture-test ratchet exceptions papering over a Tier-2 module importing the UI layer, an FSM that could disagree with the canonical store on the skip and standalone-objectives paths, a DOM-only `section-completed:` gate that broke for any unmounted section, and standalone step IDs that mutated on every reload.

This PR consolidates each of those — with the trade-offs the diff actually delivers, not aspirational ones.

### State ownership

- Delete `use-standalone-persistence.ts`. Eight step components now talk to the `completion-store` directly via `useStepCompletion` + `markStepCompleted` / `resetStep`.
- Delete the `parentCompleted` prop everywhere and the `cloneElement` enhancement that fed it. Children read their own completion from the store.
- Shrink `SectionState` to `{ acknowledged }`. The completion set lives in the canonical store; the cursor is a pure derivation via `computeCursor`. The #842 Bug 1 invariant ("ack requires completion") is preserved by `ACKNOWLEDGE` taking `completedCount` and refusing when zero.
- New `useSectionCompletion(sectionId)` selector + atomic store helpers (`markStepsCompleted`, `resetSteps`, `resetSection`, `evictSectionCache`).
- **Completion store now lives at `src/global-state/completion-store.ts` (Tier 1).** The store is the canonical persistence layer. `useStepChecker`'s FSM and the `SequentialRequirementsManager` orchestrator remain as in-memory mirrors needed to drive UI — they write through to the store on every terminal transition (manual completion, skipped, objectives auto-complete, and reset) via `writeStoreCompletion` / `writeStoreReset`, so the mirrors and the canonical store cannot disagree on either axis. `step-checker.store-bridge.test.ts` pins the contract (8 tripwires across both directions).
- New `sectionId: string | null | undefined` prop on `useStepChecker`: `null` opts out of store writes (used by the section's own objectives self-checker, which is not a real step).
- Hydration vs. reset race fix: `completion-store` tracks per-section "cleared during hydration" so a reset issued before the async storage snapshot resolves cannot be resurrected from the stale snapshot. `resetSection` persists even when the cache is empty, so the clear survives navigation mid-hydration.
- **Reset paths must evict the in-memory cache.** Moving completion state out of component lifetime into module scope creates a footgun — every storage-clear path also has to evict the cache or live subscribers keep rendering "completed." The four reset sites (`useContentReset`, `useGuidePreviewProgress.reset`, `learning-paths.hook.ts`, `MyLearningTab.handleResetAll`) pair `interactiveStepStorage.clearAllForContent` / `clearAll` with `evictContentCache` / `evictAllContentCaches`. Documented as a load-bearing invariant in `STEP_MODEL.md`.

### Events

- Delete the four legacy progress events (`interactive-step-completed`, `section-completed`, `interactive-section-completed`, `interactive-progress-saved`). One unified discriminated `pathfinder:progress` event with `kind: 'step' | 'section' | 'guide'` replaces them.
- Seven listener sites migrated: `content-renderer`, `step-checker`, `use-section-requirements`, `interactive-conditional`, `useGuidePreviewProgress`, `useAlignmentReevaluation`, and the section's own dispatch.
- `progress-events.ts` moved to `src/global-state/` (Tier 1) so the engines tier can subscribe without an architecture exception.
- Orphan `step-auto-skipped` listener removed — no dispatcher existed anywhere in the repo.
- Bulk paths now dispatch per-step `kind: 'step'` events too: `markStepsCompleted` / `resetSteps` / `resetSection`. Keeps `interactive-conditional` and other `kind: 'step'` listeners reactive after objectives auto-complete and full-section resets.
- `persistSection` dispatches `interactive-progress-cleared` whenever the *guide total* falls to zero, so tail-reset and redo-first-step paths reach the alignment and preview-reset-button consumers without each having to hand-dispatch.
- **`interactive-progress-cleared` and `pathfinder-step-progress` survive** alongside `pathfinder:progress`. They still drive ephemeral preview / alignment UI and a remaining listener site that hasn't migrated. Folding them into `kind: 'guide'` with `hasProgress: false` is tracked as a separate follow-up. The contracts-tripwire test (`interactive-section.contracts.tripwire.test.tsx`) pins the current three-event surface.

### Tier hygiene

- Move `content-renderer.tsx` from `src/docs-retrieval/` to `src/components/content-renderer/`. `docs-retrieval/` is now a pure parser/fetcher engine; the React component lives in the UI tier where it belongs. Three architecture-test ratchet exceptions removed:
  - `docs-retrieval/content-renderer.tsx -> components`
  - `docs-retrieval/content-renderer.tsx -> integrations`
  - `docs-retrieval/content-renderer.tsx -> requirements-manager`
- The two lazy-lookup workarounds (`getStepTypeLookup`, `getInteractiveStepComponentTypes`) that existed only to sidestep the cycle have collapsed into plain top-level const maps now that the cycle is gone.
- `completion-store` and `section-registry` relocated to `src/global-state/` (Tier 1). Lets `requirements-manager` import them directly without a tier violation.
- `components/interactive-tutorial/get-content-key.ts` becomes a thin re-export of `global-state/content-key.ts`. Eliminates the split-brain risk where the store and the section hooks could resolve different content keys after any future typed-API migration.
- **`sectionCompletedCheck` moved out of `lib/dom`** into `src/requirements-manager/checks/section-completed-check.ts` with a static `import { sectionDoneStorage } from '../../lib/user-storage'`. The previous dynamic `await import('../user-storage')` in `lib/dom` existed only to dodge a JSDOM/Prism flake from pulling `@grafana/runtime` eagerly via the heavily-mocked lib/dom layer — the check was domain-aware all along and didn't belong in the DOM-utility tier.

### Authoring ergonomics

- **Unified `loadTab(tabId, url, options?)` entry point** over the two existing loaders. `initializeRestoredActiveTab`, `setActiveTab`, `reloadActiveTab`, `useContentReset`, `openLearningJourney`, `navigateToNextMilestone` / `navigateToPreviousMilestone`, and the `link-handler` "Ready to begin" / fallback milestone CTA all route through `loadTab`. The legacy `loadTabContent` and `loadDocsTabContent` methods remain as `@deprecated` internal implementations the dispatcher routes into via `shouldUseDocsLoader`. Folding them together is open follow-up.
- JSON schema gains an optional `id` on interactive blocks; the runtime parser plumbs it into `ParsedElement.props.stepId`, overriding the section's positional `${sectionId}-step-N` convention. This is the recommended path for stable `section-completed:` requirement targets.
- The runtime parser additionally tolerates camelCase aliases (`targetAction`, `refTarget`, `targetValue`) when set alongside the lowercase canonical form. Schema-level acceptance of camelCase-only inputs is a deliberate follow-up — the CLI's flag-generation contract depends on the strict canonical schema, so widening it touches more surface than this PR intends.
- **Stable standalone step IDs:** every interactive-block converter (interactive / multistep / guided / quiz / terminal / terminal-connect / challenge / code-block) now emits `props.stepId` as either the author-supplied `id` or a `deriveStepId(...)` content+position hash. Iteration sites supply distinct parent contexts (top-level, section, conditional branches, assistant-wrapped) so duplicate content across branches doesn't collide. Standalone steps in sectionless guides now persist across reloads.

### Section-completed gate (mount-independent)

- New `sectionDoneStorage` namespace in `lib/user-storage.ts` (two-state — `true` or absent — mirroring `sectionAcknowledgementStorage`).
- `interactive-section` writes the done bit on the existing `isCompleted` effect and clears it on the inverse; full-reset paths sweep it via `clearAckAndCollapseStorage`.
- `sectionCompletedCheck` (now in `requirements-manager/checks/`) reads `sectionDoneStorage` first, falling back to the legacy `#sectionId.completed` DOM check. The mount-independent storage path is a NEW capability layered on top of the existing DOM path, not a replacement — both paths are live by design. Storage-first works for unmounted virtualized sections, cross-milestone gates, and conditional branches that haven't rendered yet; DOM fallback covers the transitional window before the async storage write resolves.

### Documentation

- `docs/developer/STEP_MODEL.md` rewritten end to end to describe the post-collapse model, including the **"Reset paths must evict the cache"** invariant (added in `394bccf5`) and the **"Section-completed requirement gate"** subsection pointing to the new check location.
- `.cursor/rules/tracked-step-types.mdc` updated from "4 sites" to "2 sites".

## Why

The historical pipeline accumulated parallel state at every layer. Every new step type or completion path had to keep all of those in lockstep — and didn't, which is how `ChallengeBlock` ended up registered in two of the four step-type sites but missing from the other two. Building on top of the engine had four mental translations for the same field, two listener-event names for what should be one signal, a parser tier that secretly knew about UI components, an FSM that could lie to the store about skips, and a DOM-only section gate that silently broke any time a section wasn't currently mounted.

### Net effect

Honest accounting (not aspirational):

- **One canonical persistence store** for step completion at `src/global-state/completion-store.ts`. The FSM (`useStepChecker`) and `SequentialRequirementsManager` remain as in-memory orchestration mirrors that write through to the store on every terminal transition — completion AND reset. They are not a third source of truth; the bridge keeps them honest. Pinned by the `step-checker.store-bridge.test.ts` tripwire (8 cases across both axes).
- **Two sites for step-type metadata** (was four). `step-type-registry` is authoritative; `section-child-classifier` is the only other site. Parity is pinned by `content-renderer.registry-parity.test.ts`.
- **Unified `loadTab` entry point** over the two existing loaders (was two parallel paths dispatched at the call site). The `shouldUseDocsLoader` branch moved inside the unified entry point. Folding `loadTabContent` + `loadDocsTabContent` together is open follow-up.
- **One unified `pathfinder:progress` event** replaces the four legacy named events. `interactive-progress-cleared` and `pathfinder-step-progress` survive for ephemeral preview / alignment UI and are tracked for a separate fold.
- **Zero docs-retrieval tier-violation exceptions** (was three) and **zero `lib/dom -> user-storage` edges** (the `sectionCompletedCheck` relocation removed the dynamic-import laundering).
- **`section-completed:` requirements work mount-independent** via a storage-first read with DOM fallback. Both paths are live by design — the storage path adds a new capability rather than replacing the DOM path.
- **Standalone step persistence survives reloads** (parser emits stable IDs for every interactive block type).
- **FSM and store can't disagree on terminal transitions — completion AND reset.** Single write authority for manual / skipped / objectives / reset, enforced by the `writeStoreCompletion` / `writeStoreReset` bridges in `useStepChecker` and pinned by tripwires.

### Trade-offs you should look at directly

- **Net source delta is +935 lines**, not a reduction. The PR deletes `use-standalone-persistence.ts` (−161), shrinks `section-state.ts` (−70), `use-section-persistence.ts` (−48), and `interactive-section.tsx` (−45). It adds `completion-store.ts` (+635), the relocated `section-registry.ts` (+125), `step-id.ts` (+67), `content-key.ts` (+69), and `progress-events.ts` (+63). Net tests are +933 lines, of which roughly half pin the new abstractions (hydration race, eviction parity, cache TOCTOU, version counters, store bridge) and half pin existing user-visible behaviour.
- **API surface grew.** `STANDALONE_SECTION_ID`, `useStepCompletion`, `useSectionCompletion`, `markStepCompleted`, `resetStep`, `markStepsCompleted`, `resetSteps`, `resetSection`, `evictSectionCache`, `evictContentCache`, `evictAllContentCaches`, `getGuideProgress`, `subscribeProgress`, `dispatchProgress`, `subscribeProgressEvent`, `ProgressEventDetail`, `ProgressReason`, `deriveStepId`, `StepContext`, `getContentKey`, `setActiveTabUrl`, `setContentKeyOverride`, `sectionDoneStorage`. The pre-PR persistence surface was a 3-effect hook.
- **Three of the original four commits were gap-closes for problems the collapse introduced.** Commit 2 fixed a hydration race the original design had; commit 3 closed the FSM-store skip / objectives divergences plus added `sectionDoneStorage` and stable IDs; commit 4 fixed the Reset-Guide cache eviction regression that the move from component lifetime to module scope created. The cache-eviction invariant is now documented as load-bearing in `STEP_MODEL.md`.

## Review pushbacks addressed

Static review surfaced six pushbacks. Status:

- ✅ **FSM ↔ store reset divergence.** `useStepChecker.resetStep` now writes through to the store via `writeStoreReset` (`394bccf5`). The "FSM and store can't disagree" claim now holds on both completion AND reset. Three new tripwires in `step-checker.store-bridge.test.ts` pin the contract.
- ✅ **Skip-reason hardcoded to `'manual'`.** `persistCompletion` now plumbs a `ProgressReason`, and the skip path passes `'skipped'` (`2dfb95ec`). A regression test on the quiz pins it. The mock-blind-spot that let this land is fixed too.
- ✅ **`dom-utils.ts` dynamic-import tier hack.** `sectionCompletedCheck` relocated to `src/requirements-manager/checks/section-completed-check.ts` with a static `user-storage` import (`394bccf5`). Architecture ratchet unchanged at vertical=2.
- ✅ **Reset-Guide cache invariant undocumented.** `STEP_MODEL.md` now has a "Reset paths must evict the cache" subsection listing the four reset sites and the eviction API (`394bccf5`).
- ✅ **Hydration race overwrites persisted completions** (Tom's HIGH). Closed in `7efd8d8` — `hydrationClears` + `'all'` sentinel + `.then` re-persist on resolved snapshot.
- ⚠ **Line-count framing in commit 1's message** ("~290 fewer lines of source"). The commit message can't be edited without rewriting history; the honest +935 breakdown is in the "Trade-offs you should look at directly" section above instead.
- ⚠ **`loadTab` framing.** Reframed throughout this description as "unified entry point" (not "one tab loader"). Actually folding `loadTabContent` + `loadDocsTabContent` is tracked as open follow-up.
- ⚠ **`step-id.ts` add/delete/re-add cycle.** Left in the commit history; not squashing to preserve the existing review threads on those commits. Future archaeology will show the file landing in `components/interactive-tutorial/`, getting declared dead, and coming back at `src/global-state/`.

## Test plan

- [x] `npm run check` is green (typecheck + lint + prettier + go + tests). Locally: **230 suites, 3874 tests pass** (+9 from review pushbacks: 3 reset-bridge tripwires, regression coverage on the skip-reason path, plus earlier follow-ups).
- [x] `npm run build` smoke runs cleanly.
- [ ] Manual playwright on a sectionless guide: complete a standalone step, reload, confirm persistence (C1 surface).
- [ ] Manual playwright on `prometheus-grafana-101` (section-heavy): complete several steps, reload, redo a step, confirm the #842 ack gate re-fires (C2 surface).
- [ ] Block-editor preview mode: trigger reset, confirm the "Reset guide" button reacts; remount under the same preview key starts fresh (C3 + the `evictSectionCache` path).
- [ ] All 7 shipped bundled guides render correctly (C5 + C6 smoke).
- [ ] Spot-check that `subscribeProgressEvent` listeners fire on the right `kind`:
  - `useAlignmentReevaluation` flips on `kind: 'guide'`
  - `useGuidePreviewProgress` flips on `kind: 'guide'`
  - `interactive-conditional` re-evaluates on `kind: 'step'` (exists-reftarget)
  - `step-checker` rechecks on `kind: 'section'` (`section-completed:` requirements)
- [ ] **Skip a skippable step with failing requirements → reload → still marked complete, with `reason: 'skipped'`** (FSM-store bridge + skip-reason fix).
- [ ] **Reset a completed standalone step → reload → still cleared** (FSM-store reset bridge).
- [ ] **Standalone interactive in markdown → complete → reload → still complete** (parser-derived stable IDs).
- [ ] **`section-completed:` requirement on a step that lives on a later milestone → satisfies before the gating section is scrolled into view** (storage-backed gate).
- [ ] **Reset first step before hydration finishes (throttle storage) → progress stays cleared on next render** (hydration race fix).
- [ ] **Reset guide → in-memory subscribers immediately re-render as not-completed** (cache eviction fix).
- [ ] **Package journey: next/prev milestone + "Ready to begin"** routes through `loadTab` and re-resolves the manifest (loadTab migration).

## Storage compatibility

The completion store reads/writes the existing `interactiveStepStorage` namespace. No localStorage migrations needed for it; existing user progress is preserved. The store treats an empty completed set as an explicit `clear()` on the storage key (rather than a `setCompleted({})` marker) — `hasProgress` and reset-detection stay clean.

The new `sectionDoneStorage` namespace is additive — users without an entry fall back to the DOM check on first encounter, then the section's `isCompleted` effect populates the bit going forward. No backfill needed.

Standalone steps that previously persisted under `standalone-step-N` (anonymous-counter scheme) will have their old keys orphaned in storage. This is a deliberate trade-off — the counter scheme was already losing progress on every reload, so net effect is neutral or positive for users. New, stable IDs going forward. Debug-time confusion (storage inspector showing stale keys) is acknowledged.

## Commits

1. `93793de` — original collapse (state ownership, events, tier hygiene, loadTab, JSON `id`, docs).
2. `7efd8d8` — close PR-review gaps: hydration race, unified `getContentKey`, bulk progress events, complete `loadTab` migration; medium/low fixes (legacy `interactive-progress-cleared` from tail-reset, post-write `allStepsCompleted` derivation, delete dead `deriveStepId` / `syncSectionCompletionCache`).
3. `8415056` — finish state-collapse: relocate completion store to Tier 1, FSM → store bridge (completion direction), `sectionDoneStorage` + storage-first `section-completed:`, parser-emitted stable step IDs.
4. `964fa56` — Reset guide also evicts the completion-store in-memory cache. Closes the regression where Reset-Guide cleared localStorage but left subscribers rendering "completed" until they remounted.
5. `2dfb95ec` — Preserve `'skipped'` reason on user skip. Closes Tom's MINOR: `persistCompletion` now plumbs a `ProgressReason` and the skip paths pass `'skipped'` (quiz / multi-step / guided).
6. `d10b4d7f` — Prettier reflow follow-up to `2dfb95ec`. No behaviour change.
7. `394bccf5` — Close PR-#909 review pushbacks: FSM → store reset bridge, relocate `sectionCompletedCheck` to requirements-manager, document the cache-eviction invariant.

## Out of scope

- Schema-level acceptance of camelCase-only JSON guides — would force the CLI flag-generator + Commander help-text rewrite; deferred.
- Folding `loadTabContent` + `loadDocsTabContent` into a single loader implementation — `loadTab` unifies the entry point; the underlying split stays for now.
- Migrating `interactive-progress-cleared` and `pathfinder-step-progress` into `pathfinder:progress` — three event surface kept by design for ephemeral preview / alignment UI.
- Collapsing the FSM and `SequentialRequirementsManager` into the store — they remain as orchestration mirrors; the bridge keeps them honest.
- Component-side anonymous step counters — now last-resort fallback only (parser supplies `stepId` for every JSON-sourced step). Counters are kept for test harnesses that mount step components directly without going through the parser.


tomglenn (dismissed):
Couple changes needed:
## 🐛 Hydration race overwrites persisted completions (HIGH)

  `src/components/interactive-tutorial/completion-store.ts:87-112` + `178-202`

  `ensureHydrated` kicks off an async storage read and returns immediately. If `markStepCompleted`
   is invoked between mount and hydration resolution:

  1. In-memory `bySteps` is empty.
  2. Line 188 writes the new step into the empty map.
  3. Line 190 → `persistSection` → `setCompleted(contentKey, sectionId, {newStepId})` — a **full
  overwrite**, not a merge (`user-storage.ts:1043`: `storage.setItem(key,
  Array.from(completedIds))`).
  4. Hydration's `.then` resolves (lines 95-107), merges stored IDs back into memory, but **never
  re-persists**.
  5. On the next reload, the previously-persisted step IDs are gone.

  The storage backend is `createUserStorage()` (Grafana user storage — round-trips the backend,
  not synchronous localStorage), so the race window is realistically tens to hundreds of ms —
  plenty for a fast-clicking user to lose progress.

  The deleted `use-standalone-persistence.ts` explicitly handled this: its restore callback merged
   storage with a live ref *and* re-persisted when the user had completed during the window. That
  protection isn't carried into the new store.

  **Fix options** (any one):
  - In `ensureHydrated`'s `.then`, call `persistSection(contentKey, sectionId)` whenever `changed`
   is true.
  - Have `persistSection` union with storage before writing, until hydration resolves.
  - Gate writes on a hydration `Promise` so `markStepCompleted` awaits hydration before
  persisting.

  **Test gap:** `completion-store.test.tsx` covers reads-before-hydration but not
  writes-before-hydration. Suggest adding a regression test that mocks slow storage and asserts
  pre-existing IDs survive a pre-hydration write.

  ---

  ## Skip reason reported as `'manual'` (MINOR)

  `src/components/interactive-tutorial/interactive-quiz.tsx:148-150, 385-393`

  `handleSkip` calls `persistCompletion()`, which hardcodes `reason: 'manual'` when calling
  `markStepCompleted`. The dispatched `pathfinder:progress` event therefore carries `reason:
  'manual'` even though the user pressed Skip.

  No live consumer reads `detail.reason` off the event today, and `isStepSkipped`
  (`step-state.ts:207`) reads from the separate step-checker reducer — so this is currently
  dormant. But the new unified event lies about intent, which makes it a trap for future consumers
   that switch on `reason`.

  **Fix options:**
  - Plumb a reason through `persistCompletion(reason: CompletionReason = 'manual')` and pass
  `'skipped'` from `handleSkip`.
  - Or drop `'skipped'` from `ProgressReason` in `progress-events.ts` and document that the event
  carries intent loosely.

moxious:
# Adversarial review synthesis — PR #909

_Synthesis of two independent reviews (R1, R2) plus targeted validation against the actual diff. Findings are ranked by `blast × likelihood × (1 / reversibility after merge)`, not by prior-reviewer labels._

## 1. What this PR actually does

Collapses six parallel "is step done?" sites into one Tier-1 store (`src/global-state/completion-store.ts`), shrinking `SectionState`'s reducer from 6 actions to 3 (it now owns only `{ acknowledged }`). Adds a single `pathfinder:progress` event with `kind: step|section|guide` replacing four legacy event names. Renames `InteractiveElementData` fields from lowercase to camelCase across the request pipeline. Moves `content-renderer.tsx` out of `docs-retrieval/` to satisfy tier rules. Net `-290` source lines, real consolidation — but the safety net used in the prior high-risk refactor (#885) of "tripwires first, refactor second" was not used here: commit 2 (`7efd8d8`) is a list of four post-hoc defect fixes caught in static review, with tests written *for* those fixes.

## 2. Must fix before merge

### MF-1 — Author-stable step IDs are dropped at the renderer
- `src/components/content-renderer/content-renderer.tsx:1025-1051`: every `interactive-*` case spreads an explicit prop list and **none of them forward `element.props.stepId`** from the parser.
- `src/components/interactive-tutorial/interactive-section.tsx:233`: `stepComponents` memo synthesises `${sectionId}-${schema.idPrefix}-${stepIndex+1}` — purely positional — and `toEnhancedProps` (`step-type-registry.ts:121`) clones this positional id over the child after.
- **Failure mode:** Author writes blocks `id:"create-ds"`, `id:"add-panel"`. User completes step 1 → storage records `${sectionId}-step-1`. Author inserts a new block in front. On reload, `create-ds` now occupies index 2; the user's completion stays glued to whatever block now sits at index 1. The headline "stable IDs" feature is structurally unreachable; `step-id.ts`'s `deriveStepId` fallback is also unreachable.
- **Fix scope (S):** Forward `stepId={element.props.stepId}` from each step case in `content-renderer.tsx` (~8 lines). In `interactive-section.tsx:233`, prefer the child prop: `const stepId = (child as any).props?.stepId ?? \`${sectionId}-${schema.idPrefix}-${stepIndex + 1}\``.
- **Test sketch:** Render a section whose first child has an explicit `stepId="create-ds"`. Call `markStepCompleted("create-ds", sectionId, "manual")`. Assert the rendered step shows "completed". Then re-render with a synthetic step inserted before it; assert the original step still shows "completed".

### MF-2 — `getGuideProgress` inflates past 100% on author edits; no roster filter anywhere
- `src/components/interactive-tutorial/section-state.ts:77-115`: reducer no longer filters by roster. The deleted "filters out step ids that are not in the current roster" test in `section-state.test.ts` has no replacement.
- `src/global-state/completion-store.ts:323-332` `getGuideProgress` calls `interactiveStepStorage.countAllCompleted(contentKey)` (`src/lib/user-storage.ts:1173`), which scans localStorage by prefix and never consults the roster. `ensureHydrated` writes the snapshot back as-is.
- **Failure mode:** Guide v1 (sections A=3 steps, B=2 steps) → storage holds 5 IDs. Author ships v2 deleting section B and renaming an A step's `refTarget` (which feeds `deriveStepId`). On reload: total = 3, count = 5, **percentage = 167%**. Surfaces in the progress chip via `use-document-step-progress.ts:70` and in learning-paths progress cards. Stable IDs (MF-1, once fixed) make this *strictly worse* — IDs survive identity-preserving edits but orphan silently on rename. There is no migration path; orphans persist indefinitely.
- **Fix scope (M):** Roster-aware reconciliation in `ensureHydrated` — drop snapshot IDs absent from the section's parser-provided roster, then `persistSection` the filtered set. Self-heals on first read.
- **Test sketch:** Seed storage with 5 step IDs under `contentKey`. Register the section with a 3-id roster. Call `ensureHydrated`. Assert `countAllCompleted(contentKey) === 3` and `getGuideProgress(...).percentage <= 100`.

### MF-3 — Preview "Reset guide" button is unreachable from a cold session
- `src/global-state/completion-store.ts:203-217`: preview mode suppresses both storage writes AND `kind:'guide'` dispatch (`percentage = isPreview ? undefined : ...`; gated on `percentage !== undefined`).
- `src/hooks/useGuidePreviewProgress.ts:37-41`: only listens for `kind:'guide'`. Mount-time `interactiveStepStorage.hasProgress()` always reads false because writes are suppressed.
- `src/pages/BlockEditor/BlockEditorHeader.tsx:553`: button gated on `hasPreviewProgress`.
- **Failure mode:** Open block editor → preview tab → complete a step → Reset button never appears, even after page reload. The contracts tripwire test even comments that the section event is what the editor "can react to," but the hook ignores section events.
- **Fix scope (S):** In `useGuidePreviewProgress.ts:37-41`, accept `kind:'step'|'section'` when `getContentKey()` matches `progressKey` and set `hasProgress=true`. One conditional.
- **Test sketch:** Render `BlockEditorHeader` in preview mode with no mocked `hasProgress`. Dispatch a real `pathfinder:progress` event with `kind:'step'` and matching contentKey. Assert the Reset button is in the DOM.

### MF-4 — `evictContentCache` does not protect against in-flight hydration _(novel — neither review caught this)_
- `src/global-state/completion-store.ts:489-510` deletes `entries`, `hydratedSections`, and `hydrationClears` for the key.
- `src/components/docs-panel/hooks/useContentReset.ts:56-64` calls `clearAllForContent(...)` then `evictContentCache(...)` synchronously.
- **Failure mode:** User opens a guide with prior completions; `ensureHydrated` is in flight (storage read pending). User clicks "Reset guide" before it resolves: storage is cleared, eviction wipes `hydrationClears`. The pending `.then` resolves with the pre-clear snapshot, `cleared?.has(stepId)` returns false (the marker was wiped), and the stale set is re-inserted into `entries` + `notify(contentKey)` fires. UI shows "complete" until next remount; subsequent `markStepCompleted` re-persists the resurrected set, undoing the reset on disk too. The `noteHydrationClear` protection on `resetStep`/`resetSection` was designed for this race but the eviction path bypasses it entirely.
- The new test at `completion-store.test.tsx:309` `await flushMicrotasks()` before evicting — so the in-flight window is untested.
- **Fix scope (S/M):** Either (a) `evictContentCache` should record a wildcard clear via `noteHydrationClear(contentKey, '*', 'all')` semantics, or (b) `ensureHydrated`'s `.then` should re-check `hydratedSections.has(key)` and bail if the entry was evicted while pending.
- **Test sketch:** Pre-seed storage with 2 step IDs. Start `useStepCompletion` render (kicks `ensureHydrated`). Without flushing, call `clearAllForContent` + `evictContentCache`. THEN flush. Assert `entries.get(contentKey)?.bySteps` is empty and `notify` did not re-insert stale data.

## 3. Fix in follow-up

| # | Finding | Why defer |
|---|---|---|
| F-1 | All-passive-only guide reads 0% | The 1+1 example resolves to 100% (numerator and denominator both exclude all-passive). The genuine bug — all-passive *only* — is rare and visible to author, not user. Fix with MF-2's roster work to avoid a second migration. |
| F-2 | No integration test pins section-child re-render on store flip | Production wiring uses `useSyncExternalStore` correctly. The gap is regression safety, not a live bug. Add a 10-line probe test next pass. |
| F-3 | `youtube-video-renderer.tsx:62-63` still reads window globals directly | Bypasses the typed setter the rest of the codebase migrated to. Tier-coupling smell, not a defect. Migrate during the next global-state pass. |
| F-4 | `.cursor/rules/systemPatterns.mdc:49,106`, `.cursor/rules/tracked-step-types.mdc:6,36`, `docs/developer/CONTEXT_INDEX.md:21`, `docs/developer/engines/interactive-engine.md:242` point at old `src/docs-retrieval/content-renderer.tsx`; new `STEP_MODEL.md` is orphaned from CONTEXT_INDEX | Doc-drift, ~6-line edits. Bundle with MF-1's docs touch or run `/maintain-docs`. |
| F-5 | `markStepCompleted` additive-merge contract is comment-only | Invariant currently holds; risk is a future regression. Add one test pinning "mark issued during pending hydration survives the resolve." |
| F-6 | `interactive-multi-step.tsx:65` types `data: any` | The casing flip itself is a quiet bug *fix*, not a new defect (pre-flip, `targetAction`/`refTarget` arrived as `undefined`). Tighten the parameter type to `InteractiveElementData` so future drift is a compile error. |

## 4. Dismissed findings

- **`resetStep` asymmetric with `resetSection` (R2, HIGH).** Self-heals. `noteHydrationClear` runs *before* the early-return at `completion-store.ts:301-303`; the resolver closes over `contentKey`/`sectionId` as string primitives, so navigate-away doesn't break it. The `hadClears=true` branch in the resolver calls `persistSection` and the cleared ID is correctly absent. The reviewer hedged ("the window is real but narrow") — under current code there is no window. Rejected as HIGH; at most a code comment is warranted.
- **`interactive-progress-cleared` semantics narrowed (R2, HIGH).** Misread of the change. The narrowing applies to the new backstop in `persistSection:228-232`. **Every manual reset surface** (`interactive-section.tsx:1036-1043`, `useContentReset.ts:69-73`, `useGuidePreviewProgress.ts:64-68`, `MyLearningTab.tsx:374-378`) still dispatches unconditionally and was not touched. The narrowing is the FIX — it prevents a false "progress cleared" signal while another section retains progress. Rejected.
- **Multistep casing flip is risky (R2, MEDIUM).** Backward framing. `checkRequirementsFromData` in `interactive.hook.ts:213-242` only ever read camelCase; pre-flip those fields arrived as `undefined`. Post-flip is correct. The MEDIUM rating belonged to the *prior* state, not this PR.
- **Contracts tripwire preview relaxation (R2, MEDIUM).** The suppression of `kind:'guide'` in preview is intentional and documented in code (`percentage = isPreview ? undefined : ...`). The test's missing *negative* pin is a one-line tripwire gap, not a contract relaxation.

## 5. Novel findings (neither review caught)

### N-1 — `evictContentCache` hydration race
Covered as **MF-4** above. Block-listed because: (a) "Reset guide" is a top-level affordance, (b) the race is reachable in a single user action sequence, (c) the resurrection is **silent** — no error, no telemetry, and the next write re-persists the resurrected state, making the reset durably ineffective.

### N-2 — Two-tab progress corruption: no `'storage'` event subscription
Verified: `git grep -n "addEventListener.*storage"` in `src/global-state/` returns nothing. Module-scope `entries` in `completion-store.ts` and the `completedCountCache` in `interactiveStepStorage` are populated per-tab and never invalidated by cross-tab writes. Realistic scenario: a user has the sidebar open in one Grafana tab and the block-editor preview in another (a documented workflow). Tab A's `markStepCompleted` writes its stale view back over Tab B's authoritative reset — Tab B's clear is silently undone. There's no log, no error, no UI indicator. Either (a) add a `'storage'` listener that selectively `evictContentCache`s when localStorage keys with the `INTERACTIVE_STEPS_PREFIX{contentKey}` prefix change, or (b) read-through to storage on every read (loses the perf win). Test sketch: dispatch a synthetic `StorageEvent` with `key = 'pathfinder.interactive.steps:guide-x:section-y'` and assert `entries.get('guide-x')?.bySteps.get('section-y')` is evicted.

### N-3 — Storage-quota path is silent to the user
`setCompleted` swallows `QuotaExceededError` with `console.warn`. If a heavy-user environment hits quota mid-guide, completions stop persisting; on reload everything reverts. No surface to the user, no recovery prompt. Not a blocker, but P0-debug-six-months-out exactly the kind of failure that takes a week to diagnose. Worth a `recovery/` integration in a follow-up.

## 6. Open questions for the author

1. **Stable IDs (MF-1):** Is the missing `stepId` forward at `content-renderer.tsx:1025-1051` a known not-yet-wired step, or was the renderer expected to forward all parser-provided props? The `step-id.ts` module and `interactive-step.tsx:125-126` JSDoc claim end-to-end stability the renderer doesn't provide.
2. **Roster filtering (MF-2):** Where do you intend reconciliation to live — `ensureHydrated`, `getGuideProgress`, the storage adapter, or a one-shot migration pass on first read? The deleted reducer logic had it inside `RESTORE`; nothing has it now.
3. **Multi-tab (N-2):** Is "the same guide open in two tabs" a supported workflow? If not, document the constraint; if yes, the cross-tab cache invalidation is a hole.
4. **Tripwire pattern (process):** #885 landed tripwires in a separate commit before the refactor. This PR rewrote tripwires inside commit 1 to track the new shape. Was that intentional, or worth retroactively pinning the old contracts in a frozen test before any future refactor?

Jayclifford345:
Block-list fixes landed in `8927f3b1`. Adversarial review surfaced four must-fix-before-merge bugs; this commit closes them plus the F-4 doc-drift cleanup.

## In this commit

| Finding | Fix | Tests |
| --- | --- | --- |
| **MF-1** Author-stable step IDs forwarded end to end | `content-renderer.tsx` forwards `element.props.stepId` to all 8 step cases; `interactive-section.tsx` `stepComponents` memo prefers child's `stepId` over the positional fallback so `cloneElement` no longer overwrites it | +3 tripwires in `interactive-section.contracts.tripwire.test.tsx` (prefers author ID, survives sibling insertion, positional fallback when absent) |
| **MF-2** Roster reconciliation + percentage clamp | New `reconcileSection(sectionId, roster)` in `completion-store.ts` (self-heals on first mount); `getGuideProgress` clamps to `<= 100` as a defensive ceiling | +3 in `completion-store.test.tsx` (drops orphans, no-op when aligned, 5/3 = 100 not 167) |
| **MF-3** Preview "Reset guide" button reactive | `useGuidePreviewProgress` accepts `kind: 'step' \| 'section'` events whose active content key matches the progress key (preview mode suppresses `kind: 'guide'`) | +3 in `useGuidePreviewProgress.test.ts` (step event flips, section event flips, cross-key ignored) |
| **MF-4 / N-1** evictContentCache race with in-flight hydration | `ensureHydrated.then` bails when the section's hydration marker was wiped by an eviction between scheduling and resolution | +1 race-window tripwire (no `flushMicrotasks` before evict) |
| **F-4** Doc drift to renamed `content-renderer.tsx` path | 4 docs/rules files updated to `components/content-renderer/content-renderer.tsx`; `STEP_MODEL.md` added to `CONTEXT_INDEX`; `tracked-step-types.mdc` step-7 calls out the `stepId` forwarding requirement | -- |

`npm run check` green. 230 suites / 3884 tests (+10 vs base).

## Deferred to a follow-up PR after merge

Captured here so the next branch starts from the same context:

- **N-2** cross-tab progress corruption — module-scope `completion-store` cache survives `localStorage` writes from another tab; no `'storage'` listener. Needs a product call first on whether "same guide in two tabs" is a supported workflow before adding the listener.
- **N-3** silent `QuotaExceededError` — pre-existing behavior in `user-storage.ts`; surface a one-time `alert-warning` toast in the `recovery/` integration pass.
- **F-1** all-passive-only guide reads 0% — fold into the next numerator/denominator change to avoid a second migration.
- **F-2** section-child re-render probe — regression safety only; production wiring is correct.
- **F-3** `youtube-video-renderer.tsx` reads `(window as any).__DocsPlugin*` — migrate to the typed setter alongside the next `global-state` pass.
- **F-5** `markStepCompleted` additive-merge tripwire — invariant holds today; pin it.
- **F-6** `interactive-multi-step` `data: any` → `InteractiveElementData` (compile-time guard against future casing drift).
- **N-4** STEP_MODEL.md caveat: stable IDs survive position / content edits but type changes still orphan (because `deriveStepId` doesn't include the type).
- **N-5** `evictContentCache` locality JSDoc: subscribers capture `getContentKey()` at hook-render time.

## Reviewer answers to open questions (from the adversarial synthesis)

1. **MF-1** stepId forward — yes, the renderer was expected to forward all parser-provided props. Fixed.
2. **MF-2** roster filtering — reconciliation lives in `reconcileSection(sectionId, roster)` (Tier 1, called from `InteractiveSection` once roster is known). `ensureHydrated` is too early (no roster yet); `getGuideProgress` is too late (orphans persist in storage). The clamp in `getGuideProgress` is belt-and-braces for the pre-reconcile window.
3. **Multi-tab** — deferred, see N-2 above.
4. **Tripwire pattern** — agreed. Worth retroactively pinning the old contracts in a frozen test before any future refactor; flagged for the follow-up PR.

moxious (approved):

PR #912, opened 2026-05-22
fix(interactive): cross-tab completion-store sync (N-2 follow-up)

Jayclifford345:
## Summary

Follow-up to [#909](https://github.com/grafana/grafana-pathfinder-app/pull/909) closing the N-2 deferred finding: when the same guide is open in two browser tabs, each tab has its own module-scope completion-store cache but they share localStorage. Without cross-tab sync, tab B's stale cache could silently write back over tab A's authoritative reset (or vice versa) with no error, log, or UI indicator.

https://github.com/user-attachments/assets/4cb54c68-719f-43b7-b986-e18c8eaa2c69

## What changed

- **Module-scope `storage` event listener** in `src/global-state/completion-store.ts`. Fires in every OTHER tab on the same origin on localStorage mutation. Handles `event.key === null` (cross-tab `localStorage.clear()`) as a global evict; matches `INTERACTIVE_STEPS_PREFIX` keys against the active set (entries plus hydratedSections), then `evictSectionCacheForKey` + `invalidateCountCache` + `notify`.
- **Hydration-generation counter** (`hydrationVersion`). Snapshotted by `ensureHydrated` at schedule time, re-checked in the `.then` handler. Strictly supersedes the old `!hydratedSections.has(key)` race guard: a cross-tab eviction followed by a fresh re-hydration cycle re-adds the key before the original `.then` runs, so the `has(key)` check alone passes and stale data would merge. The version check catches this.
- **`evictSectionCacheForKey(contentKey, sectionId)`** — extracted explicit-key variant. The cross-tab event may carry a key that isn't the tab's currently active content key. `evictSectionCache(sectionId)` becomes a thin wrapper.
- **`interactiveStepStorage.invalidateCountCache(contentKey)`** export in `src/lib/user-storage.ts`. The per-tab `completedCountCache` was already invalidated inline by `setCompleted` / `clear` / `clearAllForContent` / `clearAll`. Exposed publicly so the cross-tab handler can drop the stale numerator without touching localStorage.

## Tests

Four new tripwires in `src/global-state/completion-store.test.tsx` (\`cross-tab sync\` describe block):

1. Storage event evicts the in-memory section cache; subscriber re-renders against authoritative storage.
2. Storage events for unrelated localStorage keys are ignored (early-return).
3. \`event.key === null\` triggers a global evict.
4. Stale in-flight hydration is dropped when a cross-tab storage event triggers re-hydration (pins the \`hydrationVersion\` contract).

## Docs

New \"Cross-tab synchronization\" subsection in [docs/developer/STEP_MODEL.md](docs/developer/STEP_MODEL.md) documenting the listener, the hydration-version invariant, and the best-effort semantics (last-write-wins; no merge of conflicting changes).

## Test plan

- [x] \`npm run check\` clean: 230 suites / 3902 tests pass (+4 vs base).
- [ ] Manual: open the same guide in two browser tabs, complete a step in tab A, click \"Reset guide\" in tab B, return to tab A, scroll the section, confirm tab A reflects the reset on next render.
- [ ] Manual: complete a step in tab A, then in tab B click reset, complete a different step in tab A, confirm tab B does not silently resurrect tab A's first completion on next render.

## Risk and reversibility

- Storage listener fires on every cross-tab localStorage write including unrelated keys. The early-return on prefix mismatch is O(1).
- Hydration-version check strictly supersedes the prior MF-4 guard; same correctness, broader coverage.
- \`invalidateCountCache\` export is additive; no existing callers affected.
- Reversible by reverting the commit. No storage format changes.

## Deferred follow-ups from PR #909 (remaining)

- N-3 silent QuotaExceededError
- F-1 all-passive guide reads 0%
- F-3 youtube window globals
- F-5 markStepCompleted additive-merge tripwire
- F-6 multi-step \`data: any\` tighten
- F-2 section-child re-render probe
- N-4 STEP_MODEL caveat on type-change orphans
- N-5 evictContentCache locality JSDoc

These will be addressed in subsequent follow-up PRs.

### DIFF
--- a/docs/developer/STEP_MODEL.md
+++ b/docs/developer/STEP_MODEL.md
@@ -53,6 +53,23 @@ Known reset sites (each pairs the storage clear with a cache eviction):
 
 Adding a new reset path: pair the storage clear with the corresponding eviction.
 
+## Cross-tab synchronization
+
+The completion store's caches (`entries`, `hydratedSections`, `hydrationVersion`, ...) live in the module instance for the current browser tab. `localStorage` is shared across every tab on the same origin. Without cross-tab sync, tab B's stale cache could silently write back over tab A's authoritative reset.
+
+A module-scope `storage` event listener (installed via `installCrossTabSync` at module init) reacts to cross-tab writes:
+
+1. `event.key === null` — another tab called `localStorage.clear()`; drop every in-memory cache via `evictAllContentCaches`.
+2. `event.key.startsWith(StorageKeys.INTERACTIVE_STEPS_PREFIX)` — another tab wrote a `(contentKey, sectionId)` slot; resolve the contentKey against the live set of active keys (`entries` ∪ `hydratedSections`), then `evictSectionCacheForKey` + `notify` so the subscriber re-hydrates from authoritative storage on the next render.
+
+A per-section monotonic `hydrationVersion` counter closes the in-flight hydration race: every cache-clearing path bumps the version, and `ensureHydrated` snapshots the version at schedule time. When the storage read resolves, a mismatch indicates the cycle was invalidated (by an eviction or a fresh re-hydration kicked off by the listener) and the merge is dropped. This strictly supersedes the `!hydratedSections.has(key)` race guard — the version check also catches the case where a new `ensureHydrated` cycle has already re-added the key before the old `.then` runs.
+
+Limits — best effort, not transactional:
+
+- Last-write-wins. No merge of conflicting changes between tabs.
+- Tab A writing immediately before tab B writes the same key follows browser-defined ordering; the listener fires after-the-fact in each tab.
+- The `completedCountCache` in `lib/user-storage.ts` is also per-tab — the listener invalidates it via `interactiveStepStorage.invalidateCountCache(contentKey)` so the next `getGuideProgress` re-scans storage.
+
 ## Section reducer — minimal state
 
 `section-state.ts` owns one bit: `acknowledged: true | null` (the #842 gate). Cursor is a pure derivation of `(stepIds, completedSet)` via `computeCursor(stepIds, completed)`; the completion set comes from `useSectionCompletion`. Reducer actions:
--- a/src/global-state/completion-store.test.tsx
+++ b/src/global-state/completion-store.test.tsx
@@ -503,6 +503,55 @@ describe('completion-store', () => {
       expect(screen.getByTestId('completed').textContent).toBe('false');
     });
 
+    it('matches exact (contentKey, sectionId) — does not misroute when one content key is a prefix of another', async () => {
+      // Regression for the prefix-collision bug in the original
+      // listener: when one active content key (`bundled:loki-101`)
+      // is a hyphen-delimited string prefix of another
+      // (`bundled:loki-101-extended`), `stripped.startsWith(short + '-')`
+      // matched the shorter key first and evicted the wrong pair,
+      // leaving the longer guide's cache stale until the next mount.
+      const SHORT_KEY = 'bundled:loki-101';
+      const LONG_KEY = 'bundled:loki-101-extended';
+
+      setActiveTabUrl(SHORT_KEY);
+      storedCompleted.set(`${SHORT_KEY}-section-x`, new Set(['step-1']));
+      const short = render(<StepProbe stepId="step-1" sectionId="section-x" />);
+      await flushMicrotasks();
+      expect(short.getByTestId('completed').textContent).toBe('true');
+      short.unmount();
+
+      setActiveTabUrl(LONG_KEY);
+      storedCompleted.set(`${LONG_KEY}-section-y`, new Set(['step-2']));
+      const long = render(<StepProbe stepId="step-2" sectionId="section-y" />);
+      await flushMicrotasks();
+      expect(long.getByTestId('completed').textContent).toBe('true');
+
+      // Tab A clears LONG_KEY's progress; storage event fires here for
+      // the LONG key only. Exact matching evicts the LONG pair; prefix
+      // matching would have evicted a non-existent `(SHORT_KEY, "extended-section-y")`
+      // and left the LONG cache untouched.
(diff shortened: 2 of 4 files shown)


moxious (changes requested):
## Overview

Closes the N-2 deferred finding from #909: when the same guide is open in two tabs, each tab has a per-module completion-store cache while sharing `localStorage`. Without sync, tab B's stale cache silently overwrites tab A's authoritative reset.

Three load-bearing changes:

1. Module-scope `storage` event listener (`installCrossTabSync` at module init) that translates cross-tab writes into local cache evictions + subscriber notifications.
2. Per-section `hydrationVersion` counter, snapshotted in `ensureHydrated` and re-checked in the `.then` handler, so any in-flight hydration whose cycle was invalidated drops its merge instead of resurrecting stale data.
3. New `evictSectionCacheForKey(contentKey, sectionId)` (explicit key) and `interactiveStepStorage.invalidateCountCache(contentKey)` (drop per-tab numerator without touching storage).

Plus 4 new tripwires and a STEP_MODEL.md subsection.

## What's good

- **Hydration-version invariant is the right primitive.** The previous `!hydratedSections.has(key)` guard genuinely failed in the cross-tab path because the listener triggers a fresh `ensureHydrated` cycle that re-adds the key before the old `.then` runs. Bumping a monotonic counter on every cache-clearing path and snapshotting at schedule time is the textbook fix. The double-clause `!hydratedSections.has(key) || currentVersion !== expectedVersion` is belt-and-suspenders but harmless.
- **`evictAllContentCaches` bump-before-clear ordering is subtle and correct.** Clearing `hydrationVersion` outright would let a fresh `ensureHydrated` cycle's snapshot-of-0 match a stale `.then`'s captured-0 and silently merge. The comment at `completion-store.ts:736–741` calls this out — keep it.
- **Preserves the storage format.** No serialization changes, listener is additive, fully reversible.
- **Tests cover the right things.** All four tripwires are well-targeted: in-tab eviction propagation, unrelated-key ignore, `event.key === null` global evict, and the in-flight hydration race (which is the one the version invariant was actually invented for).

## Issues and risks

### 1. Content-key disambiguation is prefix-based, which can misroute when one active key is a prefix of another (`completion-store.ts:805–815`)

The listener strips the `INTERACTIVE_STEPS_PREFIX` and then iterates `activeContentKeys`, accepting the first one where `stripped.startsWith(contentKey + '-')`. Content keys can contain hyphens (URLs, bundled paths), so if two active content keys are e.g. `bundled:loki-101` and `bundled:loki-101-extended`, and the event key is `<PREFIX>bundled:loki-101-extended-section-x`, both match. The first iteration order wins, and a Set iterates in insertion order — which is `entries.keys()` first, then anything new from `hydratedSections`. Outcome: the listener evicts the wrong `(contentKey, sectionId)` pair (and computes a junk `sectionId` like `extended-section-x`).

In practice the tabbed UI can have multiple guides loaded concurrently, so this is reachable.

**Suggested fix** — match exactly. The hydrated set already knows the exact `(contentKey, sectionId)` split. Iterate it and check for an exact match instead of prefix match:

```ts
for (const hydratedKey of hydratedSections) {
  const sep = hydratedKey.indexOf('::');
  if (sep <= 0) continue;
  const contentKey = hydratedKey.slice(0, sep);
  const sectionId = hydratedKey.slice(sep + 2);
  if (contentKey + '-' + sectionId !== stripped) continue;
  evictSectionCacheForKey(contentKey, sectionId);
  interactiveStepStorage.invalidateCountCache(contentKey);
  notify(contentKey);
  return;
}
```

This also collapses the two-phase "collect content keys, then prefix-match" into one pass.

### 2. `hydrationVersion` grows unbounded over a session (`completion-store.ts:92`)

All four eviction paths (`evictSectionCacheForKey`, `evictContentCache`, `evictAllContentCaches`, and the cross-tab handler) deliberately bump but never delete from `hydrationVersion`. Deletion would reset a captured version to 0 and match a fresh cycle's snapshot — exactly the bug the comment warns about. But this means every unique `(contentKey, sectionId)` ever seen leaves a permanent map entry.

Not a leak in practice (each entry is one number, content keys are bounded to 200 chars and you'd need to evict and never re-enter to leave an orphan). Worth a one-line note in the JSDoc at line 80–91 so a future maintainer doesn't "tidy this up" by adding deletes.

### 3. Indirect concern: `hydratedSections` separator collision (low risk)

`hydratedKey.indexOf('::')` assumes `::` never appears inside a content key. `sanitize()` in `content-key.ts` only strips `..`. URLs like `https://...` contain `://` but not `::`; pathnames don't normally contain `::`. Worth flagging only because the disambiguation strategy already has correctness pressure (issue #1) and choosing a separator that can't appear in either side would foreclose both.

### 4. Listener has no removal path

`installCrossTabSync` adds the listener at module init and never removes it. Fine for production. In dev with HMR, repeated module reloads would stack listeners and the older ones would close over stale module state. Doesn't appear to be hit today, but if you ever add a `removeCrossTabSync` for tests, please also call it from a HMR `module.hot.dispose` hook.

### 5. Test gap: "unrelated keys" test verifies the negative weakly (`completion-store.test.tsx:484–502`)

The test confirms `completed` stays `true` after an unrelated event, which is what users care about. But it doesn't directly prove the early-return ran — a future regression that accidentally invalidates the cache *and* then re-hydrates from authoritative storage (which still has the completion) would pass this test. A `jest.spyOn` on `evictSectionCacheForKey` or `notify` (with `toHaveBeenCalledTimes(0)`) would tighten this. Optional.

## Style and conventions

- Comments are dense and load-bearing; that's the house style for this subsystem. No churn requested.
- Sentence case in docs is followed.
- No new tier-crossing imports (`global-state` → `lib` was already established).

## Recommendation

Request changes on issue #1 (prefix-collision in the listener). Issues 2–5 are nits or notes — fine as follow-ups or to defer. The hydration-version invariant and overall design are sound and worth landing as soon as #1 is tightened.

moxious (approved):

PR #913, opened 2026-05-22
fix(storage): surface quota-exceeded toast (N-3 follow-up)

Jayclifford345:
## Summary

Closes the deferred **N-3** finding from PR #909. Previously, all four `QuotaExceededError` catch blocks in `src/lib/user-storage.ts` (`createLocalStorage.set`, `journeyCompletionStorage.set`, `interactiveCompletionStorage.set`, `tabStorage.setTabs`) swallowed the error into a `console.warn`, so users never learned their progress had silently stopped persisting once browser storage filled up.

This PR adds a module-level `hasWarnedAboutQuota` guard plus a `warnQuotaExceededOnce()` helper that publishes a single `AppEvents.alertWarning` toast — _"Browser storage full. Your progress may not be saved. Try resetting old guide progress via My Learning to free up space."_ — the first time any of the four catch blocks fires in a page lifecycle. Subsequent quota errors are still logged but no longer re-toast (avoids spamming the user as repeated writes pile up against the same full quota).

The helper wraps `getAppEvents()` in `try/catch` because the runtime can be uninitialized in some test environments. A test-only `__resetQuotaWarningForTests()` export lets the suite exercise the once-per-lifecycle contract deterministically.

## What changed

- `src/lib/user-storage.ts`
  - New module-scope `hasWarnedAboutQuota` flag.
  - New `warnQuotaExceededOnce()` helper that publishes `AppEvents.alertWarning` once.
  - New `__resetQuotaWarningForTests()` test-only export.
  - Call site added in each of the four `QuotaExceededError` catch blocks.
- `src/lib/user-storage.test.ts`
  - Mock `@grafana/runtime` (`getAppEvents` + `usePluginUserStorage`).
  - Spy on `Storage.prototype.setItem` to throw `QuotaExceededError` deterministically.
  - Cover: one toast across many quota-exceeded writes spanning all four storage paths; reset → re-publishes; `getAppEvents()` throwing doesn't bubble to the storage caller.

## Why

PR #909 explicitly deferred this user-facing affordance to keep that PR scoped to the larger interactive-state collapse. See the deferred N-3 follow-up bullet and the rollout plan at `.cursor/plans/pr909-followups-parallel_6f07373c.plan.md` (PR 1 of 8).

## Test plan

- [x] `npm run check` (typecheck + lint + prettier + go lint/test + jest) — green, 3915 tests pass.
- [ ] Manual: fill `localStorage` to quota in a dev session, perform a tab save or interactive completion write, observe a single _"Browser storage full"_ warning toast; subsequent writes do not re-toast.

## Risk and reversibility

Low. Additive, behind a module flag that idempotently no-ops after the first publish. Reverting this commit restores prior silent behavior with no schema or storage-shape changes.

### DIFF
--- a/src/lib/index.ts
+++ b/src/lib/index.ts
@@ -8,5 +8,7 @@ export * from './analytics';
 // Hash utilities
 export * from './hash.util';
 
-// User storage
-export * from './user-storage';
+// User storage is intentionally NOT re-exported here. Consumers should
+// import from `./user-storage` directly so that test-only helpers
+// (e.g. `__resetQuotaWarningForTests`) do not leak into the lib's
+// public surface.
--- a/src/lib/user-storage.ts
+++ b/src/lib/user-storage.ts
@@ -35,7 +35,8 @@
  * SECURITY NOTE: Data is NOT encrypted. Do not store sensitive information.
  */
 
-import { usePluginUserStorage } from '@grafana/runtime';
+import { AppEvents } from '@grafana/data';
+import { getAppEvents, usePluginUserStorage } from '@grafana/runtime';
 import { useCallback, useRef, useEffect } from 'react';
 import { z } from 'zod';
 
@@ -161,6 +162,52 @@ const LIMITS = {
 } as const;
 
 // ============================================================================
+// QUOTA WARNING (N-3 follow-up from PR #909)
+// ============================================================================
+
+/**
+ * Module-level guard so the user only sees one quota-exceeded toast per
+ * page lifecycle. The previous behavior swallowed every QuotaExceededError
+ * into a `console.warn`, so the user never learned that their progress
+ * had silently stopped persisting. See PR #909.
+ */
+let hasWarnedAboutQuota = false;
+
+/**
+ * Publish a single user-visible warning toast the first time any storage
+ * write hits a `QuotaExceededError`. Subsequent calls are no-ops to avoid
+ * spamming the user as repeated writes pile up against the same full quota.
+ *
+ * Wrapped in try/catch because `getAppEvents()` can throw in environments
+ * where `@grafana/runtime` isn't fully initialized (notably some tests).
+ */
+function warnQuotaExceededOnce(): void {
+  if (hasWarnedAboutQuota) {
+    return;
+  }
+  hasWarnedAboutQuota = true;
+  try {
+    getAppEvents().publish({
+      type: AppEvents.alertWarning.name,
+      payload: [
+        'Browser storage full',
+        'Your progress may not be saved. Try resetting old guide progress via My Learning to free up space.',
+      ],
+    });
+  } catch {
+    // getAppEvents() can throw in test envs without a grafana/runtime mock.
+  }
+}
+
+/**
+ * Test-only reset of the module-level `hasWarnedAboutQuota` flag so
+ * suites can exercise the once-per-lifecycle contract deterministically.
+ */
+export function __resetQuotaWarningForTests(): void {
+  hasWarnedAboutQuota = false;
+}
+
+// ============================================================================
 // TYPE DEFINITIONS
 // ============================================================================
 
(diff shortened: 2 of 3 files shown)


moxious (commented):
## Review summary

- **Closes:** N-3 deferred from #909 — quota-exceeded toast in `lib/user-storage.ts`
- **Classification:** product-runtime (small fix + tests)
- **Reversibility:** two-way (additive module flag; no storage shape / event contract / schema change)
- **Anchor match:** exact

## Findings (ordered by severity)

### M1 — minor / correctness — flag-flip ordering

`warnQuotaExceededOnce` flips `hasWarnedAboutQuota = true` **before** the try block. If `getAppEvents()` throws on the first storage write of a page lifecycle (the uninitialized-runtime path the comment explicitly anticipates), the flag is permanently set even though no toast was shown — the user gets no toast for the remainder of the page.

Suggested fix:

```ts
function warnQuotaExceededOnce() {
  if (hasWarnedAboutQuota) return;
  try {
    getAppEvents().publish({ type: AppEvents.alertWarning.name, payload: [...] });
    hasWarnedAboutQuota = true; // only flip after successful publish
  } catch {
    // swallow — runtime not initialized yet, leave flag false so next write retries
  }
}
```

Existing test `'does not throw when getAppEvents() itself throws'` covers the non-throwing contract but doesn't pin the silent-failure-doesn't-permanently-suppress invariant. Worth a small addition.

### N1 — nit — barrel leak

`__resetQuotaWarningForTests` is re-exported through `src/lib/index.ts:12` (`export * from './user-storage'`). No internal callers consume it via the barrel today, so practical risk is zero, but the underscore-prefixed name is leaking into the library's public surface. Either restructure the barrel to omit `__`-prefixed members or add an explicit narrow re-export list.

### N2 — nit — manual test-plan unchecked

Manual checkbox in the test plan is unchecked. Acceptable given the additive low-risk change, but worth a one-session dev smoke before merge.

## Other checks (clean)

- **Four catch blocks all covered** (lines 240, 718, 866, 965 in the diff). All four invoke `warnQuotaExceededOnce()`.
- **Test fidelity:** mocks `Storage.prototype.setItem` to throw a real `QuotaExceededError`, drives the helpers end-to-end through three of the four code paths with a retry shape (`tabStorage`, `journeyCompletionStorage`, `interactiveCompletionStorage`). The fourth (`createLocalStorage.set`) re-throws after warning, so it's exercised transitively. An explicit one-liner for that path would be tidy but not required.
- **Toast copy is sentence-case compliant** ("Browser storage full. Your progress may not be saved...").
- **Once-per-lifecycle scope is correct.** Module-scope guard is the right granularity for "your storage is full." Confirmed no custom HMR plumbing in the repo; helper is idempotent on dev re-evaluation.

---

src/lib/index.ts
```diff
@@ -8,5 +8,7 @@ export * from './analytics';
 // Hash utilities
 export * from './hash.util';
 
-// User storage
-export * from './user-storage';
+// User storage is intentionally NOT re-exported here. Consumers should
+// import from `./user-storage` directly so that test-only helpers
+// (e.g. `__resetQuotaWarningForTests`) do not leak into the lib's
+// public surface.
```
tomglenn, line 14: This sort of comment is unnecessary. The absence of a thing does not need a justification for why it isn't there. Comments should explain why code exists, not why it doesn't exist.

src/lib/user-storage.ts
```diff
@@ -160,6 +161,55 @@ const LIMITS = {
...
+// QUOTA WARNING (N-3 follow-up from PR #909)
+// ============================================================================
+
+/**
+ * Module-level guard so the user only sees one quota-exceeded toast per
+ * page lifecycle. The previous behavior swallowed every QuotaExceededError
+ * into a `console.warn`, so the user never learned that their progress
+ * had silently stopped persisting. See PR #909.
+ */
+let hasWarnedAboutQuota = false;
```
tomglenn, line 174: Excessive level of commenting, we need to stop the LLMs from doing this.

src/lib/user-storage.ts
```diff
@@ -160,6 +161,55 @@ const LIMITS = {
   MAX_PERSISTED_TABS: 50, // Prevent quota exhaustion
 } as const;
 
+// ============================================================================
+// QUOTA WARNING (N-3 follow-up from PR #909)
+// ============================================================================
```
tomglenn, line 166: Advice stripping these headings, they don't serve a purpose.

tomglenn (approved):
Approving, but would like to see the comments toned down as noted.

PR #916, opened 2026-05-22
refactor(multi-step): tighten data parameter to InteractiveElementData (F-6 follow-up)

Jayclifford345:
## Summary

Replaces the `data: any` parameter on `checkRequirementsFromData` inside `checkActionRequirements` (in `src/components/interactive-tutorial/interactive-multi-step.tsx`) with the typed `InteractiveElementData`. This gives us a compile-time guard against future field-casing drift between the legacy snake_case author payload and the camelCase runtime shape produced by the parser introduced in #909.

## Why

This is the deferred F-6 finding from #909. The broader F-6 sweep tightened most call sites during that PR, but a single `data: any` signature in `interactive-multi-step.tsx` was left behind. The parser already emits camelCase, so no caller-side changes are required — `npx tsc --noEmit` and `npm run check` both pass cleanly.

## Scope

- One-line type tightening + one type import.
- Plan: `pr909-followups-parallel_6f07373c.plan.md` — PR 4.
- Refs: #909.

## Test plan

- [x] `npx tsc --noEmit` passes
- [x] `npm run check` passes (231 suites / 3912 tests)
- [ ] CI green on PR

## Risk and reversibility

Trivial. One-line signature change; revertable with a single commit if it ever surfaces a caller-side regression.

### DIFF
--- a/src/components/interactive-tutorial/interactive-multi-step.tsx
+++ b/src/components/interactive-tutorial/interactive-multi-step.tsx
@@ -11,6 +11,7 @@ import { useStepChecker, validateInteractiveRequirements } from '../../requireme
 import { reportAppInteraction, UserInteraction, buildInteractiveStepProperties } from '../../lib/analytics';
 import { INTERACTIVE_CONFIG } from '../../constants/interactive-config';
 import { InternalAction } from '../../types/interactive-actions.types';
+import type { InteractiveElementData } from '../../types/interactive.types';
 import { testIds } from '../../constants/testIds';
 import { STEP_STATES } from './step-states';
 import { markStepCompleted, resetStep, useStepCompletion } from '../../global-state/completion-store';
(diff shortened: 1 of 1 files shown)


moxious (commented):
## Review summary

- **Closes:** F-6 deferred from #909 — `data: any` → `InteractiveElementData` in `interactive-multi-step.tsx`
- **Classification:** product-runtime (type-only tightening)
- **Reversibility:** two-way
- **Anchor match:** exact

## Findings

**Clean.** Verified:

1. The lone `data: any` survivor at `interactive-multi-step.tsx:66` is the only F-6 site in that file. (One unrelated `(e: any)` at line 98 on `result.error` items is separate — not F-6 scope.)
2. `InteractiveElementData` (`src/types/interactive.types.ts:12-44`) is camelCase-only (`refTarget`, `targetAction`, `targetValue`) — matches the parser-emitted shape from #909.
3. The `actionData` literal built in `checkActionRequirements` (lines 74-81) supplies every required field and uses the same camelCase the hook reads (`interactive.hook.ts:213-221`). No silent widening.
4. Single caller (line 343) passes the hook-returned `checkRequirementsFromData`, whose signature has been typed since #909. The tightening removes the inconsistency rather than introducing one.
5. Snake-case (`reftarget`, `targetaction`, `targetvalue`) survives only as DOM attribute names and requirement-token strings — not as fields on `InteractiveElementData`. Type-level guard holds.

### Coverage note

F-6 closes this one site. The unrelated `(e: any)` at line 98 (typing recovery-result `error[]` entries) is a separate latent surface — out of F-6 scope, worth a follow-up if you want to retire `any` from this file entirely.

Disposition: **approve**.

---

moxious (approved):

PR #917, opened 2026-05-22
fix(progress): treat all-passive guide as 100% when acknowledged (F-1 follow-up)

Jayclifford345:
## Summary

Deferred **F-1** follow-up to #909 (*refactor(interactive): collapse state ownership, restore tier hygiene*). A guide whose sections are entirely passive registers no interactive steps, so the user's "Mark section complete" click never reaches `persistSection` — and the persisted `interactiveCompletionStorage` percentage stayed at 0% even after the user acknowledged every section. The progress chip and the My Learning row silently lied about a real terminal state.

This PR routes the all-passive percentage through `sectionAcknowledgementStorage` — the storage namespace the production ack writer actually writes to — instead of inventing an ack-marker entry in `interactiveStepStorage` (which `countAllCompleted` deliberately filters out).

## What changed

**`src/lib/user-storage.ts`**

- Add `sectionAcknowledgementStorage.countAllAcknowledged(contentKey)` — a synchronous localStorage scan that mirrors `interactiveStepStorage.countAllCompleted` and returns the number of acknowledged sections under `contentKey`. The completion store needs a sync read to compute progress on every change.

**`src/global-state/section-registry.ts`**

- Export `getRegisteredSectionCount()` — the denominator the completion store needs when `getTotalDocumentSteps() === 0`. Every `<InteractiveSection>` registers itself (even passive-only) so this is the canonical "how many sections is this guide?" answer.

**`src/global-state/completion-store.ts`**

- `refreshGuidePercentage` and `getGuideProgress` — when `getTotalDocumentSteps() < 1`, derive the percentage from `countAllAcknowledged / getRegisteredSectionCount` instead of returning 0. Persist via `interactiveCompletionStorage.set` so cross-tab reads + My Learning see the same value. `getGuideProgress` reports `{ completed: ackCount, total: sectionCount }` in this branch so the numbers make sense to UI consumers ("3 / 5 sections acknowledged").
- Export `refreshAndNotifyGuideProgress(contentKey)` — a public entry point that recomputes the percentage, fires `kind: 'guide'`, and notifies subscribers. Used by the section's ack handler (which doesn't go through `persistSection`).
- Extend the `interactive-progress-cleared` guard inside `persistSection` to require **both** `countAllCompleted === 0` **and** `countAllAcknowledged === 0`, so resetting the last interactive step in a mixed guide doesn't falsely tell consumers the guide is back to zero progress while passive acks remain.

**`src/components/interactive-tutorial/interactive-section.tsx`**

- Relax the `isCompleted` effect guard from `stepComponents.length > 0` to `stepComponents.length > 0 || gateAnalysis.isAllPassive`. All-passive sections now also persist `sectionDoneStorage` (so `section-completed:` requirement checks work for them) and dispatch `kind: 'section'`.
- After the section dispatch, all-passive sections call `refreshAndNotifyGuideProgress(getContentKey())` so the guide percentage gets recomputed and persisted.
- Mirror the recompute on the reset path so My Learning + the chip drop back down when the user resets an all-passive section.

**`src/test-utils/interactive-section-harness.tsx`**

- Mock `sectionAcknowledgementStorage.countAllAcknowledged` for the section-state-machine and tripwire test suites.

**`src/lib/user-storage.test.ts`** and **`src/global-state/completion-store.test.tsx`**

- Cover the new ack-count read, the four all-passive branches in `getGuideProgress` (full ack, partial ack, no ack, no sections registered), and `refreshAndNotifyGuideProgress` persistence.

## Why

PR #909 collapsed step state into the canonical completion store and moved passive acks into their own storage namespace. The original F-1 fix proposed in this PR's earlier revision reached for a synthetic ack-marker entry in `interactiveStepStorage` that no post-#909 code path actually writes (and that `countAllCompleted` deliberately filters out). Reviewer flagged it as a passing-test-broken-in-prod regression — this revision wires the fix through the storage the production ack writer uses, so it actually fires in the running app.

## Test plan

- [x] `npx jest src/global-state/completion-store.test.tsx --no-coverage` — 30/30 pass (5 new F-1 cases + 25 existing).
- [x] `npx jest src/lib/user-storage.test.ts --no-coverage` — 40/40 pass (4 new `countAllAcknowledged` cases).
- [x] `npx jest src/components/interactive-tutorial src/global-state --no-coverage` — 390 pass / 12 skipped / 5 todo.
- [x] `npm run check` — clean.

## Risk and reversibility

Low-to-moderate. The change touches three production code paths (`refreshGuidePercentage`, `getGuideProgress`, the cleared-event guard) and adds two effect branches in the section component, but every branch is gated on `getTotalDocumentSteps() < 1` (or `gateAnalysis.isAllPassive`), so guides with any interactive steps are bit-for-bit identical to pre-PR. Reversible by reverting the commit; the new `countAllAcknowledged` reader is additive.

## References

- Parent PR: #909
- PR #909 follow-up batch, PR 2 (F-1).

### DIFF
--- a/src/components/interactive-tutorial/interactive-section.tsx
+++ b/src/components/interactive-tutorial/interactive-section.tsx
@@ -450,11 +450,9 @@ export function InteractiveSection({
   }, [isCompleted, isPreviewMode, sectionId]);
 
   // Trigger reactive checks when section completion status changes.
-  // The `stepComponents.length > 0 || gateAnalysis.isAllPassive` guard
-  // lets all-passive sections (which register zero interactive steps)
-  // still flow through this effect so their "section done" bit is
-  // persisted and the guide-level progress percentage refreshes — see
-  // F-1 follow-up to PR #909.
+  // The `gateAnalysis.isAllPassive` branch lets sections with zero
+  // interactive steps still persist `sectionDoneStorage` + refresh
+  // guide progress (F-1, #909 follow-up).
   useEffect(() => {
     if (isCompleted && (stepComponents.length > 0 || gateAnalysis.isAllPassive)) {
       // Single unified event — replaces the two legacy CustomEvents
@@ -472,10 +470,8 @@ export function InteractiveSection({
         // Preview mode is sandboxed — keep the ephemeral check DOM-only.
         if (!isPreviewMode) {
           sectionDoneStorage.set(getContentKey(), sectionId, true);
-          // All-passive sections never write to `interactiveStepStorage`,
-          // so `persistSection` never runs for them and the guide
-          // percentage would otherwise never update. Refresh here so
-          // My Learning + the progress chip reflect the ack.
+          // All-passive sections bypass `persistSection`, so refresh
+          // the guide percentage explicitly.
           if (gateAnalysis.isAllPassive) {
             refreshAndNotifyGuideProgress(getContentKey());
           }
@@ -1082,11 +1078,8 @@ export function InteractiveSection({
           detail: { contentKey },
         })
       );
-      // For all-passive sections the store-based reset doesn't update
-      // the persisted guide percentage (no step writes → no
-      // `persistSection`). Recompute now so My Learning / the chip
-      // catch the drop. Safe for all sections — it's a no-op when
-      // there's nothing to recalculate.
+      // All-passive sections bypass `persistSection` on reset too;
+      // recompute so the persisted percentage drops in lockstep.
       if (!isPreviewMode) {
         refreshAndNotifyGuideProgress(contentKey);
       }
--- a/src/components/interactive-tutorial/interactive-section.tsx
+++ b/src/components/interactive-tutorial/interactive-section.tsx
@@ -70,6 +70,7 @@ import {
   markStepCompleted,
   markStepsCompleted,
   reconcileSection,
+  refreshAndNotifyGuideProgress,
   resetSection as resetSectionStore,
   resetSteps,
   useSectionCompletion,
@@ -448,9 +449,14 @@ export function InteractiveSection({
     }
   }, [isCompleted, isPreviewMode, sectionId]);
 
-  // Trigger reactive checks when section completion status changes
+  // Trigger reactive checks when section completion status changes.
+  // The `stepComponents.length > 0 || gateAnalysis.isAllPassive` guard
+  // lets all-passive sections (which register zero interactive steps)
+  // still flow through this effect so their "section done" bit is
+  // persisted and the guide-level progress percentage refreshes — see
+  // F-1 follow-up to PR #909.
   useEffect(() => {
-    if (isCompleted && stepComponents.length > 0) {
+    if (isCompleted && (stepComponents.length > 0 || gateAnalysis.isAllPassive)) {
       // Single unified event — replaces the two legacy CustomEvents
       // (`section-completed` on document + `interactive-section-completed`
       // on window). The `!hasEmittedGuideCompletionRef.current` guard
@@ -466,6 +472,13 @@ export function InteractiveSection({
         // Preview mode is sandboxed — keep the ephemeral check DOM-only.
         if (!isPreviewMode) {
           sectionDoneStorage.set(getContentKey(), sectionId, true);
+          // All-passive sections never write to `interactiveStepStorage`,
+          // so `persistSection` never runs for them and the guide
+          // percentage would otherwise never update. Refresh here so
+          // My Learning + the progress chip reflect the ack.
+          if (gateAnalysis.isAllPassive) {
+            refreshAndNotifyGuideProgress(getContentKey());
+          }
         }
       }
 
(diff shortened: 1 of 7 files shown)


moxious (commented):
## Review summary

- **Closes:** F-1 deferred from #909 — all-passive guide should read 100% when acknowledged
- **Classification:** product-runtime (frontend, state/persistence)
- **Reversibility:** two-way for the read path; write path persists `100` to `interactiveCompletionStorage` (clean via existing `useContentReset`)
- **Anchor match:** **divergent** — the read-path fix does not connect to the production ack writer

## Findings (ordered by severity)

### 🛑 H1 — HIGH — fix does not close F-1 against production code

The PR body claims: *"the passive ack still writes an ack-marker entry that `countAllCompleted` correctly sees."* In post-#909 code that statement is false.

Verified by checkout against the PR branch:

1. `grep -rn "ack-marker\|__ack-marker__" src/` returns only three families of hits:
   - The defensive filter at `src/lib/user-storage.ts:1190-1194` that **excludes** markers from the count
   - This PR's own comments (`completion-store.ts:253-257`, `:350-354`) claiming a producer exists
   - This PR's own test (`completion-store.test.tsx:178`) that **manually injects** `Set(['__ack-marker__'])` into the in-memory `storedCompleted` map
2. The actual production ack writers are `handleMarkSectionComplete` (`interactive-section.tsx:1113-1120`) → `setAcknowledgement()` (`use-section-persistence.ts:75-80`) → `sectionAcknowledgementStorage.set(...)`. They write to `SECTION_ACKNOWLEDGED_PREFIX`, a **different storage namespace** that `countAllCompleted` never reads.
3. `countAllCompleted` (`user-storage.ts:1173-1204`) scans only `INTERACTIVE_STEPS_PREFIX` keys and filters out anything ending in `::ack-marker` (line 1194).

**Net effect:** in production, an all-passive-acknowledged guide gives `completed = 0`, so the new `completed > 0` predicate in `getGuideProgress` evaluates `false` and `refreshGuidePercentage` returns `undefined`. The chip still reads 0%. **F-1 reproduces after merge.**

The test passes because it bypasses the public store API and writes directly into the in-memory `storedCompleted` map. The injected value is something no real code path produces.

Note that even the marker convention itself is contradictory: the filter at `user-storage.ts:1194` deliberately *removes* `::ack-marker` from the count, so even adding a writer that emits markers wouldn't increment `completed`. The filter and the new branch are at odds.

**Suggested fix path** (one of):

1. **Drive from `sectionDoneStorage`/`sectionAcknowledgementStorage` instead.** `refreshGuidePercentage` and `getGuideProgress` consult `sectionDoneStorage` (already populated by #909) to detect "every registered section has its done bit set, document total is 0" → return 100. No new marker, no contradiction with the filter. This is the cleanest path.
2. **Retire the marker convention entirely.** The defensive filter at `user-storage.ts:1194` is a fossil with no producer; remove it together with the comments referencing markers, and use option (1).
3. **If a separate PR in this batch adds the marker writer** (and rewires the filter), mark this as `Depends on:` that PR and don't merge alone.

Whichever path is taken, the new unit test must drive the fix **through the public ack writer** (simulate the "Mark section complete" click or call `setAcknowledgement` via a hook test), not by injecting into the storage map directly. The injection test was always going to pass — it just doesn't bind to anything real.

Additionally, `mockTotalDocumentSteps = 0` makes the `total < 1` predicate true for any input, so the renamed test also wouldn't catch a regression in the `completed > 0` arm.

### M1 — MEDIUM — `persistSection` may dispatch `interactive-progress-cleared` spuriously after the fix

Once any fix actually wires a writer for the all-passive path, walk through `persistSection` (lines 238-243): it computes `guideTotal = interactiveStepStorage.countAllCompleted(contentKey)`. The filter strips `::ack-marker` entries, so `guideTotal === 0`, so a `CustomEvent('interactive-progress-cleared')` is dispatched immediately after the user clicks "Mark section complete." Consumed by the recommendations panel and `useAlignmentReevaluation`, this turns "user completed an all-passive guide" into the same event as a reset.

Mitigation depends on the fix path chosen — option (1) above sidesteps this entirely by not adding a writer. If a marker-based path is chosen, guard the dispatch on `interactiveCompletionStorage.get(contentKey) !== 100`.

### L1 — LOW — renamed 0-case test no longer covers the interesting branch

The rename of `getGuideProgress returns 0% when total is unknown and nothing has been completed` dropped the `storedCompleted.set(...)` seed; the renamed test now asserts the easy case (both zero). The near-miss — `completed > 0` but the branch chooses `0` because the marker convention is wrong, or `completed > 0` but the percentage shouldn't be 100 — has no coverage. Keep both tests, or add a third case that seeds a non-marker step ID with `mockTotalDocumentSteps = 0` and asserts `percentage: 100` for the current branch behavior.

### Informational

- **Reversibility check:** `useContentReset.ts:57` calls `interactiveCompletionStorage.clear(progressKey)` and `evictContentCache(progressKey)` — a user who resets after the fix lands does get the persisted `100` cleared. No migration needed.
- **Cross-PR with #919:** #919 adds `persistSection(contentKey, sectionId)` inside `ensureHydrated`'s additive-merge branch. Once F-1 is properly fixed, that path will exercise the fix transitively — good coverage, no conflict.
- **Cross-PR with #914:** docs-only, no semantic interaction.

## Disposition

**Request changes.** The diff and the test are internally consistent, but they describe a system that does not exist post-#909. F-1 reproduces after merge as currently written. Recommend re-routing the fix through `sectionDoneStorage` / `sectionAcknowledgementStorage` and driving the test through the public ack writer.

---

moxious (changes requested):

tomglenn (dismissed):
Latest commit addresses concerns in @moxious review and code LGTM

moxious (approved):

PR #918, opened 2026-05-22
refactor(youtube-video): replace window globals with typed content-key accessors (F-3 follow-up)

Jayclifford345:
## Summary

Resolves the deferred **F-3** finding from PR #909
(*refactor(interactive): collapse state ownership, restore tier hygiene*),
which flagged `src/docs-retrieval/components/docs/youtube-video-renderer.tsx`
for reaching into `(window as any).__DocsPluginActiveTabUrl` and
`(window as any).__DocsPluginContentKey` instead of going through the
typed content-key module that already owns those globals.

This is the **PR 3** slice of the PR #909 follow-up batch (F-3).

## What changed

- **`src/global-state/content-key.ts`** — export typed reader functions
  `getActiveTabUrl()` and `getContentKeyOverride()` alongside the existing
  `getContentKey()`. Both prefer the in-module typed value and fall back
  to the legacy `window.__DocsPlugin*` globals so older callers keep
  working during migration. They return the raw (unsanitized) value;
  callers that need a storage-safe identifier are pointed at
  `getContentKey()`. JSDoc now explicitly names `analytics.ts` alongside
  `useGlobalActiveTabExposure` and `content-renderer.tsx` so a future
  agent does not remove the fallback prematurely.
- **`src/docs-retrieval/components/docs/youtube-video-renderer.tsx`** —
  replace the two `(window as any).__DocsPlugin*` casts inside
  `getDocumentInfo()` with calls to the new typed accessors. No
  behavioural change; the analytics event still emits the same
  `source_document`, `video_url`, and `video_title` fields (the
  `contentKey` intermediate is only used as a fallback for
  `sourceDocument`, not a payload field).
- **`src/global-state/content-key.test.ts`** — unit coverage for the new
  readers: typed setter wins, legacy window-global fallback works,
  `undefined` when neither is set, and raw (unsanitized) passthrough
  for both `getActiveTabUrl` and `getContentKeyOverride`.

## Why

`(window as any).__DocsPlugin*` casts in renderer code are exactly the
tier-hygiene smell PR #909 set out to eliminate. The writers
(`setActiveTabUrl`, `setContentKeyOverride`) already live in
`global-state/content-key.ts`; this PR adds the matching typed readers
and migrates the last `docs-retrieval` consumer over so we can remove
the `any` casts.

## Test plan

- [x] `npx jest src/global-state/content-key.test.ts --no-coverage` — 17/17 pass (10 existing + 7 new for the two new readers)
- [x] `npm run check` — clean (typecheck + lint + prettier + lint:go + test:go + test:ci all green; 3919 frontend tests pass)
- [ ] Spot-check the renderer in dev (`npm run dev` + `npm run server`) — analytics event for a YouTube video still includes `source_document`

## Risk and reversibility

Low risk. Readers fall back to the legacy globals so any caller still
writing through `window.__DocsPlugin*` keeps working. Easily reverted —
this is a 3-file, additive-then-swap change.

### DIFF
--- a/src/docs-retrieval/components/docs/youtube-video-renderer.tsx
+++ b/src/docs-retrieval/components/docs/youtube-video-renderer.tsx
@@ -1,5 +1,6 @@
 import React, { useEffect, useRef, useCallback } from 'react';
 import { reportAppInteraction, UserInteraction } from '../../../lib/analytics';
+import { getActiveTabUrl, getContentKeyOverride } from '../../../global-state/content-key';
 import { parseUrlSafely } from '../../../security/url-validator';
 
 export interface YouTubeVideoRendererProps {
@@ -59,8 +60,8 @@ export function YouTubeVideoRenderer({
   // Get document info for analytics
   const getDocumentInfo = useCallback(() => {
     try {
-      const tabUrl = (window as any).__DocsPluginActiveTabUrl as string | undefined;
-      const contentKey = (window as any).__DocsPluginContentKey as string | undefined;
+      const tabUrl = getActiveTabUrl();
+      const contentKey = getContentKeyOverride();
 
       const sourceDocument = tabUrl || contentKey || window.location.pathname || 'unknown';
 
--- a/src/global-state/content-key.test.ts
+++ b/src/global-state/content-key.test.ts
@@ -98,5 +98,11 @@ describe('content-key', () => {
     it('returns undefined when neither the typed value nor the global is set', () => {
       expect(getContentKeyOverride()).toBeUndefined();
     });
+
+    it('returns the raw value without sanitization', () => {
+      const longKey = 'bundled:' + 'a'.repeat(500);
+      setContentKeyOverride(longKey);
+      expect(getContentKeyOverride()).toBe(longKey);
+    });
   });
 });
--- a/src/global-state/content-key.test.ts
+++ b/src/global-state/content-key.test.ts
@@ -1,4 +1,11 @@
-import { getContentKey, resetContentKeyForTests, setActiveTabUrl, setContentKeyOverride } from './content-key';
+import {
+  getActiveTabUrl,
+  getContentKey,
+  getContentKeyOverride,
+  resetContentKeyForTests,
+  setActiveTabUrl,
+  setContentKeyOverride,
+} from './content-key';
 
 describe('content-key', () => {
   beforeEach(() => {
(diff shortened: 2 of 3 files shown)


moxious (commented):
## Review summary

- **Closes:** F-3 deferred from #909 — `(window as any).__DocsPlugin*` reaches in `youtube-video-renderer.tsx` replaced with typed accessors
- **Classification:** product-runtime + analytics-and-telemetry (additive typed wrappers; no payload-shape change)
- **Reversibility:** two-way (additive readers + one call-site swap; legacy window-globals are still written by their existing owners)
- **Anchor match:** exact within scope (see migration-completeness note below)

## Findings

### Migration scope narrower than "last `docs-retrieval` consumer" implies

Grep confirms the YouTube renderer was the only `(window as any).__DocsPlugin*` site **inside `src/docs-retrieval/`**, so the description is true for that subtree. However, two other consumers still reach via `(window as any)`:

- `src/lib/analytics.ts:522-523` (`__DocsPluginActiveTabUrl`, `__DocsPluginContentKey`)
- `src/components/content-renderer/content-renderer.tsx:285` (read) and `:335` (write of `__DocsPluginContentKey`)

The fallback path inside the new readers is therefore **not** dead weight — those callers can still be the source of truth. A one-line addition to the JSDoc noting "fallback exists because `analytics.ts` and `content-renderer.tsx` still write the legacy globals" would prevent a future agent from removing the fallback prematurely. Non-blocking.

A natural F-3.1 would migrate those two remaining sites; once both are typed, the `readGlobal` fallback inside `getActiveTabUrl` / `getContentKeyOverride` becomes truly dead weight and the `(window as any).__DocsPlugin*` pattern can be entirely retired.

### PR description over-claims the analytics payload shape

The body says the event "still receives the same `sourceDocument` / `contentKey` payload." Reading `getDocumentInfo()` shows the returned object has `source_document`, `video_url`, `video_title` — `contentKey` is an intermediate variable used as a fallback for `sourceDocument`, not a field in the payload. Behavior is preserved (the `tabUrl || contentKey || pathname` chain is identical pre/post); the description just over-promises. Tighten before merge.

### Test asymmetry — `getContentKeyOverride` missing raw-passthrough assertion

`getActiveTabUrl` has the "returns the raw value without sanitization" case with a 500-char string; `getContentKeyOverride` does not. Both readers share the same `??` + `readGlobal` shape, so coverage is logically equivalent — but the symmetry helps future readers and is cheap. Add a one-line mirror.

## Other checks (clean)

- **Tier hygiene:** `docs-retrieval` (Tier 2) → `global-state` (Tier 1) is a downward import, legal per the architecture test. No allowlist change needed; `architecture.test.ts` untouched.
- **Security/sanitization:** raw return is the right call. JSDoc steers callers needing a storage-safe identifier to `getContentKey()` (which applies `sanitize()` + 200-char cap). The renderer's `sourceDocument` flows into `reportAppInteraction(...)` only — same sink, same value as before. No new XSS/DOM-injection surface.
- **Behavior preservation:** identical. Pre-PR `(window as any).__DocsPluginActiveTabUrl ?? (window as any).__DocsPluginContentKey ?? window.location.pathname ?? 'unknown'`; post-PR `getActiveTabUrl() ?? getContentKeyOverride() ?? window.location.pathname ?? 'unknown'`. Each typed accessor falls back to the matching global, so when only the legacy globals are set (the current production reality — neither `analytics.ts` nor `content-renderer.tsx` call the typed setters yet) the resolution is bit-for-bit identical. Pinned by the new fallback tests.
- **Reset/reversibility:** `resetContentKeyForTests` only clears the typed module state — intentionally does not touch window globals. Existing tests that manage their own window-global teardown are unaffected.

Disposition: **approve** with PR-description tightening and the small symmetric test addition.

---

moxious (changes requested):
minor changes only, but worth making prior to merge.

tomglenn (approved):
Latest commit addresses review comments. LGTM

moxious (approved):

PR #919, opened 2026-05-22
fix(completion-store): persist additive merge from hydration resolve (F-5 follow-up)

Jayclifford345:
## Summary

PR #909 deferred F-5 as a "test-only" tripwire — the assumption being that `markStepCompleted`'s additive-merge invariant across the hydration race already held. Writing the tripwire revealed it doesn't.

## Root cause

When `markStepCompleted` fires while a section's hydration read is still in flight:

1. `ensureHydrated` early-returns (the section is already marked hydrated).
2. The synchronous write persists the new in-memory set (e.g. `Set(['step-b'])`) and **overwrites** the on-disk snapshot (`Set(['step-a'])`).
3. The hydration `.then` resolver correctly re-adds the missing IDs to the in-memory cache (skipping anything in the `cleared` set).
4. But the resolver only calls `persistSection` when `hadClears` is true — the additive path silently leaves storage out of sync with memory.
5. On next reload, the previously-stored `step-a` is gone.

## Fix

One-line change in `ensureHydrated`'s `.then` handler: when the resolve adds new IDs (the `changed` branch), call `persistSection` so storage matches the merged in-memory state. The existing `hadClears` reconcile path is preserved; in the rare case both branches fire we eat one redundant write — harmless.

```ts
if (changed) {
  bumpSectionVersion(contentKey, sectionId);
  notify(contentKey);
  persistSection(contentKey, sectionId); // <-- added
}
```

## Test

The F-5 tripwire that surfaced this bug — `'mark issued during pending hydration survives the resolve'` in the existing `hydration race` describe block of `src/global-state/completion-store.test.tsx`:

- Seeds `step-a` in storage.
- Mounts a probe for `step-b` (kicks off hydration).
- Calls `markStepCompleted('step-b', ...)` **before** flushing microtasks.
- Asserts final storage contains BOTH `step-a` AND `step-b`.

Pre-fix: fails (`Set { "step-b" }`, `step-a` dropped).
Post-fix: passes alongside the existing `resetStep` / `resetSection` / `resetSteps` race guards.

## Risk and reversibility

- Single localized change inside the hydration resolver, behind the same `changed` predicate that already exists.
- Existing race-guard tests for reset paths (3 in the same describe block) still pass; the bulk-event, eviction, and reconciliation suites all still pass — `npm run check` clean (231 suites, 3913 tests).
- Reversible by reverting this commit.

Refs PR #909 (F-5 deferred item).
Plan: `pr909-followups-parallel_6f07373c.plan.md`.

## Test plan

- [x] `npx jest src/global-state/completion-store.test.tsx --no-coverage` — 26/26 pass, including the new F-5 tripwire.
- [x] `npm run check` clean (typecheck + lint + prettier + lint:go + test:go + test:ci).

### DIFF
--- a/src/global-state/completion-store.test.tsx
+++ b/src/global-state/completion-store.test.tsx
@@ -221,6 +221,15 @@ describe('completion-store', () => {
       expect(screen.getByTestId('completed').textContent).toBe('false');
       expect(storedCompleted.get(`${CONTENT_KEY}-section-x`)).toEqual(new Set(['step-1']));
     });
+
+    it('mark issued during pending hydration survives the resolve', async () => {
+      storedCompleted.set(`${CONTENT_KEY}-section-x`, new Set(['step-a']));
+      render(<StepProbe stepId="step-b" sectionId="section-x" />);
+      // Hydration pending.
+      act(() => markStepCompleted('step-b', 'section-x', 'manual'));
+      await flushMicrotasks();
+      expect(storedCompleted.get(`${CONTENT_KEY}-section-x`)).toEqual(new Set(['step-a', 'step-b']));
+    });
   });
 
   describe('bulk progress events', () => {
--- a/src/global-state/completion-store.ts
+++ b/src/global-state/completion-store.ts
@@ -181,6 +181,7 @@ function ensureHydrated(contentKey: string, sectionId: string): void {
       if (changed) {
         bumpSectionVersion(contentKey, sectionId);
         notify(contentKey);
+        persistSection(contentKey, sectionId);
       }
       // Reconcile storage with the filtered cache. Without this, the
       // cleared IDs would persist in storage and reappear on next reload.


moxious (commented):
## Review summary

- **Closes:** F-5 deferred from #909 — additive-merge persist fix (scope expanded from test-only because the tripwire surfaced a real divergence)
- **Classification:** product-runtime (state-management / storage-coherence)
- **Reversibility:** two-way (one-line add inside an existing predicated branch)
- **Anchor match:** exact

## Findings

**Approve.** Walked all four race scenarios and the fix holds:

1. **Mark during pending hydration** (the new test case): synchronous `markStepCompleted` ran `persistSection` writing `{step-b}` over `{step-a}`; resolver's additive merge adds `step-a` to memory and now calls `persistSection`, restoring `{step-a, step-b}`. Fixed.
2. **Reset during pending hydration:** `hadClears` path unchanged.
3. **Mark + reset interleaved:** cleared IDs go into `hydrationClears`; resolver skips them, then both `changed` (additive) and `hadClears` are true — two `persistSection` calls, idempotent (same `bySteps` snapshot). `bumpSectionVersion` fires once (inside the `if (changed)` block, correctly), so subscribers re-render once.
4. **Empty resolve** (stored empty, no clears): `changed=false`, `hadClears=false`, no spurious `persistSection` call.

**`changed` predicate semantics:** `!bySteps.has(stepId)` — strictly "the resolver added an ID to the cache." Will not fire when memory already contains everything stored. Correct minimum-surface change.

**Test quality:** The tripwire asserts **storage** state (`storedCompleted.get(...)`), not just in-memory cache — that's exactly the bug. Pre-fix the synchronous persist overwrites storage with `{step-b}` and the resolver's memory-only merge doesn't reconcile → final storage `{step-b}`. Test fails loudly. Post-fix passes.

### Informational (non-blocking)

- **Double-write side-effect fan-out:** when `changed && hadClears` both fire, `persistSection` runs twice and so does `refreshGuidePercentage` → `dispatchProgress({kind:'guide', ...})`. Payload-identical re-dispatches; downstream `kind:'guide'` listeners are subscribe-and-re-read so it's a benign re-render trigger. Worth flagging if a future subscriber becomes count-sensitive. Possible follow-up: hoist into a single guarded `persistSection` call when both predicates fire.
- **Cross-PR with #917:** #917 edits `getGuideProgress` / `refreshGuidePercentage` ~25 lines higher. Textually orthogonal. Semantically: the new `persistSection` call here transitively invokes `refreshGuidePercentage`, so the F-1 ack-marker accounting (if/when #917 is correctly wired — see review on that PR) will be exercised on the resolver path too. No interaction risk.

The fact that writing the tripwire surfaced a real bug retroactively validates F-5's catalog placement.

---

moxious (approved):

PR #920, opened 2026-05-22
test(interactive-section): pin sibling re-render on completion-store flip (F-2 follow-up)

Jayclifford345:
## Summary

PR #909 collapsed `InteractiveSection`'s completion state into the global `completion-store` and re-wired the section to subscribe via `useSyncExternalStore(useSectionCompletion)`. Review surfaced **F-2**: there was no test pinning the cross-step re-render contract. If the store wiring ever regresses (missing listener, stale snapshot, or a dep-array gap on the version bump), an external `markStepCompleted` call would silently fail to refresh sibling steps, and section-level derivations (`stepsCompleted`, `isCompleted`, the Reset button gate) would never reflect the new state. That probe was deferred from #909 as follow-up **F-2**; this PR lands it.

## What changed

Appends one tripwire test to `src/components/interactive-tutorial/interactive-section.contracts.tripwire.test.tsx`:

- Mounts an `InteractiveSection` with two `InteractiveStep` children (explicit `stepId`s `step-1`, `step-2`).
- Drains the initial mount-time `pathfinder-step-progress` so the post-flip event is unambiguously store-driven.
- Calls `markStepCompleted('step-1', SECTION_ID, 'manual')` from outside React.
- Asserts the sibling harness stays addressable (no unmount / re-key).
- Asserts the section's `useDocumentStepProgress` effect re-fires `pathfinder-step-progress` for this `sectionId` — proving the section re-committed with a new `completedSteps` reference and the effect's dep array saw the change.
- Clicks `step-2`'s harness to confirm `handleStepComplete` captured the refreshed `completedSteps` snapshot (the Reset button appears, not Do Section).

Test-only — no production code touched.

## Why

Pins the `useSyncExternalStore(useSectionCompletion)` wiring introduced by #909 against silent regressions. Without this, a dep-array gap or missed subscription would compile, type-check, lint, and pass every existing test — but break the Reset / progress UX in production. The tripwire fails loudly the moment the wiring breaks.

## Test plan

- [x] `npx jest src/components/interactive-tutorial/interactive-section.contracts.tripwire.test.tsx --no-coverage` — 11/11 pass (including the new F-2 test, 8 ms).
- [x] `npm run check` — clean (231 suites / 3913 tests pass; typecheck + lint + prettier + Go lint + Go tests + frontend CI all green).

## Risk and reversibility

Test-only file; no runtime, no public API, no migration. Two-way door — revert by deleting the appended `describe` block.

## Refs

- Deferred follow-up **F-2** from #909.
- PR #909 follow-up batch, PR 6 (F-2).

### DIFF
--- a/src/components/interactive-tutorial/interactive-section.contracts.tripwire.test.tsx
+++ b/src/components/interactive-tutorial/interactive-section.contracts.tripwire.test.tsx
@@ -392,4 +392,96 @@ describe('InteractiveSection contracts — Phase 0 tripwire', () => {
       }
     });
   });
+
+  // F-2 tripwire — pin the cross-step re-render contract introduced by
+  // PR #909. The store relocation moved completion state out of the
+  // section's reducer and into the global `completion-store`; the section
+  // now subscribes via `useSyncExternalStore(useSectionCompletion)`. If
+  // that wiring regresses (missing listener, stale snapshot, dep-array
+  // gap on the version bump), an external `markStepCompleted` call will
+  // silently fail to trigger a re-render — sibling steps keep rendering
+  // against the pre-flip completion set and section-level derivations
+  // (`stepsCompleted`, `isCompleted`, the Reset button gate) never reflect
+  // the new state. PR #909 deferred this probe as follow-up F-2; this
+  // test pins the contract so the wiring can't quietly regress.
+  describe('cross-step re-render on store flips (F-2)', () => {
+    it('re-renders sibling step + section when markStepCompleted flips from outside React', async () => {
+      const { markStepCompleted } = require('../../global-state/completion-store');
+
+      const flushMicrotasks = async () => {
+        await act(async () => {
+          await Promise.resolve();
+        });
+      };
+
+      const { events, unsubscribe } = recordSectionEvents();
+      try {
+        render(
+          <InteractiveSection id="contracts" title="Contracts section" autoCollapse={false}>
+            <InteractiveStep stepId="step-1" targetAction="highlight" refTarget=".a">
+              Step one
+            </InteractiveStep>
+            <InteractiveStep stepId="step-2" targetAction="highlight" refTarget=".b">
+              Step two
+            </InteractiveStep>
+          </InteractiveSection>
+        );
+
+        // Both children mount with their author-supplied stepIds.
+        await waitFor(() => {
+          expect(screen.getByTestId(`harness-complete-step-1`)).toBeInTheDocument();
+          expect(screen.getByTestId(`harness-complete-step-2`)).toBeInTheDocument();
+        });
+
+        // Drain the initial mount's `useDocumentStepProgress` dispatch so
+        // the post-flip event is unambiguously attributable to a
+        // store-driven re-render rather than the mount-time effect.
+        await waitFor(() => {
+          expect(events.find((e) => e.name === 'pathfinder-step-progress')).toBeDefined();
+        });
+        events.length = 0;
+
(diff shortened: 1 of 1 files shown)


moxious (commented):
## Review summary

- **Closes:** F-2 deferred from #909 — pins sibling re-render on completion-store flip
- **Classification:** tests-only
- **Reversibility:** two-way
- **Anchor match:** exact

## Findings

**Approve with two minor cleanups.**

### N1 — comment/PR-description overstates assertion #2

The PR summary and the in-test comment claim the test asserts `pathfinder-step-progress` re-fires "with the new `completedSteps` set". The actual payload is `{ sectionId, totalSteps, documentStepIndex, completedCount }` — no `completedSteps` field exists (see `hooks/use-document-step-progress.ts:73-81`). The assertion at lines 463-466 only verifies the event was *re-fired* for the right `sectionId`; it does not inspect a set.

This is still a valid dep-array probe — re-firing proves the effect re-ran, which requires the `completedSteps` reference to have changed in the section's commit — but the wording invites future maintainers to add a payload check that can't be written. Tighten to something like: "a re-fire of `pathfinder-step-progress` for this `sectionId` proves the section re-committed with a new `completedSteps` reference (the only dep in the effect that could have changed)" and drop "with the new `completedSteps` set" from the PR body.

### N2 — author-machine path leak in PR body

Refs section cites `/Users/jayclifford/.cursor/plans/pr909-followups-parallel_6f07373c.plan.md`. Replace with a relative reference (e.g. "PR #909 follow-up batch, PR 6") before merge.

## Tripwire-quality checks (clean)

- **Anchor:** `id="contracts"` resolves to `sectionId = "section-contracts" = SECTION_ID`, so the external `markStepCompleted('step-1', SECTION_ID, 'manual')` flip hits the same subscriber the section reads via `useSectionCompletion`. The Reset gate proof is load-bearing: `handleStepComplete` builds `postWriteCompleted` from the closed-over `completedSteps`, so if the prior store flip hadn't refreshed the section's commit, `allStepsCompleted` → false → `isCompleted` → false → no Reset button. The test fails loudly on the regression it claims to pin.
- **Subscription proof:** the test uses a non-React caller (`require(...).markStepCompleted(...)` outside `render`/`act`-wrapped events), so the only path that lets the section see the flip is the store's `subscribe` notifier driving `useSyncExternalStore`. A top-down `parentCompleted` regression would not pick up an external mutation and would fail this test.
- **Dep-array gap detection:** `useDocumentStepProgress` deps are `[currentlyExecutingStep, stepComponents, sectionId, completedSteps]`. If `completedSteps` were dropped from deps (or stale-closured), the post-flip dispatch would never fire and the `waitFor` at 463 would time out.
- **Drain discipline:** `events.length = 0` at line 442 follows a `waitFor` proving the mount-time event already landed; the subsequent flip is wrapped in `await act(async () => { ...; await Promise.resolve(); })`. Not racy.
- **Test isolation:** `beforeEach` calls `resetSectionHarness()` + `resetInteractiveCounters()` and clears window globals; `afterEach` runs `cleanup()`. No module-scope mutation introduced. The completion-store is the real implementation (no jest.mock at the top), which is appropriate for a tripwire — it's the wiring being pinned.
- **Style:** matches the existing `recordSectionEvents` / `try/finally unsubscribe` / `waitFor` shape used by the other 10 cases in this file.

Local run: 11/11 pass (8 ms for the new case).

---

moxious (changes requested):
Minor changes only, but worth making prior to merge.

moxious (approved):

PR #921, opened 2026-05-22
fix(interactive): "Fix this" docks the nav from overlay state

### LINKED ISSUE #910: Selecting "Fix this" does not dock the left menu panel
## Summary

Clicking "Fix this" on the `navmenu-open` requirement when the left menu panel is undocked opens the panel in overlay mode, but doesn't dock it. The requirement-failed card disappears (signaling success to the user), but the step's action then fails with "Element not found".

## Steps to replicate

Image

1. `npm run server` (Grafana 13.x in Docker)
2. Sign in
3. Toggle open the tutorial by selecting the top right `?` icon
4. Select `Start` on the Welcome to Grafana card
5. Select `Show me` to showcase expected behavior
6. On the left menu panel select `Undock menu`
7. On the right-hand tutorial panel, select `Fix this` to showcase that the left-hand panel remains undocked

## Root cause (traced via breakpoint)

Image

`openAndDockNavigation()` in `src/interactive-engine/navigation-manager.ts:1507` has two early returns that bail whenever nav items are visible in the DOM. But nav items are visible in **both** docked and overlay states, so the check doesn't actually distinguish them.

  - **Line 1529** (first early return): fires when "Fix this" is invoked with the nav **already in overlay mode**. The function returns immediately. **This is the path the repro above exercises** — the "Undock menu" step leaves the nav in overlay.
  - **Line 1539** (second early return): fires when the toggle click (line 1533) opens the nav as overlay rather than docked.
  - **Line 1543**: the dock step. Never reached in either case.

## Environment
- Plugin: grafana-pathfinder-app main (reproduced at commit baf54ef)
- Grafana: 13.x via npm run server
- Browser: Firefox on Linux

Happy to open a PR if useful — wanted to surface as issue first.

EvanHStanton:
## Summary

Fixes #910. `openAndDockNavigation()` had two early returns that bailed without docking whenever nav items were visible in the DOM, which conflated the "already docked" and "in overlay mode" states. Adds a small private helper that discriminates them by the `#dock-menu-button` aria-label (`"Dock menu"` in overlay, `"Undock menu"` when docked) and clicks the button only in the overlay case.

## Demo

grafana_pathfinder_pr

## What changed

- `src/interactive-engine/navigation-manager.ts` — added `dockIfInOverlay()` private method; called at both early-return paths in `openAndDockNavigation()`. The closed-nav → toggle → poll path (line 1543) is untouched.
- `src/interactive-engine/navigation-manager.test.ts` — added a `describe('openAndDockNavigation')` block with two tests: clicks-in-overlay (the bug, now passing) and no-op-when-docked (already-correct behavior, locked in).

## Why this shape

PR #709 added the `navItemsVisible` early returns specifically to prevent `openAndDockNavigation` from undocking an already-docked sidebar (because `#dock-menu-button` exists in both states with opposite semantics — clicking it docks in overlay, undocks when docked). That fix correctly avoided the undock-regression but was too aggressive — it also skipped the dock step for the overlay case. The aria-label is the discriminator Grafana already exposes: this PR uses it.

**One caveat for i18n:** the aria-label literal (`"Dock menu"`) is en-US. Non-English Grafana instances may localize it.

## Test plan

- `npm run check` (typecheck + lint + prettier + Go gates + frontend tests)
- Manual repro from #910: nav undocked → click Fix this → confirm nav now docks (was: stayed in overlay)
- Manual no-op confirmation: nav already docked → click Fix this → confirm nav stays docked (no undock regression — locks in PR #709's intent)


moxious (approved):
thanks for the contribution!  The code looks straightforward; correct, and the issue is well specified.  Appreciate the effort!

PR #922, opened 2026-05-22
chore: remove unused custom event listeners ("grafana:location-changed")

leslie-heinzen:
Follow-up to https://github.com/grafana/grafana-pathfinder-app/pull/904

## Summary

Removes all `grafana:location-changed` custom event listeners from the codebase. I can't find evidence that this event is dispatched anywhere in Grafana. I verified by grepping the full grafana/grafana repo and all installed @grafana/* packages. There's a _super old_ `location-change` (no "d") event, but that predates React, and was removed a long time ago. Every URL change flows through locationService's shared HistoryWrapper history instance, which our existing `history.listen()` handlers use.

### DIFF
--- a/docs/developer/engines/requirements-manager.md
+++ b/docs/developer/engines/requirements-manager.md
@@ -224,7 +224,7 @@ When a requirement check fails, the system may identify an automatic fix. The `f
 
 - **Context monitoring** - Listens to ContextService (EchoSrv) for Grafana state changes
 - **DOM monitoring** - Observes specific DOM changes (plugins, datasources) via MutationObserver
-- **Navigation monitoring** - Tracks URL changes and navigation events (popstate, hashchange, grafana:location-changed)
+- **Navigation monitoring** - Tracks URL changes and navigation events (popstate, hashchange, focus)
 - **Guide response monitoring** - Reacts to guide variable updates (input blocks, user responses)
 - **Heartbeat checking** - Optional periodic rechecking for fragile requirements (configurable)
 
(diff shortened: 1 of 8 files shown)


docs/developer/engines/requirements-manager.md
```diff
@@ -224,7 +224,7 @@ When a requirement check fails, the system may identify an automatic fix. The `f
 
 - **Context monitoring** - Listens to ContextService (EchoSrv) for Grafana state changes
 - **DOM monitoring** - Observes specific DOM changes (plugins, datasources) via MutationObserver
-- **Navigation monitoring** - Tracks URL changes and navigation events (popstate, hashchange, grafana:location-changed)
+- **Navigation monitoring** - Tracks URL changes and navigation events (popstate, hashchange, focus)
```
leslie-heinzen, line 227: I let claude do this update and was skeptical about `focus`, but it turns out that is accurate: https://github.com/grafana/grafana-pathfinder-app/blob/9ba6e5d32f1cc57cddee322c332d6b533040f841/src/requirements-manager/requirements-checker.hook.ts#L491-L492

moxious (approved):
LGTM.  I verified that the event truly is dead, so this is a good "remove needless tech debt" PR.

One follow-up worth filing separately: this PR makes a pre-existing inconsistency more visible. Four of the six call sites use the `try { getHistory().listen() } catch { popstate }` pattern, but requirements-checker.hook and step-checker.hook never wired history.listen(). They rely on popstate + hashchange + (focus or 2s polling). 

Popstate doesn't fire on history.pushState().  Grafana's locationService.push() is actually pushState under the hood. So those two hooks miss most SPA navigations through their event channel.  Not in scope of this PR though.

EchoSrv + MutationObserver cover the gap in practice, so it's not a regression but the comment at requirements-checker.hook ("event-driven, no polling") is arguably misleading.

Suggested follow-up: wire the same `try { getHistory().listen() } catch { popstate }` shape into both hooks and update that comment.

PR #925, opened 2026-05-26
fix(popout): milestone navigation + theme conformance

Jayclifford345:
## Summary

https://github.com/user-attachments/assets/08a02fb5-26ab-4001-807d-376340522795

- Floating ("popout") panel now mounts the shared `LearningJourneyMilestoneToolbar` (prev/next arrows, "Milestone X of Y" label, progress bar, Open, Reset) and wires `useKeyboardShortcuts`, so users can complete an entire learning path without docking back to the sidebar.
- Wraps the standalone floating React root in `ThemeContext.Provider` driven by `config.theme2` plus a body-class `MutationObserver`, so the popout matches Grafana's light/dark theme and live-updates when the user toggles theme.

## Context

The floating panel was unusable for full learning paths: there was no UI or keyboard handler to advance milestones, and the panel always rendered in the default theme because it mounts via `createCompatRoot` outside Grafana's provider tree.

Implementation notes:
- `LearningJourneyMilestoneToolbar` gains a third `surface` variant `'floating'` (existing `'sidebar'` + `'fullscreen'` were already designed for reuse). Adds `floating_panel_milestone_progress_bar` as the analytics `interaction_location` for the Open button on this surface.
- `FloatingPanelContent` renders the toolbar gated on the toolbar inputs being present — the full-screen surface omits them and renders its own toolbar in the layout's sub-header, so there's no double-toolbar regression.
- New `useGrafanaTheme()` reads `config.theme2` and observes `<body>` class mutations (Grafana toggles `theme-dark` / `theme-light` on body during theme switches; no public theme-change event is exposed for plugins).

## Test plan

- [x] `npm run server` + `npm run dev`, open Pathfinder, start a multi-milestone learning path.
- [x] Pop out the panel: confirm the milestone toolbar (prev arrow, "Milestone X of Y", next arrow, progress bar, Open, Reset) renders above the guide content.
- [x] Click the next/prev arrows — milestone advances/retreats without docking. Disabled states match the sidebar at the start/end milestones.

### DIFF
--- a/src/components/docs-panel/components/LearningJourneyMilestoneToolbar.tsx
+++ b/src/components/docs-panel/components/LearningJourneyMilestoneToolbar.tsx
@@ -75,6 +75,14 @@ export interface LearningJourneyMilestoneToolbarProps {
    * `<Dropdown>`; the fullscreen surface omits it.
    */
   trailingActions?: React.ReactNode;
+  /**
+   * When true, suppresses the entire `milestoneActions` row (Open + Reset +
+   * `trailingActions`). The arrow nav and progress bar still render. Used
+   * by the floating panel to reclaim vertical space — the popout has a much
+   * smaller content area than the sidebar / fullscreen surfaces and the
+   * secondary actions are still reachable from the docked sidebar.
+   */
+  compact?: boolean;
 }
 
 /**
@@ -92,6 +100,7 @@ export function LearningJourneyMilestoneToolbar({
   progressKey,
   onResetGuide,
   trailingActions,
+  compact = false,
 }: LearningJourneyMilestoneToolbarProps) {
   const styles = useStyles2(getMilestoneStyles);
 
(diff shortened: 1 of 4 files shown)


tomglenn (changes requested):
Can confirm milestone navigation now works when undocked, and the floating panel respects current Grafana theme.

One fix needed; when navigating between milestones in undocked mode the state is not persisted when you re-dock. For example, if you are on Milestone 2 when docked, then you undock, navigate to milestone 4 and re-dock, you will be back at Milestone 2. The state should persist between docked/un-docked mode.

Jayclifford345:
Hey @tomglenn fully agree with your find, but I will follow up with a separate PR for this since it's out of scope of the fix and would balloon this pr.

tomglenn (approved):
Sounds good, if the bug is pre-existing then agree we should follow up with a separate PR to fix. In which case LGTM 👍

PR #926, opened 2026-05-26
fix(alignment): make starting-location prompt one-shot per guide launch

Jayclifford345:
## Summary

https://github.com/user-attachments/assets/b93ba9d2-9314-4006-93ec-97ad0ec82348

- The implied-0th-step alignment prompt grew a reactive re-evaluation loop that re-checked `startingLocation` on every Grafana `history` change while the user had no progress. Guides that step the user across pages (e.g. `/a/grafana-collector-app/fleet-management` → `/a/grafana-collector-app/alloy`) hit the loop on every step transition and got paused mid-flow.
- This PR removes the reactive loop and its supporting scaffold (begin/end interactive-navigation counter, `location_change` launch source, stale `currentPath` refresh). The prompt is now evaluated **once** at guide launch in `loadDocsTabContent`, resolved by `confirmAlignment` / `dismissAlignment`, and never re-surfaces.
- Net **-643** lines. The `useAlignmentReevaluation` hook is renamed to `useGuideProgressState` to reflect its sole remaining responsibility (tracking `hasInteractiveProgress` for the reset-guide button).

## What stays

- `recovery/{starting-location,alignment-evaluator,launch-sources}.ts` — manifest/bundled resolver + path-prefix matcher.
- `AlignmentPrompt.tsx`, `AlignmentPendingContext`, `useIsAlignmentPaused` — UI + step-checker gate so step 1 doesn't auto-execute behind the launch prompt.
- `confirmAlignment` / `dismissAlignment` on `docs-panel.tsx` (slimmed: try/finally without the counter wrapper).
- Telemetry: `AlignmentPromptShown` / `Confirmed` / `Dismissed` events fire at launch only.

## What goes

- `src/hooks/useAlignmentReevaluation.ts` — the `history.listen` reactive loop.
- `src/global-state/interactive-navigation.ts` — counter that only existed to suppress the loop.
- `reevaluateAlignment()` method on `CombinedLearningJourneyPanel` — dead with the listener gone.
- `'location_change'` launch source — only the deleted method emitted it.
- All `beginInteractiveNavigation` / `endInteractiveNavigation` call sites in `navigate-handler.ts`, `navigation-manager.ts`, `docs-panel.tsx`.
- Three test suites in `docs-panel.alignment.test.ts` (`reevaluateAlignment`, `interactive-navigation guard`, `stale currentPath refresh`).

## Regression test

[...]

PR #927, opened 2026-05-26
refactor(docs-panel): fold loadTabContent + loadDocsTabContent into loadTab

Jayclifford345:
## Summary

- Make `loadTab` the single canonical content-loader on `CombinedLearningJourneyPanel`; remove the deprecated `loadTabContent` and `loadDocsTabContent` methods from the class and the `DocsPanelModelOperations` interface.
- Extract the shared lifecycle (set-loading → success-patch → failure-patch → persist) into three private helpers (`setTabLoading`, `applyTabLoadSuccess`, `applyTabLoadFailure`), with the docs-only and journey-only divergences living in two strategy methods (`applyDocsTabSuccess`, `applyJourneyTabSuccess`).
- Rename `loadDocsTabContentResult` → `loadTabContentResult(url, { mode: 'docs' | 'journey', ... })` so both fetch pipelines share one signature; extend the unit tests with journey-mode coverage.

## What changed

- `src/components/docs-panel/docs-panel.tsx` — `loadTab` is now the body, not just a dispatcher. `openDocsPage` calls `loadTab` directly (matching `useContentReset`, `link-handler.hook`, `useAutoOpenListener`, and the tab-restoration path that already used it). Deleted `loadTabContent` and `loadDocsTabContent`. Updated the `_pendingLaunchSource` JSDoc to call out that only the docs branch consumes the carrier.
- `src/components/docs-panel/utils/docs-tab-loader.ts` — `loadTabContentResult(url, { mode, ... })` covers both branches. Journey mode surfaces a controlled error for empty URLs (the panel still guards the empty case upstream). `loadDocsTabContentResult` alias removed.
- `src/components/docs-panel/types.ts` — `DocsPanelModelOperations` now exposes `loadTab` only; the two deprecated entries are gone. `_recordAutoLaunchSource` is retained (still used by `useContentReset` and `ContextPanel` callbacks) with refreshed JSDoc.
- `src/components/docs-panel/link-handler.hook.ts` and `src/components/floating-panel/FloatingPanelContent.tsx` — model shapes drop their `loadTabContent` member.
- Tests — `docs-panel.alignment.test.ts`, `docs-panel.tab-restore-guard.test.ts`, `useContentReset.test.ts`, and `link-handler.hook.test.ts` updated to use the new function/method names. `docs-tab-loader.test.ts` rewritten to cover both `mode: 'docs'` and `mode: 'journey'` (7 cases total).
- JSDoc updates in `context-panel.tsx`, `utils/tab-validation.ts`, and `hooks/useContentReset.ts` to reference the unified `loadTab`.

## Why

`loadTab` was already the unified entry point on the read side but `openDocsPage` was the last call site bypassing it, and the two bodies (`loadTabContent`, `loadDocsTabContent`) shared ~50–60% of their lifecycle code while diverging only on alignment, launch-source consume, manifest auto-derive, empty-URL handling, and `setJourneyCompletionPercentage`. Folding them removes the divergence-by-duplication, gives one place to reason about lifecycle (loading-state map + patch + persist), and closes a small leak where the launch-source carrier could survive a journey-path reload and contaminate the next docs open.

## Test plan

- [x] `npm run test:ci` — 3924 passed, 230 suites
- [x] `npm run check` — typecheck + lint + prettier + Go all clean
- [x] Alignment regression coverage (15+ cases in `docs-panel.alignment.test.ts` for `pendingAlignment` decisions, `internal_reload` / `browser_restore` / `mcp_launch` sources, full-screen suppression, exact + prefix path matches, bundled-index fallback, etc.) still passes against the unified body
- [x] Tab-restore-guard regression coverage still passes
- [x] Package-manifest coverage extended in `docs-tab-loader.test.ts` (docs/package + docs/empty + docs/not-found + docs/non-package + journey/url + journey/empty + journey/ignores-packageInfo)

## Risk and reversibility

- Behavior: no change to the alignment evaluator, package-manifest auto-derive path, telemetry payloads, or `setJourneyCompletionPercentage` semantics. The only intentional behavior change is that the launch-source carrier is no longer leaked across a journey-path reload (previously a known footgun that `openLearningJourney` defensively drained).
- Reversibility: pure refactor; no schema, storage, or wire-format changes. Reverting the merge restores both old methods unchanged.
- Stack: depends on #926 (`fix(alignment): make starting-location prompt one-shot per guide launch`). Targeting that branch as the base so the diff stays focused on the refactor.


Jayclifford345:
on hold till larger design doc

moxious (changes requested):
Needed changes:
  - QC8 cluster in docs-panel.tsx — multiple rewritten comments are decorative (QC8.2/8.3/8.4):
    - :118-132 (_pendingLaunchSource JSDoc) — enumerates two callers that explain themselves at the call site; trim to
   the consumption invariant.
    - :222-226 and :294-298 — defend-a-non-action shape ("without this the source would leak / would prompt
  mid-tutorial"). Pick one canonical home (the LaunchSource enum / ALIGNED_BY_CONSTRUCTION_SOURCES definition) and
  shrink the call-site comments to "Tag as aligned-by-construction."
    - utils/tab-validation.ts:13-21 and utils/docs-tab-loader.ts:7-13 — JSDoc restating what the function body already
   shows.
  - One-way-door invariant uncovered by tests — the carrier-drain in openLearningJourney is a sequence-order invariant
   (drain BEFORE this.loadTab(tabId, url)) that is neither type-enforced nor tested. A future refactor moving the
  drain after loadTab would silently reintroduce the leak. Add one assertion in docs-panel.alignment.test.ts: after
  openLearningJourney resolves, _pendingLaunchSource is null even if the caller pre-stashed a non-aligned source.
  - Question — :343-345 silent no-op for empty url in journey reloads. Docs-branch + loadTabContentResult journey-mode
   both return a controlled 'Invalid URL provided' error elsewhere; this branch asymmetry isn't tested and the comment
   doesn't say why journey needs the silent path while docs doesn't. Either tighten the comment or add a test.

  Tech debt found in branch changes:
  - A1 (Cargo-cult commentary) — docs-panel.tsx:222-226, :294-298: rewritten comments paraphrase the single-line call
  below them. This file is a hotspot (#903, #924, #928 all touched it); pruning here compounds.
  - A3 (Over-specified defense, suggestive) — :343-348 silent-no-op early return: stated justification ("downstream
  returns an error rather than no-op") is the opposite of what the panel does. Either delete the early return
  (consistent behavior across branches) or extract shouldSilentlySkipEmptyJourneyReload(...) so the asymmetry is
  explicit.
  - E5 micro (suggestive) — 'Failed to load documentation' and 'Failed to load content' literals each appear twice
  inside loadTab (success-path vs catch-path fallbacks). Hoist to module-level const so the pair can't drift.

Jayclifford345:
Closing out this PR based on RFC: https://github.com/grafana/pathfinder-rfcs/pull/2#discussion_r3311419628

PR #929, opened 2026-05-26
refactor(progress-events): migrate `pathfinder-step-progress` onto unified channel (A2)

Jayclifford345:
## Summary

Migration **A2** of the legacy-event consolidation: replace the standalone
`pathfinder-step-progress` window event with a new `kind: 'document'`
variant on the unified `pathfinder:progress` channel. The panel header
chip ("X of Y steps") now reads from the unified channel like every
other progress consumer.

This is one of two siblings; **A3** (`interactive-progress-cleared`) will
follow as a separate PR.

## What changed

- `ProgressEventDetail` (`src/global-state/progress-events.ts`) gains a
  fourth variant:
  ```ts
  | { kind: 'document'; contentKey: string; sectionId: string;
      totalSteps: number; completedCount: number;
      documentStepIndex?: number }
  ```
  Field set matches the legacy payload verbatim. `contentKey` is **new**
  to the envelope — it isn't read by the existing listener, but it
  closes a latent cross-guide-contamination gap (cheap to include,
  harmless if unused).
- `useDocumentStepProgress` (the single producer) now calls
  `dispatchProgress({ kind: 'document', ... })`.
- `useStepProgressFromEvents` (the single listener) now subscribes via
  `subscribeProgressEvent` and filters on `kind === 'document'`.
- Tests:
  - `useStepProgressFromEvents.test.ts` — rewrote dispatch helper to
    use the unified envelope; added an explicit probe that the listener
    ignores `step` / `section` / `guide` variants; unsubscribe assertion
    now expects `'pathfinder:progress'`.
  - `interactive-section.contracts.tripwire.test.tsx` — dropped the
    legacy listener, replaced the `pathfinder-step-progress` payload
    assertion with a `kind: 'document'` shape assertion (key set + types).

## Why

The unified `pathfinder:progress` channel was introduced to consolidate
five legacy custom events behind a single discriminated dispatch /
subscribe pair (`dispatchProgress` / `subscribeProgressEvent`). A2
removes one of the two remaining stragglers, leaving the codebase with
one canonical event channel for progress + a single tier-safe entry
point for it.

## Out of scope

- Window globals `__DocsPluginTotalSteps` and
  `__DocsPluginCurrentStepIndex` are intentionally untouched (deferred
  to A4). They are read by analytics and are a separate contract surface.
- `interactive-progress-cleared` migration is **A3**, opening as a
  sibling PR.

## Risk and reversibility

Low. Single producer + single listener seam, both in the same conceptual
component pair (`InteractiveSection` ↔ panel chip). The contracts
tripwire pins the payload shape across the seam. Fully reversible by
reverting this commit.

## Test plan

- [x] `npm run check` (typecheck + lint + prettier + lint:go + test:go +
      test:ci — 3922 tests pass)
- [x] grep `pathfinder-step-progress` returns only the documentation
      reference in `progress-events.ts` header

### DIFF
--- a/src/components/floating-panel/FloatingPanelManager.tsx
+++ b/src/components/floating-panel/FloatingPanelManager.tsx
@@ -178,9 +178,7 @@ function FloatingPanelInner() {
   // tab is its own kind of "active content" but isn't a guide, so leave it false.
   const hasActiveGuide = activeTab != null && activeTab.id !== 'recommendations' && !isEditorTab;
 
-  // Track interactive step progress via the unified `pathfinder:progress`
-  // event (variant `kind: 'document'`) — shared subscription with
-  // FullScreenPanel via the hook.
+  // Shared with FullScreenPanel via the hook.
   const stepProgress = useStepProgressFromEvents(hasActiveGuide);
 
   // After restoration completes, if there's no guide to show and none
--- a/src/components/floating-panel/FloatingPanelManager.tsx
+++ b/src/components/floating-panel/FloatingPanelManager.tsx
@@ -178,8 +178,9 @@ function FloatingPanelInner() {
   // tab is its own kind of "active content" but isn't a guide, so leave it false.
   const hasActiveGuide = activeTab != null && activeTab.id !== 'recommendations' && !isEditorTab;
 
-  // Track interactive step progress via the `pathfinder-step-progress`
-  // event — shared subscription with FullScreenPanel via the hook.
+  // Track interactive step progress via the unified `pathfinder:progress`
+  // event (variant `kind: 'document'`) — shared subscription with
+  // FullScreenPanel via the hook.
   const stepProgress = useStepProgressFromEvents(hasActiveGuide);
 
   // After restoration completes, if there's no guide to show and none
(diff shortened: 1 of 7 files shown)


Jayclifford345:
on hold till larger design doc

moxious (changes requested):
Needed changes:
  - Tripwire weakened (QC6) — interactive-section.contracts.tripwire.test.tsx:236-238 changed from exact
  Object.keys().sort() to expect.arrayContaining([...]). That accepts stray extra keys silently, defeating the
  wire-shape pin. Restore exact key-set equality (or split into "required keys exact" + "optional documentStepIndex").
  - QC8.4 nit — FloatingPanelManager.tsx:181 comment narrates the call below; consider deleting.
  - Question (non-blocking) — confirm no external consumers (CDN guides, MCP runtime, e2e, analytics dashboards) still
   listen for the old pathfinder-step-progress event name — repo grep is clean but external surfaces aren't searchable
   here.
   
  Tech debt found in branch changes:
  - A3 (low) — use-document-step-progress.ts:71-73: empty try { ... } catch {} swallowing storage I/O errors.
  Pre-existing, but in the touched hunk.

PR #932, opened 2026-05-26
chore(agents): default-to-no-comments rule + QC8 review check

tomglenn:
## Summary

- Adds a **Comments** subsection to `AGENTS.md` with two rules: default to no comments (narrow band of exceptions for counterintuitive code, hidden invariants, and external-bug workarounds), and trim bad-shape comments inside functions you're already editing — never as a standalone sweep PR.
- Adds **QC8** to the `PR_REVIEW.md` unified detection table with a full eight-shape catalog of comment patterns to delete, plus a keep-list so reviewers and agents don't over-correct and strip useful context.
- Extends the **`correctness-and-reliability`** concern in `CONCERNS.md` with two review questions that route through QC8, so `/review` flags violations as findings rather than relying on agent self-policing alone.

## Why

The global agent system prompt already says "default to no comments," but in practice recent code has been running roughly 3 lines of comments per 1 line of actual code, with much of that explanatory prose defending decisions later reversed by refactors. Comments at that density actively make AI agents worse — they read stale comments as authoritative context and either code around them or preserve obsolete decisions, turning the comments themselves into tech debt.

This PR puts the rule, a concrete catalog of named bad shapes, and the `/review` enforcement loop in the three files agents and reviewers actually consult.

## What's deliberately not in scope

- No retroactive sweep PR that deletes comments across the repo. Cleanup rides along on code changes.
- No pre-commit hook or linter rule. Comment quality can't be reliably regex'd; enforcement lives in `/review`.
- No new `.cursor/rules/*.mdc` file. Rule is general enough to live in `AGENTS.md`, which is always loaded.

## Test plan

- [x] `npx prettier --check` passes on the three edited files
- [ ] Manually load `AGENTS.md` in a fresh agent session and confirm the new Comments rule + canonical examples are surfaced
- [ ] Open a small test PR that introduces a deliberate `// Intentionally NOT X here because Y` block and run `/review` — expect a `correctness-and-reliability` finding citing QC8
- [ ] Revisit after 3-4 PRs on this policy: does the agent (a) avoid writing new bad-shape comments and (b) trim adjacent bad-shape comments inside functions it modifies? If it over-corrects or under-corrects, adjust the trim-scope rule or example set


leslie-heinzen (dismissed):
Love the thought behind this! Agree with the dos/don'ts you've drafted.

Something this change has me thinking about is this blog post I read a couple of months ago: https://developer.upsun.com/posts/ai/agents-md-less-is-more

I am not _totally_ convinced about their premise in general, but this change does add a lot of context that is specific to the treatment of in-code comments.

I am happy with the change. IMO, we _need_ the extra context because we are trying to correct the agents from authoring the types of comments we want to avoid that already exist in the code base. But it might be worth thinking about what the detailed rules you have drafted here would look like in half as many words, or even a few sentences, which we could maybe move toward as the comment quality increases in the source code in the future.

tomglenn:
> Love the thought behind this! Agree with the dos/don'ts you've drafted.
> 
> Something this change has me thinking about is this blog post I read a couple of months ago: https://developer.upsun.com/posts/ai/agents-md-less-is-more
> 
> I am not _totally_ convinced about their premise in general, but this change does add a lot of context that is specific to the treatment of in-code comments.
> 
> I am happy with the change. IMO, we _need_ the extra context because we are trying to correct the agents from authoring the types of comments we want to avoid that already exist in the code base. But it might be worth thinking about what the detailed rules you have drafted here would look like in half as many words, or even a few sentences, which we could maybe move toward as the comment quality increases in the source code in the future.

Absolutely agree, I went back and forth on this a few times. Originally I kept the rules just in `PR_REVIEW.md` but this would mean they were only present during the `/review` skill invocation. Then I split it out so I gave a shorter enforcement text in `AGENTs.md` and the full examples in `PR_REVIEW.md` but worried we'd have an asymmetrical problem there (agent wouldn't know about some of the comment rules but `/review` would flag additional ones unknown to `AGENTS.md` as Medium severity).

Eventually landed on just putting it all in `AGENTS.md`. I'm not particularly happy about how much extra context it adds to that file because I agree it should be slimmer (i actually think we have TOO MUCH context stuffed in markdown files in the repo already, another one to tackle at a later date). 

I'm in agreement, I think we should land this change and re-evaluate where we're at in a couple weeks time. Let's see if it stops the agents producing overly verbose comments and if it starts tidying as it goes like I've asked it to. If it does, and the code base starts looking cleaner w/r/t comments, then I'm agreed that we should look to trim these instructions back.

Jayclifford345 (approved):

PR #933, opened 2026-05-26
fix(deps): drop protobufjs from tree to resolve CVE-2026-41242

moxious:
## Summary

Removes `protobufjs` (and its vulnerable `@protobufjs/utf8` sub-dep) entirely from the dependency tree via a surgical faro upgrade. After this PR, `npm ls protobufjs` and `npm ls @protobufjs/utf8` both return empty, and `npm audit` no longer reports either advisory.

## What changed

Three edits to `package.json`:

| Before | After | Reason |
|---|---|---|
| `"@grafana/faro-react": "^2.0.2"` | `"@grafana/faro-react": "2.0.2"` | Pin exact. `^2.0.2` resolves to `2.7.0`, which adds a `react-router@^7.12.0` optional peer that conflicts with our `react-router@6` and breaks `npm ci`. |
| `"@grafana/faro-web-tracing": "^2.0.2"` | `"@grafana/faro-web-tracing": "^2.7.0"` | The actual fix — `faro-web-tracing@2.7.0` pulls `@opentelemetry/otlp-transformer@^0.218.0`, which dropped its `protobufjs` dependency. |
| _(not declared)_ | `"@testing-library/dom": "10.4.1"` (devDep) | Make the previously-implicit peer of `@testing-library/react@16.3.2` explicit. Same version as what was being auto-hoisted. |

## Why this works without bumping `faro-react`

`@grafana/faro-web-tracing@2.7.0` transitively requires `@grafana/faro-web-sdk@^2.7.0`, which npm hoists to the project root. `faro-react@2.0.2` then dedups to that same hoisted `faro-web-sdk@2.7.0`, so the whole tree gets the modern faro-core / otlp-transformer chain without needing to bump `faro-react` itself:

```
faro-react@2.0.2          ┐
faro-web-tracing@2.7.0    ├─→ faro-web-sdk@2.7.0  → faro-core@2.7.0
@grafana/runtime@13.0.1   │       (hoisted)            ↓
@grafana/ui@13.0.1        ┘                    otlp-transformer@0.218.0
                                                        ↓
                                                  (no protobufjs)
```

This avoids `faro-react@2.7.0`'s `react-router@^7.12.0` optional peer (which conflicts with the `react-router@6` we get via `@grafana/ui`) while still eliminating all four `protobufjs` copies.

## What we were vulnerable to

| Resolved version | Path | Status re: CVE-2026-41242 |
|---|---|---|
| `protobufjs@8.0.1` (x1) | `faro-react → faro-web-sdk@2.6.3 → faro-core@2.6.3 → @opentelemetry/otlp-transformer@0.217.0` | Patched, but still pulled vulnerable `@protobufjs/utf8@1.1.0` (CVE-2026-44288) |
| `protobufjs@8.0.0` (x3) | `faro-web-tracing@2.2.3 → @opentelemetry/exporter-trace-otlp-http@0.211.0 → @opentelemetry/otlp-transformer@0.211.0` | **Vulnerable to CVE-2026-41242** |

- **[CVE-2026-41242](https://security.snyk.io/vuln/SNYK-JS-PROTOBUFJS-16094665)** ([GHSA-xq3m-2v4x-88gg](https://github.com/advisories/GHSA-xq3m-2v4x-88gg)) — arbitrary code execution in `codegen` when an attacker-controlled type name reaches `Function()` during reflection-based descriptor loading. Affects `>=8.0.0, <8.0.1` and `<7.5.5`. Fixed in 8.0.1 / 7.5.5.
- **[CVE-2026-44288](https://github.com/advisories/GHSA-q6x5-8v7m-xcrf)** — overlong UTF-8 decoding in `@protobufjs/utf8 <=1.1.0`. Affects all four of our copies.

Exploitability for Pathfinder is low — the `codegen` path requires loading attacker-controlled `.proto`/JSON descriptors, and faro/otel only feed protobufjs bundled OTLP descriptors. But the vulnerable copies sit in the bundle and trip security scanners.

## Considered alternatives

- **Bump `faro-react` to `^2.7.0` as well.** First attempt. Fails `npm ci` due to the react-router peer conflict described above. Would require `--legacy-peer-deps` permanently, which is a larger behavioral change than this PR needs to make.
- **`npm overrides` to pin `protobufjs@^8.0.2`.** Works, but forces a transitive resolution that faro/otel didn't ship-test. Upgrading the parents and letting them remove the dep cleanly is lower risk.

## Verification

- [x] `npm ci --dry-run` — clean, no ERESOLVE
- [x] `npm ls protobufjs @protobufjs/utf8` — both empty
- [x] `npm audit` — no protobufjs / @protobufjs/utf8 entries
- [x] `npm run typecheck` — passing
- [x] `npm run lint` — 0 errors (14 pre-existing warnings)
- [x] `npm run prettier-test` — clean
- [x] `npm run test:go` — passing
- [x] `npm run test:ci` — 231 suites, 3926 tests passing
- [ ] `npm run lint:go` — fails on 2 pre-existing `staticcheck SA5011` warnings in `pkg/plugin/package_recommendations_test.go:386,389`. Verified the same errors exist on `main` — **not introduced by this PR**. Worth filing separately.

## Test plan

- [ ] Confirm faro RUM still initializes (visit `/a/grafana-pathfinder-app/...` and check the browser console for `faro` init logs)
- [ ] Confirm faro tracing instrumentation still attaches (network panel should show `traceparent` headers on fetch calls)
- [ ] No regressions in the right-hand docs sidebar or any docs-panel interactions


moxious:
Worth noting because it's relevant to this PR that we're not currently even using the faro dependency due to a bug in faro we're still working with that team to resolve.

Jayclifford345:
Happy to see the build, okay was worried about the differnce in faro packages

Jayclifford345 (approved):

PR #935, opened 2026-05-27
refactor(block-editor): extract step field mapping utility and cover merge methods with tests

moxious:
## Summary

After main's #1031 landed an equivalent (and more thorough) extraction of `mergeBlocks` from `useBlockEditor`, this branch scopes down to two remaining contributions: a `stepFieldMapping` utility that centralises the tooltip↔description swap rule used when converting between multistep and guided blocks, and public-API tests for `mergeBlocksToMultistep` / `mergeBlocksToGuided` that complement the internal-helper tests added in #1031.

## What to look at

- `src/components/block-editor/utils/stepFieldMapping.ts` — new module with two pure helpers (`stepMultistepToGuided`, `stepGuidedToMultistep`). Each encodes the same field-swap rule that previously lived inline in two separate `.map()` callbacks inside `handleConvertType`. Trace: when converting multistep→guided, the per-step `tooltip` becomes the new `description`; when converting guided→multistep, `description` becomes `tooltip`. The two helpers express that rule symmetrically and are now the single source of truth.
- `src/components/block-editor/hooks/useBlockConversionHandlers.ts` — `handleConvertType` now delegates step transformation to the helpers. Net deletion; behavior unchanged.
- `src/components/block-editor/hooks/useBlockEditor.test.ts` — four new tests exercising the public `mergeBlocksToMultistep` / `mergeBlocksToGuided` API through the hook. These complement (do not duplicate) the helper-level tests in `useBlockEditor.merge.test.ts` that landed via #1031.

## Why

Two copies of the same field-mapping rule sitting a few dozen lines apart in `handleConvertType` were a small but real correctness hazard — any future change to the conversion rule (e.g., carrying a new field across, or making the fallback non-trivial) had to be applied in both places or the two directions would drift out of sync. Extraction puts the rule in one place. The merge tests fill a coverage gap: #1031's tests exercise the internal `mergeBlocks` helper, but the hook callbacks `mergeBlocksToMultistep` / `mergeBlocksToGuided` had no direct integration coverage.

## How to verify

1. Open the block editor, create a `multistep` block with two steps that have non-empty `tooltip` values, convert the block to `guided`, and confirm each step's tooltip text now renders in the description field.
2. Reverse: take a `guided` block with `description` values on each step, convert to `multistep`, and confirm those values now appear as tooltips.
3. Select two interactive blocks in the block editor and use "Merge to multistep"; confirm the resulting multistep block contains both as steps with their content mapped to `tooltip`. Repeat with "Merge to guided" and confirm content maps to `description`.

## Breaking changes

None. Public API and on-disk JSON shape are unchanged.


moxious:
Reviewer note:  multistep and guided are two of the most complex interactive blocks because they work differently, they're similar, and they both contain multiple interactive actions underneath of them. In the context of the block editor, we support editing a block to "switch types" and switches between these types have been problematic in the past.  A core bit of this transform is the `step.description || step.tooltip` schema difference between the two types.  Agents had been writing duplicative code to handle this difference, which I'm trying to centralize, adding tests to make sure that the lessened code is reliable.

Jayclifford345 (dismissed):

Jayclifford345 (approved):

PR #956, opened 2026-05-29
feat(block-editor): strict mode for fragile selector detection in recorder

### LINKED ISSUE #856: Block editor: add 'strict mode' to recorder that refuses low-quality selectors
## Feature request

Add a **Strict mode** toggle to the block editor's record mode that **refuses to record a step** when the recorder cannot find a high-quality selector for the clicked/hovered element.

The intent (per @pepe-grafana) is to deliberately block authors when the underlying product UI lacks stable hooks, so that the friction surfaces as upstream pressure on product teams to add `data-testid` / `aria-label` / stable roles to their components. Today the recorder will happily emit a fragile `:nth-match(...)` or deep compound CSS selector and the author has no idea they've just authored a guide step that will silently rot.

## Why this is feasible today

The selector quality signal already flows through the recording pipeline — we just don't gate on it.

**Selector picking algorithm:** `src/lib/dom/selector-generator.ts` generates candidates and ranks them via `CANDIDATE_SCORES` (lower = better):

| Method | Score |
|---|---|
| `data-testid` | 10 |
| scoped testid | 15 |
| `#id` | 50 |
| role+text / role-contains | 100–110 |
| `aria-label` | 120 |
| button text | 130 |
| placeholder/title/label/name/href | 140–180 |
| compound CSS | 500 |
| `:nth-of-type` | 10,000 |
| `:nth-match` | 50,000 |

**Recorder pipeline:** `src/utils/devtools/selector-generator.util.ts::generateSelectorFromEvent()` already returns a `selectorInfo` containing `method`, `isUnique`, `matchCount`, plus validation `warnings`. `src/utils/devtools/action-recorder.hook.ts` (around line 585) consumes this and builds the `RecordedStep` — this is the natural chokepoint to gate.

**Form-side badge:** `src/components/block-editor/SelectorHealthBadge.tsx` already encodes a quality threshold scheme (`>=80` good / `>=40` medium / `<40` poor) we could mirror or share.

## Proposed behavior

- Add a **Strict mode** toggle to `RecordModeOverlay.tsx`, alongside the existing `Auto-group` toggle.
- While strict mode is on:
  - Hover highlight turns red over any element that would be rejected, so the author sees the problem *before* clicking.
  - On click, if the picked selector's `method` is in a deny-list (e.g. `compound`, `nth-of-type`, `nth-match`, possibly `button-text` / positional methods), the step is **not recorded** and a toast explains why and suggests filing an upstream issue against the offending Grafana component (e.g. *"No stable selector — this element needs a `data-testid`. Skipped."*).
- Optionally surface a counter in the recorder banner: *"3 elements rejected — these need stable selectors upstream"*, so the friction is visible and quantifiable.

## Where the changes would land

- `src/components/block-editor/RecordModeOverlay.tsx` — toggle UI, red-highlight state for rejectable elements
- `src/components/block-editor/hooks/useRecordingActions.ts` and/or `src/utils/devtools/action-recorder.hook.ts` (~line 585) — gate `RecordedStep` creation on `selectorInfo.method` / score
- `src/components/block-editor/hooks/useRecordingState.ts` — persist the strict-mode flag alongside other recording state
- Possibly factor a shared `isHighQualitySelector(selectorInfo)` predicate so `SelectorHealthBadge` and the recorder agree on the bar.

## Open questions

1. Threshold model: deny-list of methods, or a numeric score cutoff? (Methods are likely easier to explain to authors.)
2. Hard block vs soft warning-with-override? Original request says hard block — confirming that's what we want.
3. Should rejected attempts be logged/counted (e.g. for telemetry on which Grafana surfaces have the worst selector hygiene)?
4. Default on or off? A toggle off by default seems safest, with strict mode opt-in for authors who want the discipline.

## Note

Logging only — not scheduled for implementation.

leslie-heinzen:
## Summary

Adds an opt-in **strict mode** to the recorder that warns the author about fragile selectors before they're saved into a guide. When strict mode is enabled, the overlay highlights ineligible hover targets in red as a preview, and on **Stop** the recording pauses and surfaces a modal listing every captured step whose `selectorMethod` falls outside the high-quality allow-list — the author chooses **Save** to keep the recording or **Discard** to throw it away.

Closes #856.

## What to look at

- [`src/lib/dom/selector-generator.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/855cdd908955b8490dac7f994a32f50b6462b3a1/src/lib/dom/selector-generator.ts) — the new `HIGH_QUALITY_SELECTOR_METHODS` allow-list and `isHighQualitySelectorMethod` predicate. This is the single source of truth for which methods are considered fragile. Anything outside this list (`compound`, positional fallbacks, plain button text, `nth-of-type`) is treated as fragile by both the overlay preview and the on-stop check.
- [`src/utils/devtools/action-recorder.hook.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/855cdd908955b8490dac7f994a32f50b6462b3a1/src/utils/devtools/action-recorder.hook.ts) — all four step-creation sites (Shift+click hover, Alt+click form capture, default click, formfill on change) now stamp `selectorMethod` onto the step. The hook records unconditionally; classification is deferred to Stop. A new [`action-recorder.method-stamp.test.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/855cdd908955b8490dac7f994a32f50b6462b3a1/src/utils/devtools/action-recorder.method-stamp.test.ts) covers the four paths.
- [`src/components/block-editor/BlockEditor.tsx`](https://github.com/grafana/grafana-pathfinder-app/blob/855cdd908955b8490dac7f994a32f50b6462b3a1/src/components/block-editor/BlockEditor.tsx) — the on-stop modal flow. Note the invariant: while the modal is open the overlay must be hidden and event capture halted so the hover highlight and stray clicks don't bleed through to the page underneath. Save calls `stopRecording`; Discard calls the new `discardRecording`.
- [`src/components/block-editor/RecordModeOverlay.tsx`](https://github.com/grafana/grafana-pathfinder-app/blob/855cdd908955b8490dac7f994a32f50b6462b3a1/src/components/block-editor/RecordModeOverlay.tsx) — strict-mode toggle (sibling of the existing Auto-group toggle, off by default, session-scoped, not persisted) and the red hover-highlight preview when the hovered element would be rejected.
- [`src/components/block-editor/hooks/useRecordingActions.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/855cdd908955b8490dac7f994a32f50b6462b3a1/src/components/block-editor/hooks/useRecordingActions.ts) — new `discardRecording` action that stops capture and clears state without serializing anything into the guide.

## Why

The recorder happily emitted steps backed by positional CSS selectors and other fragile strategies, which silently degraded guides after the recorded page's DOM shifted. The intent of the v1 attempt was to refuse fragile clicks at capture time with a toast, but in practice that produced two failure modes: authors lost clicks they thought were captured, and a noisy "N rejected" counter without enough context to act on. v2 inverts the flow — capture everything, classify on Stop, and put the decision back in the author's hands with the full list of fragile steps in front of them.

## How to verify

1. Open the block editor on any guide and start a recording.
2. Confirm the **Strict mode** toggle appears in the overlay next to **Auto-group**, and that it defaults to off.
3. With strict mode **off**, click around and stop — the recording is saved unconditionally as before.
4. Toggle strict mode **on**. Hover over an element that resolves to a fragile selector (e.g. a button with no `aria-label` / `data-testid` / stable text). Confirm the hover highlight turns red.
5. Click that fragile element, then click a button with a stable selector (e.g. anything with `data-testid`). Stop the recording.
6. Confirm the **Fragile Selectors Detected** modal appears listing only the fragile step, and that:
   - the overlay is no longer visible behind the modal,
   - clicking through the modal area does not record additional steps,
   - **Save** keeps the full recording (including the fragile step),
   - **Discard** drops the recording entirely (verify by starting a new recording — no leftover steps).
7. Repeat step 5 but record only stable-selector clicks. Stop. Confirm no modal appears and the recording is saved directly.
8. Toggle strict mode off mid-recording and verify the red hover-highlight returns to its non-strict appearance.

## Breaking changes

None. The feature is off by default and the new `selectorMethod` field on `RecordedStep` is optional, so older exported recordings continue to load unchanged.


leslie-heinzen:
Sorry about the noise for those of you tagged. Testing the new pr-summary skill updates on a not-ready branch (`--open` `--tidy-history` flags)

Issue #382, opened 2025-12-02
epic(devtools): Creation of Block editor


Jayclifford345:
This is an epic concerning the creation of the V1 version of the Block editor. This will be the new method for creating interactive guides within Pathfinder

PR #981, opened 2026-06-08
feat(snippets): add snippet schema + build-snippets CLI command

tomglenn:
## What

Extracts the **inert plumbing + build tooling** for the snippet feature so the upstream content repo (`grafana/interactive-tutorials`) can generate its snippet catalog in CI. This intentionally does **not** ship the user-facing feature — the runtime (snippet-engine resolver, content-renderer inlining, block-editor picker) stays on `feat/snippet-system` until it's verified end-to-end.

Two commits:
1. **JsonSnippet schema + types + `snippet-ref` block variant** (`JsonBlockSchemaNoRef` rejects nested refs at any depth). Type/schema definitions only — no renderer, parser, or engine wiring, so it's inert at runtime: nothing resolves or renders a `snippet-ref` yet.
2. **`build-snippets` CLI command** — generates `index.json` from a directory of snippet bodies. The body is the source of truth; the catalog is always rebuilt. Validates each body against `JsonSnippetSchema` and enforces filename == `id` (the resolver fetches `<id>.json`). Adds `npm run snippets:build` and fixes the CDN path in the types header.

## Why now (unblocks a cross-repo deadlock)

`grafana/interactive-tutorials` PR [#382](https://github.com/grafana/interactive-tutorials/pull/382) wires `pathfinder-cli build-snippets` into its validate + deploy workflows, which check out this repo at `ref: main`. That PR's CI can't pass until `build-snippets` exists on `main` — but we don't want to merge the full snippet feature to `main` before verifying it. This branch breaks the deadlock: it lands only the tooling/plumbing, so the tutorials catalog can be built and deployed to a **dev** CDN for verification, with the user-facing feature merging later.

## Tests

`build-snippets.test.ts` (9 cases) + the cherry-picked `json-snippet.schema.test.ts`. Typecheck, governance (architecture ratchet), and prettier all green.


Jayclifford345 (dismissed):
Overall looks good, my only nit is that the code comments are quite heavy in some places, but we haven't resolved that in our authoring workflows yet.

Jayclifford345 (approved):

PR #982, opened 2026-06-08
fix(e2e): treat step unmount as completion in guided substep loop

leslie-heinzen:
## Summary

The guide-runner hung on the final substep of multistep-guided steps when the substep's action caused the parent section to auto-collapse, unmounting the step element from the DOM. The runner then polled a detached locator until the global timeout fired. The substep loop now treats mid-flight unmount as a terminal completion signal, and the completion poller short-circuits once the step has been observed alive and is later detached.

## What changed

- `tests/e2e-runner/utils/guide-runner/execution.ts`
  - `runGuidedSubstepLoop` now returns `{ completed: boolean }`. A new `stepDetached()` helper runs at the top of each iteration and immediately after every substep action; either detection point returns `{ completed: true }`. The previous URL-change-specific inline re-query is removed — both the URL-change and the no-nav section-collapse cases are now handled uniformly.
  - `executeStep`'s guided branch captures the loop result and skips the post-loop `waitForCompletionWithObjectivePolling` when the loop already observed completion.
  - `waitForCompletionWithObjectivePolling` tracks `seenAlive`. A `count === 0` observation after the element was seen alive returns success regardless of `urlBeforeClick`. This retroactively fixes the same class of bug for non-guided multisteps whose final internal action triggers section auto-collapse without navigation.

## Test plan

- [ ] `npm run check`
- [ ] `npm run e2e -- --package <block-editor-tutorial-package>` — confirm the previously hanging multistep-guided section now completes
- [ ] Run a sample of other bundled guides via `npm run e2e -- --bundled` to confirm no regressions on happy-path guided steps (state-based completion still wins on the top-of-loop check)
- [ ] Manual: a guide whose final non-guided multistep action triggers section auto-collapse — verify it no longer hangs

## Risk and reversibility

Contained to the e2e runner; no product code changes. Reversible — revert the single commit. Behavioural change is "unmount counts as completion", which trusts a Pathfinder invariant (steps never re-mount under the same id within a run). If that invariant were ever violated by product code, the runner would falsely pass a step rather than hang — failure mode shifts from "hang on a real bug" to "miss a real bug", so monitor for unexpectedly green guides in CI after merge.

### DIFF
--- a/tests/e2e-runner/utils/guide-runner/execution.ts
+++ b/tests/e2e-runner/utils/guide-runner/execution.ts
@@ -206,22 +206,26 @@ export async function waitForCompletionWithObjectivePolling(
   const startTime = Date.now();
   const stepLocator = page.getByTestId(testIds.interactive.step(stepId));
   const urlBeforeClick = options?.urlBeforeClick;
+  let seenAlive = false;
 
   while (Date.now() - startTime < timeout) {
-    // If step element was detached (e.g. after formfill that changed URL), treat as completed
     const count = await stepLocator.count();
-    if (count === 0 && urlBeforeClick != null) {
-      const currentUrl = page.url();
-      if (currentUrl !== urlBeforeClick) {
-        return { completedViaObjectives: false };
-      }
+    if (count > 0) {
+      seenAlive = true;
+    }
+
+    // Unmount after the element was observed alive is a completion signal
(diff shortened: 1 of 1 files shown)


Jayclifford345 (approved):
LGTM!

PR #980, opened 2026-06-08
feat(ai-fix): add enableAiAutoHeal admin toggle + analytics/testIds/styles scaffolding

Jayclifford345:
## What & why

**PR1 of the PR-886 AI auto-heal breakdown** ([RFC](https://github.com/grafana/pathfinder-rfcs/blob/208096ae619519c32ff2f56aca84685bd4a2414d/plans/PR-886-AI-AUTOHEAL-BREAKDOWN.md)). [#886](https://github.com/grafana/grafana-pathfinder-app/pull/886) adds an AI-powered "Fix this" affordance to failing interactive steps in one ~3k-line, conflicting PR. The RFC splits it into nine progressively-mergeable PRs; this is the first — the **dark scaffolding** the rest build on.

## What's included

- `enableAiAutoHeal` admin toggle in **Interactive features** config (default **off**) + `DEFAULT_ENABLE_AI_AUTO_HEAL`, the `DocsPluginConfig` field, and `getConfigWithDefaults` wiring
- Four `AiFix{Offered,Accepted,Applied,Failed}` `UserInteraction` analytics enums
- `requirementAiFixButton` testId factory + `interactiveFeatures.enableAiAutoHeal` testId
- Shared `.interactive-*-ai-fix-btn` styles
- `UseStepCheckerReturn.requiresDomElement` field
- `GlobalInteractionBlocker.start/stopAdHocBlocking` — status text rendered via `textContent`, not `innerHTML`

## It lands dark

Every new symbol is **forward-declared and inert** — no active path references it until the orchestrator (PR7) and the buttons (PR8) land. The flag defaults off and nothing branches on it yet, so **behavior is unchanged for every existing user**. A grep confirms the four enums, the testId, the styles, `requiresDomElement`, and `start/stopAdHocBlocking` have zero consumers outside their own definitions; the flag is read only by the admin toggle.

## Note for reviewers

The breakdown was authored against the pre-rebase branch, where the blocker stub used lowercase `reftarget`/`targetaction`. `main` has since renamed those to camelCase `refTarget`/`targetAction` on `InteractiveElementData`; the stub was updated to match. `createMainContentOverlay` resolves the overlay via `getElementById('pageContent')` (never `querySelector(refTarget)`), so the empty-string stub is safe when PR7 invokes `startAdHocBlocking`.

Out of scope by design: the dead `FIX_TYPES.AI_FIX` constant is **not** added, and no `package.json`/lock changes (`@grafana/assistant` was already a dependency).

## Test plan

[...]

PR #983, opened 2026-06-08
feat(ai-fix): add AI-fix patch wire-format schema

Jayclifford345:
## What & why

**PR3 of the PR-886 AI auto-heal breakdown** ([RFC](https://github.com/grafana/pathfinder-rfcs/blob/208096ae619519c32ff2f56aca84685bd4a2414d/plans/PR-886-AI-AUTOHEAL-BREAKDOWN.md)). PR1 ([#980](https://github.com/grafana/grafana-pathfinder-app/pull/980)) is merged. PR2 (id-space reconciliation) was **folded into PR4** — `main` already settles the id space (`deriveStepId` + id-keyed completion + `reconcileSection` orphan self-heal), so the canonical-id/block-finder work moves into PR4 where `applyPatchToGuide` lives.

This PR adds the **AI-fix patch wire-format schema** — the Zod contract for what the Grafana Assistant is allowed to return when a user clicks the AI "Fix this" button on a failing interactive step.

## What's included

Two net-new files under `src/integrations/assistant-integration/`:
- `ai-fix-patch.schema.ts` — a `discriminatedUnion('type')` of:
  - `selector-patch` — replace a failing step's `reftarget`
  - `substep-selector-patch` — replace a `reftarget` inside a multistep/guided container (addressed by container id + positional index)
  - `prepend-step` — insert a new interactive step before the failing one; `newStep` round-trips through the authoritative `JsonInteractiveBlockSchema`
- `ai-fix-patch.schema.test.ts` — boundary tests

Every selector field passes `SafeSelectorSchema`: a deny-list (`<`, `>`, `` ` ``, `${`, `javascript:`, `data:`, `vbscript:`) + a 512-char cap. This is the security core — these strings flow into `querySelector`, so anything HTML/URL/template-shaped is rejected at the boundary.

## It lands dark

Net-new and inert — nothing imports `AiFixPatchSchema` until PR4 (`applyPatchToGuide`) and PR5 (the generation hook). No barrel export added yet (those PRs import the module directly). Zero behavior change for existing users.

## Test plan

- `npm run typecheck` ✅ — confirms `prepend-step`'s `JsonInteractiveBlockSchema.superRefine(...)` compiles against `main`'s Zod-4 schema
- `npm run lint` ✅ 0 errors · `prettier` ✅
- `jest ai-fix-patch.schema.test.ts` ✅ 12/12 — deny-list rejects HTML / `javascript:` / `${…}` / >512-char; `prepend-step` delegates to `JsonInteractiveBlockSchema` (non-`noop`-without-`reftarget` rejects, `noop`-without-`reftarget` accepts); unknown discriminator rejects
- governance/architecture (Tier 3 → Tier 0 import) ✅ 73/73

### DIFF
--- a/src/integrations/assistant-integration/ai-fix-patch.schema.test.ts
+++ b/src/integrations/assistant-integration/ai-fix-patch.schema.test.ts
@@ -11,6 +11,15 @@ describe('AiFixPatchSchema', () => {
       expect(result.success).toBe(true);
     });
 
+    it('accepts an attribute selector matching a javascript: href prefix', () => {
+      const result = AiFixPatchSchema.safeParse({
+        type: 'selector-patch',
+        targetStepId: 'step-1',
+        newReftarget: 'a[href^="javascript:"]',
+      });
+      expect(result.success).toBe(true);
+    });
+
     it('rejects an empty targetStepId', () => {
       const result = AiFixPatchSchema.safeParse({
         type: 'selector-patch',
@@ -66,6 +75,56 @@ describe('AiFixPatchSchema', () => {
         newReftarget: '[data-testid="run-query"]',
       });
       expect(result.success).toBe(true);
+    });
+
+    it('rejects a negative subStepIndex', () => {
+      const result = AiFixPatchSchema.safeParse({
+        type: 'substep-selector-patch',
+        containerId: 'multi-1',
+        subStepIndex: -1,
+        newReftarget: '[data-testid="run-query"]',
+      });
+      expect(result.success).toBe(false);
+    });
+
+    it('rejects a null subStepIndex', () => {
+      const result = AiFixPatchSchema.safeParse({
+        type: 'substep-selector-patch',
+        containerId: 'multi-1',
+        subStepIndex: null,
+        newReftarget: '[data-testid="run-query"]',
+      });
+      expect(result.success).toBe(false);
+    });
+
+    it('rejects an empty containerId', () => {
+      const result = AiFixPatchSchema.safeParse({
+        type: 'substep-selector-patch',
+        containerId: '',
+        subStepIndex: 0,
+        newReftarget: '[data-testid="run-query"]',
+      });
+      expect(result.success).toBe(false);
+    });
+
+    it('rejects a null containerId', () => {
+      const result = AiFixPatchSchema.safeParse({
+        type: 'substep-selector-patch',
+        containerId: null,
+        subStepIndex: 0,
+        newReftarget: '[data-testid="run-query"]',
+      });
+      expect(result.success).toBe(false);
+    });
+
+    it('rejects a dangerous newReftarget (SafeSelectorSchema applies to substeps)', () => {
+      const result = AiFixPatchSchema.safeParse({
+        type: 'substep-selector-patch',
+        containerId: 'multi-1',
+        subStepIndex: 0,
+        newReftarget: 'javascript:alert(1)',
+      });
+      expect(result.success).toBe(false);
     });
   });
 
--- a/src/integrations/assistant-integration/ai-fix-patch.schema.test.ts
+++ b/src/integrations/assistant-integration/ai-fix-patch.schema.test.ts
@@ -1,0 +1,137 @@
+import { AiFixPatchSchema } from './ai-fix-patch.schema';
+
+describe('AiFixPatchSchema', () => {
+  describe('selector-patch', () => {
+    it('accepts a well-formed selector patch', () => {
+      const result = AiFixPatchSchema.safeParse({
+        type: 'selector-patch',
+        targetStepId: 'step-1',
+        newReftarget: '[data-testid="save-button"]',
+      });
+      expect(result.success).toBe(true);
+    });
+
+    it('rejects an empty targetStepId', () => {
+      const result = AiFixPatchSchema.safeParse({
+        type: 'selector-patch',
+        targetStepId: '',
+        newReftarget: '.save',
+      });
+      expect(result.success).toBe(false);
+    });
+
+    it('rejects an HTML-shaped selector', () => {
+      const result = AiFixPatchSchema.safeParse({
+        type: 'selector-patch',
+        targetStepId: 'step-1',
+        newReftarget: '<script>alert(1)</script>',
+      });
+      expect(result.success).toBe(false);
+    });
+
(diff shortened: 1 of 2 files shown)


src/integrations/assistant-integration/ai-fix-patch.schema.test.ts
```diff
@@ -0,0 +1,137 @@
...
+        type: 'selector-patch',
+        targetStepId: 'step-1',
+        newReftarget: '.x'.repeat(300),
+      });
+      expect(result.success).toBe(false);
+    });
+  });
+
+  describe('substep-selector-patch', () => {
+    it('accepts a well-formed substep selector patch', () => {
```
moxious, line 70: add tests: invalid values for subStepIndex = (-1, null) should fail; that containerId = '' or containerId = null should fail.

MIght consider to prove that SafeParse gets called on the selector in the subStep example, but that's mostly pretty safe because of how it's reused.

src/integrations/assistant-integration/ai-fix-patch.schema.ts
```diff
@@ -0,0 +1,58 @@
...
+      }
+    }
+  }),
+  rationale: z.string().max(500).optional(),
+});
+
+const SubstepSelectorPatchSchema = z.object({
+  type: z.literal('substep-selector-patch'),
+  containerId: z.string().min(1),
+  subStepIndex: z.number().int().min(0),
```
moxious, line 53: maybe add comment for later PR:   TODO you have to validate that this is in bounds and not let people pass subStepIndex=100000

src/integrations/assistant-integration/ai-fix-patch.schema.ts
```diff
@@ -0,0 +1,58 @@
...
+import { JsonInteractiveBlockSchema } from '../../types/json-guide.schema';
+
+// Deny-list: reject HTML, template interpolation, and unsafe URL schemes (XSS guard); 512-char cap bounds querySelector cost.
+const DANGEROUS_SELECTOR_PATTERNS = ['<', '>', '`', '${', 'javascript:', 'data:', 'vbscript:'] as const;
+
+const SafeSelectorSchema = z
+  .string()
+  .min(1)
+  .max(512, 'Selector exceeds 512-character cap')
+  .refine((sel) => !DANGEROUS_SELECTOR_PATTERNS.some((p) => sel.toLowerCase().includes(p)), {
```
moxious: `includes` might be a little too blunt force here.   

This is a valid selector that's rejected:  `a[href^="javascript:"]`

I see what you're trying to do security wise here, you might discriminate between dangerous selectors which start with a certain string (javascript, data) vs. those that contain it.

moxious (changes requested):
very close to ready to merge imo; some very small suggestions on improving test coverage and a boundary condition before LGTM

src/integrations/assistant-integration/ai-fix-patch.schema.test.ts
Jayclifford345: Added in 262f1ef8: `subStepIndex` = -1 and `null` now reject, `containerId` = `` and `null` reject, plus a test that a dangerous `newReftarget` inside a substep patch is rejected (proves `SafeSelectorSchema` runs on substeps too, not just top-level patches).

src/integrations/assistant-integration/ai-fix-patch.schema.ts
Jayclifford345: The upper bound can't be checked at the schema layer — it has no view of the container's substep count. It's enforced one PR up in #984's `applyPatchToGuide`: `apply-ai-fix-patch.ts` returns an `out-of-range` error when `subStepIndex >= block.steps.length`, covered by the test "fails with a range error when subStepIndex is out of bounds" in `apply-ai-fix-patch.test.ts`. Skipped the TODO comment to stay within the repo's no-comment policy since the behavior already lands.

src/integrations/assistant-integration/ai-fix-patch.schema.ts
Jayclifford345: Good catch. Split the deny-list in 262f1ef8: `javascript:`/`data:`/`vbscript:` are now matched as a *prefix* (after trim + lowercase), so `a[href^="javascript:"]` passes while `javascript:alert(1)` is still rejected. HTML/template chars (`<`, `>`, backtick, `${`) stay blocked anywhere. Added a test locking in the attribute-selector case.

src/integrations/assistant-integration/ai-fix-patch.schema.ts
```diff
@@ -0,0 +1,67 @@
...
+      return !startsWithUnsafeScheme && !containsUnsafeSubstring;
+    },
+    { error: 'Selector contains a disallowed substring (HTML, template, or unsafe URL scheme)' }
+  );
+
+const SelectorPatchSchema = z.object({
+  type: z.literal('selector-patch'),
+  targetStepId: z.string().min(1),
+  newReftarget: SafeSelectorSchema,
+  rationale: z.string().max(500).optional(),
```
moxious, line 29: chatty LLM losing the whole fix because rationale ran one char long could be a problem.  Might want to allow anything and then use `.transform(s => s.slice(0, 500))` to cap effectively what will be seen instead of failing schema validation.   If an LLM gets too verbose, what behavior do you want?

moxious (approved):
I had one minor comment which shouldn't block because it's so minor, but worth taking a look at to confirm intent.

src/integrations/assistant-integration/ai-fix-patch.schema.ts
```diff
@@ -0,0 +1,67 @@
+import { z } from 'zod';
+
+import { JsonInteractiveBlockSchema } from '../../types/json-guide.schema';
+
+// XSS guard for strings flowing into querySelector: HTML/template chars are blocked
+// anywhere; URL schemes are blocked only as a prefix, so attribute matches like
+// a[href^="javascript:"] pass. 512-char cap bounds querySelector cost.
+const DANGEROUS_URL_SCHEMES = ['javascript:', 'data:', 'vbscript:'] as const;
```
leslie-heinzen, line 8: Kinda surprised the grafana CSP doesn't catch these. Does it allow unsafe-inline for some reason? 👀  

Looks fine, just curious.

Also, vbscript: 

Image

PR #984, opened 2026-06-08
feat(ai-fix): add pure applyPatchToGuide with pre/post schema validation

Jayclifford345:
## What & why

**PR4 of the PR-886 AI auto-heal breakdown.** Stacked on **PR3 (#983)** — this PR's base is `feat/ai-fix-patch-schema`; GitHub auto-retargets it to `main` once PR3 merges (a quick rebase may be needed if PR3 squash-merges).

Adds the pure **`applyPatchToGuide(guideJson, patch)`** — given a validated `AiFixPatch`, it locates the target block, applies the mutation, and re-validates, returning the new guide JSON or an error. No DOM, no React; **lands dark** (only PR7's orchestrator will call it).

## What's included

Three net-new files under `src/integrations/assistant-integration/`:
- **`apply-ai-fix-patch.ts`** — pipeline: `JSON.parse` → pre-validate (`JsonGuideSchema`) → deep-clone → materialize ids → mutate by patch type → post-validate → re-stringify. Mutators recurse sections + conditional branches; `selector-patch` rewrites `reftarget`, `substep-selector-patch` rewrites `steps[i].reftarget`, `prepend-step` splices a new block before the target. A post-validation failure discards the mutation — an invalid guide never reaches the renderer.
- **`ai-fix-step-id.ts`** — the folded-in **PR2** reconciliation. `materializeStepIds` reproduces the parser's `deriveStepId` walk so an **anonymous** block (no author id) is addressable by the same `stepId` a component dispatches. Reuses `main`'s `deriveStepId`; no parser change, no new dependency.
- **`apply-ai-fix-patch.test.ts`** — 20 cases.

## The id reconciliation (folded-in PR2), verified

`main` stores the derived `stepId` only on `props.stepId`, never on `block.id`, so the apply re-derives it. The risk is the materializer drifting from the parser — pinned by **contract tests** that parse a guide with anonymous blocks, read the real `props.stepId`, and assert `applyPatchToGuide` finds the block by it (top-level, section-nested, and multistep container). No `block.id` mutation is persisted to source; materialization happens on the apply's local clone only.

## Test plan

- `npm run typecheck` ✅
- `npm run lint` ✅ 0 errors · prettier ✅
- `jest apply-ai-fix-patch.test.ts` ✅ 20/20 — all patch types + section/conditional recursion + error paths + **anonymous-block re-derive contract tests** + post-validation-rejects
- governance/architecture (Tier-3 → Tier 0/1 imports) ✅ 73/73
- dark-landing grep: no production consumer of `applyPatchToGuide`/`materializeStepIds` (only the test)

### DIFF
--- a/src/integrations/assistant-integration/apply-ai-fix-patch.ts
+++ b/src/integrations/assistant-integration/apply-ai-fix-patch.ts
@@ -1,4 +1,4 @@
-import type { AiFixPatch } from './ai-fix-patch.schema';
+import { AiFixPatchSchema, type AiFixPatch } from './ai-fix-patch.schema';
 import { materializeStepIds } from './ai-fix-step-id';
 import { JsonGuideSchema } from '../../types/json-guide.schema';
 import {
@@ -90,6 +90,10 @@ function mutateSubstepReftarget(
 }
 
 export function applyPatchToGuide(guideJson: string, patch: AiFixPatch): ApplyResult {
+  if (!AiFixPatchSchema.safeParse(patch).success) {
+    return { ok: false, error: 'AI fix patch failed schema validation; cannot safely patch' };
+  }
+
   let parsed: unknown;
   try {
     parsed = JSON.parse(guideJson);
--- a/src/integrations/assistant-integration/apply-ai-fix-patch.ts
+++ b/src/integrations/assistant-integration/apply-ai-fix-patch.ts
@@ -1,0 +1,146 @@
+import type { AiFixPatch } from './ai-fix-patch.schema';
+import { materializeStepIds } from './ai-fix-step-id';
+import { JsonGuideSchema } from '../../types/json-guide.schema';
+import {
+  isConditionalBlock,
+  isSectionBlock,
+  type JsonBlock,
+  type JsonGuide,
+  type JsonGuidedBlock,
+  type JsonInteractiveBlock,
+  type JsonMultistepBlock,
+  type JsonStep,
+} from '../../types/json-guide.types';
+
+export type ApplyResult = { ok: true; newGuideJson: string } | { ok: false; error: string };
+
+function isInteractiveWithId(block: JsonBlock, id: string): block is JsonInteractiveBlock {
+  return block.type === 'interactive' && (block as JsonInteractiveBlock).id === id;
+}
+
+function isStepContainerWithId(block: JsonBlock, id: string): block is JsonMultistepBlock | JsonGuidedBlock {
+  return (block.type === 'multistep' || block.type === 'guided') && block.id === id;
+}
+
+function mutateMatchingBlock(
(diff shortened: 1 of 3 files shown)


src/integrations/assistant-integration/apply-ai-fix-patch.ts
```diff
@@ -0,0 +1,146 @@
...
+  try {
+    parsed = JSON.parse(guideJson);
+  } catch (e) {
+    return { ok: false, error: `Failed to parse guide JSON: ${e instanceof Error ? e.message : 'unknown'}` };
+  }
+
+  const beforeCheck = JsonGuideSchema.safeParse(parsed);
+  if (!beforeCheck.success) {
+    return { ok: false, error: 'Current guide failed schema validation; cannot safely patch' };
+  }
```
leslie-heinzen, line 107: Can we guarantee that the given `patch` object is schema-compliant always before it is provided to applyPatchToGuide?

If not, I think we should do it internally to this method as a backstop, like with JsonGuideSchema (`AiFixPatchSchema.safeParse`):

```ts
const patchCheck = AiFixPatchSchema.safeParse(patch);
if (!patchCheck.success) {
  return { ok: false, error: 'AI fix patch failed schema validation; cannot safely patch' };
}

const safePatch = patchCheck.data;
```

Namely, `selector-patch` isn't adequately safeguarded here, which I verified with a quick failing test in `apply-ai-fix-patch.test.ts`:

```ts
it('rejects an unsafe selector patch even when called directly with a typed patch', () => {
  const result = applyPatchToGuide(makeGuide([validStep]), {
    type: 'selector-patch',
    targetStepId: 'step-1',
    newReftarget: '<script>alert(1)</script>', // this returns okay, but shouldn't
  });

  expect(result.ok).toBe(false);
  if (!result.ok) {
    expect(result.error).toMatch(/patch/i);
  }
});
```

leslie-heinzen (commented):
Question / possible change request depending on the answer to my question!

Mostly looks good, just one concern.

src/integrations/assistant-integration/apply-ai-fix-patch.ts
Jayclifford345: Good catch — fixed in c11a144e. Added `AiFixPatchSchema.safeParse(patch)` at the top of `applyPatchToGuide`, mirroring the existing guide pre/post checks, plus your test.

Note this makes the `JsonGuideSchema` post-validation unreachable by a schema-valid patch (the guide treats `reftarget` as a plain string, looser than the patch's `SafeSelectorSchema`) — so I kept it as defense-in-depth against future patch/guide schema divergence rather than removing it. The old malformed-patch test now lands on this front gate instead.

leslie-heinzen (approved):

PR #986, opened 2026-06-09
refactor(docs-panel): extract shared loadTab helpers; route openDocsPage through loadTab

Jayclifford345:
## Summary

**P1 (A1)** of the [INTERACTIVE-STATE-CONSOLIDATION breakdown](https://github.com/grafana/pathfinder-rfcs/pull/7)

`docs-panel.tsx` had two near-parallel content loaders — `loadTabContent` (guide path) and `loadDocsTabContent` (docs/package path) — duplicating the same loading / success / failure / save-tabs boilerplate 4+ times. Separately, `openDocsPage` still called the deprecated `loadDocsTabContent` **directly**, bypassing the unified `loadTab` dispatcher.

## What changed

- **Extract three private helpers** adjacent to `loadTab`:
  - `setTabLoading(tabId)` — the loading-state map + `setState` (was textually identical in both loaders).
  - `finishTabSuccess(tabId, buildPatch)` — applies a per-tab patch, sets `isLoading:false/error:null`, persists, and **returns the freshly-built tab**. The callback form preserves the docs path's reads of the existing tab (`t.baseUrl`/`t.packageInfo`/`t.type`); the return value lets the docs path read the telemetry title from the pre-`setState` array (the invariant the old inline comment defended is now structural).
  - `failTab(tabId, message)` — the error map + `setState` + persist (was duplicated 4×: each loader's `else` and `catch`).
- **Route `openDocsPage` through `loadTab`** instead of calling `loadDocsTabContent` directly. `loadTab`'s `shouldUseDocsLoader`/`packageInfo` routing re-enters the docs loader with the same args, so `loadTab` is now the single entry point and the deprecated internals are reached only from there.
- **Guardrail honored:** the two loaders are *not* folded — their divergent guide-journey vs docs-alignment work stays in two purpose-named functions sharing the helpers.

Net **−130/+92** in `docs-panel.tsx`.

## Tests

- New `openDocsPage` routing tests in `docs-panel.alignment.test.ts`: loading → success and loading → error transitions, asserting the docs loader is reached through `loadTab`.
- Made the alignment harness's `shouldUseDocsLoader` mock realistic (returns `true`) now that `openDocsPage` consults it via `loadTab` — previously the bare `jest.fn()` returned `undefined`, which would mis-route `packageInfo`-less opens to the guide loader.

## Verification

- `npm run typecheck` — clean
- frontend `eslint` — 0 errors (2 pre-existing `_recordAutoLaunchSource` deprecation warnings, untouched lines)
- `prettier --check` — clean
- `npm run test:ci` — **232 suites / 3958 tests pass** (no regressions outside docs-panel from the routing change)
- governance (`architecture` / `import-graph`) — 73/73

> `npm run check`'s `lint:go` step fails on a vendored `node_modules/flatted/golang` file unrelated to this PR (zero Go files changed); all frontend gates pass.

### DIFF
--- a/src/components/docs-panel/docs-panel.alignment.test.ts
+++ b/src/components/docs-panel/docs-panel.alignment.test.ts
@@ -292,6 +292,8 @@ describe('CombinedLearningJourneyPanel — implied-0th-step alignment', () => {
   beforeEach(() => {
     jest.clearAllMocks();
     mockGetLocation.mockReturnValue({ pathname: '/explore', search: '' });
+    // openDocsPage routes through loadTab → shouldUseDocsLoader; force the docs loader.
+    jest.requireMock('./utils').shouldUseDocsLoader.mockReturnValue(true);
   });
 
   describe('loadDocsTabContent — pendingAlignment decision', () => {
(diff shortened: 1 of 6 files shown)


moxious (changes requested):
- Rebase onto main — branch silently reverts PR #985's e2e hang-fix. The diff removes bounded getAttribute({ timeout: 2000 }) guards and unmount/count() === 0 short-circuits in tests/e2e-runner/utils/guide-runner/execution.ts (waitForSubstepAdvance ~L304, waitForFormfillSettle ~L352, retry ~L364). Not mentioned in PR description — accidental revert from branching pre-#985. Every guided-step E2E will hang to global timeout again.
- Finish or drop the "deprecated" plumbing. docs-panel.tsx:359-363 still routes loadTab into loadDocsTabContent / loadTabContent behind @ts-expect/no-deprecated eslint disables, and types.ts:66-72 still exposes them on CombinedPanelInterface. No external callers remain. Either make them private and remove from the interface, or explicitly call this out as part-1 with a follow-up planned. JSDoc @deprecated is not compiler-enforced — drift waiting to happen.
- Add a test for openDocsPage with packageInfo set. New tests only cover the shouldUseDocsLoader → true branch (mockReturnValue(true) in beforeEach). The loadTab dispatcher's second trigger (options?.packageInfo != null) is unexercised, and openDocsPage is its primary call site.
- Drop the inline comment at docs-panel.tsx:781-782 ("Persist auto-derived packageInfo…") — narrates a one-line || spread default (QC8 #7).

Jayclifford345:
Thanks @moxious all four addressed.

**1. Rebase onto main**

**2. Finish the deprecated plumbing.** Took the "make them private" route. `loadTabContent` and `loadDocsTabContent` are now `private` and removed from `DocsPanelModelOperations` (`types.ts`); the `@deprecated` JSDoc and both `@typescript-eslint/no-deprecated` disables in `loadTab` are gone, so the compiler now enforces `loadTab` as the sole entry point. No external callers remained — I also dropped the now-unsatisfiable `loadTabContent` member from the link-handler's structural `model` type (it only ever called `loadTab`) plus its two dead test mocks.

**3. Test for `openDocsPage` with packageInfo.** Added `reaches the docs loader via the packageInfo trigger when shouldUseDocsLoader is false` in `docs-panel.alignment.test.ts`. It forces `shouldUseDocsLoader → false`, so reaching `loadDocsTabContentResult` can only mean `options.packageInfo != null` routed it; it also asserts the `packageInfo` is carried through to the loader.

**4. Drop the narrating comment.** Removed the two-line comment over `packageInfo ?? t.packageInfo`.

moxious (approved):

PR #987, opened 2026-06-09
feat(schema): accept camelCase aliases in JSON guides via a pre-schema normalizer

Jayclifford345:
## Summary

**P2 (A5)** of the [INTERACTIVE-STATE-CONSOLIDATION breakdown](https://github.com/grafana/pathfinder-rfcs/pull/7) — independent of P1 (#986). Author-facing only.

Hand-written JSON guides that use camelCase field names (`targetAction` / `refTarget` / `targetValue`) previously **failed validation**: the schema requires the canonical lowercase forms (`action` / `reftarget` / `targetvalue`), even though the runtime parser already tolerates the camelCase aliases via `?? block.targetAction` fallbacks. `json-parser-aliases.test.ts` literally documented this gap as "a follow-up" — this is that follow-up.

## What changed

- **New `src/validation/normalize-guide-aliases.ts`** — a pure, idempotent `normalizeJsonGuideAliases(raw)` that deep-clones and recursively rewrites the three camelCase aliases to canonical lowercase (`targetAction→action`, `refTarget→reftarget`, `targetValue→targetvalue`), **canonical wins on conflict**, and removes the alias key.
- **Wired into `validateGuide`** (`src/validation/validate-guide.ts`) — runs once, immediately before `JsonGuideSchema.safeParse` *and* the `detectUnknownFields` traversal (so a both-present guide doesn't draw a spurious "unknown field `targetAction`" warning). This is the single chokepoint all paths converge on (runtime `parseJsonGuide`, content-fetcher, CLI `validate`).
- **Unchanged on purpose:** `json-guide.schema.ts` keeps the canonical field names, and `cli/utils/schema-options.ts` walks the schema's shape — so the emitted CLI flags stay `--action` / `--reftarget` / `--targetvalue` (no `--target-action`). The runtime `?? camelCase` fallbacks in `json-parser.ts` remain as the runtime read path. One validator only.

## Tests

- New `normalize-guide-aliases.test.ts` — rename per alias, canonical-wins, recursion into `blocks[]` / multistep `steps[]` / nested section + conditional branches, idempotency, no-mutation, non-object passthrough.
- Updated `json-parser-aliases.test.ts` — the case that asserted camelCase-only is *rejected* now asserts it **validates** and the props read correctly; refreshed the module docstring. The "canonical wins when both present" case stays green.

## Verification

`npx jest src/validation src/cli` — **38 suites / 854 tests pass** (CLI option-generation suites confirm the flag set is unchanged) · typecheck clean · eslint 0 errors · prettier clean · governance 73/73.

### DIFF
--- a/src/validation/normalize-guide-aliases.ts
+++ b/src/validation/normalize-guide-aliases.ts
@@ -1,0 +1,36 @@
+/**
+ * Pre-schema normalizer: rewrites the camelCase field aliases the runtime
+ * parser tolerates (`targetAction`/`refTarget`/`targetValue`) to their
+ * canonical lowercase schema names, so hand-written camelCase guides pass
+ * `JsonGuideSchema`. Canonical wins when both are present. Pure and
+ * idempotent; runs immediately before the schema in `validateGuide`.
+ */
+
+const FIELD_ALIASES: ReadonlyArray<readonly [alias: string, canonical: string]> = [
+  ['targetAction', 'action'],
+  ['refTarget', 'reftarget'],
+  ['targetValue', 'targetvalue'],
+];
+
+export function normalizeJsonGuideAliases(raw: unknown): unknown {
+  if (Array.isArray(raw)) {
+    return raw.map(normalizeJsonGuideAliases);
+  }
+  if (raw === null || typeof raw !== 'object') {
+    return raw;
+  }
+
+  const out: Record<string, unknown> = {};
+  for (const [key, value] of Object.entries(raw as Record<string, unknown>)) {
+    out[key] = normalizeJsonGuideAliases(value);
+  }
+  for (const [alias, canonical] of FIELD_ALIASES) {
+    if (alias in out) {
+      if (!(canonical in out)) {
+        out[canonical] = out[alias];
+      }
+      delete out[alias];
+    }
+  }
+  return out;
+}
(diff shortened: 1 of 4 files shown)


src/validation/normalize-guide-aliases.ts
```diff
@@ -0,0 +1,36 @@
...
+ * idempotent; runs immediately before the schema in `validateGuide`.
+ */
+
+const FIELD_ALIASES: ReadonlyArray<readonly [alias: string, canonical: string]> = [
+  ['targetAction', 'action'],
+  ['refTarget', 'reftarget'],
+  ['targetValue', 'targetvalue'],
+];
+
+export function normalizeJsonGuideAliases(raw: unknown): unknown {
```
moxious, line 15: This function is very weirdly written considering that we have explicit guidance in the repo to use a clean functional style.  This entire function can easily be rewritten with map/filter/reduce functions on side-effect free arrays, which was the style that we adopted.

Makes me scratch my head and wonder why the agent wrote it this way, given this.   It is not clear to me that this comment is a big enough deal to request changes on code that works, but it's something I noticed.

Image

Jayclifford345: Good catch!

moxious (approved):

PR #993, opened 2026-06-09
feat(ai-fix): add useAiFixGeneration hook + useIsAssistantAvailable

Jayclifford345:
## What & why

**PR5 of the PR-886 AI auto-heal breakdown** ([RFC](https://github.com/grafana/pathfinder-rfcs/pull/5)). PR1 ([#980](https://github.com/grafana/grafana-pathfinder-app/pull/980)) and PR3 ([#983](https://github.com/grafana/grafana-pathfinder-app/pull/983)) are merged; this stacks on PR4 ([#984](https://github.com/grafana/grafana-pathfinder-app/pull/984)).

Adds the React hook that asks the Grafana Assistant for an AI-fix patch when a failing interactive step has no deterministic fix, plus a standalone `useIsAssistantAvailable` hook for PR8's button gates.

## What's included

- `useAiFixGeneration.hook.ts` — `SYSTEM_PROMPT`, `buildUserPrompt`, the no-confident-fix sentinel detection, `parseAssistantPatch`, and `useAiFixGeneration(contentKey)` over `main`'s existing `useAssistantGeneration` (`assistantId: 'ai-fix'`). Parsing runs sentinel-check → `AiFixPatchSchema.safeParse` (PR3's schema).
- `useIsAssistantAvailable` added to `assistant-dev-mode.ts` (subscribes to `getIsAssistantAvailable()`), barrel-exported from `index.ts`.
- `useAiFixGeneration.hook.test.ts` — pure-function tests for `parseAssistantPatch` / `buildUserPrompt`.

RFC §5.4 trims applied in this PR: deleted the dead `screenshotDataUrl` plumbing, dropped the file/JSDoc headers per the repo's comment policy, and **kept the one load-bearing comment** — sentinel detection MUST run before schema validation, because the literal `<unchanged>` the prompt emits trips `SafeSelectorSchema`'s `<`/`>` deny-list. No `package.json` change — `@grafana/assistant` is already a dependency.

## It lands dark

Net-new and inert. Nothing imports `useAiFixGeneration` until PR7 (orchestrator); `useIsAssistantAvailable` has no consumer until PR8 (buttons). Zero behavior change for existing users.

## Test plan

- `npm run typecheck` ✅ · `npx eslint` ✅ 0 errors · `prettier --check` ✅
- `jest useAiFixGeneration.hook.test.ts` ✅ 8/8 — bare/fenced patch parse, non-JSON + schema-failure error mapping, prepend-step delegation, the `<unchanged>` sentinel-before-validation must-keep, and `buildUserPrompt` envelope shape
- `assistant-integration` suite ✅ 46/46 · architecture/governance ✅ 15/15
- dark verified by grep: only the hook + its test reference `useAiFixGeneration`; `useIsAssistantAvailable` only the definition + barrel

### DIFF
--- a/src/integrations/assistant-integration/useAiFixGeneration.hook.test.ts
+++ b/src/integrations/assistant-integration/useAiFixGeneration.hook.test.ts
@@ -1,4 +1,19 @@
-import { buildUserPrompt, parseAssistantPatch } from './useAiFixGeneration.hook';
+import { renderHook, act } from '@testing-library/react';
+
+import { buildUserPrompt, parseAssistantPatch, useAiFixGeneration } from './useAiFixGeneration.hook';
+
+const mockRawGenerate = jest.fn();
+
+jest.mock('./useAssistantGeneration.hook', () => {
+  const actual = jest.requireActual('./useAssistantGeneration.hook');
+  return {
+    ...actual,
+    useAssistantGeneration: jest.fn(() => ({
+      isAssistantAvailable: true,
+      generate: mockRawGenerate,
+    })),
+  };
+});
 
 describe('parseAssistantPatch', () => {
   it('accepts a bare selector-patch JSON object', () => {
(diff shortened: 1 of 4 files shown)


src/integrations/assistant-integration/useAiFixGeneration.hook.ts
```diff
@@ -0,0 +1,183 @@
...
+          } else {
+            setError(result.error);
+          }
+          setIsGenerating(false);
+        },
+        onError: (err) => {
+          setError(err);
+          setIsGenerating(false);
+        },
+      });
```
leslie-heinzen: potentially minor state bug here.

If rawGenerate bails before onError is invoked for some reason then we are stuck with `isGenerating = true`. Should wrap this in a try/catch or try/finally to make sure we reset isGenerating if there is a failure outside the callback path.

src/integrations/assistant-integration/useAiFixGeneration.hook.test.ts
```diff
@@ -0,0 +1,95 @@
+import { buildUserPrompt, parseAssistantPatch } from './useAiFixGeneration.hook';
+
+describe('parseAssistantPatch', () => {
```
leslie-heinzen: Any reason not to build the hook tests here like other ones we have (renderHook, act, etc)? This seems to only test the major helper methods but skips over testing the hook itself. Not super critical, just seems like a little gap in testing.

```ts
// for example...
    const { result } = renderHook(() => useAiFixGeneration(mockContentKey));

    mockRawGenerate.mockImplementation(({ onComplete }) => {
      onComplete?.(
        JSON.stringify({ type: 'selector-patch', targetStepId: 'step-1', newReftarget: '[data-testid="x"]' })
      );
    });

    await act(async () => {
      await result.current.generate({
        guideJson: '{}',
        failingStepId: 'step-1',
        failingReftarget: '.old',
        failingAction: 'button',
        domHint: 'none',
      });
    });

// etc...

```

leslie-heinzen (commented):
Looks pretty good. Couple of thoughts.

src/integrations/assistant-integration/useAiFixGeneration.hook.ts
Jayclifford345: Good catch! Wrapped the `rawGenerate` await in a `try/catch` that resets `isGenerating` and surfaces the error if `rawGenerate` rejects before either callback fires.

Worth noting I deliberately used `catch` rather than `finally`: `generate` resolves *before* its streaming `onComplete`/`onError` callbacks run (the mock schedules `onComplete` via `setTimeout` and resolves immediately), so a `finally` would clear the loading flag mid-stream. The `catch` only fires on an actual rejection.

src/integrations/assistant-integration/useAiFixGeneration.hook.test.ts
Jayclifford345: Agreed, added New `describe('useAiFixGeneration')` block uses `renderHook`/`act` (mocking `useAssistantGeneration` to control `rawGenerate`)

leslie-heinzen (approved):

PR #994, opened 2026-06-09
feat(cli): add --clean flag to e2e command for isolated docker-compose runs

leslie-heinzen:
## Summary

Adds `--clean` to `pathfinder-cli e2e`. When set, the runner spins up a dedicated `pathfinder-e2e` docker-compose project on host port 3010, resets between guides, and tears it down at the end. Developers can keep their normal dev stack running on port 3000 in parallel. `--clean` never touches it.

## What to look at

- **[`src/cli/commands/e2e.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/db812b8b0858a689b5becc953594809dae100f6a/src/cli/commands/e2e.ts)** — new `--clean` and `--clean-ready-timeout-ms` options. `runDockerCompose` / `teardownDockerSync` always invoke compose with `-f docker-compose.yaml -f docker-compose.e2e.yaml -p pathfinder-e2e`, so `down -v` only nukes `pathfinder-e2e_grafana-data` and never the dev's `grafana-data`. When `--clean` is set and `--grafana-url` is left at the default, the URL auto-switches to `http://localhost:3010` so the pre-flight health check and Playwright runs hit the isolated stack. Teardown is registered via `process.on('exit')` plus `SIGINT`/`SIGTERM` handlers and uses `execSync` so it runs cleanly inside the synchronous exit listener — looks unusual at a glance but is intentional: `process.exit()` does not await async listeners.
- **[`docker-compose.e2e.yaml`](https://github.com/grafana/grafana-pathfinder-app/blob/db812b8b0858a689b5becc953594809dae100f6a/docker-compose.e2e.yaml)** (new) — overlay that overrides the hardcoded `container_name` in `.config/docker-compose-base.yaml` (renamed to `grafana-pathfinder-app-e2e` to avoid the daemon-wide name collision) and remaps host ports (Grafana 3010, Prometheus 9091, Loki 3101, Alloy 12346). Uses Compose's `!override` YAML tag on each `ports:` list so the overlay fully replaces the inherited port mapping instead of appending to it — without `!override`, the resolved config would contain both 3000 and 3010, and the e2e container would fail to bind 3000 whenever a dev stack was up. The `!override` tag requires Docker Compose v2.20+ (Sep 2023); Docker Desktop has shipped with this for ~2 years.

## Why

* For bundled e2e runs across many guides, residual state from one guide (provisioned data sources, dashboards, plugin settings) leaks into the next via the persisted `grafana-data` volume. 
* Testing the same state-mutating guide repeatedly creates different states on repeat runs. 

This can introduce unexpected failures.

The isolated `pathfinder-e2e` project gives clean-slate runs without affecting the local dev environment — both stacks share the same bind-mounted `../dist` plugin code, so an iterative `npm run dev` cycle works transparently across both.

## How to verify

1. **Dev stack survives a `--clean` run.** Start your normal dev stack: `npm run server` (Grafana on `:3000`). In another terminal: `pathfinder-cli e2e bundled:block-editor-tutorial --clean`. While it's running, confirm `docker ps` shows both `grafana-pathfinder-app` (dev) and `grafana-pathfinder-app-e2e` (clean). After it finishes, confirm the dev container is still up and `docker volume ls` shows `grafana-pathfinder-app_grafana-data` but not `pathfinder-e2e_grafana-data` (the latter was wiped on teardown).
2. **Between-guide reset works.** `pathfinder-cli e2e bundled:block-editor-tutorial bundled:welcome-to-grafana --clean --verbose` and watch the output for `🧹 Resetting docker compose between guides...` between each guide.
3. **Ctrl+C still tears down.** Start `pathfinder-cli e2e bundled:block-editor-tutorial --clean`, press Ctrl+C mid-run, confirm `docker ps -a --filter project=pathfinder-e2e` returns no containers and the volume is gone.
4. **Baseline path unchanged.** `pathfinder-cli e2e bundled:block-editor-tutorial` (no `--clean`) runs against the existing local stack on `:3000` like before.

## Breaking changes

None. The flag is opt-in; existing invocations without `--clean` behave exactly as before. Requires Docker Compose v2.20+ for the `!override` YAML tag, but only when `--clean` is actually used — runs without the flag are unaffected.

### DIFF
--- a/docker-compose.e2e.yaml
+++ b/docker-compose.e2e.yaml
@@ -1,0 +1,20 @@
+# Compose overlay used by `pathfinder-cli e2e --clean` to run the e2e stack
+# in isolation from any local dev `docker compose up` stack on the same machine.
+#
+# Used as:
+#   docker compose -f docker-compose.yaml -f docker-compose.e2e.yaml \
+#                  -p pathfinder-e2e up -d
+services:
+  grafana:
+    container_name: grafana-pathfinder-app-e2e
+    ports: !override
+      - '3010:3000'
+  prometheus:
+    ports: !override
+      - '9091:9090'
+  loki:
+    ports: !override
+      - '3101:3100'
+  alloy:
+    ports: !override
+      - '12346:12345'
--- a/docs/developer/E2E_TESTING.md
+++ b/docs/developer/E2E_TESTING.md
@@ -24,6 +24,9 @@ npx pathfinder-cli e2e --bundled
 
 # Test a specific bundled guide by name
 npx pathfinder-cli e2e bundled:welcome-to-grafana
+
+# Run against an isolated, clean-slate docker-compose stack (see below)
+npx pathfinder-cli e2e --bundled --clean
 ```
 
 ## CLI reference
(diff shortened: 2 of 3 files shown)


moxious (changes requested):
I don't find any issues with the code per se, but worth a few lines to document the new flag here:

https://github.com/grafana/grafana-pathfinder-app/blob/main/docs/developer/E2E_TESTING.md

Jayclifford345 (approved):
LGTM with new docs

moxious (approved):

PR #995, opened 2026-06-09
refactor(ai-fix): extract DOM-context collection + confidence gate into testable modules

Jayclifford345:
## What & why

**PR6 of the PR-886 AI auto-heal breakdown** ([RFC](https://github.com/grafana/pathfinder-rfcs/pull/5)). Stacks on PR5 ([#993](https://github.com/grafana/grafana-pathfinder-app/pull/993)). Depends only on the merged patch schema (PR3 #983) + `src/lib/dom` — nothing from PR4/PR5.

In PR #886 all the prompt/gate helpers live **inline** in the 716-line `AiFixOrchestrator.tsx`, and the safety-critical confidence gate (`evaluatePatchConfidence` — it decides whether an AI patch is written into a guide) has **never had a test**, with a surrounding comment that **lies** (claims a token-overlap "Bar chart → Bar gauge" rejection the code doesn't perform). This PR extracts those helpers into three testable non-React modules and adds the unit tests, so PR7's orchestrator shell stays thin and imports tested helpers.

## What's included

Three net-new modules under `src/integrations/assistant-integration/` (+ co-located tests):
- `ai-fix-confidence.ts` — `evaluatePatchConfidence` (accepts iff the proposed selector resolves to ≥1 live element via `resolveSelector` → `querySelectorAllEnhanced`). **Must-fixes applied:** dropped the unused `originalReftarget` parameter; rewrote the doc to describe only the live-DOM check (no fictional token-overlap).
- `ai-fix-dom-context.ts` — `collectDomContext` + `nearMatches`, `scoreCandidate`, `isNavPollution`, `isPathfinderInternal`, `describeElement`, `tokensFromSelector`, `tagFromSelector`.
- `ai-fix-step-content.ts` — `extractStepContent` + `findGuideBlockById` (**renamed** from `findBlockById` to avoid shadowing `cli/utils/package-io/tree.ts`).

Function-level JSDoc and the "runtime audit logs / CONTEXT-VS-DOM logs" dead-process comments stripped per the repo's comment policy; code logic preserved.

## It lands dark

Net-new and inert — nothing imports these modules until PR7 wires the orchestrator. No barrel export, no `package.json` change. Note: the call-site comment the RFC flags at `AiFixOrchestrator.tsx:648-653` lives in the orchestrator (PR7's net-new file); PR7 must call `evaluatePatchConfidence(patch)` (no second arg) with an accurate comment.

## Test plan

- `npm run typecheck` ✅ · `eslint` ✅ 0 errors · `prettier --check` ✅
- `jest` ✅ 25/25 across the three suites — confidence gate (live-match → ok, zero-match → reject, prepend-step-without-reftarget → ok, invalid-CSS → reject, no-selector → reject); DOM-context (near-match surfacing, nav/pathfinder-internal filtering, headings, plus unit tests for the pure helpers); step-content (nested section/conditional lookup, sub-step comment/hint, 1500-char cap, invalid-JSON/missing-block → '')
- architecture/governance ✅ 15/15 · dark verified by grep (no consumers outside the modules + tests)

### DIFF
--- a/src/integrations/assistant-integration/ai-fix-confidence.test.ts
+++ b/src/integrations/assistant-integration/ai-fix-confidence.test.ts
@@ -1,9 +1,16 @@
 jest.mock('../../lib/dom', () => ({
   resolveSelector: (selector: string) => selector,
-  querySelectorAllEnhanced: (selector: string) => ({
-    elements: Array.from(document.querySelectorAll(selector)),
-    usedFallback: false,
-  }),
+  querySelectorAllEnhanced: (selector: string) => {
+    try {
+      return {
+        elements: Array.from(document.querySelectorAll(selector)),
+        usedFallback: false,
+        originalSelector: selector,
+      };
+    } catch {
+      return { elements: [], usedFallback: true, originalSelector: selector, effectiveSelector: 'ERROR' };
+    }
+  },
 }));
 
 import { evaluatePatchConfidence } from './ai-fix-confidence';
@@ -49,12 +56,12 @@ describe('evaluatePatchConfidence', () => {
     expect(evaluatePatchConfidence(patch)).toEqual({ ok: true });
   });
 
-  it('rejects an invalid CSS selector', () => {
+  it('rejects an invalid CSS selector (it resolves to no live element)', () => {
     const patch: AiFixPatch = { type: 'selector-patch', targetStepId: 's1', newReftarget: '[[[not-valid' };
     const result = evaluatePatchConfidence(patch);
     expect(result.ok).toBe(false);
     if (!result.ok) {
-      expect(result.reason).toMatch(/not valid CSS/);
+      expect(result.reason).toMatch(/does not match any element/);
     }
   });
 
--- a/src/integrations/assistant-integration/ai-fix-confidence.test.ts
+++ b/src/integrations/assistant-integration/ai-fix-confidence.test.ts
@@ -1,0 +1,74 @@
+jest.mock('../../lib/dom', () => ({
+  resolveSelector: (selector: string) => selector,
+  querySelectorAllEnhanced: (selector: string) => ({
+    elements: Array.from(document.querySelectorAll(selector)),
+    usedFallback: false,
+  }),
+}));
+
+import { evaluatePatchConfidence } from './ai-fix-confidence';
+import type { AiFixPatch } from './ai-fix-patch.schema';
+
+describe('evaluatePatchConfidence', () => {
+  afterEach(() => {
+    document.body.innerHTML = '';
+  });
+
+  it('accepts a selector-patch whose selector resolves to a live element', () => {
+    document.body.innerHTML = '<button data-testid="run-query">Run</button>';
+    const patch: AiFixPatch = { type: 'selector-patch', targetStepId: 's1', newReftarget: '[data-testid="run-query"]' };
+    expect(evaluatePatchConfidence(patch)).toEqual({ ok: true });
+  });
+
+  it('rejects a selector-patch whose selector matches nothing on the page', () => {
+    document.body.innerHTML = '<button data-testid="other">x</button>';
+    const patch: AiFixPatch = { type: 'selector-patch', targetStepId: 's1', newReftarget: '[data-testid="missing"]' };
+    const result = evaluatePatchConfidence(patch);
+    expect(result.ok).toBe(false);
+    if (!result.ok) {
+      expect(result.reason).toMatch(/does not match any element/);
+    }
+  });
+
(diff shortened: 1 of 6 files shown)


src/integrations/assistant-integration/ai-fix-confidence.ts
```diff
@@ -0,0 +1,44 @@
...
+  let matchCount = 0;
+  let cssError: string | null = null;
+  try {
+    matchCount = querySelectorAllEnhanced(resolved).elements.length;
+  } catch (e) {
+    cssError = e instanceof Error ? e.message.slice(0, 200) : 'unknown';
+  }
+  if (cssError) {
+    return { ok: false, reason: 'proposed selector is not valid CSS' };
+  }
```
leslie-heinzen: Hm, based on what I am grokking with querySelectorAllEnhanced it won't actually ever throw here, so this catch block and `if (cssError)` block are just dead code. The internal `document.querySelectorAll` is handled with try/catch, and a SelectorResult object is always returned.

We will still handle it in the following `if` block (`mathCount === 0`) since there will be an empty elements array returned. The failure reason just won't be differentiated, i.e., we'll only ever see 'proposed selector does not match any element on the current page'.

I think this is okay since the bad selector is still prevented from being evaluated as `ok`? 

But the related testing code is misleading (will make a related comment over there).

src/integrations/assistant-integration/ai-fix-confidence.test.ts
```diff
@@ -0,0 +1,74 @@
+jest.mock('../../lib/dom', () => ({
+  resolveSelector: (selector: string) => selector,
+  querySelectorAllEnhanced: (selector: string) => ({
+    elements: Array.from(document.querySelectorAll(selector)),
+    usedFallback: false,
+  }),
```
leslie-heinzen: Kind of a nit (related to my other comment): this mock doesn't match the real code's behavior; normally I wouldn't say it matters a ton, but we are asserting based on try/catch behavior of this method, so the inaccuracy causes false expectations in testing.

In this mock, `document.querySelectorAll(selector)` can throw as the result of a bad selector, but that is not how the actual `querySelectorAllEnhanced` works. In the real version, the internal code that calls `document.querySelectorAll` is handled with a try/catch, and ultimately a SelectorResult is returned; `querySelectorAllEnhanced` will _never_ throw.

This just means that the "rejects an invalid CSS selector" is somewhat incorrect (the asserted `reason` is wrong). You can verify this by updating the mock to mirror the real behavior:

```ts
  querySelectorAllEnhanced: (selector: string) => {
    try {
      return {
        elements: Array.from(document.querySelectorAll(selector)),
        usedFallback: false,
        originalSelector: selector,
      };
    } catch {
      return {
        elements: [],
        usedFallback: true,
        originalSelector: selector,
        effectiveSelector: 'ERROR',
      };
    }
  }
```

Then watch the aforementioned test fail.

leslie-heinzen: TBH I don't think it's super important in this case to make the production method handle the different failure reasons, but the mock should match reality in this way.

leslie-heinzen (changes requested):
Looks good overall. 👍🏻  

At the very least, I'd like some test/mocking code updated, and possibly a small refactor of `evaluatePatchConfidence` can be considered.

src/integrations/assistant-integration/ai-fix-confidence.ts
Jayclifford345: Good catch — confirmed. Both `querySelectorAllEnhanced` and `resolveSelector` never throw: the entire body of `querySelectorAllEnhanced` is wrapped in a try/catch that returns an empty `SelectorResult` (`usedFallback: true`, `effectiveSelector: 'ERROR'`) on any error, and `resolveSelector` returns plain CSS as-is. So the `try/catch` + `if (cssError)` branch was unreachable.

Removed it in 56613b4a. An invalid selector now falls through to the `matchCount === 0` rejection, which preserves the safety property (a bad selector is never returned as `ok`) — we just drop the differentiated reason, which per your other thread isn't worth carrying.

src/integrations/assistant-integration/ai-fix-confidence.test.ts
Jayclifford345: Fixed in 56613b4a. The mock now mirrors the real helper — it try/catches `document.querySelectorAll` and returns `{ elements: [], usedFallback: true, originalSelector, effectiveSelector: 'ERROR' }` instead of throwing.

leslie-heinzen (approved):

PR #996, opened 2026-06-10
feat(ai-fix): add AiFixOrchestrator shell + lazy mount in docs-panel

Jayclifford345:
## What & why

**PR7 of the PR-886 AI auto-heal breakdown** ([RFC](https://github.com/grafana/pathfinder-rfcs/pull/5)). Stacks on PR6 ([#995](https://github.com/grafana/grafana-pathfinder-app/pull/995)); depends on PR4 (apply), PR5 (hook), and PR6 (extracted helpers).

This is the thin React shell that glues the AI-powered "Fix this" affordance to the docs-panel Scene model, plus the docs-panel lazy mount. In PR #886 the orchestrator was a 716-line file with all the DOM-context / confidence / step-content helpers inline; PR6 extracted and unit-tested those ~510 lines, so this shell just imports them.

## What's included

- **`src/components/docs-panel/AiFixOrchestrator.tsx`** — thin shell (3 effects: request listener, apply, error) consuming PR6's helpers, PR5's `useAiFixGeneration`, and PR4's `applyPatchToGuide`. Renders nothing.
- **`docs-panel.tsx`** — `lazy()` import + `handleAiFixPatchApplied` tab-content writer + `<Suspense>` mount as the first child of the `data-pathfinder-content` container.
- **`ai-fix-step-id.ts`** — adds `materializeStepIdsInJson` (a string entry point on PR2's id module) so the shell can feed the assistant + content extraction the same canonical ids a component dispatched. The id-augmented form drives `extractStepContent` (whose `findGuideBlockById` resolves anonymous blocks by their derived id) and the assistant prompt.
- Co-located unit tests for the orchestrator and for `materializeStepIdsInJson`.

### Must-fixes applied (from §5.4 PR7)

- Call `evaluatePatchConfidence(patch)` with **no second arg** and an **accurate** one-line comment (the old code carried a comment claiming a token-overlap "Bar chart → Bar gauge" rejection that never existed; PR6 fixed the helper, this fixes the call site).
- **Dropped** the outer `synthesizeStepIdsInJson(...)` wrap on the apply path — `applyPatchToGuide` materializes ids internally, so the apply effect passes the raw guide string directly.
- **Deleted** the screenshot `NOTE` (defends a v1-disabled feature with no code left).
- **Trimmed** the 22-line file docstring / numbered runbook (moved to the commit body); kept one copy of the `lazy()`/jest-init-crash workaround note.

## It lands dark

- Nothing dispatches `pathfinder-ai-fix-request` yet — the button sites land in PR8. The orchestrator only mounts a listener.
- Even if an event fired, the handler early-returns while `enableAiAutoHeal` is false (default) — defense-in-depth alongside the (future) button-site gate.
- The `lazy()` import keeps `@grafana/assistant` out of the docs-panel init chain.

Verified dark by grep: the only `dispatchEvent('pathfinder-ai-fix-request')` in `src/` is in the orchestrator's own test.

## Test plan

- `npm run typecheck` ✅ · `eslint` ✅ 0 errors · `prettier --check` ✅
- Governance/architecture tests ✅ (no new tier violations; imports flow Tier 3 → Tier 4 downward)
- `jest` ✅ — new `AiFixOrchestrator` suite (7): renders-nothing, flag-off drop, assistant-unavailable drop, generate wiring, confident-patch apply + success toast, low-confidence rejection, error→warning toast; `materializeStepIdsInJson` suite (4): anonymous-block id, author-id preserved, invalid-JSON passthrough, non-guide passthrough
- Full frontend suite ✅ 240/240 suites, 4039 passed

### DIFF
--- a/src/components/docs-panel/AiFixOrchestrator.tsx
+++ b/src/components/docs-panel/AiFixOrchestrator.tsx
@@ -1,0 +1,191 @@
+import { useCallback, useEffect, useMemo, useRef } from 'react';
+
+import { AppEvents, usePluginContext } from '@grafana/data';
+import { getAppEvents } from '@grafana/runtime';
+
+import { getConfigWithDefaults } from '../../constants';
+import { evaluatePatchConfidence } from '../../integrations/assistant-integration/ai-fix-confidence';
+import { collectDomContext, tagFromSelector } from '../../integrations/assistant-integration/ai-fix-dom-context';
+import { AI_FIX_REQUEST_EVENT, type AiFixRequestDetail } from '../../integrations/assistant-integration/ai-fix-event';
+import { materializeStepIdsInJson } from '../../integrations/assistant-integration/ai-fix-step-id';
+import { extractStepContent } from '../../integrations/assistant-integration/ai-fix-step-content';
+import { applyPatchToGuide } from '../../integrations/assistant-integration/apply-ai-fix-patch';
+import { useAiFixGeneration } from '../../integrations/assistant-integration/useAiFixGeneration.hook';
+import { GlobalInteractionBlocker } from '../../interactive-engine';
+import { reportAppInteraction, UserInteraction } from '../../lib/analytics';
+import type { LearningJourneyTab } from '../../types/content-panel.types';
+
+// Safety net: the assistant can resolve without a patch or error, which would wedge the gate forever.
+const AI_FIX_REQUEST_TIMEOUT_MS = 30_000;
+
+function publishToast(severity: 'success' | 'warning', title: string, body?: string) {
+  try {
+    const event = severity === 'success' ? AppEvents.alertSuccess : AppEvents.alertWarning;
+    getAppEvents().publish({ type: event.name, payload: body ? [title, body] : [title] });
+  } catch {
+    // Non-Grafana host (jsdom): a missing toast is harmless.
+  }
+}
+
+interface PendingAiFixRequest {
+  detail: AiFixRequestDetail;
+  tabId: string;
+  guideJson: string;
+}
+
+interface AiFixOrchestratorProps {
+  activeTab: LearningJourneyTab | null;
+  onPatchApplied: (tabId: string, newGuideJson: string) => void;
+}
+
+function AiFixOrchestrator({ activeTab, onPatchApplied }: AiFixOrchestratorProps): null {
+  const contentKey = activeTab?.id ?? 'pathfinder-ai-fix-orchestrator';
+  const { generate, patch, error, reset, isAssistantAvailable } = useAiFixGeneration(contentKey);
+
+  const pluginContext = usePluginContext();
+  const aiAutoHealEnabled = useMemo(
+    () => getConfigWithDefaults(pluginContext?.meta?.jsonData || {}).enableAiAutoHeal,
+    [pluginContext?.meta?.jsonData]
+  );
+
+  const pendingRequestRef = useRef<PendingAiFixRequest | null>(null);
+  const timeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null);
+  // Latest tab via ref so the listener isn't re-bound each render (would drop in-flight requests).
+  const activeTabRef = useRef<LearningJourneyTab | null>(activeTab);
+  useEffect(() => {
+    activeTabRef.current = activeTab;
+  }, [activeTab]);
+
+  const clearPending = useCallback(() => {
+    pendingRequestRef.current = null;
+    if (timeoutRef.current !== null) {
+      clearTimeout(timeoutRef.current);
+      timeoutRef.current = null;
+    }
+    GlobalInteractionBlocker.getInstance().stopAdHocBlocking();
+    reset();
+  }, [reset]);
+
+  useEffect(() => () => clearPending(), [clearPending]);
+
+  useEffect(() => {
+    const handler = async (e: Event) => {
+      const detail = (e as CustomEvent<AiFixRequestDetail>).detail;
+      const tab = activeTabRef.current;
+      if (!tab?.content?.content) {
+        return;
+      }
+      if (!aiAutoHealEnabled) {
+        reportAppInteraction(UserInteraction.AiFixFailed, { reason: 'feature_flag_disabled' });
+        return;
+      }
+      if (!isAssistantAvailable) {
+        reportAppInteraction(UserInteraction.AiFixFailed, { reason: 'assistant_unavailable' });
+        return;
+      }
+      if (pendingRequestRef.current) {
+        return;
+      }
+      pendingRequestRef.current = { detail, tabId: tab.id, guideJson: tab.content.content };
+      reportAppInteraction(UserInteraction.AiFixOffered, {
+        step_id: detail.stepId ?? '',
+        rendered_step_id: detail.renderedStepId ?? '',
+      });
+
+      GlobalInteractionBlocker.getInstance().startAdHocBlocking('Asking Grafana Assistant for a fix…');
+      timeoutRef.current = setTimeout(() => {
+        if (!pendingRequestRef.current) {
+          return;
+        }
+        reportAppInteraction(UserInteraction.AiFixFailed, { step_id: detail.stepId ?? '', reason: 'timeout' });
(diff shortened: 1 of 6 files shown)


src/components/docs-panel/AiFixOrchestrator.tsx
```diff
@@ -0,0 +1,180 @@
...
+    window.addEventListener('pathfinder-ai-fix-request', handler);
+    return () => window.removeEventListener('pathfinder-ai-fix-request', handler);
+  }, [generate, isAssistantAvailable, aiAutoHealEnabled]);
+
+  useEffect(() => {
+    if (!patch) {
+      return;
+    }
+    const tab = activeTabRef.current;
+    const request = pendingRequestRef.current;
```
moxious, line 132: reads activeTabRef.current inside the patch effect, but pendingRequestRef only captures the request detail — not the tab id the request was issued against. If the user switches tabs between dispatching pathfinder-ai-fix-request and the assistant resolving, the patch — generated against tab A's guide — is applied to tab B's content via applyPatchToGuide(tab.content.content, patch) and surfaced under tab.id (line 132). Best case applyPatchToGuide returns {ok:false} and the user gets a confusing "couldn't update this step" toast on the wrong tab; worst case the targeted step id collides and a foreign patch lands silently.

The test suite never exercises a tab switch between the request and the patch arrival, so this slips through.

Suggest: capture tabId at the moment the request is accepted and in the patch/error effects either look up that tab in model.state.tabs or no-op if activeTabRef.current?.id !== request.tabId. 

Add a regression test where re-render swaps activeTab before flipping patch.

src/components/docs-panel/AiFixOrchestrator.tsx
```diff
@@ -0,0 +1,180 @@
...
+
+      GlobalInteractionBlocker.getInstance().startAdHocBlocking('Asking Grafana Assistant for a fix…');
+
+      // Materialize canonical ids so the assistant + extraction see the ids a component dispatched.
+      const augmentedJson = materializeStepIdsInJson(tab.content.content);
+      const domHint = collectDomContext(detail.refTarget ?? '');
+      const failingStepContent = extractStepContent(augmentedJson, detail.stepId ?? '', detail.containerInfo);
+      const failingTag = tagFromSelector(detail.refTarget ?? '');
+
+      await generate({
```
moxious, line 111: If useAiFixGeneration.generate resolves with neither patch nor error (assistant timeouts, network drop after the in-flight promise was awaited inside the hook, etc.), pendingRequestRef.current stays populated and the gate at line 67 (if (pendingRequestRef.current) return;) silently drops every subsequent fix request for the lifetime of the docs panel. There's no timeout, no abort, and no test covering the "generate resolves quietly" path.

Suggestion: drive the gate off isGenerating from the hook instead of an orchestrator-local ref. This option makes the source of truth the hook and removes the need for a manual reset.

moxious (changes requested):
Line-specific comments added; these comments here are cross-cutting:

B1 — Repeated cleanup triplet across effects

- Locations:
  - AiFixOrchestrator.tsx:111-114 — patch-effect "tab/request missing" branch
  - AiFixOrchestrator.tsx:122-125 — low-confidence branch
  - AiFixOrchestrator.tsx:145-148 — apply effect tail (success or fail)
  - AiFixOrchestrator.tsx:159-161 — error effect tail
- Evidence: The triplet pendingRequestRef.current = null; GlobalInteractionBlocker.getInstance().stopAdHocBlocking(); reset(); appears verbatim four times across two effects. Bodies are normalized (no JSX scaffolding, no comments) and identical.
- Disqualifiers checked: Not generated code; not adjacent-but-different-meaning (every call site means "we're done with this in-flight request"). Flagged.
- Suggested action: Extract a clearPending() helper closure inside the component (or accept it from a custom hook later). Three lines → one. Also makes F2's fix (add a timeout-driven clear) trivial.

E5 — Cross-file string-literal contract introduced without a const

- Locations:
  - AiFixOrchestrator.tsx:75, 88 — 'pathfinder-ai-fix-request' listener registration
  - AiFixOrchestrator.test.tsx:55 — dispatcher in tests
- Evidence: The custom event name is a string literal in two files today and the PR description tells us PR8 will dispatch it from new button sites in other files. The same shape (E5) is what bit interactive-action-completed / user-action-detected etc. earlier in the project — a typo at one dispatcher silently drops the event.
- Disqualifiers checked: Not auto-generated; literal is used as a public event-bus contract (the PR's whole architecture); two call sites already exist with more incoming. Flagged.
- Suggested action: Land this PR with export const AI_FIX_REQUEST_EVENT = 'pathfinder-ai-fix-request' (in AiFixOrchestrator.tsx or integrations/assistant-integration/index.ts) and use it at both the addEventListener and the test's dispatch. PR8 then imports the same constant instead of retyping the string. Also opens a single grep target for future audits.

src/components/docs-panel/AiFixOrchestrator.tsx
Jayclifford345: Fixed in `55c0c805`. The orchestrator now captures `{ detail, tabId, guideJson }` at the moment the request is **accepted** (not just `detail`), and the patch/error effects key off that snapshot instead of `activeTabRef.current`.

I went a step beyond the no-op suggestion: because `onPatchApplied` writes **by tab id** (`model.state.tabs.map(t => t.id === tabId ? … : t)`), applying against the captured `guideJson` and writing back to the captured `tabId` lands the patch on **tab A** even after the user switches to tab B — so the user's fix isn't silently dropped, it just lands on the tab it was generated for. `onPatchApplied('tab1', …)` and `applyPatchToGuide(tabA.content, patch)` (not tab B).

src/components/docs-panel/AiFixOrchestrator.tsx
Jayclifford345: Fixed in `55c0c805` — but with a **watchdog timeout** rather than driving the gate off `isGenerating`, because the hook route doesn't actually close this hole:

- `useAiFixGeneration` (PR5) never exposes a way to reset `isGenerating`, and on the exact failure you describe — a quiet resolve where neither `onComplete` nor `onError` fires — `isGenerating` itself stays `true` forever (look at the hook: it's only flipped back inside those two callbacks + the `catch`). So gating on it would wedge the same way, and the only fix would be reaching back into PR5's hook from PR7.

Instead the orchestrator arms a **30s watchdog** (matching `DEFAULT_GUIDED_STEP_TIMEOUT`) when it accepts a request. If neither a patch nor an error arrives, it clears the gate, stops the blocker, `reset()`s, and reports `AiFixFailed { reason: 'timeout' }`. A late patch arriving after the watchdog fires is safely dropped — the patch effect early-returns on a null pending ref. It's a safety net, not a deadline: long enough not to kill a legitimately-slow generation, since killing a valid-but-slow LLM call and then dropping its late patch is worse UX than waiting.

Regression test added — `self-heals the in-flight gate when generation never resolves`: accepts a request, asserts a second request is dropped while in flight, advances fake timers past the watchdog, asserts the `timeout` report fires, then asserts a fresh request goes through (gate cleared).

Jayclifford345:
Thanks for the review — all four addressed 

**B1 — cleanup triplet** → extracted a single `clearPending()` `useCallback` (nulls the pending ref, cancels the watchdog, stops the blocker, `reset()`). All four sites call it now, and it's also what the new watchdog and the unmount cleanup reuse — so the F2 timeout fix dropped in for one line, as you predicted.

**E5 — event-name contract** → extracted `export const AI_FIX_REQUEST_EVENT` + the `AiFixRequestDetail` shape into a new `src/integrations/assistant-integration/ai-fix-event.ts`. The orchestrator (listener) and its test (dispatcher) import the constant instead of retyping the literal — single grep target.

  > **Stack note:** PR8 ([#997](https://github.com/grafana/grafana-pathfinder-app/pull/997)) already creates this exact file (byte-identical) for its button dispatchers. Since #997 stacks on this branch, its next rebase turns that into a no-op add and its button sites import from here — so the constant now originates in PR7 where the listener lives, and PR8 consumes it. I'll handle the #997 rebase as part of the stack.

The two line-level findings (tab-switch race, stuck gate) are answered in-thread with the chosen approach + the new regression tests.

moxious (approved):

PR #997, opened 2026-06-10
feat(ai-fix): render "Fix this" buttons in interactive renderers + requiresDomElement

Jayclifford345:
## What & why

**PR8 of the PR-886 AI auto-heal breakdown** ([RFC](https://github.com/grafana/pathfinder-rfcs/pull/5)) — the final code PR. Stacks on PR7 ([#996](https://github.com/grafana/grafana-pathfinder-app/pull/996)); depends on PR5 (`useIsAssistantAvailable`) and PR7 (the orchestrator + event contract).

PRs 1–7 landed every part of the AI auto-heal flow except the trigger — **nothing dispatched `pathfinder-ai-fix-request` yet**. This PR completes the wire: the three interactive renderers now render an AI-powered "Fix this ✨" button on a failing step that dispatches the event the PR7 orchestrator consumes.

The RFC framed this as a hard re-base, but `main` already forwards the canonical `stepId` from the section, so dropping the original branch's parallel `sourceStepId` prop means **`interactive-section.tsx`, its registry, and the symmetry tripwire are untouched**.

## What's included

**Shared extractions** (replacing what the original copy-pasted across all 5 sites):
- **`useAiFixEnabled()`** (`src/integrations/assistant-integration/use-ai-fix-enabled.ts`) — the `assistant-available && enableAiAutoHeal` gate. Imported from its **deep module path**, not the barrel (the barrel re-exports `@grafana/assistant`, which crashes under jsdom).
- **`dispatchAiFixRequest()` + `<AiFixButton>`** (`src/components/interactive-tutorial/ai-fix-button.tsx`) — fires the `AiFixAccepted` analytics + the `CustomEvent`.
- **`ai-fix-event.ts`** — single source of truth for the event name + `AiFixRequestDetail` shape; PR7's orchestrator now imports it (its inline interface + the two event-name string literals are gone).

**Five button sites:** interactive-step (lazy-scroll runtime + pre-exec requirement), interactive-multi-step (runtime substep), interactive-guided (pre-exec + runtime substep). Substep sites carry `containerInfo {containerId, containerKind, subStepIndex}`; all dispatch the canonical `stepId`.

**`step-checker`:** derives + returns `requiresDomElement` (the type field shipped in PR1 but was never populated) and adds `refTarget` to the recheck effect deps so a runtime selector swap re-runs the requirements check.

**testId disambiguation (must-fix):** adds `guidedAiFixButton`; the two sibling buttons in step/guided get `-lazy`/`-runtime` suffixes so they're distinguishable.

## Follow-up fixes (found during live E2E)

Two runtime bugs surfaced while testing the end-to-end flow; both are fixed here, in the layer that owns the buttons + the step-checker signal:

- **Patch applied but the step didn't recover** (`fix(ai-fix): re-check + recover…`). After the orchestrator applies a patch, the step's id is stable (apply bakes a derived id into the guide), so the step updates *in place* rather than remounting. `checkStep` didn't list `refTarget`/`targetAction` in its deps → the re-check fired but re-ran the **stale** selector, so the `exists-reftarget` requirement never passed; and the runtime `lazyScrollError` persisted across the in-place update. Fix: add `refTarget`/`targetAction` to `checkStep`'s deps (restores the original #886 behaviour) and clear `lazyScrollError`/`postVerifyError` on `refTarget` change.
- **"Fix this" shown on sequentially-blocked steps** (`fix(ai-fix): don't offer "Fix this" on sequentially-blocked steps`). The pre-execution gate keyed on `requiresDomElement && !isEnabled`, but `requiresDomElement` is derived from the *static* `requirements` string and `!isEnabled` is also true for a "complete previous step" block. Fix: add `isEligibleForChecking` to the gate in interactive-step (Site B) + interactive-guided (Site A), matching the sibling Retry button. The runtime sites are unaffected (they need an execution attempt that can't happen while sequentially blocked).

## It lands dark

`useAiFixEnabled` is false unless an admin sets `enableAiAutoHeal` (default false) **and** the assistant is present; the orchestrator independently re-gates. Verified by grep: the only `pathfinder-ai-fix-request` producer in `src/` is `<AiFixButton>` (behind the gate), plus the orchestrator listener.

The banned multistep "auto-heal watcher" effect the RFC flagged (§5.4) **was never merged to `main`** — nothing to delete. A patch-applied content remount resets the multistep failure state, symmetric with guided (no watcher needed).

## Test plan

- `npm run typecheck` ✅ · `eslint` ✅ 0 errors (changed files) · `prettier --check` ✅ · governance/architecture ✅ (no new tier violations)
- Full frontend suite ✅ **243/243 suites, 4090 passed**
- New tests: `ai-fix-button` (4); `use-ai-fix-enabled` (3); `step-checker` `requiresDomElement` (4); **interactive-guided AI-fix gating** (2 — sequentially-blocked → no button, eligible + failing requirement → button shows)

> Note: `npm run check` is currently blocked at `lint:go` by a **pre-existing** staticcheck issue in `pkg/plugin/package_recommendations_test.go` (byte-identical on the base branch; this PR makes zero Go changes). All frontend gates pass.

### DIFF
--- a/src/components/interactive-tutorial/interactive-guided.tsx
+++ b/src/components/interactive-tutorial/interactive-guided.tsx
@@ -757,7 +757,7 @@ export const InteractiveGuided = forwardRef<{ executeStep: () => Promise<boolean
               >
                 {checker.canFixRequirement ? 'Fix this' : 'Check again'}
               </button>
-              {aiFixEnabled && checker.requiresDomElement && !checker.canFixRequirement && (
+              {isEligibleForChecking && aiFixEnabled && checker.requiresDomElement && !checker.canFixRequirement && (
                 <AiFixButton
                   className="interactive-guided-ai-fix-btn"
                   testId={testIds.interactive.guidedAiFixButton(renderedStepId)}
--- a/src/components/interactive-tutorial/interactive-guided.tsx
+++ b/src/components/interactive-tutorial/interactive-guided.tsx
@@ -17,8 +17,11 @@ import { getConfigWithDefaults } from '../../constants';
 import { findButtonByText, querySelectorAllEnhanced } from '../../lib/dom';
 import { GuidedAction } from '../../types/interactive-actions.types';
 import { testIds } from '../../constants/testIds';
+// Deep import (not the barrel): the barrel re-exports @grafana/assistant, which crashes under jsdom.
+import { useAiFixEnabled } from '../../integrations/assistant-integration/use-ai-fix-enabled';
 import { sanitizeDocumentationHTML } from '../../security';
 import { STEP_STATES } from './step-states';
+import { AiFixButton } from './ai-fix-button';
 import { markStepCompleted, resetStep, useStepCompletion } from '../../global-state/completion-store';
 import type { ProgressReason } from '../../global-state/progress-events';
 
@@ -264,6 +267,8 @@ export const InteractiveGuided = forwardRef<{ executeStep: () => Promise<boolean
       onStepComplete, // Pass through for objectives auto-completion
       onComplete, // Pass through for objectives auto-completion
     });
+
+    const aiFixEnabled = useAiFixEnabled();
 
     // Combined completion state: objectives always win
     const isCompletedWithObjectives = storedCompleted || checker.completionReason === 'objectives';
@@ -752,6 +757,18 @@ export const InteractiveGuided = forwardRef<{ executeStep: () => Promise<boolean
               >
                 {checker.canFixRequirement ? 'Fix this' : 'Check again'}
               </button>
+              {aiFixEnabled && checker.requiresDomElement && !checker.canFixRequirement && (
+                <AiFixButton
+                  className="interactive-guided-ai-fix-btn"
+                  testId={testIds.interactive.guidedAiFixButton(renderedStepId)}
+                  detail={{
+                    stepId: stepId ?? renderedStepId,
+                    renderedStepId,
+                    refTarget: firstActionRefTarget,
+                    action: firstActionTargetAction,
+                  }}
+                />
+              )}
             </div>
           )}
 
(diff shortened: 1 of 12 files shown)


src/components/interactive-tutorial/interactive-guided.tsx
```diff
@@ -837,6 +854,23 @@ export const InteractiveGuided = forwardRef<{ executeStep: () => Promise<boolean
               >
                 ↻ Try again
               </Button>
+              {aiFixEnabled && currentStepIndex >= 0 && (
```
leslie-heinzen, line 857: Hm, won't `currentStepIndex >= 0` always be true since it initializes to 0?

src/components/interactive-tutorial/interactive-guided.tsx
```diff
@@ -837,6 +854,23 @@ export const InteractiveGuided = forwardRef<{ executeStep: () => Promise<boolean
...
+                  testId={testIds.interactive.guidedAiFixButton(`${renderedStepId}-runtime`)}
+                  detail={{
+                    stepId: stepId ?? renderedStepId,
+                    renderedStepId,
+                    refTarget: internalActions[currentStepIndex]?.refTarget,
+                    action: internalActions[currentStepIndex]?.targetAction,
+                    containerInfo: {
+                      containerId: stepId ?? renderedStepId,
+                      containerKind: 'guided',
+                      subStepIndex: currentStepIndex,
```
leslie-heinzen, line 869: I notice that interactive-multi-step uses a separate `failedStepIndex` property in the equivalent code.

Can you explain the difference? Why is `currentStepIndex` okay here but we track the failed index independently in the other component?

leslie-heinzen (approved):
Couple of questions but lgtm 🚢

src/components/interactive-tutorial/interactive-guided.tsx
Jayclifford345: Ah you're right — it's always true. 

Fixed in the stacked PR #998: guided now tracks its own `failedStepIndex`, so the guard is meaningful again. Leaving this PR as-is to not reset the merge permission.

src/components/interactive-tutorial/interactive-guided.tsx
Jayclifford345: Ah, correct and a bug ... Multi-step is the correct pattern — guided was the odd one out. Multi-step runs all substeps in one batch and resets `currentActionIndex` to `-1` in its `finally`, so it keeps a separate `failedStepIndex` to remember which substep failed for the error UI.

Guided's `finally` also resets `currentStepIndex` to `0`, so reusing it meant the error title *and* the AiFix `detail` (`refTarget`/`action`/`subStepIndex`) always pointed at substep `0`, not the one that actually failed .

Mirrored multi-step in #998: added `failedStepIndex` (init `-1`, reset at execution start, set in the timeout branch), now used for both the title and the AiFix payload.

PR #1008, opened 2026-06-11
refactor(mcp): in-memory authoring sessions (replace GCS backend)

tomglenn:
## What

Simplifies the `gcs-storage` authoring-session backend to an **in-memory store with a sliding TTL**, per the discussion on whether the MCP needs a durable external store. Net **−2,500 lines** and one fewer runtime dependency.

> Based on `gcs-storage` — this PR targets that branch so the diff is just the simplification delta, not the full session work.

## Why

The authoring MCP is low-QPS, human-paced, and authoring sessions are short-lived drafts. Running it as a single Cloud Run instance means a process-local map is sufficient; GCS's durability and horizontal-scale guarantees were more than the workload needs.

## Changed

**Removed (~2,500 lines):** `session-store-gcs.ts`, `lib/gcs/*`, its test, `smoke-gcs-sessions.ts`, the `@google-cloud/storage` dependency, and the bucket / service-account / IAM / lifecycle bootstrap in `deploy-mcp.sh`.

**Kept unchanged:** the `SessionStore` interface, session tokens, `Mcp-Session-Id` pinning, generation-based optimistic concurrency, two-mode (`{artifact}` / `{sessionToken}`) tool input, and the fine-grained read tools.

**Added:** a sliding TTL on `InMemorySessionStore` (default 24h, tunable via `PATHFINDER_SESSION_TTL_HOURS`) replacing the bucket lifecycle rule. Lazy eviction on access + sweep on write — no background timer.

## Deployment trade-off (the load-bearing decision)

In-memory sessions are **process-local**, so the service **must run as a single instance**. `deploy-mcp.sh` pins `--min/--max-instances=1`, and the invariant is surfaced in the deploy output and the runtime startup log. A redeploy or instance recycle drops in-flight sessions — acceptable for short-lived drafts ("start over"). The `{artifact}` stateless mode remains the path for any multi-instance / OSS deploy.

Also hardens the deploy: the service now runs under a **dedicated zero-permission service account** instead of the default compute SA, so a compromise of the public unauthenticated endpoint inherits no project permissions.

## Verification

- `tsc --noEmit` clean; MCP suite **226 passing**; eslint + prettier clean; `bash -n deploy-mcp.sh` clean.
- Lockfile change confined to the removed dependency's tree.


moxious (changes requested):
Fix in this PR (both agents agree this is the strongest finding)

1. Stale "7-day" references — user-facing. src/cli/mcp/tools/result.ts:111 tells the agent "the session expired (7-day TTL)"; src/cli/mcp/tools/finalize.ts:266-269 comments reference the GCS lifecycle rule. The real default is 24h sliding. After this PR, the error message actively misdirects callers. Either substitute the configured TTL at runtime (export DEFAULT_SESSION_TTL_HOURS from session-store.ts and interpolate) or hardcode the new value. Cheap, blocks no architecture, prevents a real foot-gun.

2. bindMcpSessionId throws SessionPreconditionFailedError(0, 1) for a pin collision (session-store.ts:213). Synthetic generation numbers for a non-generation event — a debugger reading expected=0/actual=1 will be misled. Inherited from gcs-storage, not introduced here, but a SessionPinConflictError (or a plain Error with a distinct code) is two lines and removes the lie. (Both agents flagged this independently.)

3. Single-instance invariant has no runtime guard. deploy-mcp.sh pins --min/--max-instances=1 and the post-deploy banner warns, but a later gcloud run services update --max-instances=N silently scatters sessions across replicas → SESSION_NOT_FOUND mid-authoring with no signal. Suggestion: an MCP_SERVER.md note that operator drift on this knob is a contract violation. Not blocking — deploy-mcp.sh is the only sanctioned path today.

4. Zero observability on store cardinality / eviction. sweepExpired runs only on save, drops entries silently. No counter, no log line, no stats() on the interface — on a single Cloud Run instance where memory is the only bound, this is the highest-leverage gap in the new design. size() exists but is marked test-only. Add an evictions counter and emit {liveSessions, evictedThisSweep} on each sweep (piggyback on access-log.ts). Cheap; catches a leak before an OOM.

5. TTL magic number scattered across session-store.ts:120 (ms), session-store-factory.ts:14 (hours→ms re-derivation), deploy-mcp.sh:20 (comment), .env.example, and the stale error message in #1. Single source of truth in session-store.ts (DEFAULT_SESSION_TTL_HOURS = 24, derive ms from it). Naturally addressed if #1 is fixed properly.

6. Factory test uses jest.spyOn(Date, 'test.ts:54-83) instead of an injectedclock — the rest of the suite cleanly uses the constructor's now option. Two-line change: thread now through getDefaultSessionStore({ now }). Minor

7. Sweep-on-save asymmetry. Documented g traffic, but a long read-only burst
against many expired tokens won't reclae lands. Worth keeping in mind if the TTL
 ever climbs or a "list active sessions

8. Oversized JSDoc preambles on session-store.ts (32 lines), state-bridge.ts (42 lines), session-token.ts (24 lines). Per the repo's QC8 catalog, these are the headers most likely to rot. Useful today because the design is fresh and they cover load-bearing rationale, but several blocks duplicate docs/design/HOSTED-AUTHORING-MCP.md. Tr

tomglenn:
Thanks @moxious — addressed all 8 in `db1902e6`.

**Fixed**
1. **Stale "7-day" strings** — `result.ts` SESSION_NOT_FOUND message and the `finalize.ts` delete-failure warn no longer cite the old GCS lifecycle. Went **duration-free** rather than hardcoding "24h": the TTL is env-configurable *and* sliding, so any fixed number would itself drift the moment a session is touched or an operator sets `PATHFINDER_SESSION_TTL_HOURS`.
2. **Pin collision** — new `SessionPinConflictError` (`code: SESSION_PIN_CONFLICT`); `bindMcpSessionId` throws it instead of `SessionPreconditionFailedError(0, 1)`. Raw token kept off `.message` (carried on a field) per the `tokenLogPrefix` discipline. Nothing branched on the old throw, so no downstream change.
4. **Observability** — added an `evictions` counter + `stats(): { liveSessions, evictions }` on `InMemorySessionStore`, kept **off the `SessionStore` interface** (concrete-only, mirroring `size()` — observability is an in-memory-backend concern, not the tool-facing contract). The HTTP access log now emits `liveSessions`/`evictions` **on `tools/call` only** (off the high-frequency health/`tools/list` lines). A climbing `liveSessions` with flat `evictions` is the leak signal.
5. **TTL single source of truth** — `DEFAULT_SESSION_TTL_HOURS` + `MS_PER_HOUR` in `session-store.ts`; factory reuses `MS_PER_HOUR`.
6. **Factory test** — threads `now` through `getDefaultSessionStore({ now })`; dropped the `jest.spyOn(Date, 'now')`.
8. **JSDoc** — trimmed conservatively: dropped the rot-prone "Tracked in P3 deviations" process reference and the deployment paragraph that duplicated the design doc; kept the load-bearing tmpdir-deviation rationale.

**Doc-only**
3. **Single-instance guard** — added an MCP_SERVER.md note that the `--min/--max-instances=1` pin is a *correctness contract*, not a tuning knob: raising `--max-instances` scatters sessions and yields `SESSION_NOT_FOUND` mid-authoring with no deploy-time signal. `deploy-mcp.sh` remains the only sanctioned path.

**Acknowledged, no change**
7. **Sweep-on-save asymmetry** — left as a documented tradeoff for this low-QPS workload. `stats()` sweeps on read, so an accurate live count is available on demand even during a read-only burst; worth revisiting only if the TTL climbs or a "list active sessions" surface lands.

New tests cover the pin-conflict throw + idempotent rebind, `stats()` live-count and cumulative/lazy evictions, and an access-log assertion that `tools/call` carries the stats while `tools/list` omits them. `npm run check`-relevant gates green (tsc, 232 MCP tests, eslint, prettier).

moxious (approved):

Issue #1007, opened 2026-06-11
feat: clicking pathfinder in popout mode closes grafana models


Jayclifford345:
This is less of a bug and more of a feature. It is standard across the grafana interface that a modal will shut when you interact with another DOM element other than the DOM container itself. This is problematic for pathfinder when someone wants to scroll to the new instruction but this closes the modal. See video

https://github.com/user-attachments/assets/f7d3cf2d-0416-4978-9e3f-8f6ba53f3311

There are potentially options we could explore in grafana/grafana to prevent this behaviour when Pathfinder is involved but needs further exploration

moxious:
Category: design decision needed (root fix likely lives in grafana/grafana, not this repo)

Implementation basis & cause

- Popout is portaled twice: once to <div id="pathfinder-floating-root"> (src/module.tsx:207-209), and the panel itself uses createPortal(panelContent, document.body) at src/components/floating-panel/FloatingPanel.tsx:268. Definitively outside Grafana's modal portal tree.
- No outside-click suppression. Only stopPropagation in floating-panel is for drag-resize: src/components/floating-panel/useDragResize.ts:131. No capture-phase listener marking the popout as "inside" for Grafana's modal logic.
- The data-pathfinder-content="true" attribute exists (FloatingPanel.tsx:199) but only Pathfinder consumes it (src/global-state/utils.link-interception.ts:92). Grafana's modal doesn't know about it.

Cause: Grafana modals dismiss on outside-click via DOM ancestry. Because Pathfinder's popout is a document.body sibling, every click in it is "outside" the modal. Z-index doesn't affect this — only ancestry does.

Design decision needed

Three locations for the fix:
- Upstream (grafana/grafana): teach the modal's outside-click detector to ignore clicks inside [data-pathfinder-content]. Issue author already flagged this option.
- In-plugin capture-phase suppression: attach a pointerdown capture listener that stopPropagations when a Grafana modal is open. Risks Grafana focus/blur and a11y semantics.
- Re-parent the popout into Grafana's modal portal container. Conflicts with the current document.body design shared with #1003.

Shares root architecture with #1003 — both decisions could be made together.

PR #1009, opened 2026-06-11
feat(companion): add inert companion section flag

Jayclifford345:
Adds an authored `companion?: boolean` flag to section blocks, threaded
through type -> zod schema (+ KNOWN_FIELDS canary) -> parser -> component
props -> content renderer -> block-editor section form (+ test-id).

Inert: nothing consumes it yet. The runtime controller (auto pop-out beside
native modals, dock-back, modal-aware positioning) lands in later layers of
this stack. First layer of the companion-mode stack.

### DIFF
--- a/src/components/block-editor/forms/SectionBlockForm.tsx
+++ b/src/components/block-editor/forms/SectionBlockForm.tsx
@@ -66,6 +66,7 @@ export function SectionBlockForm({
   const [requirements, setRequirements] = useState(initial?.requirements?.join(', ') ?? '');
   const [objectives, setObjectives] = useState(initial?.objectives?.join(', ') ?? '');
   const [autoCollapse, setAutoCollapse] = useState(initial?.autoCollapse ?? true);
+  const [companion, setCompanion] = useState(initial?.companion ?? false);
 
   // Preserve nested blocks when editing (but don't display them in the form)
   const nestedBlocks = useRef<JsonBlock[]>(initial?.blocks ?? []);
@@ -90,8 +91,9 @@ export function SectionBlockForm({
       ...(reqArray.length > 0 && { requirements: reqArray }),
       ...(objArray.length > 0 && { objectives: objArray }),
       ...(autoCollapse === false && { autoCollapse: false }),
+      ...(companion === true && { companion: true }),
     };
-  }, [sectionId, title, requirements, objectives, autoCollapse]);
+  }, [sectionId, title, requirements, objectives, autoCollapse, companion]);
 
   const handleSubmit = useCallback(
     (e: React.FormEvent) => {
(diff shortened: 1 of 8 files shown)


Jayclifford345:
**Update — gating clarified.** This layer (the inert `companion` section flag) is version-independent and can merge **without** waiting for the upstream fix. Only #1012 is gated on grafana/grafana#126261. Part of the companion-mode stack addressing #1007.

moxious:
## Review: inert companion section flag

_Automated review (companion-mode stack, layer 1 of 4). Classification: contracts-and-schemas + product-runtime. Coverage confidence: high._

**Summary:** Clean, minimal, end-to-end plumbing of an inert `companion?: boolean` section flag. Threaded consistently through all layers, the KNOWN_FIELDS canary is satisfied, tests cover both present and absent parser paths, and inertness is verified. No blocking or should-fix issues.

### Blocking
None.

### Should-fix
None.

### Nit
- `src/components/content-renderer/content-renderer.tsx:1056` — `companion={element.props.companion}` is passed to `<InteractiveSection>`, but `InteractiveSection` (`src/components/interactive-tutorial/interactive-section.tsx:132-145`) does not destructure or consume `companion`, so the prop is silently dropped at runtime. This is correct and intentional for a deliberately-inert flag (the consumer lands in #1010+); the type only exists so passing it typechecks. Flagging only so it's a conscious staging decision, not accidental dead wiring — no action needed for this PR.

### Praise
- The **KNOWN_FIELDS canary** is handled correctly: `companion` was added to both `SectionProps` (`json-guide.schema.ts:779`) and `KNOWN_FIELDS.section` (`json-guide.schema.ts:1044`). The `type-coupling.test.ts:89-93` sync test asserts schema keys equal the KNOWN_FIELDS set exactly, so omitting either side would have failed CI. Right contract discipline.
- `SectionProps` is spread into all three recursive section schema definitions (`json-guide.schema.ts:848,882,909`), so the flag propagates to every nesting level with a single addition.
- Schema field is correctly `z.boolean().optional()` with a `.describe()` matching sibling `autoCollapse`; the parser (`json-parser.ts:538`) passes `block.companion` straight through with no coercion — faithful round-trip. The two new parser tests cover both `companion: true` and unset→`undefined`.

### Notes on focus questions
- **Sentence case:** "Companion mode" and its description are correctly sentence-cased. No violation.
- **i18n:** the new `Field` label/description are hardcoded English, but this is the established, un-linted pattern for the entire `SectionBlockForm` (Section ID, Requirements, Objectives, Auto-collapse are all hardcoded). No `@grafana/i18n` usage or `no-untranslated-strings` rule on this authoring UI, so no i18n handling is expected. Consistent.
- **QC8:** the `{/* Companion mode */}` marker is technically shape-1, but matches the pre-existing house style in this file (`{/* Section ID */}`, etc.). Not worth flagging in isolation.
- **Inertness confirmed:** the only would-be consumer (`InteractiveSection`) ignores the prop; nothing else reads `companion`.
- **Reversibility:** fully reversible — old guides omit the field → `undefined`; a revert cleanly removes an unconsumed flag with no persisted-state/telemetry footprint.

**Disposition: Approve.**

Jayclifford345:
Closing: with grafana/grafana#126261 shipped in **Grafana v13.1.0**, we've trimmed the companion-mode stack to its two bug-fix layers (resubmitted as #1283). The authored `companion` section flag is **deferred pending real user demand** — it would permanently commit the guide schema for an auto-pop-out UX we haven't validated, and every guide with a modal step would need re-authoring. The `companion/authoring` branch remains if we revive this.

PR #1010, opened 2026-06-11
feat(companion): render floating panel into getPortalContainer + exclude from auto-detect/highlight-clear

Jayclifford345:
Render the floating panel via getPortalContainer() instead of document.body, so it
becomes a child of Grafana's portal container — which native modals exempt from
their aria-hidden pass (and, once grafana/grafana#126261 lands, from outside-press
dismiss). Falls back to document.body when the container is absent (e.g. tests).

Also exclude [data-pathfinder-content="true"] from the devtools auto-detect filter
(shouldCaptureElement) and the highlight click-outside handler, so panel clicks
don't record spurious interactions or clear an active highlight.

Second layer of the companion-mode stack.


Jayclifford345:
**Update — gating clarified.** Renders the floating panel into `getPortalContainer()` — version-independent and safe on any Grafana (it's the basis for the #1007 dismiss fix once grafana/grafana#126261 ships). Can merge **without** waiting; only #1012 is gated. Part of the #1007 work.

moxious:
## Review: render floating panel into `getPortalContainer()` + exclude from auto-detect/highlight-clear

_Automated review (companion-mode stack, layer 2 of 4)._

Solid, well-scoped layer. The core mechanism is correct and the exclusion changes are consistent with an existing repo-wide convention. Two things worth attention (both non-blocking): a stacking-context behavior change inherent to the portal move, and a coverage gap for the minimized pill.

### Blocking
None.

### Should-fix

**1. Minimized pill is not covered by the new exclusion marker** — `src/components/floating-panel/MinimizedPill.tsx:68-87`
The pill is rendered as a *sibling* of the marked panel div (`FloatingPanel.tsx:184-186`), not a child, and its wrapper carries no `data-pathfinder-content`. When the panel is minimized (marked div `display:none`, pill visible), clicking the pill to drag or restore it will:
- pass `shouldCaptureElement` (`action-detector.ts`) → recorded as a spurious interaction while record mode / action-monitor is active, and
- fall through the highlight `clickOutsideHandler` (`navigation-manager.ts:456-463`) → clears an active highlight.

This is exactly the class of bug this PR sets out to prevent, just for the minimized state. Add `data-pathfinder-content="true"` to the pill wrapper div (`MinimizedPill.tsx:69`). _(correctness-and-reliability)_

**2. Stacking context change is real and worth a note** — `FloatingPanel.tsx:268` + `floating-panel.styles.ts:18`
`grafana-portal-container` is `position: fixed; z-index: 1061` (`theme.zIndex.portal`). A `position: fixed` child inside a positioned+z-indexed parent is scoped to that parent's stacking context, so the panel's `zIndex: 9990` is no longer global — effective stacking vs. the rest of the page is governed by 1061, not 9990. This is the intended tradeoff for coexisting with native modals (which also live in the container), and most Grafana overlays go through the portal so they still order correctly. But any non-portaled `document.body` content with a z-index between 1061 and 9990 could now paint over the panel. Low-risk, but please confirm during manual verification that dropdowns/tooltips/toasts still render correctly relative to the panel. _(correctness-and-reliability / cross-cutting-architecture; easy revert)_

### Nit

**3. Stale comment in MinimizedPill (trim-on-touch, QC8)** — `MinimizedPill.tsx:66-67`
`// No <Portal> needed — the parent (FloatingPanel) already renders via createPortal into document.body` is now inaccurate (it renders into the portal container). Not in the diff, so optional — but if you touch the pill for finding #1, drop or correct this line.

**4. `getPortalContainer()` fallback is real but not authored here** — `FloatingPanel.tsx:268`
The summary describes "falls back to document.body when the container is absent." Confirmed accurate, but the fallback lives *inside* `getPortalContainer()` (`@grafana/ui` does `getElementById('grafana-portal-container') ?? document.body`, return type `HTMLElement`, never null). The bare call is correct and needs no local `|| document.body`. Flagging only because the summary could read as if the PR implements the fallback — it doesn't, and doesn't need to.

### Praise
- **`data-pathfinder-content="true"` is the right vehicle.** Already a convention on the docs-panel root (`docs-panel.tsx:1207`), full-screen layout (`FullScreenLayout.tsx:80`), link interception, and assistant integration, and allowlisted in the sanitizer (`html-sanitizer.ts:248`). The floating panel was the outlier; this aligns it.
- **Auto-detect exclusion is complete via a single choke point.** Both `action-monitor.ts:298` and `action-recorder.hook.ts:450` route through `shouldCaptureElement`, so editing that one function covers all recording paths. Excluding the panel doesn't hide legitimate recording — it's Pathfinder's own chrome, never user-under-test surface.
- **Drag/resize/dodge unaffected.** All use viewport coords (`window.innerWidth`, `getBoundingClientRect`) with `position: fixed`, and `useHighlightDodge` observes the `document.body` subtree (portal container is a body descendant), so none depend on being a direct child of body.
- **`.gitignore` entry is appropriate** — `/.playwright-mcp/` is a local Playwright-MCP scratch dir, correctly anchored to repo root.

### Tests
No test added for the portal-target change; the `getPortalContainer()` fallback makes jsdom tests pass regardless (container absent → `document.body`), so the container-present case is unverified. Acceptable for this dormant layer, but companion-mode coexistence (panel + modal in the same container) will need an integration/E2E assertion by the time #1011/#1012 activate it.

No high-confidence tech-debt patterns in the changed files.

Jayclifford345:
Closing: the upstream prerequisite (grafana/grafana#126261) shipped in **Grafana v13.1.0**, so we've simplified the companion-mode stack. This layer's change (render into `getPortalContainer()` + auto-detect/highlight exclusions) is resubmitted unchanged as part of the combined PR #1283.

Issue #1003, opened 2026-06-11
Bug: Pop out pathfinder z-axis hides modals


Jayclifford345:
### Instance

All

### Page URL

Any

### Steps to record

<img width="2160" height="1398" alt="Image" src="https://github.com/user-attachments/assets/40abb810-cc6e-458f-ba2b-6dafd6717f7c" />

Pathfinder in popout mode uses a 9999 z-axis I believe, which is above all other components in Grafana's dom. Currently by design, but I need to think of a way to prevent Pathfinder from hiding critical content underneath it, for example, configuration modals.

### Observed behavior

See steps to record

### Interactive JSON

```json

```

### Console logs

```shell

```

### Severity

Low - Minor inconvenience

moxious:
Category: design decision needed

Implementation basis & cause

- Popout = "floating" mode. PanelMode defined at src/global-state/panel-mode.ts:6.
- Z-index: FLOATING_PANEL: 9990 at src/constants/interactive-z-index.ts:12, applied at src/components/floating-panel/floating-panel.styles.ts:18.
- Mount point: <div id="pathfinder-floating-root"> is document.body.appendChild'd at src/module.tsx:197-216. Renders into its own React root, sibling of Grafana's app root.
- Grafana's own theme z-indices cap at modal: 1060, portal: 1061 (@grafana/data zIndex tokens) — so any Grafana modal renders ~8000 stacking layers below Pathfinder.
- Already exists: useHighlightDodge (src/components/floating-panel/useHighlightDodge.ts) animates the popout away from interactive highlights; no equivalent for modals.

Design decision needed

The high z-index is intentional — interactive guides need to stay above Grafana chrome so highlights work. Tradeoffs (not a recommendation, just the option space):
- Detect open Grafana modals (e.g., MutationObserver on [role="dialog"]) and dim/minimize/lower the popout while a modal is open.
- Extend useHighlightDodge to also dodge modal rects.
- Lower FLOATING_PANEL below theme.zIndex.modal=1060 and accept that guides highlighting elements inside modals will need per-step boosts.
- Auto-collapse to MinimizedPill on modal open.

Whichever path is chosen interacts with global-interaction-blocker.ts overlays (9999/10000/10001/10002) when a guide is mid-step.

PR #1011, opened 2026-06-11
feat(companion): add modal-watcher + extend useHighlightDodge to dodge native modals

Jayclifford345:
Add src/interactive-engine/modal-watcher.ts: detectModalActive() (shared modal
detection that excludes Pathfinder's own floating panel, which is a role=dialog)
plus refcounted startModalWatch/stopModalWatch that emit pathfinder-modal-state-changed.
GlobalInteractionBlocker.isModalActive now delegates to detectModalActive (single
source; also fixes the panel self-trigger).

Extend useHighlightDodge with a dodgeModal flag: when set, the largest visible
native modal rect is folded into the dodge union and a re-check runs on
pathfinder-modal-state-changed, so the floating panel repositions beside the
modal (or compacts when no corner is clear). Dormant until a consumer passes
dodgeModal / emits pathfinder-companion-armed (the controller, next layer).

Third layer of the companion-mode stack.

### DIFF
--- a/src/components/floating-panel/FloatingPanel.tsx
+++ b/src/components/floating-panel/FloatingPanel.tsx
@@ -3,6 +3,7 @@ import { createPortal } from 'react-dom';
 import { IconButton, useStyles2, getPortalContainer } from '@grafana/ui';
 import { reportAppInteraction, UserInteraction } from '../../lib/analytics';
 import { buildPathfinderShareUrl } from '../../utils/pathfinder-search-params';
+import { startModalWatch, stopModalWatch } from '../../interactive-engine';
 import { getFloatingPanelStyles } from './floating-panel.styles';
 import { useDragResize } from './useDragResize';
 import { useHighlightDodge } from './useHighlightDodge';
@@ -61,8 +62,13 @@ export function FloatingPanel({
   const [isDodging, setIsDodging] = useState(false);
   const { geometry, setPosition, drag, resize } = useDragResize();
 
-  // Auto-reposition when interactive highlights overlap the panel
-  useHighlightDodge(geometry, panelState === 'minimized');
+  // Auto-reposition to dodge interactive highlights and any open native modal
+  useHighlightDodge(geometry, panelState === 'minimized', true);
+
+  useEffect(() => {
+    startModalWatch();
+    return () => stopModalWatch();
+  }, []);
 
   const handleMinimize = useCallback(() => {
     setPanelState('minimized');
--- a/src/components/floating-panel/useHighlightDodge.test.ts
+++ b/src/components/floating-panel/useHighlightDodge.test.ts
@@ -1,0 +1,87 @@
+import { renderHook } from '@testing-library/react';
+
+import { useHighlightDodge } from './useHighlightDodge';
+
+const GEOMETRY = { x: 50, y: 50, width: 400, height: 400 };
+
+function mockRect(el: HTMLElement, left: number, top: number, width: number, height: number) {
+  el.getBoundingClientRect = () =>
+    ({
+      left,
+      top,
+      right: left + width,
+      bottom: top + height,
+      width,
+      height,
+      x: left,
+      y: top,
+      toJSON: () => ({}),
+    }) as DOMRect;
+}
+
+function addModal(left: number, top: number, width: number, height: number, attrs: Record<string, string> = {}) {
+  const el = document.createElement('div');
+  el.setAttribute('role', 'dialog');
+  el.style.opacity = '1';
+  Object.entries(attrs).forEach(([k, v]) => el.setAttribute(k, v));
+  document.body.appendChild(el);
+  mockRect(el, left, top, width, height);
+  return el;
+}
+
+function captureEvents() {
+  const events: string[] = [];
+  const dodge = () => events.push('dodge');
+  const compact = () => events.push('compact');
+  document.addEventListener('pathfinder-floating-dodge', dodge);
+  document.addEventListener('pathfinder-floating-compact', compact);
+  return {
+    events,
+    cleanup: () => {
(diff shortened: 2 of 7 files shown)


Jayclifford345:
**Update — now also fixes #1003.** Beyond the modal-watcher, the floating panel now **dodges any open native modal in all popout** (repositions beside it instead of covering it). That addresses #1003, is **version-independent**, and can merge **without** waiting for grafana/grafana#126261. Only #1012 is gated. Also part of the #1007 work.

moxious:
## Review: modal-watcher + `useHighlightDodge` dodgeModal

_Automated review (companion-mode stack, layer 3 of 4)._

Solid, well-factored layer with genuinely good test coverage. The refcounting, listener symmetry, and observer teardown are correct. My one real concern is that the "dormant" claim in the PR description does not match the code: this layer ships **active** behavior for floating-panel-mode users, and it wires up an over-broad modal detector. Nothing here is irreversible (a pure revert restores prior behavior), so these are should-fix-before-merge, not blockers to the stack's design.

### Blocking
None.

### Should-fix

- **"Dormant" claim is inaccurate; behavior is live for floating-mode users.** `FloatingPanel.tsx:66` passes `useHighlightDodge(geometry, ..., true)` and `FloatingPanel.tsx:69-70` calls `startModalWatch()` on mount unconditionally. There is no `pathfinder-companion-armed` gate or feature flag anywhere (grep for `pathfinder-companion-armed` / `dodgeModal` in `src/` returns nothing outside this file). The PR body says "Dormant until a consumer passes dodgeModal / emits pathfinder-companion-armed" — but this PR *is* that consumer. Scope is narrower than "all users": `FloatingPanelManager.tsx` returns `null` unless `mode === 'floating'` (opt-in, localStorage-persisted, default `sidebar`), so impact is limited to users in floating panel mode. For those users this is a real, immediate behavior change: opening any native modal while the panel is open now repositions/compacts it, and an always-on `MutationObserver` + 1s `setInterval` poll runs the whole time the panel is mounted. Recommend either gating `dodgeModal`/`startModalWatch` behind the same arming signal the controller (#1012) will emit, or updating the description to state plainly that floating-mode dodging goes live in this PR. _(correctness-and-reliability / cross-cutting-architecture; fully reversible)_

- **`detectModalActive()` selector is over-broad for Grafana overlays** — `modal-watcher.ts:14` matches `[role="dialog"], [aria-modal="true"], [data-overlay-container="true"]`. In Grafana these attributes are emitted by many *non-modal* overlays via react-aria — notably `TimeRangePicker`'s calendar (`useDialog` → `role="dialog"`, mounted inside an `OverlayContainer` → `data-overlay-container="true"`), plus other pickers/popovers. So the watcher emits `pathfinder-modal-state-changed` and drives panel re-checks (and, for `role="dialog"` matches, actual dodging via `largestVisibleModalRect` at `useHighlightDodge.ts:80-88`) for ordinary time pickers and dropdowns, not just true dialogs. The old passive `isModalActive` had the same selector, but it was only read on-demand by the interaction blocker; this PR turns it into an always-on driver of panel movement, so the imprecision now has user-visible consequences. Consider tightening to `[aria-modal="true"]` or excluding known picker/overlay signatures.

### Nit

- **Coalescing vs. polling redundancy** (`modal-watcher.ts:38-48, 71`) — `scheduleCheck`'s `setTimeout(0)` coalescing is driven by the MutationObserver, and there's *also* a 1s `setInterval(check)`. The poll is a reasonable safety net for changes the observer's `attributeFilter` misses, but worth a one-line comment saying why both exist, or dropping the poll if the observer is sufficient.
- **`largestVisibleModalRect` doesn't honor `data-overlay-container` the same way `detectModalActive` does** (`useHighlightDodge.ts`) — the dodge-geometry function queries only `[role="dialog"], [aria-modal="true"]`, while the watcher's trigger uses the broader set. A `data-overlay-container`-only overlay can *fire* a re-check that finds no dodge rect — harmless, but a slight asymmetry.

### Praise
- **Refcounting is correct.** `startModalWatch`/`stopModalWatch` (`modal-watcher.ts:50-89`) balance start/stop, only create the observer+timer on the 0→1 transition and tear down on 1→0, and guard against underflow. No leak on unmount race, concurrent watchers handled (directly tested). `lastState` is re-seeded on each fresh `startModalWatch`, so no stale-state bug across teardown/restart.
- **Listener symmetry in `useHighlightDodge`.** The `pathfinder-modal-state-changed` handler is added/removed within the same effect, the observer is disconnected on cleanup, and `dodgeModal` is read from a ref so stable deps `[isMinimized]` are preserved — no stale closure, no missing deps (R1/R2/R13 clean).
- **The panel self-trigger fix is real and correct.** Excluding `.closest('[data-pathfinder-content="true"]')` drops Pathfinder's own `role="dialog"` panel from modal detection; the panel portals into `#grafana-portal-container` (not inside `[data-overlay-container]`), so the exclusion is robust even against nested dialogs (tested).
- **Good test coverage.** `useHighlightDodge.test.ts` covers dodge, compact (full-viewport modal), panel self-exclusion, and the `dodgeModal=false` regression guard. `modal-watcher.test.ts` covers visibility filtering, nested-dialog exclusion, open/close emission, and refcounting.

No high-confidence tech-debt patterns in the changed files.

**Bottom line:** No blockers. The two should-fix items (inaccurate "dormant" framing + over-broad modal selector) are worth resolving before merge because they change behavior for floating-mode users today, but both are fully reversible and the underlying implementation quality is high.

Jayclifford345:
Closing: the companion-mode stack was simplified now that grafana/grafana#126261 shipped in **Grafana v13.1.0**. This layer's change (modal-watcher + `dodgeModal`, fixes #1003) is resubmitted unchanged as part of the combined PR #1283.

PR #1012, opened 2026-06-11
feat(companion): add companion-mode controller + version gate (dormant)

### LINKED ISSUE #1003: Bug: Pop out pathfinder z-axis hides modals

### LINKED ISSUE #1007: feat: clicking pathfinder in popout mode closes grafana models

Jayclifford345:
Add useCompanionMode: while an authored companion section is the live (expanded,
incomplete) one, pop the panel out to floating and start the modal-watch, emit
pathfinder-companion-armed, and dock back / disarm on completion or unmount (only
docking what it popped out). FloatingPanel consumes pathfinder-companion-armed to
drive modal-aware dodging (useHighlightDodge dodgeModal). modal-watcher is consumed
via the interactive-engine barrel.

Gated by isModalCoexistenceSupported() (src/lib/grafana-version.ts), held at a
sentinel min version until grafana/grafana#126261 (modal portal-container dismiss
fix) merges and releases — so companion mode is fully dormant until then and can
never dismiss a native modal on an unfixed Grafana.

Final layer of the companion-mode stack.

### DIFF
--- a/src/components/interactive-tutorial/hooks/use-companion-mode.test.ts
+++ b/src/components/interactive-tutorial/hooks/use-companion-mode.test.ts
@@ -1,0 +1,61 @@
+import { renderHook } from '@testing-library/react';
+
+import { panelModeManager } from '../../../global-state/panel-mode';
+import { isModalCoexistenceSupported } from '../../../lib/grafana-version';
+
+import { useCompanionMode } from './use-companion-mode';
+
+jest.mock('../../../lib/grafana-version', () => ({
+  isModalCoexistenceSupported: jest.fn(),
+}));
+
+const mockSupported = isModalCoexistenceSupported as jest.Mock;
+
+let dispatchSpy: jest.SpyInstance;
+const dispatchedTypes = () => dispatchSpy.mock.calls.map((c) => (c[0] as Event).type);
+
+const ACTIVE = { companion: true, isActive: true, isCompleted: false, isPreviewMode: false };
+
+beforeEach(() => {
+  jest.clearAllMocks();
+  mockSupported.mockReturnValue(true);
+  jest.spyOn(panelModeManager, 'getMode').mockReturnValue('sidebar');
+  dispatchSpy = jest.spyOn(document, 'dispatchEvent');
+});
+
+afterEach(() => jest.restoreAllMocks());
+
+describe('useCompanionMode', () => {
+  it('pops the panel out when supported + companion section is active + docked', () => {
+    renderHook(() => useCompanionMode(ACTIVE));
+    expect(dispatchedTypes()).toContain('pathfinder-request-pop-out');
+  });
+
+  it('is a no-op when modal coexistence is unsupported (the version gate)', () => {
+    mockSupported.mockReturnValue(false);
+    renderHook(() => useCompanionMode(ACTIVE));
+    expect(dispatchedTypes()).not.toContain('pathfinder-request-pop-out');
+  });
+
+  it('is a no-op when companion is false or in preview mode', () => {
+    renderHook(() => useCompanionMode({ ...ACTIVE, companion: false }));
+    renderHook(() => useCompanionMode({ ...ACTIVE, isPreviewMode: true }));
+    expect(dispatchedTypes()).not.toContain('pathfinder-request-pop-out');
+  });
+
+  it('docks back on unmount when it owns the floating panel', () => {
+    const { unmount } = renderHook(() => useCompanionMode(ACTIVE));
+    dispatchSpy.mockClear();
+    unmount();
+    expect(dispatchedTypes()).toContain('pathfinder-request-dock');
+  });
+
+  it('does not dock a panel the user already had floating', () => {
+    (panelModeManager.getMode as jest.Mock).mockReturnValue('floating');
+    const { unmount } = renderHook(() => useCompanionMode(ACTIVE));
+    expect(dispatchedTypes()).not.toContain('pathfinder-request-pop-out');
+    dispatchSpy.mockClear();
+    unmount();
+    expect(dispatchedTypes()).not.toContain('pathfinder-request-dock');
+  });
+});
--- a/src/components/interactive-tutorial/hooks/use-companion-mode.ts
+++ b/src/components/interactive-tutorial/hooks/use-companion-mode.ts
@@ -1,0 +1,39 @@
+import { useEffect, useRef } from 'react';
+
+import { panelModeManager } from '../../../global-state/panel-mode';
+import { isModalCoexistenceSupported } from '../../../lib/grafana-version';
+
+export interface UseCompanionModeArgs {
+  companion: boolean | undefined;
+  isActive: boolean;
+  isCompleted: boolean;
+  isPreviewMode: boolean;
+}
+
+/**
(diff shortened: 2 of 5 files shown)


Jayclifford345:
🚧 **The only PR gated on grafana/grafana#126261.** Enables companion auto-pop-out/dock, gated via `isModalCoexistenceSupported()` (sentinel version in `src/lib/grafana-version.ts`) so it stays dormant — and can't dismiss native modals — until the upstream fix releases. #1009–#1011 can merge independently. Addresses #1007.

moxious:
## Review: companion-mode controller + version gate

_Automated review (companion-mode stack, layer 4 of 4 — the capstone)._

**Summary:** A well-scoped, genuinely dormant controller. The version gate is correct and blocks all activation, `weOwnFloatingRef` correctly implements "only dock what we popped out," and `panelModeManager.setMode`'s no-op-on-equal behavior neutralizes the manual-dock race. The failing e2e / CI Gate check is a **stale superseded run**, not a real failure. Two should-fix items: the PR description overclaims (arming/modal-dodge is not in this PR's code), and the ownership ref is set optimistically before the pop-out is confirmed.

### Verdict on the failing e2e check — NOT related to this change; stale/infra
Two CI runs exist on head `c2153cf`:
- Older run `27418034774` (13:15:34Z): **CI Gate = failure**, e2e matrix = **cancelled**. The gate log shows every upstream job reporting `cancelled` (`Lint: cancelled`, `Test: cancelled`, …) — the run was superseded/cancelled mid-flight, so the gate failed purely because its dependencies were cancelled, not because a test failed.
- Newer run (13:20:03Z): **CI Gate = success**, and **all 5 e2e matrix jobs passed** (`grafana-dev@13.1.0`, `grafana-enterprise@13.0.2 / 12.4.4 / 12.3.7`, `grafana@dev-preview-react19`).

The red "e2e test …" / "CI Gate" entries in `pr checks` are placeholder/aggregate rows from the cancelled run. No action needed on the change itself; not a companion-controller regression.

### Blocking
None.

### Should-fix

- **PR description overclaims vs. actual code (description drift).** The description says this PR emits `pathfinder-companion-armed`, starts the modal-watch, and drives `useHighlightDodge(dodgeModal)`. None of that is in the diff: `use-companion-mode.ts:32-37` dispatches only `pathfinder-request-pop-out` / `pathfinder-request-dock`; there is no `pathfinder-companion-armed` emitter or listener anywhere in the tree, and `useHighlightDodge` has no `dodgeModal` parameter gated by an armed event — layer #1011 wired dodging on unconditionally instead. The actual PR is a clean pop-out/dock controller — fine and correctly dormant behind the gate — but the description will mislead the merge reviewer about what safety surface is being activated. Trim the description to match the code, or confirm the armed/dodge wiring is intended for a later change. _(testing-and-verification / cross-cutting-architecture; reversible)_

- **`weOwnFloatingRef` set before the pop-out is confirmed** — `use-companion-mode.ts:29-32` sets `weOwnFloatingRef.current = true` and then dispatches `pathfinder-request-pop-out`. But the receiver `usePopOutHandoff.ts:64-70` can *reject* the request (no active guide / recommendations tab / no URL) — it publishes an alert-info and returns without switching to floating. The hook then still believes it owns the floating panel, and on unmount/complete dispatches `pathfinder-request-dock` for a panel it never popped out. Low-probability (a live companion section usually implies an active guide tab), and `panelModeManager.setMode` is a no-op when mode already equals target (`panel-mode.ts:70-72`) so the spurious dock is usually harmless — but if the user had *independently* popped out meanwhile, this would dock them. Re-check `panelModeManager.getMode() === 'floating'` in cleanup before dispatching dock, or drive ownership off the actual mode transition rather than the optimistic pre-dispatch assumption. _(correctness-and-reliability; reversible)_

### Nit
- **QC8** — the exported-hook JSDoc (`use-companion-mode.ts:11-16`) documents a non-obvious cross-surface invariant (version gate + docking semantics) and is load-bearing — keep it. The inline `// Only dock back what we popped out…` (`use-companion-mode.ts:26`) restates what `weOwnFloatingRef` + the `getMode() === 'sidebar'` check already say; trim on touch. The `TODO(companion)` + sentinel rationale in `grafana-version.ts:3-7` is exactly the right kind of comment (external-dependency workaround with an upstream link) — keep.

### Praise
- **Version gate is genuinely airtight (one-way-door safety).** `isModalCoexistenceSupported()` is the *first-class* guard in the single effect (`use-companion-mode.ts:19`), and every activation path lives past that guard — there is no second code path. Sentinel `[999,0,0]` (`grafana-version.ts:8`) with a `parseVersion` that fails closed to `false` on missing/unparseable versions means companion mode cannot fire on any real Grafana today, so it cannot dismiss a native modal before `grafana/grafana#126261` ships. The TODO + test asserting current versions return `false` and `999.x` flips it on is exactly what's needed to safely lower the sentinel later.
- **Ownership + gate + disarm-on-unmount are all covered** by `use-companion-mode.test.ts` (gate off → no pop-out; `companion:false`/preview → no-op; unmount docks when owned; already-floating → neither pop nor dock).
- **Effect deps are correct** (`[companion, isActive, isCompleted, isPreviewMode]`); the module-singleton `panelModeManager` and `isModalCoexistenceSupported` are stable and correctly omitted. Collapse/expand and complete transitions re-run cleanly with symmetric cleanup.

### Tech debt
No high-confidence patterns in the changed files. One latent seam: three DOM-event string literals (`pathfinder-request-pop-out`/`-dock`/`-companion-armed`) are referenced as raw strings across `use-companion-mode.ts`, `popout-handler.ts`, `usePopOutHandoff.ts`, `FloatingPanelManager.tsx`, and `FullScreenPanel.tsx`; a typo silently no-ops. Pre-existing pattern, non-blocking, but a shared constants module would remove a class of silent-failure bugs as the stack lands.

Jayclifford345:
Closing: with grafana/grafana#126261 shipped in **Grafana v13.1.0**, we've trimmed the companion-mode stack to its two bug-fix layers (resubmitted as #1283 — with those merged, a manually popped-out panel coexists with and dodges native modals on 13.1+). The auto-pop-out controller is **deferred pending real user demand**: auto-moving the panel is a UX commitment we're not ready to make. Design decisions recorded for any future revival: trigger TBD (authored flag vs automatic on modal-open), and **no auto-dock** — once popped out, the panel stays floating until the user docks it manually. The `companion/controller` branch remains.

PR #1016, opened 2026-06-12
feat(e2e): add dependency-aware guide execution planner

leslie-heinzen:
## PR Stack Overview

This is the bottom of a 3-PR stack adding dependency-aware guide chaining to the E2E runner:

1. **#1016 — planner core (this PR)**: the `planGuideExecution` planner, repository-index loaders, and a graph-helper export. No runtime wiring.
2. **#1017 — e2e command integration**: builds the plan in `pathfinder-cli e2e`, runs guides in dependency order, resets the `--clean` docker stack once per chain (not between guides in a chain), adds `--repository`, and reports prerequisite-skips.
3. **#1018 — docs**: updates `docs/developer/E2E_TESTING.md` and `docs/design/e2e-test-runner-design.md`.

## Summary

Adds `planGuideExecution`, a planner that turns a set of selected guides plus a `repository.json` index into an ordered list of dependency **chains**. Guides linked by a hard `depends` prerequisite are grouped together and sorted so prerequisites run first. Missing prerequisites are auto-included if available, and `depends` cycles are rejected. 

This is the foundation the e2e runner uses in `#1017` to keep a prerequisite's state alive for its dependents under `--clean`. This PR does not include any runtime calling code.

## What to look at

- [`src/cli/utils/guide-chains.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/765b72c08d0516e90d781f1bbf9c26d1f74f6a76/src/cli/utils/guide-chains.ts) — the planner.
- [`src/cli/utils/file-loader.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/765b72c08d0516e90d781f1bbf9c26d1f74f6a76/src/cli/utils/file-loader.ts) — `loadRepositoryIndex` (schema-validated; returns `{ repository: null, error }` on failure) and `bundledRepositoryPath`. The null-plus-error contract is what #1017 relies on to decide hard-fail vs graceful degradation.
- [`src/cli/commands/build-graph.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/765b72c08d0516e90d781f1bbf9c26d1f74f6a76/src/cli/commands/build-graph.ts) — `detectCycles` is now exported for reuse by the planner.

## Why

Under `--clean`, the e2e runner resets the docker stack between every guide, wiping `grafana-data`. Guides such as `loki-grafana-101` declare `depends: ["prometheus-grafana-101"]` and rely on state the prerequisite created, so they fail on a freshly-reset environment. This planner computes the ordering and grouping needed to run a chain in one environment.

## How to verify

No runtime entrypoint in this PR, just unit tests to verify.

1. Run the specs for this layer: `npx jest src/cli/utils/guide-chains.test.ts src/cli/utils/file-loader.test.ts`
2. Confirm the key planner behaviors hold: `prometheus-grafana-101` is ordered before `loki-grafana-101` regardless of selection order; a missing prerequisite is auto-included and runs first; a `depends` target that is a virtual capability resolves to its `provides` provider; and a `depends` cycle returns a planning error with empty chains.

## Breaking changes

None. This change is purely additive — a new utility module, one newly-exported function, and new tests. No existing signatures, schemas, or contracts change.

### DIFF
--- a/src/cli/utils/guide-chains.ts
+++ b/src/cli/utils/guide-chains.ts
@@ -1,0 +1,345 @@
+/**
+ * Dependency-aware guide execution planner.
+ *
+ * Groups guides into chains over their hard `depends` prerequisites and orders
+ * each chain so prerequisites run first. Only `depends` participates;
+ * `recommends`/`suggests` are advisory. Dependency targets may be real package
+ * IDs or virtual capabilities resolved through `provides`.
+ */
+
+import type { GraphEdge, GraphEdgeType, RepositoryEntry, RepositoryJson } from '../../types/package.types';
+import { detectCycles } from './graph-cycles';
+import type { LoadedGuide } from './file-loader';
+
+export interface PlannedGuide {
+  id: string;
+  guide: LoadedGuide;
+  dependencies: string[];
+  autoIncluded: boolean;
+}
+
+/** Guides ordered so prerequisites run before dependents. */
+export type GuideChain = PlannedGuide[];
+
+export interface ExecutionPlan {
+  chains: GuideChain[];
+  autoIncludedIds: string[];
+  errors: string[];
+}
+
+export interface PlanGuideExecutionOptions {
+  guides: LoadedGuide[];
+  repository: RepositoryJson;
+  loadGuideById?: (id: string, entry: RepositoryEntry) => LoadedGuide | null;
+}
+
+/**
+ * Derive a guide's bare package ID. Prefers the `id` field in content.json,
+ * falling back to the directory/file name for legacy guides.
+ */
+export function deriveGuideId(guide: LoadedGuide): string {
+  try {
+    const parsed = JSON.parse(guide.content) as { id?: unknown };
+    if (parsed && typeof parsed.id === 'string' && parsed.id.length > 0) {
+      return parsed.id;
+    }
+  } catch {
+    // Not JSON or no id — fall through to the path-derived name.
+  }
+
+  const segments = guide.path.split('/').filter(Boolean);
+  const last = segments[segments.length - 1] ?? guide.path;
+  const parent = segments[segments.length - 2];
+  if (last === 'content.json' && parent !== undefined) {
+    return parent;
+  }
+  return last.replace(/\.json$/, '');
+}
+
+/**
+ * Build a capability → provider-IDs index from a repository, sorted for
+ * deterministic provider selection.
+ */
+function buildProvidesIndex(repository: RepositoryJson): Map<string, string[]> {
+  const index = new Map<string, string[]>();
+  for (const [id, entry] of Object.entries(repository)) {
+    for (const capability of entry.provides ?? []) {
+      const providers = index.get(capability) ?? [];
+      providers.push(id);
+      index.set(capability, providers);
+    }
+  }
+  for (const providers of index.values()) {
+    providers.sort();
+  }
+  return index;
+}
+
+/**
+ * Resolve a single dependency candidate (a real package ID or virtual
+ * capability) to the concrete provider package IDs that could satisfy it.
+ */
+function resolveCandidateProviders(
+  candidate: string,
+  repository: RepositoryJson,
+  providesIndex: Map<string, string[]>
+): string[] {
+  if (repository[candidate]) {
+    return [candidate];
+  }
+  return providesIndex.get(candidate) ?? [];
+}
+
+function clauseCandidates(clause: string | string[]): string[] {
+  return typeof clause === 'string' ? [clause] : clause;
+}
+
+/**
+ * Plan guide execution: resolve dependencies, auto-include missing
+ * prerequisites, detect cycles, and group guides into topologically ordered
+ * dependency chains.
+ */
+export function planGuideExecution(options: PlanGuideExecutionOptions): ExecutionPlan {
+  const { guides, repository, loadGuideById } = options;
+  const errors: string[] = [];
+
+  const providesIndex = buildProvidesIndex(repository);
+
+  const idToGuide = new Map<string, LoadedGuide>();
+  const directDeps = new Map<string, string[]>();
+  const autoIncluded = new Set<string>();
+  const runSet = new Set<string>();
+
+  for (const guide of guides) {
+    const id = deriveGuideId(guide);
+    const existing = idToGuide.get(id);
+    if (existing) {
+      if (existing.path !== guide.path) {
+        // Fail loudly on duplicate IDs from different paths
+        errors.push(`duplicate guide id "${id}" derived from "${existing.path}" and "${guide.path}"`);
+      }
+      continue;
+    }
+    idToGuide.set(id, guide);
+    runSet.add(id);
+  }
+
+  // Load a prereq into the run set on demand.
+  const ensureLoaded = (requesterId: string, providerId: string): boolean => {
+    if (runSet.has(providerId)) {
+      return true;
+    }
+    const providerEntry = repository[providerId];
+    const loaded = providerEntry ? (loadGuideById?.(providerId, providerEntry) ?? null) : null;
+    if (!loaded) {
+      errors.push(`${requesterId}: could not load auto-included prerequisite "${providerId}"`);
+      return false;
+    }
+    idToGuide.set(providerId, loaded);
+    runSet.add(providerId);
+    autoIncluded.add(providerId);
+    return true;
+  };
+
+  // Resolve one clause to a concrete provider: prefer an alternative already in
+  // the run set (so chains reuse it), otherwise the first resolvable candidate.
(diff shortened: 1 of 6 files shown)


src/cli/utils/guide-chains.ts
```diff
@@ -0,0 +1,301 @@
...
+      // Otherwise auto-include the first resolvable provider.
+      if (!providerId) {
+        for (const candidate of candidates) {
+          const providers = resolveCandidateProviders(candidate, repository, providesIndex);
+          if (providers.length > 0) {
+            providerId = providers[0]!;
+            break;
+          }
+        }
+      }
```
leslie-heinzen: This is probably the most opinionated decision here. Basically, if a guide has a prerequisite but it is not in the set of given guides, try to auto-include it from repository.json. 

For example, the command is called with positional args, and a prereq is missing: `pathfinder-cli e2e a.json --clean`; a.json depends on another guide but it is not provided.

IMO this approach has the best runtime "ergonomics", but I could be convinced otherwise. 👀

src/cli/utils/guide-chains.ts
```diff
@@ -0,0 +1,305 @@
...
+    }
+
+    const deps: string[] = [];
+    for (const clause of (entry.depends ?? []) as DependencyList) {
+      const candidates = clauseCandidates(clause);
+
+      // Prefer an alternative that is already part of the run set.
+      let providerId: string | null = null;
+      for (const candidate of candidates) {
+        const provider = resolveCandidateProviders(candidate, repository, providesIndex).find((p) => runSet.has(p));
```
Jayclifford345: [suggestion] Preferring a provider already in `runSet` is the right instinct, but `runSet` mutates during the BFS, which makes the plan depend on selection *order*, not just the selection *set*. Concrete case: `x` has `depends: [["alt-a", "alt-b"]]`, `y` has `depends: ["alt-b"]`, neither alt selected.

- Selection `[x, y]`: `x` is processed first, finds neither alt in `runSet`, auto-includes `alt-a`; then `y` auto-includes `alt-b`. Result: 4 guides, **two chains** — `[alt-a, x]`, `[alt-b, y]`.
- Selection `[y, x]`: `y` auto-includes `alt-b` first; `x` then finds `alt-b` in `runSet` and picks it. Result: 3 guides, **one chain** — `[alt-b, x, y]`.

Same selection set, different chain structure — and in #1017 chain boundaries become `--clean` docker-reset boundaries, so argv order would change how many resets happen and which guides share an environment. The existing "regardless of selection order" test only covers the non-OR case.

A deterministic shape that also keeps the reuse you're after: resolve all forced (single-target) clauses first — they're going to be included no matter what — then resolve OR clauses against the now-stable run set, iterating IDs in sorted order. That makes the plan a pure function of the selection set and picks the shared alternative (`alt-b`) in both orderings. Non-blocking here since nothing calls the planner yet, but worth fixing before #1017 wires it up.

src/cli/utils/guide-chains.ts
```diff
@@ -0,0 +1,305 @@
...
+
+  const idToGuide = new Map<string, LoadedGuide>();
+  const directDeps = new Map<string, string[]>();
+  const autoIncluded = new Set<string>();
+  const runSet = new Set<string>();
+
+  const initialIds: string[] = [];
+  for (const guide of guides) {
+    const id = deriveGuideId(guide);
+    if (!idToGuide.has(id)) {
```
Jayclifford345: [suggestion] If two selected files derive the same ID — say two `content.json` files whose `id` fields collide, or a path-derived name matching another guide's `id` — the second file is silently dropped here. For an e2e runner, silently not executing a guide the user asked for is the kind of thing that turns into a confusing CI investigation. Cheap guard: when `idToGuide.has(id)` and the existing entry's `path` differs from the incoming guide's `path`, push a plan error (or at minimum surface a warning field) instead of ignoring it.

src/cli/utils/guide-chains.ts
```diff
@@ -0,0 +1,305 @@
...
+
+    const entry = repository[id];
+    if (!entry) {
+      // Unmanaged guide (no repository entry). Dependency-free singleton.
+      directDeps.set(id, []);
+      continue;
+    }
+
+    const deps: string[] = [];
+    for (const clause of (entry.depends ?? []) as DependencyList) {
```
Jayclifford345: [nit] `entry.depends` is already `DependencyList | undefined` on `RepositoryEntry` (via `PackageMetadataFields`), so the `as DependencyList` cast is redundant — and if the field's type ever drifts, the cast would keep this compiling instead of surfacing the mismatch. `entry.depends ?? []` alone typechecks.

src/cli/utils/guide-chains.ts
```diff
@@ -0,0 +1,305 @@
...
+ */
+
+import type {
+  DependencyList,
+  GraphEdge,
+  GraphEdgeType,
+  RepositoryEntry,
+  RepositoryJson,
+} from '../../types/package.types';
+import { detectCycles } from '../commands/build-graph';
```
Jayclifford345: [suggestion] This makes a `cli/utils` module depend on a `cli/commands` module — inverted from the usual direction (commands compose utils), and it transitively pulls `build-graph.ts`'s module-level `commander` import into anything that touches the planner. `detectCycles` is a pure graph helper with no command coupling; a `src/cli/utils/graph.ts` home (with `build-graph.ts` importing from there) keeps the layering clean. Totally fine to do as a small follow-up within the stack rather than churning this diff — flagging mainly so `utils → commands` doesn't become a precedent.

Jayclifford345 (commented):
Reviewed via `/review` (routing: `cli-and-e2e-runner` + always-on concerns; classification: `product-runtime` (dev tooling), purely additive).

**No blocking findings.** The planner's core algorithms check out: `detectCycles` reuse is sound (verified it handles self-loops and is direction-agnostic), the topological sort is correct DFS post-order with deterministic tie-breaking, weakly-connected-component grouping is right for "shared environment" semantics, and the `loadRepositoryIndex` error contract lines up exactly with `readJsonFile`'s `not_found`/`invalid_json`/`schema_validation` codes. `bundledRepositoryPath()` follows the existing cwd-based convention in `file-loader.ts`, so no new portability surface. Test coverage is genuinely strong — ordering, transitive auto-include, virtual capabilities, OR-groups, cycle rejection, and both hard gates are all exercised. CI green across the board.

**Reversibility: `reversible`.** Additive module + one export visibility change; no runtime wiring, no persisted state, no contract changes.

Four inline comments below. The one worth resolving before #1017 merges is the OR-group order-dependence — chain structure feeds `--clean` reset boundaries downstream, so plans should be a pure function of the selection *set*, not argv order. The rest are minor.

**Doc-drift:** none — `src/cli/utils/**` is already covered by the `cli-and-e2e-runner` concern's trigger paths, and docs land in #1018 by design.

src/cli/utils/guide-chains.ts
Jayclifford345: Auto-include is the right default for this tool, I think. The alternative behaviors are all worse for the e2e use case: hard-failing on a missing prereq makes positional-arg usage (`pathfinder-cli e2e a.json --clean`) miserable, and silently skipping the dependent would quietly shrink coverage. You've also already built the two things that make auto-include safe to live with: `autoIncludedIds` is surfaced in the plan, and unresolvable/unloadable prereqs hard-fail rather than degrade.

Two things would round it out in #1017: print the auto-included IDs prominently at plan time (so a CI log answers "why did prometheus run when I only asked for loki" without archaeology), and consider a `--no-auto-include` strict flag for anyone who wants exact-selection semantics — cheap to add since the planner already distinguishes the cases.

Jayclifford345 (dismissed):

src/cli/utils/guide-chains.ts
moxious: Cross-tier import inversion: utils/ should not import from commands/

src/cli/utils/guide-chains.ts:14 imports detectCycles from ../commands/build-graph. But src/cli/commands/build-graph.ts:27 already imports from ../utils/file-loader. So we now have:

- commands/build-graph.ts → utils/file-loader.ts
- utils/guide-chains.ts → commands/build-graph.ts

That's a tier inversion (commands orchestrate, utils provide), and it sets the stage for a real circular import the moment anything in build-graph.ts starts using planner output. detectCycles is pure structural graph logic with no command surface — extract it to src/cli/utils/graph-cycles.ts and have both build-graph.ts and guide-chains.ts import from there. The diff is mechanical (move the function, change two import paths) and removes both the inversion and a future C-category architectural debt finding.

leslie-heinzen: Good catch, totally missed that. Extracted detectCycles into `graph-cycles.ts` per this suggestion

Jayclifford345 (approved):

PR #1017, opened 2026-06-12
feat(e2e): run dependency chains in the e2e command

leslie-heinzen:
## PR Stack Overview

Part of the dependency-aware E2E guide-chaining stack: #1016 (planner core) → **#1017 (this PR)** → #1018 (docs). Based on #1016.

## Summary

Wires the planner from #1016 into `pathfinder-cli e2e`. Guides run in dependency order on every run, and under `--clean` the isolated docker stack resets once per dependency chain instead of between every guide, so a prerequisite's state survives for its dependents. Adds an optional `--repository <path>` flag, auto-includes missing prerequisites, and marks dependents of a failed prerequisite as skipped rather than running them against a broken setup.

## What to look at

- [`src/cli/commands/e2e.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/8edac4f812982dab8782a227e68f3939d4ad6294/src/cli/commands/e2e.ts) several load-bearing changes to the command execution code.
- [`src/cli/utils/e2e-reporter.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/609047794afd723b0347f9e0c8a2ec8803261a29/src/cli/utils/e2e-reporter.ts) - updates the reporter to ensure guides skipped as a result of a failed prerequisite are documented in the report.

## Why

Under `--clean` the runner reset the docker stack between every guide, wiping `grafana-data`, which broke any guide that `depends` on another (e.g. `loki-grafana-101` needs the Prometheus setup). Resetting per chain keeps prerequisite state available within a chain while still isolating unrelated chains.

## How to verify

Build the CLI first: `npm run build:cli`.

1. Dependency ordering and shared environment: `pathfinder-cli e2e --clean bundled:loki-grafana-101`. Confirm `prometheus-grafana-101` is auto-included and runs first, and that docker resets once at the start — not between the two guides.
2. Explicit bad `--repository` fails fast: `pathfinder-cli e2e --repository /tmp/does-not-exist.json --bundled` exits with code 2 (`CONFIGURATION_ERROR`).
3. Skip propagation: force a prerequisite to fail and confirm dependents in the same chain print "Skipped (prerequisite failed)" and the run exits 1.

## Breaking changes

No flags removed and no exit-code changes. Behavior change to be aware of: dependency-aware ordering now applies to every run (guide execution order can differ when a `depends` relationship exists), and `--clean` resets per chain rather than per guide.

## Reversibility

Reversible — reverting this PR restores the prior per-guide behavior, and nothing persists beyond the ephemeral docker stack. One note for output-parsing consumers: the console now emits a new `skipped_prereq` outcome and resets at chain boundaries, so any tooling that scraped the old per-guide reset cadence or guide-result lines should be re-checked.

### DIFF
--- a/src/cli/commands/e2e.ts
+++ b/src/cli/commands/e2e.ts
@@ -940,6 +940,16 @@ export const e2eCommand = new Command('e2e')
               exitCode: ExitCode.SUCCESS,
               autoIncluded: planned.autoIncluded,
               failedPrerequisite: blockingDep,
+              // Include a result so the skipped guide is represented in the JSON report
+              resultsData: {
+                guide: { id: planned.id, title: planned.id, path: planned.guide.path },
+                grafanaUrl: options.grafanaUrl,
+                timestamp: new Date().toISOString(),
+                results: [],
+                aborted: true,
+                abortReason: 'SKIPPED_PREREQ',
+                abortMessage: `Prerequisite "${blockingDep}" did not pass`,
+              },
             });
             continue;
           }
--- a/src/cli/commands/e2e.ts
+++ b/src/cli/commands/e2e.ts
@@ -7,13 +7,20 @@
 
 import { execSync, spawn } from 'child_process';
 import { existsSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from 'fs';
-import { join } from 'path';
+import { dirname, isAbsolute, join, resolve } from 'path';
 import { tmpdir } from 'os';
 
 import { Command, Option } from 'commander';
 
 import { validateGuideFromString, toLegacyResult } from '../../validation';
-import { loadGuideFiles, loadBundledGuides, type LoadedGuide } from '../utils/file-loader';
+import {
+  loadGuideFiles,
+  loadBundledGuides,
+  loadRepositoryIndex,
+  bundledRepositoryPath,
+  type LoadedGuide,
+} from '../utils/file-loader';
+import { planGuideExecution } from '../utils/guide-chains';
 import {
   generateReport,
   writeReport,
@@ -29,7 +36,7 @@ import {
   type PreflightOutcome,
   type CurrentTier,
 } from '../utils/manifest-preflight';
-import type { ManifestJson } from '../../types/package.types';
+import type { ManifestJson, RepositoryEntry, RepositoryJson } from '../../types/package.types';
 
 import { randomUUID } from 'crypto';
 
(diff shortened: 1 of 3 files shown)


src/cli/commands/e2e.ts
```diff
@@ -702,6 +729,82 @@ export const e2eCommand = new Command('e2e')
...
+      if (plan.errors.length > 0) {
+        console.error('\n❌ Failed to plan guide execution:');
+        for (const planError of plan.errors) {
+          console.error(`   • ${planError}`);
+        }
+        process.exit(ExitCode.CONFIGURATION_ERROR);
+      }
+
+      // Validate prerequisites that were auto-included to satisfy dependencies.
+      const autoIncludedGuides = plan.chains
```
moxious, line 778: I really like this style and this is a good clean application of what right looks like to me

src/cli/commands/e2e.ts
```diff
@@ -702,6 +729,82 @@ export const e2eCommand = new Command('e2e')
...
+        .flat()
+        .filter((p) => p.autoIncluded)
+        .map((p) => p.guide);
+      if (autoIncludedGuides.length > 0) {
```
moxious, line 782: style weirdness; right after an ultra-clean list pipeline functional operation, the agent I guess wrote a pyramid of doom with nested ifs & fors, which would be readily changeable into map/filter operations.

src/cli/commands/e2e.ts
```diff
@@ -702,6 +729,82 @@ export const e2eCommand = new Command('e2e')
...
+          }
+          process.exit(ExitCode.CONFIGURATION_ERROR);
+        }
+        console.log(
+          `\n➕ Auto-included ${autoIncludedGuides.length} prerequisite guide(s): ${plan.autoIncludedIds.join(', ')}`
+        );
+      }
+
+      if (options.verbose) {
+        console.log(`\n🔗 Execution plan: ${plan.chains.length} chain(s)`);
```
moxious, line 801: not in scope of this PR but we need a separate logging module to replace all of these `console` calls, with selectable log levels, JSON output if the caller wants it, etc.  For later, not now

moxious (approved):
nits, but nothing worth blocking over

src/cli/commands/e2e.ts
leslie-heinzen: Definitely agree 😆, the `.length > 0` guards are super goofy. There's another block that was used to emulate this pattern, I think. I'll do a follow-up cleanup PR for this.

src/cli/commands/e2e.ts
leslie-heinzen: Making a note ✍🏻

PR #1024, opened 2026-06-12
feat(mcp): authoring-session storage lib (stack 1/4)

tomglenn:
**Part 1 of 4** of the authoring-session stack — decomposes the never-opened `gcs-storage` branch (a single +5.5k diff) into reviewable layers.

Foundation: pure new lib modules with **no wiring into the tools yet** (intentionally dormant, each fully unit-tested in isolation):
- `session-store.ts` — in-memory `SessionStore`: sliding TTL, generation optimistic concurrency, eviction stats
- `session-token.ts` — opaque 110-bit Crockford-base32 tokens + log-safe hashing
- `session-store-factory.ts` — memoized process singleton
- `constants.ts`
Plus the Cloud Run `deploy-mcp.sh`, `.env.example`, and the P7 phase plan.

Base `main`. Next: `stack/session-tools` (part 2). Review/merge bottom-up.


src/cli/mcp/lib/constants.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Shared constants used across the MCP server and tool surface.
+ * Hoisted here so cross-file consumers reference one canonical value
+ * rather than redeclaring the literal at each call site.
+ */
+
+/** Grafana plugin viewer base path. Combined with `?doc=…` for guide links. */
+export const PLUGIN_VIEWER_BASE = '/a/grafana-pathfinder-app';
```
moxious, line 8: n.b. this URL is probably defined a second time in the plugin code BUT in this case I think duplication is actually warranted, because for packaging of the CLI we want to avoid causing the CLI to depend on the plugin code directly, unless it's at a very high architectural tier; i.e. we don't want the CLI to depend on react code in any way.

moxious (changes requested):
1. session-store.ts:88-128 — bindMcpSessionId / readMcpSessionPin live on SessionStore. Confidentiality pinning gets fused to artifact storage forever once parts 2-4 import the interface. Cheap fix: split into interface SessionPinStore and have InMemorySessionStore implement both.
2. session-store.ts:111 — delete(token) has no ifGenerationMatch. Adding a precondition later is a signature change every caller must absorb. Cheap fix: accept optional ifGenerationMatch?: number now; impl can no-op until needed.
3. session-store-factory.ts — getDefaultSessionStore is async with no await, and options.now only applies on the cache-miss call. Silent footgun once a transport or test tries to inject a clock against the memoized instance. Either drop options.now from the factory (tests already bypass via __resetSessionStoreFactoryForTests) or rename + document construction-only semantics.

docs/design/phases/ai-authoring-7-gcs-sessions.md
```diff
@@ -0,0 +1,224 @@
+# P7 — GCS-backed authoring sessions
```
Jayclifford345, line 1: hmm this was kind of what I was doing in the other repo. I am fine with them but just seems like a lot of extra markdown in the repo, which will go out of date if not updated? More of a nti than an issue but also bulked your PR but by an extra 224 lines

src/cli/mcp/lib/session-store.ts
```diff
@@ -0,0 +1,260 @@
...
+ * and holds the authoring artifact (`content` + optional `manifest`) and a
+ * monotonic `generation` number used for `ifGenerationMatch` optimistic
+ * concurrency.
+ *
+ * `InMemorySessionStore` is the only implementation: the deployed MCP runs
+ * as a single instance (see `docs/developer/MCP_SERVER.md`), so a
+ * process-local map holds every session and a sliding TTL bounds growth.
+ *
+ * Concurrency model:
+ *   - Every `save` requires an `ifGenerationMatch` value.
```
Jayclifford345: Perhaps we could reduce the comments a little? :)

tomglenn:
Thanks @moxious — all three addressed (the changes ripple up the stack; #1025/#1026/#1027 were rebased + re-pushed accordingly).

1. **Split the interface.** `SessionStore` is now artifact-only (`load`/`save`/`delete`); `SessionPinStore` holds `bindMcpSessionId`/`readMcpSessionPin`; `InMemorySessionStore implements SessionStore, SessionPinStore`. Added `type AuthoringSessionStore = SessionStore & SessionPinStore` as the combined contract the tools wire through — so a future artifact-only backend needn't implement pinning, and the pure-dispatch path (`state-bridge`) stays on plain `SessionStore`. `enforceMcpSessionPin` now types its store as `SessionPinStore`.
2. **`delete` precondition.** Signature is now `delete(token, ifGenerationMatch?: number)`; the in-memory impl ignores it (deletes unconditionally), so adding a real precondition later is non-breaking.
3. **Factory clock footgun.** Dropped `options.now` from `getDefaultSessionStore` entirely — a process singleton shouldn't take a per-call clock. Exported `readTtlMs` and unit-test the env→ms parsing directly; clock-driven eviction stays a store-level test (`session-store.test.ts`, via the constructor's `now`). This also reconciles the earlier request (#1008) to avoid `jest.spyOn(Date)` — neither the factory seam nor a global `Date` mock is needed now.

On the `constants.ts` inline note: agreed, leaving the URL duplicated — the CLI must not reach into plugin/React code, so a small CLI-local constant is the right call. No change.

Verified across the stack: `tsc` clean, 232 MCP tests green, eslint clean.

docs/design/phases/ai-authoring-7-gcs-sessions.md
tomglenn: @moxious what are your thoughts on this? I left your design files mostly untouched when splitting your original PR into these stacked ones. Happy to remove them from the PR entirely tbh but what are your thoughts?

src/cli/mcp/lib/session-store.ts
tomglenn: Yeah these are a bit nuts haha

tomglenn: Done — trimmed hard in 43751668. `session-store.ts` 277→193 and `session-token.ts` 129→78 lines: cut the prose preambles and per-method narration, kept the load-bearing bits (single-instance/TTL, generation optimistic concurrency, the pin 404-not-403 confidentiality boundary, the SessionPinStore segregation rationale, never-log-raw-tokens). No code change. Cascaded through the stack so #1025–#1027 carry it too.

docs/design/phases/ai-authoring-7-gcs-sessions.md
moxious: @Jayclifford345 @tomglenn what I think about this is that this code branch started at a time when I was aggressively checking in design docs in docs/design and Tom inherited the branch; that we let it slide for Tom's PR stack (I didn't review this markdown bit because I knew exactly what it was doing/why it was there).

My take is that when Tom's PR stack is done we do one last PR to savagely cut out the design docs and basically `rm -f docs/design/phases/ai-authoring*`

moxious: This way we can let his PR stack roll as-is, but team decision is going in the direction that we probably don't want phase documents in git; we can't keep PRs small & intact by doing it anyway, so we kill them and stop doing this at the end of this PR stack.

moxious (dismissed):

moxious (approved):

PR #1025, opened 2026-06-12
feat(mcp): activate session-mode in authoring tools (stack 2/4)

tomglenn:
**Part 2 of 4.** Activates session-mode across the MCP authoring surface: two-mode `{artifact}|{sessionToken}` input on mutation/inspection/create tools, fine-grained session read tools (list_blocks/get_block/get_manifest_session), `Mcp-Session-Id` pin enforcement, the state-bridge dispatcher (load→run→save, retry-once-on-412), and the structured HTTP access log. Stateless `{artifact}` mode preserved on every tool.

⚠️ Stacked on `stack/session-lib` (https://github.com/grafana/grafana-pathfinder-app/pull/1024) — **review/merge bottom-up**; the diff shown is just this layer.


moxious (changes requested):
M1 — inspection-tools.ts handlers are not wrapped in withToolErrorEnvelope (concern: correctness-and-reliability)
mutation-tools.ts, artifact-tools.ts, and session-read-tools.ts all wrap handlers in withToolErrorEnvelope so an uncaught throw from a CLI runner becomes a structured INTERNAL_ERROR. pathfinder_inspect and pathfinder_validate do not (src/cli/mcp/tools/inspection-tools.ts:60, :83). A throw in runInspect or runValidate (or in withArtifact's tmpdir setup under disk pressure) would escape as an MCP framework error rather than the well-formed envelope the rest of the surface guarantees. The wrapper is cheap; applying it here closes a small consistency gap.
Reversibility: reversible. Suggested: wrap both handlers the same way the mutation and session-read tools are wrapped.

M2 — finalize.ts documented as a two-mode consumer but not actually wired (concern: cross-cutting-architecture, ai-subsystem)
Two docstrings claim finalize is part of the two-mode contract:

- src/cli/mcp/tools/two-mode-input.ts:8 — "Three tool surfaces speak this contract — read-only (inspect / validate), mutation (add_block / edit_block / …), and finalize."
- src/cli/mcp/tools/read-input.ts:5 — "Used by inspection-tools.ts (inspect / validate) and by finalize.ts."

finalize.ts was not updated in this PR — it still accepts only {artifact}. Either (a) update finalize to accept sessionToken in this PR (the natural extension — the agent that just mutated under session-mode currently has to switch back to stateless mode to publish), or (b) trim the doc claims so they describe what the code actually does. As written, future readers chasing the "finalize is in the contract" claim will be confused.

tomglenn:
Both addressed in 9f20e3a7:

**M1** — `pathfinder_inspect` and `pathfinder_validate` now wrap their handlers in `withToolErrorEnvelope` (with the inbound `sessionToken`), matching `mutation` / `artifact` / `session-read`. An uncaught throw from `runInspect`/`runValidate` (or `withArtifact`'s tmpdir setup) now surfaces as a structured `INTERNAL_ERROR` instead of an MCP framework error.

**M2** — went with option (b): finalize's session-mode wiring lands in the next stack layer, not here, so the docstrings shouldn't claim it already participates. `two-mode-input.ts` and `read-input.ts` now read "finalize adopts it in the next stack layer (session-mode finalize)" rather than listing finalize as a current two-mode consumer.

`tsc` clean, 212 MCP tests green.

moxious (dismissed):

moxious (approved):

PR #1026, opened 2026-06-12
feat(mcp): session-mode finalize + guidance (stack 3/4)

tomglenn:
**Part 3 of 4.** Session-mode `finalize_for_app_platform` (delete-on-success; a mismatched pin does not delete), and authoring-start/SERVER instructions teaching session-mode as the primary workflow. Adds the pin test suite.

⚠️ Stacked on `stack/session-tools` — review/merge bottom-up.

### DIFF
--- a/src/cli/mcp/__tests__/finalize.test.ts
+++ b/src/cli/mcp/__tests__/finalize.test.ts
@@ -11,6 +11,8 @@ import { Client } from '@modelcontextprotocol/sdk/client/index.js';
 import { InMemoryTransport } from '@modelcontextprotocol/sdk/inMemory.js';
 
 import { CURRENT_SCHEMA_VERSION } from '../../../types/json-guide.schema';
+import { InMemorySessionStore, SESSION_GENERATION_ABSENT } from '../lib/session-store';
+import { generateSessionToken } from '../lib/session-token';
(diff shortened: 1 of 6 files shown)


moxious (dismissed):
Observations:
 
- The tool description string grew by ~30% (new fallback explainer). tools/list ships descriptions on every MCP connect — fine for a single tool, but the broader "session-mode primer" pattern is now duplicated across description, inputSchema field descriptions, SERVER_INSTRUCTIONS, and pathfinder_authoring_start. If a future edit changes the canonical phrasing, four call sites drift. Worth tracking but not worth fixing in this PR.
- registerFinalizeTool(server, options) captures mcpSessionId at registration time. This works because the rest of the stack already assumes the HTTP transport builds a fresh McpServer per request (the session-pin.test.ts header comment makes this explicit). Hidden invariant: if a future refactor moves to a long-lived McpServer instance, this capture-once pattern silently disables pin enforcement on every tool that took the same shape. The existing pin tests would catch it. Consider extracting an invariant test ("a long-lived server with two different headers must enforce both pins") at the architectural seam, or documenting the per-request assumption in tools/index.ts. Non-blocking.

moxious (dismissed):

moxious (approved):

PR #1027, opened 2026-06-12
feat(mcp): session observability + hardening tests (stack 4/4)

tomglenn:
**Part 4 of 4.** HTTP access log emits `liveSessions`/`evictions` on tools/call (leak signal on the single always-on instance); logging-discipline + tool-error-envelope tests; MCP_SERVER docs (sessions, single-instance contract, observability).

⚠️ Stacked on `stack/session-finalize` — review/merge bottom-up. With this layer the stack equals the old `gcs-storage` branch, minus the unrelated skill-reference test (separate PR).


moxious (dismissed):
- ⚠️ Minor — logging-discipline.test.ts:372: the token regex is /\b(?:sessionToken|\.pin|\btoken\b)\b/. The bare-word \btoken\b will also fire on innocent identifiers named literally token that have nothing to do with session tokens (e.g. a parser token, a CSRF token). For the current MCP surface this is fine — there's no other token in scope — but the test will be a tripwire for anyone adding lexers or oauth flows under src/cli/mcp/. If that happens, scope the check to a more specific name (sessionToken and .pin only, plus an explicit "treat a local token alias as session token" rule per-file).
- ⚠️ Minor — logging-discipline.test.ts:392-407: the "bare arg" pattern only catches the token as the first or template-interpolated arg. console.warn("hop", "x", sessionToken) slips this second test but is still caught by the first (the wrap-in-prefix check). Effectively redundant guard #2, which is fine — just don't read it as standalone coverage.

moxious (dismissed):

moxious (dismissed):

moxious (approved):

PR #1030, opened 2026-06-12
refactor(user-storage): extract bounded-record factory, type the grafana boundary, hoist event names

moxious:
## Summary

Collapse the duplicated quota-retry CRUD shape shared by `journeyCompletionStorage` and `interactiveCompletionStorage` into a single `createBoundedRecordStorage({ storageKey, limit, label })` factory. While the file is open, also type the previously-`any` `grafanaStorage` boundary as a narrow `GrafanaUserStorage` interface so future `@grafana/runtime` signature drift fails at typecheck, and hoist the two CustomEvent contract names (`learning-progress-updated`, `guide-response-changed`) to a shared `StorageEvents` constants module.

No public API change — every named export from `src/lib/user-storage.ts` keeps the same signature. `user-storage.ts` shrinks by ~170 lines.

## Why

Each fix to the quota-retry path in completion storage had to land twice — `journeyCompletionStorage` and `interactiveCompletionStorage` were structurally identical, only differing in `StorageKeys.*`, `LIMITS.*`, and the diagnostic log strings. PR #909 is a recent example: it patched the once-per-lifecycle quota toast into both namespaces. Future fixes (telemetry on quota-exceeded, jittered retry, alternative cleanup strategies) now have one place to land.

The `grafanaStorage: any` was a known typing gap on three hot functions (`createHybridStorage`, `readGrafanaKeyWithMigration`, `syncFromGrafanaStorage`); narrowing it is mechanical and catches the next `@grafana/runtime` signature drift at typecheck instead of silently.

The eight string-literal event names were a producer/consumer coupling across packages with no compile-time check; centralizing them protects rename safety.

This PR deliberately leaves the higher-risk surgical targets in `user-storage.ts` for follow-ups — see Out of scope.

## What to look at

- `src/lib/storage/bounded-record-storage.ts` — the new factory. The load-bearing detail is the `QuotaExceededError` recursion in `set()`: on quota error it calls `cleanup()` then re-invokes `api.set(key, percentage)`. Trace this against the original journey/interactive paths it replaces; the existing `warnQuotaExceededOnce` test at `src/lib/user-storage.test.ts:431` covers both namespaces through the factory.
- `src/lib/user-storage.ts` — `journeyCompletionStorage` and `interactiveCompletionStorage` are now factory calls. The module-level singleton, sync engine, `__timestamp` migration code, and `useUserStorage` hook are intentionally untouched. `warnQuotaExceededOnce` is now `export`ed so the factory shares the once-per-lifecycle flag.
- `src/types/storage.types.ts` — new `GrafanaUserStorage` interface used by `createHybridStorage`, `readGrafanaKeyWithMigration`, and `syncFromGrafanaStorage`.
- `src/lib/event-names.ts` — new constants module. Consumers in `requirements-checker.hook.ts`, `input-block.tsx`, `useBadgeCelebrationQueue.ts`, and `learning-paths.hook.ts` switched to it. The literal strings in `useBadgeCelebrationQueue.test.ts` are deliberately retained as tripwires for the contract value.

## How to verify

1. **Quota-retry path** — open DevTools, populate enough localStorage to push the journey completion record past 100 entries, complete one more journey, and confirm the percentage saves rather than silently dropping, and that the "Browser storage full" toast appears exactly once per page load.
2. **Cross-device sync** — with a fresh browser session, complete an interactive guide; observe the percentage round-trips through Grafana user storage on next page load (i.e., `syncFromGrafanaStorage` still receives properly-typed `GrafanaUserStorage` calls).
3. **Event-name consumers** — start a guide with input blocks, toggle a value, confirm `requirements-checker` re-evaluates downstream steps. Complete a guide that unlocks a badge and confirm the badge-celebration toast renders.

## Breaking changes

None. Every public export of `src/lib/user-storage.ts` retains the same name, signature, and behavior. Storage key names, envelope format, and CustomEvent string values are unchanged.

## Reversibility

Fully reversible by `git revert`. No storage format change, no migration, no telemetry shape change — the on-disk shape under `StorageKeys.JOURNEY_COMPLETION`, `StorageKeys.INTERACTIVE_COMPLETION`, and the rest is identical before and after.

## Out of scope

Deliberately left out per the high-risk-refactor one-way-door warnings; queued for follow-up PRs:

- **Module-level mutable singletons** (`globalStorageInstance`, `hasSynced`, `hasWarnedAboutQuota`, `completedCountCache`) — init-order sensitive and observable across React and non-React code paths. Refactor to a `StorageRuntime` context object should be a standalone PR.
- **Hybrid-storage write queue + debounce machinery** (`createHybridStorage`) — operational state machine that should move to its own module before being re-platformed.
- **`__timestamp` legacy-format migration** (`getTimestampKey`, `readGrafanaKeyWithMigration`, old-format branches in `syncFromGrafanaStorage`) — load-bearing for users who haven't reopened the panel since envelope migration. Needs a telemetry probe on the `migrated: true` branch before removal.
- **Sync raw-localStorage reads inside async namespaces** (`countAllCompleted`, `countAllAcknowledged`, `hasProgress`) — bypass the `UserStorage` abstraction. Cleanup requires a `UserStorage.keys()` + `getItemSync()` API extension.


Jayclifford345 (commented):
Reviewed via `/review` (concern routing: `reversibility-and-one-way-door`, `correctness-and-reliability`, `testing-and-verification`, `security`, `cross-cutting-architecture`; classification: `product-runtime`).

**No blocking findings.** Behavioral equivalence of the factory against both replaced namespaces checks out line-by-line: clamp, cap-on-write, quota-retry, `getAll`/`cleanup`/`clearAll` semantics, storage keys, envelope shape, and CustomEvent string values are all unchanged. The `any` removal is real — `usePluginUserStorage()`'s return is now structurally checked against `GrafanaUserStorage` at the `createHybridStorage` call site, so runtime signature drift fails at typecheck as claimed. Event-name centralization is complete: the only remaining literals are the deliberate tripwires in `useBadgeCelebrationQueue.test.ts`. The new factory test covers clamp, instance isolation, quota retry, trim, and `clearAll`. CI green across typecheck, unit, and all five e2e matrices.

**Reversibility: `reversible`.** No storage format, key, or event-value changes; revert restores cleanly. `journeyCompletionStorage` gains a `clearAll()` method it didn't have before (factory superset) — additive and harmless, just noting it since the PR body says "same signature."

Three non-blocking inline comments below: the new `user-storage ⇄ bounded-record-storage` import cycle, a doc/behavior mismatch on the quota-retry bound, and QC8 comment hygiene.

**Doc-drift note** (per `prevent-doc-drift` review mode): `docs/design/CONCERNS.md` routes `reversibility-and-one-way-door` on `src/lib/user-storage.ts`, but the quota/completion write path now also lives in `src/lib/storage/**`. A future PR touching only `bounded-record-storage.ts` would not path-activate that concern, and keyword coverage is weak there (the file contains neither `localStorage` nor `StorageKeys` literally). Suggest adding `src/lib/storage/**` to that concern's `trigger_paths` — fine as a ride-along here or a follow-up. `event-names.ts` is already keyword-covered via the `learning-progress-updated` value string.

**Tech debt:** nothing new beyond the cycle comment; the remaining debt in `user-storage.ts` (module-level singletons, sync engine, `__timestamp` migration) is explicitly acknowledged in the PR's out-of-scope section with sensible sequencing.

src/lib/storage/bounded-record-storage.ts
```diff
@@ -0,0 +1,115 @@
+import { createUserStorage, warnQuotaExceededOnce } from '../user-storage';
```
Jayclifford345, line 1: [suggestion] This import creates a static cycle: `user-storage.ts` imports `createBoundedRecordStorage` and calls it at module-eval time, while this module imports `createUserStorage` + `warnQuotaExceededOnce` back from `user-storage.ts`. It's runtime-safe today — both directions only touch hoisted function declarations, and calls are deferred to method invocation — but it's latent load-order sensitivity that neither the tier model nor `architecture.test.ts` will catch (they only model cross-directory edges), and it will bite if either module later adds an eval-time call into the other.

Two cheap ways out: inject the dependencies via config (`createBoundedRecordStorage({ storageKey, limit, label, getStorage, onQuotaExceeded })`, with `user-storage.ts` passing its own functions), or fold the resolution into the planned `StorageRuntime` follow-up you've already queued. If you take the follow-up route, fine as-is — just flagging so the cycle doesn't get copied as a pattern by the next extraction.

src/lib/storage/bounded-record-storage.ts
```diff
@@ -0,0 +1,115 @@
...
+        const storage = createUserStorage();
+        const data = (await storage.getItem<Record<string, number>>(storageKey)) || {};
+        data[key] = Math.max(0, Math.min(100, percentage));
+        await writeWithCap(data);
+      } catch (error) {
+        if (error instanceof Error && error.name === 'QuotaExceededError') {
+          console.warn(`Storage quota exceeded, clearing old ${label} data`);
+          warnQuotaExceededOnce();
+          await api.cleanup();
+          await api.set(key, percentage);
```
Jayclifford345: [suggestion] The interface doc on `set()` says "retries once after `cleanup()`", but this recursion has no depth guard. When the record is under `limit` (so `cleanup()` is a no-op) and the quota is consumed by *other* keys, this loops forever: `set → cleanup → set → …`, logging a `console.warn` per iteration. That's faithfully preserved from both old namespaces — preserving behavior is this PR's contract, so non-blocking — but the stated motivation for the factory is that quota-path fixes now land once. A `retried` flag (internal `setOnce(key, percentage, hasRetried)`) would bound it and make the doc true. Alternatively, fix the doc to say "retries until the write fits" and leave the bound for a follow-up.

src/lib/event-names.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Window CustomEvent names dispatched by storage helpers and consumed by hooks.
+ *
+ * Centralized here so producers and listeners reference the same constant and
+ * TypeScript catches typos in either direction. Kept separate from `storage-keys.ts`
+ * so it can be imported by code paths that don't otherwise depend on storage.
+ */
```
Jayclifford345: [nit] QC8 (comment hygiene, per AGENTS.md catalog), non-blocking: this 7-line preamble on a 15-line module is shape 8, and the "Kept separate from `storage-keys.ts` so…" sentence is shape-2 decision-defense that belongs in the commit message (where it already is). Same family elsewhere in the PR: the one-line interface docs in `bounded-record-storage.ts` (`/** Removes a single entry. */`, `/** Returns the entire record. */`) are shape 1, and the added "Exported so storage helpers extracted to sibling modules…" paragraph on `warnQuotaExceededOnce` in `user-storage.ts` is shape 4 — Find References shows the consumer, and the comment goes stale when consumers change. The `set()` doc is the one worth keeping (it states contract semantics), modulo the retry-bound wording in the other comment.

Jayclifford345 (dismissed):

Jayclifford345 (approved):

PR #1041, opened 2026-06-16
fix(react-compiler): resolve globals + immutability rules

Jayclifford345:
Part of the React Compiler rule migration, stacked on #1038 (which adopted `@grafana/eslint-config` v10 and temporarily deferred react-hooks v7's React Compiler ruleset).

**Stack:** #1038 → **#1041 (this)** → #1042 → #1043

Re-enables `react-hooks/globals` and `react-hooks/immutability` after migrating the call sites they flag.

## Changes
- **interactive-guided.tsx**: replace the module-counter + lazy-ref anonymous step-id generation with `useId()`, preserving the `guided-step-` prefix (E2E `data-step-id` contract) by stripping colons. Removes the now-dead `anonymousGuidedCounter` / `resetGuidedCounter` and their wiring in `interactive-section.tsx` and the test.
- **docs-panel.tsx**: move the `window.__pathfinderPluginConfig` assignment out of render into an effect (App/module already set it at plugin init).
- **step-checker.hook.ts**: move the `checkStep` latest-ref write into an effect; the `immutability` flag on the intentional latest-callback-ref escape hatch is suppressed inline with justification.
- **terminal-connect-step.tsx**: drop a dangling `react-hooks/globals` disable that no longer corresponds to a fired diagnostic.

## Verification
lint (globals + immutability enforced) ✅ · typecheck ✅ · full suite ✅ (253 suites, 4261 tests)

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -117,13 +117,11 @@ export default defineConfig([
   ...baseConfig,
 
   // eslint-config v10 turns on react-hooks v7's React Compiler ruleset. These
-  // four flag pre-existing call sites; deferred to a follow-up. Other new rules stay on.
+  // two flag pre-existing call sites; migrated progressively in the rest of this stack.
   {
     rules: {
       'react-hooks/refs': 'off',
       'react-hooks/set-state-in-effect': 'off',
-      'react-hooks/immutability': 'off',
-      'react-hooks/globals': 'off',
     },
   },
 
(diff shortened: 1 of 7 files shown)


moxious (approved):
- Migrations are mechanically correct. The remaining inline eslint-disable on a latest-callback ref pattern is load-bearing and properly explained.
- useId() swap still satisfies the E2E data-test-* contract (prefix check only).
- The window.__pathfinderPluginConfig global is now set inside useEffect — safe because module.tsx and App.tsx pre-initialize it before first render.

PR #1042, opened 2026-06-16
fix(react-compiler): resolve set-state-in-effect rule

Jayclifford345:
Part of the React Compiler rule migration, stacked on #1038.

**Stack:** #1038 → #1041 → **#1042 (this)** → #1043

Re-enables `react-hooks/set-state-in-effect` after migrating the 14 call sites it flags.

## Restructured (derivable / init-from-props / async-orchestration)
- **content-renderer.tsx**: derive the active tab during render instead of an effect.
- **PrTester.tsx**: seed `fetchState` from restored files via a lazy `useState` initializer.
- **FloatingPanelManager.tsx** / **FullScreenPanel.tsx**: fold the synchronous restoration-done set-state into the shared async continuation.
- **BlockEditorHeader.tsx**: sync `titleDraft` to the external title with the "adjust state during render" pattern.
- **learning-paths.hook.ts**: seed `isDynamicLoading` from whether any URL paths exist.

## Suppressed inline with justification (genuinely effect-appropriate, matching the existing convention in this codebase)
AttendeeJoin (modal lifecycle), GitHubPRModal (reset on reopen), usePanelMode (fullscreen self-heal), interactive-guided (reset on `resetTrigger` with a store write), context.hook (apply suggestions / fetch on context change), TerminalPanel (mirror shared terminal-context expand flag).

## Verification
lint (set-state-in-effect enforced) ✅ · typecheck ✅ · full suite ✅ (253 suites, 4261 tests)

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -116,12 +116,11 @@ export default defineConfig([
   },
   ...baseConfig,
 
-  // eslint-config v10 turns on react-hooks v7's React Compiler ruleset. These
-  // two flag pre-existing call sites; migrated progressively in the rest of this stack.
+  // eslint-config v10 turns on react-hooks v7's React Compiler ruleset.
+  // react-hooks/refs flags pre-existing call sites; migrated in the top of this stack.
   {
     rules: {
       'react-hooks/refs': 'off',
-      'react-hooks/set-state-in-effect': 'off',
     },
   },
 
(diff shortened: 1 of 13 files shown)


moxious (dismissed):
- All 14 sites resolved either via proper derivation (clean R14 fixes — e.g., the TabsWrapper derived-state collapse) or justified inline suppressions for modal-lifecycle resets.
- BlockEditorHeader.tsx correctly uses the React-documented "adjust state during render" pattern (the getDerivedStateFromProps equivalent); the comment is load-bearing.
- learning-paths.hook.ts lazy useState initializer correctly seeds the loading state — no race.

moxious (approved):

leslie-heinzen (approved):

Issue #603, opened 2026-02-21
Epic: Mechanical constraint enforcement for Agents


moxious:
### Problem

Our repo has a rich documentation layer starting with `AGENTS.md` as a navigational index. We did this to shape agent behavior through context injection: read the right file, you'll produce better output.

But agents can read a rule and still violate it. A security guideline about `innerHTML` doesn't stop an agent from using it. The gap between "documented constraint" and "mechanically enforced constraint" is the gap between **hoping agents do the right thing** and **making it structurally impossible for them to do the wrong thing**.

<img width="986" height="497" alt="Image" src="https://github.com/user-attachments/assets/5058efdd-2742-4351-9762-3f52095ae5ce" />

The [OpenAI Harness Engineering team](https://openai.com/index/harness-engineering/) arrived at the same conclusions, and wrote: *"Agents are most effective in environments with strict boundaries and predictable structure."*  We don't always need smarter models, but custom linters, structural tests, enforced dependency directions, and feedback loops that make the repository correct agent execution in real time.

### Value proposition

Every constraint we add pays dividends on every future agent interaction. It's a silent "acceptance criteria" expressed on all agent output. A lint rule with a great error message teaches the agent how to fix its mistake at the exact moment the mistake occurs. A TypeScript compiler error is the most reliable form of agent constraint -- it is impossible to ignore. An architectural boundary encoded in ESLint prevents an entire class of structural error that would otherwise require expensive unwinding.

**This compounds**. If 100 agents build features over a quarter, and each encounters an average of 3 constraint violations per session that would otherwise require human correction, that's 30 correction cycles eliminated. Each constraint we add reduces the correction rate for every subsequent session.

### Strategic framing

The `maintain-docs` skill treats documentation as a materialized view of the codebase -- a precomputed projection that trades freshness for accessibility. This initiative extends that metaphor: we are promoting constraints from the materialized view (docs) into the source of truth (code). A constraint that lives only in a markdown file can drift, be missed, or be misinterpreted. A constraint that lives in `eslint.config.mjs` or `tsconfig.json` is verified on every build, every lint run, every agent interaction. It's not extra context to agents, it's context delivered only when relevant, exactly at the moment needed.

The hierarchy of constraint reliability, from least to most effective:
1. **Prose in documentation** -- agent must find, read, and follow it. 
2. **Rules in `.cursor/rules/`** -- injected into context, but still advisory
[...]

PR #1043, opened 2026-06-16
fix(react-compiler): resolve refs rule and remove the disable block

Jayclifford345:
Top of the React Compiler rule migration, stacked on #1038. **Removes the disable block entirely** — after this PR all four rules (`refs`, `set-state-in-effect`, `immutability`, `globals`) are fully enforced.

**Stack:** #1038 → #1041 → #1042 → **#1043 (this)**

Re-enables `react-hooks/refs` and deletes the React Compiler disable block from `eslint.config.mjs`.

## Restructured (lazy-init refs that don't need to be refs)
- **docs-panel.tsx**: create the attendee `NavigationManager` via a `useState` lazy initializer (stable for the component's life) instead of a render-time lazy-ref write.
- **step-checker.hook.ts**: move the dynamic-import `NavigationManager` init into a mount effect.

## Suppressed inline with justification (intentional latest-value / latest-callback refs whose synchronous update is load-bearing — existing comments already document why)
step-checker eligibility/alignment/check-state/objectives refs, PrTester `manifestFilesRef`, content-renderer `currentContentUrlRef`, FullScreenPanel `dockInputsRef` + `handleExitToSidebarRef`, and the interactive-section `stepRefs`/`multiStepRefs` Maps read inside the ref callback.

> The React Compiler `refs` rule has no clean alternative to the latest-ref pattern without the experimental `useEffectEvent`; these escape hatches are documented and now individually justified rather than blanket-ignored.

## Verification
lint (all four rules enforced, no disable block) ✅ · typecheck ✅ · full suite ✅ (253 suites, 4261 tests)

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -116,14 +116,6 @@ export default defineConfig([
   },
   ...baseConfig,
 
-  // eslint-config v10 turns on react-hooks v7's React Compiler ruleset.
-  // react-hooks/refs flags pre-existing call sites; migrated in the top of this stack.
-  {
-    rules: {
-      'react-hooks/refs': 'off',
-    },
-  },
-
   // Phase 6: Security and architecture lint rules (Epic #603)
   // Mechanically enforce security patterns (F1, F5) and architecture patterns.
   // Test files are excluded — they legitimately use innerHTML for DOM setup/teardown.
--- a/src/components/PrTester/PrTester.tsx
+++ b/src/components/PrTester/PrTester.tsx
@@ -203,6 +203,7 @@ export function PrTester({ onOpenDocsPage }: PrTesterProps) {
    * up-to-date file list when it does run.
    */
   const manifestFilesRef = useRef(manifestFiles);
+  // eslint-disable-next-line react-hooks/refs -- intentional latest-value ref read inside the preload effect (see comment above)
   manifestFilesRef.current = manifestFiles;
 
   /**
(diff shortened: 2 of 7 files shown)


moxious (dismissed):
love it. 

Before this PR, a blanket react-hooks/refs: 'off' meant any ref pattern an agent wrote would pass lint silently — including the render-time lazy-init antipattern (if (!ref.current) { ... }) that the PR explicitly fixes. The rule was effectively disabled for everyone, including agents.

After this PR:

- Agents get immediate, actionable failures when they write a problematic ref pattern. A lint error is a much cleaner signal than a subtle runtime bug from a stale closure or a React Compiler optimization bypass.
- The inline suppressions act as documented examples. When an agent reads the codebase to learn local patterns, it now sees: "here's what a justified latest-value ref looks like, here's what a properly-structured useState lazy initializer looks like." Previously the global disable obscured which sites were intentional and which were just missed.
- New components are compiler-optimizable by default. React Compiler can auto-memoize components that follow the rules. Agents generating new hooks or components within the rules get that optimization for free. Code that violates the rules is either excluded or risks incorrect compilation.
- The escape hatch surface is now explicit and minimal. An agent auditing a file for ref safety can see at a glance whether any suppressions exist and whether they're justified. The blanket disable made that analysis impossible.

Ref: https://github.com/grafana/grafana-pathfinder-app/issues/603

moxious (dismissed):
- The useState(() => new NavigationManager()) lazy-init pattern in docs-panel.tsx is the textbook React Compiler-compliant replacement for a render-time lazy-ref write.
- Moving the dynamic NavigationManager import in step-checker.hook.ts into a mount-only useEffect is behaviorally identical to the prior async import — all callers already guard.
- The six remaining inline suppressions (latest-value / latest-callback refs, ref-callback Maps) are all in the "no clean alternative without experimental useEffectEvent" category, each individually justified rather than blanket-ignored

leslie-heinzen (approved):

PR #1045, opened 2026-06-16
feat(guide-reader): full-screen overlay + interactive-mode foundation

Jayclifford345:
Bottom of the two-tab controller stack — the foundation the rest builds on. No read-only mode (the original read-only viewer was dropped in favor of the interactive controller).

## What it provides

- **`GuideReaderOverlay`**: fetches a guide and renders it full-screen as a portal over `document.body` (the kiosk pattern), behind theme + feature providers — a dedicated reader tab. The controller layers render this same overlay in `mode="controller"`.
- **`InteractiveModeContext` / `useInteractiveMode`**: the per-tab mode the controller layers extend (`'interactive' | 'controller'`).
- **Deterministic anonymous step IDs**: replaces `useId()` with a module-level counter (`resetGuidedCounter` / `resetInteractiveCounters`) so step IDs are stable across renders — needed for the overlay to re-render guides predictably.

## Testing

- Typecheck, lint, prettier — clean; architecture ratchet green.
- Specs: the interactive-mode hook, and `GuideReaderOverlay` (fetch → render → close → error).


src/components/guide-reader/guide-reader.styles.ts
```diff
@@ -0,0 +1,51 @@
...
+  closeButton: css({
+    position: 'fixed',
+    top: theme.spacing(2),
+    right: theme.spacing(2),
+    zIndex: 1,
+    display: 'flex',
+    alignItems: 'center',
+    justifyContent: 'center',
+    width: '36px',
+    height: '36px',
```
leslie-heinzen, line 30: Question/nit: Is there a standard for some of this CSS that's already available? I kinda figured we'd have a mostly ready-to-go large icon button styling available to inherit. If not, no problem. But it'd nice to avoid big bespoke CSS definitions like this where possible.

leslie-heinzen (commented):
**Edit to add:** @Jayclifford345 Just to clarify, this PR review is _mainly_ asking a question -- proposed changes are just if you're like "oh yeah, we _should_ do that." 😄

 
Right now, the useInteractiveReadonly hook seems to be used purely for hiding certain view elements, which is necessary. 👍🏻 

Question though: if we're in readonly mode, should we prevent writing state altogether? There are possibly ways the app can still update the user's guide state that don't involve the hidden UI element interaction (e.g., section completion based on grafana environment state?). Do we want to allow that, or should readonly just full-stop not try to write state at all in the underlying engine code?

If we want to prevent it, I'd like to propose something like this:

* Gate useStepChecker so that if we're in readonly mode, we **cannot** write guide state (early returns in writeStoreCompletion / onStepComplete / onComplete):

```ts
import { useIsInteractiveReadonly } from '../global-state/interactive-readonly-context';

const isReadonly = useIsInteractiveReadonly();

// ...blah

// writeStoreCompletion bails early if isReadonly === true
const writeStoreCompletion = useCallback(
  (reason: 'manual' | 'skipped' | 'objectives'): void => {
    if (isReadonly || storeSectionTarget === null) {
      return;
    }
    markStepCompleted(stepId, storeSectionTarget ?? STANDALONE_SECTION_ID, reason);
  },
  [stepId, storeSectionTarget, isReadonly]
);

// ...blah

useEffect(() => {
  // similar change in this effect that observes step state
  if (isReadonly) {
    return;
  }
  if (state.completionReason === 'objectives' && !disabled) {
    writeStoreCompletion('objectives');
    onStepComplete?.(stepId);
    onComplete?.();
  }
}, [isReadonly, state.completionReason, stepId, disabled, onStepComplete, onComplete, writeStoreCompletion]);

```


I'd also like to propose another view-specific change that is mostly a nit: maybe consider a `<ReadonlyGate />` wrapper component. This centralizes the presentation logic:

```ts
// readonly-gate.tsx

import { useIsInteractiveReadonly } from '../../global-state/interactive-readonly-context';

export function ReadonlyGate({ children }: { children: React.ReactNode }): React.ReactElement | null {
  const isReadonly = useIsInteractiveReadonly();
  return isReadonly ? null : <>{children}</>;
}
``` 

For me is a bit easier to grok when looking at the tsx for a specific component. It's clear to see where the component-specific view logic is versus the global readonly state. The component-specific vs. global state thing helps make the case for the wrapper component for me, but again, take with a grain of salt, definitely just a nit:
```tsx
// terminal-step.tsx

{isEnabled && !isCompleted && (
  <ReadonlyGate>
    <div className={styles.actions}></div>
  </ReadonlyGate>
)}
```

src/components/guide-reader/guide-reader.styles.ts
Jayclifford345: Good call. The full-screen backdrop and the fixed top-corner positioning have to stay bespoke — this is a standalone-root portal rendered over all of Grafana (kiosk pattern), not an in-app panel — but the close button itself doesn't. I can swap it for Grafana's `<IconButton name="times" />` and keep only the fixed-position wrapper, which drops most of this block. Happy to make that change here; let me know if you'd prefer it.

Jayclifford345:
@leslie-heinzen thanks — and your questions actually shaped where this landed. Since this review we dropped the read-only viewing mode entirely: it was a stepping stone, and the interactive controller (the upper PRs in this stack) supersedes it. In the current version of this PR:

- `useIsInteractiveReadonly` is gone — the context is now `InteractiveMode = 'interactive' | 'controller'` (`useInteractiveMode`), with no hide-UI gating anywhere.
- Your state-writing question is exactly why I was comfortable dropping read-only rather than hardening it: you were right that the hook only hid UI, so a read-only guide could still auto-complete via the objectives path (the `writeStoreCompletion('objectives')` effect in `useStepChecker` wasn't gated) — a real footgun. There's no longer a *passive* viewer that can silently mutate progress; the controller is fully interactive (the user is actively driving the live tab), so writing completion there is intended.
- The `<ReadonlyGate>` centralization was a nice idea, but with the gating removed there's nothing left to wrap.

So this PR is now just the full-screen overlay + the interactive-mode context foundation. Appreciate the review!

moxious:

# PR #1045: feat(guide-reader): full-screen overlay + interactive-mode foundation

**Verdict:** Request changes

## Findings

1. **[high] `useGrafanaTheme` — stale-closure bug / spurious re-subscription**
   `GuideReaderOverlay.tsx`, `useGrafanaTheme`, the `useEffect`
   The effect has `[theme]` in its dependency array. Every time the theme object changes (i.e. on every theme switch) the observer is torn down and re-created with the new `theme` value closed over — which is correct by accident, but it also means the inner `config.theme2 !== theme` reference-equality guard will pass on the *first* mutation after every remount, even if the theme didn't actually change (because `config.theme2` is the same object reference). More importantly, the intent is *"observe body-class changes and sync theme"* — the observer should be created once and the handler should always read `config.theme2` fresh rather than capturing stale `theme`. Fix: empty dep array `[]`; remove the captured `theme` from the guard (`setTheme(config.theme2)` unconditionally, or compare strings if deduplication matters).

2. **[medium] `{<JSX />}` no-op wrapping in five files — dead scaffolding left in**
   `interactive-step.tsx` (two sites), `interactive-section.tsx`, `interactive-guided.tsx`, `interactive-multi-step.tsx`, `interactive-quiz.tsx`, `challenge-block.tsx`
   Every modified action-button block is wrapped in a bare `{ <div>...</div> }` expression — a JSX no-op that generates no additional DOM but clearly signals "this will become `{interactiveMode === 'interactive' && <div>...}` later." None of the modified files import or call `useInteractiveMode`. Shipping scaffolding that silently has no effect, with the actual conditional left for a later PR, makes this PR's diff misleading and the intent invisible to reviewers of those later PRs. Either: (a) apply the conditional now (`{mode === 'interactive' && ...}`), or (b) revert the wrapping and add it in the PR that actually wires up the context check.

3. **[medium] `interactive-guided.tsx` still uses `useId()` — counter migration incomplete**
   `interactive-guided.tsx`, line 144: `const generatedStepId = \`guided-step-${useId().replace(/:/g, '')}\``
   All other components in this PR (`interactive-step`, `interactive-multi-step`) were migrated from `useId()` to a module-level deterministic counter with a `resetXxxCounter()` export. `interactive-guided.tsx` was not. The PR description says the counter replacement covers "~5 files" but the diff only converts two. If the overlay's predictable re-render requirement applies here too (and guided steps can appear inside the overlay), this is a correctness gap. If guided steps can't appear in the overlay, document why this file was intentionally skipped.

4. **[medium] `GuideReaderOverlay` comment — QC8 violation (defends a non-action / whole-file docstring)**
   `GuideReaderOverlay.tsx`, line 48–53: the JSDoc on `GuideReaderOverlay` explains what it *is* (full-screen viewer, new tab, portal, high z-index, covers chrome) — information already apparent from the 8 lines of JSX below it. The "not a second live Grafana" rationale belongs in the commit message, not the file. Delete or compress to one line naming the invariant that isn't visible in code (e.g. the z-index rationale if that's load-bearing).
   Similarly the inline comment on `backdrop` in `guide-reader.styles.ts` (lines 12–15) is a QC8-shape-2 "defends a non-action" — why it's opaque rather than blurred. Belongs in the commit message.

5. **[medium] `useGrafanaTheme` comment — QC8 violation (references dead process artifacts / narrates)**
   `GuideReaderOverlay.tsx`, line 24: `// Mirrors FloatingPanelManager: this tree is mounted in a standalone root (createCompatRoot)...` — "Mirrors FloatingPanelManager" is a renamed-along-with-symbol shape; if `FloatingPanelManager` is renamed or removed the comment becomes a lie. The load-bearing fact is *"this component lives outside Grafana's provider tree and must supply its own theme"*. Trim to that one sentence.

6. **[low] `interactive-readonly-context.ts` — `InteractiveMode` type is a union of one**
   `src/global-state/interactive-readonly-context.ts`, line 3: `export type InteractiveMode = 'interactive'`
   A union of one member is not a union — it's a string literal alias with no discriminant value. The PR description acknowledges it will become `'interactive' | 'controller'` in PR #1062. Shipping a type that cannot yet be consumed meaningfully (there is no `Provider` wrapping any tree in this PR, and no site reads it) creates churn: every file that imports the type gets a spurious re-check when the union widens. Consider either landing the full two-value union now (even if `'controller'` is unused until #1062), or deferring the context file entirely to the PR that first uses it.

7. **[low] Test: no Escape-key close case**
   `GuideReaderOverlay.test.tsx` — three cases cover fetch→render, close-button click, and fetch error. The Escape key handler (`document.addEventListener('keydown', ...)`) in `GuideReaderInner` is untested. Add a `fireEvent.keyDown(document, { key: 'Escape' })` case alongside the close-button test.

8. **[low] `document.body.style.overflow = 'hidden'` side-effect not restored on error path**
   `GuideReaderOverlay.tsx`, the keyboard `useEffect` sets `document.body.style.overflow = 'hidden'` and restores it in the cleanup. The cleanup runs correctly on unmount. However, if the component is unmounted synchronously (e.g. a parent catches an error boundary before the effect cleanup fires) the scroll lock leaks. This is an edge case, but it's worth noting that the overflow mutation and the keydown listener are two distinct concerns in one effect — splitting them makes each cleanup easier to reason about and test independently.

moxious:

Context: reviewing the full 16-PR two-tab stack (#1133) bottom-first against `main@fb90805` before any merge. This is one of the two foundation PRs, so findings here are deliberately load-bearing. Posting as context, not a change request yet.

**What's good:** `GuideReaderOverlay` correctly mounts outside Grafana's provider tree with its own `ThemeContext` (mirrors FloatingPanelManager), cancels its in-flight fetch on unmount, restores `body.overflow`, and has solid a11y (`role="dialog"`, `aria-modal`, `aria-label`) plus testIds. The genuinely new code — the overlay, its styles, and the context definition — reads cleanly.

### 1. ~210 lines of this PR are no-op JSX churn (stack health — high)

`interactive-step.tsx` (`+105/−101`) and `interactive-section.tsx` (`+107/−105`) account for most of the diff, but the change is almost entirely **wrapping existing JSX blocks in bare `{ … }` expression braces and re-indenting**. The braces are never filled with a conditional — at the top of the stack the final file still reads:

```tsx
<div className="interactive-step-actions">
  {
    <div className="interactive-step-action-buttons">
```

The actual controller-mode behavior lives in the handlers (`mode === 'controller'`), not in this JSX. So the real change to these two files in this PR is essentially nil.

Why it matters for the stack: these exact files are re-touched by six PRs above this one — `interactive-step.tsx` by #1063 / #1070 / #1072, and `interactive-section.tsx` / guided / multi-step by #1068 / #1073 / #1084. The re-indentation churn here will produce noisy, avoidable rebase conflicts all the way up the cascade. Dropping the braces now makes the whole stack cheaper to land.

### 2. Unrelated `StorageEvents` → raw-string change (scope creep)

`interactive-section.tsx` swaps one dispatch from `StorageEvents.InteractiveProgressCleared` to the raw literal `'interactive-progress-cleared'` and drops the import — but **line 1077 of the same file still uses the constant**, and 15+ call sites across the repo use it. The values match, so there's no runtime break, but it leaves the file internally inconsistent and is unrelated to the two-tab feature. Suggest reverting to the constant.

### 3. `interactive-readonly-context` is misnamed and ships degenerate (A3)

The seam is a fine mechanism (a React context carrying a mode enum), but in this PR `type InteractiveMode = 'interactive'` is a single-member union with no provider and no consumer; #1062 widens it to `'interactive' | 'controller'`. The concept that actually emerged is "controller," not "readonly." Since this naming propagates through the rest of the stack, renaming it now (while it's cheap) would reduce confusion later.

### Smaller nits

- `handleClose` uses `window.close()`, which is a no-op for tabs the script didn't open (a directly-navigated or bookmarked `?doc=` URL). Escape and the close button then silently do nothing — consider a fallback.
- `useGrafanaTheme` re-subscribes its `MutationObserver` on every theme change (`[theme]` dependency); a ref would avoid the re-subscribe.

Tracking IDs `F-1045-1..5` in the stack review ledger.

moxious:
**Fix-pass outcome (re-stacked, gate-green, signed, pushed `bddec459`)**

Foundation branch. Removed ~210 lines of no-op JSX brace churn (the stack's worst rebase hotspot), renamed `interactive-readonly-context` → `interactive-mode-context` (the readonly concept was dead), migrated guided steps off mount-specific `useId()` to a deterministic counter, and fixed the overlay theme re-subscribe + a `window.close()` fallback for bookmarked deep links.

leslie-heinzen (dismissed):

leslie-heinzen (approved):

PR #1055, opened 2026-06-17
refactor(storage): phase 0-1 groundwork — characterization tests + key inventory

moxious:
## Summary

Groundwork for the user-storage refactor, split into two independently reviewable phases:

- **Phase 0 (tests)** — characterization/tripwire tests that pin the highest-risk, previously-untested async paths in `user-storage.ts` (the hybrid write queue and Grafana sync) before any decomposition begins.
- **Phase 1 (refactor)** — centralize all storage keys so we can see the sprawl of localStorage across the code base.  Lock a contract down so the storage keys can't change.

No production behavior changes, and no stored key-string values change.

## Why

The pre-refactor analysis flagged the hybrid write-queue drain as "the highest-risk path in the entire storage surface" with zero tests, and found ~25 storage keys defined as inline literals outside `StorageKeys` (block editor, Coda terminal, assistant, devtools, feature-flag overrides). Locking that behavior with tests and completing the key inventory are the two lowest-risk, highest-leverage prerequisites for the larger refactor: they de-risk later extractions and give one complete map of every persisted key. Renaming the stored key strings themselves was deliberately excluded — that would orphan existing user data for cosmetic gain.

## What to look at

- [`src/lib/user-storage.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/0be02597744a206f92fbf785f42aec592dae7c2d/src/lib/user-storage.ts) — the only additions are four test-only exports plus `__resetSyncedForTests`, mirroring the existing `__resetQuotaWarningForTests`. Confirm the queue/sync bodies themselves are untouched.
- [`src/lib/user-storage.plumbing.test.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/0be02597744a206f92fbf785f42aec592dae7c2d/src/lib/user-storage.plumbing.test.ts) — the anti-stranding "re-check drain" test is the subtle one: it deliberately lets the second debounce timer fire and bail while the first drain is still awaiting, proving the post-drain re-check rescues the stranded item.
- [`src/lib/storage-keys.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/0be02597744a206f92fbf785f42aec592dae7c2d/src/lib/storage-keys.ts) + [`storage-keys.test.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/0be02597744a206f92fbf785f42aec592dae7c2d/src/lib/storage-keys.test.ts) — new keys carry their exact prior literal values; the contract test asserts the full map so any future rename fails loudly (Playwright E2E injects by key string and stored user data is addressed by these strings, so a rename is a silent, data-orphaning break).
- Call-site repoints — block editor (`constants.ts`, health/condition UI prefs), Coda [`terminal-storage.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/0be02597744a206f92fbf785f42aec592dae7c2d/src/integrations/coda/terminal-storage.ts), assistant integration (removes the triplicated `pathfinder-assistant-` literal via `buildAssistantStorageKey`), devtools panels, `openfeature.ts` flag overrides, and `experiment-utils.ts` legacy key. Each swaps an inline literal for the shared constant; values are byte-identical.

## How to verify

1. With existing guide progress in localStorage, reload Pathfinder and confirm completion / tabs / streaks still load — key strings are unchanged, so prior data must still resolve.
2. Open the Coda terminal, set a height and toggle open/closed, then refresh — confirm open state and height persist (keys now sourced from `StorageKeys.CODA_TERMINAL_*`).
3. In the block editor, toggle the health panel and the condition field's raw-mode toggle, then reload — confirm both preferences persist.
4. Set a feature-flag override through the console flow and confirm it still applies on the next load.

## Breaking changes

None. No key string values changed — this is a literal-to-constant swap plus new tests.

## Reversibility

Fully reversible. No persisted data shape, stored key string, or telemetry payload changes, so a straight revert restores prior state with no migration or manual cleanup.

## Out of scope

- The `module.tsx` entry-point legacy auto-open literal — left as-is to avoid adding an import to the bootstrap path (the refactor bootstrap-contamination gate).
- A highlighted-guide `console.log` hint string that mentions a key prefix — cosmetic, not a functional key.
- Renaming stored key strings to a single prefix, routing raw `localStorage` through named domains, and the badge/analytics dependency inversion — these are later phases of the refactor.

### DIFF
--- a/src/components/block-editor/forms/ConditionChipsField.tsx
+++ b/src/components/block-editor/forms/ConditionChipsField.tsx
@@ -28,9 +28,10 @@ import {
   isValidRequirement,
 } from '../../../types/requirements.types';
 import { isAutoRecoverableRequirement } from '../../../recovery';
+import { StorageKeys } from '../../../lib/storage-keys';
 import { HELPER_BY_PREFIX } from './condition-helpers';
 
-const RAW_MODE_PREFERENCE_KEY = 'pathfinder.blockEditor.conditionField.rawMode';
+const RAW_MODE_PREFERENCE_KEY = StorageKeys.BLOCK_EDITOR_CONDITION_RAW_MODE;
 
 export type ConditionFieldMode = 'requirements' | 'objectives' | 'conditions' | 'verify';
 
--- a/src/lib/storage-keys.test.ts
+++ b/src/lib/storage-keys.test.ts
@@ -1,0 +1,89 @@
+/**
+ * Contract / stability test for the centralized storage key registry.
+ *
+ * These exact string values are a load-bearing contract: Playwright E2E tests
+ * inject localStorage by key, `experiment-debug` sweeps by prefix, and existing
+ * user data in the wild is addressed by these strings. A rename is a silent,
+ * data-orphaning breaking change — this test makes any change to a key VALUE
+ * fail loudly so it must be a deliberate, reviewed decision (with a migration).
+ *
+ * When ADDING a new key, add it here too. When CHANGING an existing value, stop
+ * and consider the migration implications before updating this test.
+ */
+import { StorageKeys } from './storage-keys';
+
+describe('StorageKeys — stable string contract', () => {
+  it('matches the locked key values exactly', () => {
+    expect(StorageKeys).toEqual({
+      // Core learner / progress state
+      JOURNEY_COMPLETION: 'grafana-pathfinder-app-journey-completion',
+      INTERACTIVE_COMPLETION: 'grafana-pathfinder-app-interactive-completion',
+      TABS: 'grafana-pathfinder-app-tabs',
+      ACTIVE_TAB: 'grafana-pathfinder-app-active-tab',
+      INTERACTIVE_STEPS_PREFIX: 'grafana-pathfinder-app-interactive-steps-',
+      WYSIWYG_PREVIEW: 'grafana-pathfinder-app-wysiwyg-preview',
+      WYSIWYG_PREVIEW_JSON: 'grafana-pathfinder-app-wysiwyg-preview-json',
+      E2E_TEST_GUIDE: 'grafana-pathfinder-app-e2e-test-guide',
+      SECTION_COLLAPSE_PREFIX: 'grafana-pathfinder-app-section-collapse-',
+      SECTION_ACKNOWLEDGED_PREFIX: 'grafana-pathfinder-app-section-acknowledged-',
+      SECTION_DONE_PREFIX: 'grafana-pathfinder-app-section-done-',
+      FULLSCREEN_MODE_STATE: 'grafana-pathfinder-app-fullscreen-mode-state',
+      FULLSCREEN_BUNDLED_STEPS: 'grafana-pathfinder-app-fullscreen-bundled-steps',
+      FULLSCREEN_BUNDLING_ACTION: 'grafana-pathfinder-app-fullscreen-bundling-action',
+      FULLSCREEN_SECTION_INFO: 'grafana-pathfinder-app-fullscreen-section-info',
+      MILESTONE_COMPLETION: 'grafana-pathfinder-app-milestone-completion',
+      LEARNING_PROGRESS: 'grafana-pathfinder-app-learning-progress',
+      GUIDE_RESPONSES: 'grafana-pathfinder-app-guide-responses',
+
+      // Experiment / feature-flag state
+      EXPERIMENT_AUTO_OPEN: 'grafana-pathfinder-app-experiment-auto-open',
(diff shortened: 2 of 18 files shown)


src/components/block-editor/forms/ConditionChipsField.tsx
```diff
@@ -28,9 +28,10 @@ import {
   isValidRequirement,
 } from '../../../types/requirements.types';
 import { isAutoRecoverableRequirement } from '../../../recovery';
+import { StorageKeys } from '../../../lib/storage-keys';
 import { HELPER_BY_PREFIX } from './condition-helpers';
 
-const RAW_MODE_PREFERENCE_KEY = 'pathfinder.blockEditor.conditionField.rawMode';
+const RAW_MODE_PREFERENCE_KEY = StorageKeys.BLOCK_EDITOR_CONDITION_RAW_MODE;
```
leslie-heinzen, line 34: Super nit: this intermediary variable doesn't seem super valuable now that we're importing type-safe strings from StorageKeys and this var is not exported, but for the sake of reducing blast radius I like the choice to keep it (same for similar changes in other files). Maybe a future chore PR to clean up pass-through constants like this.

src/lib/storage-keys.test.ts
```diff
@@ -0,0 +1,87 @@
...
+ * data-orphaning breaking change — this test makes any change to a key VALUE
+ * fail loudly so it must be a deliberate, reviewed decision (with a migration).
+ *
+ * When ADDING a new key, add it here too. When CHANGING an existing value, stop
+ * and consider the migration implications before updating this test.
+ */
+import { StorageKeys, buildAssistantStorageKey } from './storage-keys';
+
+describe('StorageKeys — stable string contract', () => {
+  it('matches the locked key values exactly', () => {
```
leslie-heinzen, line 16: 👍🏻

src/lib/storage-keys.ts
```diff
@@ -47,13 +47,65 @@ export const StorageKeys = {
   // not session — auto-open fires once per browser, not once per session.
   HIGHLIGHTED_GUIDE_AUTO_OPEN_PREFIX: 'grafana-pathfinder-highlighted-guide-auto-open-',
   HIGHLIGHTED_GUIDE_RESET_PROCESSED_PREFIX: 'grafana-pathfinder-highlighted-guide-reset-processed-',
+  // Legacy global treatment auto-open key (kept for backwards compatibility).
+  // Used with a `{hostname}` suffix. Not migrated — see G5/G10 notes.
```
leslie-heinzen, line 51: Question: "G5/G10" - are these docs/notes to come? I can find G5 in the PR_REVIEW skill for `.go` files (which I don't think is what this is supposed to reference), and nothing for G10.

src/utils/experiments/experiment-utils.ts
```diff
@@ -23,7 +23,7 @@ export const getStorageKeys = (hostname: string) => ({
   // sessionStorage - cleared when browser closes
   autoOpened: `${StorageKeys.EXPERIMENT_SESSION_AUTO_OPENED_PREFIX}${hostname}`,
   // Legacy global treatment key (kept for backwards compatibility)
-  treatmentOpened: `grafana-pathfinder-experiment-treatment-opened-${hostname}`,
+  treatmentOpened: `${StorageKeys.EXPERIMENT_TREATMENT_OPENED_LEGACY_PREFIX}${hostname}`,
   // Per-page treatment prefix
   treatmentPagePrefix: `${StorageKeys.EXPERIMENT_TREATMENT_PAGE_PREFIX}${hostname}-`,
 });
```
leslie-heinzen, line 29: Note:

The test file for this module still uses hard-coded strings to mock `StorageKeys`, but I think it makes sense to use the shared constants. It's almost missing an entry for the `treatmentOpened` value: https://github.com/grafana/grafana-pathfinder-app/blob/bbd5e10333d2ca09ceafa401e10443bdca4b633c/src/utils/experiments/experiment-utils.test.ts#L15-L19


And `treatmentOpened` is not asserted in the `getStorageKeys` spec like the other values: https://github.com/grafana/grafana-pathfinder-app/blob/bbd5e10333d2ca09ceafa401e10443bdca4b633c/src/utils/experiments/experiment-utils.test.ts#L69-L75)

Anyway, unrelated enough to this PR's purpose that it makes sense to defer. I could go either way.

leslie-heinzen (approved):
A few non-blocking questions/comments 👍🏻

src/lib/storage-keys.ts
moxious: this is bleed over referencing a larger planning document I'm not checking in, and this keys back to high risk migration guidance patterns identified in this file

PR #1056, opened 2026-06-17
feat(twotab): add same-origin cross-tab transport

Jayclifford345:
A BroadcastChannel-backed transport linking a popped-out controller tab
to the live Grafana tab. It carries typed CrossTabMessage payloads (step
commands + heartbeats), stamps a per-tab senderId to drop a tab's own
echoes, and takes an injectable channel factory for deterministic tests.
Plumbing only — no UI; the controller-emit and live-executor layers sit
on top of this.

---

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -189,19 +189,21 @@ Excluded from tier analysis (not tiered): `test-utils/`, `cli/`, `bundled-intera
 
 **Key dependency edges** (where the load-bearing wiring lives):
 
-| Edge                                           | Why                                                            |
-| ---------------------------------------------- | -------------------------------------------------------------- |
-| `context-engine` → `docs-retrieval`            | Fetches content for the recommendations it surfaces            |
-| `docs-retrieval` → `bundled-interactives`      | Fallback when the online CDN is unavailable                    |
-| `docs-retrieval` → `package-engine`            | Resolves package manifests + content                           |
-| `docs-retrieval` → `snippet-engine`            | Inlines `snippet-ref` blocks against the CDN at parse time     |
-| `components/docs-panel` → `interactive-engine` | Executes step actions when the user clicks "Show me" / "Do it" |
-| `interactive-engine` → `requirements-manager`  | Checks prereqs before enabling / executing a step              |
-| `interactive-engine` → `lib/dom`               | Selector resolution + element detection                        |
-| `components/docs-panel` → `context-engine`     | Reads and renders recommendations                              |
-| `components/docs-panel` → `global-state`       | Sidebar, panel-mode, tab persistence                           |
-| `learning-paths` → `lib/user-storage`          | Persists progress + streak data                                |
-| `recovery` → `requirements-manager`            | Decides whether a failed requirement is auto-recoverable       |
+| Edge                                                          | Why                                                                                |
+| ------------------------------------------------------------- | ---------------------------------------------------------------------------------- |
+| `context-engine` → `docs-retrieval`                           | Fetches content for the recommendations it surfaces                                |
+| `docs-retrieval` → `bundled-interactives`                     | Fallback when the online CDN is unavailable                                        |
+| `docs-retrieval` → `package-engine`                           | Resolves package manifests + content                                               |
+| `docs-retrieval` → `snippet-engine`                           | Inlines `snippet-ref` blocks against the CDN at parse time                         |
+| `components/docs-panel` → `interactive-engine`                | Executes step actions when the user clicks "Show me" / "Do it"                     |
+| `interactive-engine` → `requirements-manager`                 | Checks prereqs before enabling / executing a step                                  |
+| `interactive-engine` → `lib/dom`                              | Selector resolution + element detection                                            |
+| `components/docs-panel` → `context-engine`                    | Reads and renders recommendations                                                  |
+| `components/docs-panel` → `global-state`                      | Sidebar, panel-mode, tab persistence                                               |
+| `learning-paths` → `lib/user-storage`                         | Persists progress + streak data                                                    |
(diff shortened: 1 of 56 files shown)


moxious:

# PR #1056: feat(twotab): add same-origin cross-tab transport

**Verdict:** Request changes

## Findings

1. **[Bug] `stop()` clears listeners, making restart impossible** — `src/lib/cross-tab-transport.ts`, `stop()`
   `this.listeners.clear()` inside `stop()` means any caller that holds an `onMessage` unsubscribe handle and later calls `start()` again will receive nothing. The transport is presented as a restartable lifecycle (`start`/`stop`/`start`), so this is a real bug: the second `start()` silently drops all registered subscribers. Fix: remove `this.listeners.clear()` from `stop()`. Subscribers own their own lifecycle via the returned unsubscribe function; `stop()` should only tear down the channel.

2. **[Minor bug] `stop()` does not guard against a null channel** — `src/lib/cross-tab-transport.ts`, `stop()`
   `start()` guards with `if (this.channel) return`, but `stop()` calls `this.channel.onmessage = null` unconditionally. If `stop()` is called before `start()` (or after a double-stop) this throws. One null-guard at the top of `stop()` fixes it.

3. **[API contract] `channelFactory` return type should be documented as non-nullable here** — `src/lib/cross-tab-transport.ts`
   PR 1063 changes `defaultChannelFactory` to return `BroadcastChannelLike | null` when `BroadcastChannel` is unavailable. This PR types `ChannelFactory` as returning `BroadcastChannelLike` (non-nullable), so `start()` does not null-check the result. When 1063 lands, the two signatures will be in conflict. Either widen the type now and guard in `start()`, or add a note in the type alias that the factory must guarantee a non-null return (and defer the widening to 1063). Either is fine — just needs to be a conscious choice, not accidental divergence.

4. **[Low] `createSenderId` fallback is not a security concern, but the comment overstates the guarantee** — `src/lib/cross-tab-transport.ts`, `createSenderId()`
   `Math.random()` is not cryptographically random, but the sender ID is only used to filter echo messages (same-sender suppression). It is not a secret, not used for authorization, and a collision merely means a tab briefly ignores its own message — a cosmetic glitch, not a security issue. No change needed on the implementation. However, if there is a comment claiming uniqueness or randomness guarantees, trim it; the code already shows the fallback path clearly.

5. **[Pass] Test coverage is solid** — `cross-tab-transport.test.ts`
   The `FakeBus`/`FakeChannel` double correctly models BroadcastChannel fan-out semantics. Echo-suppression, multi-listener dispatch, and unsubscribe are all exercised. The one gap that follows from finding #1 above: there is no test for `stop()` followed by `start()` verifying that previously registered listeners still fire. Add that case once the `listeners.clear()` bug is fixed.

moxious:

Context: reviewing the full 16-PR two-tab stack (#1133) bottom-first against `main@fb90805` before any merge. This PR is one of the two foundation PRs, so findings here are deliberately load-bearing — cheaper to settle now than above the stack. Posting as context, not a change request yet.

**What's good:** clean transport/protocol separation (`cross-tab-transport.ts` stays ~90 LOC and never changes again across the stack), solid lifecycle (`start` idempotent; `stop` nulls `onmessage` + closes + clears listeners; `onMessage` returns an unsubscribe), and disciplined Tier-0 typing — defining local `CrossTabAction` / `RemoteRequirementError` structural mirrors instead of importing up-tier types respects the import-tier model.

### 1. Receive gate validates the envelope, not the message body (security — likely blocking)

`isPathfinderMessage` checks only `source === 'pathfinder'` and that `senderId` is a string. It does not validate `kind` or the per-message payload before forwarding to listeners. BroadcastChannel is same-origin, but Grafana runs many plugins on that origin — any co-resident plugin, content script, or injected script can post a `step-command` or `fix-requirement` envelope and drive the live-tab executor's DOM sink (Navigate / Button / FormFill). That's a cross-plugin command-injection path into a side-effecting sink.

Suggested fix: schema-validate the discriminated union per `kind` before dispatch (here at the transport boundary and/or at the executor sink in #1064).

### 2. No protocol version on the envelope (moderate)

The envelope carries `source` / `senderId` / `timestamp` but no `protocolVersion`. Two tabs on different plugin builds (a stale tab plus an upgrade) can't negotiate, and an unknown `kind` is silently dropped. Cheap to add a version field now; awkward to retrofit once the protocol has shipped. Worth a conscious decision: add it, or explicitly accept the same-build assumption.

### 3. `CrossTabPayload` is triple-maintained and has already drifted (nit, but telling)

Each message type must be added in three places — the interface, the `CrossTabMessage` union, and the `CrossTabPayload` union — and the payload union has already drifted in ordering (`StepProgress` / `StepComplete`) relative to the message union. Derive it instead:

```ts
type CrossTabPayload = CrossTabMessage extends infer M ? Omit<M, keyof CrossTabEnvelope> : never;
```

### Smaller nits

- `RemoteRequirementError` is a structural mirror of `CheckResultError` / `RequirementsCheckResult` and can silently diverge — add a compile-time assignability assertion (`satisfies`) to catch drift.
- `handleMessage`'s `forEach`: a single throwing listener blocks the remaining listeners — wrap each in try/catch.
- `post()` silently no-ops before `start()`; a pre-start post is lost with no signal. Confirm that's intentional.

Tracking IDs `F-1056-1..6` in the stack review ledger.

moxious:
**Fix-pass outcome (re-stacked, gate-green, signed, pushed `a94618f2`)**

**T1 PART A — the security spine.** Added `validateCrossTabMessage`, a per-kind discriminated-union gate that checks the envelope *and* the kind-specific shape before any dispatch. Also derived `CrossTabPayload` distributively (removes a hand-maintained drift site) and wrapped listeners in try/catch. Decided *against* a `protocolVersion` field — the same-build assumption is documented instead.

leslie-heinzen (approved):

PR #1057, opened 2026-06-17
test(storage): add characterization tests for untested storage modules

moxious:
## Summary

Adds tripwire-level characterization tests for five storage modules that previously had no tests, so the rest of this PR stack lands on a real safety net instead of dead reckoning. No production code changes.

## What to look at

- **`src/lib/user-storage.hook.test.ts`** — pins the `useUserStorage` hook's backend selection (hybrid vs localStorage) and the `setGlobalStorage` bridge that non-React storage helpers depend on. This is the most load-bearing of the five — it's the seam that the user-storage decomposition (later in the roadmap) sits on.
- **`src/integrations/coda/terminal-storage.test.ts`** — covers height clamping (`MIN_HEIGHT`/`MAX_HEIGHT`), scrollback truncation (`MAX_SCROLLBACK_SIZE`), and the localStorage/sessionStorage split. Includes failure-swallowing tests so quota / storage-unavailable paths can't regress silently.
- **`src/components/block-editor/hooks/useBlockPersistence.test.ts`** — pins the 1000 ms debounced autosave, `autoSavePaused` gating, blockIds round-trip, and the v2 version-mismatch warning.
- **`src/components/block-editor/hooks/useRecordingPersistence.test.ts`** — pins the mount-time `onRestore`-once guard (Strict Mode safe) and the "save no-ops when not recording" branch.
- **`src/docs-retrieval/content-fetcher.bundled-localstorage.test.ts`** — covers the `bundled:wysiwyg-preview` and `bundled:e2e-test` paths through `fetchContent`, including malformed-JSON title fallback and key isolation between the two.

## Why

`USER_STORAGE_ANALYSIS.md` flagged five modules with zero tests and load-bearing persistence logic. Each was a silent-regression hazard for any future refactor — the kind that ships fine in CI and breaks in production weeks later when somebody changes a key shape. PR #1055 locked the key catalog; this PR locks the behavior behind each key.

PrTester component tests are deliberately deferred — its storage keys are already locked by `storage-keys.test.ts`, and the cleaner surface for testing PrTester's storage behavior is the `usePersistedLocalState` hook that lands in PR3 of this stack.

## How to verify

1. Run the new test files in isolation and confirm they pass:
   ```
   npx jest src/lib/user-storage.hook.test.ts src/integrations/coda/terminal-storage.test.ts src/components/block-editor/hooks/useBlockPersistence.test.ts src/components/block-editor/hooks/useRecordingPersistence.test.ts src/docs-retrieval/content-fetcher.bundled-localstorage.test.ts
   ```
   Expect 64 tests passing.
2. Confirm no production source files were touched in this PR — every change is under a `*.test.ts` path.

## Breaking changes

None. This change is additive: no production code modified, no existing tests changed.

## Out of scope

- PrTester component tests — deferred to PR3 (`usePersistedLocalState`) where the right test surface lives.
- The `useUserStorage` hook tests deliberately do not cover the `syncFromGrafanaStorage` queue + conflict-resolution paths — those are already pinned in [`user-storage.plumbing.test.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/b6cff99a15d7cf9829f76983c194c5db82b2c12c/src/lib/user-storage.plumbing.test.ts) (added in PR #1055).

Refs `.cursor/local/USER_STORAGE.md`, `.cursor/local/USER_STORAGE_ANALYSIS.md`

### DIFF
--- a/src/components/block-editor/hooks/useBlockPersistence.test.ts
+++ b/src/components/block-editor/hooks/useBlockPersistence.test.ts
@@ -1,0 +1,226 @@
+/**
+ * Characterization / tripwire tests for useBlockPersistence.
+ *
+ * Pins behavior of the block-editor autosave/restore loop before any
+ * refactor that routes it through `UserStorage`:
+ *   - 1000 ms debounce on guide changes
+ *   - per-guide-snapshot dedup via `lastGuideRef`
+ *   - `autoSavePaused` halts saves but preserves the persisted snapshot
+ *   - mount-time `onLoad` receives both `guide` and `blockIds`
+ *   - version mismatch logs a warning but still returns the parsed guide
+ *   - clear / hasSavedGuide / getLastSaveTime contracts
+ */
+import { act, renderHook } from '@testing-library/react';
+
+import { StorageKeys } from '../../../lib/storage-keys';
+import type { JsonGuide } from '../types';
+
+import { useBlockPersistence } from './useBlockPersistence';
+
+const STORAGE_KEY = StorageKeys.BLOCK_EDITOR_STATE;
+
+function guide(title = 'g'): JsonGuide {
+  return {
+    title,
+    sections: [],
+  } as unknown as JsonGuide;
+}
+
+beforeEach(() => {
+  localStorage.clear();
+  jest.useFakeTimers();
+});
+
+afterEach(() => {
+  jest.useRealTimers();
+});
+
+describe('useBlockPersistence — debounced auto-save', () => {
+  it('writes to localStorage exactly once after the 1000 ms debounce', () => {
+    const { rerender } = renderHook(({ g }) => useBlockPersistence({ guide: g }), {
+      initialProps: { g: guide('a') },
+    });
+
+    // No write yet (initial guide ≠ lastGuideRef but timer hasn't fired).
+    expect(localStorage.getItem(STORAGE_KEY)).toBeNull();
+
+    rerender({ g: guide('b') });
+    rerender({ g: guide('c') });
+
+    // Still no write until debounce elapses.
+    expect(localStorage.getItem(STORAGE_KEY)).toBeNull();
+
+    act(() => {
+      jest.advanceTimersByTime(1000);
+    });
+
+    const stored = JSON.parse(localStorage.getItem(STORAGE_KEY)!);
+    expect(stored.guide.title).toBe('c');
+    expect(stored.version).toBe(2);
+  });
+
+  it('does not write when autoSavePaused is true', () => {
+    const { rerender } = renderHook(
+      ({ g, paused }) => useBlockPersistence({ guide: g, autoSavePaused: paused }),
+      { initialProps: { g: guide('a'), paused: true } }
+    );
+
+    rerender({ g: guide('b'), paused: true });
+
+    act(() => {
+      jest.advanceTimersByTime(2000);
+    });
+
+    expect(localStorage.getItem(STORAGE_KEY)).toBeNull();
+  });
+
+  it('does not write when autoSave is false', () => {
+    const { rerender } = renderHook(({ g }) => useBlockPersistence({ guide: g, autoSave: false }), {
+      initialProps: { g: guide('a') },
+    });
+
+    rerender({ g: guide('b') });
+
+    act(() => {
+      jest.advanceTimersByTime(2000);
+    });
+
+    expect(localStorage.getItem(STORAGE_KEY)).toBeNull();
+  });
+
+  it('stores blockIds alongside guide when provided', () => {
+    const { rerender } = renderHook(
+      ({ g, ids }: { g: JsonGuide; ids: string[] }) => useBlockPersistence({ guide: g, blockIds: ids }),
+      { initialProps: { g: guide('a'), ids: ['b1', 'b2'] } }
+    );
+
+    rerender({ g: guide('b'), ids: ['b1', 'b2', 'b3'] });
+
+    act(() => {
+      jest.advanceTimersByTime(1000);
+    });
+
+    const stored = JSON.parse(localStorage.getItem(STORAGE_KEY)!);
+    expect(stored.blockIds).toEqual(['b1', 'b2', 'b3']);
+  });
+});
+
+describe('useBlockPersistence — mount-time restore via onLoad', () => {
+  it('calls onLoad with stored guide AND blockIds when both are present', () => {
+    localStorage.setItem(
+      STORAGE_KEY,
+      JSON.stringify({
+        guide: guide('restored'),
+        blockIds: ['b1', 'b2'],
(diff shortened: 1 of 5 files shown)


moxious:
## Round 1 automated review

**Classification:** `tests-only` — five new `*.test.ts` files, zero production code changes, +946/-0.

**Activated concerns:** `testing-and-verification` (AO), `correctness-and-reliability` (AO, test-correctness only), `reversibility-and-one-way-door` (AO — storage keys touched in tests), `cross-cutting-architecture` (AO), `security` (AO — passive, no new sinks), `coda-terminal` (sub — `terminal-storage.test.ts`), `docs-retrieval-and-rendering` (sub — `content-fetcher.bundled-localstorage.test.ts`).

**Verification:** ran `npx jest` against all five files at the PR head — 64/64 pass in ~0.84s. Cross-checked each test's expected strings (`'Preview: WYSIWYG Guide'`, `'No preview content available'`, `'No E2E test content available'`, `'E2E Test Guide'`) against the production source in `content-fetcher.ts` at the head ref — all match. Verified `__resetSyncedForTests`, `setGlobalStorage`, `createLocalStorage`, `createUserStorage`, `useUserStorage` all exist in `src/lib/user-storage.ts` at the head ref.

### Overall

This is a clean tripwire-tests PR. The five files cover exactly the modules `USER_STORAGE_ANALYSIS.md` flagged as zero-coverage and load-bearing for the downstream refactor. Tests are small, focused, and pin behavior (not implementation). They will catch the regressions they're designed to catch — backend selection in `useUserStorage`, height clamping and scrollback truncation in `terminal-storage`, debounce window and v2 mismatch warning in `useBlockPersistence`, mount-once restore in `useRecordingPersistence`, and the malformed-JSON fallback titles in `content-fetcher`.

The only findings are minor — a couple of small characterization gaps that, if filled, would make the safety net a touch tighter before the rest of the stack lands. None are blocking.

---

### Findings (ordered by severity, then confidence)

#### 1. `low` · `testing-and-verification` · QC2 — "Strict Mode safe" comment isn't actually verified by the test

`src/components/block-editor/hooks/useRecordingPersistence.test.ts:240–249, 364–383`

The file-level docstring promises *"mount-time `onRestore` fires exactly once (Strict Mode safe)"*, and the test case is titled `'hasRestoredRef gate prevents double restore on rerender'`. But the test only rerenders the same hook instance with different props — it does not wrap in `<React.StrictMode>`, which is the actual condition where double-mount-effects happen. A rerender by itself does not re-run a `[load, onRestore]`-deps effect (deps are stable). So the test passes today even if the `hasRestoredRef` gate is removed, because the deps trick — not the ref — is what suppresses re-fires.

**Why it matters:** the `hasRestoredRef` guard is the load-bearing protection against Strict Mode's intentional double-mount. The test claims to pin that protection but actually pins something weaker (effect-deps stability). A future refactor that changes the deps (or moves restore to a non-`useEffect` site) could remove `hasRestoredRef` and still pass this test, then break in Strict Mode.

**Reversibility:** `reversible` — tests-only.

**Suggested action:** either (a) rename the test to "does not double-restore across rerenders" and adjust the docstring, or (b) wrap with `{ wrapper: StrictMode }` so the test actually exercises the guard. Option (b) is the higher-leverage change because that's the property the comment claims.

#### 2. `low` · `testing-and-verification` · coverage gap — missing pin on the "no-change save notification" branch in `useBlockPersistence`

`src/components/block-editor/hooks/useBlockPersistence.ts:163–167` is a load-bearing branch:

```ts
if (currentGuideStr === lastGuideRef.current) {
  onSave?.();  // clears isDirty even when nothing changed
  return;
}
```

The PR's tests don't cover this branch. A refactor that drops the `onSave?.()` call here (easy to do by accident when porting to a generic `usePersistedLocalState`) would silently leave `isDirty` set in the block editor — the kind of "ships fine in CI, breaks in production weeks later" failure the PR is specifically designed to prevent.

**Reversibility:** `reversible` — tests-only.

**Suggested action:** add one assertion: trigger an effect run where the serialized guide is unchanged and assert `onSave` was called. Two lines plus a `jest.fn()`.

#### 3. `low` · `correctness-and-reliability` · QC2 — misleading comment about module re-import

`src/lib/user-storage.hook.test.ts:950–952`

```ts
// Defer the import so the mocked module is fresh per test.
const { createUserStorage } = await import('./user-storage');
```

`jest.mock` at the top of the file is set up at module-evaluation time, and `await import()` returns the same cached module instance — it does not give you a "fresh" module per test. The test still works (because `setGlobalStorage` is set via the rendered hook earlier in the test, and `createUserStorage()` reads from that module-level global). But the comment misdescribes what it does, which is exactly the QC8 "stale-in-waiting" shape — the next reader will believe the import is fresh and act on that belief.

**Reversibility:** `reversible` — comment-only.

**Suggested action:** either delete the comment (the static import at the top of the file already covers the test path the same way) or rewrite it to "use the same module instance — `createUserStorage()` reads the module-level `globalStorageInstance` that the rendered hook installed above". Per QC8, deletion is preferable.

---

### Adversarial verification

Two `medium`-or-higher candidate findings were considered and dropped after skeptic review:

- *Considered:* "Test at line 45–66 of `useBlockPersistence.test.ts` may rely on undocumented order between effect re-run + setTimeout reset." **Dropped** — verified against `useBlockPersistence.ts:170–184`: each rerender's effect clears the prior timeout and schedules a new one, and `save` closes over the latest `guide` via `useCallback` deps. The test is correct.
- *Considered:* "Monkey-patching `Storage.prototype.setItem` may leak between tests." **Dropped** — the `try/finally` blocks in both monkey-patch tests restore the original `Storage.prototype.{setItem,getItem}` even on assertion failure. No leak.

---

### Tech debt

Tech debt: no high-confidence patterns found in the changed files. Scope is test-only; per `.cursor/skills/techdebt/SKILL.md` §3, suggestive findings are suppressed on tests-only changes. The three findings above are the only ones worth surfacing, and they are characterization-quality issues, not debt.

---

### Doc-drift check

No drift detected. This PR introduces no new subsystems, scripts, skills, plugin routes, feature flags, or architecture changes — only behavioral pins on existing modules. `AGENTS.md`, `CLAUDE.md`, and `.cursor/rules/` need no updates.

---

### Reversibility summary

All changes are `reversible`. Reverting the PR removes only test files and restores the previous (zero-test) state of the five modules. No persisted state, no telemetry, no contract surface touched.

---

### Coverage note

`coverage_confidence: high`. The PR's center of gravity (storage characterization) is well-modeled by `reversibility-and-one-way-door` and `testing-and-verification`, both of which are AO concerns. Subsystem coverage from `coda-terminal` and `docs-retrieval-and-rendering` triggered cleanly on path + keyword signals.

leslie-heinzen (approved):
:shipit:

PR #1059, opened 2026-06-17
refactor(storage): extract three shared storage utilities

moxious:
## Summary

Extracts three shared storage utilities that deduplicate patterns flagged as B1 (near-duplicate function) and B4 (convergent re-bugging) in `USER_STORAGE_ANALYSIS.md`. The two prefix-sweep helpers and the extension-sidebar parser are prerequisites for the next PR in this stack (PR4 / badge eviction); the `usePersistedLocalState` hook is a UI-prefs convenience that retires hand-rolled state-from-localStorage boilerplate in five components.

## What to look at

- **[`src/lib/storage/key-utils.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/7f91a8cec36114adec2e7cc1fda6b1f567dbfafe/src/lib/storage/key-utils.ts)** — `collectKeysByPrefix(storage, prefix)` / `clearKeysByPrefix(storage, prefix)`. Iterates descending so callers can `removeItem` during the sweep without skipping siblings. Migrates `highlighted-guide-utils.clearHighlightedGuideMarkers`, `experiment-utils.resetExperimentState`, and three sites in `experiment-debug` (`listExposureMarkers`, `clearCache`, `showCache`).
- **[`src/lib/storage/extension-sidebar.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/7f91a8cec36114adec2e7cc1fda6b1f567dbfafe/src/lib/storage/extension-sidebar.ts)** — `parseExtensionSidebarDocked()` plus four narrow helpers (`isExtensionSidebarOwnedByOther`, `isExtensionSidebarOwnedByPathfinder`, `isExtensionSidebarInUse`, `clearExtensionSidebarDocked`). One canonical parser for the `grafana.navigation.extensionSidebarDocked` key, which previously had three subtly-different parse paths (`module.tsx` restore, `module.tsx` suggest-reject, `experiment-utils.isExtensionSidebarOwnedByOther`) — the JSON-then-legacy-string fallback varied between them. The legacy call sites delegate.
- **[`src/hooks/usePersistedLocalState.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/7f91a8cec36114adec2e7cc1fda6b1f567dbfafe/src/hooks/usePersistedLocalState.ts)** — generic + `usePersistedString` + `usePersistedBoolean`. **Note the first-effect-run skip**: the write-on-change effect deliberately skips the initial render so the default value never gets persisted unprompted. Without that guard, `ConditionChipsField` and `HealthStatusBar` would start writing `'false'` to localStorage just because the component mounted, which would change the observable contract for those keys — `ConditionChipsField.test.tsx:197` explicitly tests the "absent until user toggles" behavior.
- **Component migrations**: `UrlTester.tsx`, `SelectorDebugPanel.tsx`, `HealthStatusBar.tsx`, `ConditionChipsField.tsx`. The `hooks/index.ts` barrel adds the three exports so consumers import from `'../../hooks'`, not the internal path.

## Why

`USER_STORAGE_ANALYSIS.md` identified these three patterns as cross-cutting debt that was blocking cleaner extraction in user-storage and learning-paths. The prefix-sweep utility is a direct prerequisite for PR4 — the badge coordinator's deletion of `learningProgressStorage.markGuideCompleted` from `user-storage.ts` doesn't itself need the helper, but the user-storage decomposition that follows will. The extension-sidebar parser closes a real B4 bug class: the three parsers previously disagreed on whether to treat a legacy plain-string value as a `pluginId`.

## How to verify

1. Open the dev panel (with dev mode enabled). Type a URL into URL Tester, refresh the page — confirm the URL persists. Toggle PR Tester and URL Tester sections open / closed, refresh, confirm state restores.
2. Open the block editor. Toggle the health-status bar expanded / collapsed, refresh, confirm restore. In a condition field (e.g., requirements), click "View raw" → "Use chip editor", refresh, confirm the chosen mode restores.
3. With another plugin currently docked in the Grafana extension sidebar, dispatch a `pathfinder-suggest` event and confirm it's rejected with `reason: 'sidebar_in_use'`. Then with Pathfinder docked, confirm a fresh suggest succeeds.

## Breaking changes

None. All migrations preserve the existing storage key strings and the existing read/write contracts. The `usePersistedLocalState` hook intentionally does NOT write the default value on mount, preserving the call-site behavior where `localStorage.getItem(KEY)` returns `null` until the user actually toggles the preference.

## Out of scope

- **PrTester migration is deferred.** Its storage is a multi-key coupled cache (PR URL + fetched files JSON + the URL that the files were fetched for — all three must invalidate atomically when the URL changes). That fits a typed multi-key reducer, not a single-key hook. Will be a separate follow-up.
- **`user-storage.ts` internal prefix sweeps not migrated.** Six existing inline sweeps in `clearAllForContent`, `clearAll`, `hasProgress`, `countAllCompleted`, `countAllAcknowledged` are intentionally left for the user-storage decomposition PR where they'll be revisited alongside the broader L1/L2 split.

### DIFF
--- a/src/hooks/usePersistedLocalState.ts
+++ b/src/hooks/usePersistedLocalState.ts
@@ -1,0 +1,118 @@
+/**
+ * `usePersistedLocalState` — a small `useState` wrapper that lazy-reads its
+ * initial value from `localStorage` and writes through to localStorage on
+ * every change.
+ *
+ * Replaces the hand-rolled `useState(() => localStorage.getItem(KEY) || x)`
+ * + `useEffect(() => localStorage.setItem(KEY, ...))` pattern that was
+ * duplicated across the dev/UI-pref components surveyed in
+ * `.cursor/local/USER_STORAGE_ANALYSIS.md`:
+ *   - PrTester, UrlTester, SelectorDebugPanel (devtool prefs)
+ *   - HealthStatusBar, ConditionChipsField (block-editor UI prefs)
+ *
+ * Designed for ephemeral *UI preferences* — anything where a localStorage
+ * read miss should fall back to a fixed default without surfacing an error.
+ * Quota and parse failures are deliberately swallowed in line with the
+ * existing call-site contract; UI state is local-only and not synced to
+ * Grafana user storage.
+ *
+ * Three flavors are provided so callers don't have to write the same
+ * deserialize/serialize boilerplate every time:
+ *
+ *   - `usePersistedString`    → string values, stored as-is
+ *   - `usePersistedBoolean`   → boolean values, stored as `"true"` / `"false"`
+ *   - `usePersistedLocalState` → arbitrary `T`, callers provide
+ *                                serialize / deserialize callbacks
+ */
+import { useCallback, useEffect, useRef, useState, type Dispatch, type SetStateAction } from 'react';
+
+export interface UsePersistedLocalStateOptions<T> {
+  /** Storage key — must come from `StorageKeys` for new call sites. */
+  key: string;
+  /** Default returned when the key is absent or deserialize fails. */
+  defaultValue: T;
+  /** Convert the storage string into `T`. Throwing falls back to `defaultValue`. */
+  deserialize: (raw: string) => T;
+  /** Convert `T` into the string to store. */
+  serialize: (value: T) => string;
+}
+
+/**
+ * Generic persisted-state hook. Reads `key` once on mount via the lazy
+ * initializer; writes on every state change via an effect. Returns the
+ * same `[value, setValue]` tuple as `useState`.
+ */
+export function usePersistedLocalState<T>({
+  key,
+  defaultValue,
+  deserialize,
+  serialize,
+}: UsePersistedLocalStateOptions<T>): [T, Dispatch<SetStateAction<T>>] {
+  const [value, setValue] = useState<T>(() => {
+    try {
+      const stored = localStorage.getItem(key);
+      if (stored === null) {
+        return defaultValue;
+      }
+      return deserialize(stored);
+    } catch {
+      return defaultValue;
+    }
+  });
+
+  // Skip the write on the first effect run. The mount write would otherwise
+  // turn the default value into a persisted value, changing observable
+  // behavior for call sites that previously only wrote on user-driven change.
+  const isFirstEffectRun = useRef(true);
+
+  useEffect(() => {
+    if (isFirstEffectRun.current) {
+      isFirstEffectRun.current = false;
+      return;
+    }
+    try {
+      localStorage.setItem(key, serialize(value));
+    } catch {
+      // Storage unavailable / quota — UI preferences degrade silently.
+    }
+  }, [key, value, serialize]);
+
+  return [value, setValue];
+}
+
+/**
+ * String-valued persisted state. The default is `''` unless overridden.
+ */
+export function usePersistedString(
+  key: string,
+  defaultValue = ''
+): [string, Dispatch<SetStateAction<string>>] {
+  const deserialize = useCallback((raw: string) => raw, []);
+  const serialize = useCallback((value: string) => value, []);
+  return usePersistedLocalState({
+    key,
+    defaultValue,
+    deserialize,
+    serialize,
+  });
+}
+
+/**
+ * Boolean-valued persisted state. Stored as `"true"` / `"false"` strings
+ * for compatibility with existing call sites. Only `"true"` is treated as
+ * `true` on read — any other value defaults to `false` (or `defaultValue`
+ * when the key is absent).
+ */
+export function usePersistedBoolean(
+  key: string,
+  defaultValue = false
+): [boolean, Dispatch<SetStateAction<boolean>>] {
+  const deserialize = useCallback((raw: string) => raw === 'true', []);
+  const serialize = useCallback((value: boolean) => String(value), []);
+  return usePersistedLocalState({
+    key,
+    defaultValue,
+    deserialize,
+    serialize,
+  });
+}
--- a/src/hooks/usePersistedLocalState.ts
+++ b/src/hooks/usePersistedLocalState.ts
@@ -83,10 +83,7 @@ export function usePersistedLocalState<T>({
 /**
  * String-valued persisted state. The default is `''` unless overridden.
  */
-export function usePersistedString(
-  key: string,
-  defaultValue = ''
-): [string, Dispatch<SetStateAction<string>>] {
+export function usePersistedString(key: string, defaultValue = ''): [string, Dispatch<SetStateAction<string>>] {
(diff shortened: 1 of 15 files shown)


moxious:
## Round 1 automated review

**Classification**: `mixed` — `product-runtime` (touches `module.tsx` sidebar-restore behavior and five UI components) + `contracts-and-schemas` (introduces a new public surface in `lib/storage/` and `hooks/`). Reviewed against base `refactor/storage/2-progress-cleared-event` (PR-scoped diff only).

**Activated concerns**: `correctness-and-reliability`, `testing-and-verification`, `reversibility-and-one-way-door`, `cross-cutting-architecture`, `security` (always-on; no triggers fired).

**Coverage**: high. This PR is a tight, well-bounded extraction PR with good test scaffolding around each new utility. One subtle behavior delta found in `module.tsx`'s restore path; everything else looks clean.

---

### Findings

#### 1. Sidebar-restore loses a legacy-string fallback case (low severity, high confidence)

- **Concern**: `reversibility-and-one-way-door` / `cross-cutting-architecture`
- **Where**: `src/module.tsx:80` → `isExtensionSidebarOwnedByPathfinder()` in `src/lib/storage/extension-sidebar.ts:96-104`
- **Problem**: The old inline check in `module.tsx` recognized Pathfinder ownership in four cases for a docked value:
  1. JSON `pluginId === pluginJson.id`
  2. JSON `componentTitle === 'Interactive learning'`
  3. Legacy bare string `=== pluginJson.id`
  4. **Legacy bare string `=== 'Interactive learning'`** ← dropped

  The new `parseExtensionSidebarDocked()` puts a legacy bare string into `pluginId` only. `isExtensionSidebarOwnedByPathfinder` then checks `docked.pluginId === myPluginId || docked.componentTitle === componentTitleMatch`. Case 4 — older Grafana storing just the bare string `'Interactive learning'` as the value — would have matched the old inline code's componentTitle clause but is not recognized by the new helper (it lands in `pluginId`, not `componentTitle`).
- **Why it matters**: A user on an old enough Grafana version that wrote the componentTitle as a bare string would lose the browser-restore path on the next reload — Pathfinder would not call `setPendingOpenSource('browser_restore', 'restore')`. The PR description explicitly claims "All migrations preserve the existing storage key strings and the existing read/write contracts." This drift contradicts that.
- **Reversibility**: `reversible` — no persisted state shape changes; reverting restores the prior behavior.
- **Confidence in real-world impact**: low. This shape is almost certainly extinct on supported Grafana versions; the original fallback may have been defensive belt-and-suspenders. Worth a quick decision either way.
- **Suggested action**: Either (a) intentionally drop the case and note it in the PR body, or (b) extend `isExtensionSidebarOwnedByPathfinder` to also match `docked.pluginId === componentTitleMatch` when no JSON shape was found, and add a regression test in `extension-sidebar.test.ts` (alongside the existing "legacy plain string as pluginId" case).

#### 2. UrlTester no longer writes the empty default to localStorage on mount (low severity, medium confidence)

- **Concern**: `correctness-and-reliability` / `reversibility-and-one-way-door`
- **Where**: `src/components/UrlTester/UrlTester.tsx:74` (via `usePersistedString`) and `src/hooks/usePersistedLocalState.ts:50-58` (first-effect-run skip)
- **Problem**: The previous `UrlTester` wrote `testUrl` to localStorage on every render via `useEffect`, including the initial empty default. The new `usePersistedLocalState` deliberately skips the first effect run (with a clear, documented rationale that this is what makes `ConditionChipsField`/`HealthStatusBar` keep their "absent until toggle" contract — and the PR body calls this out as load-bearing). The net effect for UrlTester is: on a fresh install, `localStorage.getItem(StorageKeys.DEVTOOLS_URL_TESTER_URL)` now stays `null` until the user actually types something, where previously it would become `''` after first mount.
- **Why it matters**: This is almost certainly an improvement (the new behavior is more honest and matches the explicit invariant the hook is designed to enforce), and no external code reads this key. But the PR body's claim "All migrations preserve the existing storage key strings and the existing read/write contracts" is technically inaccurate for this one site. If the PR description gets used later as evidence for "this was a pure refactor," that matters.
- **Reversibility**: `reversible`.
- **Suggested action**: Either add a one-liner to the PR description acknowledging the intentional drift, or accept as-is. No code change needed.

---

### Tech debt

No high-confidence patterns found in the changed files. The extraction reduces duplication across 7+ call sites and the new helpers each have focused unit tests. One observation (informational, not a finding):

- `usePersistedLocalState`'s effect depends on `serialize` identity. The typed wrappers (`usePersistedString`/`usePersistedBoolean`) memoize via `useCallback([])` so this is stable, but a direct caller of `usePersistedLocalState` who passes an inline `serialize` arrow will re-run the effect every render. The current generic-mode test fixture happens to do this. Worth a sentence in the JSDoc nudging direct callers to memoize, or — cheaper — wrap the serialize ref internally in the hook.

### Doc-drift

`.cursor/local/USER_STORAGE.md:14` already documents `src/lib/storage/` as the home for storage sub-modules, so the new files fit the existing structure without prose updates. No `AGENTS.md` / `CLAUDE.md` / `.cursor/rules/` changes are required for this PR — the architecture is unchanged, the tier-1 (`lib/`) placement matches the documented tier model, and no new public conventions are introduced.

Optionally: the inventory in `.cursor/local/USER_STORAGE.md` could mention `key-utils.ts` and `extension-sidebar.ts` alongside the existing `experiment-auto-open-storage.ts` entry, since that doc explicitly enumerates `src/lib/storage/` sub-modules. Not blocking.

---

### Summary

Clean, well-tested extraction PR with one subtle correctness drift in the sidebar-restore path (legacy-string-as-componentTitle case, very low real-world likelihood) and one undocumented but probably-correct contract change in `UrlTester`. Neither blocks merge. The shape of the new utilities — `Storage`-parameterized, key-strategy-only, no `UserStorage` coupling — is the right design for the extraction goal and lines up cleanly with the stack's later PRs.

src/hooks/usePersistedLocalState.ts
```diff
@@ -0,0 +1,112 @@
+/**
+ * `usePersistedLocalState` — a small `useState` wrapper that lazy-reads its
+ * initial value from `localStorage` and writes through to localStorage on
+ * every change.
+ *
+ * Replaces the hand-rolled `useState(() => localStorage.getItem(KEY) || x)`
+ * + `useEffect(() => localStorage.setItem(KEY, ...))` pattern that was
+ * duplicated across the dev/UI-pref components surveyed in
+ * `.cursor/local/USER_STORAGE_ANALYSIS.md`:
```
leslie-heinzen, line 9: tech debt nit: this references a local file you're using, I'm assuming? Shows up in a couple of other comments.

leslie-heinzen (approved):

PR #1062, opened 2026-06-17
feat(twotab): open an interactive guide in a controller tab

Jayclifford345:
Stacked on #1056.

Adds an **"Open in interactive window"** button to the docs content-meta toolbar (sibling of the docs / learning-journey open buttons), shown only for guides of type `interactive` (`pickControllerTabOpenAction`). It opens a same-origin `?doc=<guide>&controller=1` tab; `module.tsx` gains a matching short-circuit that mounts `GuideReaderOverlay` in controller mode.

`GuideReaderOverlay` gains a `mode` prop and `InteractiveMode` gains `'controller'`; `controller` joins the `PATHFINDER_PARAMS` parse/strip set. Clicking the visible step actions doesn't drive the live tab yet — the cross-tab emit wiring lands in the next layer.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -189,19 +189,21 @@ Excluded from tier analysis (not tiered): `test-utils/`, `cli/`, `bundled-intera
 
 **Key dependency edges** (where the load-bearing wiring lives):
 
-| Edge                                           | Why                                                            |
-| ---------------------------------------------- | -------------------------------------------------------------- |
-| `context-engine` → `docs-retrieval`            | Fetches content for the recommendations it surfaces            |
-| `docs-retrieval` → `bundled-interactives`      | Fallback when the online CDN is unavailable                    |
-| `docs-retrieval` → `package-engine`            | Resolves package manifests + content                           |
-| `docs-retrieval` → `snippet-engine`            | Inlines `snippet-ref` blocks against the CDN at parse time     |
-| `components/docs-panel` → `interactive-engine` | Executes step actions when the user clicks "Show me" / "Do it" |
-| `interactive-engine` → `requirements-manager`  | Checks prereqs before enabling / executing a step              |
-| `interactive-engine` → `lib/dom`               | Selector resolution + element detection                        |
-| `components/docs-panel` → `context-engine`     | Reads and renders recommendations                              |
-| `components/docs-panel` → `global-state`       | Sidebar, panel-mode, tab persistence                           |
-| `learning-paths` → `lib/user-storage`          | Persists progress + streak data                                |
-| `recovery` → `requirements-manager`            | Decides whether a failed requirement is auto-recoverable       |
+| Edge                                                          | Why                                                                                |
+| ------------------------------------------------------------- | ---------------------------------------------------------------------------------- |
+| `context-engine` → `docs-retrieval`                           | Fetches content for the recommendations it surfaces                                |
+| `docs-retrieval` → `bundled-interactives`                     | Fallback when the online CDN is unavailable                                        |
+| `docs-retrieval` → `package-engine`                           | Resolves package manifests + content                                               |
+| `docs-retrieval` → `snippet-engine`                           | Inlines `snippet-ref` blocks against the CDN at parse time                         |
+| `components/docs-panel` → `interactive-engine`                | Executes step actions when the user clicks "Show me" / "Do it"                     |
+| `interactive-engine` → `requirements-manager`                 | Checks prereqs before enabling / executing a step                                  |
+| `interactive-engine` → `lib/dom`                              | Selector resolution + element detection                                            |
+| `components/docs-panel` → `context-engine`                    | Reads and renders recommendations                                                  |
+| `components/docs-panel` → `global-state`                      | Sidebar, panel-mode, tab persistence                                               |
+| `learning-paths` → `lib/user-storage`                         | Persists progress + streak data                                                    |
+| `recovery` → `requirements-manager`                           | Decides whether a failed requirement is auto-recoverable                           |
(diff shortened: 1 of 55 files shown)


moxious:

# PR #1062: feat(twotab): open an interactive guide in a controller tab

**Verdict:** Request changes

## Findings

**1. [Must fix] Wrong default for `mode` prop — `GuideReaderOverlay.tsx`**

`mode = 'controller'` is a surprising default for a component named `GuideReaderOverlay`. The existing interactive surface (floating panel, kiosk) mounts this component in `'interactive'` mode. Any caller that forgets to pass `mode` — likely, because the prop is optional — silently gets controller behaviour, which suppresses the interactive step runner. The call site in `module.tsx` does pass `mode='controller'` explicitly, so the default is never exercised today; but it is a trap for every future caller. Flip the default to `'interactive'`, which is the existing/ambient meaning of the overlay.

**2. [Must fix] `tabType` is untyped in `pickControllerTabOpenAction` — `controller-tab-open-action.ts`**

The signature accepts `tabType: string | undefined`. The call site passes `activeTab.type`, which is `LearningJourneyTab['type']` — a strict union `'learning-journey' | 'docs' | 'devtools' | 'interactive' | 'editor' | undefined`. Widening to `string` at the function boundary throws away the compiler's ability to catch a future rename of `'interactive'`. Accept the narrower type directly: `tabType: LearningJourneyTab['type']`. That way the `tabType !== 'interactive'` check stays in sync with the union by construction.

**3. [Must fix] `?controller=1` is not stripped from the URL — `module.tsx` / `pathfinder-search-params.ts`**

`PATHFINDER_PARAMS` does not include `'controller'`, and `parsePathfinderDeepLink` does not return a `controller` field — the PR adds `controllerParam` by reading `params.get('controller')` directly instead of going through the shared contract. This means `?controller=1` lingers in the URL after the overlay mounts. Add `'controller'` to `PATHFINDER_PARAMS` and add a typed `controller` field to `DeepLinkParams` (and `parsePathfinderDeepLink`) to match every other param. `stripPathfinderParams` will then clean it automatically.

**4. [Must fix] The "Open" button label is ambiguous — `DocsPanelContentArea.tsx`**

The toolbar already has a `<button>` that renders `<span>Open</span>` for "Open this page in new tab" (`pickGrafanaDocsOpenAction`). The new controller button renders an identical `<span>Open</span>` right next to it. Two buttons with the same visible text and the same icon (`external-link-alt`) in the same toolbar are indistinguishable for sighted users, screen readers, and automated tests. The `aria-label` differs, but that only helps assistive tech. Use a distinct label — "Open interactive window", "Interactive window", or similar — and update the i18n key accordingly.

**5. [Should fix] Double-ID guard is a non-atomic read-then-write — `module.tsx`**

```ts
if (!document.getElementById('pathfinder-controller-root')) {
  import('./components/guide-reader/GuideReaderOverlay')
    .then(async ({ GuideReaderOverlay }) => {
      if (document.getElementById('pathfinder-controller-root')) { return; }
      // ... appendChild + render
    })
```

The outer check and the inner check are separated by an `await`-able chunk load. In practice `plugin.init` runs once synchronously so a true race is impossible, but the defensive inner check creates the false impression it guards against concurrent callers. The pattern is consistent with how `pathfinder-kiosk-root` and `pathfinder-floating-root` are guarded elsewhere in `module.tsx`, so this is an inherited idiom and not something to fight alone — but document why the outer check is sufficient (it is, because `init` runs once) or move the `getElementById` inside the `.then()` only, removing the outer one. As-is it looks like a half-applied race guard and will confuse a future reader.

**6. [QC violation] File-level docstring on a small module — `controller-tab-open-action.ts`**

Based on the diff, the file is small (one exported function, one type). If there is a multi-line module preamble comment on it, that falls under QC8-shape-8 (whole-file docstring on a small module). Trim to a single-line or remove entirely — the function name and its return type already describe the contract.

moxious:

**What's good:** `controller` is added to the strict `PATHFINDER_PARAMS` allowlist and parsed as `=== '1'` (follows the existing deep-link pattern); the trigger button uses `window.open(url, '_blank', 'noopener,noreferrer')` with analytics; the overlay is lazy-imported for code-splitting.

### 1. `?controller=1` entry skips the enablement gate (moderate)

In `module.tsx`, the `if (docsParam && controllerParam)` branch returns early **before** `shouldMountSidebar(pathfinderEnabled, …)`, so it never consults the enablement gate. A shared or bookmarked controller URL will mount the overlay even when Pathfinder is disabled for the user. Worth deciding whether the controller entry should respect enablement. (Related: the live-tab executor install in #1064 has the same gap — see that PR.)

### 2. `pathfinder-controller-root` guard is not race-safe (low / TOCTOU)

The idempotency guard `if (!document.getElementById('pathfinder-controller-root'))` is checked, then two `await`s follow (`import(...)`, `createCompatRoot`) before the container is appended. Two concurrent `plugin.init` calls could both pass the guard and append duplicate roots/overlays. Claim the id synchronously (create + append the container) before awaiting, or use a synchronous flag.

### 3. `mode = 'controller'` default (nit)

The default is duplicated in both the `GuideReaderOverlay` and `GuideReaderInner` signatures, and it defaults to the *privileged* mode while the context's own default is `'interactive'`. Single-source the default and document why the privileged one is correct here.

Tracking IDs `F-1062-1..3` in the stack review ledger.

moxious:
**Fix-pass outcome (re-stacked, gate-green, signed, pushed `3873f57a`)**

Gated the `?controller=1` entry on `pathfinderEnabled` (same policy as the executor install gate), narrowed the `tabType` param back to its union, made the controller-root id claim atomic, added the missing i18n key, and dropped the privileged `'controller'` default in favor of `'interactive'`.

leslie-heinzen (approved):

PR #1063, opened 2026-06-17
feat(twotab): drive the live tab from controller-mode step actions

Jayclifford345:
Stacked on #1062. Wire the controller overlay to emit step-commands instead of executing them locally.

A new global-state `ControllerChannelProvider` owns a `CrossTabTransport` and is mounted by `GuideReaderOverlay` only in controller mode. In controller mode, the Show me / Do it handlers post a step-command over the channel and skip local execution; completion is marked optimistically (persist + `onStepComplete` + `onComplete`) so the controller reflects progress immediately.

A controller tab drives a *different* Grafana tab, so requirements that probe this tab's DOM/URL can't be evaluated here. `useStepChecker` strips tab-local requirements (`exists-reftarget`, `navmenu-open`, `on-page`, `form-valid`) in controller mode and lets the live tab enforce them, while session/permission requirements (`is-admin`, `has-datasources`, ...) still run so genuine failures surface. This covers every step type that gates through `useStepChecker`.

The transport degrades to a no-op where `BroadcastChannel` is unavailable. The live tab doesn't consume these commands yet — that's the next layer (live-tab executor).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -189,19 +189,21 @@ Excluded from tier analysis (not tiered): `test-utils/`, `cli/`, `bundled-intera
 
 **Key dependency edges** (where the load-bearing wiring lives):
 
-| Edge                                           | Why                                                            |
-| ---------------------------------------------- | -------------------------------------------------------------- |
-| `context-engine` → `docs-retrieval`            | Fetches content for the recommendations it surfaces            |
-| `docs-retrieval` → `bundled-interactives`      | Fallback when the online CDN is unavailable                    |
-| `docs-retrieval` → `package-engine`            | Resolves package manifests + content                           |
-| `docs-retrieval` → `snippet-engine`            | Inlines `snippet-ref` blocks against the CDN at parse time     |
-| `components/docs-panel` → `interactive-engine` | Executes step actions when the user clicks "Show me" / "Do it" |
-| `interactive-engine` → `requirements-manager`  | Checks prereqs before enabling / executing a step              |
-| `interactive-engine` → `lib/dom`               | Selector resolution + element detection                        |
-| `components/docs-panel` → `context-engine`     | Reads and renders recommendations                              |
-| `components/docs-panel` → `global-state`       | Sidebar, panel-mode, tab persistence                           |
-| `learning-paths` → `lib/user-storage`          | Persists progress + streak data                                |
-| `recovery` → `requirements-manager`            | Decides whether a failed requirement is auto-recoverable       |
+| Edge                                                          | Why                                                                                |
+| ------------------------------------------------------------- | ---------------------------------------------------------------------------------- |
+| `context-engine` → `docs-retrieval`                           | Fetches content for the recommendations it surfaces                                |
+| `docs-retrieval` → `bundled-interactives`                     | Fallback when the online CDN is unavailable                                        |
+| `docs-retrieval` → `package-engine`                           | Resolves package manifests + content                                               |
+| `docs-retrieval` → `snippet-engine`                           | Inlines `snippet-ref` blocks against the CDN at parse time                         |
+| `components/docs-panel` → `interactive-engine`                | Executes step actions when the user clicks "Show me" / "Do it"                     |
+| `interactive-engine` → `requirements-manager`                 | Checks prereqs before enabling / executing a step                                  |
+| `interactive-engine` → `lib/dom`                              | Selector resolution + element detection                                            |
+| `components/docs-panel` → `context-engine`                    | Reads and renders recommendations                                                  |
+| `components/docs-panel` → `global-state`                      | Sidebar, panel-mode, tab persistence                                               |
+| `learning-paths` → `lib/user-storage`                         | Persists progress + streak data                                                    |
+| `recovery` → `requirements-manager`                           | Decides whether a failed requirement is auto-recoverable                           |
+| `global-state/controller-channel` → `lib/cross-tab-transport` | BroadcastChannel link between a popped-out controller tab and the live Grafana tab |
+| `integrations/cross-tab` → `interactive-engine`               | Live tab replays a controller tab's step commands against its own DOM              |
 
 For per-subsystem entry points, public surfaces, and key files, load `.cursor/rules/systemPatterns.mdc`.
 
--- a/docs/developer/CONTEXT_INDEX.md
+++ b/docs/developer/CONTEXT_INDEX.md
@@ -56,14 +56,15 @@ Load these files **only when working in the relevant domain**.
 
 ## Dev mode, local dev, live sessions
 
-| File                       | When to load                                            | Auto-triggered by globs        |
-| -------------------------- | ------------------------------------------------------- | ------------------------------ |
-| `DEV_MODE.md`              | Dev mode configuration and debugging tools              | `src/utils/dev-mode.ts`        |
-| `LOCAL_DEV.md`             | Local development setup, prerequisites, Docker workflow | --                             |
-| `LIVE_SESSIONS.md`         | Live sessions feature (WebRTC, PeerJS)                  | `src/components/LiveSession/*` |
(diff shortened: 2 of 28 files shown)


moxious:

# PR #1063: feat(twotab): drive the live tab from controller-mode step actions

**Verdict:** Request changes

## Findings

**1. [High] `handleDoAction` optimistic completion is inconsistent with `handleShowAction`**

`handleShowAction` guards `persistCompletion()` behind `if (!doIt)` — matching the local-execution path exactly. `handleDoAction` calls `persistCompletion()` unconditionally in controller mode, which is correct for a do-only step but also means a step that has both Show me and Do it will persist on the Show me click (via `handleShowAction`'s `!doIt` = false path skips it) and then again on Do it. The local-execution path avoids double-persisting because completion state is idempotent, but the controller path produces two `onStepComplete` + `onComplete` calls for a show+do step when Do it is clicked — the show post fires (no completion), then Do it fires and calls all three callbacks. That's fine for now, but the guard asymmetry is a latent confusion trap. At minimum, add a comment on the `handleDoAction` controller block explaining why the `!doIt` guard is absent (it's always a "do" click, so `doIt` is always true here). Without that, the next reader will assume it's a bug or copy the wrong pattern.

**2. [High] `setIsShowRunning` / `setIsDoRunning` not set before early return — spinner state inconsistency**

In controller mode, both handlers return early before calling `setIsShowRunning(true)` / `setIsDoRunning(true)`. That's intentional — there's nothing async to show a spinner for. But the guard at the top of both callbacks checks `isShowRunning` / `isDoRunning` as a debounce. Because these are never set in controller mode, the debounce is still effective (they're always false). No bug, but `isShowRunning` / `isDoRunning` contribute to the button's disabled/loading UI. If a parent ever re-enters the callback before the React re-render from `persistCompletion`, there's no in-flight guard. This is probably fine given the synchronous post, but it's worth a comment or, better, a brief `setIsShowRunning(true); return;` pattern is not needed — just confirm this is deliberate with a one-liner like `// synchronous post — no spinner needed`.

**3. [Medium] Comment in `step-checker.hook.ts` — borderline QC8 violation**

```ts
// A controller tab drives a *different* Grafana tab, so requirements that probe
// this tab's DOM/URL (exists-reftarget, on-page, ...) can't be evaluated here.
// Strip them and let the live tab enforce them; session/permission requirements
// still run so genuine failures surface.
```

This is four lines of justification for a one-liner `useMemo`. The *what* is already expressed by `stripTabLocalRequirements`'s name. The *why* (cross-tab architecture) is a legitimate hidden invariant — but it doesn't need four lines. The same comment is already present as a block comment at the top of `controller-requirements.ts`, which is the canonical place for it. Compress this to one line: `// controller tab can't probe a different tab's DOM — strip those requirements`. The detailed rationale lives in `controller-requirements.ts`.

**4. [Medium] `ControllerChannelProvider` `useState` lazy-init is correct but `useMemo` redundant**

```tsx
const [active] = useState<ChannelTransport>(() => transport ?? new CrossTabTransport(createSenderId()));
const channel = useMemo<ControllerChannel>(() => ({ post: (payload) => active.post(payload) }), [active]);
```

The lazy `useState` is the right pattern — the fallback is created once at mount, not on every render. Good. But `channel` wraps `active.post` in a new object stabilised by `useMemo([active])`. `active` is a state value — it never changes (it's `[active]` with no setter). The `useMemo` is therefore a no-op stabilisation: the identity of `channel` would be stable with a plain `useMemo` call, but `active` itself never triggers a re-memo. This works, it's just noise. Replace with:

```tsx
const channel = useMemo<ControllerChannel>(() => ({ post: (payload) => active.post(payload) }), [active]);
```

...actually that's what it already is. The deeper point: because `active` is a stable state ref, the `useMemo` provides no memoisation benefit over defining `channel` directly as `useRef({ post: ... }).current`. Not a bug, not worth a change request on its own, but note it if you're touching the file.

**5. [Low] `controllerChannel?.post(...)` optional chain papers over a missing invariant**

In `handleShowAction` and `handleDoAction`, the mode guard is `if (mode === 'controller')`, but the post call is `controllerChannel?.post(...)`. If `mode === 'controller'` and `controllerChannel` is null (i.e., the component is not wrapped in `ControllerChannelProvider`), the command is silently dropped and the user sees completion state advance with nothing sent to the live tab. This is a misconfiguration rather than a runtime scenario — `GuideReaderOverlay` always wraps in `ControllerChannelProvider` when `mode === 'controller'`. But the silent drop is dangerous: the controller marks the step done, the live tab never executes anything. Either assert (`if (!controllerChannel) { console.error(...); return; }`) or make the channel non-optional inside the `mode === 'controller'` branch so the invariant is expressed in code.

**6. [Low] Security: optimistic completion in controller mode is acceptable for the current scope**

Marking steps complete before live-tab confirmation means a user can advance a guide without the live tab ever executing the step (e.g., if the BroadcastChannel message is lost). For single-step interactions this is a deliberate UX tradeoff (PR description notes PR 1073 adds confirmation for multi-step/guided). No security concern — step completion is local learning-path state, not a privileged operation. No action required.

**7. [Low] `handleDoAction` dep array includes `targetComment` but `handleShowAction` also uses it — verify symmetry**

`targetComment` was added to `handleDoAction`'s dep array in this PR but was already in `handleShowAction`. Confirmed symmetric. No issue.

moxious:

**What's good:** clean `ControllerChannelProvider` lifecycle (lazy transport via `useState`, `start`/`stop` in a single effect); the `CrossTabAction` Tier-0 decoupling drops the `InteractiveAction` up-tier import from the type file; thorough `useCallback` dependency arrays; `stripTabLocalRequirements` is well-reasoned, and its comment (session/permission requirements deliberately retained so genuine failures still gate) earns its place.

### 1. Optimistic fire-and-forget completion (high)

Both handlers do:

```ts
if (mode === 'controller') {
  controllerChannel?.post({ kind: 'step-command', ... });
  persistCompletion();
  onStepComplete?.(stepId);
  onComplete?.();
  return;
}
```

The step is marked ✓ complete immediately, with no confirmation that a live tab received the command, that one even exists, or that the action succeeded. Because of the optional chaining on `controllerChannel?.post`, if there is no channel or no connected live tab, the step *still* completes and nothing happens anywhere — a silent no-op success the user reads as "done."

I see `step-complete` (#1073) and the requirement round-trip (#1070) add a reverse channel upstack; the open question is whether the **simple (non-composite) step path** ever gets gated on a live-tab acknowledgement, or stays optimistic. Flagging here as the origin of the pattern.

### 2. `TAB_LOCAL_REQUIREMENTS` drift risk (moderate)

The allowlist of DOM-probing requirement IDs (`exists-reftarget`, `navmenu-open`, `on-page`, `form-valid`) must stay in sync with the real set of tab-local requirements in requirements-manager. If a new DOM-probing requirement is added elsewhere and not added here, it won't be stripped and will be evaluated against the wrong tab. Consider deriving it from a single source or adding an assertion. (Also worth clarifying how this relates to #1070's round-trip — does stripping become partly redundant once requirements are evaluated on the live tab?)

### 3. Cross-context `stepId` determinism is a hidden invariant (moderate)

The posted `stepId: renderedStepId` only routes correctly if `renderedStepId` is deterministic **and identical across two different render contexts** — the controller overlay (`GuideReaderOverlay` → `ContentRenderer`) and the live sidebar (docs-panel). If the deterministic counter ever depends on mount-point-specific ordering, commands will silently address the wrong step. Worth an explicit invariant/test that the same guide yields identical step IDs in both contexts.

Tracking IDs `F-1063-1..3` in the stack review ledger.

moxious:
**Fix-pass outcome (re-stacked, gate-green, signed, pushed `d5b8dc27`)**

Made controller-mode steps fail loud when an anonymous step lacks a stable id (mis-addressing risk), and added a drift tripwire test so any new DOM-probing requirement must be added to the tab-local set or fail CI. Decision recorded: simple-step completion stays optimistic (no live ack), while composites are ack-gated.

leslie-heinzen (approved):

PR #1064, opened 2026-06-17
feat(twotab): execute controller commands against the live tab

Jayclifford345:
Add a module-level live-tab executor (integrations/cross-tab) that
subscribes to the cross-tab channel and replays step-commands against
this tab's Grafana DOM via the interactive-engine handlers (highlight /
button / formfill / navigate / hover): 'show' highlights, 'do' performs.
Lazily installed from module.tsx on the normal path only, so controller
and read-only tabs (which return early) never execute. Degrades to a
no-op where the channel is unavailable.

The engine barrel now re-exports the handler classes the executor needs,
keeping the import graph's barrel discipline intact.

This closes the round trip: click Show me / Do it in a popped-out
controller tab and the action runs on your live Grafana tab.

---

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -189,19 +189,21 @@ Excluded from tier analysis (not tiered): `test-utils/`, `cli/`, `bundled-intera
 
 **Key dependency edges** (where the load-bearing wiring lives):
 
-| Edge                                           | Why                                                            |
-| ---------------------------------------------- | -------------------------------------------------------------- |
-| `context-engine` → `docs-retrieval`            | Fetches content for the recommendations it surfaces            |
-| `docs-retrieval` → `bundled-interactives`      | Fallback when the online CDN is unavailable                    |
-| `docs-retrieval` → `package-engine`            | Resolves package manifests + content                           |
-| `docs-retrieval` → `snippet-engine`            | Inlines `snippet-ref` blocks against the CDN at parse time     |
-| `components/docs-panel` → `interactive-engine` | Executes step actions when the user clicks "Show me" / "Do it" |
-| `interactive-engine` → `requirements-manager`  | Checks prereqs before enabling / executing a step              |
-| `interactive-engine` → `lib/dom`               | Selector resolution + element detection                        |
-| `components/docs-panel` → `context-engine`     | Reads and renders recommendations                              |
-| `components/docs-panel` → `global-state`       | Sidebar, panel-mode, tab persistence                           |
-| `learning-paths` → `lib/user-storage`          | Persists progress + streak data                                |
-| `recovery` → `requirements-manager`            | Decides whether a failed requirement is auto-recoverable       |
+| Edge                                                          | Why                                                                                |
+| ------------------------------------------------------------- | ---------------------------------------------------------------------------------- |
+| `context-engine` → `docs-retrieval`                           | Fetches content for the recommendations it surfaces                                |
+| `docs-retrieval` → `bundled-interactives`                     | Fallback when the online CDN is unavailable                                        |
+| `docs-retrieval` → `package-engine`                           | Resolves package manifests + content                                               |
+| `docs-retrieval` → `snippet-engine`                           | Inlines `snippet-ref` blocks against the CDN at parse time                         |
+| `components/docs-panel` → `interactive-engine`                | Executes step actions when the user clicks "Show me" / "Do it"                     |
+| `interactive-engine` → `requirements-manager`                 | Checks prereqs before enabling / executing a step                                  |
+| `interactive-engine` → `lib/dom`                              | Selector resolution + element detection                                            |
+| `components/docs-panel` → `context-engine`                    | Reads and renders recommendations                                                  |
+| `components/docs-panel` → `global-state`                      | Sidebar, panel-mode, tab persistence                                               |
+| `learning-paths` → `lib/user-storage`                         | Persists progress + streak data                                                    |
+| `recovery` → `requirements-manager`                           | Decides whether a failed requirement is auto-recoverable                           |
+| `global-state/controller-channel` → `lib/cross-tab-transport` | BroadcastChannel link between a popped-out controller tab and the live Grafana tab |
+| `integrations/cross-tab` → `interactive-engine`               | Live tab replays a controller tab's step commands against its own DOM              |
 
 For per-subsystem entry points, public surfaces, and key files, load `.cursor/rules/systemPatterns.mdc`.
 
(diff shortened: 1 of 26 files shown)


moxious:

# PR #1064: feat(twotab): execute controller commands against the live tab

**Verdict:** Request changes

## Findings

1. **[High] Module-level `installed` flag + `resetLiveTabExecutorForTests` — testing anti-pattern**
   `live-tab-executor.ts`: `let installed = false` and the exported `resetLiveTabExecutorForTests()` are a canonical module-state anti-pattern. The module is now globally stateful; tests must remember to call the reset or they leak state across suites. The correct alternative is to convert to a factory function or class that callers hold a reference to, OR inject the singleton guard via a ref that tests can replace. At minimum the `resetLiveTabExecutorForTests` export should be marked `// @internal` and gated behind a check like `process.env.NODE_ENV !== 'production'` — exporting it at production module scope is a footgun.

2. **[High] `installed = true` is set before handler construction — failure leaves executor permanently dead**
   `live-tab-executor.ts`, lines ~50–65: `installed = true` is set before `new InteractiveStateManager()`, `new FocusHandler()` etc. are constructed. If any constructor throws, `installed` stays `true` and subsequent calls to `installLiveTabExecutor()` return the no-op uninstall function silently. The live tab never gets a working executor. Move `installed = true` to after all handlers are successfully constructed (or wrap the whole setup in a try/finally that resets `installed` on failure).

3. **[Medium] `module.tsx` comment is wrong — "controller / read-only tabs returned early above" is false**
   The comment on the `installLiveTabExecutor` call says "controller / read-only tabs returned early above". The controller path does return early (the `docsParam && controllerParam` guard). But the read-only deep-link path (the `docsParam && !sidebarMountable` guard, which handles `?readonly=1`) appears *below* the executor install in the file — meaning read-only tabs also install the executor before returning. Either the guard order needs fixing (move the read-only return above the executor install) or the comment needs correcting. Verify the intended execution order and fix whichever is wrong.

4. **[Medium] In-flight `runStepCommand` can outlive teardown — highlights reappear after uninstall**
   `installLiveTabExecutor` returns an uninstall function that calls `navigationManager.clearAllHighlights()` synchronously. But `runStepCommand` is `async` and dispatched with `void runStepCommand(...)` — fire-and-forget. If teardown fires while a step command is mid-execution, `clearAllHighlights()` runs first and then the async step finishes and calls `focusHandler.execute()`, re-applying highlights on a "stopped" executor. Add an abort mechanism (e.g. a `cancelled` boolean set on uninstall; check it at the start of each action in `runStepCommand`).

5. **[Low] `tagName: 'button'` hardcoded for all action types — `live-tab-executor.ts`**
   The synthetic `InteractiveElementData` sets `tagName: 'button'` for every action, including `navigate` and `formfill`. If `NavigateHandler` or `FormFillHandler` branches on `tagName`, this will silently misbehave. Verify that no handler uses `tagName` for dispatch logic, or set it dynamically from `command.action.targetAction`.

6. **[Low] `textContent` is synthetic and does not reflect the actual element — `live-tab-executor.ts`**
   `textContent: \`${isShow ? 'Show me' : 'Do'}: ${command.action.refTarget}\`` is a synthetic string built from the selector, not from the DOM element. If any handler logs or surfaces `textContent` in an error message, users will see "Do: #save-button" rather than "Save" — confusing. Acceptable for now, but worth noting.

moxious:

**What's good:** the executor is lazy-imported (keeps the interactive engine out of the entry bundle); style init and each command run inside try/catch; handlers share state/navigation managers correctly.

### 🔴 1. Remote DOM execution from envelope-only-validated messages (BLOCKING)

This is the sink for a trust gap that starts in #1056. The transport's receive gate (`isPathfinderMessage`) checks only `source === 'pathfinder'` and that `senderId` is a string — both trivially forgeable — and does **not** validate `kind` or the action payload. This executor then runs `navigate` / `button` / `formfill` / `hover` / `highlight` handlers against the live Grafana DOM for any `step-command` it receives.

Two things make the blast radius large:

- **It installs unconditionally.** In `module.tsx`, `installLiveTabExecutor()` runs in `plugin.init` with no `pathfinderEnabled` gate, before `sidebarMountable` is computed — so every Grafana page in a session with the plugin loaded is listening.
- **No sender authentication.** Pairing (#1073) selects which live tab the controller drives, but `senderId` is observable and forgeable on the shared channel, so pairing does not authorize the sender.

Anyone with same-origin script execution — a malicious or compromised co-installed plugin, an XSS in any panel / dashboard / data source, or a browser-extension content script — can post a `step-command` and drive navigation, clicks, and form fills on the user's authenticated Grafana. This turns any contained same-origin foothold into UI control plus navigation.

Suggested remediation:
- Schema-validate each message per `kind` before dispatch (reject unknown/malformed shapes), ideally at the transport boundary in #1056.
- Gate the executor install on `pathfinderEnabled`.
- Require an authorization handshake before accepting remote DOM commands — e.g., a user gesture on the live tab to "accept control," or a nonce exchanged out-of-band when the controller tab is opened, rather than trusting any same-origin sender.

### 2. No reverse error channel (moderate)

`runStepCommand` swallows failures (`console.error`) and never reports back. Combined with the optimistic completion in #1063, a remote action that fails leaves the controller showing the step as complete. The later `step-complete` / `requirement-result` messages add a reverse path — worth confirming failures (not just successes) flow back.

### 3. Parallel action-dispatch table (moderate, drift)

The `switch (command.action.targetAction)` here duplicates the interactive engine's own action routing. When actions are added or renamed in the engine, this switch silently goes stale (unknown action → `console.warn`) and that capability just stops working over the channel. Consider routing through the engine's existing dispatch instead of a parallel table.

### Smaller items

- `module.tsx` discards the teardown returned by `installLiveTabExecutor()`, so it's a page-lifetime singleton (`installed` module global + `resetLiveTabExecutorForTests`). Confirm that's intended; it's coupled to the enablement-gate gap above.
- `InteractiveElementData.tagName` is hardcoded `'button'` for every action.

Tracking IDs `F-1064-1..3` (and `F-1056-1`) in the stack review ledger.

moxious:
**Fix-pass outcome (re-stacked, gate-green, signed, pushed `e34c6c66`)**

**T1 PART B.** Gated `installLiveTabExecutor` on `pathfinderEnabled` and ran the shared per-kind validator at the executor sink before any dispatch. Also serialized replays onto a single chain with a teardown cancel flag, and moved the install-slot claim to after handler construction (so a throwing constructor can't permanently brick the executor). The reverse error channel was deferred to #1073, where the `step-complete` kind exists.

leslie-heinzen (approved):

PR #1065, opened 2026-06-17
feat(twotab): show whether the controller is linked to a live tab

Jayclifford345:
The controller overlay heartbeats over the channel and the live-tab
executor acks; the controller tracks the ack and surfaces a status badge
("Connected to your Grafana tab" / "Waiting for your Grafana tab…").
Connection goes stale if no ack arrives within the timeout, so closing or
navigating away from the live tab flips the badge back to waiting.

---

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -189,19 +189,21 @@ Excluded from tier analysis (not tiered): `test-utils/`, `cli/`, `bundled-intera
 
 **Key dependency edges** (where the load-bearing wiring lives):
 
-| Edge                                           | Why                                                            |
-| ---------------------------------------------- | -------------------------------------------------------------- |
-| `context-engine` → `docs-retrieval`            | Fetches content for the recommendations it surfaces            |
-| `docs-retrieval` → `bundled-interactives`      | Fallback when the online CDN is unavailable                    |
-| `docs-retrieval` → `package-engine`            | Resolves package manifests + content                           |
-| `docs-retrieval` → `snippet-engine`            | Inlines `snippet-ref` blocks against the CDN at parse time     |
-| `components/docs-panel` → `interactive-engine` | Executes step actions when the user clicks "Show me" / "Do it" |
-| `interactive-engine` → `requirements-manager`  | Checks prereqs before enabling / executing a step              |
-| `interactive-engine` → `lib/dom`               | Selector resolution + element detection                        |
-| `components/docs-panel` → `context-engine`     | Reads and renders recommendations                              |
-| `components/docs-panel` → `global-state`       | Sidebar, panel-mode, tab persistence                           |
-| `learning-paths` → `lib/user-storage`          | Persists progress + streak data                                |
-| `recovery` → `requirements-manager`            | Decides whether a failed requirement is auto-recoverable       |
+| Edge                                                          | Why                                                                                |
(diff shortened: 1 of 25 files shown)


moxious:

# PR #1065: feat(twotab): show whether the controller is linked to a live tab

**Verdict:** Request changes (minor)

## Findings

1. **[Medium] `connected` in `useMemo` causes all step components to re-render on connect/disconnect**
   `controller-channel.tsx`: `channel = useMemo(() => ({ post, connected }), [active, connected])`. The `channel` object identity changes on every heartbeat ack (connected: false → true) and stale event (connected: true → false). Since `ControllerChannelProvider` passes `channel` as a context value, every component that calls `useControllerChannel()` re-renders on each connection state change — including every `InteractiveStep`. This may be 10–30 components. Split `connected` into a separate context (e.g. `ControllerConnectionContext`) so only `ControllerStatusBadge` re-renders on connect/disconnect changes, while step components remain stable.

2. **[Medium] `className` string concatenation — prefer `cx()` from Grafana/Emotion**
   `GuideReaderOverlay.tsx`, `ControllerStatusBadge`: 
   ```tsx
   className={`${styles.controllerStatusDot} ${connected ? styles.controllerStatusConnected : styles.controllerStatusWaiting}`}
   ```
   The Grafana codebase uses `cx(styles.controllerStatusDot, connected ? styles.controllerStatusConnected : styles.controllerStatusWaiting)` (from `@emotion/css` or `@grafana/ui`). String concatenation is a code smell here — it produces a different class string format and can misbehave with Emotion's hash-based class names. Use `cx()`.

3. **[Low] `channel?.connected ?? false` null guard is unreachable — `GuideReaderOverlay.tsx`**
   `ControllerStatusBadge` is rendered exclusively inside `<ControllerChannelProvider>`, so `useControllerChannel()` will never return null here. The `?.` and `?? false` are defensive but mislead readers into thinking null is possible at this call site. Use a non-nullable assertion or restructure: `const { connected } = useControllerChannel()!`.

4. **[Low] Heartbeat timing is not documented and uses magic numbers**
   `HEARTBEAT_INTERVAL_MS = 2000` and `HEARTBEAT_STALE_MS = 5000` are module-level constants with no co-location in `INTERACTIVE_CONFIG`. The 2.5× ratio (stale = 2.5 × interval) is the implicit design contract: 2 missed heartbeats = disconnected. If `HEARTBEAT_INTERVAL_MS` is ever tuned, `HEARTBEAT_STALE_MS` must be tuned in tandem. A comment like `// stale after ~2 missed heartbeats` next to `HEARTBEAT_STALE_MS` makes the invariant explicit.

5. **[Low] `tick()` is called immediately on mount, firing before any live tab can respond**
   The initial `tick()` call posts a controller heartbeat and immediately checks staleness. Since `lastLiveSeenRef.current` starts at 0, the stale check is `0 > 0 && ...` which is false — so no false stale on mount. Good. But the badge starts as "Waiting for your Grafana tab…" (connected=false), which is the right UX default. No bug; noting for clarity.

moxious:
**Fix-pass outcome (re-stacked, gate-green, signed, pushed `d062805c`)**

Split connection state into its own context so every `InteractiveStep` no longer re-renders on each heartbeat tick — only the status badge consumes the connection context now.

leslie-heinzen (approved):

PR #1067, opened 2026-06-17
feat(twotab): hand the sidebar between the main window and the controller tab

Jayclifford345:
A controller tab posts a sidebar-handoff over the channel: on open the
live tab closes its Pathfinder sidebar (close-extension-sidebar); on close
(pagehide / unmount) it posts a reopen so the live tab restores the
sidebar — but only if this tab gave it up and no other plugin now occupies
the slot.

---

### DIFF
--- a/docs/developer/CROSS_TAB_CONTROLLER.md
+++ b/docs/developer/CROSS_TAB_CONTROLLER.md
@@ -48,8 +48,9 @@ GuideReaderOverlay mode="controller"          module.tsx normal path
 | Transport                      | `src/lib/cross-tab-transport.ts` (`CrossTabTransport`)                                                                      |
 | Message protocol               | `src/types/cross-tab.types.ts`                                                                                              |
 | Controller emit + presence     | `src/global-state/controller-channel.tsx` (`ControllerChannelProvider`, `useControllerChannel`)                             |
-| Per-step emit                  | `src/components/interactive-tutorial/interactive-step.tsx` (handlers branch on `useInteractiveMode()`)                      |
-| Live-tab executor              | `src/integrations/cross-tab/live-tab-executor.ts` (`installLiveTabExecutor`)                                                |
+| Per-step emit                  | `interactive-step.tsx`, `interactive-multi-step.tsx`, `interactive-guided.tsx` (handlers branch on `useInteractiveMode()`)  |
+| Requirement round-trip         | `src/requirements-manager/step-checker.hook.ts` (controller branch), `controller-requirements.ts` (tab-local set)           |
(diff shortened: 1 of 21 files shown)


moxious:

# PR #1067: feat(twotab): hand the sidebar between the main window and the controller tab

**Verdict:** Merge as-is

## Findings

1. **[Info] pagehide + React unmount double-post is safe.** When the tab is navigated away, `pagehide` fires and `handBack()` posts `sidebar-handoff: reopen`. React then runs the cleanup function, which calls `handBack()` a second time. Both posts arrive on the `BroadcastChannel` at the live tab. The live executor handles this correctly: the first `reopen` fires `sidebarState.openSidebar` and sets `handedOffSidebar = false`; the second `reopen` hits the `if (handedOffSidebar && ...)` guard, which is now `false`, so nothing happens. No double-open, no error. The idempotency is incidental but real — no code change needed, but a brief comment on the `handBack()` call in the cleanup ("safe to double-post; executor guards on handedOffSidebar") would make the intent clear to the next reader.

2. **[Info] `sidebarState.openSidebar('Interactive learning')` is the correct call.** `openSidebar` takes a `componentTitle: string` as its first argument. `'Interactive learning'` matches the string used in `openWithGuide` and by `handleReturnToSidebar` elsewhere in the codebase — it's the canonical Pathfinder sidebar title.

3. **[Info] `handedOffSidebar` closure scope is correct.** It lives inside `installLiveTabExecutor`, which is called once per live tab load (guarded by the `installed` singleton flag). Each live tab gets exactly one closure, so the flag correctly tracks whether _this_ tab gave up its sidebar. A module-level variable would incorrectly survive across test resets and would not survive the `installed = false` reset on uninstall; the closure placement is right.

4. **[Minor] Multiple simultaneous controller tabs.** If two controller tabs open at the same time, both post `sidebar-handoff: close`. The live executor closes the sidebar on the first and sets `handedOffSidebar = true`; the second close arrives when `getIsSidebarMounted()` is already `false`, so it skips the close and does _not_ set `handedOffSidebar` again (the `if` guard prevents it). When the first tab closes and posts `reopen`, `handedOffSidebar` is `true` and the sidebar reopens — correct. When the second tab closes and posts `reopen`, `handedOffSidebar` is `false` (already cleared), so the reopen is skipped — also correct. The logic holds for the simultaneous case without any changes. Worth noting: the v1 doc already acknowledges "multiple live tabs all execute a command (no targeting)" as acceptable; the same pragmatism applies here.

moxious:
**Fix-pass outcome (re-stacked, gate-green, signed, pushed `ffabc16c`)**

Re-assert the sidebar `close` once the first live heartbeat arrives — the one-shot close on mount could be silently dropped if the live tab mounts after the controller. Flag-guarded so it doesn't re-post on every reconnect. Added the `sidebar-handoff` validator case.

leslie-heinzen (approved):

PR #1068, opened 2026-06-17
feat(twotab): extend controller mode to multi-step, guided, and sections

Jayclifford345:
Multi-step and guided steps now emit their internalActions sequence over
the channel in controller mode (instead of executing locally and failing
to resolve elements in the popped-out tab); the live-tab executor replays
the sequence in order. Sections filter tab-local requirements
(navmenu-open, on-page, ...) on their own requirement path
(useSectionRequirements + handleDoSection), so they stop blocking in
controller mode while session requirements still gate.

---

### DIFF
--- a/docs/developer/CROSS_TAB_CONTROLLER.md
+++ b/docs/developer/CROSS_TAB_CONTROLLER.md
@@ -48,8 +48,9 @@ GuideReaderOverlay mode="controller"          module.tsx normal path
 | Transport                      | `src/lib/cross-tab-transport.ts` (`CrossTabTransport`)                                                                      |
 | Message protocol               | `src/types/cross-tab.types.ts`                                                                                              |
 | Controller emit + presence     | `src/global-state/controller-channel.tsx` (`ControllerChannelProvider`, `useControllerChannel`)                             |
-| Per-step emit                  | `src/components/interactive-tutorial/interactive-step.tsx` (handlers branch on `useInteractiveMode()`)                      |
-| Live-tab executor              | `src/integrations/cross-tab/live-tab-executor.ts` (`installLiveTabExecutor`)                                                |
+| Per-step emit                  | `interactive-step.tsx`, `interactive-multi-step.tsx`, `interactive-guided.tsx` (handlers branch on `useInteractiveMode()`)  |
+| Requirement round-trip         | `src/requirements-manager/step-checker.hook.ts` (controller branch), `controller-requirements.ts` (tab-local set)           |
+| Live-tab executor              | `src/integrations/cross-tab/live-tab-executor.ts` (`installLiveTabExecutor`: replay, requirement eval, remote fix)          |
 
 ## Message protocol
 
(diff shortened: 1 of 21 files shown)


moxious:

# PR #1068: feat(twotab): extend controller mode to multi-step, guided, and sections

**Verdict:** Request changes

## Findings

1. **[Medium] `cross-tab.types.ts` — inline anonymous type for `internalActions` elements**

   `internalActions?: Array<{ targetAction: string; refTarget?: string; targetValue?: string; targetComment?: string }>` uses an ad-hoc structural type that is a strict subset of `InternalAction` (already in `src/types/interactive-actions.types.ts`). The subset drops `requirements`, which is intentional for the cross-tab wire format, but the narrowing is implicit and undocumented. Either reuse `InternalAction` directly (the extra `requirements` field is harmless on the wire), or extract a named `CrossTabInternalAction` type and export it. An unnamed inline type is invisible to `Find References`, blocks future narrowing with a named union, and silently diverges from `InternalAction` when that type gains a field.

2. **[Medium] `interactive-section.tsx` — strip logic at section level may diverge from step-checker strip in PR 1070**

   Section-level `controllerRequirements = stripTabLocalRequirements(requirements)` is its own independent call site. PR 1063's `useStepChecker` strips tab-local requirements before DOM checks on the step path; PR 1070 replaces that strip with live-tab round-trips. Once PR 1070 lands, the section path will still be stripping while the step path round-trips — two divergent behaviours for the same requirements string. Unless sections are deliberately kept as strip-only forever (which should be documented), this is a latent consistency bug. At minimum, add a co-change note so whoever lands PR 1070 knows to revisit this site.

3. **[Medium] `interactive-guided.tsx` / `interactive-multi-step.tsx` — optimistic completion is not flagged as provisional**

   `persistCompletion()` / `onStepComplete()` / `onComplete()` are called immediately in controller mode before the live tab confirms. The PR description notes this is fixed in PR 1073, but within this PR the code reads as final. A reader landing on this PR in isolation has no signal that the completion is deliberately optimistic and will be corrected. A one-line inline note ("provisional — superseded by live-tab confirmation in PR 1073") or a TODO in the controller-mode branches would prevent the next person from treating it as a deliberate design choice rather than a known debt.

4. **[Low] `live-tab-executor.ts` — `targetAction: 'multistep'` or `'guided'` reaching `runAction` falls through to warning**

   The refactor routes composite step types via `internalActions`, so a top-level `CrossTabAction` with `targetAction: 'multistep'` and no `internalActions` array will hit the `default: console.warn(...)` branch in `runAction`. That is probably the right silent failure for a malformed command, but the warning message should call out which `targetAction` values are not dispatchable as single actions (i.e., `'multistep'` and `'guided'` are container types that require `internalActions`). Currently the warning is generic and gives the on-call no triage path. Sharpen the message or add those values to an explicit guard-and-warn case.

5. **[Low] Comment quality — multiple bad-shape comments in changed hunks**

   - `interactive-section.tsx` line ~205: `// Section-level requirements polling (initial check, 4 event listeners, 5-second setInterval) is owned by useSectionRequirements. Pattern F timing contract — the 5s cadence is load-bearing for "auto-pickup of out-of-band state changes that don't fire one of the 4 events".` — process/pattern artefact reference (QC3 and QC6). The cadence constraint belongs in `useSectionRequirements` itself, not duplicated here.
   - `interactive-multi-step.tsx` line ~630: `// setExecutionError is a stable React setter, isCancelledRef is a ref — neither needs to appear in the deps array.` — defends a non-action (QC2). Delete; ESLint exhaustive-deps already enforces this.
   - The large block-comment banners (`═══ STATE: IDLE ═══` etc.) inside both components are new and repeat the `data-test-step-state` value visible three lines away (QC1 / QC5). They add visual noise without load-bearing information. Trim to a plain heading or nothing.

moxious:

**What's good:** the `runAction`/`runStepCommand` split is clean, `noop` is handled, and composites carry an ordered `internalActions` array replayed sequentially so ordering is preserved.

### 1. Optimistic completion extends to composites (see #1063)

Guided and multi-step use the same fire-and-forget pattern as the simple step: post the `step-command`, then immediately `persistCompletion()` + `onStepComplete` + `onComplete`. Composites take *longer* to replay on the live tab (especially once #1069 paces them), so the controller marks the step ✓ complete while the live tab is still mid-sequence — and if no live tab is connected, it completes with nothing happening anywhere. I see #1073's `step-complete` is meant to gate this; flagging the composite case as part of that thread.

### 2. Top-level `targetAction: 'guided' | 'multistep'` is decorative (low)

The executor routes composites purely by `internalActions?.length`; it never switches on `targetAction === 'guided'`. Those values aren't in `runAction`'s `switch`, so if one ever reached `runAction` it'd hit the `console.warn` default. Works today, but it's a slightly fragile coupling — the top-level action label is unused on the live side.

### 3. `internalActions` dependency (resolved, noting for the record)

The dep arrays switch from `internalActions.length` to `internalActions`. That's correct per exhaustive-deps, and since `internalActions` originates in the content parser (parsed-once content passed via props) it's referentially stable, so the `useCallback` identity won't churn. No action needed.

Tracking IDs `F-1068-1..3` in the stack review ledger.

moxious:
**Fix-pass outcome (re-stacked, gate-green, signed, pushed `17f65954`)**

Extracted an exported `CrossTabInternalAction` type, and **fixed a real latent bug**: `'guided'` was missing from the known-verb set, so guided composites would have failed the receive gate (only `'multistep'` was seeded). The validator now accepts `'guided'` and validates every `internalActions[]` element's verb.

leslie-heinzen (approved):

PR #1069, opened 2026-06-17
feat(twotab): pace multi-step replay on the live tab

Jayclifford345:
Stacked on #1068.

Composite steps (multi-step, guided) emit a single `do`-command carrying their ordered `internalActions`. The live-tab executor replayed them in a tight loop, so the live tab burst through every action at once instead of the staged highlight → perform cadence a normal multi-step shows.

Now each internal action replays the way a live multi-step paces itself: **show (highlight) → pause → do (perform) → settle → inter-step pause**, using `INTERACTIVE_CONFIG.delays.multiStep`. Pacing is injectable so tests stay deterministic.

### DIFF
--- a/docs/developer/CROSS_TAB_CONTROLLER.md
+++ b/docs/developer/CROSS_TAB_CONTROLLER.md
@@ -48,8 +48,9 @@ GuideReaderOverlay mode="controller"          module.tsx normal path
 | Transport                      | `src/lib/cross-tab-transport.ts` (`CrossTabTransport`)                                                                      |
 | Message protocol               | `src/types/cross-tab.types.ts`                                                                                              |
 | Controller emit + presence     | `src/global-state/controller-channel.tsx` (`ControllerChannelProvider`, `useControllerChannel`)                             |
-| Per-step emit                  | `src/components/interactive-tutorial/interactive-step.tsx` (handlers branch on `useInteractiveMode()`)                      |
-| Live-tab executor              | `src/integrations/cross-tab/live-tab-executor.ts` (`installLiveTabExecutor`)                                                |
+| Per-step emit                  | `interactive-step.tsx`, `interactive-multi-step.tsx`, `interactive-guided.tsx` (handlers branch on `useInteractiveMode()`)  |
+| Requirement round-trip         | `src/requirements-manager/step-checker.hook.ts` (controller branch), `controller-requirements.ts` (tab-local set)           |
+| Live-tab executor              | `src/integrations/cross-tab/live-tab-executor.ts` (`installLiveTabExecutor`: replay, requirement eval, remote fix)          |
 
 ## Message protocol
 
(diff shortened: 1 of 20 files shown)


moxious:

# PR #1069: feat(twotab): pace multi-step replay on the live tab

**Verdict:** Request changes (minor)

## Findings

1. **[Medium] `live-tab-executor.ts` — `settleMs: 200` is a hardcoded magic number**

   `settleMs: 200` in `DEFAULT_PACING` has no home in `INTERACTIVE_CONFIG`. Every other settle/DOM-settling delay in the codebase is named: `technical.navigationDock` (200 ms), `debouncing.stateSettling` (100 ms), `delays.navigation.domSettlingDelay` (300 ms). The 200 ms value here matches `technical.navigationDock` by coincidence, but they are semantically different. Add a `multiStep.settleAfterActionMs` (or similar) entry to the `INTERACTIVE_CONFIG_DEFAULTS.delays.multiStep` block so the value is visible, overridable, and survives a grep. A bare literal in `DEFAULT_PACING` is the only timing constant in this subsystem that is not centralised.

2. **[Low] `DEFAULT_PACING.showToDoMs` — derived by arithmetic that re-does what `interactive-multi-step.tsx` already does inline**

   `showToDoMs: INTERACTIVE_CONFIG.delays.multiStep.showToDoIterations * INTERACTIVE_CONFIG.delays.multiStep.baseInterval` recomputes `18 * 100 = 1800 ms` at module initialisation. This is the same product the existing `interactive-multi-step.tsx` loop iterates over manually (PR 1068 still runs the cancellation-safe loop on the live tab). That is fine for now, but the two representations will silently diverge if either `showToDoIterations` or `baseInterval` changes independently in the config. Consider adding a derived constant `INTERACTIVE_CONFIG_DEFAULTS.delays.multiStep.showToDoMs` so both consumers share a single computed value rather than each multiplying independently.

3. **[Low] `settleDom()` — double `requestAnimationFrame` pattern has no explanation**

   ```ts
   const settleDom = (): Promise<void> =>
     new Promise((resolve) => requestAnimationFrame(() => requestAnimationFrame(resolve)));
   ```

   The double-rAF pattern is a well-known React idiom for flushing two render passes (schedule → commit → paint), but it is not obvious to a reader who doesn't recognise it. Per the project's comment policy this is exactly the counterintuitive-but-correct case that earns a one-line comment: `// Two rAF passes: first schedules React's commit, second waits for the browser paint.` The existing `interactive-multi-step.tsx` (line ~393) uses the same pattern without a comment — add it there too if that file is in the changeset, but at minimum add it here where the named function is the canonical definition.

4. **[Low] `ExecutorPacing` / `DEFAULT_PACING` / `sleep` / `settleDom` are all module-level in a singleton module**

   The module uses a `let installed = false` singleton install guard (from prior PRs in the stack). `sleep` and `settleDom` are pure module-level constants with no state, so there is no interaction risk. However `DEFAULT_PACING` is also module-level and references `INTERACTIVE_CONFIG` at import time. If a test hot-patches `INTERACTIVE_CONFIG` values after import (a pattern used in a few timing tests in this codebase), `DEFAULT_PACING` will silently hold the pre-patch values. The injectable `pacing` parameter in `runComposite` already mitigates this for tests, but it's worth noting the gap for the next person adding a timing test: always pass `pacing` explicitly rather than relying on the module-level default.

moxious:

**What's good:** the show → pause → do → settle → inter-step cadence mirrors how a live multi-step paces itself, so the user watches a staged sequence rather than an instant burst. The `pacing` parameter is injectable (testable), it reuses `INTERACTIVE_CONFIG`, and `settleDom` (double `requestAnimationFrame`) is a sound way to wait for a DOM settle. The explaining comment earns its place.

### 1. Replays are neither serialized nor cancelable (moderate, concurrency)

The executor handles each message with `void runStepCommand(message)` and there's no queue or mutex. Two consequences:

- **Overlapping commands interleave.** If the user clicks quickly, or a new `step-command` arrives while a composite is still replaying, the two `runComposite` loops run concurrently and their DOM actions interleave on the live tab.
- **No cancellation.** A composite replay is a sequence of `await`s with `sleep`s; if the user closes the controller or moves on mid-replay, it runs to completion regardless.

Worth a per-executor serialization (a promise chain or in-flight guard) and ideally a cancellation token tied to a newer command / disconnect.

### 2. `runComposite` ignores `command.phase` (low)

It always runs each action as show then do. That's fine today because composites are always posted with `phase: 'do'`, but it's an implicit assumption worth a comment or guard.

Tracking IDs `F-1069-1` in the stack review ledger.

moxious:
**Fix-pass outcome (re-stacked, gate-green, signed, pushed `c09891fe`)**

Confirmed the replay serialization (landed at #1064) survives this pacing layer, with the cancel guard inside the composite loop. Centralized the `settleAfterActionMs` pacing literal in the interactive config rather than leaving it inline.

leslie-heinzen (approved):

PR #1070, opened 2026-06-17
feat(twotab): evaluate controller requirements on the live tab

Jayclifford345:
Stacked on #1069.

**Problem.** Controller mode stripped tab-local requirements (`navmenu-open`, `exists-reftarget`, `on-page:`, `form-valid`) so they never gated — a step whose live-tab prerequisite was unmet looked ready in the controller, and its "Fix this" affordance never appeared. (Reported: the nav-menu requirement + Fix never shows in the controller the way it does in the sidebar.)

**Fix.** Evaluate them where they're real. The controller round-trips the requirement to the live tab over the existing channel and the live-tab executor runs `checkRequirements` / `dispatchFix` against its own DOM.

- New messages: `check-requirements` / `requirement-result` / `fix-requirement` / `fix-result`, correlated by `requestId` (reply-message pattern, no transport API change).
- The reply is a plain `RequirementsCheckResult` and flows through the **unchanged** `createRequirementsState` path — so the warning + "Fix this" render exactly as in-tab.
- "Fix this" routes to the live tab; a controller-mode heartbeat polls while a fragile step is **blocked** so the warning clears once the prerequisite is met.
- Falls back to the prior strip-and-pass behavior when no live tab answers, so a disconnected controller never hangs.

**Tiers.** Evaluation/fixes live on the executor side (tier 3 → requirements-manager, downward). `controller-channel` (tier 1) only correlates replies and carries a tier-0 result type — no upward import. Governance unchanged (vertical=2 lateral=8 barrel=0).

### DIFF
--- a/docs/developer/CROSS_TAB_CONTROLLER.md
+++ b/docs/developer/CROSS_TAB_CONTROLLER.md
@@ -48,8 +48,9 @@ GuideReaderOverlay mode="controller"          module.tsx normal path
 | Transport                      | `src/lib/cross-tab-transport.ts` (`CrossTabTransport`)                                                                      |
 | Message protocol               | `src/types/cross-tab.types.ts`                                                                                              |
 | Controller emit + presence     | `src/global-state/controller-channel.tsx` (`ControllerChannelProvider`, `useControllerChannel`)                             |
-| Per-step emit                  | `src/components/interactive-tutorial/interactive-step.tsx` (handlers branch on `useInteractiveMode()`)                      |
-| Live-tab executor              | `src/integrations/cross-tab/live-tab-executor.ts` (`installLiveTabExecutor`)                                                |
+| Per-step emit                  | `interactive-step.tsx`, `interactive-multi-step.tsx`, `interactive-guided.tsx` (handlers branch on `useInteractiveMode()`)  |
+| Requirement round-trip         | `src/requirements-manager/step-checker.hook.ts` (controller branch), `controller-requirements.ts` (tab-local set)           |
+| Live-tab executor              | `src/integrations/cross-tab/live-tab-executor.ts` (`installLiveTabExecutor`: replay, requirement eval, remote fix)          |
 
 ## Message protocol
 
@@ -58,8 +59,21 @@ the `pathfinder-cross-tab` channel. Every message carries an envelope
 (`source: 'pathfinder'`, a per-tab `senderId` used to drop self-echoes, and a
 `timestamp`):
 
-- `step-command` — `{ phase: 'show' | 'do', stepId, action: { targetAction, refTarget, targetValue?, targetComment? } }`
+- `step-command` — `{ phase: 'show' | 'do', stepId, action: { targetAction, refTarget, targetValue?, targetComment?, internalActions? } }`.
+  Composite steps carry their ordered sub-actions in `internalActions`. A
+  `multistep` replays with staged pacing (see [Replay pacing](#replay-pacing));
+  a `guided` step runs through the live tab's `GuidedHandler` instead — it
+  highlights each target and waits for the user.
+- `step-complete` — `{ stepId, ok }`, live → controller, signals a composite
+  actually finished so the controller marks completion only then.
+- `step-progress` — `{ stepId, index, total }`, live → controller, reports which
+  internal action a composite is replaying so the controller can animate per-step
+  progress while it runs on the live tab.
 - `heartbeat` — `{ role: 'controller' | 'live' }`
+- `check-requirements` / `requirement-result` — the requirement round-trip
+  (controller → live → controller), correlated by `requestId`.
+- `fix-requirement` / `fix-result` — a "Fix this" routed to the live tab,
+  correlated by `requestId`.
 
 The transport degrades to a no-op where `BroadcastChannel` is unavailable. The
 provider is inert until a tab actually drives one; the executor, however, is a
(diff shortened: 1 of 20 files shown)


moxious:

# PR #1070: feat(twotab): evaluate controller requirements on the live tab

**Verdict:** Request changes

## Findings

**1. [Minor] `settle` captures a snapshot of `pendingRef.current`; the timeout path in `request` does not — they refer to the same underlying Map, so there is no real aliasing bug, but the asymmetry is a trap.**

`settle` closes over `const pending = pendingRef.current` (captured once at effect mount). The timeout handler inside `request` calls `pendingRef.current.delete(requestId)` directly. Both always point to the same `Map` instance because `pendingRef.current` is assigned exactly once (`useRef(new Map())`), so the behaviour is correct today. However, if a future refactor ever reassigns `pendingRef.current` (e.g. to reset state on reconnect), the snapshot in `settle` would silently operate on the old map while the timeout arm operates on the new one. A one-line comment on the `const pending = ...` line explaining the invariant — or just having the timeout arm call `settle(requestId, fallback)` directly — would eliminate the trap. The timeout arm resolving without going through `settle` also means a timed-out request skips the `clearTimeout` call on its own timer (the timer fires itself, but if `settle` were called instead it would be idempotent and self-documenting).

**2. [Moderate] `Promise<any>` return type on `attemptCheck` in `step-checker.hook.ts` (line 255) is pre-existing but this PR adds the remote branch that returns a `RemoteRequirementResult` through that same untyped slot.**

The function now returns either a `RequirementsCheckResult` (local path) or a `RemoteRequirementResult` (remote path) or a hand-rolled `{ requirements, pass, error }` literal from the catch blocks — all structurally compatible but invisible to the compiler. The remote branch result is then passed directly to `createRequirementsState`, which accepts `RequirementsCheckResult`. Since `RemoteRequirementResult` is a structural mirror this works, but a type annotation on `attemptCheck` — `Promise<RequirementsCheckResult>` — would let the compiler catch a future mirror divergence at the call site rather than at runtime. This PR is a good moment to add it since it widens the return surface.

**3. [Moderate] Silent fail-open on timeout is the wrong default for tab-local DOM requirements.**

When `requestRequirementCheck` returns `null` (4 s timeout, no live tab), the fallback strips tab-local tokens and runs a local check that will almost certainly pass (nothing to probe). The step is therefore silently enabled on a disconnected controller. This is documented in the PR description as intentional — "a disconnected controller never hangs" — but the UX consequence is that a user can click "Do it" on a step whose DOM precondition has never actually been verified. A tighter alternative: resolve the timeout as `{ pass: false, requirements, error: [{ requirement: requirements, pass: false, error: 'No live tab' }] }` so the step stays blocked with a "Fix this" / "Connect to live tab" affordance. If the current fail-open behaviour is an explicit product decision, it deserves a comment on the fallback path explaining why open beats closed, not just that it prevents hangs.

**4. [Minor] `requestSeqRef` rolls over silently if the controller is live for a very long session or the component remounts many times without a page refresh.**

`requestSeqRef.current` is a plain JS number incrementing with `+=`. JS numbers are IEEE 754 doubles; integer precision is safe up to 2^53 so there is no practical collision risk in a single session. The concern is semantic: if the component unmounts and remounts (React StrictMode double-invoke in dev, route change that keeps the provider alive), `requestSeqRef.current` resets to 0 because `useRef` reinitialises. Meanwhile the cleanup in the `useEffect` return resolves all pending requests with `null` and clears the map, so there is no double-settle risk — the IDs are just reused across mount cycles with no overlap. This is fine; noting it so reviewers can dismiss it explicitly.

**5. [Minor] QC8 comment shapes — a few remain in the diff that don't earn their place.**

- Lines 113–119 in `step-checker.hook.ts` (`// A controller tab drives a *different* Grafana tab...`): this is a QC8-shape 2 ("Intentionally NOT X here because Y") attached to `const requirements = rawRequirements`. The doc and the round-trip logic itself carry this reasoning; the comment is defending a non-action against a future revert. Trim.
- The `// Disable internal retry since we're handling it here` inline on `retryCount: 0` / `maxRetries: 0` in the local-check branch (line ~288) survived the refactor; the new remote branch above it makes the same choices for different reasons. Neither needs narration — the parameter names are self-documenting.

**6. [Low] No test coverage for the timeout→fallback path in `controller-channel.test.tsx`.**

The tests added in this PR cover the happy path (result arrives before timeout). The fallback path — timeout fires, `null` is resolved, consumer falls back to stripped local check — is the most operationally important branch (disconnected controller) and is currently untested. Not a blocker but worth a follow-up issue or a `TODO` comment on the test file.

moxious:

Context: reviewing the full 16-PR two-tab stack (#1133) bottom-first against `main@fb90805`. This is the largest PR in the stack and one of the cleanest. Posting as context, not a change request yet.

**What's good:**
- `requestId`-correlated pending map with proper timeout cleanup (delete + resolve fallback).
- **Late replies after the timeout are safe** — `settle` no-ops when the pending entry is already gone, so there's no double-resolve.
- The fail-open is sensible: on timeout the controller falls back to a stripped **local** check, so session/permission requirements still gate and only tab-local DOM checks are skipped — not a blind pass. (This also makes `stripTabLocalRequirements` the fallback path, not dead code.)
- `maxRetries = 0` for remote checks (the live tab runs its own retry loop) is correctly reasoned and commented.
- `pendingRef` / `requestSeqRef` refs avoid stale closures; the pending map is bounded, so no leak.

### 1. `requestId` is not globally unique across controller tabs (moderate, concurrency)

`requestId` is a per-instance counter (`req-${requestSeqRef += 1}`). If two controller tabs are open, both mint `req-1`, `req-2`, … A `requirement-result` / `fix-result` is broadcast to every tab and matched on `requestId` alone, so a reply intended for controller A can settle controller B's unrelated pending request — delivering the wrong requirement result to the wrong controller.

This PR's title for #1073 mentions scoping replies to one tab, which may address this; if it doesn't fully (replies still matched by `requestId` only), prefix the id with the transport's `senderId` (a UUID) or generate a UUID per request.

### 2. No fast-path when disconnected; overlapping polls (low)

`isRemote` gates on `controllerChannel != null` but not `connected`, so a check still posts and waits the full 4s before falling back even when no live tab is present. And the controller-mode fragile-requirement poll now runs while blocked — each poll issues a fresh round-trip with no in-flight dedupe, so several `check-requirements` for the same step can be outstanding at once. Both are wasteful rather than incorrect.

### 3. Security surface grew (tracked under the #1056/#1064 blocking item)

`check-requirements` and `fix-requirement` are now remotely triggerable, and `runRemoteFix` → `dispatchFix` performs DOM / navigation mutations on the live tab. Same root cause as the envelope-only validation gap flagged on #1056/#1064 — noting that the remote-execution surface now includes fixes, not just step actions.

### 4. Structural-mirror drift (minor)

The executor posts `checkRequirements`'s `RequirementsCheckResult` directly as the `RemoteRequirementResult` tier-0 mirror. If those shapes ever diverge, the mismatch is silent. A compile-time assignability assertion between the real type and the mirror would catch it.

Tracking IDs `F-1070-1..2` (and `F-1056-1`, `F-1056-4`) in the stack review ledger.

moxious:
## Independent verification — branch tip `12ef932a`

Read-only re-derivation against the source, types, tests, and a clean gate run (not the implementing agent's summary). Verdict per check:

**Branch integrity**
- ✅ Tip `12ef932a` == origin; parent is the fixed #1069 `c09891fe`; 3 commits in order (feat→fix→test)
- ✅ All 3 commits show a Good signature
- ✅ No conflict markers in `src/`; no stale `interactive-readonly-context` imports anywhere

**Security (the requirement round-trip is the high-risk surface)**
- ✅ `onMessage` validates first and dispatches every side-effecting branch off `validated`, never the raw message (`live-tab-executor.ts:256-280`)
- ✅ `evaluateRequirements` / `runRemoteFix` (DOM probe + navigation/DOM mutation) are invoked only inside post-validation branches — no other call site
- ✅ `KIND_VALIDATORS` is `Record<CrossTabMessage['kind'], …>` with all 7 cases incl. the 4 new ones (compile-enforced, not a Partial)
- ✅ `isValidCheckRequirements` / `isValidFixRequirement` require `requestId`/`stepId`/`requirements` as strings and validate every field flowing into `checkRequirements`/`dispatchFix`; forged messages missing/mistyping `requirements` are rejected
- ✅ Adversarial: gate enforces shape exactly — residual risk is "shape validated, sender not authorized" (§6.1), no weaker hole found

**Merge correctness**
- ✅ `CrossTabPayload` kept as the distributive form; #1070's hand-written union dropped; no duplicate; union gained all 4 kinds + `RemoteRequirementResult` mirrors
- ✅ `controller-channel.tsx`: #1065 connection-split survives (`connected` off the channel value/memo), #1067 re-assert-close intact, round-trip helpers are `useCallback`-stable, pending map torn down on cleanup (clearTimeout + resolve(null) + clear)

**Finding mitigations**
- ✅ F-1070-1: `requestId` minted with `crypto.randomUUID()`; old `requestSeqRef` counter gone
- ✅ NEW-1070-1: `attemptCheck` annotated `Promise<RequirementsCheckResult>`
- ✅ F-1056-4: mutual-assignability guard is non-vacuous (collapses to `never` on required-field drift)
- ✅ §6.5: fail-open behavior matches the comment — strips tab-local tokens on timeout; session/permission/role still gate
- ✅ NEW-1068-1: section-level strip documented as deliberately strip-only
- ✅ §6.1 handshake leave-behind deliberately deferred to #1073/#1074 (not added here)

**Tests + gate**
- ✅ Genuine negative tests: forged `fix-requirement` / `check-requirements` dropped at the gate (would fail if the gate were removed); positive counterparts prove handlers fire when well-formed
- ✅ Consumer fail-open test advances fake timers past the round-trip timeout and asserts the step becomes usable
- ✅ `typecheck` clean · `lint` 0 errors / 15 pre-existing warnings · `test:ci` **279 suites / 4522 passed**

**Verdict: PASS** — every security and merge-correctness check passes; no bugs or regressions found.

moxious:
**Fix-pass outcome (re-stacked, gate-green, signed, pushed `12ef932a`)**

**Highest-risk branch.** This is where the remote surface grows from "replay step actions" to "trigger requirement *checks and fixes*" — i.e. navigation and DOM mutation. Routed both new handlers through the validator (`validated`), closing the F-1056-1 hole on the new sink; added 4 new validator cases; made `requestId` globally unique (`crypto.randomUUID`) so replies can't cross-settle between two controller tabs; kept fail-open with a load-bearing rationale comment.

**Independently verified and accepted** — see https://github.com/grafana/grafana-pathfinder-app/pull/1070#issuecomment-4784980851.

leslie-heinzen (approved):

PR #1071, opened 2026-06-17
docs(twotab): document replay pacing and the requirement round-trip

Jayclifford345:
Stacked on #1070. Ends the controller stack with documentation.

Brings `docs/developer/CROSS_TAB_CONTROLLER.md` up to the current state:
- composite steps replay with staged pacing,
- requirements (including tab-local ones) are evaluated and fixed on the live tab via a `requestId`-correlated round-trip,
- the message protocol gained `check-requirements` / `requirement-result` / `fix-requirement` / `fix-result`.

Replaces the stale "only `interactive-step` is routed" v1 scope.

### DIFF
--- a/docs/developer/CROSS_TAB_CONTROLLER.md
+++ b/docs/developer/CROSS_TAB_CONTROLLER.md
@@ -60,9 +60,15 @@ the `pathfinder-cross-tab` channel. Every message carries an envelope
 `timestamp`):
 
 - `step-command` — `{ phase: 'show' | 'do', stepId, action: { targetAction, refTarget, targetValue?, targetComment?, internalActions? } }`.
-  Composite steps (multi-step / guided) carry their ordered sub-actions in
-  `internalActions`; the executor replays those with the same staged pacing a
-  normal multi-step uses (see [Replay pacing](#replay-pacing)).
+  Composite steps carry their ordered sub-actions in `internalActions`. A
+  `multistep` replays with staged pacing (see [Replay pacing](#replay-pacing));
+  a `guided` step runs through the live tab's `GuidedHandler` instead — it
+  highlights each target and waits for the user.
+- `step-complete` — `{ stepId, ok }`, live → controller, signals a composite
+  actually finished so the controller marks completion only then.
+- `step-progress` — `{ stepId, index, total }`, live → controller, reports which
+  internal action a composite is replaying so the controller can animate per-step
+  progress while it runs on the live tab.
 - `heartbeat` — `{ role: 'controller' | 'live' }`
 - `check-requirements` / `requirement-result` — the requirement round-trip
   (controller → live → controller), correlated by `requestId`.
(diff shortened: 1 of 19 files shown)


moxious:

# PR #1071: docs(twotab): document replay pacing and the requirement round-trip

**Verdict:** Request changes (one factual error to fix before merge)

## Findings

**1. [Blocker] Stale filename in the key files table.**

The "Interactive" affordance row still reads:

```
src/components/docs-panel/utils/readonly-tab-open-action.ts
```

The actual file on this branch (and in the repo) is:

```
src/components/docs-panel/utils/controller-tab-open-action.ts
```

This was renamed in an earlier PR in the stack. Any reader following the table will get a 404. Fix before merge.

**2. [Minor] Round-trip flow diagram omits the heartbeat dependency.**

The diagram shows `requestRequirementCheck` firing a `check-requirements` message and the live tab replying. What it doesn't show is that the controller only has `connected: true` once it has received at least one `live` heartbeat — if the controller posts a `check-requirements` before the first heartbeat reply, the message will be received by the live tab (the transport doesn't gate on `connected`), but the controller's `requestRequirementCheck` API is available regardless of connection state. This is probably fine operationally, but the doc implies a tighter coupling ("The reply is a plain `RequirementsCheckResult`; it flows through the same `createRequirementsState` path") without noting that `connected` is purely a UI badge concern and does not gate the round-trip. Worth one sentence of clarification to avoid confusion when someone debugs a "connected badge is off but checks still work" report.

**3. [Nit] "v1 scope" section says "if no live tab answers within the timeout, the check falls back to stripping the tab-local tokens and passing."**

"Passing" is slightly misleading — the fallback runs a full local `checkRequirements` call on the stripped string, which can still fail for non-tab-local requirements (e.g. `section-completed:`). "Falls back to stripping tab-local tokens and evaluating locally" is more precise.

moxious:

**What's good:** the replay-pacing and requirement round-trip sections are clear and match the implementation, including the tier reasoning (live-side evaluation, tier-0 result type, no upward import).

### Fail-open wording (minor accuracy)

The round-trip section says that if no live tab answers, "the check falls back to stripping the tab-local tokens **and passing**." The code actually strips the tab-local tokens and then **still evaluates the remaining** requirements (session/permission) locally via `checkRequirements` — so genuine session/permission failures still gate; it's not a blanket pass. Tightening the wording to "stripping the tab-local tokens and evaluating the rest locally" would match the behavior.

Tracking ID `F-DOCS-3` in the stack review ledger.

moxious:
**Fix-pass outcome (re-stacked, gate-green, signed, pushed `ffb20f26`)**

Docs: reworded the fail-open description ("strip the tab-local tokens and evaluate the rest locally — session/permission requirements still gate"), and noted that the requirement round-trip is independent of the `connected` heartbeat badge.

leslie-heinzen (approved):

PR #1072, opened 2026-06-17
feat(twotab): re-check requirements at click time in controller mode

Jayclifford345:
Stacked on #1071.

Cross-tab requirement state can be stale by the time a controller user clicks **Show me** / **Do it** (e.g. they just closed the nav menu on the live tab). Adds `checker.revalidate()` — it re-runs the requirements check (a live-tab round-trip in controller mode), folds the result into the FSM, and returns whether the action may proceed. The controller-mode Show me / Do it handlers call it first, so a regressed prerequisite **gates** and re-surfaces its warning + "Fix this" instead of acting on a stale "enabled".

Tests: a passing re-check still emits; a failing re-check gates and surfaces "Fix this" without emitting.

### DIFF
--- a/docs/developer/CROSS_TAB_CONTROLLER.md
+++ b/docs/developer/CROSS_TAB_CONTROLLER.md
@@ -60,9 +60,15 @@ the `pathfinder-cross-tab` channel. Every message carries an envelope
 `timestamp`):
 
 - `step-command` — `{ phase: 'show' | 'do', stepId, action: { targetAction, refTarget, targetValue?, targetComment?, internalActions? } }`.
-  Composite steps (multi-step / guided) carry their ordered sub-actions in
-  `internalActions`; the executor replays those with the same staged pacing a
-  normal multi-step uses (see [Replay pacing](#replay-pacing)).
+  Composite steps carry their ordered sub-actions in `internalActions`. A
+  `multistep` replays with staged pacing (see [Replay pacing](#replay-pacing));
+  a `guided` step runs through the live tab's `GuidedHandler` instead — it
+  highlights each target and waits for the user.
+- `step-complete` — `{ stepId, ok }`, live → controller, signals a composite
+  actually finished so the controller marks completion only then.
+- `step-progress` — `{ stepId, index, total }`, live → controller, reports which
+  internal action a composite is replaying so the controller can animate per-step
+  progress while it runs on the live tab.
 - `heartbeat` — `{ role: 'controller' | 'live' }`
 - `check-requirements` / `requirement-result` — the requirement round-trip
   (controller → live → controller), correlated by `requestId`.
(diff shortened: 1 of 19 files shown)


moxious:

# PR #1072: feat(twotab): re-check requirements at click time in controller mode

**Verdict:** Request changes

## Findings

**1. [Bug / Render loop] `checker` in dependency arrays — `interactive-step.tsx`**

`useStepChecker` returns a new object literal on every render (line 998: `return { ...state, checkStep, ... }`). The object is not memoized. Adding `checker` to `handleShowMe` and `handleDoIt` dep arrays means both callbacks recreate on every render, not just on real state transitions. This is the exact problem the existing comment on line 806–808 calls out for `checker.resetStep`: *"including 'checker' would cause infinite re-creation since checker depends on component state."* The fix used there — referencing `checker.revalidate` (the stable individual callback) instead of `checker` — should be applied here too. Alternatively, extract `const { revalidate } = checker` above the callbacks and list `revalidate` in the dep arrays.

**2. [QC8 — narrating comment] `hooks.types.ts`**

```ts
// Re-check requirements at action time and return whether the step may proceed.
revalidate: () => Promise<boolean>;
```

The comment restates the function signature word-for-word. Delete it.

**3. [Design question] `targetAction || 'button'` fallback in `revalidate`**

Phase 3 of `checkStep` also passes `targetAction || 'button'` (line 488), so the fallback is consistent with the polling path. Not a bug, but worth confirming: `'button'` is the default selector strategy used by `checkRequirements`, so steps without an explicit `targetAction` (which usually means they aren't DOM-targeted) get a benign no-op check. Fine as-is; no action required.

**4. [Fail-open behavior, undocumented] `revalidate` returns `true` on transport failure**

When the remote check times out or the live-tab check strips down to a stripped (local-only) check that passes, `revalidate()` returns `true` and the controller proceeds to post the command. This is a deliberate fail-open choice — the PR description mentions preventing acting on *stale enabled state*, not preventing action when state is *unknown*. The behavior is reasonable for UX (the live tab will re-verify on receipt), but the asymmetry is non-obvious: `revalidate` can return `true` both when requirements are met *and* when they could not be determined. A one-line code comment at the call site in controller mode would make the intent clear and prevent a future reader from "fixing" this into a fail-closed gate that breaks the happy path.

moxious:

**What's good:** `revalidate()` re-checks requirements at click time (via the round-trip), folds the result back into the FSM so a regressed prerequisite re-surfaces, and gates the emit — a clean fix for stale cross-tab state. This PR also trims several verbose comments, which is a nice touch.

### 1. Click-gate is wired only into `InteractiveStep` (moderate)

The `await checker.revalidate()` gate is added to the simple step's show/do handlers, but not to `interactive-guided.tsx` or `interactive-multi-step.tsx`. Composites still post their `step-command` without a click-time revalidation, so they can emit on stale cross-tab state. If that's intentional (e.g. composites rely on their own per-substep checks), a one-line note would help; otherwise the gate should extend to them.

### 2. Handlers depend on the whole `checker` object (low)

Both handlers add `checker` to their dependency arrays. `checker` is the `useStepChecker` return value, rebuilt each render, so the `useCallback` identity churns every render. Depending on `checker.revalidate` specifically would keep the handlers stable.

### Note

`revalidate` gates *entry* but not *completion* — the simple-step path still marks completion optimistically after emitting (see #1063). And in controller mode each click now awaits a round-trip (up to the 4s timeout) before the action fires, which is a reasonable correctness/latency trade-off worth being aware of.

Tracking IDs `F-1072-1..2` in the stack review ledger.

moxious:
**Fix-pass outcome (re-stacked, gate-green, signed, pushed `9de6b6ca`)**

Depend on the stable `revalidate` callback in the click handlers (not the whole `checker` object, which is rebuilt each render). **Decision recorded: entry-only click-gate** — composites are *not* re-gated at click time, because a requirement round-trip can cost ~4 s and would compound per click on top of the staged replay; documented at both composite call sites.

leslie-heinzen (approved):

PR #1073, opened 2026-06-17
feat(twotab): run guided as guided, finish composites, scope replies to one tab

Jayclifford345:
Stacked on #1072. Three controller-mode robustness fixes that share the cross-tab plumbing:

- **Guided runs as guided.** A `guided` command replays through the live tab's `GuidedHandler` (highlight + wait for the user) instead of the multi-step auto show→do replay.
- **No early completion.** Composites post `step-complete` back to the controller when the live tab actually finishes, so a guided step is marked done only once the user has walked through every step.
- **Reply scoping.** A controller binds to the first live tab that answers and ignores requirement/fix/completion replies from any other tab, so a second Grafana tab can't answer a requirement check with a different DOM state. Commands stay broadcast (every live tab executes them).

Note: an earlier revision of this PR also addressed commands to one tab via a `targetId`, but that silently dropped Show me / Do it on an id mismatch, so command-level targeting was removed — only the replies the controller trusts are scoped.

Tests cover first-tab binding + ignoring other tabs' replies, guided routing, and `step-complete`. Governance unchanged (vertical=2 lateral=8 barrel=0).

### DIFF
--- a/docs/developer/CROSS_TAB_CONTROLLER.md
+++ b/docs/developer/CROSS_TAB_CONTROLLER.md
@@ -60,9 +60,15 @@ the `pathfinder-cross-tab` channel. Every message carries an envelope
 `timestamp`):
 
 - `step-command` — `{ phase: 'show' | 'do', stepId, action: { targetAction, refTarget, targetValue?, targetComment?, internalActions? } }`.
-  Composite steps (multi-step / guided) carry their ordered sub-actions in
-  `internalActions`; the executor replays those with the same staged pacing a
-  normal multi-step uses (see [Replay pacing](#replay-pacing)).
+  Composite steps carry their ordered sub-actions in `internalActions`. A
+  `multistep` replays with staged pacing (see [Replay pacing](#replay-pacing));
+  a `guided` step runs through the live tab's `GuidedHandler` instead — it
+  highlights each target and waits for the user.
+- `step-complete` — `{ stepId, ok }`, live → controller, signals a composite
+  actually finished so the controller marks completion only then.
+- `step-progress` — `{ stepId, index, total }`, live → controller, reports which
+  internal action a composite is replaying so the controller can animate per-step
+  progress while it runs on the live tab.
 - `heartbeat` — `{ role: 'controller' | 'live' }`
 - `check-requirements` / `requirement-result` — the requirement round-trip
   (controller → live → controller), correlated by `requestId`.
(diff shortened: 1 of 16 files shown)


moxious:

# PR #1073: feat(twotab): run guided as guided, finish composites, scope replies to one tab

**Verdict:** Request changes

## Findings

**1. [Bug] `awaitStepComplete` — no timeout, `isExecuting` can stick forever**

`awaitStepComplete` returns a `Promise<boolean>` that resolves only when the live tab posts `step-complete`. If the live tab crashes, navigates away, or the channel drops mid-execution, the promise never resolves. `setIsExecuting(true)` stays true permanently, locking the button with no user escape. Add a timeout (e.g. via `Promise.race` with a generous ceiling like 60 s) that resolves `false` and clears `isExecuting`. The same `failAllPending()` mechanism that fires on stale re-pairing should also cover this, but that only helps if the paired tab actually goes stale — a tab that hangs without going stale leaves the promise dangling.

**2. [Bug] Optional-chain on `controllerChannel?.awaitStepComplete` silently drops completion**

```tsx
const finished = (await controllerChannel?.awaitStepComplete(renderedStepId)) ?? false;
if (finished) { persistCompletion(); ... }
```

If `controllerChannel` is null when the button is clicked, `finished` is `false`, `persistCompletion` is never called, but `setIsExecuting(true)` was already called before the `await`. The component stays in executing state with no resolution. The outer `finally { setIsExecuting(false) }` does clean up the spinner, but the step silently never completes. The null channel case should be guarded before `setIsExecuting` — fail fast rather than letting the user watch a spinner that clears with no result.

**3. [Design question] Stale-tab answer after re-pairing**

On stale detection: `pairedLiveIdRef.current = null` → `failAllPending()` → next heartbeat re-pairs with a new `senderId`. If the original stale tab recovers and posts a `step-complete` reply, it now arrives from a `senderId` that no longer matches `pairedLiveIdRef.current`, so the reply is ignored. That's correct. However, requirement-check *requests* are still broadcast, and the brief window between `pairedLiveIdRef = null` and the next heartbeat means a requirement reply from the old tab (sent before it went stale) could arrive and be accepted if the old tab happens to respond first. This is a race condition at the pairing boundary rather than a persistent issue; worth a comment acknowledging the window.

**4. [UX] Silent drop when `failAllPending` unblocks `awaitStepComplete` with `ok=false`**

When the paired tab goes stale mid-execution, `failAllPending` resolves `awaitStepComplete` with `false`. The caller checks `if (finished)` and skips `persistCompletion`. The spinner clears (via `finally`), but no message tells the user what happened — the step just resets to its idle state. For a multi-step or guided execution this is confusing. Even a transient "connection to live tab lost — please retry" error state would be meaningful. Not a correctness bug, but a UX gap that will generate support noise.

moxious:

**What's good:** the pairing + completion machinery is strong. `pairedLiveIdRef` binds to the first live tab that answers, ignores the rest, and re-pairs on stale; result / `step-complete` messages are scoped to the paired `senderId`; `failAllPending` resolves both in-flight requests and step-completion waiters on stale/unmount so nothing strands; the `step-complete` ack now gates composite completion (guided + multi-step `await awaitStepComplete` and only persist if finished, with an executing spinner); `awaitStepComplete` supersedes an earlier waiter on re-run; and `GuidedHandler` drives a genuinely interactive guided flow on the live tab.

### 1. `requestId` still collides across two controllers on one live tab (moderate)

Scoping replies by the paired live `senderId` fixes the two-*live*-tab case, but not two *controller* tabs paired to the same live tab. Each controller mints its own `req-1`, `req-2`, … from a per-instance counter, and both accept that live tab's broadcast `requirement-result` / `fix-result`, so a reply can still settle the wrong controller's request. Including the controller's `senderId` in the `requestId` (or echoing the requester id in the reply) closes this.

### 2. `awaitStepComplete` is unbounded and spins the UI (moderate)

`awaitStepComplete` has no timeout (by design, since guided is human-paced), and the component sets `isExecuting(true)` while awaiting. If the user abandons a guided step but leaves the live tab open, heartbeats keep the pairing alive so `failAllPending` never fires — the controller spins indefinitely with no cancel affordance. Worth a cancel path, or a live-side "abandoned" signal.

### 3. Remote guided drops the per-step timeout (moderate)

`runGuided` calls `guidedHandler.executeGuidedStep(action, i, actions.length)` without the `stepTimeout` argument the in-tab guided path passes. Remote guided steps therefore have no per-step timeout and can hang the live side too, compounding #2.

### 4. Pairing is first-responder and spoofable (security)

A result / `step-complete` from any sender can establish the pairing slot if none is set yet, so a forged message could claim the pairing before the real live tab heartbeats and then feed bogus results. This is the same root cause as the envelope-only validation gap flagged on #1056/#1064 — sender authentication would close it.

### Nit

`action.targetAction as GuidedAction['targetAction']` is an unchecked cast from a free string; an unsupported value would slip past the type system.

Tracking IDs `F-1073-1..2` (and `F-1070-1`, `F-1056-1`) in the stack review ledger.

moxious:
**Fix-pass outcome (re-stacked, gate-green, signed, pushed `f624d530`)**

**T1 PART C — security-critical.** Pair-on-heartbeat-only: dropped reply-adoption so the controller binds to a live tab solely on a `heartbeat`, never on a reply — closing the first-responder spoof. The `step-complete` validator is compile-forced; resolved #1064's deferred reverse error channel (`step-complete{ok}` posts on success *and* failure). Added a cancel affordance plus null-channel and lost-connection feedback. The v1 "no sender handshake" residual risk is left as a `TODO(twotab)` at the pairing site (docs half landed at #1074).

leslie-heinzen (approved):

PR #1077, opened 2026-06-17
refactor(cli): extract clean-environment + grafana-health modules

leslie-heinzen:
## Stack Overview
#1076 -> #1077 (this PR) -> #1078 -> #1079 -> #1080 -> #1081

## Summary
Move the --clean docker lifecycle out of e2e.ts into a deep CleanEnvironment module (src/cli/utils/clean-environment.ts) that encapsulates the single "did we start the stack?" bit. Ownership is claimed inside reset() before `up -d`, and teardownIfOwned() keeps the flag private — so the mid-reset teardown-ordering bug is now impossible by construction rather than patched at the call sites.

Also split the Grafana health probe into src/cli/utils/grafana-health.ts, shared by the CLI pre-flight check and the clean reset's readiness poll (avoids a cycle back into e2e.ts).

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/src/cli/commands/e2e.ts
+++ b/src/cli/commands/e2e.ts
@@ -5,7 +5,7 @@
  * into localStorage and verify they load correctly in the docs panel.
  */
 
-import { execSync, spawn } from 'child_process';
+import { spawn } from 'child_process';
 import { existsSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from 'fs';
 import { dirname, isAbsolute, join, resolve } from 'path';
 import { tmpdir } from 'os';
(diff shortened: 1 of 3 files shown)


leslie-heinzen:
Note for the reviewer: right now, I've chucked the new more specific modules from this and subsequent PRs in this stack into `utils/`, but it isn't really the best home for them. 

The plan after this `e2e.ts` refactor is to add an e2e-specific dir for the supporting modules the e2e command needs, something like this:

```
src/cli/
  commands/e2e.ts           ← stays, cli entry point
  e2e/                      ← new feature dir
    clean-environment.ts
    grafana-health.ts
    exit-codes.ts
    playwright-runner.ts
    e2e-runner-contract.ts
    e2e-reporter.ts         ← was already in utils, move it too
    e2e-reporter.test.ts
```

That PR will just be moving files/updating imports.

tomglenn (dismissed):
**Non-blocking.** The teardown-ordering fix looks correct — making `startedByUs`
private and setting it as `reset()`'s first statement makes the old hazard
impossible to reintroduce from a caller. 👍

One gap: nothing tests that invariant, so a future refactor could silently
re-break the ordering with no failing test. Not a regression (this surface had no
tests before), and not worth holding the merge — but if it's cheap, a small
`clean-environment.test.ts` asserting ownership is set *before* `up -d` and that
`teardownIfOwned()` is idempotent would lock it in. Fine as a follow-up too.

tomglenn (approved):

PR #1079, opened 2026-06-17
refactor(cli): encapsulate CLI<->Playwright e2e env-var protocol

leslie-heinzen:
## Stack Overview
#1076 -> #1077 -> #1078 -> #1079 (this PR) -> #1080 -> #1081

## Summary
Introduce src/cli/utils/e2e-runner-contract.ts as the single owner of the environment-variable contract between the `pathfinder-cli e2e` command and the Playwright guide runner. Both sides now reference E2E_ENV keys and the encodeEnvFlag/isEnvFlagEnabled helpers instead of duplicating literal env names and the 'true'/'false' convention, so a rename cannot silently break the cross-process protocol.

Also remove the hardcoded Playwright trace slug: the runner records the actual <outputDir>/trace.zip path (E2E_ENV.TRACE_OUTPUT_FILE) and the CLI reads it, so the trace location no longer breaks if the test title or project name changes.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/src/cli/utils/e2e-runner-contract.ts
+++ b/src/cli/utils/e2e-runner-contract.ts
@@ -1,0 +1,47 @@
+/**
+ * Contract between the `pathfinder-cli e2e` command and the Playwright guide
+ * runner it spawns. The CLI (`src/cli/commands/e2e.ts`) sets these environment
+ * variables; the runner (`tests/e2e-runner/guide-runner.spec.ts`) and its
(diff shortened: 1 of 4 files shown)


tomglenn (dismissed):

tomglenn (approved):
Re-approving after stale-dismissal — content is unchanged from my prior review (mechanical cascade rebase, only the SHA moved). Carrying the GO forward.

PR #1074, opened 2026-06-17
docs(twotab): document tab pairing, guided handling, and step completion

Jayclifford345:
Stacked on the composite/pairing PR. Updates `CROSS_TAB_CONTROLLER.md` for the latest controller behavior: a Tab pairing section, the `step-complete` message, guided running through `GuidedHandler`, and the re-check-at-click-time / complete-on-finish notes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1084, opened 2026-06-17
feat(twotab): stream per-step progress for controller composites

Jayclifford345:
Stacked on #1074.

In controller mode a multi-step / guided step emits to the live tab and awaits completion, so its progress bar sat at 0 until the whole replay finished. The live-tab executor now posts a `step-progress` message before each internal action; the controller subscribes via a new `ControllerChannel.onStepProgress` and advances the step counter as the live tab works through the sequence.

Also removes the temporary completion diagnostics (the round-trip is confirmed working) and trims a few self-evident inline comments.

### DIFF
--- a/docs/developer/CROSS_TAB_CONTROLLER.md
+++ b/docs/developer/CROSS_TAB_CONTROLLER.md
@@ -66,6 +66,9 @@ the `pathfinder-cross-tab` channel. Every message carries an envelope
   highlights each target and waits for the user.
 - `step-complete` — `{ stepId, ok }`, live → controller, signals a composite
   actually finished so the controller marks completion only then.
+- `step-progress` — `{ stepId, index, total }`, live → controller, reports which
+  internal action a composite is replaying so the controller can animate per-step
+  progress while it runs on the live tab.
 - `heartbeat` — `{ role: 'controller' | 'live' }`
 - `check-requirements` / `requirement-result` — the requirement round-trip
   (controller → live → controller), correlated by `requestId`.
--- a/src/components/docs-panel/components/DocsPanelContentArea.tsx
+++ b/src/components/docs-panel/components/DocsPanelContentArea.tsx
@@ -26,6 +26,7 @@ import { Button, Icon, IconButton } from '@grafana/ui';
 import { t } from '@grafana/i18n';
 import { PLUGIN_BASE_URL } from '../../../constants';
 import { testIds } from '../../../constants/testIds';
+import { TWOTAB_CONTROLLER_ENABLED } from '../../../constants/interactive-config';
 import type { LearningJourneyTab, PackageOpenInfo, ContextPanelState } from '../../../types/content-panel.types';
 import type { getStyles as getDocsPanelStyles } from '../../../styles/docs-panel.styles';
 import { isDocsLikeTab, pickGrafanaDocsOpenAction, pickControllerTabOpenAction } from '../utils';
@@ -315,6 +316,9 @@ export function DocsPanelContentArea(props: DocsPanelContentAreaProps): React.Re
                       );
                     })()}
                     {(() => {
+                      if (!TWOTAB_CONTROLLER_ENABLED) {
+                        return null;
+                      }
                       const guideUrl = activeTab.content?.url || activeTab.baseUrl;
                       const action = pickControllerTabOpenAction(guideUrl, activeTab.type);
                       if (!action.shouldShow || !action.controllerUrl) {
--- a/src/components/guide-reader/GuideReaderOverlay.tsx
+++ b/src/components/guide-reader/GuideReaderOverlay.tsx
@@ -18,7 +18,6 @@ import type { RawContent } from '../../types/content.types';
 import { getGuideReaderStyles } from './guide-reader.styles';
 
 interface GuideReaderOverlayProps {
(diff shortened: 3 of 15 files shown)


moxious:

# PR #1084: feat(twotab): stream per-step progress for controller composites

**Verdict:** Request changes (one must-fix, rest nits)

## Findings

1. **[Must fix] Comma expression for `runComposite` return value — correctness is fine, readability is not.**
   `(await runComposite(internalActions, postProgress), true)` is valid JS — the comma operator evaluates both operands and returns the right-hand value — but it reads as a bug at a glance. A reader has to stop and reason about operator precedence to confirm `true` is the expression result, not an argument. The previous `if/else` was clearer. Prefer:
   ```ts
   await runComposite(internalActions, postProgress);
   ok = true;
   ```
   or wrap in a small inline block if you want to keep the ternary shape:
   ```ts
   ok = command.action.targetAction === 'guided'
     ? await runGuided(internalActions, postProgress)
     : await runComposite(internalActions, postProgress).then(() => true);
   ```

2. **[Nit] `total: internalActions?.length ?? 0` — the `?? 0` branch is unreachable.**
   `postProgress` is only constructed inside `if (internalActions?.length)`, so `internalActions` is defined and non-empty at that point. The optional chaining and nullish fallback are dead code. `total: internalActions.length` is sufficient (TypeScript will also confirm this is safe inside the guard).

3. **[Low] Single-subscriber-per-stepId limitation — fine for current callers, but fragile contract.**
   `stepProgressRef` holds one callback per stepId; a second `onStepProgress` call for the same stepId silently overwrites the first. Today `interactive-multi-step` and `interactive-guided` subscribe to different stepIds (each renders its own `renderedStepId`), so there is no collision in practice. But the interface contract doesn't document this. Either add a note — "at most one subscriber per stepId" — or upgrade to a `Set` per key. The silent-overwrite behaviour is the hazard; the unsubscribe guard (`if (ref.get(stepId) === cb)`) correctly handles the stale-closure case, so cleanup is safe.

4. **[Nit] JSDoc on `onStepProgress` interface method is borderline per QC8.**
   `/** Subscribe to a composite step's per-action progress; returns an unsubscribe. */`
   The return type `() => void` already signals "unsubscribe function". "Subscribe to … progress" is what the name says. The only load-bearing word here is "per-action" (distinguishes this from step-level completion). If you keep any comment, trim to: `/** Per-action progress within a composite step; returns an unsubscribe. */` or delete and let the name + signature speak.

5. **[Low] `stripTabLocalRequirements` loses its JSDoc entirely.**
   The removed block was admittedly verbose, but the function signature `(requirements: string | undefined): string | undefined` does not communicate *why* stripping is safe (session/permission requirements are deliberately excluded). The module-level comment above `TAB_LOCAL_REQUIREMENTS` now carries that reasoning, which is the right place for it — so this is fine as long as callers read the module. Acceptable trade-off, flagging only for awareness.

6. **[Confirmed OK] `step-progress` is correctly pairing-gated.**
   The `onMessage` handler reaches the `step-progress` branch only after the `pairedLiveIdRef` check, so a rogue second live tab cannot drive the controller's progress bar. No action needed.

7. **[Confirmed OK] React batching / extra renders.**
   `stepProgressRef.current.get(message.stepId)?.(message.index, message.total)` calls `setCurrentActionIndex` / `setCurrentStepIndex` from outside a React event handler. In React 18 with automatic batching this is batched with any other state updates in the same microtask. Even if it weren't, a single `setState` per progress tick is one render — not a correctness issue, just a render per action step. No action needed.

moxious:

**What's good:** a tidy additive PR. `StepProgressMessage` is added cleanly as the 9th message kind (the discriminated-union envelope has held up well across the whole stack); `onStepProgress` returns an unsubscribe and checks callback identity before deleting; the guided/multi-step components subscribe and unsubscribe in a `finally`; the live executor posts progress before each action. It also trims a few stale comments. No correctness findings.

### Doc drift: `step-progress` is undocumented

This PR adds a new message kind after the last docs PR (#1074), so `docs/developer/CROSS_TAB_CONTROLLER.md` now describes an 8-kind protocol while the code ships 9 — the "Message protocol" section has no `step-progress` entry. Worth a one-line addition there in this PR to keep the doc in sync.

### Minor

- The components register `onStepProgress` *after* posting the `step-command`. It's safe in practice (registration is synchronous and BroadcastChannel delivery is async, so the listener is always in place before any reply arrives), but registering before the post would make the ordering intent explicit.

### Note (re-confirms a #1056 nit)

The `CrossTabPayload` union now lists `StepProgress` before `StepComplete`, while the `CrossTabMessage` union lists them in the opposite order — a small visible symptom of the payload union being maintained by hand in parallel with the message union (flagged on #1056; derivable via a distributive `Omit`).

Tracking ID `F-DOCS-4` (and `F-1056-2`) in the stack review ledger.

moxious:
**Fix-pass outcome (re-stacked, gate-green, signed, pushed `bff1478e`)**

Adds the 9th wire kind, `step-progress`. The `Record<kind>` validator table **forced** its validator case to be authored at compile time — T1 working exactly as designed. The dispatch sits behind the pair-on-heartbeat-only guard, so a forged `step-progress` from an unpaired sender is dropped (a test proves it). Also added the protocol-doc bullet and made composites subscribe to progress before posting the command.

leslie-heinzen (approved):

PR #1087, opened 2026-06-18
feat(e2e): add a test-target resolver for package-aware runs

leslie-heinzen:
## Stack overview

#1086 -> #1087 (this PR) -> #1088 -> #1089 -> #1090 -> #1091

## Summary

Adds `resolveTarget(testEnvironment, { grafanaUrl, currentTier })`, which maps a guide's manifest `testEnvironment` to a concrete Grafana target or a structured skip. Tier rules are delegated to the existing `checkTier` so there's one source of truth: `local`/absent/unknown tier runs against the configured Grafana URL; `cloud` is skipped (`skipped_tier_mismatch` on a local environment, `skipped_no_auth` on a cloud one). Pure function, no I/O — consumed by the remote resolver in #1088.

## What to look at

- `src/cli/utils/e2e-targets.ts` — `resolveTarget` reuses `checkTier` rather than re-deriving tier logic. The cloud branches always skip for now (no credentials), but the returned `ResolvedTarget` keeps optional `username`/`password` fields so the shape is stable when cloud auth is added.

## Why

Routing a guide to the right Grafana (and skipping what can't run here) is a distinct decision from *fetching* the guide. Isolating it as a pure resolver keeps the remote resolver (#1088) and command (#1090) simple, and reusing `checkTier` avoids a second copy of the tier rules drifting from the manifest-preflight path.

## How to verify

Behavior is covered by unit tests (`e2e-targets.test.ts`: local/cloud/unknown tiers, both skip reasons, instance passthrough). The end-to-end effect is observable once #1090 wires it in: run `pathfinder-cli e2e --remote` against a local Grafana and confirm `local`-tier guides run while `cloud`-tier guides are skipped (not failed).

## Breaking changes

None. New module; nothing else changes.

### DIFF
--- a/src/cli/utils/e2e-targets.test.ts
+++ b/src/cli/utils/e2e-targets.test.ts
@@ -1,0 +1,66 @@
+/**
+ * Tests for the e2e target resolver. Covers tier → target mapping and the two
+ * skip reasons, including the forward-looking cloud-on-cloud skip that the
+ * cloud-auth follow-on will turn into a runnable target.
+ */
+
+import { resolveTarget } from './e2e-targets';
+
+const LOCAL_URL = 'http://localhost:3000';
+
+describe('resolveTarget', () => {
+  it('runs local-tier guides against the configured Grafana URL', () => {
+    const target = resolveTarget({ tier: 'local' }, { grafanaUrl: LOCAL_URL, currentTier: 'local' });
+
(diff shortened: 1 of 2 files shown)


moxious (approved):
The `ResolvedTarget` interface seems like it might duplicate some Zod schema definitions, but this use is sufficiently different it makes sense as plumbing here

PR #1088, opened 2026-06-18
feat(e2e): add a remote package resolver for --package and --remote

leslie-heinzen:
## Stack overview

#1086 -> #1087 -> #1088 (this PR) -> #1089 -> #1090 -> #1091

## Summary

Adds the remote package resolver behind `--package <id>` and `--remote`. `resolveRemotePackage` resolves a bare ID via the recommender, then resolves and fetches the target's transitive `depends` prerequisites from the CDN index so the run chains them the same way a local selection auto-includes its prerequisites. `resolveRemoteRepository` fetches the whole CDN `repository.json` and resolves every package. Each guide is mapped to either a runnable guide (target resolved to the local Grafana) or a structured skip — network, validation, tier, and unsupported-type failures become skip outcomes rather than throwing, so a batch degrades gracefully.

## What to look at

- `src/cli/utils/e2e-package.ts`:
  - `discoverPrerequisiteIds` runs the existing planner (`planGuideExecution`) as a pure oracle — a throwaway stub loader lets it report `autoIncludedIds` so we don't re-implement OR-group / `provides` resolution. Only the IDs are used; stub content is discarded.
  - Single-package mode returns the target as `runnable` and its prerequisites as a separate `prerequisites` list (so the planner auto-includes them in #1090, matching local behavior); batch mode returns all guides as `runnable`.
  - Shared `resolveIndexEntry` + `buildGuideOrSkip` keep the single and batch paths consistent. Fallbacks: if the CDN index is unreachable or the target isn't in it, the target runs as its own chain.

## Why

The point of the stack is to test the published ecosystem, which means resolving guides over the network. Mapping every failure to a structured skip (rather than throwing) is what lets `--remote` test a whole repository without one bad package aborting the batch. Reusing the planner for prerequisite discovery avoids duplicating the dependency-resolution logic that already lives in `guide-chains.ts`.

## How to verify

Behavior is covered by unit tests with mocked recommender/CDN/`fetch` (`e2e-package.test.ts`: direct + transitive + `provides` prerequisites, cloud/tier skips, fetch/validation/resolution failures, target-absent fallback). End-to-end (once #1090 wires the flags): `pathfinder-cli e2e --package <local-tier-id>` resolves the guide, pulls in its `depends` prerequisite, and runs the prerequisite first; `--remote` runs the local-tier guides and skips cloud ones.

## Breaking changes

None. New module; consumed by #1090.

## Out of scope

- Cloud-tier execution (credentials / auth) — cloud guides are skipped here.
- Path/journey (`milestones`) expansion — non-`guide` package types are skipped as `unsupported_type`.

### DIFF
--- a/src/cli/utils/e2e-package.ts
+++ b/src/cli/utils/e2e-package.ts
@@ -1,0 +1,315 @@
+/**
+ * Remote package resolution for the e2e CLI.
+ *
+ * Thin orchestration over CLI-local clients:
+ *   - `recommender-resolver` resolves a bare ID via the recommender
+ *     (`GET /api/v1/packages/{id}`) for `--package <id>`; the target's `depends`
+ *     prerequisites are then resolved from the CDN index and chained.
+ *   - `repository-client` fetches the CDN `repository.json` for `--remote` and
+ *     for prerequisite resolution.
+ *
+ * Each resolved package is mapped to either a runnable guide or a structured
+ * skip outcome. Network and validation failures never throw; they become skip
+ * outcomes so a batch run degrades gracefully.
+ */
+
+import { validateGuideFromString } from '../../validation';
+import type { RepositoryEntry, RepositoryJson, TestEnvironment } from '../../types/package.types';
+import { resolvePackageById } from './recommender-resolver';
+import { fetchRepositoryIndex, buildPackageFileUrl, type RepositoryPackage } from '../mcp/lib/repository-client';
+import { planGuideExecution } from './guide-chains';
+import type { LoadedGuide } from './file-loader';
+import { resolveTarget } from './e2e-targets';
+import type { CurrentTier } from './manifest-preflight';
+
+const FETCH_TIMEOUT_MS = 15_000;
+
+/**
+ * Reasons a remote package is skipped before execution. This is the single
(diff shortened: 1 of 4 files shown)


src/cli/utils/e2e-package.ts
```diff
@@ -0,0 +1,326 @@
...
+ *
+ * Out of scope (handled elsewhere or deferred): cloud credentials (cloud guides
+ * are skipped) and path/journey milestone expansion (non-`guide` packages are
+ * skipped as unsupported).
+ */
+
+import { validateGuideFromString } from '../../validation';
+import type { ManifestJson, RepositoryEntry, RepositoryJson, TestEnvironment } from '../../types/package.types';
+import { RecommenderPackageResolver } from '../../package-engine/recommender-resolver';
+import { fetchRepositoryIndex, buildPackageFileUrl, type RepositoryPackage } from '../mcp/lib/repository-client';
```
leslie-heinzen, line 19: 👀 Aware of this import transgression from `../mcp` - plan to resolve soon with a shared module (also [mentioned here](https://github.com/grafana/grafana-pathfinder-app/pull/1086/changes#r3433142474))

leslie-heinzen:
Closing in favor of of #1103 and #1104

PR #1089, opened 2026-06-18
feat(e2e): run remote-resolved packages from the e2e command

leslie-heinzen:
#1103 -> #1104 -> #1089 (this PR)

## Summary

Wires package-aware remote resolution into the e2e runner. The CLI can now resolve published guides by bare package ID (`--package <id>`) or run the whole published CDN repository (`--remote`), then plan/run/report those guides through the existing e2e flow. Local-tier guides run against the configured Grafana URL; cloud-tier guides are logged as skips until cloud auth lands.

## What to look at

- `src/cli/commands/e2e.ts` — `resolveRunInputs` is the load-bearing seam: local files/bundled guides, remote single-package resolution, and remote repository resolution all normalize into the same `{ guides, repoSource, preRunSkipped, packageMetaById }` shape before planning. Trace how remote prerequisites are made available to `planGuideExecution` without being directly selected, so `--package <id>` auto-includes its `depends` chain the same way local guide selection does.
- `src/cli/commands/e2e.ts` — the new status vocabulary distinguishes real failures from non-failing pre-run skips. `validation_failed` is intentionally a test failure; `skipped_no_auth`, `skipped_tier_mismatch`, `resolution_failed`, `fetch_failed`, and `unsupported_type` are logged outcomes that let repository-wide runs degrade gracefully.
- `src/cli/utils/e2e-reporter.ts` — package metadata (`packageId`, `tier`, `instance`, `targetUrl`, `sourceUrl`) and `preRunSkipped` are additive optional fields, so local reports remain backward-compatible while remote reports capture package provenance and pre-execution skip reasons.
- `docs/developer/E2E_TESTING.md` and `docs/design/e2e-test-runner-design.md` — the docs describe the shipped local-tier slice: `--package <id>`, `--remote`, `--repo-url`, `--resolver-url`, cloud skip behavior, and the deferred cloud-auth / path-expansion work.

## Why

This PR delivers user-facing CLI behavior from previously added pieces. The main constraint is keeping remote and local runs on the same execution path once guides are resolved, so dependency ordering, clean-slate behavior, Playwright execution, and JSON reports stay consistent regardless of where the guide content came from.

The rejected alternative was to keep `--package <id>` as a one-guide-only fetch. That would break parity with local guide selection: a published guide with `depends` prerequisites would fail or behave differently than the same guide selected locally. Instead, the resolver returns prerequisites separately and the existing planner performs the auto-inclusion.

## How to verify

1. Run a local-tier published guide by ID:
   `pathfinder-cli e2e --package <local-tier-guide-id>`
   Confirm the guide resolves, its `depends` prerequisite (if any) is auto-included, and the prerequisite runs before the target.
2. Run the repository mode:
   `pathfinder-cli e2e --remote`
   Confirm local-tier guides run and cloud-tier / unsupported package types are reported as skips rather than aborting the batch.
3. Run a cloud-tier package on a local environment:
   `pathfinder-cli e2e --package <cloud-tier-guide-id>`
   Confirm it exits successfully with a structured skip reason (`skipped_tier_mismatch` or `skipped_no_auth`), without launching Playwright unnecessarily.
4. Generate a report:
   `pathfinder-cli e2e --remote --output results.json`
   Confirm tested guides include package metadata and pre-run skips are recorded under `preRunSkipped`.
5. Re-run existing local modes (`./guide.json`, `--bundled`, `bundled:<name>`, and local `--package ./dir`) and confirm behavior is unchanged.

## Breaking changes

None. Existing local and bundled e2e invocations remain supported. The new remote flags and report fields are additive.

## Reversibility

Reversible. This change adds CLI resolution paths and optional report fields; it does not migrate state or alter persisted guide data. Reverting removes the remote modes and returns the reporter to the previous shape. The only compatibility consideration is external consumers that start reading the new optional report fields — they would need to tolerate their absence after a revert.

### DIFF
--- a/src/cli/commands/e2e.ts
+++ b/src/cli/commands/e2e.ts
@@ -5,7 +5,7 @@
  * into localStorage and verify they load correctly in the docs panel.
  */
 
-import { existsSync } from 'fs';
+import { existsSync, statSync } from 'fs';
 import { dirname, isAbsolute, join, resolve } from 'path';
 
 import { Command, Option } from 'commander';
@@ -26,6 +26,7 @@ import {
   writeMultiGuideReport,
   formatMultiGuideSummary,
   type TestResultsData,
+  type PreRunSkip,
 } from '../utils/e2e-reporter';
 import {
   checkTier,
@@ -34,6 +35,15 @@ import {
   type PreflightOutcome,
   type CurrentTier,
 } from '../utils/manifest-preflight';
+import {
+  resolveRemotePackage,
+  resolveRemoteRepository,
+  REMOTE_SKIP_REASONS,
+  type RemoteResolution,
+  type RemoteSkipReason,
+  type ResolvedRemoteGuide,
+  type SkippedPackage,
+} from '../utils/e2e-package';
 import { checkGrafanaHealth } from '../utils/grafana-health';
 import { CleanEnvironment, CLEAN_COMPOSE_PROJECT, CLEAN_GRAFANA_URL } from '../utils/clean-environment';
 import { ExitCode } from '../utils/exit-codes';
@@ -59,11 +69,70 @@ interface E2ECommandOptions {
   clean: boolean;
   cleanReadyTimeoutMs: number;
   repository?: string;
+  /** Resolve and test every package in the CDN repository index. */
+  remote: boolean;
+  /** CDN base URL override for --remote. */
+  repoUrl?: string;
+  /** Recommender base URL for --package <id> resolution. */
+  resolverUrl: string;
 }
 
 const DEFAULT_GRAFANA_URL = 'http://localhost:3000';
+const DEFAULT_RESOLVER_URL = 'https://recommender.grafana.com';
 
-type GuideStatus = 'passed' | 'failed' | 'auth_expired' | 'skipped_prereq';
+/** How guide inputs are resolved for a run. */
+type RunMode = 'local' | 'remote-package' | 'remote-repository';
+
+/** Package metadata attached to a remotely-resolved guide's report. */
+interface PackageMeta {
+  packageId: string;
+  tier?: string;
+  instance?: string;
+  targetUrl?: string;
+  sourceUrl?: string;
+}
+
+/**
+ * A guide's terminal status: either a run outcome (the runner executed or aborted
+ * it) or a remote skip reason (the resolver skipped it before execution). The
+ * skip half is derived from the resolver's `RemoteSkipReason` so the two
+ * vocabularies cannot drift.
+ */
+type GuideStatus = 'passed' | 'failed' | 'auth_expired' | 'skipped_prereq' | RemoteSkipReason;
+
+/** Statuses that count as a test failure (non-zero exit). */
+const FAILURE_STATUSES: ReadonlySet<GuideStatus> = new Set<GuideStatus>(['failed', 'validation_failed']);
+
+/** Pre-run skips (the resolver's skip reasons) recorded in the JSON report; excludes skipped_prereq, which carries step data. */
+const PRE_RUN_SKIP_STATUSES: ReadonlySet<GuideStatus> = new Set<GuideStatus>(REMOTE_SKIP_REASONS);
+
+/** Summary line labels in display order. */
+const GUIDE_STATUS_LABELS: ReadonlyArray<readonly [GuideStatus, string]> = [
+  ['passed', '✅ Passed'],
+  ['failed', '❌ Failed'],
+  ['validation_failed', '❌ Validation failed'],
+  ['auth_expired', '🔐 Auth expired'],
+  ['skipped_prereq', '⊘ Skipped (prerequisite failed)'],
+  ['skipped_tier_mismatch', '⊘ Skipped (tier mismatch)'],
+  ['skipped_no_auth', '⊘ Skipped (no cloud auth)'],
+  ['unsupported_type', '⊘ Skipped (unsupported type)'],
+  ['fetch_failed', '⊘ Skipped (fetch failed)'],
+  ['resolution_failed', '⊘ Skipped (resolution failed)'],
+];
+
+/** Per-guide listing icons. */
+const GUIDE_STATUS_ICONS: Record<GuideStatus, string> = {
+  passed: '✅',
+  failed: '❌',
+  validation_failed: '❌',
+  auth_expired: '🔐',
+  skipped_prereq: '⊘',
+  skipped_tier_mismatch: '⊘',
+  skipped_no_auth: '⊘',
+  unsupported_type: '⊘',
+  fetch_failed: '⊘',
+  resolution_failed: '⊘',
+};
 
 interface GuideRunResult {
   guide: string;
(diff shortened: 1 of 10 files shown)


leslie-heinzen:
Low-severity clanker review items. Considering these for follow-ups:

---

### 1. `PreRunSkip.sourceUrl` is always populated but documented as "when known"

- **Concern**: `guide-schema-and-contracts` / `reversibility-and-one-way-door`
- **Severity**: low · **Confidence**: high · **Reversibility**: reversible (contract is additive and still optional in the type system)
- **Where**: `src/cli/commands/e2e.ts:840-843`, `932-942`; type at `src/cli/utils/e2e-reporter.ts:142-156`

The exported `PreRunSkip` type documents `sourceUrl` as *"Source content.json URL, when known"*. In practice the field is always set:

```ts path="/Users/leslieheinzen/Source/grafana-pathfinder-app/src/cli/commands/e2e.ts" start=840 end=844
function preRunSkipsFromResults(results: GuideRunResult[]): PreRunSkip[] {
  return results
    .filter((r) => PRE_RUN_SKIP_STATUSES.has(r.status))
    .map((r) => ({ id: r.id, reason: r.status, message: r.abortMessage ?? '', tier: r.tier, sourceUrl: r.guide }));
}
```
`r.guide` traces back through `skipToResult` to `skip.sourceUrl ?? skip.id` — so when the resolver did not provide a content URL (e.g. `resolution_failed` from `resolveRemotePackage`, which omits `sourceUrl`), the JSON report's `sourceUrl` ends up as a bare package ID string.

```ts path="/Users/leslieheinzen/Source/grafana-pathfinder-app/src/cli/commands/e2e.ts" start=932 end=935
function skipToResult(skip: SkippedPackage): GuideRunResult {
  return {
    guide: skip.sourceUrl ?? skip.id,
    ...
```
**Why it matters**: This is a public JSON contract field. A downstream consumer that parses it as a URL (for clickable links in a dashboard, link checking, etc.) will hit unexpected non-URL strings on resolution failures — exactly the case where the link would be most useful.

**Suggested action**: Carry the URL distinctly on `GuideRunResult` (e.g. a separate optional `sourceUrl?: string` rather than overloading `guide`), and only set `PreRunSkip.sourceUrl` when the resolver actually produced one. Bonus: it also avoids the small impedance mismatch where `guide: string` is sometimes a path, sometimes a URL, sometimes an id.

---

### 2. Silent precedence between conflicting CLI flags

- **Concern**: `cli-and-e2e-runner`
- **Severity**: low · **Confidence**: medium · **Reversibility**: reversible
- **Where**: `src/cli/commands/e2e.ts:921-929` (mode resolution); `253-288` (input resolution)

`resolveRunMode` silently picks `--remote` > `--package` > local; `resolveGuides` silently picks `--package <dir>` > `--bundled` > positional files. Combinations like `pathfinder-cli e2e --remote --package alerting-101 ./foo.json` produce a remote-repository run with no warning, ignoring two of three inputs. Similarly `--remote` silently makes `--repository <path>` a dead option (the local repo index is never consulted once `repoSource` is set in `buildExecutionPlan`).

**Why it matters**: This concern's one-way doors include *"Structured output changes; new preflight network calls before an expected early exit"* — silent flag dropping makes CI failures harder to diagnose, since two engineers running the "same" command can get different reports depending on which flag wins.

**Suggested action**: Fail fast with a clear message when mutually-exclusive flags are combined (`--remote` with `--package`, `--bundled`, positional files, or `--repository`; `--package <dir>` with `--bundled` or positional files). Or at least emit a warning naming which flag(s) are being ignored.

---

### 3. No direct tests for the new command-side orchestration

- **Concern**: `testing-and-verification`
- **Severity**: low · **Confidence**: high · **Reversibility**: n/a (tests gap)
- **Where**: `src/cli/commands/e2e.ts` — new helpers `resolveRunMode`, `skipToResult`, `buildPackageMetaMap`, `remoteLoadGuideById`, `preRunSkipsFromResults`, `applyPackageMeta`, plus the new exit-code semantics in `exitFromResults`.

The resolver layer is thoroughly tested in `src/cli/utils/e2e-package.test.ts`. The command-side glue — which owns the JSON-report contract additions, the mode-precedence rule, the dual classification of `validation_failed`, and the mapping from `SkippedPackage` to `GuideRunResult` — has no direct tests. The functions are also currently un-exported, so they can't be unit-tested in place.

**Why it matters**: Several findings above (the `sourceUrl` mapping, the dual classification, the mode precedence) are exactly the kind of issue tests on these helpers would catch immediately.

**Suggested action**: Lift the pure helpers (`resolveRunMode`, `skipToResult`, `preRunSkipsFromResults`, `buildPackageMetaMap`, `applyPackageMeta`, the status-label/icon tables) into a `src/cli/commands/e2e-results.ts` (or similar) module and add a small unit test for the JSON-report fields and the precedence/exit rules. Keeps the test surface small and avoids dragging Playwright into the test setup.

---

src/cli/commands/e2e.ts
```diff
@@ -534,6 +667,16 @@ async function runChains(
         blocked.add(planned.id);
         console.log(`\n📚 ${planned.guide.path}`);
         console.log(`   ⊘ Skipped: prerequisite "${blockingDep}" did not pass`);
+        const prereqResultsData: TestResultsData = {
+          guide: { id: planned.id, title: planned.id, path: planned.guide.path },
+          grafanaUrl: options.grafanaUrl,
```
moxious: kinda confused about the difference between different usages in this PR between options.grafanaUrl and the targetUrl.  Hard for tired brain to see all of the trace possibilities of whether/if/when these things can differ, but I would try to quick audit whether they can ever be different and whether inputs/outputs/log messages always match on the basis of the use of both confusingly similar fields in different places this PR.

src/cli/commands/e2e.ts
```diff
@@ -754,18 +894,151 @@ function writeJsonReport(results: GuideRunResult[], options: E2ECommandOptions):
...
+  try {
+    return existsSync(value) && statSync(value).isDirectory();
+  } catch {
+    return false;
+  }
+}
+
+/** Determine how guide inputs should be resolved for this run. */
+function resolveRunMode(options: E2ECommandOptions): RunMode {
+  if (options.remote) {
```
moxious, line 946: if --remote && --package, you get --remote behavior.  One dominates the other.  Probably intentional and probably ok, but odd.  In options parsing, you might want to make it impossible to state conflicting things.

leslie-heinzen: Noted (see point 2 in the claude review comment above; great minds think alike 🤖). I think an error is ideal in that case -- I don't want a run that's intended for a single package chain to accidentally churn an entire repository. I plan to do it in a follow-up.

src/cli/commands/e2e.ts
leslie-heinzen: Yeah, there's definitely some ambiguity here (even for a not-tired brain!). I would say they are misnamed on some of the interfaces, so the meaning is muddied.

What's intended:

* `grafanaUrl`: a grafana url passed in via CLI command option for local tier guides (`--grafana-url` option / default value)
* `targetUrl` is a specific guide's resolved target url, which, once cloud auth is added in the next phase, can be derived from a few places: 

  * `TestEnvironment.instance` from cloud tier package meta
  * `--cloud-url` for cloud tier package with no instance meta
  * `grafanaUrl` for local tier guides
 

Given that, IMO `ResolvedTarget` and `RunGuideOptions` should have `targetUrl` properties to accurately reflect their purpose (instead of `grafanaUrl`). With those renames, if you see `grafanaUrl` in code, it will match up with the definition above. `targetUrl` should also be more clear.

Let me think a little more, and I will do an update shortly!

leslie-heinzen: Ah, and `TestResultsData.grafanaUrl` itself is a bit dated with these changes as well (on a multi-guide run, there are potentially many instance urls), so I need to decide what to do with that.

leslie-heinzen: Address these issues [here](https://github.com/grafana/grafana-pathfinder-app/pull/1089/changes/1123747db17917d26048e5ad43f5421288a314f1) and [here](https://github.com/grafana/grafana-pathfinder-app/pull/1089/changes/ece9499822ba7d0d4af44069674df688e67c0fd8)

moxious (approved):

PR #960, opened 2026-06-03
fix: validate image block URLs

miachillgood:
Fixes #826.

## What changed
- Validate image block URLs before submit, requiring an absolute `http` or `https` URL.
- Show inline field errors for invalid URL syntax, unsupported URL schemes, and preview image load failures.
- Keep the submitted `src` trimmed so valid pasted URLs are normalized before saving.
- Add focused coverage for relative paths, `javascript:` URLs, preview load failures, and successful trimmed URL submission.

## Why
`ImageBlockForm` previously only checked that the image URL was non-empty, so relative paths, typos, and `javascript:` URLs could be accepted silently. The preview also hid broken images without explaining what went wrong.

## Validation
- `npm exec prettier -- --config .prettierrc.js --check src/components/block-editor/forms/ImageBlockForm.tsx src/components/block-editor/forms/ImageBlockForm.test.tsx`
- `npm exec eslint -- src/components/block-editor/forms/ImageBlockForm.tsx src/components/block-editor/forms/ImageBlockForm.test.tsx`
- `npm run typecheck`
- `npm run test:ci -- src/components/block-editor/forms/ImageBlockForm.test.tsx`
- `git diff --check`

PR #1096, opened 2026-06-19
fix: validate image block URLs before submit

miachillgood:
## Summary
- reject relative and non-http(s) image URLs in the image block form
- surface a clear inline error when the preview image fails to load
- trim the submitted URL and add regression coverage for the validation paths

## Validation
- npm test -- --runInBand src/components/block-editor/forms/ImageBlockForm.test.tsx
- npm run typecheck -- --pretty false


miachillgood:
Closing this as redundant: the same ImageBlockForm URL-validation fix was already merged via #960 from the same branch.

Issue #1051, opened 2026-06-16
fix: "Return to my Learning" and "My Learning" should stay in sidebar panel


moxious:
When users finish a guide and click "Return to my Learning" or click the "My Learning" navigation item, Pathfinder currently performs a full-page navigation away from the Grafana UI. Users want to:
1. Remain in the sidebar panel view (not navigate away to a separate Pathfinder page)
2. Preserve scroll position in the guide list when navigating back

**Feedback from field:** Multiple event participants reported context loss while working through guides alongside Grafana UI.

**Ref:** slack discussion, 2026-06-16, Heds

**Related:** Linked to #1002 (FE event feedback epic)

Issue #887, opened 2026-05-15
Reset Guide button doesn't visually reset sections; partial storage clear


moxious:
## Symptom

Clicking the top-level **Reset guide** button (in the docs panel header) reloads the guide content, but sections that were previously marked completed continue to render as completed. Only a hard browser reload (Cmd-R / F5) makes them appear fresh.

Individual-step **Undo** works correctly. The bug is specific to the global Reset guide path.

## Repro

1. Open a guide with at least one section (e.g. `first-dashboard` → "Explore your data").
2. Complete the section (Do section, or step-by-step).
3. Click **Reset guide** at the top of the panel.
4. Observe: the section still shows step checkmarks / completed styling. localStorage is partially cleared but the UI hasn't reacted.
5. Reload the page. Now the section correctly shows as fresh.

## Root cause

Three contributing pieces, all on `main` today:

### 1. `useContentReset` only clears 2 of 4 storage namespaces

`src/components/docs-panel/hooks/useContentReset.ts:53-55`:

```ts
await interactiveStepStorage.clearAllForContent(progressKey);
await interactiveCompletionStorage.clear(progressKey);
```

`InteractiveSection` reads from **four** storage namespaces on mount-only restore (see `src/components/interactive-tutorial/hooks/use-section-persistence.ts`):

| Namespace | Cleared on Reset guide? |
|---|---|
| `interactiveStepStorage` | ✅ |
| `interactiveCompletionStorage` | ✅ |
| `sectionAcknowledgementStorage` | ❌ |
| `sectionCollapseStorage` | ❌ |

The per-section Reset path (`handleResetSection` → `clearAllStorage()`) correctly sweeps all four, so the gap is specific to the global Reset guide path.

### 2. `ContentRenderer` doesn't remount on reset

`src/components/docs-panel/docs-panel.tsx:2500`:

```tsx
<ContentRenderer
  key={activeTab?.currentUrl || stableContent.url}
  ...
/>
```

`useContentReset` calls `loadDocsTabContent(activeTab.id, activeTab.currentUrl || activeTab.baseUrl)` — the **same** URL as before. The key doesn't change, so `ContentRenderer` and all `InteractiveSection` descendants are not remounted. The in-memory React reducer state (`completed: Set<all step ids>`, `acknowledged: true`) survives the click.

This keying strategy was introduced 2026-02-18 in #587 (`fix(interaction): fixed learning journey progress`). The bug has been latent since.

[...]

Issue #822, opened 2026-04-30
InputBlockForm: validation pattern field accepts invalid regex silently


Jayclifford345:
The `Validation pattern` field in `src/components/block-editor/forms/InputBlockForm.tsx` accepts any string. If the user enters an invalid regex (e.g. `[unclosed`), it's saved without complaint and only blows up later at runtime when the input block tries to use it.

Validate that the pattern compiles before submit and surface an inline error to the author.

> Interview task — the body is intentionally light. Ask clarifying questions or expand the scope as you go.

SafinMahmud:
I would like to work on this. We can validate on blur, not onChange or submit, as onChange would create errors halfway through typing a valid regex and would feel noisy. For the error message, should I surface the JavaScript SyntaxError message or keep it generic like the other validation messages in the form? I'll go with generic unless you prefer otherwise.

moxious:
## Agent analysis (leave-behind for next contributor)

**Status: Confirmed.** The "Validation pattern" field in `InputBlockForm.tsx` has no pre-save validation. At runtime, `new RegExp(pattern)` throws, the error is caught silently, and users see a generic "Invalid format" message rather than knowing their regex was malformed.

**Difficulty: Easy**

**Fix:** Add a helper:
```ts
function isValidRegexPattern(pattern: string): boolean {
  if (!pattern) return true; // field is optional
  try { new RegExp(pattern); return true; } catch { return false; }
}
```
Wire `invalid` / `error` props on the pattern `Field` component — the exact same pattern the variable name field in this same component already uses. Also gate the submit button on `!patternRegexError`. ~15 lines, 1 file.

PR #1098, opened 2026-06-21
fix: tighten block editor validation and preserve sidebar navigation

miachillgood:
## Summary
- validate input block regex patterns inline and block submit when the pattern is invalid
- remove guided <-> multistep from the generic type switch so the dedicated converter keeps step description semantics intact
- make interactive sections react to guide-wide reset events by clearing local ack/collapse state immediately
- route `Return to my learning` through Grafana client-side navigation so the sidebar stays mounted instead of forcing a full page reload

## Issue coverage
- addresses #822
- addresses #804
- addresses #887
- addresses #1051

## Validation
- `npx jest --runInBand src/components/block-editor/forms/InputBlockForm.test.tsx src/components/block-editor/utils/block-conversion.test.ts src/components/interactive-tutorial/hooks/use-section-persistence.test.ts src/components/docs-panel/hooks/useContentReset.test.ts`
- `npm run typecheck` currently fails on upstream `src/utils/openfeature.ts` because `disableVisibilityRefresh` is not part of `OFREPWebProviderOptions`; no new type errors introduced by this patch


src/components/interactive-tutorial/hooks/use-section-persistence.ts
```diff
@@ -91,6 +91,28 @@ export function useSectionPersistence({
...
+      const activeContentKey = getContentKey();
+      if (customEvent.detail?.contentKey !== activeContentKey) {
+        return;
+      }
+
+      dispatch({ type: 'CLEAR_ACK' });
+      void clearAckAndCollapseStorage();
+    };
+
+    window.addEventListener('interactive-progress-cleared', handleProgressCleared);
```
moxious: 2 instances; we recently introduced global constants in another file, use StorageEvents.InteractiveProgressCleared from src/lib/event-names.ts instead

src/components/interactive-tutorial/hooks/use-section-persistence.ts
```diff
@@ -91,6 +91,28 @@ export function useSectionPersistence({
...
+    return () => {
+      window.removeEventListener('interactive-progress-cleared', handleProgressCleared);
```
moxious: see comment above, same thing here

moxious (changes requested):
Thank you for the contribution!   See comments -- there is one tiny thing that needs to be addressed but otherwise this looks good to merge.  Make those changes, we'll verify the build is green and get it in.

moxious:
looks like some prettier/lint items to address as well

miachillgood:
Addressed the formatting follow-up and pushed the fix. I also reran the targeted local tests for the touched editor/reset paths; waiting on the refreshed CI now.

miachillgood:
Fixed the lint follow-up from the refreshed run and pushed it. The issue was the submit callback missing  in its dependency list. I also reran the same targeted local Jest coverage for the touched block-editor/docs-panel paths; all 4 suites passed.

miachillgood:
Fixed the lint follow-up from the refreshed run and pushed it. The issue was the submit callback missing patternError in its dependency list. I also reran the same targeted local Jest coverage for the touched block-editor/docs-panel paths; all 4 suites passed.

moxious:
@miachillgood I've allowed the build jobs to re-run, I expect you're green and will pass, but there's still the outstanding substantive comment on the PR to replace those hard-coded strings with references.  That should finish it.

moxious:
We also recently merged a change to fix a reporting step so I'm merging main into your branch to see if we can fix that last build step.

miachillgood:
Addressed the remaining review comment as well: the section persistence listener now uses `StorageEvents.InteractiveProgressCleared` from `src/lib/event-names.ts` instead of hard-coded event-name strings. I reran the targeted section-persistence and related reset/input tests after the change.

moxious (approved):

moxious (approved):
Thanks for the contribution!  Sorry it took us a while to get to this one.  The remaining CI jobs which aren't passing are due to this being an external contribution, but they do things like post bundle size reports to the PR and summarize, they don't block merge

moxious (changes requested):
Oh no, I'm terribly sorry but apparently we need more revisions - I'd forgotten we recently instituted a new org-wide protection where git commits must be signed, otherwise I cannot merge this.

Recommend that you update latest from main, and then revise your commits to be signed, and re-push.  I apologize for the back and forth.

PR #1103, opened 2026-06-22
feat(e2e): add CLI-local recommender client

leslie-heinzen:
## Overview

Part of the package-aware e2e testing changes. #1086 and #1087 are merged to `main`; this PR adds the CLI recommender client that the package resolver will use.

## Summary

Adds `resolvePackageById(resolverUrl, id)` — a small, CLI-local client for the recommender's `GET /api/v1/packages/{id}` endpoint. It resolves a bare package id to its CDN content URL and fetches the package manifest for routing metadata (type, testEnvironment). Failures never throw: network, HTTP, and parse errors return `{ ok: false, message }`, and a missing/unreachable/malformed manifest is non-fatal. Nothing imports it yet — the package resolver in the next PR consumes it.

## What to look at

- `src/cli/utils/recommender-resolver.ts` — the design boundary is the main review point: this deliberately does **not** reuse `src/package-engine`'s `RecommenderPackageResolver`. The packaged CLI (`Dockerfile.cli`) copies only `src/cli`, `src/types`, and `src/validation`, so importing `src/package-engine` breaks `npm run build:cli`. Keeping a CLI-local client preserves that isolation, mirroring the existing `mcp/lib/repository-client.ts`. Also confirm the URL is built via the `URL` API + `encodeURIComponent` (F3), the request is bounded by `AbortSignal.timeout`, and every failure path returns a structured `{ ok: false }` rather than throwing.

## Why

The e2e runner's upcoming `--package <id>` mode needs recommender resolution, but it must stay inside the CLI's packaging boundary. Rejected alternatives: (1) copy `src/package-engine` into the CLI image — risks dragging its `@grafana/runtime`-coupled siblings (e.g. `online-cdn-resolver`) into a Node-only artifact; (2) invert the dependency so package-engine consumes a CLI module — worse, it pulls CLI code into the browser bundle and inverts the tier order. A small CLI-local client (shared contract types already live at Tier-0 `package.types` / `package.schema`) keeps both the dependency direction and the CLI footprint correct. A future shared package-data module may consolidate these clients; this is the documented interim.

## How to verify

1. `npm run build:cli` succeeds — the new module compiles within the CLI's restricted build (`tsconfig.cli.json`), staying inside the `src/cli` + `src/types` + `src/validation` subtree.
2. Unit tests (`recommender-resolver.test.ts`) cover: successful resolve + manifest parse, resolver 404 → failure, network error → failure, manifest-fetch failure (non-fatal — resolves without a manifest), and a package with no manifestUrl.

## Breaking changes

None. New, currently-unconsumed module; the package resolver wires it in next.

### DIFF
--- a/src/cli/utils/recommender-resolver.ts
+++ b/src/cli/utils/recommender-resolver.ts
@@ -33,7 +33,8 @@ export async function resolvePackageById(resolverUrl: string, id: string): Promi
   let response: Response;
   try {
     // SECURITY: build the URL via the URL API and encode the id (F3).
-    const endpoint = new URL(`/api/v1/packages/${encodeURIComponent(id)}`, resolverUrl);
+    const base = resolverUrl.endsWith('/') ? resolverUrl : `${resolverUrl}/`;
+    const endpoint = new URL(`api/v1/packages/${encodeURIComponent(id)}`, base);
     response = await fetch(endpoint.toString(), {
       signal: AbortSignal.timeout(FETCH_TIMEOUT_MS),
       headers: { Accept: 'application/json' },
@@ -46,14 +47,31 @@ export async function resolvePackageById(resolverUrl: string, id: string): Promi
     return { ok: false, message: `HTTP ${response.status} ${response.statusText}` };
   }
 
-  let data: PackageResolutionResponse;
+  let data: Partial<PackageResolutionResponse>;
   try {
-    data = (await response.json()) as PackageResolutionResponse;
+    data = (await response.json()) as Partial<PackageResolutionResponse>;
   } catch (err) {
     return { ok: false, message: err instanceof Error ? err.message : 'Invalid resolution response' };
   }
 
-  return { ok: true, id: data.id, contentUrl: data.contentUrl, manifest: await fetchManifest(data.manifestUrl) };
+  // contentUrl must be be a valid http(s) URL
+  if (typeof data.contentUrl !== 'string' || !isHttpUrl(data.contentUrl)) {
+    return { ok: false, message: 'resolution response did not include a valid contentUrl' };
+  }
+
+  const resolvedId = typeof data.id === 'string' && data.id !== '' ? data.id : id;
+  const manifest = typeof data.manifestUrl === 'string' ? await fetchManifest(data.manifestUrl) : undefined;
+  return { ok: true, id: resolvedId, contentUrl: data.contentUrl, manifest };
+}
+
+/** True only for absolute http(s) URLs; guards against non-http schemes (file:, data:, etc.). */
+function isHttpUrl(value: string): boolean {
+  try {
+    const url = new URL(value);
+    return url.protocol === 'http:' || url.protocol === 'https:';
+  } catch {
+    return false;
+  }
 }
 
 /**
@@ -62,7 +80,7 @@ export async function resolvePackageById(resolverUrl: string, id: string): Promi
  * so the guide can still run under default targeting.
  */
 async function fetchManifest(manifestUrl: string): Promise<ManifestJson | undefined> {
-  if (!manifestUrl) {
+  if (!manifestUrl || !isHttpUrl(manifestUrl)) {
     return undefined;
   }
   let raw: unknown;
--- a/src/cli/utils/recommender-resolver.ts
+++ b/src/cli/utils/recommender-resolver.ts
@@ -1,0 +1,83 @@
+/**
+ * CLI recommender client.
+ *
+ * Resolves a bare package id to its CDN content/manifest URLs via the
+ * recommender's `GET /api/v1/packages/{id}` endpoint, returning the parsed
+ * manifest metadata (type, testEnvironment) the e2e runner routes on.
+ */
+
+import { ManifestJsonObjectSchema } from '../../types/package.schema';
+import type { ManifestJson } from '../../types/package.types';
+
+const FETCH_TIMEOUT_MS = 15_000;
+
+/** Subset of the recommender `GET /api/v1/packages/{id}` response we consume. */
+interface PackageResolutionResponse {
+  id: string;
+  contentUrl: string;
+  manifestUrl: string;
+}
+
+/** Outcome of resolving a bare package id. Failures never throw. */
+export type RecommenderResolution =
+  | { ok: true; id: string; contentUrl: string; manifest?: ManifestJson }
+  | { ok: false; message: string };
+
+/**
+ * Resolve a bare package id through the recommender, then fetch its manifest
+ * for routing metadata. Network, HTTP, and parse failures map to
+ * `{ ok: false }` rather than throwing, so the caller can record a structured
+ * skip outcome.
+ */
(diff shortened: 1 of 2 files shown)


src/cli/utils/recommender-resolver.ts
```diff
@@ -0,0 +1,83 @@
+/**
+ * CLI recommender client.
+ *
+ * Resolves a bare package id to its CDN content/manifest URLs via the
+ * recommender's `GET /api/v1/packages/{id}` endpoint, returning the parsed
+ * manifest metadata (type, testEnvironment) the e2e runner routes on.
+ */
```
leslie-heinzen, line 13: My first (lazy) instinct was to just import from package-engine (`RecommenderPackageResolver`), but after A) reviewing how the package-engine is consumed and B) encountering a Dockerfile build error in CI (`/package-engine` isn't copied into the CLI build docker container!), I decided it's best not to do that. Hence, a new client module.

This file won't live in utils for long, and I'm hoping it will either become or be replaced by some shared module/package that is responsible for exposing a recommender data-fetching client to cross-boundary consumers (CLI, mcp, package-engine, etc).

A few things talking to the recommender now:

* `src/cli/mcp/lib/repository-client.ts` — CDN repository.json + content/manifest fetch (that I temporarily hacked into the CLI e2e code)
* `src/cli/utils/recommender-resolver.ts` — the CLI recommender client I just added
* `src/package-engine/*resolver.ts` — the frontend resolvers
* `src/lib/package-recommendations-client.ts` — the runtime client

src/cli/utils/recommender-resolver.ts
```diff
@@ -0,0 +1,83 @@
...
+ * Resolve a bare package id through the recommender, then fetch its manifest
+ * for routing metadata. Network, HTTP, and parse failures map to
+ * `{ ok: false }` rather than throwing, so the caller can record a structured
+ * skip outcome.
+ */
+export async function resolvePackageById(resolverUrl: string, id: string): Promise<RecommenderResolution> {
+  let response: Response;
+  try {
+    // SECURITY: build the URL via the URL API and encode the id (F3).
+    const endpoint = new URL(`/api/v1/packages/${encodeURIComponent(id)}`, resolverUrl);
```
moxious: I think there might be a conceptual confusion here.  Not sure where the `/api/v1/packages/` is coming from, unless I've forgotten some detail.  This probably doesn't fetch a package off of the CDN that we have, and the URL also doesn't match the k8s-style API that would back the app platform stored guides, so I'm not sure what this is trying to do

src/cli/utils/recommender-resolver.ts
```diff
@@ -0,0 +1,83 @@
...
+ * non-fatal: a missing, unreachable, or malformed manifest yields `undefined`
+ * so the guide can still run under default targeting.
+ */
+async function fetchManifest(manifestUrl: string): Promise<ManifestJson | undefined> {
+  if (!manifestUrl) {
+    return undefined;
+  }
+  let raw: unknown;
+  try {
+    const res = await fetch(manifestUrl, {
```
moxious, line 94: note that for security, the package URL is carefully built by encoding the URL component, but the manifest URL is fully fetched just as-is, meaning this function is subject to injection by the caller

moxious (changes requested):
no problem with the plumbing you're trying to do but I think there's a conceptual confusion.  Double check the end URLs you'd expect out of the CDN URL, (e.g. URLs relative to `https://interactive-learning.grafana.net/packages/repository.json`)

src/cli/utils/recommender-resolver.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  if (!response.ok) {
+    return { ok: false, message: `HTTP ${response.status} ${response.statusText}` };
+  }
+
+  let data: PackageResolutionResponse;
+  try {
+    data = (await response.json()) as PackageResolutionResponse;
+  } catch (err) {
+    return { ok: false, message: err instanceof Error ? err.message : 'Invalid resolution response' };
```
moxious, line 60: If the recommender returns a well-formed-JSON-but-wrong-shape 200 (e.g. {} or { id } with no contentUrl), the function reports `{ ok: true, contentUrl: undefined }` — a success the downstream e2e runner will act on by fetching `undefined`.

src/cli/utils/recommender-resolver.ts
moxious: Because the first argument is an absolute path, a resolverUrl like https://host/api-prefix resolves to https://host/api/v1/packages/{id}, discarding /api-prefix.

leslie-heinzen: I was implementing this approach basically: https://github.com/grafana/grafana-pathfinder-app/blob/main/docs/design/e2e-test-runner-design.md#package-resolution. Specifically, this is for the `--package` arg that accepts an individual package id.

This resolver module won't be used to fetch the package json from the CDN directly. It just resolves where the package lives given a bare id (like `alerting-101`), and returns a response containing the CDN URLs that the `e2e-package` module (subsequent PR) uses to get the content. 

Here's a sample req to our prod endpoint:

```
curl -sS -i -H 'Accept: application/json' \
    https://recommender.grafana.com/api/v1/packages/alerting-101

HTTP/2 200 
access-control-allow-headers: Content-Type, Authorization
access-control-allow-methods: GET, HEAD, OPTIONS
access-control-allow-origin: *
cache-control: public, max-age=300
content-type: application/json
x-request-id: 4f090d569af51520
date: Mon, 22 Jun 2026 21:51:56 GMT
content-length: 243
via: 1.1 google
alt-svc: h3=":443"; ma=2592000

{"id":"alerting-101","contentUrl":"https://interactive-learning.grafana.net/packages/alerting-101/content.json","manifestUrl":"https://interactive-learning.grafana.net/packages/alerting-101/manifest.json","repository":"interactive-tutorials"}
```

The "resolver" url can be defined as an argument, and will fall back to `https://recommender.grafana.com` if not provided.

Does that make sense? Or were you thinking about different approaches?

leslie-heinzen: Maybe we could have a discussion about the app platform-stored guides. I don't really have a clear picture of how that will work.

leslie-heinzen: Good catch. Updated to handle this.

src/cli/utils/recommender-resolver.ts
leslie-heinzen: I added some simple validation following this bit.

src/cli/utils/recommender-resolver.ts
moxious: That does clear it up.  That's the resolution endpoint (which I had forgotten about).  The resolution endpoint is there explicitly to separate knowing a package ID from knowing where it's located on the net (now a recommender responsibility). You're using it for what it was intended to do, I just got confused about CRDs vs. repo URLs.

Revise the comment to at most -- flag this with a comment and it's good to go. This is a coupling point; as such we should document that this depends on a recommender API endpoint (which is fair game, this is documented in its API docs)

moxious (dismissed):

moxious (approved):

PR #1104, opened 2026-06-22
feat(e2e): add remote package resolver for --package and --remote

leslie-heinzen:
## Overview

#1103 (recommender client) -> #1104 (this PR) -> reporter/command/docs. Sits on the package-aware e2e testing changes.

## Summary

Adds the remote package resolver behind `--package <id>` and `--remote`. `resolveRemotePackage` resolves a bare id via the CLI-local recommender client (#1103), then resolves and fetches the target's transitive `depends` prerequisites from the CDN index so the run chains them the same way a local selection auto-includes its prerequisites. `resolveRemoteRepository` fetches the whole CDN `repository.json` and resolves every package. Each guide maps to either a runnable guide (target resolved to the local Grafana) or a structured skip — network, validation, tier, and unsupported-type failures become skip outcomes rather than throwing, so a batch degrades gracefully.

## What to look at

- `src/cli/utils/e2e-package.ts`:
  - `discoverPrerequisiteIds` runs the existing planner (`planGuideExecution`) as a pure oracle — a throwaway stub loader lets it report `autoIncludedIds` so OR-group / `provides` resolution isn't re-implemented. Only the ids are used; stub content is discarded.
  - Single-package mode returns the target as `runnable` and its prerequisites as a separate `prerequisites` list (the command auto-includes them via the planner, matching local behavior); batch mode returns all guides as `runnable`.
  - Shared `resolveIndexEntry` + `buildGuideOrSkip` keep the single and batch paths consistent. Fallbacks: if the CDN index is unreachable or the target isn't in it, the target runs as its own chain.
  - It consumes the recommender client from #1103 (`./recommender-resolver`) and the CDN client from `mcp/lib/repository-client` — no `src/package-engine` import, preserving the CLI packaging boundary.

## Why

The point of the stack is to test the published ecosystem, which means resolving guides over the network. Mapping every failure to a structured skip (rather than throwing) is what lets `--remote` test a whole repository without one bad package aborting the batch. Reusing the planner for prerequisite discovery avoids duplicating the dependency-resolution logic that already lives in `guide-chains.ts`.

## How to verify

1. Unit tests (`e2e-package.test.ts`) with mocked recommender/CDN/`fetch`: direct + transitive + `provides` prerequisites, cloud/tier skips, fetch/validation/resolution failures, and the target-absent fallback.
2. End-to-end arrives once the command PR wires the flags: `--package <local-tier-id>` resolves the guide, pulls in its `depends` prerequisite, and runs the prerequisite first; `--remote` runs local-tier guides and skips cloud ones.

## Breaking changes

None. New module; consumed by the command PR next.

### DIFF
--- a/src/cli/utils/e2e-package.ts
+++ b/src/cli/utils/e2e-package.ts
@@ -36,6 +36,7 @@ export const REMOTE_SKIP_REASONS = [
   'resolution_failed',
   'validation_failed',
   'unsupported_type',
+  'prerequisite_failed',
 ] as const;
 
 /** Why a remote package was skipped instead of run. */
@@ -216,11 +217,74 @@ async function resolveIndexEntry(
  * Enumerate the target's transitive `depends` prerequisites by running the
  * planner as a pure oracle: a throwaway stub loader lets it report
  * `autoIncludedIds` without re-implementing OR-group / `provides` resolution.
- * The stub content is never read, only the discovered ids.
+ * Planner errors (cycles, unresolvable clauses) are surfaced alongside the ids
+ * so callers can cascade them; the stub content is never read.
  */
-function discoverPrerequisiteIds(targetGuide: LoadedGuide, repository: RepositoryJson): string[] {
+function discoverPrerequisites(
+  targetGuide: LoadedGuide,
+  repository: RepositoryJson
+): { ids: string[]; errors: string[] } {
   const stubLoader = (stubId: string): LoadedGuide => ({ path: stubId, content: '{}' });
-  return planGuideExecution({ guides: [targetGuide], repository, loadGuideById: stubLoader }).autoIncludedIds;
+  const plan = planGuideExecution({ guides: [targetGuide], repository, loadGuideById: stubLoader });
+  return { ids: plan.autoIncludedIds, errors: plan.errors };
+}
+
+function prerequisiteFailedSkip(target: ResolvedRemoteGuide, detail: string): SkippedPackage {
+  return {
+    id: target.id,
+    reason: 'prerequisite_failed',
+    message: `Prerequisite(s) did not resolve: ${detail}`,
+    sourceUrl: target.sourceUrl,
+    tier: target.tier,
+  };
+}
+
+function resolutionFailedSkip(target: ResolvedRemoteGuide, message: string): SkippedPackage {
+  return {
+    id: target.id,
+    reason: 'resolution_failed',
+    message,
+    sourceUrl: target.sourceUrl,
+    tier: target.tier,
+  };
+}
+
+function skippedResolution(skipped: SkippedPackage[], repository: RepositoryJson = {}): RemoteResolution {
+  return { runnable: [], prerequisites: [], skipped, repository };
+}
+
+/**
+ * Fetch and classify each discovered prerequisite id against the repository
+ * index, collecting runnable guides and structured skips. A planner-included
+ * id that's missing from the repository becomes a `resolution_failed` skip.
+ */
+async function resolvePrerequisites(
+  ids: string[],
+  repository: RepositoryJson,
+  baseUrl: string,
+  options: RemoteResolveOptions
+): Promise<{ prerequisites: ResolvedRemoteGuide[]; skipped: SkippedPackage[] }> {
+  const prerequisites: ResolvedRemoteGuide[] = [];
+  const skipped: SkippedPackage[] = [];
+  for (const prerequisiteId of ids) {
+    const entry = repository[prerequisiteId];
+    if (!entry) {
+      skipped.push({
+        id: prerequisiteId,
+        reason: 'resolution_failed',
+        message: 'Prerequisite missing from repository index',
+      });
+      continue;
+    }
+    const built = await resolveIndexEntry(prerequisiteId, entry, baseUrl, options);
+    if (built.runnable) {
+      prerequisites.push(built.runnable);
+    }
+    if (built.skipped) {
+      skipped.push(built.skipped);
+    }
+  }
+  return { prerequisites, skipped };
 }
 
 /**
@@ -234,14 +298,8 @@ export async function resolveRemotePackage(id: string, options: RemoteResolveOpt
   // Resolve metadata (URLs + manifest) only; the raw content is fetched in
   // buildGuideOrSkip so the injected guide stays byte-faithful.
   const resolution = await resolvePackageById(options.resolverUrl, id);
-
   if (!resolution.ok) {
-    return {
-      runnable: [],
-      prerequisites: [],
-      skipped: [{ id, reason: 'resolution_failed', message: resolution.message }],
-      repository: {},
-    };
+    return skippedResolution([{ id, reason: 'resolution_failed', message: resolution.message }]);
   }
 
   const built = await buildGuideOrSkip(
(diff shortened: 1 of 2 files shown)


src/cli/utils/e2e-package.ts
```diff
@@ -0,0 +1,315 @@
...
+    return {
+      skipped: {
+        id,
+        reason: 'fetch_failed',
+        message: 'Could not construct content.json URL',
+        tier: entry.testEnvironment?.tier,
+      },
+    };
+  }
+  return buildGuideOrSkip(id, entry.type, entry.testEnvironment ?? {}, contentUrl, options);
```
moxious, line 213: Problem: When a hard depends prerequisite is skipped (fetch_failed, skipped_tier_mismatch, unsupported_type), the target is still returned in runnable (see the reports a prerequisite whose content cannot be fetched test: runnable=[loki-101], skipped=[prom-101]). The target might then execute later without the state its prerequisite was meant to establish.

src/cli/utils/e2e-package.ts
```diff
@@ -0,0 +1,315 @@
...
+
+/**
+ * Enumerate the target's transitive `depends` prerequisites by running the
+ * planner as a pure oracle: a throwaway stub loader lets it report
+ * `autoIncludedIds` without re-implementing OR-group / `provides` resolution.
+ * The stub content is never read, only the discovered ids.
+ */
+function discoverPrerequisiteIds(targetGuide: LoadedGuide, repository: RepositoryJson): string[] {
+  const stubLoader = (stubId: string): LoadedGuide => ({ path: stubId, content: '{}' });
+  return planGuideExecution({ guides: [targetGuide], repository, loadGuideById: stubLoader }).autoIncludedIds;
```
moxious: Problem: planGuideExecution returns { chains, autoIncludedIds, errors }. discoverPrerequisiteIds reads only .autoIncludedIds. Unresolvable depends clauses and dependency cycles push to errors and are dropped here — a cyclic or broken dependency chain vanishes with no signal at resolution time

src/cli/utils/e2e-package.ts
```diff
@@ -0,0 +1,315 @@
...
+  const index = await fetchRepositoryIndex(options.repoUrl);
+  if (!index.ok) {
+    return { runnable: [target], prerequisites: [], skipped: [], repository: {} };
+  }
+
+  const repository = indexToRepository(index.packages);
+  const prerequisites: ResolvedRemoteGuide[] = [];
+  const skipped: SkippedPackage[] = [];
+  for (const prerequisiteId of discoverPrerequisiteIds(target.guide, repository)) {
+    const builtPrereq = await resolveIndexEntry(prerequisiteId, repository[prerequisiteId]!, index.baseUrl, options);
```
moxious: If discoverPrerequisiteIds returns an ID satisfied via planner logic but absent from repository, repository[prerequisiteId] is undefined and resolveIndexEntry will read entry.path, causing a runtime error instead of a structured skip. That contradicts the PR’s stated goal that network/validation failures “never throw” and batch/single runs degrade gracefully.

src/cli/utils/e2e-package.ts
```diff
@@ -0,0 +1,315 @@
...
+
+  // Target itself skipped (tier mismatch, unsupported type, fetch/validation): no chain.
+  if (!built.runnable) {
+    return { runnable: [], prerequisites: [], skipped: built.skipped ? [built.skipped] : [], repository: {} };
+  }
+  const target = built.runnable;
+
+  // Resolve dependencies from the CDN index. Without it fall back to running the target alone.
+  const index = await fetchRepositoryIndex(options.repoUrl);
+  if (!index.ok) {
```
moxious, line 318: convenient, but risky?  What if the target depends on setup guides and produce misleading failures later in execution?   Scoping -- might need a simplifying assumption that everything needed for a given guide has to come from the same repo.  Agree/disagree?

src/cli/utils/e2e-package.ts
```diff
@@ -0,0 +1,315 @@
...
+  resolverUrl: string;
+  /** CDN base URL override for `--remote` batch resolution. */
+  repoUrl?: string;
+}
+
+/** Fetch a URL as raw text with a timeout. Never throws. */
+async function fetchText(url: string): Promise<{ ok: true; text: string } | { ok: false; message: string }> {
+  try {
+    const res = await fetch(url, {
+      signal: AbortSignal.timeout(FETCH_TIMEOUT_MS),
```
moxious, line 97: this is probably not a real problem, but AbortSignal isn't available in all node runtimes.  Might be something to just check quickly, and then pin that the CLI has a minimum node requirement, but it's probably fine.

src/cli/utils/e2e-package.ts
```diff
@@ -0,0 +1,315 @@
...
+ */
+async function buildGuideOrSkip(
+  id: string,
+  type: string | undefined,
+  testEnvironment: TestEnvironment,
+  contentUrl: string,
+  options: RemoteResolveOptions
+): Promise<{ runnable?: ResolvedRemoteGuide; skipped?: SkippedPackage }> {
+  // Only single guides are testable; paths/journeys (milestone expansion) are deferred.
+  if (type && type !== 'guide') {
```
moxious, line 122: see docs/design the package design spec. There are "metapackages" like learning journeys with `milestones`.  Arguably you don't have to handle those, but it might be worth thinking intentionally about how you want to handle them, whether to "descend" into the learning journey structure via ordered milestones, or just "skip".  It's fine to scope this however you like right now, this doesn't have to be the all-powerful first PR, just flagging a boundary of what it will/won't support.

moxious (changes requested):

src/cli/utils/e2e-package.ts
leslie-heinzen: Great point. I have a note specifically about supporting milestone in a future phase but wanted to start focused.

src/cli/utils/e2e-package.ts
leslie-heinzen: Good callout. I think it's a non-issue for this specifically. We have other deps that force us to be on a more recent node versions anyway (e.g., `commander` version requires node >= 20 apparently).

Did a quick search for compat just in case, which says Node 14.17.0 is when AbortSignal became globally available (Nov 2020). That version is end-of-life for a few years, so I think we're okay. 

But maybe a min node version documented makes sense since the CLI tool is shipped independently.

src/cli/utils/e2e-package.ts
leslie-heinzen: Gah! Thank you for spotting this, this is actually a bug in the behavior, which somehow I missed.

It actually deviates from the way guide chains handle `skipped_prereq` at runtime (i.e., if a prereq is skipped, the dependent is also skipped).

This is related to a couple of your other comments, so I will push a fix that addresses them all.

src/cli/utils/e2e-package.ts
leslie-heinzen: I think that makes sense. I want to avoid misleading playwright failures downstream as a result of things not being set up correctly here. This was just me lazily handling what I assumed would be edge-casey.

I'm keen to (for now):
* Surface skip with `resolution_failed` if we can't fetch the repository index for the given repoUrl.
* If the index is fetched by our target isn't in it, also skip early. We won't just fall back to testing the guide that has dependents.

TBH I'm not fully grokking the purpose of the cross-repo resolution that the recommender does, just "seems cool". Would we basically need to duplicate that logic when resolving packages here?

moxious: The real purpose of cross-repo resolution is to basically allow us to assert a universe of package IDs (a given package depends on id: `foo`) and to not bury into package metadata the resolution mechanics. 

Imagine we have 2 package repos (we kinda already do; embedded in pathfinder guides, and the main CDN).  Let's say you want to say that A depends on B (by ID) but they're on different repos.  Without resolution, you're doomed, there's no way to deference B.  That's the use case.  But I'm ok with it being an edge case for now, this design might have been optimistic future-proofing, I can't cite a case right now where "if you don't handle this you're going to fail"

moxious (approved):

Issue #1133, opened 2026-06-23
Epic: Two-tab interactive guides — pop out a guide window and drive it from your Grafana tab


moxious:
## Problem

Users with two monitors want to follow interactive Pathfinder guides on one screen while operating Grafana on the other. Until now, interactive guides required the Pathfinder sidebar to share screen space with the Grafana UI, making dual-monitor workflows awkward. External-source guides could open in new tabs but private and interactive guides could not.

## Solution

A cross-tab interactive controller. The guide opens in a dedicated window (`?controller=1`), sends step commands over `BroadcastChannel`, and the live Grafana tab executes them against the real DOM.

The controller tab acts as a remote control: it dispatches step commands, evaluates requirements against the live tab's DOM via a round-trip message protocol, tracks presence via heartbeat, coordinates sidebar open/close state, animates step progress, and pairs itself to the correct live tab when multiple Grafana tabs are open.

## PR stack (in merge order)

| # | PR | What it adds |
|---|---|---|
| 1 | #1045 | Overlay shell (`GuideReaderOverlay`), `interactive-readonly-context`, deterministic component counters |
| 2 | #1056 | `CrossTabTransport` — injectable `BroadcastChannel` wrapper with start/stop lifecycle and echo suppression |
| 3 | #1062 | `?controller=1` URL param, module routing to mount overlay in controller mode |
| 4 | #1063 | `ControllerChannelProvider`, `useControllerChannel`, `stripTabLocalRequirements`, step command dispatch |
| 5 | #1064 | `installLiveTabExecutor` — live-tab-side handler dispatch (Focus, Button, FormFill, Navigate, Hover) |
| 6 | #1065 | Heartbeat loop, `connected` state, `ControllerStatusBadge` presence indicator |
| 7 | #1066 | Initial `CROSS_TAB_CONTROLLER.md` architecture doc, `AGENTS.md` + `CONTEXT_INDEX.md` updates |
| 8 | #1067 | Sidebar handoff — controller closes Pathfinder sidebar on open, reopens on close/unmount |
| 9 | #1068 | Composite step support — multi-step and guided steps post `internalActions` arrays to the live tab |
| 10 | #1069 | Replay pacing — show→pause→do→settle→inter-step cadence for composite steps on the live tab |
| 11 | #1070 | Requirement round-trip — controller sends `check-requirements`; live tab evaluates against its DOM and replies; 4 s fail-open timeout |
[...]

PR #1147, opened 2026-06-25
fix(cross-tab): unicast executable commands to paired live tab only

### LINKED ISSUE #1140: [security] Two-tab controller: unicast executable commands to paired tab only
## Context

The two-tab controller ships dark pending resolution of this and 5 sibling items.

## Problem

`BroadcastChannel` delivers messages to **all** tabs on the same origin, not just the intended recipient. Executable commands (`step-command`, `check-requirements`, `fix-requirement`) are broadcast to every live Grafana tab. If two live tabs are both paired (even with different controller sessions), both execute the same commands.

The current pair-on-heartbeat-only mitigation requires each live tab to have independently paired before it acts. But in a multi-tab scenario, both could legitimately be paired, and both would execute.

## What needs to happen

Ensure executable commands reach only the intended paired live tab. Options:

1. **Scoped channel name** — include a pairing token in the `BroadcastChannel` name so controller and its paired live tab are on a dedicated channel that other tabs are not subscribed to.
2. **Recipient filter** — embed the target `liveTabId` in the envelope and have each live tab drop messages not addressed to its own ID (requires live tabs to know their own ID, which the heartbeat already establishes).

Option 2 is lower friction; option 1 provides stronger isolation.

## Gate

This issue is a hard gate before `TWOTAB_CONTROLLER_ENABLED` is flipped to `true`.

/label security

moxious:
## Summary

Executable cross-tab commands (`step-command`, `check-requirements`, `fix-requirement`) were being broadcast to every live Grafana tab subscribed to the channel. In a multi-tab scenario, all tabs with the executor installed would simultaneously execute the same DOM mutations — navigating, clicking, and mutating requirements on every open tab instead of just the paired one. This PR closes that gap by adding recipient filtering (Option 2 from #1140).

## What to look at

- **`src/types/cross-tab.types.ts`** — `targetTabId?: string` added to the three executable message interfaces; the three corresponding validators each accept the new optional field via `isOptionalString(message.targetTabId)`. The `CrossTabPayload` distributed type automatically includes it — no third union to maintain.
- **`src/lib/cross-tab-transport.ts`** — new public `getSenderId()` getter exposes the transport's own sender ID so the executor can read it at install time without reflection.
- **`src/integrations/cross-tab/live-tab-executor.ts`** — stores `myTabId = transport.getSenderId()` after `installed = true`, then drops any targeted message (`targetTabId !== undefined && targetTabId !== myTabId`) for the three executable kinds before reaching the dispatch chain. Messages without `targetTabId` are still accepted (broadcast, backward compatible).
- **`src/global-state/controller-channel.tsx`** — `post()` callback now injects `targetTabId: pairedLiveIdRef.current` for the three executable kinds whenever the controller is paired. `requestRequirementCheck` and `requestFix` both go through `post()` via `request()`, so they gain targeting automatically with no additional changes.

## Why

`BroadcastChannel` is an origin-scoped broadcast primitive: every subscriber on the channel receives every message. The Pathfinder two-tab flow installs an executor on every live Grafana tab opened on the same origin. Without recipient filtering, a controller driving tab A would simultaneously trigger `focusHandler.execute`, `checkRequirements`, and `dispatchFix` on tab B — causing unintended navigation and DOM mutations on tabs the user has not paired. The fix embeds the intended recipient's sender ID in the message so every non-paired executor silently discards it.

## How to verify

1. Open two Grafana tabs side-by-side. Enable the two-tab controller in one tab.
2. Start a guided tutorial from the controller. Confirm that only the paired live tab executes the step actions (highlight, button click, navigation) — the second live tab should remain unaffected.
3. Confirm that a live tab receiving a `step-command`/`check-requirements`/`fix-requirement` with a `targetTabId` matching its own sender ID executes normally (no regression in the single-tab case — messages without `targetTabId` are also accepted).
4. In DevTools > Application > BroadcastChannel, verify the channel messages now carry a `targetTabId` field for the three executable kinds when a live tab is paired.

## Breaking changes

None. `targetTabId` is optional on the wire; messages without it continue to be accepted and dispatched as broadcast. The executor only filters when `targetTabId` is explicitly present and does not match.

## Out of scope

- The two-tab feature ships with `TWOTAB_CONTROLLER_ENABLED = false`. This PR closes one of four gate items required before the flag can be flipped. The remaining gate items are tracked in #1133.
- The pairing handshake itself (sender-gesture acceptance, out-of-band nonce) is a separate gate item noted in `CROSS_TAB_CONTROLLER.md` "Known limitations" — not addressed here.

Fixes #1140
Refs #1133


moxious:
build blocked -- this PR needs to be merged & resolved first to unblock: https://github.com/grafana/grafana-pathfinder-app/pull/1150

tomglenn (changes requested):
On the whole looks good, but a few things that need changed.

 The targeting logic itself is right and the validator / `getSenderId` plumbing is clean. My concern is the broadcast fallback, and I think it needs to come out before this merges.

**We shouldn't be adding a backward-compat path for a feature that never shipped.** The two-tab controller is behind `TWOTAB_CONTROLLER_ENABLED = false` and has never been enabled in a release, so there is no older build in the wild that needs to fall back to broadcast. The fallback is defending a population that doesn't exist. It also contradicts what we already say in `cross-tab.types.ts`:

> Same-build assumption: the controller and live tabs are the same plugin build... Cross-version compatibility is not a goal; a mismatched build is out of scope.

So we'd be writing compat code for a case the protocol explicitly rules out a few lines away.

**The bigger issue is that the fallback makes the fix fail-open.** The whole protection rests on the producer always stamping `targetTabId`, and the executor will broadcast-execute anything that omits it. That leaves #1140 reachable. The controller only stamps `targetTabId` when `pairedLiveIdRef.current !== null` (`controller-channel.tsx:96`), and that ref is null in two reachable states:

- before the first live heartbeat, and
- after the paired tab goes stale, where `tick()` resets it to null (`controller-channel.tsx:202`) until a new heartbeat re-pairs.

If an executable command goes out in the gap between going stale and re-pairing (a reactive requirement check, or the user clicking through a step while the live tab missed a heartbeat), it ships with no `targetTabId` and every live executor runs it. That's the exact multi-tab DOM mutation this PR is meant to kill, just transient instead of permanent.

Worth noting the single-tab case doesn't need the fallback either: one live tab heartbeats, the controller pairs to it, and `targetTabId` is stamped and matches. Single-tab already runs through the paired path, not broadcast. The broadcast path only gets hit when unpaired, which is the bug window.

What I'd like instead, fail-closed:

- **Executor:** require `targetTabId` for `step-command` / `check-requirements` / `fix-requirement`, i.e. drop them when it's absent as well as when it mismatches. An untargeted executable command is ambiguous, and dropping it is strictly safer than mutating every tab.
- **Controller:** don't post executable commands while unpaired at all, so an unpaired-window command no-ops cleanly instead of broadcasting.

Given the same-build assumption, there's no cost to this.

Two things that come with it:

- The three `executes ... with no targetTabId (broadcast, backward compat)` tests currently lock in the behavior we want to remove. They should flip to asserting the message is dropped.
- The `CROSS_TAB_CONTROLLER.md` update describes the broadcast fallback as a compat feature, which contradicts the same-build assumption. Update it to match the fail-closed behavior.

Everything else looks good to me: validators as `isOptionalString`, `getSenderId()` instead of reflection, the filter placed after validation, reply filtering untouched. Happy to pair on the fail-closed change if it helps.

moxious:
Closing without merge — superseded by #1153.

This PR added plaintext recipient filtering (a `targetTabId` field on the three executable kinds, injected controller-side and filtered executor-side). #1153 (merged after this PR opened) solved the same problem cryptographically: every executable command is ECDSA-signed with `liveTabId` bound into the signature body, and `verifySignedMessage` rejects any command whose `liveTabId` doesn't match the receiving tab (`pairing-manager.ts:354`). That delivers the unicast guarantee this PR set out to provide, and more strongly — keeping `targetTabId` as well would be a redundant unauthenticated filter beneath the crypto gate.

A merge of `main` produced four conflicts, all tracing to this same root cause (the signing path replaced the plaintext-targeting and heartbeat-pairing model this PR built on). Full rationale in #1140, which #1153 resolves.

PR #1148, opened 2026-06-25
fix(cross-tab): gate live-tab executor pairing behind gesture-to-accept (#1139)

### LINKED ISSUE #1139: [security] Two-tab controller: add sender authentication before re-enable (gesture-to-accept or out-of-band nonce)
## Context

The two-tab controller ships dark (`TWOTAB_CONTROLLER_ENABLED = false`) pending resolution of this and its sibling gate items. Do not flip the flag until this issue is closed by a PR that authenticates the controller session before the live tab accepts remote commands.

## Threat model

`BroadcastChannel('pathfinder-cross-tab')` is same-origin. A non-Grafana external website cannot post to it directly.

The relevant attacker is untrusted JavaScript running in the Grafana origin, such as a malicious or compromised co-plugin, or an XSS payload in a same-origin context. If attacker code is already executing inside the exact live Grafana document, the browser cannot give Pathfinder a strong isolation boundary against direct DOM mutation. This issue is about preventing Pathfinder from becoming an unauthenticated same-origin remote-control sink that any channel participant can use by knowing the channel name and message shapes.

## Problem

The existing mitigations validate message shape and constrain when the sink is installed, but they do not authenticate the sender. Any same-origin channel participant that can produce Pathfinder-shaped messages can currently attempt to pair or send controller-to-live messages that drive the authenticated live tab's DOM, including requirement checks, navigation, clicks, form fill, and requirement fixes.

Pairing on first heartbeat is not sufficient sender authentication. A live-tab user gesture is useful only if it confirms a non-spoofable controller identity or session capability. It is not sufficient to whitelist a self-asserted `senderId`, and a global `window` CustomEvent cannot be the trust signal because same-origin scripts can forge it.

## Required security properties

- The live tab must be unauthorized by default.
- Every side-effecting controller-to-live message must be gated at the live-tab dispatch point. This includes `step-command`, `check-requirements`, `fix-requirement`, and any other message kind that can close/open UI, probe the DOM, navigate, click, fill forms, or mutate state.
- Authorization must bind to a non-spoofable session identity or capability, not to `senderId` alone.
- The user gesture, if used, must confirm that session identity/capability inside trusted module state. It must not round-trip through a forgeable global DOM event as the trust boundary.
- Signed or capability-bearing messages must include enough context to prevent cross-talk and replay, such as session id, target live tab id, message kind, request/run id, and a nonce or monotonic sequence.
- Message shape validation remains required defense in depth, but it is not authorization.
- Routing fields such as `senderId` or `targetTabId` may be useful for correlation, but they are not security boundaries unless protected by the session authentication scheme.

## Acceptable approaches

Either of these can satisfy the issue if implemented with the required properties above:

1. **Gesture-to-accept on the live tab.** The live tab shows a prompt when an unknown controller session asks to pair. Accepting confirms a cryptographic public key, challenge-response result, or equivalent unguessable capability. Commands are accepted only from that verified session.
2. **Out-of-band nonce/capability at controller open.** The live tab and controller exchange a one-time capability through a path that a random same-origin channel participant cannot guess. Commands are accepted only when they prove possession of that capability.

Workshop mode already has useful Web Crypto building blocks (`generateSessionKeyPair`, `generateNonce`, `signChallenge`, `verifyChallenge`) and a challenge-response pattern. Reuse the pure crypto helpers where practical, but move them to an import-graph-appropriate lower tier such as `src/security/` or `src/lib/`. Do not copy the workshop PeerJS/session protocol wholesale.

## Done criteria

- The live tab drops unsigned, invalidly signed, unpaired, wrong-target, stale, or replayed side-effecting controller messages without touching the DOM.
- Tests cover negative cases for each side-effecting kind, especially `step-command`, `check-requirements`, and `fix-requirement`.
- Tests cover the positive paired flow from controller handshake through accepted command execution.
- Reject/cancel has explicit behavior and does not silently leave a sender partially trusted.
- The feature flag remains off until this issue and the sibling gate items are resolved.

/label security

moxious:
## Summary

Before a controller tab can drive interactive steps on the live Grafana tab, the live tab now requires an explicit user gesture to accept the pairing. Previously, any same-origin script (or XSS payload) could forge a `heartbeat` with `role: controller`, claim the pairing slot, and then dispatch `check-requirements` / `fix-requirement` commands — which run navigation and DOM mutations on the authenticated Grafana tab. This PR closes that gate.

## What to look at

- **`src/integrations/cross-tab/live-tab-executor.ts`** — the core change: unknown senders are no longer auto-responded to. Instead, a `pathfinder-pairing-request` CustomEvent is dispatched on `window`; the executor only sends a live heartbeat after `pathfinder-pairing-accepted` arrives. Once a sender is accepted its subsequent heartbeats are auto-responded to without re-prompting (tracked in `acceptedSenderIds`). The `handlePairingAccepted` listener is removed on teardown, so post-uninstall acceptance events have no effect.

- **`src/integrations/cross-tab/PairingRequestBanner.tsx`** — the thin React root that bridges `window` CustomEvents to the user. It renders a fixed-position `role="alertdialog"` banner (bottom-right, z-index 99999) via `createPortal` to `document.body`, using the same standalone-root `useGrafanaTheme` pattern as `GuideReaderOverlay` so it stays in sync with Grafana's theme toggle. Accept dispatches `pathfinder-pairing-accepted`; Reject dispatches `pathfinder-pairing-rejected`. A second pairing request from a different sender replaces (not stacks) the current prompt. The Accept button receives `autoFocus` on mount; Escape on the dialog root triggers reject — both required by the `alertdialog` ARIA role.

- **`src/module.tsx`** — the live-tab code path now mounts `PairingRequestBanner` as a parallel lightweight root (`pathfinder-pairing-banner-root`) **before** calling `installLiveTabExecutor()`, so the `pathfinder-pairing-request` listener is live before the transport starts. The banner is mounted only once (id-guarded) and imported concurrently with `create-root-compat` via `Promise.all`.

- **`src/global-state/controller-channel.tsx`** — the `// PAIRING SITE. TODO(twotab)` comment is updated to reflect that the gesture-to-accept is now implemented. T1 PART C (pairing only on heartbeat, never from a reply) is unchanged.

## Why

The `BroadcastChannel` is same-origin, so any co-plugin or XSS payload running in the Grafana tab can send a `heartbeat` with `role: controller`. The old pair-on-first-heartbeat behavior meant the first sender claimed the pairing slot unconditionally — authenticating the channel without any user involvement. Once paired, that sender could issue `check-requirements` and `fix-requirement` messages, which probe and mutate the authenticated live tab's DOM.

The gesture must happen on the **live tab** (not the controller tab) because that is the tab whose DOM will be mutated. Moving the prompt to the controller tab would let the forged sender bypass it entirely by operating on a different tab.

Rejected alternative: an out-of-band nonce at controller-open time (exchanged over `window.opener` or URL fragment). This works when the controller is opened via `window.open`, but breaks for manually navigated controller URLs and adds per-launch coordination complexity. The gesture approach is simpler, more robust, and requires no coordination between the two tabs beyond the existing BroadcastChannel.

## How to verify

1. Open Grafana with Pathfinder mounted. In DevTools console, run:
   ```js
   const ch = new BroadcastChannel('pathfinder-cross-tab');
   ch.postMessage({ source: 'pathfinder', senderId: 'forge-test', timestamp: Date.now(), kind: 'heartbeat', role: 'controller' });
   ```
   Confirm: the Accept/Reject banner appears in the bottom-right of the Grafana tab. Confirm: no `heartbeat` reply is sent to the channel before you click Accept.

2. Click **Accept** on the banner. Confirm: a live heartbeat is now posted to the channel. Send the forged heartbeat a second time — confirm the banner does not reappear (auto-respond path).

3. Open the real two-tab controller flow (when `TWOTAB_CONTROLLER_ENABLED` is flipped on). Confirm the pairing prompt appears on the live Grafana tab, not the controller tab, and that accepting it completes the handshake (controller status badge shows "Connected").

4. Click **Reject** on the banner. Confirm: no live heartbeat is posted, and a second controller heartbeat re-triggers the prompt.

5. After accepting, send a heartbeat from a **different** senderId. Confirm: the prompt reappears for the new sender.

6. With the banner visible, press **Escape**. Confirm: the banner dismisses and no live heartbeat is posted.

## Breaking changes

None. `PairingRequestBanner` and the `acceptedSenderIds` gate are additive. The feature ships dark (`TWOTAB_CONTROLLER_ENABLED = false`), so no live user is affected. The existing behavior of the live-tab executor for all non-heartbeat message kinds is unchanged.

## Out of scope

- **Post-gesture command injection:** `acceptedSenderIds` gates heartbeat replies only — `check-requirements`, `fix-requirement`, and `step-command` messages from *any* BroadcastChannel sender are executed without a sender check once any pairing is established. Separately, the live heartbeat is an untargeted broadcast; any listener on the channel receives it without needing to know which controller was accepted. Full per-message sender authorization (checking all command kinds against `acceptedSenderIds`, and targeting the live heartbeat to the accepted controller) is deferred to a future PR.
- The controller-tab side of the pairing UI (status indicators, timeout feedback) — out of scope for this PR, tracked in #1133.

Fixes #1139
Refs #1133


moxious:
## Review suggestions

- **[blocking] Reorder `module.tsx` mount sequence** — move the `PairingRequestBanner` import and mount *before* the `installLiveTabExecutor()` call, so the banner's `pathfinder-pairing-request` listener is live before the executor starts the transport. The first controller heartbeat can otherwise arrive in the async gap and be silently dropped.

- **[suggestion] Restore the `queue` one-liner comment** in `live-tab-executor.ts`, stripped of the ticket reference: `// Serialized: concurrent replays race on shared highlight state.` The ticket reference (`F-1069-1`) was correctly removed, but the *why* (concurrent replays racing on highlight state) is non-obvious from the variable name alone.

- **[suggestion] Add autoFocus + Escape-to-reject to `BannerInner`** — autoFocus the Accept button on mount and add `onKeyDown={(e) => { if (e.key === 'Escape') handleReject(); }}` to the root div. The `alertdialog` role makes AT expectations that the implementation doesn't meet without these.

- **[suggestion] Clarify the untargeted live heartbeat in the out-of-scope note** — the PR description says the post-pairing risk is "a compromised sender that *learned the accepted ID*," but the actual race is any BroadcastChannel listener present at broadcast time (no ID knowledge needed). Either update the known-limitations note to name this, or consider adding the accepted `senderId` to the live heartbeat payload so the controller side can correlate before binding.

moxious:
build blocked -- this PR needs to be merged & resolved first to unblock: https://github.com/grafana/grafana-pathfinder-app/pull/1150

tomglenn (changes requested):
## Review: NO-GO

I'm blocking this one, and I want to be clear that it's a hard no rather than a "fix on follow-up." The PR is presented as closing #1139, but as written it does not close that boundary at all. The gesture is gating the wrong message kind, and the control it does add is bypassable three separate ways. The only reason this is safe to leave on the branch is that the feature is dead code behind `TWOTAB_CONTROLLER_ENABLED = false`. It must not merge under the #1139 framing, and the flag must not be flipped while the code looks like this.

One framing note first, because it changes how we reason about the whole thing. `BroadcastChannel('pathfinder-cross-tab')` is same-origin. A non-Grafana external website cannot reach the channel at all. The actual threat #1139 is about is an untrusted script running inside the Grafana origin: an XSS payload or a malicious/compromised co-plugin. That's the attacker we have to keep out, and it's the attacker this PR doesn't stop.

### Why it's a no-go

**1. The gesture gates the heartbeat, not the commands.** In `live-tab-executor.ts:320-333`, the three dangerous kinds dispatch on shape validation alone:

- `step-command` -> `runStepCommand`, no sender check
- `check-requirements` -> `evaluateRequirements`, no sender check
- `fix-requirement` -> `runRemoteFix` (navigation + DOM mutation), no sender check

Only the `heartbeat` branch consults `acceptedSenderIds`. So an attacker skips pairing entirely and posts a single `fix-requirement` directly. The banner never renders, the gesture never runs, and the command executes against the authenticated tab. The "Out of scope" note undersells this: it isn't "ungated once paired," it's ungated with no pairing required at all.

**2. `senderId` is self-asserted, so even the heartbeat gate is spoofable.** `cross-tab-transport.ts:21-27` and `:66-72` stamp a sender-chosen `senderId`, and validation only checks `typeof === 'string'` (`cross-tab.types.ts:163`). Heartbeats broadcast the accepted controller's id in the clear, so an attacker reads it and sets their own envelope `senderId` to match. Copying the `acceptedSenderIds` check onto the command branches would still leave a gate that authenticates a string the attacker fully controls.

**3. The "user gesture" is a forgeable `window` CustomEvent.** `live-tab-executor.ts:301-306` treats `pathfinder-pairing-accepted` as the trust signal. Any same-origin script can run `window.dispatchEvent(new CustomEvent('pathfinder-pairing-accepted', { detail: { senderId: 'attacker' } }))` and self-accept with no banner and no click. A same-origin DOM event cannot represent a trust decision against a same-origin attacker.

Each of these defeats the control on its own.

### Supporting issues (worth fixing regardless)

- `module.tsx:174-184` has a double-mount race that the sibling controller path at `:146-151` was explicitly hardened against. The id-guard reads `getElementById` but the append happens after two awaited dynamic imports, so two `plugin.init` calls can both pass the guard. Claim the id synchronously like the controller path does.
- The banner root is created and appended but never unmounted or removed. Page-scoped leak, low impact, but no teardown symmetry.
- Reject is a no-op. `pathfinder-pairing-rejected` has no listener anywhere, so clicking Reject just hides the banner and takes no protective action on the executor.

### We already solved this in workshop mode

This is the part I most want us to sit with before redoing it from scratch. Workshop mode already authenticates peers properly, and this PR is a weaker reinvention of it. `src/integrations/workshop/session-crypto.ts:57-109` has transport-agnostic, zero-dependency Web Crypto primitives (`generateSessionKeyPair`, `generateNonce`, `signChallenge`, `verifyChallenge`), with the challenge-response reference pattern in `session-manager.ts:237-253` and `:524-569`. That scheme binds trust to an unspoofable keypair instead of a self-declared string, which is exactly what's missing here.

Two caveats so we reuse it correctly rather than naively:

- Promote the crypto, not the protocol. `session-crypto.ts` sits in tier-3 `integrations/`, and cross-tab transport is tier-1 `lib/`. Reuse means moving those pure helpers down to `lib/` or `security/`, since the import graph forbids the other direction. The PeerJS, join-code, and attendee machinery does not transfer; that's remote-session plumbing irrelevant to a same-origin channel.
- Workshop's model is one-sided. Attendees verify the presenter, but there's no per-message attendee authentication. Cross-tab needs the controller -> live direction authenticated per message, which is more than workshop does today. So we copy the building blocks and the handshake pattern, not the protocol wholesale.

### What "ready" looks like

1. Move the authorization check onto every side-effecting kind rements`, `fix-requirement`), dropping silently otherwise. Thisalone closes the largest hole.
2. Bind authorization to an unspoofable identity. Reuse the wored envelopes verified at the dispatch points), with the singleuser gesture confirming the key exchange rather than whitelisting a `senderId`.                                                                        3. Move the trust signal off the `window` CustomEvent. The bannide the module closure, never round-trip through a forgeableglobal event.                                                                                                                                          4. Add tests that assert the security property: each command kited or unsigned sender. The current suite only covers the bannerUI and the heartbeat reply, and the pre-existing tests actively assert that commands from an unaccepted `controller` sender execute. Those lock in the vulnerability and need to flip.

The UI work itself is clean and I don't want that effort lost. the whole point of this PR, and until 1 through 4 land, #1139 isstill open.

PR #1152, opened 2026-06-25
refactor(cli): modularize the e2e command into src/cli/e2e

leslie-heinzen:
## Summary

Reorganizes the `e2e` CLI command and `src/cli/utils/` into focused modules under a dedicated `src/cli/e2e/` feature directory. Behavior is unchanged (same flags, exit codes, skip/fail/pass outcomes, and JSON report), while some of the command's pure logic moves into a tested results-model module and a console-reporter module.

## What to look at

- **Module boundary** — `src/cli/e2e/` now holds the e2e-only modules; `src/cli/utils/` stays the shared layer (`file-loader`, `graph-cycles`, `output`) that the other commands (`validate`, `build-*`) and `e2e/` both depend on. Confirm `e2e/` only imports "down" into `utils/`, never the reverse.
- **New tested results model** ([`src/cli/e2e/e2e-results.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/92e6f6a580b9418a7755de0b12ce746e7c36531d/src/cli/e2e/e2e-results.ts)) — the command-side glue the PR #1089 review flagged as untested: run-mode precedence, `SkippedPackage`→`GuideRunResult` mapping (including `validation_failed` counting as both a skip reason and a failure), pre-run-skip report fields, and the exit-code rules.
- **New console reporter** ([`src/cli/e2e/e2e-console-reporter.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/92e6f6a580b9418a7755de0b12ce746e7c36531d/src/cli/e2e/e2e-console-reporter.ts)) — owns all terminal output plus the optional JSON-report write, and takes narrow render-focused inputs (e.g. `RunConfigurationView`) so it no longer depends on the command's full option bag.
- **`summarizeSteps`** in the results model replaces five parallel `reduce` passes with one switch-based tally — behavior-identical, but removes a copy-paste surface where a step status could be mis-bucketed.
- **Cross-process contract** ([`tests/e2e-runner/guide-runner.spec.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/92e6f6a580b9418a7755de0b12ce746e7c36531d/tests/e2e-runner/guide-runner.spec.ts)) — the Playwright runner imports `e2e-runner-contract` and `e2e-reporter` from the relocated `src/cli/e2e/` path, so this move spans `src/` and `tests/`; the env-var protocol between the command and the spawned runner is the thing to confirm still resolves.
- **Reviewer gotcha** — commits 1–3 are intentionally non-compiling intermediates (history is split per refactor step); only the branch tip typechecks. Review the final tree rather than checking out intermediate commits.

## Why

`src/cli/utils/` had become a junk drawer for e2e supporting modules: ~13 e2e-only modules sat alongside helpers shared by the other CLI commands, so the e2e surface wasn't legible and shared-vs-e2e ownership was implicit. 

Separately, the PR #1089 review noted the command-side orchestration (mode precedence, skip→result mapping, exit codes) had no direct tests because the helpers were un-exported inside the command file. This lifts those helpers into a tested module, isolates console printing in a reporter module that mirrors the existing JSON reporter, and makes the e2e cluster an explicit feature directory.

## How to verify

1. Run the `e2e` command against a local or bundled guide and confirm the summary output and process exit code match `main` (pass → 0, validation/test failure → 1, configuration error → 2, auth expiry → 4).
2. Run a `--package <id>` / `--remote` resolution that yields a mix of runnable and skipped packages; confirm the skip-vs-fail classification and the JSON report's `preRunSkipped` entries are unchanged — in particular that `validation_failed` still counts as a failure.
3. Run one guide end-to-end through Playwright (`tests/e2e-runner`) and confirm the command still spawns the runner — this exercises the relocated `e2e-runner-contract` env-var protocol that crosses `src/` ↔ `tests/`.

## Breaking changes

None. Pure restructure: identical CLI flags, exit-code semantics, skip/fail/pass classification, and JSON report shape. Every module move is a content-preserving rename (git reports R098–R100); only import paths changed.

## Out of scope

- Narrowing the remaining engine-side functions (`runChains`, `buildExecutionPlan`, `runPreflightChecks`) off the whole `E2ECommandOptions` bag — deferred to a follow-up so this PR stays a move + extraction.
- Dropping the now-redundant `e2e-` filename prefixes (e.g. `e2e/reporter.ts`) — left for a follow-up to keep this diff rename-clean and git-blame intact.

Refs #1089.


docs/design/e2e-test-runner-design.md
```diff
@@ -1174,7 +1174,7 @@ For multi-guide reports (batch or path), each guide entry includes these package
 
 ### Dependency-aware chaining
 
-The runner builds a dependency-aware execution plan from a `repository.json` index before running any guide (`src/cli/utils/guide-chains.ts`). For `--bundled` runs it defaults to `src/bundled-interactives/repository.json`; an optional `--repository <path>` flag points at a custom index.
+The runner builds a dependency-aware execution plan from a `repository.json` index before running any guide (`src/cli/e2e/guide-chains.ts`). For `--bundled` runs it defaults to `src/bundled-interactives/repository.json`; an optional `--repository <path>` flag points at a custom index.
```
moxious, line 1177: not blocking on this PR, but this file is probably pretty long in the tooth and may have been superceded by the RFC we already discussed.  Consider looking with a skeptical eye at docs/design/(anything e2e related) as it's likely to have been overcome by the RFC.  It's possible that a big "rip out" PR could be warranted, but not in scope here.

moxious (approved):
LGTM.  One non-blocking comment.

docs/design/e2e-test-runner-design.md
leslie-heinzen: Yeah agree -- it's not _super_ useful at this point, just touching it while it's here and we're mid-build

PR #1153, opened 2026-06-25
fix(cross-tab): authenticated controller pairing with ECDSA challenge-response

### LINKED ISSUE #1139: [security] Two-tab controller: add sender authentication before re-enable (gesture-to-accept or out-of-band nonce)

moxious:
## What this PR does

Defines and enforces Pathfinder's **cross-tab controller pairing protocol**: how a popped-out controller tab proves it was opened by the live Grafana tab, how the user consents, how the accepted controller's commands are authenticated and replay-protected, and how the pairing behaves across rejection, reconnect, and shutdown.

Before this work the live tab's executor had no cryptographic gating — any script on the same origin could broadcast `step-command` / `check-requirements` / `fix-requirement` / `sidebar-handoff` and have it run against the user's authenticated Grafana DOM. This PR makes the live tab **unauthorized by default**: every side-effecting message is dropped until the user accepts a pairing backed by a verifiable session keypair, and every command after that is individually signed.

Supersedes #1148 (right idea — a consent prompt — but authorized on the spoofable `senderId`). Fixes #1139.

## The trust boundary

Authority to drive the live tab **is possession of the controller's ECDSA P-256 private key** — nothing else. The key is:

- minted per session inside the controller tab, **non-extractable**, never exported (only the public key crosses the wire);
- bound to the live tab at pairing time and unrecoverable once the controller tab closes.

The protocol has four enforced layers, all on the **controller → live command direction**:

| Layer | Guarantee |
|---|---|
| **Launch binding** | The live tab mints a one-time launch (`pairingId` + HMAC secret + 6-digit code) and embeds it in the controller tab's URL fragment. The controller proves possession via an HMAC over the canonical challenge `{pairingId, publicKeyB64, sessionId}`. The live tab accepts only challenges matching a registered, unexpired launch. |
| **Consent gesture** | Pairing is accepted only on a trusted user gesture; programmatic/untrusted accepts are no-ops. Competing valid challenges fail closed; rejected sessions stay suppressed; expired challenges are retryable. |
| **Signed commands** | Every side-effecting message is ECDSA-signed over a **canonical JSON of the entire message body** plus `sessionId`, `liveTabId`, `sigNonce`, and `sigTs` — so the action payload, target tab, and session are all bound into one signature. |
| **Replay / staleness** | `verifySignedMessage` enforces a session-wide `(sessionId:sigNonce)` replay ledger and a ±30s `sigTs` window, on top of the signature + session/liveTab checks. |

Per-kind structural validation (`validateCrossTabMessage`, at both the receive gate and the executor sink) and `pathfinderEnabled` install-gating remain as **defense-in-depth** beneath the auth gate.

### Knowingly out of scope (accepted limitations)

- **Unsigned replies.** `requirement-result` / `fix-result` / `step-progress` / `step-complete` / `live` heartbeat are not signed — the controller scopes them by `senderId`, which is forgeable. A same-origin script can spoof reply *content* (mislead the controller's UI) but **cannot** issue commands or drive the live tab. Signing replies would need a second keypair to defend against an attacker who already has same-origin code execution.
- **No post-accept revoke API.** Intentional — closing the controller tab destroys the only signing key; the accepted session lingering in live memory is inert without it.
- **Same-origin code execution (XSS in either tab).** Out of scope; an in-tab attacker already holds the session/key.

Durable record: `docs/developer/CROSS_TAB_CONTROLLER.md` → "Security / trust model" + "Known limitations" (updated in this PR).

## Acceptance criteria ↔ proof

The canonical proof is a centralized, integration-style suite: **`src/integrations/cross-tab/cross-tab-protocol.acceptance.test.tsx`** (33 tests, 8 criteria). End-to-end flows drive the **real** `ControllerChannelProvider`, the **real** `installLiveTabExecutor`, and the **real** `pairing-manager` authgate over a shared in-process bus — **WebCrypto is never mocked**; only DOM-touching deps (action handlers, requirements-manager, sidebar, Grafana runtime) are. Field-level assertions call the real `verifySignedMessage` with a real controller keypair.

| # | Criterion | Proven by |
|---|---|---|
| 1 | **Launch binding** — challenge carries `pairingId`+proof; only registered launches accepted; wrong HMAC / mutated `sessionId` / mutated `publicKeyB64` / unknown `pairingId` rejected; URL-fragment round-trip | ✅ |
| 2 | **No prompt capture** — unproved hostile challenge can't take the prompt slot; valid one still promptable; competing valid challenges fail closed and leave neither acceptable | ✅ |
| 3 | **User consent** — trusted gesture required; untrusted/programmatic accept blocked; rejection suppresses; no side-effects before accept; signed command runs end-to-end after accept | ✅ |
| 4 | **Session binding** — bound to session / live tab / key / body / nonce / ts; wrong session, wrong liveTab, wrong key, body mutation, and sig-reuse-over-mutated-body all rejected, across all four command families | ✅ |
| 5 | **Replay & staleness** — exact replay rejected; session-wide nonce reuse rejected; stale and future-dated `sigTs` rejected | ✅ |
| 6 | **Reconnect** — same accepted session re-pairs without reload; live re-emits `pairing-accept`; signed commands resume | ✅ |
| 7 | **Shutdown hand-back** — prepared signed `sidebar-handoff:reopen`; `pagehide` posts it synchronously (no awaited WebCrypto); unsigned/stale handoffs rejected; doesn't steal the sidebar when another plugin owns it | ✅ |
| 8 | **Revocation** — rejected stays suppressed; expired pending retryable; authority is key-bound and dies with the tab (other keys + exact replays rejected; key is non-extractable) | ✅ |

Backed by the focused unit suites these consolidate from: `pairing-manager.test.ts`, `live-tab-executor.test.ts`, `controller-channel.test.tsx`, `cross-tab.types.test.ts`, `pathfinder-search-params.test.ts`.

## Manual verification

1. **Happy path** — open a Grafana tab (live), open the controller in a second tab → pairing banner appears in the live tab → Accept → run an interactive step from the controller → it executes on the live tab.
2. **Rejection** — repeat, click Reject / press Escape → no steps execute afterward.
3. **Forged command dropped** — in the live tab's DevTools, broadcast a `step-command` without a valid `sig` → not executed.
4. **Stale sig dropped** — pair, then fabricate a signed command with `sigTs = Date.now() - 35000` → dropped.
5. **Reconnect** — close and reopen the controller tab → live re-pairs cleanly without a live-tab reload.

## Breaking changes

None. Additive and reversible: the executor gains an optional `authGate` param (defaults to the module singleton); the message shape gains optional auth fields absent from unsigned messages.


tomglenn (changes requested):
## Requesting changes

First off, thank you for picking this up and going the ECDSA route. The approach is right, the WebCrypto primitives are implemented correctly, and the gate now covers all four side-effecting kinds. I do need to request changes though, because as written I don't think #1139 is actually closed yet. The signature authenticates the wrong scope, and there are a couple of supporting gaps that need to land before the flag can ever be flipped. The feature being dark (`TWOTAB_CONTROLLER_ENABLED = false`) is the only reason this isn't urgent.

Quick framing note that the rest hangs on: `BroadcastChannel('pathfinder-cross-tab')` is same-origin, so the attacker we care about is a script running inside the Grafana origin (XSS payload or a compromised co-plugin), and it can read every broadcast message, including the signed ones.

### Blocking: the signature does not cover the message body

The signed canonical payload, on both the signing side (`controller-channel.tsx:104`) and the verify side (`pairing-manager.ts:25-27`), is:

```
`${sessionId}|${liveTabId}|${kind}|${sigTs}`
```

That binds session, target tab, message kind, and timestamp, but not the body that actually drives the DOM (`action` / `refTarget` / `targetValue` for `step-command`, `requirements` / `fixType` / `targetHref` for `fix-requirement`, and so on). The executor then dispatches off the full attacker-supplied `validated` message (`live-tab-executor.ts:363-368`).

So a same-origin attacker can:

1. Capture any one legitimately signed message of a given kind off the channel while the user is running a guide.
2. Re-broadcast it reusing the same `sig` / `sessionId` / `liveTabId` / `kind` / `sigTs`, but swap the body (for example change a `fix-requirement`'s `targetHref`, or a `step-command`'s `action`).
3. `verifySignedMessage` recomputes `sessionId|liveTabId|kind|sigTs`, which is unchanged, so the ECDSA verify passes.
4. The attacker's body runs against the authenticated tab's DOM.

The signature proves "the paired controller sent some message of this kind recently," not "this instruction came from it." That is the exact attack this PR set out to stop, so the crypto being correct doesn't help here. This is the main blocker.

### Also needed before the flag flips

- **No replay protection.** The only anti-replay control is the +/- 30s `sigTs` window (`pairing-manager.ts:104`). There is no nonce, counter, or seen-set, so any captured signed message can be replayed verbatim within 30s, and since the controller emits fresh signed commands during normal use, an attacker can keep harvesting fresh signatures for the whole session. The `Math.abs` also accepts timestamps up to 30s in the future, which we don't need on a single machine.
- **Challenge trust is not bound to the controller the user opened.** A co-plugin can broadcast its own `pairing-challenge` with its own keypair. `setPendingChallenge` is replace-not-stack (`pairing-manager.ts:44-50`) and the controller re-announces on every tick, so an attacker can clobber the real challenge right before the user clicks. The banner only shows a truncated random `senderTabId`, so the user can't tell which controller they are accepting. If they accept the attacker's, `acceptedSession` binds to the attacker's key.
- **The security logic is untested.** `cross-tab-crypto.ts` and `pairing-manager.ts` have no test files, and the 8 new auth-gate tests inject a stubbed gate, so the real `verifySignedMessage` / `verifyPayload` (tamper, stale, replay, wrong tab, wrong session) is asserted nowhere. This is the load-bearing code and it currently has no coverage.

### Smaller things worth fixing in the same pass

- The private key is generated `extractable: true` (`cross-tab-crypto.ts:25`) but only ever needs `sign`. Set it to `false` as defense in depth.
- The live tab stays `acceptedSession` for the page lifetime, so a same controller reconnecting after going stale silently re-authorizes with no new consent. Worth a TTL or an explicit expiry.
- `SIDE_EFFECTING_KINDS` (`live-tab-executor.ts:59`) and `SIGNED_KINDS` (`controller-channel.tsx:95`) are duplicated security-critical constants that have to stay in lockstep. If they drift, the controller signs a kind the executor doesn't gate, or vice versa. Please hoist to one shared constant in `types/cross-tab.types.ts`.
- `aria-modal="true"` is on a non-modal corner toast in `PairingRequestBanner.tsx`. Either make it a real focus-trapped modal or drop the attribute.

### We already maintain this crypto in workshop mode

`src/security/cross-tab-crypto.ts` is roughly 85% byte-identical to the existing `src/integrations/workshop/session-crypto.ts`. `toBase64url`, `fromBase64url`, and `generateSessionKeyPair` are character-for-character copies, and the ECDSA params and verify/import logic are identical. The only real difference is one line of how the bytes to sign are derived. The tiers allow sharing cleanly, since `security` is tier 1 and workshop is tier 3, so workshop can import down. I'd like us to extract the shared primitives into a neutral `src/security/ecdsa.ts` (`signBytes` / `verifyBytes` plus the base64url helpers and keygen) and have both features keep thin adapters. Shipping as-is leaves two diverging ECDSA copies where a fix to one silently misses the other.

### What gets this to an approve

1. Sign the full body. Include a canonical, sorted-key serialization of every action-bearing field in the signed payload, changing `controller-channel.tsx` and `pairing-manager.canonicalPayload` in lockstep, with a body-tamper negative test.
2. Add replay protection: a monotonic `seq` or nonce in the signed payload, rejected if not strictly newer, and tighten the window to forward-only.
3. Bind the challenge to the controller the user opened (a short code in the controller URL/UI that the banner displays and matches), or at minimum surface the competing-challenge condition instead of silently last-writer-wins.
4. Test the real crypto and pairing manager (round-trip, tamper, stale, replay, wrong tab, wrong session) before the flag flips.
5. Consolidate the crypto with workshop and set `extractable: false`.

The direction is good and the UI work is clean, so none of this is a knock on the effort. I just don't want us to call #1139 closed or enable the flag until the command body is actually signed. Happy to pair on the body-signing and nonce changes if that helps.

moxious:
## Proposed Acceptance Criteria

Separate it into protocol invariants:

- Launch binding: only the controller opened by the live tab can produce an acceptable pairing challenge.
- No prompt capture: invalid or competing same-origin challenges cannot block or replace the expected controller prompt.
- User consent: no side-effecting command is accepted before a trusted user acceptance action.
- Session binding: accepted commands are bound to session ID, controller public key, live tab ID, timestamp, nonce, and full command body
- Replay/staleness: stale commands and exact nonce replays are rejected.
- Reconnect: the same accepted controller can recover after heartbeat loss without a full reload.
- Shutdown hand-back: controller close/navigation reliably hands the sidebar back without weakening authentication.
- Revocation semantics: define exactly what “revoked” means for this PR. Right now it covers rejection, pending challenge expiry, competing-challenge denial, and controller shutdown; it does not add a post-accept “disconnect this controller” user control unless we choose to include that.

moxious:
## Relationship to Workshop / Live Sessions auth — shared surface vs. why the protocols diverge

A colleague raised a good question while reviewing this: Workshop signs the *handshake* and then sends an unsigned body, so why does cross-tab sign *every* message? I dug through both subsystems to substantiate it. Short version: he's right, and the reason is structural (transport), not stylistic.

### Workshop's model (confirmed in code)

Workshop runs over **WebRTC / PeerJS `DataConnection`** (`session-manager.ts:7,119,507`) — a DTLS-encrypted, ordered, point-to-point channel. It signs **only the handshake**: attendee sends a nonce (`session-manager.ts:530`), presenter signs with the per-session ECDSA P-256 key (`signChallenge`, `session-manager.ts:249`), attendee verifies against the public key in the join code (`verifyChallenge`, `session-manager.ts:545`). After that, actions go out raw — `broadcastToAttendees` → `conn.send(event)` (`session-manager.ts:651`), and `InteractiveStepEvent` carries no signature. Data-channel "replay protection" is only a 100/500ms dedupe window (`action-capture.ts:303`, `action-replay.ts:351`).

That's safe **because of the transport**: prove identity once *over the encrypted channel*, and DTLS guarantees every subsequent message on that channel comes from the same peer (with integrity + anti-replay on the wire). The channel itself is the binding.

### Why cross-tab can't reuse that model

`BroadcastChannel` is a same-origin **bus**, not a connection (`cross-tab-transport.ts:19`): plaintext, no peer identity, any same-origin script can `postMessage`. "Authenticate once" is meaningless when there's no connection to anchor the authentication to — every message is an independent, anonymous post. This PR's doc says it directly (`CROSS_TAB_CONTROLLER.md:88`): a message on `pathfinder-cross-tab` is *not* proof it came from a Pathfinder controller.

So this PR has to rebuild, at the application layer, the guarantees DTLS gives Workshop for free: **sign every side-effecting message** + bind to `sessionId`/`liveTabId`/body/`sigNonce`/`sigTs` + a **session-wide replay ledger** + a ±30s window. Tellingly, Workshop's unsigned body — perfectly safe over DTLS — would be a straight drive-the-DOM vuln if dropped onto BroadcastChannel. This PR correctly **inverts** the model rather than porting it.

### Shared surface (what *can* be reused)

The crypto primitive overlap is real — in fact it already happened by copy. `security/cross-tab-crypto.ts` is a near-literal fork of `workshop/session-crypto.ts`:

| Primitive | Status |
|---|---|
| `toBase64url` / `fromBase64url` | **byte-identical** (~19 lines) |
| `ECDSA_PARAMS` / `SIGN_PARAMS` | identical |
| `generateSessionKeyPair` | identical **except the `extractable` flag** — cross-tab `false` (non-extractable, deliberate), workshop `true` |
| ECDSA sign/verify core | same `crypto.subtle` calls; differ only in input marshalling (payload-string-encode vs nonce-base64url-decode) |
| nonce generator | a third dupe pair: `createSignatureNonce` (`pairing-manager.ts:72`) vs `generateNonce` |

Consolidatable core ≈ **45–50 lines**, collapsible into one ~45-line `security/` module. Tier rule already permits it: `security/` is Tier 1, `workshop/` is Tier 3, so workshop importing the shared primitive is a legal downward edge.

What stays separate: the **HMAC launch-binding helpers** (cross-tab only), the sign/verify *wrappers*, and — crucially — the entire protocol layer (handshake-only vs sign-every-message + replay ledger).

### Take

Net LOC win from unifying the primitive is modest (~40–55 lines) and probably not worth a standalone PR. The value is **a single audited crypto primitive and one source of truth for the wire format** so the two features can't silently drift in how they encode keys/signatures. One caveat if anyone does it: the `extractable: false` divergence is load-bearing security here (it's what makes "authority dies with the tab" true), so a unified `generateSessionKeyPair` must keep it as an explicit parameter.

The colleague's instinct — "two features, two protocols, no requirement to share the handshake" — holds up. The building blocks overlap; the protocols don't, and shouldn't. This PR's per-message signing isn't reinventing Workshop's wheel — it's compensating for a transport that, unlike WebRTC, provides none of the channel guarantees that make "sign the handshake only" safe.

tomglenn (approved):
## Approving

This is a strong turnaround. All three blockers from my earlier review are genuinely closed, not just renamed, and I verified each by reading the new code and running the suites (67 passing, including the new acceptance suite).

- **Body is signed.** `canonicalSignedPayload` covers every field except `{source, senderId, timestamp, sig}` via a stable sorted-key serialization, and the verifier recomputes over the received body, so mutating `action` / `targetHref` / `refTarget` breaks the signature. The substitution forgery is gone.
- **Replay is closed.** `sigNonce` plus the `${sessionId}:${sigNonce}` ledger, with the signature check ordered before the ledger insert so it can't be grown by an unauthenticated flood.
- **The challenge is bound to the controller the user opened.** The `pairingSecret` rides the controller's URL hash (never the bus), and the HMAC `pairingProof` covers `publicKeyB64`, so a same-origin script that can read the channel still can't forge or key-swap an acceptable challenge. The visible `pairingCode` and trusted-gesture accept are good additions.

`extractable: false`, the deduped signed-kinds constant, the competing-challenge denial, and the new tests all address the rest of the review. The workshop comparison is settled too: the protocols diverge for transport reasons and the primitive overlap is small, so not sharing the handshake is the right call.

Approving on that basis. #1139 is closed for the stated same-origin threat model.

### Follow-ups (non-blocking, please track before the flag flips)

None of these block merge, and the feature is flag-dark, but they are worth a tracked issue so they aren't lost:

1. **Forged `pairing-accept` DoS (medium, availability-only).** `pairing-accept` is unsigned and matched only on the broadcast `sessionId` (`controller-channel.tsx:207-216`). A same-origin attacker can post a forged accept faster than the user clicks and latch `pairedLiveIdRef` onto a bogus sender. No integrity or confidentiality impact (the real live tab rejects the attacker-bound commands via the `liveTabId` check), but it leaves the controller dead. Worth binding or signing `pairing-accept`, or at least surfacing a stuck/unpaired state instead of silently latching the first sender.
2. **Future-dated timestamps (low).** `Math.abs(now - sigTs)` accepts `sigTs` up to 30s in the future (`pairing-manager.ts:358`, mirrored in the controller). Two same-origin tabs share a clock, so reject `sigTs > now + small_skew` rather than allowing the forward window.
3. **URL-hash strip timing (low).** The secret strip runs inside the async `plugin.init` path, so the hash lingers in the address bar / history briefly and persists if init fails. Strip it before the async imports.
4. **Single crypto primitive (maintainability).** `security/cross-tab-crypto.ts` still overlaps `workshop/session-crypto.ts` by ~40 to 55 lines. Not worth a standalone PR, but folding the shared base64url + keygen + sign/verify core into one audited `security/` module would stop the two copies drifting on key/signature encoding. Keep `extractable` an explicit parameter if you do.

Nice work getting this to a real authenticated boundary.

tomglenn:
Filed the security/robustness follow-ups from my approval as a tracking issue so they aren't lost before the flag flips: https://github.com/grafana/grafana-pathfinder-app/issues/1158 (the forged `pairing-accept` DoS is the one with a security dimension; the timestamp/hash-strip/crypto-consolidation items are folded in there too).

PR #1154, opened 2026-06-25
feat(e2e): authenticate cloud-tier guides with ephemeral instance auth

leslie-heinzen:
## Summary

Cloud-tier E2E guide runs can now authenticate against real Grafana Cloud targets by provisioning short-lived service-account tokens from explicit per-instance admin-token mappings. The E2E CLI resolves each remote guide's target from package metadata, provisions ephemeral credentials per target origin and dependency chain, and keeps admin tokens and minted runner tokens out of package metadata and JSON reports.

This replaces the earlier direct-token/form-login shape in the PR with a single origin-scoped provisioning model: `--cloud-instance-admin-token host=ENV_VAR`.

## What to look at

- **E2E command orchestration** — [`src/cli/commands/e2e.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/7f98a9c0a979da1f496db055275ec48b8b908581/src/cli/commands/e2e.ts) now resolves remote inputs, builds dependency chains, provisions cloud credentials per target origin, runs Playwright, and reports results without carrying credential values in package metadata.
- **Cloud auth and provisioning policy** — [`src/cli/e2e/cloud-auth.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/7f98a9c0a979da1f496db055275ec48b8b908581/src/cli/e2e/cloud-auth.ts), [`src/cli/e2e/cloud-provisioning.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/7f98a9c0a979da1f496db055275ec48b8b908581/src/cli/e2e/cloud-provisioning.ts), and [`src/cli/e2e/cloud-environment.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/7f98a9c0a979da1f496db055275ec48b8b908581/src/cli/e2e/cloud-environment.ts) split credential policy from side-effectful provisioning: auth policy exposes provisionable origins, provisioning mints/tears down service accounts, and `CloudEnvironment` owns Grafana's service-account API calls for one target stack.
- **Remote target/package resolution** — [`src/cli/e2e/e2e-targets.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/7f98a9c0a979da1f496db055275ec48b8b908581/src/cli/e2e/e2e-targets.ts) and [`src/cli/e2e/e2e-package.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/7f98a9c0a979da1f496db055275ec48b8b908581/src/cli/e2e/e2e-package.ts) treat cloud auth as target capability (`CloudAuthTargets`), not secret material. This is the key invariant: a cloud guide only becomes runnable when its resolved origin is provisionable.
- **Playwright runner auth contract** — [`src/cli/e2e/playwright-runner.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/7f98a9c0a979da1f496db055275ec48b8b908581/src/cli/e2e/playwright-runner.ts), [`src/cli/e2e/e2e-runner-contract.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/7f98a9c0a979da1f496db055275ec48b8b908581/src/cli/e2e/e2e-runner-contract.ts), [`tests/e2e-runner/playwright.config.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/7f98a9c0a979da1f496db055275ec48b8b908581/tests/e2e-runner/playwright.config.ts), and [`tests/e2e-runner/auth/auth.setup.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/7f98a9c0a979da1f496db055275ec48b8b908581/tests/e2e-runner/auth/auth.setup.ts) preserve per-guide runner isolation. The remaining `GRAFANA_TOKEN` env key is internal runner transport for the minted short-lived token, not a user-supplied long-lived token.
- **Docs and focused tests** — [`docs/developer/E2E_TESTING.md`](https://github.com/grafana/grafana-pathfinder-app/blob/7f98a9c0a979da1f496db055275ec48b8b908581/docs/developer/E2E_TESTING.md), [`docs/design/e2e-test-runner-design.md`](https://github.com/grafana/grafana-pathfinder-app/blob/7f98a9c0a979da1f496db055275ec48b8b908581/docs/design/e2e-test-runner-design.md), and the new/updated tests under [`src/cli/e2e/`](https://github.com/grafana/grafana-pathfinder-app/tree/7f98a9c0a979da1f496db055275ec48b8b908581/src/cli/e2e) document and verify the shipped cloud-auth behavior.

## Why

Remote package-aware E2E tests already knew which guides target `local` vs `cloud`, but cloud-tier guides could not be exercised safely against real Cloud stacks. Direct long-lived service-account tokens and username/password form login were rejected for the final shape because they either do not work for grafana.com SSO stacks, put durable credentials into browser traffic, or do not provide per-chain identity isolation.

The resulting design uses admin tokens only at the CLI provisioning boundary. Each dependency chain gets fresh service-account identity per target origin, Playwright receives only the minted short-lived token, and the CLI tears down provisioned service accounts at chain end with TTL and startup orphan sweep as additional safety nets.

## How to verify

1. First, in Grafana, generate an admin Service Account token with `serviceaccounts:write` from Grafana > Administration > Service Accounts. Pick a cloud guide with no declared instance in its manifest. Set an admin token env var for your specific cloud stack, then run the built CLI with `--tier cloud`, `--cloud-url <stack-url>`, and `--cloud-instance-admin-token <stack-host>=<ENV_VAR_NAME>`. Confirm the guide runs against the stack and the temporary service account is deleted after the chain completes.
2. Pick or create a guide whose manifest declares `testEnvironment.instance: "play.grafana.org"`. Run it with `--cloud-instance-admin-token play.grafana.org=<ENV_VAR_NAME>` and confirm the Playwright request header uses the minted token for `play.grafana.org`, not an admin token or a token for another stack.
3. Run the same `instance` guide without a matching `--cloud-instance-admin-token` mapping. Confirm the package is reported as `skipped_no_auth` and no browser run starts for that guide. **Note: this requires that we are able to general SA tokens on the targeted instance.**
4. Start a multi-guide run with two different cloud target origins and confirm the CLI provisions and tears down one temporary service account per origin for the dependency chain.
5. Interrupt a cloud run during execution, then start another run for the same target. Confirm the startup orphan sweep attempts to remove leftover `pathfinder-e2e-*` service accounts before provisioning new ones.

## Breaking changes

None relative to `main`. This is additive CLI behavior for cloud-tier E2E runs.

## Reversibility

Reverting this PR stops new cloud service-account provisioning. If a run crashes after creating a temporary service account, the account is bounded by token TTL and the next run's orphan sweep attempts cleanup, but a manual cleanup in the target Grafana stack may still be needed for any already-created `pathfinder-e2e-*` service accounts.

## Out of scope

- Path and journey milestone expansion remains deferred; non-`guide` package types still skip as `unsupported_type`.
- Interactive SSO/Okta browser login is not implemented; cloud auth goes through explicit admin-token mappings and minted service-account tokens.
- Ephemeral Grafana Cloud stacks and org-global data cleanup are not handled here; per-chain provisioning isolates identity state only.


src/cli/e2e/cloud-environment.ts
```diff
@@ -0,0 +1,154 @@
...
+ */
+
+import { randomUUID } from 'crypto';
+
+/** Prefix for provisioned SAs — used for identification and orphan cleanup. */
+const SA_NAME_PREFIX = 'pathfinder-e2e-';
+/** Token TTL: a safety net so a leaked token expires even if teardown never runs. */
+const TOKEN_TTL_SECONDS = 3600;
+/** Role granted to the ephemeral SA. Admin so guide steps can exercise any action. */
+const SA_ROLE = 'Admin';
```
leslie-heinzen: It seems there are some guides that require admin role perms in order to execute correctly, so I think this would be necessary to successfully test them.

On one hand, willy-nilly provisioning admin role service accounts is 👀 👀 👀 .

On the other hand, the provisioned account is ephemeral, and is torn down when the test completes.

moxious: I think your assumption is correct and you should go with Admin and note this as a need/requirement

Again if we're using ephemeral cloud stacks the impact of this decision goes way way down.  If you're trying to test live on long-lived production instances, the impact of this decision goes way up.

src/cli/e2e/cloud-environment.ts
```diff
@@ -0,0 +1,154 @@
...
+   * Delete service accounts left over from crashed runs (matched by name
+   * prefix). Best-effort: a failure here never aborts the run.
+   */
+  async sweepOrphans(): Promise<void> {
+    try {
+      const result = await this.api<ServiceAccountSearchResult>(
+        'GET',
+        `/api/serviceaccounts/search?query=${encodeURIComponent(SA_NAME_PREFIX)}&perpage=100`
+      );
+      const orphans = (result.serviceAccounts ?? []).filter((sa) => sa.name.startsWith(SA_NAME_PREFIX));
```
moxious: What are reasonable intended assumptions about how many threads might be running against the same instance?  If you decide for an architecture where (e.g.) you're testing multiple read-only guides against the same instance, they may run parallel and this orphan scanning by prefix will bite you.  If it's one guide per one thread, it probably doesn't matter.

Flagging this as a decision to make rather than a change that's needed.

src/cli/e2e/cloud-provisioning.ts
```diff
@@ -0,0 +1,87 @@
...
+}
+
+export async function provisionCloudTargetsForChain(options: {
+  targetUrls: string[];
+  cloudAuth: CloudAuthPolicy | undefined;
+  verbose: boolean;
+}): Promise<ProvisionedCloudTargets> {
+  const provisionedTargets = new ProvisionedCloudTargets();
+
+  for (const targetUrl of options.targetUrls) {
```
moxious: you need an outer loop that's something like "tearDownEverything" on fail.  If one of the items in this loop fails, you can error out and leave behind all the successful ones you've provisioned so far.

tests/e2e-runner/playwright.config.ts
```diff
@@ -34,20 +43,14 @@ export default defineConfig<PluginOptions>({
...
-      testMatch: [/.*\.js/],
-    },
-    // 2. Run tests in Google Chrome. Every test will start authenticated as admin user.
+    ...(useToken ? [] : [authProject]),
     {
       name: 'chromium',
       use: {
         ...devices['Desktop Chrome'],
-        storageState: join(projectRoot, 'playwright/.auth/admin.json'),
+        ...(useToken ? { extraHTTPHeaders: { Authorization: `Bearer ${serviceAccountToken}` } } : { storageState }),
```
moxious: every request to every origin gets the service account token.  Again provided that we're using ephemeral instances this isn't a heartstoppingly bad security finding, but good practice would be to limit the ability to send that token to only origins associated with the token.

src/cli/e2e/cloud-environment.ts
moxious: Suggested way you can get around this: add a per-run nonce segment to the SA name (pathfinder-e2e-<runId>-<rand>) and have the startup sweep only delete accounts older than the token TTL (age-gated), or only those matching this run's id at teardown. Keep the broad prefix only for an explicit --sweep/manual cleanup path

src/cli/e2e/cloud-provisioning.ts
```diff
@@ -0,0 +1,87 @@
...
+  for (const planned of chain) {
+    const targetUrl = packageMetaById.get(planned.id)?.targetUrl;
+    if (targetUrl && cloudAuth?.needsProvisioningFor(targetUrl)) {
+      targetUrls.add(new URL(targetUrl).toString());
+    }
+  }
+  return [...targetUrls];
+}
+
+export async function provisionCloudTargetsForChain(options: {
```
moxious, line 56: go hard on error handling logic in here

cloud-provisioning.ts:56-74: provisionCloudTargetsForChain adds each origin's SA to a local ProvisionedCloudTargets and returns it at the end. If origin 2's provisionChain() throws, the exception propagates before the return, so origin 1's already-minted SA is never returned to runChains and never reaches the finally teardown — it leaks (TTL/sweep only)

moxious (changes requested):

src/cli/e2e/cloud-environment.ts
leslie-heinzen: I like it. Pushed a change.

tests/e2e-runner/playwright.config.ts
leslie-heinzen: I debated delaying this for a follow-up PR because it changes the approach quite a bit but I think it's worth it https://github.com/grafana/grafana-pathfinder-app/pull/1154/changes/9a3b6cd1cb86772851d59a0e47fa78a0528a90f7

moxious (approved):

PR #1155, opened 2026-06-26
feat(reftarget): support ordered fallback selector chains for interactive blocks

jackw:
Part of grafana/grafana-frontend-platform#194 — selector-logic exploration, **Job 1 of 3**.

`refTarget` now accepts an ordered array of selectors and resolves them selector-major, so an interactive step survives UI churn by falling back to the next candidate. The element picker generates a strategy-diverse fallback chain automatically and the block editor carries it into the step form.

## What's in here
- **Schema + runtime** — guide JSON accepts `refTarget` arrays; the union is carried through the interactive runtime.
- **Resolution** — selector-major fallback resolution; `exists-reftarget` passes if *any* candidate resolves.
- **Picker → form** — `generateSelectorChain` produces a strategy-diverse chain; `generateSelectorFromEvent` returns it; the block editor authors the chain into the form and the manual fallback editor is removed in favor of it.
- **CLI** — repeated `--reftarget` builds a fallback chain.
- **Tests** — end-to-end parse + resolve coverage for chains.

## Validation
`npm run check` green.

### DIFF
--- a/src/cli/__tests__/authoring-integration.test.ts
+++ b/src/cli/__tests__/authoring-integration.test.ts
@@ -27,6 +27,7 @@ import { runCreate } from '../commands/create';
 import { runInspect } from '../commands/inspect';
 import { runSetManifest } from '../commands/set-manifest';
 import { readPackage } from '../utils/package-io';
+import { findBlockById } from '../utils/package-io/tree';
 import { validatePackage } from '../../validation/validate-package';
 
 function tempDir(): string {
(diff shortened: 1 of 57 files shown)


moxious:
Thanks for this proposal!  Here are some thoughts and discussion points about design.

Several things I really appreciate about this design is the backwards compat in the zod schema design, 

## E2E testing

We have plans to e2e test pathfinder guides that are underway. If you want the full gory details, they're here.  https://github.com/grafana/pathfinder-rfcs/blob/main/rfc/E2E-GUIDE-TESTING.md. Anyway - if a guide expresses a selector chain, what would "pass" mean?  Suppose a guide has 5 selectors, the first 4 are bad, the 5th works.  The guide "passes" because the 5th worked.  But pathfinder has a retry mechanism in it with a backoff; if the first fails, it gets retried, so the deeper the "succeed" is in the chain, the concern might be that even allowing the selector to be correct is a compounding performance penalty.  Do we transmute the problem of a wrong selector into a performance problem?

Ref evidence: (resolveSelectorPipeline loops candidates, calling resolveSingleCandidate per entry). The default budget is delays: [200, 600, 1800] (selector-retry.ts:17). So a candidate that genuinely doesn't exist burns ~2.6 seconds of backoff before the chain advances.

## Axes of consideration for selectors

Our selector algorithm tries to pick best available based on a static DOM snapshot, and the proposed code has a picker that's operating on the same basis.  Challenge here is it could be valuable to have alternative selectors in these scenarios I can tell:

1. Grafana version skew (we're trying to support OSS & cloud)
2. Feature-flag / rollout-gated UI
3. Distribution differences

This is challenging because none of those can be addressed from the record UI, or an in-app picker because they're all based on the static DOM snapshot we have this moment.  The feature-flag/rollout-gated UI one would be really compelling, but e.g. a tech writer won't have a datasource to know what the second or third selector could be.  The grafana version skew one is easier addressed by adopting the other PR (reverse lookup to e2e selectors) and then expanding e2e selector coverage.

This leads me to...

## When to use selector chains?

One of the e2e tests suggests we could use this to get around i18n.  We could target a button as "Save" or as "Speichern" (German).  This probably isn't the strongest framing of the idea though, because this would induce us to couple i18n translations into a selector chain, and how many languages deep should we go?  Anyway, i18n isn't the point of the code, that was just an example in a test.  

Still ... when would I use this?  If we have a selector algo to pick the best available, would we extend the selector chain by using increasingly imprecise selectors?  Would we have scenarios where we know there are two equally valid/good ones and use those?  Or something else?  This is an important question because we have a content authoring ecosystem including tech writers, we'd need to tell them how/when to use a feature like this.

jackw:
I'm going to go ahead and close this PR as discussions led to the conclusion fallback selectors aren't a direction we want to take.

PR #1156, opened 2026-06-26
feat(dom): emit version-stable grafana: selectors from the element picker

jackw:
Part of grafana/grafana-frontend-platform#194 — selector-logic exploration.

> Originally stacked on #1155 (fallback selector chains), which was closed after discussion. This PR has been rebased onto `main` and reworked to stand alone.

When the picked element (or an ancestor) carries a Grafana `data-testid`, the picker reverse-looks-up the element to a `grafana:` selector path, producing version-stable selectors that don't break on CSS-class churn.

- Reverse-lookup element → `grafana:` selector path (`findGrafanaSelectorPath`): a memoized reverse index over the `components.*` and `pages.*` trees of `@grafana/e2e-selectors`, matching `data-testid` (then `aria-label`) and extracting parameters for parameterized selectors.
- The picker emits the `grafana:` selector as the top-ranked candidate; the raw testid selector remains available as a fallback via `generateFallbackSelectors`. Uniqueness testing resolves `grafana:`/`panel:` prefixes before matching.
- **Version-aware resolution**: both directions resolve the selector tree against the running Grafana's `config.buildInfo.version` — the emitted `grafana:` path is the stable identity, and its CSS value is computed per-version at replay time.
- **Ambiguity rejection**: a value claimed by more than one selector path (8 of 637 values in the current bundle), or matching more than one parameterized template, is never reverse-matched — the picker degrades to its own CSS strategies rather than emitting a confidently wrong path.
- `@grafana/e2e-selectors` is now a direct pinned dependency (previously transitive-only), with the bundle-freshness coupling documented in `.cursor/rules/techContext.mdc`.

## Validation
`npm run check` green.

### DIFF
--- a/src/lib/dom/grafana-selector.ts
+++ b/src/lib/dom/grafana-selector.ts
@@ -146,8 +146,9 @@ interface ReverseIndex {
   templates: Array<{ regex: RegExp; path: string }>;
 }
 
-// A control character that never appears in a real selector value, used to
-// locate the parameter position inside a parameterized selector's output.
+// Delimited by U+E000 private-use characters that never appear in a real
+// selector value, used to locate the parameter position inside a
+// parameterized selector's output.
 const TEMPLATE_SENTINEL = 'PARAM';
 const TESTID_PREFIX = /^data-testid\s*/;
 
--- a/src/lib/dom/grafana-selector.ts
+++ b/src/lib/dom/grafana-selector.ts
@@ -4,8 +4,27 @@
  * Based on @grafana/plugin-e2e approach for cross-version compatibility
  */
 
-import { selectors as grafanaSelectors } from '@grafana/e2e-selectors';
+import { resolveSelectors, versionedComponents, versionedPages } from '@grafana/e2e-selectors';
+import { config } from '@grafana/runtime';
 import { querySelectorAllEnhanced } from './enhanced-selector';
+
+type SelectorNode = { [key: string]: SelectorNode | string | ((...args: never[]) => unknown) };
+
+let resolvedSelectors: { version: string; tree: SelectorNode } | null = null;
+
+/**
+ * The selector tree resolved for the running Grafana version, so both forward
+ * resolution and reverse lookup use the values this instance actually renders,
+ * not the newest values in the bundled package.
+ */
+function getResolvedSelectors(): SelectorNode {
+  const version = config.buildInfo.version || 'latest';
+  if (resolvedSelectors?.version !== version) {
+    const tree = resolveSelectors({ components: versionedComponents, pages: versionedPages }, version);
+    resolvedSelectors = { version, tree: tree as unknown as SelectorNode };
+  }
+  return resolvedSelectors.tree;
+}
 
 /**
  * Convert a Grafana selector path to a CSS selector string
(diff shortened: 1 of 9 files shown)


moxious:
I like the direction; Anchoring our picker to @grafana/e2e-selectors is the right instinct, and a couple of things about it are particularly good.

The version coverage is the part I'm most drawn to. I went and looked at what's actually in the package, and it's richer than I expected: MIN_GRAFANA_VERSION is 8.5.0 ... 13.0.0. So the archaeological record is good. Static maintained lookup table of known-good selectors across Grafana versions is useful because we expect to have to support a range of OSS and Cloud versions over time, inheriting that table is a strong addition we definitely can't build ourselves.

It also gives the core Grafana team a "way in" to our codebase, which I like. Every time upstream e2e-selectors improves, we bump the bundle and inherit.

Some things about the PR worth discussing ... 

1. It needs to actually use the version machinery. Right now the PR imports the flat selectors export, which is resolveSelectors(versioned) with no version argument — that defaults to "latest" and collapses every versioned map to its newest value. This would mean if someone is running something in the 12 series, the oracle lookup table could be wrong, despite having the right answer.

2. We need to own pathfinder's "bundle, not Grafana version" coupling. The stability is only as good as how current our bundled @grafana/e2e-selectors is relative to the fleet of Grafanas we run against. We should write that dependency down somewhere visible and have a plan to keep the bundle fresh, because nothing in the code enforces it today.

3. We should enumerate the ways the lookup can be wrong and mitigate each one in the PR. If there's a total miss in e2e-selectors that's arguably already safe-ish, because it falls through to the selector algorithm  we built.  The risk of this change is in the case where we match a selector but it's wrong.
- Version skew author on latest, replay on older Grafana
- Mis-attribution: the exact index is first-path-wins on duplicate string values (components before pages), so two selectors sharing a string collapse to one path and can silently track the wrong selector's evolution.
- Template over-match / wrong parameter, greedy `^prefix(.+)suffix$` regexes, first-match-wins, can capture the wrong parameter or match multiple templates.
- aria-label / i18n — we match against aria-label too, and older selector values are aria-label text; on a non-English Grafana the author's path may not resolve against a translated DOM.

For each of these we should plan a mitigation, and where we can't fully mitigate, make sure we degrade cleanly to the existing picker rather than emitting a confident wrong. We can never do worse than "use the algorithm we already have," 

I think some version of this though clearly does become a meaningful durability win.

src/lib/dom/grafana-selector.ts
```diff
@@ -113,3 +136,144 @@ export function existsByGrafanaSelector(selectorPath: string, selectorId?: strin
...
+    for (const template of index.templates) {
+      const match = template.regex.exec(value);
+      if (!match || !match[1]) {
+        continue;
+      }
+      if (matched) {
+        matched = null;
+        break;
+      }
+      matched = `grafana:${template.path}:${match[1]}`;
```
Jayclifford345, line 271: If the captured param contains `:` or `'` (dashboard titles make both realistic — a breadcrumb for "Home: Overview" or "Jay's board"), the emitted path can't round-trip: `resolveSelector` splits on `lastIndexOf(':')` and `toGrafanaSelector` interpolates into single quotes. The uniqueness gate does drop the broken candidate, but only after a logged resolve error. Since the design already prefers degrading over emitting a confidently-wrong path, worth skipping such params here explicitly, in the same spirit as the ambiguity rejection?

Jayclifford345:
Reviewed — no blocking changes from my side. Verified the load-bearing invariants: `admitCandidate` drops any `grafana:` candidate whose resolution fails or matches nothing, so the picker can't emit a broken path; the versioned test suite exercises the real `resolveSelectors` end-to-end; and the `@grafana/e2e-selectors` pin matches what was already resolved transitively in the lockfile. One non-blocking suggestion open inline: skip reverse-matched params containing `:` or `'` explicitly rather than relying on the downstream resolve failure.

Jayclifford345 (approved):

PR #1157, opened 2026-06-26
feat(dom): anchor identity-less elements via stable ancestors and :has() descendants

jackw:
Part of grafana/grafana-frontend-platform#194 — selector-logic exploration, **Job 3 of 3**.

> **Stacked on Job 1** (#1155). Review/merge that first; GitHub retargets this to `main` once it lands. The diff shown here is scoped to Job 3 only.

Identity-less wrapper elements (no id/testid/stable class) previously produced brittle positional selectors like `div:nth-match(4)`. This job anchors them on stable structure instead:

- **Up** — anchor on the nearest stable ancestor.
- **Down** — borrow a stable descendant's identity via `:has()` *without* retargeting the intent element (e.g. `li>div:has(a[data-testid='...'][href='/explore'])`).

It also unifies the block-editor health-badge scoring with the generator, so the badge reflects the same quality signal the engine uses.

## Validation
`npm run check` green; `:has()` path validated in-browser against the `/explore` nav.

### DIFF
--- a/src/components/block-editor/SelectorHealthBadge.tsx
+++ b/src/components/block-editor/SelectorHealthBadge.tsx
@@ -11,59 +11,25 @@ import { GrafanaTheme2 } from '@grafana/data';
 import { css } from '@emotion/css';
 import { querySelectorAllEnhanced } from '../../lib/dom/enhanced-selector';
 import { resolveSelector } from '../../lib/dom/selector-resolver';
+import { analyzeSelectorString } from '../../lib/dom/selector-generator';
 import { useDebouncedValue } from './useDebouncedValue';
 
 interface SelectorHealthBadgeProps {
   reftarget: string;
 }
 
-interface SelectorInfo {
+const QUALITY_COLORS: Record<string, string> = {
+  good: '#73BF69',
+  medium: '#FF9830',
+  poor: '#F2495C',
+};
+
+interface BadgeInfo {
   method: string;
-  score: number;
+  stabilityScore: number;
+  quality: string;
   matchCount: number;
   warnings: string[];
-}
-
-function analyzeSelectorPattern(reftarget: string): Omit<SelectorInfo, 'matchCount'> {
-  const warnings: string[] = [];
-  let method = 'compound';
-  let score = 40;
-
-  if (reftarget.includes('data-testid')) {
-    method = 'data-testid';
-    score = 100;
-  } else if (reftarget.includes('aria-label')) {
-    method = 'aria-label';
-    score = 85;
-  } else if (reftarget.includes('#')) {
-    method = 'id';
-    score = 90;
-  } else if (reftarget.includes(':text(') || reftarget.includes(':contains(')) {
-    method = 'text';
-    score = reftarget.includes(':text(') ? 65 : 55;
-    warnings.push('Text-based selectors are fragile and may break if the UI text changes');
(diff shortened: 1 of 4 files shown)


moxious:
I really like this concept and think we should do some variant of it.  Basically I interpret this as raising the quality floor of the worst selector cases, by anchoring them on something "nearest nearby stable".

Conceptually, this strengthens the points I made on the other PR comment (https://github.com/grafana/grafana-pathfinder-app/pull/1155) about "when would I use selector chains".  The tension between these two PRs is that whenever we have the option to strengthen the quality of a selector, it'll always be preferable to being able to express more (potentially bad) selectors.  I like this PR because it pushes in the direction of improving quality of selectors, even in cases where we can't necessarily have a global optimum.  I think it may also help with version skew, because if we chain a selector off of, e.g. `a[href='/explore']` that's more likely to be version stable since paths don't change as much.

Anchoring downward with the `:has()` operator is also a pretty nice approach.

src/components/block-editor/SelectorHealthBadge.tsx
```diff
@@ -11,73 +11,41 @@ import { GrafanaTheme2 } from '@grafana/data';
...
+interface BadgeInfo {
   method: string;
-  score: number;
+  stabilityScore: number;
+  quality: string;
   matchCount: number;
   warnings: string[];
 }
 
-function analyzeSelectorPattern(reftarget: string): Omit<SelectorInfo, 'matchCount'> {
```
jackw, line 27: Replacing this with `analyzeSelectorString` to unify health badge scoring with the generator.

src/lib/dom/selector-generator.ts
```diff
@@ -1199,3 +1384,96 @@ function computeQuality(stabilityScore: number, flags: StabilityFlag[]): Selecto
...
+  warnings: string[];
+}
+
+/**
+ * Infer the method + internal score of a *typed* selector string, in the same
+ * priority order the candidate generators use. The strongest signal present wins;
+ * `computeStabilityFlags` then corrects for fragility (e.g. a testid-scoped
+ * selector that also uses `:nth-child` is downgraded via the structural flag).
+ */
+function inferMethodAndScore(selector: string): { method: string; score: number } {
```
Jayclifford345, line 1424: `grafana:` / `panel:` reftargets fall through every branch here and come out as `compound` / 50 / medium — and once #1156 lands, `grafana:` becomes the picker's top emission, so the badge would rate our most stable selector as mediocre. Probably wants an early branch for the two prefixes. Related nit: `button#save` misses the id regex (it requires a space/combinator before the `#`), which the old badge heuristic caught.

Jayclifford345:
Reviewed — one small change I'd make before merge: teach `inferMethodAndScore` about `grafana:` / `panel:` reftargets (see inline comment), otherwise the badge rates the picker's best output as mediocre once #1156 lands. Everything else checked out: `:has()` resolution is native-first with the custom parser only as fallback (and that path predates this PR, so already-authored guides stay safe on revert), and the descendant scan runs only on recorded actions, never hover.

One housekeeping notes this PR and #1156 edit the same hunks of `generateCandidates` / `admitCandidate`, so whichever merges second needs a semantic rebase (the unique-match branch ends up needing both `'grafana'` and `'has-descendant'`).

Jayclifford345 (dismissed):

Jayclifford345 (approved):

PR #1170, opened 2026-06-29
fix(cross-tab): authenticate pairing-accept to prevent forged-accept DoS (#1158)

### LINKED ISSUE #1158: cross-tab: bind/sign pairing-accept to prevent forged-accept DoS (follow-up to #1153)
## Summary

Follow-up from the review of #1153 (authenticated cross-tab controller pairing). Non-blocking for that PR, but should be resolved **before `TWOTAB_CONTROLLER_ENABLED` is flipped on**.

`pairing-accept` is unsigned and matched only on the broadcast `sessionId` (`src/global-state/controller-channel.tsx:207-216`). The `sessionId` travels in the clear inside the `pairing-challenge`, so a same-origin script (XSS payload or compromised co-plugin) that reads `BroadcastChannel('pathfinder-cross-tab')` can post a forged `pairing-accept` faster than the user clicks Accept. The controller binds `pairedLiveIdRef.current` to the first accept and never rebinds.

## Impact

Availability only. The controller latches onto a bogus sender, then signs commands with `liveTabId = <attacker senderId>`. The real live tab rejects those (the `liveTabId !== ownTabId` check in `verifySignedMessage`), and the genuine user accept is ignored because `pairedLiveIdRef` is already set. Result: the legitimately-paired session can never drive the live tab — a dead controller.

No integrity or confidentiality impact: the attacker gains no ability to execute commands (they lack the private key, and the live tab gates on `liveTabId`).

## Suggested fix

- Bind or sign `pairing-accept` so the controller only latches onto an accept it can attribute to the tab it actually paired with, or
- Surface a stuck/unpaired state instead of silently latching the first sender (allow rebinding / re-pairing rather than first-writer-wins).

## Lower-priority items from the same review (fold in if cheap)

- **Future-dated timestamps (low):** `Math.abs(now - sigTs)` accepts `sigTs` up to 30s in the future (`src/lib/pairing-manager.ts:358`, mirrored in the controller). Two same-origin tabs share a clock — reject `sigTs > now + small_skew`.
- **URL-hash strip timing (low):** the `pairingSecret` strip runs inside the async `plugin.init` path, so the hash lingers in the address bar / history briefly and persists if init fails. Strip before the async imports.
- **Single crypto primitive (maintainability):** `src/security/cross-tab-crypto.ts` overlaps `src/integrations/workshop/session-crypto.ts` by ~40-55 lines. Fold the shared base64url + keygen + sign/verify core into one audited `security/` module (keep `extractable` an explicit parameter) so the two copies can't drift on key/signature encoding.

Refs #1153, #1139, #1133

Jayclifford345:
## What

Authenticates the `pairing-accept` message so a same-origin script can no longer forge one and dead-lock the two-tab controller (a forged-accept availability DoS), and tightens the signed-message timestamp window against future-dated `sigTs`. Both are "before re-enable" items for `TWOTAB_CONTROLLER_ENABLED`.

Closes #1158. Part of #1133.

## The problem

Before this change the controller bound its pairing slot on a bare `sessionId` match:

```ts
if (message.kind === 'pairing-accept') {
  if (message.sessionId !== sessionId) return;
  if (pairedLiveIdRef.current === null) {
    pairedLiveIdRef.current = message.senderId;   // first-writer-wins, never rebinds
    ...
  }
}
```

`sessionId` travels in cleartext inside the broadcast `pairing-challenge`, so a co-plugin / panel-XSS / extension script that reads `BroadcastChannel('pathfinder-cross-tab')` can post a forged `pairing-accept` faster than the user clicks Accept. The controller latches `pairedLiveIdRef` to the attacker's `senderId` and never rebinds, so the genuine session can never drive the live tab — a dead controller. (Integrity was already safe: the attacker lacks the controller's private key and the live tab gates commands on `liveTabId`. This is availability-only — but it blocks the feature.)

## The fix

`pairing-accept` now carries an HMAC proof keyed by the **launch secret** — the same `pairingSecret` already used for the challenge proof, delivered to the controller via its URL fragment and **never broadcast**:

- **Live tab** (`pairing-manager.createPairingAcceptForSession` → `live-tab-executor`): emits `{ pairingId, acceptProof = HMAC(pairingSecret, {pairingId, sessionId, liveTabId}) }`. The launch secret is captured into the accepted session at `acceptSession` time (and persists for reconnect re-emits); it never leaves the live tab's memory and is not exposed by `getAcceptedSession`.
- **Controller** (`controller-channel`): verifies `pairingId` matches its own launch and `verifyPairingAcceptProof(...)` passes **before** binding `pairedLiveIdRef`. Without a launch it has no secret to verify against, so it **fails closed** and never pairs.

A forger lacks the secret, so it cannot produce a valid proof for any `liveTabId` — it can neither take the slot nor bind the controller to a tab of its choosing. Replaying a genuine accept only re-confirms the genuine pairing. The accept proof binds `{pairingId, sessionId, liveTabId}` while the challenge proof binds `{pairingId, publicKeyB64, sessionId}`, so the two HMACs are domain-separated and non-interchangeable.

### Future-dated `sigTs`

`verifySignedMessage` previously accepted `sigTs` up to 30 s in the **future** (symmetric `Math.abs(now - sigTs) > 30s`). Both tabs share one clock, so a future timestamp is jitter or forgery; the window is now asymmetric — up to 30 s old, at most 1 s future skew — which shrinks the replay-validity window an attacker could claim.

## Tests

- `controller-channel.test.tsx`: new `ignores a forged pairing-accept and still pairs with the genuine live tab`; existing `pairWithLive` helper now produces a valid proof and awaits the (now-async) binding.
- `cross-tab-protocol.acceptance.test.tsx`: new boundary case for the tightened skew (a `now + 5s` sigTs that the old 30 s window accepted is now rejected); the real-flow suite exercises the new accept proof end-to-end with real WebCrypto.
- `interactive-step.test.tsx`: controller-mode pairing helper updated to the authenticated accept.
- `npm run test:ci` green; typecheck, eslint, prettier clean.

## Review

Ran an adversarial review (crypto-attacker, correctness/lifecycle, regression/contract lenses). The crypto and regression lenses found nothing; the one correctness note (a cosmetic symmetric-vs-asymmetric timestamp check in the controller's own prepared-handback self-guard) was verified to be structurally unreachable — `sigTs` is always in the past at verify time — and left unchanged to keep scope tight.

### DIFF
--- a/docs/developer/CROSS_TAB_CONTROLLER.md
+++ b/docs/developer/CROSS_TAB_CONTROLLER.md
@@ -81,9 +81,16 @@ DOM sink and is **not** unconditionally safe to install — see the trust model.
 
 ## Entry gate and mount policy
 
-Both the controller overlay (`?controller=1` path) and the live-tab executor install are gated on `pathfinderEnabled` (the `pathfinder.enabled` kill-switch) — the same policy that controls whether the main Pathfinder sidebar mounts. Because the executor drives the user's authenticated Grafana DOM, it must follow the same mount gate as the rest of the plugin.
+Both the controller overlay (`?controller=1` path) and the live-tab executor install are gated on two conditions: the `enableTwoTabController` admin setting (plugin config → Interactive features, default **off**), and `pathfinderEnabled` (the `pathfinder.enabled` kill-switch) — the same policy that controls whether the main Pathfinder sidebar mounts. The admin setting ships the feature dark until an instance opts in; the kill-switch keeps it aligned with the rest of the plugin. Because the executor drives the user's authenticated Grafana DOM, it must follow the same mount gate as the rest of the plugin.
 
 ## Security / trust model
+
+> **Reviewing a change in this subsystem?** The `cross-tab-controller` concern in
+> [`docs/design/CONCERNS.md`](../design/CONCERNS.md) is the canonical review checklist
+> for these files. It fires on any single touch of the cross-tab files and enumerates
+> the trust invariants below — signed commands, gesture-to-accept pairing, the per-kind
+> validation gate, and the `enableTwoTabController` re-enable one-way door — that a
+> reviewer must confirm before merge.
 
 `BroadcastChannel` is shared by every script running on the same origin, so a
 message on `pathfinder-cross-tab` is **not** proof it came from a Pathfinder
@@ -133,8 +140,9 @@ Defense in depth on top of authentication:
   set — at the transport receive gate _and_ again at the executor sink. Malformed
   or unknown-kind messages are dropped before the auth gate is even consulted.
 - **Install/entry gating.** The executor is installed, and the controller
-  overlay mounted, only when `pathfinderEnabled` is true — a disabled plugin
-  exposes neither the sink nor the driver.
+  overlay mounted, only when the `enableTwoTabController` admin setting is on
+  _and_ `pathfinderEnabled` is true — a disabled plugin, or an instance that
+  hasn't opted in, exposes neither the sink nor the driver.
 - **Same-build / same-origin / one-session assumption.** Controller and live
   tabs are the same plugin build in the same browser profile and session; there
   is no protocol-version negotiation and cross-version compatibility is not a
--- a/docs/developer/CROSS_TAB_CONTROLLER.md
+++ b/docs/developer/CROSS_TAB_CONTROLLER.md
@@ -101,7 +101,11 @@ drive that Grafana. The controller→live command path is therefore
   an HMAC over the canonical challenge `{pairingId, publicKeyB64, sessionId}`.
   The live tab accepts a challenge only when it matches a registered, unexpired
   launch — a wrong secret, a mutated `sessionId`/`publicKeyB64`, or an unknown
-  `pairingId` is dropped (`pairing-manager.ts`).
+  `pairingId` is dropped (`pairing-manager.ts`). The reverse `pairing-accept`
+  carries an HMAC over `{pairingId, sessionId, liveTabId}` keyed by the same
+  launch secret, so the controller binds its pairing slot only to an accept it
+  can attribute to the launched tab — a forged accept (`sessionId` is readable
+  off the wire) can't take the slot.
 - **Per-session keypair.** The controller generates a **non-extractable** ECDSA
   P-256 keypair per session; only the public key crosses the wire
   (`cross-tab-crypto.ts`). Authority to drive the live tab _is_ possession of
@@ -116,8 +120,9 @@ drive that Grafana. The controller→live command path is therefore
   `check-requirements`, `fix-requirement`, `sidebar-handoff` — is ECDSA-signed
   and bound to `sessionId`, `liveTabId`, the command body, a fresh `sigNonce`,
   and a `sigTs`. The executor's auth gate (`verifySignedMessage`) checks the
-  accepted session, the `sessionId`/`liveTabId` match, a ±30s timestamp window,
-  the signature against the accepted public key, and a session-wide
+  accepted session, the `sessionId`/`liveTabId` match, the timestamp window (up
+  to 30s old, at most 1s future-dated since both tabs share a clock), the
+  signature against the accepted public key, and a session-wide
   `(sessionId:sigNonce)` replay ledger before any action runs.
 
 Defense in depth on top of authentication:
--- a/docs/developer/CROSS_TAB_CONTROLLER.md
+++ b/docs/developer/CROSS_TAB_CONTROLLER.md
@@ -167,8 +167,15 @@ window plus the replay ledger kill any captured in-flight message.
 **Out of scope: same-origin code execution.** XSS or a malicious script running
 _inside_ the controller or live tab is out of scope — it already holds the
 session (and, in the controller, the signing key), so the cross-tab boundary is
-not the relevant control. Non-extractable keys and per-kind validation are
-hardening, not a defense against an already-compromised origin.
+not the relevant control. This includes capturing the launch secret: it reaches
+the controller through a `window.open` URL fragment, so a same-page script
+(panel/datasource XSS) or an extension content script with page access can read
+it and forge a valid challenge or accept proof. The authenticated pairing path
+therefore defends specifically against a **separate same-origin tab that only
+observes the channel** — it can post forged messages but cannot read the secret,
+so it can neither pair as the controller nor claim the live tab's pairing slot.
+Non-extractable keys and per-kind validation are hardening, not a defense
+against an already-compromised origin.
 
 ## Tab pairing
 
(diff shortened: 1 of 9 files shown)


moxious (changes requested):
**Requesting changes.** The core design is solid and correct for its scope, but there's a blocker: the forged-message availability DoS this PR exists to close is still fully open on the sibling `pairing-challenge` path, so merging this as-is doesn't actually clear the hole before re-enable.

First, the good part — I verified the accept-proof design against the code and it holds: genuine pairing survives (`session.liveTabId` == the transport-stamped `senderId` the controller verifies against), the accept proof and challenge proof are domain-separated (different key sets → no canonical collision, so the on-wire challenge proof can't be replayed as an accept proof), and the post-verify `pairedLiveIdRef` re-check closes the TOCTOU window. The `sigTs` tightening is correct too. 👍

**1. BLOCKER — the same forged-message availability DoS still exists on the `pairing-challenge` path.**

The challenge proof is HMAC'd over `{pairingId, publicKeyB64, sessionId}` only — `senderTabId` isn't signed — but `sameChallenge` compares `senderTabId` (`pairing-manager.ts:142`), and a "competing" fresh challenge runs `denyCompetingChallenge`, which deletes the launch and clears the pending challenge (`pairing-manager.ts:236`, `:262`).

Repro: an attacker reads the genuine `pairing-challenge` off the channel and re-posts it **verbatim, including the valid `pairingProof`, with a different `senderId`**. It verifies (senderId isn't in the signed payload), is treated as competing, and deletes the launch from `expectedLaunches`. The controller keeps re-broadcasting every heartbeat, but `verifiedLaunchForChallenge` then finds no launch and returns early forever → the pairing banner never appears → permanent dead controller, repeatable on every re-open.

This is the exact DoS class #1158 is about, just on the challenge side. Authenticating `pairing-accept` while leaving this open means a same-origin observer can still dead-lock the controller, so the "before re-enable" goal isn't met. I'd like it closed in this PR (or in a sibling PR landed alongside) rather than deferred.

Fix directions (your call): exclude the unauthenticated `senderTabId` from the competing-challenge comparison — two challenges with identical `{pairingId, publicKeyB64, sessionId}` are the same logical challenge regardless of envelope sender — or don't let an unauthenticated replay revoke an already-pending verified challenge. Please add a test: valid challenge → replay same proof under a different `senderId` → user accept still succeeds.

**2. Threat-model precision on the launch secret.**

The PR text lists panel-XSS and extension scripts among the defended-against attackers, but the secret is minted in the live tab and travels through `window.open` + the controller fragment, so a same-page script (panel-XSS) or an extension content script can capture it and forge a valid accept proof. The guarantee is really against a *separate same-origin tab that only observes the BroadcastChannel*. Suggest narrowing the wording in `CROSS_TAB_CONTROLLER.md` to that scope — the design is still a clear improvement, I just want the claim to match the boundary.

**3. (nit) No length cap before HMAC verify.**

`isValidPairingAccept` only checks `typeof acceptProof === 'string'`, and `verifyHmacPayload` base64-decodes arbitrary input. Pre-pairing, a flood of oversized proofs is alloc/crypto pressure. Capping `acceptProof`/`pairingId` to expected protocol lengths in the validator would harden it (the same gap exists on `pairingProof`, so a shared follow-up is fine rather than scope creep here).

Item 1 is the one holding the review. 2 and 3 can ride along or split out as you prefer.

Jayclifford345:
All three done in the latest push:

1. `sameChallenge` no longer compares the unauthenticated `senderTabId`, so a verbatim challenge replay under a different senderId gets deduped instead of tripping `denyCompetingChallenge` — no more launch-deletion dead-lock. Added the replay-under-a-new-senderId test.
2. Narrowed the wording in `CROSS_TAB_CONTROLLER.md`: the guarantee is against a separate observer tab; a same-page script or extension that can read the launch secret is the out-of-scope same-origin-execution case.
3. Length-capped the proof/id fields in both pairing validators before the HMAC verify.

Good catch on the challenge path being the real hole.

moxious:
Last one:  src/lib/pairing-manager.ts:327 can accept a still-visible pending prompt after its launch secret has been
  pruned. setPendingChallenge verifies the launch and stores only pairingCode, then acceptSession looks the launch up again later. Because the controller rebroadcasts challenges while unpaired (/private/tmp/pr1170/src/global-state/controller-channel.tsx:272), a rebroadcast just after the 5-minute launch TTL can prune expectedLaunches while the 30-second pending prompt is still fresh. A user click then sets acceptedSession with acceptedLaunchSecret = null; createPairingAcceptForSession returns null, so no authenticated pairing-accept is posted and that controller never pairs. Capture the verified launch secret with the pending challenge, or fail/clear the pending prompt if the launch is no longer available at accept time.

Jayclifford345:
done :)

moxious:
## Review — security-sensitive (pairing handshake)

Adds an HMAC `acceptProof` over `{pairingId, sessionId, liveTabId}` keyed by the launch secret so a forged `pairing-accept` can't dead-lock the controller; controller fails closed without a valid proof, and tightens the `sigTs` window to asymmetric (≤30s old / ≤1s future).

**Verdict: nothing blocks merge.** Security claims verified against checked-out code (pairing-manager / controller-channel / acceptance / interactive-step suites green):

- Launch secret never leaves the live tab — `getAcceptedSession` exposes only `{sessionId, publicKeyB64, liveTabId, pairingId}`; the secret lives module-private and only ever leaves as an HMAC. Added `pairingId` field is additive/non-breaking (sole consumer reads only `sessionId`).
- Challenge proof (`{pairingId, publicKeyB64, sessionId}`) and accept proof (`{liveTabId, pairingId, sessionId}`) are domain-separated by distinct field sets → no cross-protocol replay.
- Controller fails closed: `!pairing` short-circuits the accept handler before any bind; verify must return `ok` first.
- No TOCTOU/double-bind: the async `.then` re-checks `pairedLiveIdRef.current !== null` before assigning.
- Tightened `sigTs` window can't reject legit same-clock messages (prepared hand-back is re-signed every heartbeat and past-dated).

### Medium (non-blocking)
- No direct negative test asserting a challenge proof is **rejected** by `verifyPairingAcceptProof` — domain separation is currently only proven structurally + via the forged-accept integration test. Worth one explicit cross-protocol negative test in `pairing-manager.test.ts` to lock the invariant against future canonical-payload refactors.
- Latent footgun (not a bug): `challengeKey` (rejected-ledger) still includes `senderTabId` while `sameChallenge` no longer does. Correct today, but a one-line comment noting the divergence is intentional would help future readers.

Jayclifford345:
Addressed both non-blocking notes in c39c2f08:

- Added an explicit cross-protocol negative test in `pairing-manager.test.ts`: a challenge proof (`{pairingId, publicKeyB64, sessionId}`) is now asserted to fail `verifyPairingAcceptProof` (`{pairingId, sessionId, liveTabId}`) even with the same secret and a shared pairingId/sessionId, alongside a positive control so the rejection can't pass vacuously. That locks the domain separation against a future canonical-payload refactor.
- Added a short comment on `challengeKey` noting the `senderTabId` divergence from `sameChallenge` is intentional (the rejected-ledger is keyed per originating tab; challenge identity is the crypto triple).

Heads up: I also rebased this onto latest main and unstacked it, so it's now a standalone PR against main rather than the bottom of the two-tab controller stack.

moxious (approved):

PR #1171, opened 2026-06-29
refactor(security): consolidate duplicated WebCrypto into one audited module (#1159)

### LINKED ISSUE #1159: Consolidate duplicated WebCrypto core into a single audited security/ module
Part of the two-tab controller epic #1133. Follow-up from review of #1153 — explicitly deferred out of that PR to keep the authenticated-pairing security change tightly scoped.

### Problem
Two modules carry overlapping WebCrypto implementations (~40–55 lines of overlap):
- `src/security/cross-tab-crypto.ts` (ECDSA sign/verify + HMAC for launch binding)
- `src/integrations/workshop/session-crypto.ts` (ECDSA challenge-response for presenter auth)

The `toBase64url`/`fromBase64url` helpers are byte-identical, and `ECDSA_PARAMS` / `SIGN_PARAMS` / `generateSessionKeyPair` / the ECDSA sign+verify pair are near-identical. Two copies of crypto-encoding logic can silently drift on key or signature encoding, which is the kind of divergence that turns into a subtle interop or security bug.

**The drift has already begun:** the private-key `extractable` flag differs between the copies — `cross-tab-crypto.ts` mints non-extractable keys (`generateKey(..., false, ...)`), while `session-crypto.ts` mints extractable ones (`true`). The workshop copy doesn't need extractability to export its public key, so this is also a latent hardening opportunity.

### Proposal
Fold the shared base64url + ECDSA keygen + sign/verify core into one audited module under `src/security/` (Tier 1), and have both `cross-tab-crypto.ts` and `workshop/session-crypto.ts` consume it. HMAC helpers can stay in or alongside the shared module since cross-tab is the only consumer.

**Must-haves:**
- Keep `extractable` an **explicit parameter** of the shared keygen — do not paper over the current divergence; decide per call site and default to non-extractable.
- Preserve both existing test suites (`session-crypto.test.ts`, plus the cross-tab acceptance/unit coverage) green with no behavioral change — pure refactor.
- Respect the import-tier model: shared crypto lives in Tier 1 `security/`; `integrations/workshop` (Tier 3) importing downward from it is allowed.

### Out of scope
- Any protocol or behavioral change. This is a non-functional consolidation.

### Acceptance criteria
- [ ] Single shared crypto module; no duplicated base64url/keygen/sign-verify between the two files.
- [ ] `extractable` is an explicit keygen parameter; workshop's key extractability decision is made deliberately.
- [ ] All existing crypto tests pass unchanged; `npm run check` green.

Jayclifford345:
## What

Consolidates the duplicated WebCrypto core shared by `src/security/cross-tab-crypto.ts` and `src/integrations/workshop/session-crypto.ts` into one audited Tier-1 module, `src/security/webcrypto.ts`, and resolves the `extractable`-flag divergence the two copies had drifted into.

Closes #1159. Part of #1133.

## The duplication

The two modules carried byte-identical `toBase64url`/`fromBase64url`, identical `ECDSA_PARAMS`/`SIGN_PARAMS`, and near-identical ECDSA keygen/sign/verify. The copies had already drifted on the private-key `extractable` flag: cross-tab minted **non-extractable** keys (`false`), workshop minted **extractable** ones (`true`), justified by an incorrect comment ("extractable is set to true so we can export the public key") — public keys export regardless of the private key's flag, which the cross-tab copy already proves.

## The change

`src/security/webcrypto.ts` (new) owns the shared primitive:

- `toBase64url` / `fromBase64url`
- `generateEcdsaKeyPair(extractable: boolean)` — `extractable` is an **explicit parameter**, no hidden default
- `signBytes(privateKey, bytes)` / `verifyBytes(publicKeyB64, bytes, sig)` — byte-level ECDSA P-256, so each caller prepares its own input

Both consumers now delegate to it:

- `cross-tab-crypto.ts`: `signPayload`/`verifyPayload` wrap `signBytes`/`verifyBytes` over `TextEncoder().encode(payload)`. HMAC helpers (`signHmacPayload`/`verifyHmacPayload`) stay here — cross-tab is their only consumer — and now import the shared base64url helpers.
- `session-crypto.ts`: `signChallenge`/`verifyChallenge` wrap the same primitive over `fromBase64url(nonce)`.

**Extractability decided deliberately:** both call sites now pass `false`. Workshop's private key is hardened to non-extractable — it only ever signs and exports its public key, never the private key, so it never needed extractability (a latent hardening, transparent to behavior).

## Equivalence / behavior

Pure refactor — the bytes signed and verified are unchanged on both sides:

- cross-tab: `sign(encode(payload))` before and after.
- workshop: `sign(fromBase64url(nonce))` before and after.

The crypto is self-consistent and ephemeral (per-session keys, no persisted/cross-version material), so the passing real-WebCrypto suites prove interop. New `src/security/webcrypto.test.ts` covers the shared module directly (base64url round-trip, the `extractable` flag, sign/verify happy-path + cross-key/garbage rejection). `session-crypto.test.ts` and the cross-tab acceptance suite pass unchanged. `npm run test:ci`, typecheck, eslint, prettier, and the import-tier governance tests are green (the new `security/` → consumer edges are legal: cross-tab-crypto stays Tier 1; workshop Tier 3 imports downward).

### DIFF
--- a/src/integrations/workshop/session-crypto.ts
+++ b/src/integrations/workshop/session-crypto.ts
@@ -1,108 +1,27 @@
 /**
- * Cryptographic utilities for workshop session authentication
- *
- * Implements asymmetric challenge-response presenter verification using ECDSA P-256
- * via the browser-native Web Crypto API. No external dependencies.
- *
- * Security model:
- *   - Presenter generates an ECDSA key pair at session creation
- *   - The PUBLIC key is embedded in the join code (safe to share)
- *   - The PRIVATE key never leaves the presenter's browser session
- *   - Attendees challenge the presenter with a random nonce; only the holder of
- *     the private key can produce a valid signature — knowing the join code
- *     (and therefore the public key) does not allow impersonation
+ * Workshop presenter authentication: ECDSA P-256 challenge-response. The public
+ * key ships in the join code; the non-extractable private key stays in the
+ * presenter's session, so knowing the join code can't impersonate the presenter.
  */
+import { fromBase64url, generateEcdsaKeyPair, signBytes, toBase64url, verifyBytes } from '../../security/webcrypto';
 
-// ============================================================================
-// Base64url helpers
-// ============================================================================
-
-function toBase64url(bytes: Uint8Array): string {
-  let binary = '';
-  for (let i = 0; i < bytes.byteLength; i++) {
-    binary += String.fromCharCode(bytes[i]!);
-  }
-  return btoa(binary).replace(/\+/g, '-').replace(/\//g, '_').replace(/=+$/, '');
-}
-
-function fromBase64url(str: string): Uint8Array {
-  const padded = str.replace(/-/g, '+').replace(/_/g, '/');
-  const padLength = (4 - (padded.length % 4)) % 4;
-  const base64 = padded + '='.repeat(padLength);
-  const binary = atob(base64);
-  const bytes = new Uint8Array(new ArrayBuffer(binary.length));
-  for (let i = 0; i < binary.length; i++) {
-    bytes[i] = binary.charCodeAt(i);
-  }
-  return bytes;
-}
-
-// ============================================================================
-// Public API
-// ============================================================================
-
-const ECDSA_PARAMS = { name: 'ECDSA', namedCurve: 'P-256' } as const;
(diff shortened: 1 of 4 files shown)


moxious (approved):
nice PR.  Easy & simple, was needed.

Issue #1149, opened 2026-06-25
[docs/architecture] Two-tab controller: add cross-tab-controller concern entry to CONCERNS.md before re-enable


moxious:
## Problem

`src/integrations/cross-tab/**`, `src/lib/cross-tab-transport.ts`, `src/types/cross-tab.types.ts`, and `src/global-state/controller-channel.tsx` are not listed in any subsystem concern's `trigger_paths` in `docs/design/CONCERNS.md`. The `security` concern is always-on so review coverage is not zero, but the cross-tab subsystem is one of the most security-sensitive surfaces in the codebase:

- The `BroadcastChannel` is same-origin and forgeable by any co-plugin, panel XSS, or extension content script
- The live-tab executor turns inbound messages into real DOM mutations and navigation on an authenticated Grafana session
- The pairing handshake (`gesture-to-accept`, `acceptedSenderIds`), the double-validation gate (transport receive + executor sink), and the `KNOWN_TARGET_ACTIONS` allowlist are all load-bearing trust invariants with no dedicated review routing

Without an explicit concern entry, a PR that widens `KNOWN_TARGET_ACTIONS`, weakens a per-kind validator, or alters mount ordering has no automated concern activation telling a reviewer to look closely at those invariants.

Discovered during review of #1148 — see the review comment there.

### Routing table row (subsystem block in `docs/design/CONCERNS.md`)

```
| `cross-tab-controller` | sub | N | strong | 1 | 8 | `src/integrations/cross-tab/**`, `src/lib/cross-tab-transport.ts`, `src/types/cross-tab.types.ts`, `src/global-state/controller-channel.tsx`, `src/utils/pathfinder-search-params.ts` | `CrossTabTransport`, `installLiveTabExecutor`, `validateCrossTabMessage`, `KNOWN_TARGET_ACTIONS`, `KIND_VALIDATORS`, `acceptedSenderIds`, `pathfinder-cross-tab`, `pathfinder-pairing-accepted`, `step-command`, `check-requirements`, `fix-requirement`, `TWOTAB_CONTROLLER_ENABLED`, `ControllerChannelProvider` |
```

### Concern review table entry

| field | value |
|---|---|
[...]

PR #1165, opened 2026-06-29
docs(concerns): add cross-tab-controller concern entry (#1149)

Jayclifford345:
## What

Adds a dedicated `cross-tab-controller` subsystem concern so security-relevant changes to the two-tab interactive controller route to a focused reviewer **before `TWOTAB_CONTROLLER_ENABLED` is flipped on** (it is still `false`).

- **`docs/design/CONCERNS.md`** — new routing-table row + new review-table entry, placed next to `workshop-collaboration` (the other cross-tab/collaboration trust-boundary subsystem).
- **`docs/developer/CROSS_TAB_CONTROLLER.md`** — short pointer note at the top of the *Security / trust model* section naming the concern as the canonical review checklist.

`min_signals = 1`: every touch of these files is load-bearing for the trust model, so the concern fires on a single file hit (same rationale as `build-and-ci`).

## Reconciled the proposed entry against current code

The issue's proposed entry was written against an earlier, simpler trust model. The cross-tab subsystem has since gained a full **cryptographic authentication** layer (`pairing-manager.ts`, `cross-tab-crypto.ts`, ECDSA-signed commands). I verified every path/symbol/constant in the proposal against the codebase and reconciled it so the entry routes reviewers to things that **actually exist** — shipping dead symbols into a security-routing doc would defeat its purpose. Changes vs. the verbatim proposal:

| Proposed | Action | Why |
|---|---|---|
| `acceptedSenderIds` | **removed** | No longer exists anywhere in `src/`. |
| `pathfinder-pairing-accepted` | **fixed → `pairing-accept`** | Real `KIND_VALIDATORS` kind is `pairing-accept` (`cross-tab.types.ts`); the proposed string matches nothing. |
| — | **added** `src/lib/pairing-manager.ts`, `src/security/cross-tab-crypto.ts` to `trigger_paths`/`load_code` | HMAC launch-binding + ECDSA signing are now the core trust mechanism. |
| — | **added** keywords `verifySignedMessage`, `createControllerPairingLaunch`, `acceptSession`, `sidebar-handoff` | Executor auth gate, launch binding, gesture-accept, and the 4th signed command kind. |
| — | **added** review questions + one-way doors for the signature/auth gate (`verifySignedMessage` — session binding, ±30s window, replay-nonce ledger), HMAC pairing challenge, and ECDSA key extractability | The trust model these defend didn't exist when the issue was filed. |

[...]

PR #1173, opened 2026-06-29
test(cross-tab): real-browser adversarial specs for the two-tab controller (#1142)

### LINKED ISSUE #1142: [testing] Two-tab controller: real-browser adversarial smoke tests before re-enable
## Context

The two-tab controller ships dark pending resolution of this and 5 sibling items.

## Problem

All security-critical tests for the two-tab controller run in jsdom, which does not implement `BroadcastChannel` natively — the channel is mocked. The following adversarial scenarios are only exercised in mock-land:

- A forged `step-command` is rejected (not dispatched to the DOM sink).
- A second live tab that is NOT the paired tab does not execute commands.
- Cancel/retry ignores stale `step-complete` and `step-progress` messages.
- A forged pre-heartbeat reply cannot claim the pairing slot.

In a real browser, `BroadcastChannel` delivers messages asynchronously across actual JS contexts. The timing and delivery semantics differ from the mock in ways that could mask bugs.

## What needs to happen

Add real-browser E2E smoke tests (using the existing Playwright/E2E runner — see `docs/developer/E2E_TESTING.md`) covering:

1. **Forged command rejected**: open a live tab, send a `step-command` from a non-paired sender, confirm no DOM action occurs.
2. **Second live tab isolation**: open a controller and two live tabs; confirm only the paired tab executes.
3. **Cancel/retry stale-message safety**: start a composite, cancel mid-run, retry; confirm the old `step-complete` does not settle the new run.
4. **Pairing slot protection**: send a forged reply before any heartbeat; confirm it cannot claim the pairing slot.

These tests should pass before `TWOTAB_CONTROLLER_ENABLED` is flipped to `true`.

## Gate

This issue is a hard gate before `TWOTAB_CONTROLLER_ENABLED` is flipped to `true`.

Jayclifford345:
## What

Adds real-browser adversarial Playwright specs for the two-tab controller — `tests/two-tab-controller.spec.ts` — covering the scenarios that jsdom (where `BroadcastChannel` is mocked and synchronous) cannot reproduce.

Closes #1142. Part of #1133.

## Skipped until the flag flips

The specs are a `test.describe.fixme` group, so they are **registered but skipped**. The controller overlay, the live-tab executor, and the "open in interactive window" affordance are all gated on the hardcoded `TWOTAB_CONTROLLER_ENABLED = false`, with no runtime override — so they aren't reachable in a normal build. To run them: build with the flag enabled and remove the `.fixme`. The file header documents this.

This matches the issue framing — these are the gate that must pass *before* the flag is flipped — and lands the concrete scenarios now so they're ready to run at that moment, without shipping a runtime toggle of the security gate (which #1149's concern flags as a one-way door).

## Scenarios

1. **Forged command rejected** — a raw, unsigned `step-command` posted on `pathfinder-cross-tab` does not drive the live tab (asserts no navigation).
2. **Second-tab isolation** — controller paired with live A; a driven command reaches A but not an unpaired live B (the signed `liveTabId` binds the command to the paired tab).
3. **Stale-message safety** — a composite run cancelled then retried is not settled by a stale `step-complete` carrying the old `runId`.
4. **Pairing-slot protection** — a forged `pairing-accept` (correct `sessionId` read off the wire, bogus proof) cannot claim the controller's slot; the genuine accept still pairs (exercises #1158's HMAC-bound accept in a real browser).

Forging is done via `page.evaluate` posting raw messages on a real `BroadcastChannel`, and multi-tab scenarios use multiple pages in one browser context (shared same-origin channel).

## Status / caveat

Because the feature is off in this build, these specs cannot be executed here — they are a faithful scaffold built on the real selectors (`testIds.docsPanel.openControllerTabButton`, `testIds.guideReader.controllerStatus`, the `Accept` pairing button) and the real channel/message contracts. The pairing orchestration (cross-tab `window.open` capture + fragment handoff) should be validated when the flag is enabled and the group is unskipped. Typecheck and prettier are green; `tests/` is outside the eslint config (repo convention); jest (`test:ci`) does not run Playwright specs.

### DIFF
--- a/tests/two-tab-controller.spec.ts
+++ b/tests/two-tab-controller.spec.ts
@@ -1,0 +1,183 @@
+import { test, expect } from './fixtures';
+import { testIds } from '../src/constants/testIds';
+import { CROSS_TAB_CHANNEL } from '../src/types/cross-tab.types';
+import type { Page } from '@playwright/test';
+
+/**
+ * Real-browser adversarial coverage for the two-tab controller (#1142).
+ *
+ * The security-critical cross-tab tests elsewhere run in jsdom, where
+ * BroadcastChannel is mocked and message delivery is synchronous. These specs
+ * exercise the same trust boundary across real, separate JS contexts where the
+ * channel delivers asynchronously — the conditions the mock can't reproduce.
+ *
+ * SKIPPED (`test.describe.fixme`) until the `enableTwoTabController` admin
+ * setting is turned on (plugin config → Interactive features): the controller
+ * overlay, the live-tab executor, and the "open in interactive window"
+ * affordance are all gated on that setting (default off), so none of this is
+ * reachable on a default instance. To run these, enable the setting (or seed it
+ * in the test instance's plugin jsonData) and remove the `.fixme`.
+ */
+
+const SENTINEL_PATH = '/a/grafana-pathfinder-app/__forged_nav__';
+
+async function postRawCrossTabMessage(page: Page, message: Record<string, unknown>): Promise<void> {
+  await page.evaluate(
+    ({ channel, msg }) => {
+      const ch = new BroadcastChannel(channel);
+      ch.postMessage(msg);
+      ch.close();
+    },
+    { channel: CROSS_TAB_CHANNEL, msg: message }
+  );
+}
+
+async function startCapturingSessionId(page: Page): Promise<void> {
+  await page.evaluate((channel) => {
+    const w = window as unknown as { __pfCapturedSessionId?: string };
+    const ch = new BroadcastChannel(channel);
+    ch.onmessage = (event) => {
+      const data = event.data as { kind?: string; sessionId?: string };
+      if (data?.kind === 'pairing-challenge' && typeof data.sessionId === 'string') {
+        w.__pfCapturedSessionId = data.sessionId;
+        ch.close();
+      }
+    };
+  }, CROSS_TAB_CHANNEL);
+}
+
+async function readCapturedSessionId(page: Page): Promise<string> {
+  return page.evaluate(() => {
+    const w = window as unknown as { __pfCapturedSessionId?: string };
+    return w.__pfCapturedSessionId ?? '';
+  });
+}
+
+async function gotoLiveTab(page: Page): Promise<void> {
+  await page.goto('/');
+  await page.waitForLoadState('networkidle');
+}
+
+async function openController(liveTab: Page): Promise<Page> {
+  await liveTab.locator('button[aria-label="Help"]').click();
+  await expect(liveTab.getByTestId(testIds.docsPanel.container)).toBeVisible();
+
+  const openButton = liveTab.getByTestId(testIds.docsPanel.openControllerTabButton);
+  await expect(openButton).toBeVisible();
+
+  const controllerPromise = liveTab.context().waitForEvent('page');
+  await openButton.click();
+  const controller = await controllerPromise;
+  await controller.waitForLoadState('networkidle');
+  return controller;
+}
+
+async function acceptPairing(liveTab: Page): Promise<void> {
+  await liveTab.getByRole('button', { name: 'Accept' }).click();
+}
+
+test.describe.fixme('two-tab controller — real-browser adversarial (pending enableTwoTabController)', () => {
+  test('a forged, unsigned step-command does not drive the live tab', async ({ page }) => {
+    await page.goto('/');
+    await page.waitForLoadState('networkidle');
+    const urlBefore = page.url();
+
+    await postRawCrossTabMessage(page, {
+      source: 'pathfinder',
+      senderId: 'attacker-tab',
+      timestamp: Date.now(),
+      kind: 'step-command',
+      phase: 'do',
+      stepId: 'forged',
+      runId: 'forged-run',
+      action: { targetAction: 'navigate', refTarget: SENTINEL_PATH },
+    });
+
+    await page.waitForTimeout(500);
+    expect(page.url()).toBe(urlBefore);
+  });
+
+  test('only the paired live tab executes a controller command', async ({ page, context }) => {
+    const liveA = page;
+    await gotoLiveTab(liveA);
+    const controller = await openController(liveA);
+    await acceptPairing(liveA);
+
+    const liveB = await context.newPage();
+    await liveB.goto('/');
+    await liveB.waitForLoadState('networkidle');
+    const liveBUrlBefore = liveB.url();
+
+    await controller.getByRole('button', { name: /do it/i }).first().click();
+
+    await expect.poll(() => liveA.url()).not.toBe(liveBUrlBefore);
+    expect(liveB.url()).toBe(liveBUrlBefore);
+  });
+
+  test('a stale step-complete from a cancelled run does not settle the retried run', async ({ page }) => {
+    const liveA = page;
+    await gotoLiveTab(liveA);
+    const controller = await openController(liveA);
+    await acceptPairing(liveA);
+
+    const composite = controller.getByRole('button', { name: /do section|do it/i }).first();
+    await composite.click();
+    await controller
+      .getByRole('button', { name: /cancel/i })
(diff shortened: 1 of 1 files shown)


moxious (changes requested):
**Request changes** — two issues should be fixed before this scaffold merges. The overall shape is right (skipped `fixme` group, no runtime override of the security gate, fully reversible), and I verified the channel name, selectors, flag, fixtures, and the `step-command` / `step-complete` / `pairing-challenge` shapes all match the real contracts. But because the specs are skipped and unrunnable in this build, nothing mechanically guards their correctness until the flag flips — so faithfulness to the real contract is the bar to clear now, and scenario 4 doesn't clear it.

---

### 1. Scenario 4 never asserts the forged accept was rejected (blocking)

The test is titled *"a forged pairing-accept cannot claim the controller pairing slot,"* but the only assertion is:

```ts
await expect(status).toContainText(/connected/i);
```

run **after** both the forged accept and the genuine `acceptPairing(liveA)`. That checks only that the genuine accept eventually connects — it never asserts the forged accept *failed* to pair. The test would pass whether or not the forge succeeded, so it does not exercise the security property it claims.

This is a live concern against `main`, not just a future one: `isValidPairingAccept` requires only `typeof message.sessionId === 'string'` — there is no proof check on accept today. A forged accept carrying the correct `sessionId` (which the test reads off the wire) is a structurally valid accept, so the forge would actually succeed and the test would still pass on its weak assertion.

**Fix:** add a negative assertion — e.g. that immediately after the forged accept (before the genuine one) the controller is *not* `connected`, or that the paired tab is specifically `liveA`.

### 2. `acceptProof` / `pairingId` on the forged accept are invented fields (blocking)

The forged `pairing-accept` posts:

```ts
kind: 'pairing-accept', sessionId, pairingId: 'guessed', acceptProof: 'forged',
```

`acceptProof` exists nowhere in the codebase, and `PairingAcceptMessage` is `{ kind, sessionId }` only — `isValidPairingAccept` ignores both extra fields. The PR body says this *"exercises #1158's HMAC-bound accept,"* but that proof-on-accept contract hasn't landed in `main` (per the open #1158 gate), so the "bogus proof" is a no-op today, and the guessed field name is unlikely to match what #1158 actually ships (the challenge side uses `pairingProof`, not `acceptProof`).

This drift is invisible to typecheck because `postRawCrossTabMessage` takes `Record<string, unknown>`, which is exactly why a skipped scaffold needs it caught now rather than after unskip.

**Fix:** either drop the speculative fields until #1158 lands, or align the field name with #1158's actual accept-proof contract and add a comment tying the two together so they're reconciled when that PR merges.

---

These two are coupled: fixing #1 properly (asserting the forge is rejected) requires the forged message in #2 to actually represent an invalid accept, which it can't until the accept-proof contract exists. Worth sequencing this behind #1158 so scenario 4 tests a real rejection path.

Jayclifford345:
Both fixed in the latest push:

1. Scenario 4 now actually asserts the rejection — after the forged accept plus a `live` heartbeat from the attacker, it checks the controller does *not* connect, then the genuine accept pairs. It only passes if the forge is rejected.
2. The accept-proof contract landed in #1158, which is now stacked directly below this PR — so `acceptProof`/`pairingId` are real fields and `acceptProof: 'forged'` is a genuinely invalid proof the controller rejects. Added a comment tying it to #1158.

moxious:
## Review — security-sensitive (this is the pre-flip acceptance gate)

Adds `tests/two-tab-controller.spec.ts` — 4 real-browser Playwright adversarial specs (forged command, second-tab isolation, stale-runId, forged `pairing-accept`) for scenarios jsdom can't reach, all in a `test.describe.fixme` block so they're registered but never executed. Selectors and the `acceptProof` field verify against the real base contract.

**This PR has blockers.** An acceptance gate that has never once executed is unverified — and scenarios 1–2 below show concretely that unexecuted specs hide real assertion bugs.

### [HIGH] The `.fixme` rationale is now factually wrong
The header/PR body justify skipping with "the feature is gated on `TWOTAB_CONTROLLER_ENABLED = false` with no runtime override in this build." But the stack order is **#1174 → #1165 → #1173**: #1174 (already below this PR) removes that const and adds the runtime `enableTwoTabController` admin toggle. The build this lands on **does** have a runtime override, so the stated reason no longer holds. Either (a) un-`fixme`, wire enablement via `jsonData.enableTwoTabController = true` (feasible — `tests/fixtures.ts` already uses `gotoAppConfigPage`), and land at least one green run; or (b) if there's a real CI blocker, state that concrete blocker instead.

### [MEDIUM] Scenario 1 passes vacuously
The forged unsigned `step-command` is posted into a `page.goto('/')` tab that was never paired and has no executor installed. The URL stays put because **nothing is listening**, not because the signature gate rejected it. Run it against a paired tab with the live-tab executor installed (same setup as scenarios 2–4), then assert rejection there.

### [MEDIUM] Scenario 2 isolation assertion is weak / inverted
It never captures `liveA`'s starting URL and asserts `liveA.url()` `.not.toBe(liveBUrlBefore)` — comparing A's URL to B's baseline. If A and B start on different URLs this passes without any command executing. Capture `liveAUrlBefore`, assert A navigated to the expected target while B is unchanged from its own baseline.

### [MEDIUM] Coverage gaps in the adversarial suite
Omits attack vectors the contract explicitly defends: no **replay** test (re-post a genuine signed command → assert ledger ignores it), no **future/expired `sigTs`** test, no **signature-stripping** test. At minimum add the replay spec — the replay ledger is a named element of the contract and exactly what jsdom's synchronous mock can't exercise. The rest acceptable as tracked follow-up if scope is held to #1142's four.

**On the never-run `.fixme`:** normally a faithful registered-but-skipped scaffold is fine to land ahead of flag-flip. Not here — this suite is the explicit pre-flip acceptance criterion, and the dependency the `.fixme` blames is resolved by #1174 within this same stack, below this PR. Recommend: un-`fixme`, wire enablement, get one green run, fix scenarios 1–2 before merge.

No other blockers — typecheck/prettier green, `tests/` correctly outside eslint, selector/message-contract references verified.

Jayclifford345:
Addressed all four in the latest push (3ca8c710):

**[HIGH] `.fixme` rationale** — corrected, but keeping the specs skipped, because the premise had drifted. This branch is **not** stacked on #1174 — it's standalone on `main` (a single commit; `git merge-base --is-ancestor` confirms #1174 is not an ancestor). Here the controller is still gated by the compile-time `TWOTAB_CONTROLLER_ENABLED = false` const with **no runtime override**; `enableTwoTabController` only exists in #1174. So there's no build this branch can produce that reaches the overlay / live-tab executor, and per the re-enable gate the #1142 specs stay registered-but-skipped until the flag flips. I rewrote the header + `describe` title to state that concrete blocker instead of the stale `enableTwoTabController` reference, and left the enablement recipe (seed `jsonData.enableTwoTabController = true` via `gotoAppConfigPage`, remove `.fixme`) for when #1174 lands and the flag flips.

**[MEDIUM] Scenario 1 vacuous** — now runs against a paired tab with the live-tab executor installed (`gotoLiveTab → openController → acceptPairing`) before posting the forged unsigned `step-command`. A stalled URL is now attributable to the signature gate rejecting it, not to nothing listening.

**[MEDIUM] Scenario 2 inverted** — now captures `liveAUrlBefore` and asserts A moved from *its own* baseline while B is unchanged from *B's own* baseline (was comparing A's URL against B's baseline).

**[MEDIUM] Coverage gap** — added the replay spec: it captures a genuine signed `step-command` off the channel, re-posts it verbatim (same `{sessionId, sigNonce}`), and asserts no re-dispatch — exercising the nonce ledger (`seenSignedMessages`) that jsdom's synchronous mock can't reach. Future/expired-`sigTs` and signature-stripping specs remain tracked follow-up, per your note about holding scope to #1142's four.

Good catch on scenarios 1–2 — both were passing for the wrong reason. Typecheck + prettier green; specs remain `.fixme` (unexecuted by design on this branch).

moxious (approved):

Issue #1144, opened 2026-06-24
[hardening] Two-tab controller: deepen wire validation for internalActions fields, requirement-result.error[], and progress indices


moxious:
## Context

Part of the two-tab interactive controller epic (#1133).

This is a **post-re-enable hardening item**, not a gate. The feature can be re-enabled once the 5 gate items are resolved; this issue addresses incremental validation depth.

## Problem

The current `KIND_VALIDATORS` table validates the presence and top-level shape of each message kind but does not validate:

1. **`internalActions` element fields** — each element is checked for a recognized `targetAction`, but `refTarget` (string), `targetValue` (string | undefined), and `phase` (`show` | `do`) are not type-checked. A malformed `internalActions` element passes the gate.

2. **`requirement-result.error[]`** — the `errors` array in requirement results is not shape-validated. A forged error array with unexpected types could reach consumer code.

3. **Progress index bounds** — `step-progress` carries `{ stepId, index, total }` but does not validate that `index >= 0`, `total >= 1`, or `index <= total`. Out-of-range values reach the UI progress display directly.

## What needs to happen

Extend the relevant validator cases in `src/types/cross-tab.types.ts`:

- `step-command`: validate each `internalActions[]` element's `refTarget: string`, `phase` in `{show, do}`, and optionally `targetValue: string | undefined`.
- `requirement-result`: validate `errors` is an array of `{ kind: string, message: string }` (or whatever the actual shape is).
- `step-progress`: validate `index >= 0 && total >= 1 && index <= total`.

PR #1174, opened 2026-06-29
feat(cross-tab): make the two-tab controller an admin plugin-config toggle (default off)

### LINKED ISSUE #1004: feat: External browser tab mode for custom guides

Jayclifford345:
## What

Replaces the compile-time `TWOTAB_CONTROLLER_ENABLED` constant with an admin-controlled plugin-config setting, `enableTwoTabController` (default **off**), so an instance admin can enable the two-tab interactive controller without a custom build. Mirrors the existing `enableAiAutoHeal` opt-in.

Part of #1133.

## Why

The gate was a hardcoded `const = false` with no runtime override — the feature could only be enabled by editing source and rebuilding. Moving it to `jsonData` (plugin config → **Interactive features**) lets admins opt in per-instance while the feature still ships dark by default.

## Changes

- **`src/constants.ts`** — `DEFAULT_ENABLE_TWO_TAB_CONTROLLER = false`, `enableTwoTabController?: boolean` on `DocsPluginConfig`, and the `getConfigWithDefaults` resolution (exactly the `enableAiAutoHeal` shape).
- **`src/components/AppConfig/InteractiveFeatures.tsx`** — a "Two-tab interactive controller" toggle with a warning Alert (it drives the authenticated DOM), wired through draft state / reset / save / `hasChanges` like the other toggles.
- **`src/constants/testIds.ts`** — `interactiveFeatures.enableTwoTabController`.
- **Gate sites now read `config.enableTwoTabController`:** `module.tsx` (controller-overlay entry + live-tab-executor install — `config` is already computed there from `meta.jsonData`) and `DocsPanelContentArea.tsx` (the "Open in interactive window" affordance, via `usePluginContext()` + `getConfigWithDefaults`, the same pattern other in-panel components use — no `as any` global).
- **`src/constants/interactive-config.ts`** — the `TWOTAB_CONTROLLER_ENABLED` const removed.
- **`docs/developer/CROSS_TAB_CONTROLLER.md`** — gating sections updated: both mount sites now require the admin setting **and** `shouldMountSidebar(...)`.

## Default & security

Default is **off** — enabling is an explicit admin action. Note this changes the nature of the re-enable one-way door: instead of a code flip, any admin who toggles this makes the live-tab executor reachable in production. The trust-boundary work this gate protects (authenticated pairing #1153/#1158, deeper validation #1144, and the real-browser adversarial tests #1142) should be landed before admins are encouraged to turn it on. The config UI carries an experimental/security warning to that effect.

## Cross-PR note

The `cross-tab-controller` concern (#1149 / PR #1165) and the E2E specs (#1142 / PR #1173) reference `TWOTAB_CONTROLLER_ENABLED` by name; whichever merges after this should update those references to the `enableTwoTabController` config setting.

## Verification

`npm run test:ci` (incl. new `getConfigWithDefaults` coverage for the flag), typecheck, eslint, prettier, and `npm run build` all green. No dangling references to the removed const.

### DIFF
--- a/src/components/AppConfig/InteractiveFeatures.tsx
+++ b/src/components/AppConfig/InteractiveFeatures.tsx
@@ -12,6 +12,7 @@ import {
   DEFAULT_ENABLE_KIOSK_MODE,
   DEFAULT_KIOSK_RULES_URL,
   DEFAULT_ENABLE_AI_AUTO_HEAL,
+  DEFAULT_ENABLE_TWO_TAB_CONTROLLER,
   getConfigWithDefaults,
 } from '../../constants';
 import { updatePluginSettings } from '../../utils/utils.plugin';
@@ -26,6 +27,7 @@ type State = {
   enableKioskMode: boolean;
   kioskRulesUrl: string;
   enableAiAutoHeal: boolean;
+  enableTwoTabController: boolean;
 };
 
 export interface InteractiveFeaturesProps extends PluginConfigPageProps<AppPluginMeta<JsonData>> {}
@@ -44,6 +46,7 @@ const InteractiveFeatures = ({ plugin }: InteractiveFeaturesProps) => {
     enableKioskMode: jsonData?.enableKioskMode ?? DEFAULT_ENABLE_KIOSK_MODE,
     kioskRulesUrl: jsonData?.kioskRulesUrl ?? DEFAULT_KIOSK_RULES_URL,
     enableAiAutoHeal: jsonData?.enableAiAutoHeal ?? DEFAULT_ENABLE_AI_AUTO_HEAL,
+    enableTwoTabController: jsonData?.enableTwoTabController ?? DEFAULT_ENABLE_TWO_TAB_CONTROLLER,
   }));
   const [isSaving, setIsSaving] = useState<boolean>(false);
   const [validationErrors, setValidationErrors] = useState<Record<string, string>>({});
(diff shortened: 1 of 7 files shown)


leslie-heinzen (dismissed):
:shipit:

moxious:
## Review — security-sensitive (changes the re-enable one-way door)

Replaces the compile-time `TWOTAB_CONTROLLER_ENABLED` const with an admin `jsonData` setting `enableTwoTabController` (default off), mirroring `enableAiAutoHeal`, and converts all three gate sites (`module.tsx` controller-overlay entry + live-tab-executor install, `DocsPanelContentArea.tsx` affordance) to read it.

**Verdict: nothing blocks merge** — the gate plumbing is correct and complete:

- All `TWOTAB_CONTROLLER_ENABLED` references gone (`git grep` clean across `src/`, `docs/`, `.cursor/`).
- `getConfigWithDefaults` resolves the flag via `?? DEFAULT_ENABLE_TWO_TAB_CONTROLLER` (false); a missing/malformed `jsonData` value cannot resolve truthy.
- All three gates read resolved config; no `as any` global reads (`DocsPanelContentArea` uses the same `usePluginContext()` pattern as the sibling AI flag).
- Toggle wired through draft/reset/save/`hasChanges` identically to sibling toggles; save preserves other tabs' fields. Coverage added for default-off and explicit-true.

### Medium — sequencing (owner decision, not a code defect)
This converts the re-enable one-way door from "edit source + rebuild" to "any admin flips a toggle," making the authenticated DOM-driving executor reachable in production **before** the gate items it protects (#1142 real-browser adversarial tests, #1144 deeper validation) have all landed. The code correctly ships dark and warns on opt-in, so this is a release/product decision — recommend the epic owner explicitly confirm landing the runtime toggle ahead of #1142/#1144 is intended. Consider strengthening the in-product Alert from "experimental" to "not yet fully hardened (see #1133)."

### Medium — cross-PR housekeeping (already noted in PR body)
This PR correctly removes the const; ensure the merge order resolves the stale `TWOTAB_CONTROLLER_ENABLED` references in #1165 and #1173 so a dangling const reference doesn't ship. (#1165 verified already updated; #1173 still references the old reason — see its review.)

moxious (approved):

PR #1181, opened 2026-06-29
fix(cross-tab): write formfill via Monaco model API so it lands on a backgrounded live tab

### LINKED ISSUE #1004: feat: External browser tab mode for custom guides

Jayclifford345:
## What

Fixes formfill steps that target the Monaco query editor not taking effect on the **live tab when driven from the two-tab controller** (they worked fine in the in-sidebar path). Discovered on the `prometheus-advanced-queries` guide, whose query steps are `formfill` into `textarea.inputarea`.

Stacked on top of the two-tab controller stack (#1133).

## Root cause

`FormFillHandler.setMonacoEditorValue` wrote to Monaco by setting the **hidden textarea's native `.value`** and dispatching synthetic `input`/`keydown` events, relying on Monaco to mirror the textarea into its model. That mirroring only happens while the editor's tab is **focused/foreground**. When the guide is driven from the popped-out controller window, the live Grafana tab is **backgrounded**, so Monaco ignores the synthetic input and the query never lands — even though the executor runs the formfill end-to-end (element resolves, Monaco is detected, clear+set run).

Confirmed by tracing the live-tab path: `dispatching step-command (formfill) → runAction → formfill handleDoMode { isMonacoEditor: true, shouldClear: true }` all fire, but the editor stays empty. Plain `element.click()` actions (button/highlight) aren't foreground-sensitive, which is why those worked.

## Fix

Write through the **Monaco model API** instead of the hidden textarea — `editor.setValue()` / `model.setValue()` go straight to the editor model and are foreground-independent. New `trySetMonacoModelValue(element, value)` in `code-block-handler.ts` (where `window.monaco` access already lives — same pattern as the existing `clearAndInsertCode`); `setMonacoEditorValue` calls it first and **falls back to the existing synthetic-textarea path** when the Monaco API or a matching model isn't found.

Behavior-equivalent for the in-tab path: `setMonacoEditorValue` already did a full clear-then-set, and `setValue()` is likewise a full replace — so this is strictly more robust, not a semantic change.

## Tests

New `code-block-handler.test.ts` covers `trySetMonacoModelValue`: API-unavailable → false, no `[data-uri]` editor → false, matching editor → `setValue` + `focus`, and the `getModels()` fallback. `form-fill-handler.test.ts` and the interactive-step suite stay green (jsdom has no `window.monaco`, so they exercise the synthetic fallback unchanged). `npm run test:ci`, typecheck, eslint, prettier, and `npm run build` all pass.

## Out of scope

The `prometheus-advanced-queries` guide is poorly authored (bare top-level steps; should use sections) — a separate guide-quality PR will address that. This PR is the engine fix that makes its formfill steps work under the controller regardless.

### DIFF
--- a/src/interactive-engine/action-handlers/code-block-handler.test.ts
+++ b/src/interactive-engine/action-handlers/code-block-handler.test.ts
@@ -46,12 +46,18 @@ describe('trySetMonacoModelValue', () => {
     expect(focus).toHaveBeenCalled();
   });
 
-  it('falls back to the matching model when no editor matches', () => {
+  it('falls back to the matching model when no editor matches, without touching the wrong editor', () => {
     const modelSetValue = jest.fn();
+    const wrongEditorSetValue = jest.fn();
+    const wrongEditorFocus = jest.fn();
     const monaco: MonacoMock = {
       editor: {
         getEditors: () => [
-          { getModel: () => ({ uri: { toString: () => 'other' } }), setValue: jest.fn(), focus: jest.fn() },
+          {
+            getModel: () => ({ uri: { toString: () => 'other' } }),
+            setValue: wrongEditorSetValue,
+            focus: wrongEditorFocus,
+          },
         ],
         getModels: () => [{ uri: { toString: () => URI }, setValue: modelSetValue }],
       },
@@ -60,5 +66,8 @@ describe('trySetMonacoModelValue', () => {
 
     expect(trySetMonacoModelValue(textareaUnder(URI), 'v')).toBe(true);
     expect(modelSetValue).toHaveBeenCalledWith('v');
+    // The non-matching editor is left alone, and the model-fallback path never focuses.
+    expect(wrongEditorSetValue).not.toHaveBeenCalled();
+    expect(wrongEditorFocus).not.toHaveBeenCalled();
   });
 });
--- a/src/interactive-engine/action-handlers/code-block-handler.test.ts
+++ b/src/interactive-engine/action-handlers/code-block-handler.test.ts
@@ -1,0 +1,64 @@
+import { trySetMonacoModelValue } from './code-block-handler';
+
+type MonacoMock = NonNullable<Window['monaco']>;
+
+describe('trySetMonacoModelValue', () => {
+  const URI = 'inmemory://model/1';
+
+  function textareaUnder(uri: string): HTMLElement {
+    const root = document.createElement('div');
+    root.setAttribute('data-uri', uri);
+    const textarea = document.createElement('textarea');
+    root.appendChild(textarea);
+    document.body.appendChild(root);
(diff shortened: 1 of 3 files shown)


moxious:
## Review

Adds `trySetMonacoModelValue(element, value)` in `code-block-handler.ts` and calls it first in `FormFillHandler.setMonacoEditorValue`, writing through Monaco's model API (foreground-independent) instead of synthetic textarea events, so query-editor formfill lands on a **backgrounded** live tab driven from the controller; falls back to the synthetic path when the API/model isn't found.

**Verdict: nothing blocks merge.** Correct, minimal, well-targeted. Verified:

- **Wrong-editor targeting safe** — both loops match `model.uri.toString() === dataUri`, the same per-model URI disambiguation `tryMonacoApi` already uses; multiple Monaco instances handled correctly.
- **Element resolution correct** — formfill passes `textarea.inputarea`, so `element.closest('[data-uri]')` walks up to the editor root; the `?? querySelector` covers the code-block case.
- **Fallback preserved** — returns `false` on every miss → drops to the unchanged synthetic path. jsdom has no `window.monaco`, so existing suites exercise the fallback unchanged (claim verified).
- **`focus()` benign** on a backgrounded tab; **no new security surface** — raw text to model, ECDSA command contract untouched, and the textarea path never sanitized either.

### Medium (optional, not required to merge)
- **DRY:** `trySetMonacoModelValue` largely duplicates the URI-match loop in `tryMonacoApi` (`code-block-handler.ts:78-95`). Consider having `tryMonacoApi` delegate to it to collapse the two copies.
- **Test gap:** the multi-editor test gives a non-matching editor a mock `setValue` but never asserts it stayed **uncalled**, so the "doesn't write the wrong editor" guarantee isn't actually pinned. One extra assertion (plus asserting the model-fallback branch does not call `focus()`) would lock it.

Jayclifford345:
Addressed both non-blocking notes in the latest push (91df3e7a):

- **DRY:** `tryMonacoApi` now delegates to `trySetMonacoModelValue` instead of carrying a second copy of the editor/model URI-match loops. Its sole caller (`clearAndInsertCode`) reads only `.success` and discards the error on failure (falls through to the textarea path), so collapsing the per-branch error strings into one is non-breaking. As a small bonus the delegated path is slightly more robust — `element.closest('[data-uri]') ?? element.querySelector('[data-uri]')` also handles the case where the resolved container *is* the editor element, which the old `container.querySelector` missed.
- **Test gap:** the model-fallback case now gives the non-matching editor named mocks and asserts both `setValue` and `focus` on it stay **uncalled**, pinning "doesn't write the wrong editor" and "model-fallback branch doesn't focus."

Targeted jest (4/4), typecheck, and prettier all green.

moxious (approved):

PR #1183, opened 2026-06-29
[SPIKE] feat(e2e): add cloud stack isolation for guide tests

leslie-heinzen:
## Summary

Adds classifier-driven isolation for cloud-tier E2E guide tests so guides that may mutate Grafana Cloud state can run against disposable stacks instead of polluting a shared target. The runner can now cold-provision ephemeral stacks, lease pre-provisioned hot-pool stacks, mint short-lived runner tokens, retire leased stacks after use, and immediately create replacements for the pool.

## What to look at

- E2E cloud routing and side-effect detection: [`side-effects.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/6ea1044fe293f0d9b1ea5f430e6ff37dee553096/src/cli/e2e/side-effects.ts), [`e2e-package.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/6ea1044fe293f0d9b1ea5f430e6ff37dee553096/src/cli/e2e/e2e-package.ts), and [`e2e-results.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/6ea1044fe293f0d9b1ea5f430e6ff37dee553096/src/cli/e2e/e2e-results.ts) classify remote guide content, propagate the classification into reports, and add the `skipped_unsafe_shared_stack` outcome for unsafe chains that cannot get an isolated stack.
- Stack provisioning and leasing: [`cloud-stack-environment.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/6ea1044fe293f0d9b1ea5f430e6ff37dee553096/src/cli/e2e/cloud-stack-environment.ts), [`cloud-stack-pool.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/6ea1044fe293f0d9b1ea5f430e6ff37dee553096/src/cli/e2e/cloud-stack-pool.ts), and [`cloud-provisioning.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/6ea1044fe293f0d9b1ea5f430e6ff37dee553096/src/cli/e2e/cloud-provisioning.ts) contain the Terraform-backed cold stack path, pool discovery, per-chain target/token selection, teardown, and immediate replacement behavior.
- CLI integration: [`e2e.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/6ea1044fe293f0d9b1ea5f430e6ff37dee553096/src/cli/commands/e2e.ts) wires the new Cloud stack flags into remote resolution, skips shared-stack preflight for isolated chains, and routes Playwright to the leased/provisioned stack.
- Operator documentation: [`E2E_TESTING.md`](https://github.com/grafana/grafana-pathfinder-app/blob/6ea1044fe293f0d9b1ea5f430e6ff37dee553096/docs/developer/E2E_TESTING.md) documents required Cloud Access Policy scopes, pool labels, Terraform dependency, side-effect classifications, and follow-up gaps such as manifest-declared plugin installation.

## Why

Cloud auth on the current branch isolates users and sessions per dependency chain, but it does not isolate org-global Grafana Cloud state such as dashboards, folders, data sources, and alerting resources. We considered relying only on service-account isolation plus cleanup, but that would be incomplete for guides that mutate existing or product-specific state; this spike instead gives unsafe cloud chains a disposable stack while preserving the faster shared-stack path for read-only guides.

## How to verify

1. Label one or more pre-created Grafana Cloud stacks with `pathfinder-e2e-pool=true`, then run a mutating cloud package with `--cloud-stack-access-policy-token <env var>` and confirm the runner leases one of those stacks, executes the guide, retires the leased stack, and creates a replacement.
2. Run the same package with no pool stack available but with `--cloud-stack-region <region>` and confirm the runner falls back to cold Terraform stack provisioning.
3. Run a cloud package that the classifier marks read-only with a matching `--cloud-instance-admin-token host=env` and confirm it keeps using the shared-stack service-account path.
4. Run an unsafe cloud package without pool labels or `--cloud-stack-region` and confirm it exits with `skipped_unsafe_shared_stack` instead of running against the shared stack.
5. Inspect a JSON report from a remote package run and confirm package metadata includes the side-effect classification used for routing.

## Breaking changes

None. This change is additive and keeps existing local, bundled, remote, and shared-stack cloud E2E flows available.

## Reversibility

Reverting this PR removes the new isolated-stack routing and returns cloud package E2E runs to the previous service-account-only behavior. Any Grafana Cloud stacks created or destroyed while using the new flags are external side effects and cannot be restored by a git revert; failed teardown leaves labeled stacks discoverable for manual cleanup.

## Out of scope

- Installing plugins declared by `testEnvironment.plugins` on ephemeral stacks is deferred. Today those plugin requirements are still checked by preflight and requirements logic, but they are not automatically installed.
- Durable cross-process lease coordination for hot pools is deferred. The current pool lease tracking is process-local, so concurrent CLI invocations still require external coordination.
- Manifest schema changes such as `stateIsolation` or `sideEffects` are deferred because `testEnvironment` is an external manifest contract.

### DIFF
--- a/docs/developer/E2E_TESTING.md
+++ b/docs/developer/E2E_TESTING.md
@@ -56,6 +56,11 @@ npx pathfinder-cli e2e [options] [files...]
 | `--resolver-url <url>`                    | Recommender base URL for `--package <id>` resolution                                                                                                     | `https://recommender.grafana.com` |
 | `--cloud-instance-admin-token <host=env>` | Admin service-account token env var for a cloud target. Repeat for multiple cloud instances.                                                             | None                              |
 | `--cloud-url <url>`                       | Default Grafana Cloud instance URL for cloud-tier guides without a manifest `instance`.                                                                  | `https://learn.grafana.net/`      |
+| `--cloud-stack-access-policy-token <env>` | Cloud Access Policy token env var for provisioning ephemeral Grafana Cloud stacks for state-mutating cloud guide chains.                                 | None                              |
+| `--cloud-stack-region <region>`           | Grafana Cloud region for ephemeral stacks. Required when stack provisioning is enabled.                                                                  | None                              |
+| `--cloud-stack-slug-prefix <prefix>`      | Alphanumeric prefix for generated ephemeral stack slugs. Hyphens are stripped because Cloud stack slugs must be alphanumeric.                            | `pfe2e`                           |
+| `--cloud-stack-plugin-version <version>`  | Pathfinder plugin version to install on ephemeral stacks.                                                                                                | `latest`                          |
+| `--cloud-stack-pool-id <id>`              | Optional pool namespace. When omitted, any stack labeled `pathfinder-e2e-pool=true` is eligible.                                                         | None                              |
 
 ### Input formats
 
@@ -382,6 +387,10 @@ These variables are consumed by the CLI or passed to the spawned Playwright proc
 
 For cloud targets, pass `--cloud-instance-admin-token host=ENV_VAR_NAME`; the named env var contains an admin service-account token for that exact host. The env var name is user-defined, for example `GRAFANA_PLAY_ADMIN_TOKEN`.
 
+For ephemeral Cloud stacks, pass `--cloud-stack-access-policy-token ENV_VAR_NAME` and `--cloud-stack-region <region>`. The named env var contains a Grafana Cloud Access Policy token with stack lifecycle and plugin-install permissions. The CLI passes this token to Terraform through `TF_VAR_cloud_access_policy_token`; it is not written into generated Terraform configuration.
+
+For a manual hot pool, pass `--cloud-stack-access-policy-token ENV_VAR_NAME`. The CLI discovers stacks labeled `pathfinder-e2e-pool=true`, leases one, and mints a short-lived service-account token for the runner. Add `--cloud-stack-pool-id <id>` to restrict leasing to stacks whose `pathfinder-e2e-pool-id` label matches that value. Pool leases are tried before cold stack provisioning.
+
 ## Error classification
 
 When a step fails, the runner assigns an error classification to help with triage:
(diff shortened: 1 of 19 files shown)


leslie-heinzen:
## `/review` — PR #1183: cloud stack isolation for guide tests

**Disposition: Approve with minor comments.** This is a well-structured spike with strong security hygiene and good test coverage on the routing/classifier core. No blocking issues. One functional gap in the hot-pool replenishment path is worth addressing or explicitly documenting before anyone relies on the pool; the rest are non-blocking.

**Routing.** Classified as `mixed` (CLI/E2E runner tooling + external stateful effects + token/URL handling + tests + docs). Concerns reviewed: always-on (`security`, `correctness-and-reliability`, `testing-and-verification`, `reversibility-and-one-way-door`, `cross-cutting-architecture`) + subsystem `cli-and-e2e-runner` + standalone deep-security lens (tokens, credential minting, URL/HCL construction, external trust boundary).

### Findings (by severity, then confidence)

**1. [medium · high confidence] Auto-replenished pool stacks never get Pathfinder installed → unusable replacements**
`concern: cli-and-e2e-runner / correctness-and-reliability` · reversibility: `partially_reversible` (creates orphaned, non-functional Cloud stacks)
- `CloudStackPoolLease.teardownChain()` deletes the leased stack and calls `createCloudStackPoolStack(...)` to "immediately create a replacement" (advertised in the PR summary). But `poolStackModule()` in `cloud-stack-pool-create.ts` only creates a bare `grafana_cloud_stack` with labels — no `grafana_cloud_plugin_installation` resource.
- The lease path (`CloudStackPool.lease` → `tokenModule`) only mints a service-account token; unlike the cold path (`CloudStackEnvironment.provisionChain`), it never probes/installs the plugin.
- Net effect: every auto-created replacement enters the pool **without** `grafana-pathfinder-app`, contradicting the documented invariant ("Each stack must have Pathfinder installed or pre-provisioned"). The next run that leases it will fail at preflight/requirements (plugin missing) instead of skipping cleanly, and the pool silently degrades over time.
- Suggested action: either install/probe Pathfinder in the replacement module (reuse the cold-path probe+install), or explicitly scope replacements out in the docs/"Out of scope" and disable auto-replacement until the plugin-install path covers pool stacks.

**2. [low · high confidence] CLI inputs interpolated into generated HCL without validation (inconsistent with `slug-prefix`)**
`concern: security (defense-in-depth) / correctness-and-reliability` · reversibility: `reversible`
- `--cloud-stack-region` → `region_slug = "${config.region}"`, `--cloud-stack-plugin-version` → `version = "${pluginVersion || 'latest'}"` (cloud-stack-environment.ts), and `--cloud-stack-pool-id` → HCL label value (cloud-stack-pool-create.ts `hclStringMap`) are all interpolated raw. `slugPrefix` is normalized/validated but these are not.
- Security impact is minimal: these are operator-supplied and the operator already holds the Cloud Access Policy token, so this is not a privilege-escalation vector. The real cost is robustness/consistency — a stray `"` or `${` yields a cryptic Terraform parse error instead of a clear CLI validation message (and is a latent injection seam if a flag value is ever sourced from less-trusted CI input).
- Suggested action: validate `region`/`pluginVersion`/`poolId` against simple allowlist patterns (as done for `slug-prefix`) and fail fast in the config builders.

**3. [low · medium confidence] Replenishment ordering and leaked runner-token SAs**
`concern: reversibility-and-one-way-door` · reversibility: `partially_reversible`
- Replacement is created *after* the leased stack is deleted; if `createCloudStackPoolStack` fails, the pool net-shrinks by one (warn-only). Acceptable for a spike but worth a tracking note.
- In pool-only runs (no `--cloud-stack-region`), there is no `sweepCloudStacks` pass (it's gated on `inputs.cloudStack`), so runner-token service accounts left by a failed lease teardown are only bounded by the 3600s TTL, not swept. Fine given TTL, but note it.

**4. [low · medium confidence] Direct test coverage gaps on the CLI wiring**
`concern: testing-and-verification`
- `createCloudStackPoolStack` is only exercised indirectly (via the lease-teardown `['destroy','init','apply']` assertion); the generated pool HCL (labels, plugin absence) isn't asserted.
- The new `e2e.ts` branches — `coldStackConfigRequested`, `shouldUseEphemeralStackForMeta`, and the `targetUrlsToCheck` `size === 0 && packageMetaById.size === 0` change — have no direct unit test. The routing helpers in `cloud-provisioning.ts` and `e2e-targets.ts` *are* well covered.

### What's done well
- Access Policy token is passed to Terraform via `TF_VAR_cloud_access_policy_token` (env), never written into generated `.tf`; the TF variable is `sensitive = true`.
- Terraform error output is redacted for the access policy token (`redact`/`assertTerraformSuccess`), and tests assert HCL never contains the secret.
- All outbound `fetch` calls use `AbortSignal.timeout(15s)`; provision failures tear down the partially-created stack; TTL (3600s) + grace + `sweepCloudStacks` provide a safety net for cold stacks.
- The side-effect classifier is conservative, recursive over all block types, and explainable (carries `reasons` with JSON paths); missing/unknown classification fails safe to "unsafe" (`skipped_unsafe_shared_stack`) rather than mutating a shared stack. `GuideStatus`/`FAILURE_STATUSES`/`PRE_RUN_SKIP_STATUSES` are all updated consistently for the new outcome.

### Tech debt (non-blocking)
- **QC3/QC4 duplication** across `cloud-stack-environment.ts`, `cloud-stack-pool.ts`, and `cloud-stack-pool-create.ts`: `defaultCommandRunner`, `redact`, `assertCommandSuccess`/`assertTerraformSuccess`, `terraformEnv`/`runnerEnv`, `normalizeSlugPrefix`, `generatedSlug`/`stackSlug`, the `terraform { required_providers ... } provider "grafana"` HCL preamble, and the `TOKEN_TTL_SECONDS` / `TERRAFORM_PROVIDER_VERSION` constants are duplicated verbatim. `CloudStackListResponse` is also defined twice. Consider a small shared `cloud-stack-terraform.ts` helper module. Comment hygiene (QC8) is good — the new files are appropriately comment-light.

### Documentation drift (non-blocking)
- The `cli-and-e2e-runner` concern's `trigger_paths` (CONCERNS.md:57) list `src/cli/commands/**` and `docs/developer/E2E_TESTING.md` but **not** `src/cli/e2e/**`, which is where nearly all of this PR's logic (and the pre-existing `cloud-*.ts` files) lives. Recommend adding `src/cli/e2e/**` to that concern's `trigger_paths` so future cloud-runner changes route to a subsystem reviewer rather than only the always-on concerns.
- `docs/developer/E2E_TESTING.md` itself is updated thoroughly (flags table, side-effect classification, hot-pool workflow, required Access Policy scopes, the new outcome row) — no drift there.

### Coverage note
`cli-and-e2e-runner` activated via `docs/developer/E2E_TESTING.md` and `src/cli/commands/e2e.ts`, but the bulk of the new code sits in `src/cli/e2e/**`, which the concern registry maps only weakly (see doc-drift above). I read the full source of all new/changed files, so confidence is high, but this is the kind of cluster the routing table should model explicitly.

Issue #1187, opened 2026-06-29
Provision ephemeral Grafana Cloud stacks for unsafe E2E chains


leslie-heinzen:
Parent: #1184

## Problem

Unsafe cloud guide chains need clean Grafana Cloud stacks when no hot-pool stack is available.

## Scope

- Provision a Grafana Cloud stack with `delete_protection=false`.
- Mint a short-lived Admin service-account token for the runner.
- Probe for `grafana-pathfinder-app` and install it only when missing.
- Destroy the stack after the guide chain completes.
- Redact Cloud Access Policy tokens from logs and errors.
- Keep setup and teardown errors clear without hiding the actual guide test result.

## Acceptance criteria

- The E2E runner can cold-provision an isolated stack and run a cloud guide chain against it.
- Stack teardown is best-effort and reports cleanup failures.
- Tests cover token redaction, plugin probe/install behavior, and destroy-on-failure paths.

Issue #1184, opened 2026-06-29
epic(e2e): Isolate Grafana Cloud E2E guide runs from shared stack state


leslie-heinzen:
## Problem

Cloud E2E guide testing now supports isolated logins by minting short-lived service accounts per guide dependency chain. That isolates per-user state, but it does **not** isolate Grafana Cloud stack state.

Guides that create or mutate org-global resources — dashboards, folders, data sources, alerting resources, plugin config, or other stack-level state — can pollute the shared Cloud test stack and cause later guide tests to fail for reasons unrelated to the guide under test.

We need a way to run state-mutating cloud guides against isolated Grafana Cloud stacks while keeping read-only guides fast and cheap.

## Solution

Build a Cloud E2E isolation layer with three pieces:

1. **Side-effect classification**
   - Statically classify remote guide content as `readonly`, `possibly_mutating`, `mutating`, or `unknown`.
   - Use the classification to decide whether a guide can safely run on a shared stack or needs an isolated stack.

2. **Ephemeral stack execution**
   - For unsafe cloud guide chains, provision or lease a disposable Grafana Cloud stack.
   - Mint a short-lived runner service-account token for the stack.
   - Run the guide chain against that stack.
   - Destroy the used stack after the run.

3. **Hot pool support**
   - Keep a small pool of pre-provisioned Cloud stacks labeled for E2E use.
   - Lease a pool stack before falling back to cold provisioning.
   - After a pool stack is used, retire it and immediately create a replacement.
   - A later scheduled replenisher can reconcile capacity and recover from crashed runs.

Read-only cloud guides should continue to use the faster shared-stack service-account path when safe.

## PRs

| PR | Purpose | Status |
| --- | --- | --- |
| #1183 | Spike: side-effect classifier, cold stack provisioning, hot-pool leasing, immediate pool replacement, and E2E docs | Draft |
| #1196 | feat(e2e): add cloud guide side-effect routing — classifier and shared-stack routing (closes #1185, #1186) | Merged |
| #1211 | refactor(e2e): clarify shared cloud stack targeting | Merged |
| #1212 | feat(e2e): add cold cloud stack lifecycle | Merged |
| #1213 | feat(e2e): wire cold cloud stack routing | Merged |

## Key design decisions

- **Read-only guides should stay fast.** If a cloud guide is classified as read-only and a matching shared-stack service-account token is available, it should keep using the existing shared-stack path.
[...]

PR #1211, opened 2026-06-30
refactor(e2e): clarify shared cloud stack targeting

leslie-heinzen:
> [!NOTE]
> For broader context, see the **E2E Cloud isolation epic**: https://github.com/grafana/grafana-pathfinder-app/issues/1184
>
> This is the first setup PR for #1187 **(Part 1 of 3)**

## Summary

Renames the existing cloud E2E service-account lifecycle to `SharedCloudStackEnvironment` and clarifies target resolution around shared stack auth; it also keeps instance-declared guides pinned to their declared shared stack instead of allowing them to fall back to future isolated-stack execution.

## What to look at

- **CLI E2E target resolution**: [`e2e-targets.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/794f4af0309c63fd736b40622fc006e2af1294f0/src/cli/e2e/e2e-targets.ts) now models cloud execution as target capabilities (`sharedStackUrls`, future `isolatedStack`) and requires matching shared-stack auth for manifest-declared `instance` guides.
- **Shared stack lifecycle naming**: [`shared-cloud-stack-environment.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/794f4af0309c63fd736b40622fc006e2af1294f0/src/cli/e2e/shared-cloud-stack-environment.ts) is the old per-chain service-account lifecycle under a more precise name.
- **Remote package resolution**: [`e2e-package.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/794f4af0309c63fd736b40622fc006e2af1294f0/src/cli/e2e/e2e-package.ts) threads the renamed target capabilities into package resolution so later cold-stack routing can build on the same contract.

## Why

Some Cloud guides must be tested against a specific shared stack, such as manifests that declare `play.grafana.org`. This PR makes that invariant explicit before adding cold isolated stack execution: an instance-declared guide needs matching shared-stack auth, while non-instance cloud guides can later be routed to an isolated stack when appropriate.

## Out of scope

- Cold isolated stack provisioning is handled in the next PR.
- CLI routing from unsafe cloud chains to cold stacks is handled after the cold-stack lifecycle module lands.

## How to verify

No manual runtime verification is needed for this rename-focused PR. Review the target-resolution tests for the key behavior: instance-declared cloud guides skip without matching shared-stack auth, while non-instance cloud target resolution remains available for later isolated-stack routing.

## Breaking changes

Refs #1184
Refs #1187

### DIFF
--- a/src/cli/commands/e2e.ts
+++ b/src/cli/commands/e2e.ts
@@ -21,7 +21,12 @@ import {
 import { planGuideExecution } from '../e2e/guide-chains';
 import type { TestResultsData } from '../e2e/e2e-reporter';
 import { checkTier, loadManifestFromDir, runManifestPreflight, type CurrentTier } from '../e2e/manifest-preflight';
-import { resolveRemotePackage, resolveRemoteRepository, type ResolvedRemoteGuide } from '../e2e/e2e-package';
+import {
+  resolveRemotePackage,
+  resolveRemoteRepository,
+  type RemoteResolveOptions,
+  type ResolvedRemoteGuide,
+} from '../e2e/e2e-package';
 import { checkGrafanaHealth } from '../e2e/grafana-health';
 import { CleanEnvironment, CLEAN_COMPOSE_PROJECT, CLEAN_GRAFANA_URL } from '../e2e/clean-environment';
 import { ExitCode } from '../e2e/exit-codes';
@@ -741,13 +746,13 @@ async function resolveRunInputs(files: string[], options: E2ECommandOptions): Pr
   const cloudAuth = createCloudAuthPolicy({
     cloudInstanceAdminTokenSpecs: options.cloudInstanceAdminToken,
   });
-  const remoteOptions = {
+  const remoteOptions: RemoteResolveOptions = {
     grafanaUrl: options.grafanaUrl,
     currentTier: options.tier,
     resolverUrl: options.resolverUrl,
     repoUrl: options.repoUrl,
     cloudUrl: options.cloudUrl,
-    cloudAuthTargets: cloudAuth.targets,
+    cloudTargetCapabilities: cloudAuth.targets,
   };
   const resolution =
     mode === 'remote-package'
@@ -857,7 +862,7 @@ export const e2eCommand = new Command('e2e')
       await maybeCleanStart(cleanEnv, options);
 
       await sweepCloudTargets({
-        targetUrls: inputs.cloudAuth?.targets.provisionable ?? [],
+        targetUrls: inputs.cloudAuth?.targets.sharedStackUrls ?? [],
         cloudAuth: inputs.cloudAuth,
         verbose: options.verbose,
       });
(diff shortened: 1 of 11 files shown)


moxious (dismissed):
## Request changes — one blocking fix

Nice, well-scoped refactor and the `resolveTarget` capability seam reads correctly. One blocking issue before merge.

### 🔴 Blocking — stale option key silently disables all cloud E2E targeting

`e2e-package.ts` renames the resolve option `cloudAuthTargets` → `cloudTargetCapabilities` on `RemoteResolveOptions`, and every reader inside that module now reads `options.cloudTargetCapabilities`. But the sole production caller wasn't updated:

```ts
// src/cli/commands/e2e.ts:744
const remoteOptions = {
  ...
  cloudUrl: options.cloudUrl,
  cloudAuthTargets: cloudAuth.targets,   // ← stale key; should be cloudTargetCapabilities
};
const resolution = mode === 'remote-package'
  ? await resolveRemotePackage(options.package!, remoteOptions)
  : await resolveRemoteRepository(remoteOptions);
```

**Why this compiles but breaks at runtime:** `remoteOptions` is assigned to an un-annotated `const` and *then* passed to the resolver functions. TypeScript only applies excess-property checking to fresh object literals at the assignment/call site — once it's a variable, the extra `cloudAuthTargets` property is permitted and the optional `cloudTargetCapabilities` is simply absent. So `npm run check` passes, but at runtime `options.cloudTargetCapabilities` is `undefined` in both `resolveRemotePackage` and `resolveRemoteRepository`.

**Impact:** `resolveTarget` then defaults to `{ sharedStackUrls: [] }`, so **every cloud-tier guide resolves to `skipped_no_auth`** regardless of configured `--cloud-instance-admin-token`, defeating the cloud E2E path entirely. The unit tests don't catch it because each constructs its own options object with the new key directly — the `e2e.ts` wiring is the one spot the rename missed, and it's untested.

**Fix** — rename the key at `e2e.ts:750`:

```ts
    cloudTargetCapabilities: cloudAuth.targets,
```

Consider annotating `const remoteOptions: RemoteResolveOptions = {…}` so a future rename surfaces as a compile error instead of a silent drop.

---

### 🟢 Nits (non-blocking)

- `e2e-targets.ts` — the loop variable in `hasSharedStackAuthFor` is `sharedStackUrlsTargetUrl`; suggest `sharedStackUrl`.
- Test names read awkwardly after the mechanical rename: `cloud-auth.test.ts` "declares a cloud URL sharedStackUrls when…" and `cloud-provisioning.test.ts` "deduplicates sharedStackUrls targets by URL".

Otherwise: security clean (no trust-boundary change, admin tokens never flow through `CloudTargetCapabilities`), fully reversible (additive flag + rename), and no doc drift. The new `isolatedStack` runnable branch is currently test-only with no production caller, which is expected for Part 1 of 3.

leslie-heinzen:
Gah, a late-changing refactor after agent review resulted in the bug you caught. This is what I get for not explicitly typing variables (`remoteOptions`)!

Requested fix (and nits) here: https://github.com/grafana/grafana-pathfinder-app/compare/defdfe822a9b017b6bc9c13a6cdd56ef6d401386..2011109cfe4c4b1f9113625ae78371486ff0b628

Thanks!

Jayclifford345 (approved):

PR #1212, opened 2026-06-30
feat(e2e): add cold cloud stack lifecycle

leslie-heinzen:
> [!NOTE]
> For broader context, see the **E2E Cloud isolation epic**: https://github.com/grafana/grafana-pathfinder-app/issues/1184. 
>
> This change is part of the cold isolated stack lifecycle work in https://github.com/grafana/grafana-pathfinder-app/issues/1187 **(Part 2 of 3)**.

## Summary

Adds the standalone `ColdCloudStackEnvironment` lifecycle for creating disposable Grafana Cloud stacks, minting short-lived runner tokens, probing/installing Pathfinder, and best-effort destroying the stack after use. This PR only introduces the cold-stack implementation and tests; it does not yet route E2E guide chains through it.

## What to look at

- **Cold Cloud stack lifecycle**: [`cold-cloud-stack-environment.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/3bd098ab8995d5fa5df78f74b808efa629e7eee7/src/cli/e2e/cold-cloud-stack-environment.ts) owns the Terraform module generation, env-var based Cloud Access Policy token handling, plugin probe/install pass, teardown memoization, and cleanup registry used for signal-triggered teardown.

## Why

Unsafe Cloud guide chains need a disposable stack path before the runner can stop skipping them. Keeping the cold-stack lifecycle isolated in its own module makes the Terraform, token-redaction, plugin-install, and teardown behavior reviewable before CLI routing is added.

## How to verify

No manual runtime verification is expected for this PR because the module is not wired into the E2E command yet. Review the unit coverage for config parsing, generated Terraform safety, plugin probe/install behavior, partial setup teardown, token redaction, and concurrent teardown.

## Breaking changes

## Out of scope

- Routing unsafe Cloud guide chains to this lifecycle is handled separately.
- Hot-pool leasing and durable pool coordination are not included here.
- Installing manifest-declared plugins beyond `grafana-pathfinder-app` remains future work.

Refs #1184
Refs #1187

### DIFF
--- a/src/cli/e2e/cold-cloud-stack-environment.ts
+++ b/src/cli/e2e/cold-cloud-stack-environment.ts
@@ -1,0 +1,417 @@
+import { spawn } from 'child_process';
+import { randomUUID } from 'crypto';
+import { chmodSync, mkdtempSync, rmSync, writeFileSync } from 'fs';
+import { tmpdir } from 'os';
+import { join } from 'path';
+import { CLOUD_STACK_FETCH_TIMEOUT_MS } from './shared-cloud-stack-environment';
+
+const TERRAFORM_PROVIDER_VERSION = '~> 4.5';
+const PLUGIN_ID = 'grafana-pathfinder-app';
+const TOKEN_TTL_SECONDS = 3600;
+export const DEFAULT_CLOUD_STACK_SLUG_PREFIX = 'pfe2e';
+
+export interface ColdCloudStackConfigInput {
+  accessPolicyTokenEnvVar?: string;
+  region?: string;
+  slugPrefix?: string;
+  pluginVersion?: string;
+  env?: NodeJS.ProcessEnv;
+}
+
+export interface ColdCloudStackProvisioningConfig {
+  accessPolicyTokenEnvVar: string;
+  accessPolicyToken: string;
+  region: string;
+  slugPrefix: string;
+  pluginVersion?: string;
+}
+
+export interface ProvisionedCloudStack {
+  targetUrl: string;
+  token: string;
+  stackSlug: string;
+}
+
+export interface CommandResult {
+  exitCode: number;
+  stdout: string;
+  stderr: string;
+}
+
+export type CommandRunner = (
+  command: string,
+  args: string[],
+  options: { cwd: string; env: NodeJS.ProcessEnv }
+) => Promise<CommandResult>;
+
+interface TerraformOutput {
+  stack_url?: { value?: unknown };
+  stack_slug?: { value?: unknown };
+  service_account_token?: { value?: unknown };
+}
+
+interface CloudStackTeardownTarget {
+  teardownChain(): Promise<string[]>;
+}
+
+export class ColdCloudStackCleanupRegistry {
+  private readonly targets = new Set<CloudStackTeardownTarget>();
+
+  track(target: CloudStackTeardownTarget): void {
+    this.targets.add(target);
+  }
+
+  untrack(target: CloudStackTeardownTarget): void {
+    this.targets.delete(target);
+  }
+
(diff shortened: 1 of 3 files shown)


src/cli/e2e/cold-cloud-stack-environment.ts
```diff
@@ -0,0 +1,415 @@
+import { spawn } from 'child_process';
+import { randomUUID } from 'crypto';
+import { mkdtempSync, rmSync, writeFileSync } from 'fs';
+import { tmpdir } from 'os';
+import { join } from 'path';
+
+const TERRAFORM_PROVIDER_VERSION = '>= 4.5.3';
```
Jayclifford345: `>= 4.5.3` has no upper bound, so the day the grafana provider hits 5.x, CI pulls it and E2E could break with no change on our side. `~> 4.5` or a pinned version keeps runs reproducible.

src/cli/e2e/cold-cloud-stack-environment.ts
```diff
@@ -0,0 +1,415 @@
...
+
+  private async isPathfinderPluginInstalled(stack: ProvisionedCloudStack): Promise<boolean> {
+    const url = new URL(`/api/plugins/${PLUGIN_ID}/settings`, stack.targetUrl).toString();
+    const response = await this.fetchImpl(url, {
+      method: 'GET',
+      headers: {
+        Authorization: `Bearer ${stack.token}`,
+        Accept: 'application/json',
+      },
+      signal: AbortSignal.timeout(15_000),
```
Jayclifford345: nit: the sibling `shared-cloud-stack-environment.ts` names this `FETCH_TIMEOUT_MS` — same 15s, files right next to each other. Worth a shared constant.

src/cli/e2e/cold-cloud-stack-environment.ts
```diff
@@ -0,0 +1,415 @@
...
+    private readonly verbose: boolean,
+    private readonly runner: CommandRunner = defaultCommandRunner,
+    private readonly fetchImpl: typeof fetch = fetch
+  ) {
+    this.secrets = [config.accessPolicyToken];
+  }
+
+  async provisionChain(): Promise<ProvisionedCloudStack> {
+    const slug = generatedStackSlug(this.config.slugPrefix);
+    const moduleDir = mkdtempSync(join(tmpdir(), 'pathfinder-e2e-stack-'));
```
Jayclifford345, line 317: Non-blocking, just flagging: `terraform apply` drops the minted token into `terraform.tfstate` here in plaintext. It stays contained — 0700 temp dir, cleaned on teardown, 1h TTL if we crash first — and the long-lived access policy token never hits state since it comes in via `TF_VAR_*`. So this is the only secret-at-rest surface, and it looks fine.

Jayclifford345 (commented):
Solid module — redaction is thorough, operator strings go through `TF_VAR_*` instead of interpolated HCL, teardown is memoized against concurrent calls, and failures always clean up. Mirrors `SharedCloudStackEnvironment` nicely.

Nothing blocking. One design question on the two-phase apply, the rest is minor.

leslie-heinzen:
@Jayclifford345 Thanks for the review. I addressed the nits.

You mention "One design question on the two-phase apply, the rest is minor." but I don't see a question about two-phase apply in there. Can you clarify? 

I came to this approach actually because I was running into an error with _always_ including the pathfinder plugin in the apply step -- if the plugin is installed on the stack already, terraform apply fails. Hence the "sniff and re-apply if necessary" second step. 

But maybe the plugin will _always_ be applied for grafana cloud, so this is unnecessary (possibly unused)?

Jayclifford345 (approved):

PR #1213, opened 2026-06-30
feat(e2e): wire cold cloud stack routing

### LINKED ISSUE #1187: Provision ephemeral Grafana Cloud stacks for unsafe E2E chains

leslie-heinzen:
> [!NOTE]
> For broader context, see the **E2E Cloud isolation epic**: https://github.com/grafana/grafana-pathfinder-app/issues/1184. 
>
> This change completes the cold isolated stack lifecycle work in https://github.com/grafana/grafana-pathfinder-app/issues/1187 **(Part 3 of 3)**.

## Summary

Routes unsafe or unauthenticated non-instance Cloud E2E guide chains through cold-provisioned Grafana Cloud stacks when cold-stack config is available. Read-only shared-stack runs remain on the existing service-account path, while unsafe chains without isolated-stack config still skip with `skipped_unsafe_shared_stack`.

## What to look at

- **E2E command orchestration**: [`e2e.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/d89d544cb534b7c43a9f768bca86af9b527f707a/src/cli/commands/e2e.ts) adds the cold-stack CLI flags, exposes isolated-stack capability during remote resolution, skips preflight against the original Cloud URL for chains that will be cold-provisioned, and registers active cold-stack cleanup for signals.
- **Cloud target provisioning**: [`cloud-provisioning.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/d89d544cb534b7c43a9f768bca86af9b527f707a/src/cli/e2e/cloud-provisioning.ts) decides whether a chain needs a cold stack, maps all Cloud guides in that chain to the provisioned target URL/token, and preserves the shared-stack service-account path for read-only/shared-auth runs.
- **Result reporting and docs**: [`e2e-console-reporter.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/d89d544cb534b7c43a9f768bca86af9b527f707a/src/cli/e2e/e2e-console-reporter.ts) and [`E2E_TESTING.md`](https://github.com/grafana/grafana-pathfinder-app/blob/d89d544cb534b7c43a9f768bca86af9b527f707a/docs/developer/E2E_TESTING.md) surface best-effort cleanup warnings and document the cold-stack operator flags and cleanup limits.

## Why

Phase 1 made unsafe Cloud guide chains skip rather than pollute shared stacks. This change wires the cold-stack lifecycle into the runner so those chains can execute against disposable Grafana Cloud stacks while preserving the fast shared-stack path for read-only or instance-specific guides.

## How to verify

1. Run an unsafe Cloud package with `--tier cloud`, `--cloud-stack-access-policy-token <envVar>`, and `--cloud-stack-region <region>`; confirm the runner provisions an ephemeral Cloud stack, runs Playwright against the generated stack URL, and attempts teardown after the chain.
2. Run the same unsafe Cloud package without cold-stack config; confirm it reports `skipped_unsafe_shared_stack` instead of running against the shared target.
3. Run a read-only Cloud package with matching `--cloud-instance-admin-token host=ENV_VAR_NAME`; confirm it continues to use the shared-stack service-account path.
4. Start a cold-stack run and interrupt it with Ctrl-C; confirm the CLI attempts best-effort teardown for any active cold-provisioned stack before exiting.

## Breaking changes

## Reversibility

Reverting this change disables cold-stack routing and returns unsafe Cloud chains to the existing skip behavior. Cloud stacks created while using the new flags are external side effects; normal completion, setup failure, guide failure, Ctrl-C, and SIGTERM attempt cleanup, but a process killed before signal handling or Terraform teardown may leave a labeled stack that requires manual cleanup.

## Out of scope

- Hot-pool discovery, leasing, and replacement.
- Durable stale-stack reconciliation after crashes.
- Installing manifest-declared plugins beyond `grafana-pathfinder-app`.

Refs #1184
Refs #1187

Closes #1187

### DIFF
--- a/src/cli/commands/e2e.ts
+++ b/src/cli/commands/e2e.ts
@@ -35,6 +35,7 @@ import {
   applyPackageMeta,
   buildPackageMetaMap,
   exitCodeFromResults,
+  provisioningFailureResults,
   resolveRunMode,
   skipToResult,
   type GuideRunResult,
@@ -249,22 +250,45 @@ interface ChainRunOutcome {
   cleanupWarnings: string[];
 }
 
+const REPEATED_SIGNAL_FORCE_EXIT_GRACE_MS = 30_000;
+
 /**
  * Install exit/signal handlers that tear down owned isolated environments.
  * On SIGINT/SIGTERM, re-raise with the conventional 128+signal code.
  */
 function installTeardownHandlers(cleanEnv: CleanEnvironment, cloudStackCleanup: ColdCloudStackCleanupRegistry): void {
-  let exiting = false;
+  let cleanupStartedAtMs: number | undefined;
   const exitHandler = () => cleanEnv.teardownIfOwned();
+  const exitCodeForSignal = (signal: NodeJS.Signals) => (signal === 'SIGINT' ? 130 : signal === 'SIGTERM' ? 143 : 1);
   const signalHandler = async (signal: NodeJS.Signals) => {
-    if (exiting) {
+    const code = exitCodeForSignal(signal);
+    if (cleanupStartedAtMs !== undefined) {
+      const elapsedMs = Date.now() - cleanupStartedAtMs;
+      if (elapsedMs >= REPEATED_SIGNAL_FORCE_EXIT_GRACE_MS) {
+        console.warn('\n⚠ Force exiting before cleanup completed; Cloud stacks may require manual teardown.');
+        process.exit(code);
+        return;
+      }
+      const remainingSeconds = Math.ceil((REPEATED_SIGNAL_FORCE_EXIT_GRACE_MS - elapsedMs) / 1000);
+      console.warn(
+        `\n⚠ Cleanup is still running; ignoring repeated ${signal}. Press again in ${remainingSeconds}s to force exit, which may leave Cloud stacks running.`
+      );
       return;
     }
-    exiting = true;
-    await cloudStackCleanup.teardownAll();
-    cleanEnv.teardownIfOwned();
-    const code = signal === 'SIGINT' ? 130 : signal === 'SIGTERM' ? 143 : 1;
-    process.exit(code);
+    cleanupStartedAtMs = Date.now();
+    try {
+      const warnings = await cloudStackCleanup.teardownAll();
+      for (const warning of warnings) {
+        console.warn(`   ⚠ ${warning}`);
+      }
+    } catch (err) {
+      console.warn(
+        `   ⚠ Failed to tear down active Cloud stacks: ${err instanceof Error ? err.message : 'Unknown error'}`
+      );
+    } finally {
+      cleanEnv.teardownIfOwned();
+      process.exit(code);
+    }
   };
   process.on('exit', exitHandler);
   process.on('SIGINT', signalHandler);
@@ -569,15 +593,24 @@ async function runChains(
       results.push(...unsafeSharedStackSkipResults(chain, packageMetaById, message));
       continue;
     }
-    const provisionedTargets = await provisionCloudTargetsForChain({
-      targetUrls: cloudTargetsInChain(chain, packageMetaById, cloudAuth),
-      cloudAuth,
-      chain,
-      packageMetaById,
-      cloudStack,
-      cloudStackCleanup,
-      verbose: options.verbose,
-    });
+    let provisionedTargets: Awaited<ReturnType<typeof provisionCloudTargetsForChain>>;
+    try {
+      provisionedTargets = await provisionCloudTargetsForChain({
+        targetUrls: cloudTargetsInChain(chain, packageMetaById, cloudAuth),
+        cloudAuth,
+        chain,
+        packageMetaById,
+        cloudStack,
+        cloudStackCleanup,
+        verbose: options.verbose,
+      });
+    } catch (err) {
+      const message = `Cloud target provisioning failed: ${err instanceof Error ? err.message : 'Unknown error'}`;
+      console.error(`\n❌ ${message}`);
+      results.push(...provisioningFailureResults(chain, packageMetaById, options.grafanaUrl, message));
+      allPassed = false;
+      continue;
+    }
     try {
       // IDs in this chain that failed or were skipped; their dependents skip.
       const blocked = new Set<string>();
--- a/src/cli/commands/e2e.ts
+++ b/src/cli/commands/e2e.ts
@@ -18,7 +18,7 @@ import {
   bundledRepositoryPath,
   type LoadedGuide,
 } from '../utils/file-loader';
-import { planGuideExecution } from '../e2e/guide-chains';
+import { planGuideExecution, type ExecutionPlan } from '../e2e/guide-chains';
 import type { TestResultsData } from '../e2e/e2e-reporter';
 import { checkTier, loadManifestFromDir, runManifestPreflight, type CurrentTier } from '../e2e/manifest-preflight';
 import {
@@ -55,6 +55,12 @@ import { randomUUID } from 'crypto';
 import { createCloudAuthPolicy, type CloudAuthPolicy } from '../e2e/cloud-auth';
 import { cloudTargetsInChain, provisionCloudTargetsForChain, sweepCloudTargets } from '../e2e/cloud-provisioning';
 import { unsafeCloudGuidesInChain, unsafeSharedStackMessage, unsafeSharedStackSkipResults } from '../e2e/cloud-routing';
+import {
+  ColdCloudStackCleanupRegistry,
+  createColdCloudStackProvisioningConfig,
+  type ColdCloudStackProvisioningConfig,
+} from '../e2e/cold-cloud-stack-environment';
+import { preflightTargetUrlsForPlan } from '../e2e/preflight-targets';
 
 /**
  * CLI options for the e2e command
@@ -83,6 +89,14 @@ interface E2ECommandOptions {
   cloudInstanceAdminToken: string[];
   /** Default cloud instance URL for cloud-tier guides without an `instance`. */
   cloudUrl: string;
+  /** Env var containing a Grafana Cloud Access Policy token for cold stack provisioning. */
+  cloudStackAccessPolicyToken?: string;
+  /** Grafana Cloud region slug for cold stack provisioning. */
+  cloudStackRegion?: string;
+  /** Slug prefix for cold-provisioned Grafana Cloud stacks. */
+  cloudStackSlugPrefix?: string;
+  /** Pathfinder plugin version to install on cold-provisioned stacks. */
+  cloudStackPluginVersion?: string;
 }
 
 function collectOption(value: string, previous: string[]): string[] {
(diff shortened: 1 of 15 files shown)


src/cli/e2e/cloud-provisioning.ts
```diff
@@ -31,13 +57,57 @@ export class ProvisionedCloudTargets {
...
+
+function lacksSharedAuth(meta: PackageMeta | undefined, cloudAuth: CloudAuthPolicy | undefined): boolean {
+  return isCloudGuide(meta) && !cloudAuth?.needsProvisioningFor(meta?.targetUrl);
+}
+
+function cloudGuideIds(chain: PlannedGuideRef[], packageMetaById: Map<string, PackageMeta>): string[] {
+  return chain.filter((planned) => isCloudGuide(packageMetaById.get(planned.id))).map((planned) => planned.id);
+}
+
+export function chainNeedsCloudStack(options: {
```
Jayclifford345, line 97: **Instance-pinned cloud guides are silently rerouted to a generic cold stack.**

`PackageMeta` has an `instance` field, and `resolveTarget` in `e2e-targets.ts` deliberately refuses the isolated-stack capability for instance-pinned guides ("runnable only with shared-stack auth for that exact instance"). But `chainNeedsCloudStack` and `cloudGuideIds` never look at `meta.instance`:

- A **mutating guide pinned to e.g. `play.grafana.org`** (with valid shared auth) now provisions a cold stack and runs against a fresh empty `pfe2e*` stack — its declared instance is ignored. Before this PR it was skipped with `skipped_unsafe_shared_stack`, a truthful signal.
- A **readonly instance-pinned guide** that merely shares a chain with an unsafe guide is likewise rerouted off its declared instance, where its preconditions (datasources, dashboards) likely don't exist.

This contradicts the resolution layer's own rule and the PR summary's "non-instance" framing. Suggest either excluding chains containing instance-pinned guides from cold routing (keep the skip), or explicitly deciding and documenting that cold routing overrides `instance`.

src/cli/commands/e2e.ts
```diff
@@ -582,6 +572,10 @@ async function runChains(
     const provisionedTargets = await provisionCloudTargetsForChain({
```
Jayclifford345: **A mid-run cold-stack provisioning failure aborts the entire run and discards completed results.**

This call sits outside the per-chain `try`, and nothing in `runChains` catches it. A Terraform failure on chain 3 of 5 (quota, transient API error — far likelier than the shared path's SA-minting failure) propagates to the command's `catch`, which prints `❌ Error`, exits with `CONFIGURATION_ERROR`, writes **no JSON report**, and throws away the earlier chains' results.

Consider converting a per-chain provisioning failure into failed/skipped results for that chain and continuing, so one flaky Terraform apply doesn't nuke the whole run's report.

src/cli/commands/e2e.ts
```diff
@@ -225,25 +241,28 @@ function resolveGuides(files: string[], options: E2ECommandOptions): LoadedGuide
...
-function installTeardownHandlers(cleanEnv: CleanEnvironment): void {
+function installTeardownHandlers(cleanEnv: CleanEnvironment, cloudStackCleanup: ColdCloudStackCleanupRegistry): void {
+  let exiting = false;
   const exitHandler = () => cleanEnv.teardownIfOwned();
   const signalHandler = async (signal: NodeJS.Signals) => {
+    if (exiting) {
+      return;
+    }
+    exiting = true;
+    await cloudStackCleanup.teardownAll();
```
Jayclifford345: **SIGINT during `terraform apply` races destroy against the aborting apply.**

On Ctrl-C the terminal delivers SIGINT to the whole process group, so an in-flight `terraform apply` child starts its own graceful abort while this handler concurrently runs `terraform destroy` in the same module dir. Destroy will hit the local state lock still held by the dying apply and fail immediately (Terraform doesn't wait without `-lock-timeout`), leaking the stack as a mere warning.

Passing `-lock-timeout=60s` (or similar) to destroy in `runTeardown` would make the interrupt path actually work; alternatively wait for the in-flight command to exit before destroying. (The destroy command lives in the base PR, but the concurrent-invocation path is introduced by this handler.)

src/cli/commands/e2e.ts
```diff
@@ -225,25 +241,28 @@ function resolveGuides(files: string[], options: E2ECommandOptions): LoadedGuide
...
- * Install exit/signal handlers that tear down the isolated --clean docker stack.
+ * Install exit/signal handlers that tear down owned isolated environments.
  * On SIGINT/SIGTERM, re-raise with the conventional 128+signal code.
  */
-function installTeardownHandlers(cleanEnv: CleanEnvironment): void {
+function installTeardownHandlers(cleanEnv: CleanEnvironment, cloudStackCleanup: ColdCloudStackCleanupRegistry): void {
+  let exiting = false;
   const exitHandler = () => cleanEnv.teardownIfOwned();
   const signalHandler = async (signal: NodeJS.Signals) => {
+    if (exiting) {
```
Jayclifford345: **Second Ctrl-C is a complete no-op.**

The `exiting` guard swallows all subsequent signals while `teardownAll()` runs. If a destroy hangs (network, cloud API), the user cannot interrupt at all short of `kill -9`. The usual convention: first signal → graceful teardown, second signal → force `process.exit` immediately.

src/cli/commands/e2e.ts
```diff
@@ -627,21 +625,17 @@ async function runChains(
...
         const runGuideOptions: RunGuideOptions = {
-          targetUrl: meta?.targetUrl ?? options.grafanaUrl,
+          targetUrl,
           verbose: options.verbose,
           trace: options.trace,
           headed: options.headed,
           artifacts: options.artifacts,
           alwaysScreenshot: options.alwaysScreenshot,
-          token: runnerAuth?.token,
+          token: isCloudTarget ? provisionedTargets.tokenForGuide(planned.id, meta?.targetUrl) : undefined,
```
Jayclifford345, line 671: **`runnerAuthFor` (and `RunnerAuth`) are now dead production code.**

Runner-token selection moved here to `provisionedTargets.tokenForGuide(...)`, leaving `cloudAuth.runnerAuthFor` with no production caller (test mocks only implement it because the interface demands it). Per the repo's refactor-hygiene convention — delete what you replace in the same PR — remove it from `CloudAuthPolicy` and the factory in `cloud-auth.ts`.

src/cli/e2e/cloud-provisioning.ts
```diff
@@ -56,10 +126,43 @@ export function cloudTargetsInChain(
 export async function provisionCloudTargetsForChain(options: {
```
Jayclifford345, line 127: **API clarity: `chain` and `packageMetaById` are optional but always passed.**

They're optional only to spare old test call sites; the one production caller always passes them, and their absence silently disables cold routing. Making them required would remove the `options.cloudStack!` non-null assertion pressure and the "half-configured" call shape.

Relatedly, `ProvisionedCloudTarget.cleanupTarget` always duplicates `env` — tracking `{env, cleanupRegistry}` and untracking `env` (typed as the teardown target) would drop a field.

src/cli/e2e/preflight-targets.ts
```diff
@@ -0,0 +1,79 @@
...
+    if (unsafeCloudGuidesInChain(chain, packageMetaById).length > 0) {
+      for (const planned of chain) {
+        ids.add(planned.id);
+      }
+    }
+  }
+  return ids;
+}
+
+function idsUsingColdCloudStack(
```
Jayclifford345, line 46: **Preflight gap for mixed-tier chains** (low severity — moot if chains are always tier-homogeneous).

`idsUsingColdCloudStack` excludes *every* guide in a cold chain from preflight, but only the chain's *cloud* guides get rerouted (`cloudGuideIds`); a local guide in that chain still runs against `--grafana-url`, which may then never be health-checked if no other guide references it. If mixed-tier chains are possible, exclude only the cloud guide ids.

Jayclifford345 (commented):
Review of the cold cloud stack routing wiring. Overall solid: routing is centralized in `chainNeedsCloudStack`, preflight exclusion mirrors the runtime decision through the same function, provisioning failures are internally torn down (no leak on setup failure), and the memoized `teardownPromise` makes the registry and per-chain teardown safely share one destroy. Tests cover the routing matrix well.

Findings are inline, most severe first:

1. Instance-pinned cloud guides silently rerouted to a generic cold stack (`chainNeedsCloudStack`)
2. Mid-run cold-stack provisioning failure aborts the whole run and discards completed results (`runChains`)
3. SIGINT during `terraform apply` races destroy against the aborting apply on the local state lock (signal handler)
4. Second Ctrl-C is a no-op (signal handler)
5. `runnerAuthFor`/`RunnerAuth` are now dead production code (anchored at the replacement site; `cloud-auth.ts` isn't in this diff)
6. `chain`/`packageMetaById` optional but always passed; `cleanupTarget` duplicates `env`
7. Preflight gap for mixed-tier chains (`idsUsingColdCloudStack`)

Nothing found in the security or performance categories: the access-policy token stays env-var-referenced, Terraform receives it via `TF_VAR_*` rather than generated HCL, and error output is redacted against known secrets.

Jayclifford345 (dismissed):

src/cli/e2e/preflight-targets.ts
leslie-heinzen: mixed-tier chains are not possible. 👍🏻

src/cli/commands/e2e.ts
leslie-heinzen: Going to deliberately not do what's suggested here. Think of the case where a user spams ctrl+c. The second exit results in cloud stack resources not being properly cleaned up.

Going to add a grace period and console warning instead.

src/cli/e2e/cloud-provisioning.ts
leslie-heinzen: Pushing back on this one for now because the solution is a bit overwrought, and I'd rather not add complexity unnecessarily.

Couple of assumptions that seem mostly true based on our guides:

* there should never be a chain of guides that have a mix of `instance` and non-instance. If there are dependents, they should have have the same instance declared.
* If a specific instance is declared, assume the guide is readonly; e.g. play.grafana.org and learn.grafana.net guides have `instance` but should not mutate.

Given the above points, I don't think we need to deal with the problem that is framed here at this time.

Jayclifford345 (approved):

src/cli/e2e/cloud-provisioning.ts
Jayclifford345: y thats very fair!

PR #1218, opened 2026-07-01
fix(auto-launch): buffer delivery via reusable latched channel so the guide always renders

Jayclifford345:
## Problem

Some sessions record the experiment open (`source: highlighted_guide_experiment`) but the guide never renders. It's a race in the auto-launch handshake:

1. The experiment fires → the sidebar **shell** mounts, records analytics, fires `pathfinder-sidebar-mounted`.
2. The orchestrator dispatches the one-shot `auto-launch-tutorial` DOM event **500 ms later**.
3. The listener (`useAutoLaunchTutorial`) lives in the **lazy-loaded** panel and only attaches after its chunk loads.

On slow chunk loads the listener isn't attached within 500 ms, so the one-shot event fires into the void — open recorded, guide never rendered. The `?doc=` deep-link path has the same race.

## Fix — one reusable latched channel (no copy-paste)

`src/lib/latched-broadcast.ts` — `createLatchedBroadcast<T>({ ttlMs })`:
- `emit` broadcasts to current subscribers; if none are subscribed, it **latches** the value.
- `subscribe` drains a fresh latched value once on attach, then receives future emits.
- A **TTL** (30 s) bounds the latch so a stale value can't re-open a guide on a much-later manual open (the "opens for no reason" class we've been fixing).

The `auto-launch-tutorial` mechanic now flows through a shared `autoLaunchChannel` (`src/global-state/auto-launch.ts`) instead of a one-shot DOM event: producers `emit`, `useAutoLaunchTutorial` subscribes. Delivery is **order-independent** — whether the listener attaches before or after the emit, the guide opens exactly once. The `?doc=` deep-link and navigate-action paths get the same reliability for free.

Only the final delivery was swapped — the surrounding coordination (`pathfinder-sidebar-mounted` / `-panel-mounted` / `pathfinder-auto-launch-pending`, empty-state suppression, analytics, the mount/timer logic) is unchanged.

## Why not route it through `pathfinder-suggest`?

`pathfinder-suggest` only populates the Featured/recommendations zone with clickable cards; the experiment needs the guide **rendered directly** as a tab. So instead of reusing that event, this extracts its *buffering* behavior into a reusable primitive and applies it to the direct-open path.

## Tests
- `latched-broadcast.test.ts` — live delivery, latch-within-TTL, dropped-after-TTL, broadcast-to-all, single late-subscriber drain, unsubscribe, last-write-wins.
- Migrated the 3 auto-launch tests to the channel, including a **race-fix test**: emit *before* the hook mounts → the guide still opens on subscribe.

`npm run check` green: typecheck, lint (0 errors), prettier, tests (311 suites / 4872), coverage thresholds met.

### DIFF
--- a/src/global-state/sidebar.ts
+++ b/src/global-state/sidebar.ts
@@ -3,6 +3,7 @@ import { BusEventWithPayload } from '@grafana/data';
 import { reportAppInteraction, UserInteraction } from '../lib/analytics';
 import pluginJson from '../plugin.json';
 import { panelModeManager } from './panel-mode';
+import { autoLaunchChannel } from './auto-launch';
 
 interface OpenExtensionSidebarPayload {
   pluginId: string;
@@ -111,16 +112,12 @@ class GlobalSidebarState {
       document.removeEventListener('pathfinder-panel-mounted', dispatch);
       // Small delay so docs-panel's useEffect listener is registered after mount
       setTimeout(() => {
-        document.dispatchEvent(
-          new CustomEvent('auto-launch-tutorial', {
-            detail: {
-              url: `bundled:${guideId}`,
-              title: guideId,
-              type: 'docs-page',
-              source: 'mcp_launch',
-            },
-          })
-        );
+        autoLaunchChannel.emit({
+          url: `bundled:${guideId}`,
+          title: guideId,
+          type: 'docs-page',
+          source: 'mcp_launch',
+        });
       }, 300);
     };
 
--- a/src/global-state/sidebar.ts
+++ b/src/global-state/sidebar.ts
@@ -69,7 +69,7 @@ class GlobalSidebarState {
   public openSidebar(componentTitle: string, props?: Record<string, unknown>): void {
     const mode = panelModeManager.getMode();
     // In floating mode, the panel is already mounted and listening for
-    // auto-launch-tutorial events. No sidebar open needed.
+    // auto-launch requests. No sidebar open needed.
     // In full-screen mode, the dedicated full-screen page owns the active
     // session; opening the sidebar would mount a *second*
     // CombinedLearningJourneyPanel that races on tabStorage. The user can
--- a/src/components/docs-panel/docs-panel.tsx
+++ b/src/components/docs-panel/docs-panel.tsx
@@ -31,7 +31,7 @@ import { isDevModeEnabled } from '../../utils/dev-mode';
 
 import { reportAppInteraction, UserInteraction, getContentTypeForAnalytics } from '../../lib/analytics';
 import { tabStorage, useUserStorage } from '../../lib/user-storage';
-import { useGuideProgressState, useAutoLaunchTutorial } from '../../hooks';
+import { useGuideProgressState, useAutoLaunchTutorial, type AutoLaunchTutorialDetail } from '../../hooks';
 import {
   fetchContent,
   getNextMilestoneUrlFromContent,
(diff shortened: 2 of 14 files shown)


moxious (dismissed):
Thanks for this — replacing the one-shot `auto-launch-tutorial` DOM event with the latched `autoLaunchChannel` is a clean fix for the render race, and it nicely reduces coupling (one typed channel instead of a string literal scattered across producers) and narrows the security surface. CI is green and the design is sound.

Requesting changes on a few cheap, in-cycle items — all small — that harden the load-bearing invariant here (**a latched value is drained by exactly one subscriber and cleared on delivery**) and keep the touched files honest. These are quick enough to fold into this PR rather than defer.

### 1. Memoize the consumer callbacks
The three consumers pass unstable inline `onIncoming` / `onLaunched` / `skipLaunch`, so `useAutoLaunchTutorial` unsubscribes and resubscribes on every render:
- `src/components/docs-panel/docs-panel.tsx:1169-1185`
- `src/components/floating-panel/FloatingPanelManager.tsx:150-154`
- `src/components/full-screen/FullScreenPanel.tsx:170-181`

This is safe *today* only because `subscribe()` clears the latch the instant it delivers — but if that invariant ever changes (e.g. to support multiple late surfaces), each of these silently becomes a double-open on the next render within the TTL. Wrapping the callbacks in `useCallback`/`useMemo` so the hook subscribes once per mount removes the fragility.

### 2. Add a single-drain regression test
Lock in the invariant #1 depends on: emit with no subscriber; subscribe + unsubscribe (drains the latch); subscribe again within TTL; assert the second handler is **not** called. The existing test at `src/lib/latched-broadcast.test.ts:62-73` covers two *distinct* late subscribers, not unsubscribe→resubscribe of the same consumer, so this property isn't currently pinned.

### 3. Trim the now-stale comments in the touched files
Several comments still describe the retired `auto-launch-tutorial` DOM event / "dispatch"; the mechanism is now the in-memory `autoLaunchChannel` (`emit`). Per trim-on-touch:
- `src/hooks/useAutoLaunchTutorial.ts:1-17` and `:52-57`
- `src/components/docs-panel/docs-panel.tsx:1160-1168`
- `src/components/full-screen/FullScreenPanel.tsx:167`
- `src/components/floating-panel/FloatingPanelManager.tsx:147`

(Keep the `auto-launch-complete` window-event reference — that one is still accurate.)

---

**Non-blocking note (no change required for this PR):** a value latched during a brief close-in-launch-window can be replayed to a later *manual* open within the 30s TTL, firing a spurious `OpenResourceClick(trigger_source: auto_launch_tutorial)`. Net telemetry coherence still improves over the bug being fixed here — just flagging it: if it shows up in the data, tagging the latch payload with an emit timestamp / `replayed` marker (or clearing the latch on explicit close) would let downstream distinguish a live auto-launch from a replay.

Nothing above is architectural — security, reversibility, and cross-cutting architecture all reviewed clean, and there's no tech debt introduced. These are just quick hardening + hygiene wins worth doing in-cycle. Thanks again!

Jayclifford345:
Thanks @moxious — all three items applied in b7132b2c.

**1. Memoized the consumer callbacks.** Wrapped `onIncoming` (and fullscreen's `skipLaunch`) in `useCallback` across the three surfaces so `useAutoLaunchTutorial` subscribes once per mount instead of resubscribing every render:
- `docs-panel.tsx` — hoisted the analytics `onIncoming` into a stable `handleAutoLaunchIncoming` (`useCallback([])`).
- `FloatingPanelManager.tsx` / `FullScreenPanel.tsx` — hoisted the `guideOpenInFlightRef` flips (and fullscreen's `skipLaunch`) into stable callbacks.

This removes the latent double-open you flagged — the safety no longer rests solely on `subscribe()` clearing the latch on delivery.

**2. Added the single-drain regression test.** `latched-broadcast.test.ts` → *"clears the latch on the first drain even when that subscriber then unsubscribes"*: emit with no subscriber → subscribe + unsubscribe (drains) → advance the clock but stay within TTL → resubscribe → assert the second handler is **not** called. That pins the drain-clears-latch invariant #1 depends on, which the existing two-distinct-subscribers test didn't cover.

**3. Trimmed the stale comments.** Fixed the ones you listed (hook header + `skipLaunch` doc, and the three components). I also extended the trim to the other files this PR touches, since they still narrated the retired DOM event and trim-on-touch applies to everything the PR modifies:
- `highlighted-guide-orchestrator.ts`, `pathfinder-deep-link-handler.ts`, `navigate-handler.ts`, `sidebar.ts` — *"dispatch `auto-launch-tutorial`"* → *"emit … via `autoLaunchChannel`"*, and dropped two dead references to a `module.tsx` `dispatchAutoLaunch` seam that no longer exists.
- Kept the `auto-launch-complete` window event and the ~500 ms deep-link references — both still accurate (the deep-link handler still does `setTimeout(() => autoLaunchChannel.emit(…), 500)`).
- Left the genuinely-untouched files (`panel-mode.ts`, `recovery/launch-sources.ts`, `pathfinder-search-params.ts`) alone per the no-sweep rule — happy to fold those in if you'd prefer them here.

**On the non-blocking note (replayed latch → spurious `auto_launch_tutorial` telemetry):** agreed, left as-is per your call. Noting it for whoever picks it up — clearing the latch on explicit close, or tagging the payload with an emit timestamp / `replayed` marker, would let downstream tell a live auto-launch from a replay. Can file a follow-up if it surfaces in the data.

While I was in here I also confirmed there are no orphaned DOM-event producers/listeners left — nothing still dispatches or listens for `auto-launch-tutorial`, so the channel migration is complete.

Verified locally: `typecheck`, `eslint` (0 errors), `prettier`, and the full frontend suite (**311 suites / 4874 passed**) all green.

src/global-state/sidebar.ts
```diff
@@ -111,16 +112,12 @@ class GlobalSidebarState {
...
-              source: 'mcp_launch',
-            },
-          })
-        );
+        autoLaunchChannel.emit({
+          url: `bundled:${guideId}`,
+          title: guideId,
+          type: 'docs-page',
+          source: 'mcp_launch',
+        });
```
leslie-heinzen, line 120: Several of these emits are inside scheduled setTimeouts of various intervals (300-500ms). Kinda minor, I think, but it's possible we end up emitting stale events as a result.

```
component mounts -> setTimeout schedules -> component unmounts -> this event is emitted after the time out elapses.
```

Maybe something like this to intercept:

```ts
setTimeout(() => {
  if (!sidebarState.getIsSidebarMounted()) {
    return;
  }
  autoLaunchChannel.emit({
    url: detail.url,
    title: detail.title,
    type: detail.type,
    source: 'highlighted_guide_experiment',
  });
}, 500);
```

Don't think it's super critical unless we see a lot of weird results but I think this is technically a vector. Would suggest not fixing at this point but maybe make a note about it.

leslie-heinzen (approved):
LGTM :shipit: 

I was also going to comment on the stale docs but it looks like you caught them 👍🏻

src/global-state/sidebar.ts
Jayclifford345: Good catch — addressed in #1219 (kept it a separate PR so it didn't expand this one). It guards each delayed emit on `getIsSidebarMounted()` at fire time across the three mount-driven sites (`sidebar.openWithGuide`, the highlighted-guide orchestrator, and the `?doc=` deep link), so a surface that unmounts during the 300–500 ms window no longer lands a stale value that the channel latches and replays. Regression tests cover the orchestrator + deep-link paths. Left `navigate-handler` alone since that emit is the user's explicit navigate-and-open intent. Thanks for flagging!

PR #1219, opened 2026-07-01
fix(auto-launch): skip stale channel emits after surface unmount

Jayclifford345:
## Problem

Follow-up to #1218. In review, @leslie-heinzen [flagged](https://github.com/grafana/grafana-pathfinder-app/pull/1218#discussion_r3507682078) that the auto-launch emits are scheduled inside 300–500 ms `setTimeout`s:

```
component mounts → setTimeout schedules → component unmounts → the event is emitted after the timeout elapses
```

Because `autoLaunchChannel` **latches** when no subscriber is present, a stale post-unmount emit isn't a harmless no-op — it lingers for up to the 30 s TTL and can **replay on a later manual open**, causing a guide to open unprompted plus a spurious `auto_launch_tutorial` analytics event.

(This is specific to the new latched channel — the old one-shot DOM event just fired into the void after unmount, which is why it only surfaces now.)

## Fix

Guard each delayed emit on `getIsSidebarMounted()` at **fire time** — the symmetric completion of the `getIsSidebarMounted()` check these sites already do at dispatch-*setup* time:

- `sidebar.ts` `openWithGuide` (`mcp_launch`, 300 ms)
- `highlighted-guide-orchestrator.ts` (`highlighted_guide_experiment`, 500 ms)
- `pathfinder-deep-link-handler.ts` (`?doc=`, 500 ms)

This does **not** defeat the latch's purpose: in the normal flow the sidebar/panel **shell** keeps the flag `true` while the lazy panel's subscriber is still loading, so the emit still latches and drains correctly. The guard only suppresses the emit when *no* surface is mounted at all.

`navigate-handler.ts` is intentionally left alone — there the emit is the user's explicit navigate-and-open intent, not a mount-driven auto-launch.

## Tests

- New regression tests in `highlighted-guide-orchestrator.test.ts` and `pathfinder-deep-link-handler.test.ts` for the mount → unmount → timer-fires → **no emit** path.
- Corrected the orchestrator emit-dispatch block's `getIsSidebarMounted` default from `false` to `true`, since those tests simulate a mounted surface (which sets the flag `true` in production, via `openSidebar`/mount effects).
- `openWithGuide` has no pre-existing test harness, so its (identical) guard isn't covered by a new test here.

Verified locally: typecheck, lint (0 errors), prettier, and the full frontend suite (**311 suites / 4890 passed**).

### DIFF
--- a/src/global-state/sidebar.ts
+++ b/src/global-state/sidebar.ts
@@ -112,6 +112,11 @@ class GlobalSidebarState {
       document.removeEventListener('pathfinder-panel-mounted', dispatch);
       // Small delay so docs-panel's useEffect listener is registered after mount
       setTimeout(() => {
+        // If the surface unmounted during the delay, skip the emit — the channel
+        // latches, so a stale value would replay on a later manual open.
+        if (!this.getIsSidebarMounted()) {
+          return;
+        }
         autoLaunchChannel.emit({
           url: `bundled:${guideId}`,
           title: guideId,
--- a/src/utils/experiments/highlighted-guide-orchestrator.test.ts
+++ b/src/utils/experiments/highlighted-guide-orchestrator.test.ts
@@ -244,7 +244,8 @@ describe('setupHighlightedGuideAutoOpen → auto-launch-tutorial dispatch', () =
     jest.runAllTimers();
     jest.clearAllMocks();
     mockIsExtensionSidebarOwnedByOther.mockReturnValue(false);
-    mockGetIsSidebarMounted.mockReturnValue(false);
+    // A surface is mounted when the delayed emit fires; the emit is now guarded on it.
+    mockGetIsSidebarMounted.mockReturnValue(true);
     mockFindDocPage.mockReturnValue({
       url: 'https://grafana.com/docs/onboarding',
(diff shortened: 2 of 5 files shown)


leslie-heinzen (approved):
LGTM :shipit: Let's get this fix out.

One comment/question for future consideration. Now that I'm looking more closely, the setTimeouts + getIsSidebarMounted guard looks like a bit of a code smell/timing workaround.

Since the channel already latches, the mount-event + fixed-delay bit seems mostly unnecessary. Could we drop the setTimeouts altogether?

Something like:
1. Drop the setTimeout and emit once, relying on the latch for late/lazy subscribers
2. Clear the latch when a surface unmounts (or shorten/scope the TTL) instead of guarding each emit site:
```ts
// src/global-state/sidebar.ts
public setIsSidebarMounted(isSidebarMounted: boolean): void {
  // ...
  if (!isSidebarMounted) {
    autoLaunchChannel.clearLatch(); // new channel clear method
  }
}
```
3. Extract `installAutoLaunchOnMount` helper so there's a single shared emit path to reason about

Food for thought for a future refactor.

PR #1224, opened 2026-07-02
feat(guide-reader): add document outline extraction hook

Jayclifford345:
## Summary

Adds a pure extraction layer for a future GitHub-style document outline in the two-tab guide reader: a slugify/uniqueSlug util for generating stable heading ids, and a `useDocumentOutline` hook that scans a rendered guide's DOM for headings and interactive-guide sections. No UI changes yet — this lands the logic the outline rail PR builds on.

## What to look at

- `src/hooks/useDocumentOutline.ts` — queries `h2, h3, h4, [data-interactive-section="true"]` in document order, assigns generated ids onto heading elements that don't already have one, and reads existing ids/section titles as-is.
- The dedup rule in the same file: when a plain heading is immediately followed by an interactive section sharing the same title text (the common `## Explore your data` + `<InteractiveSection title="Explore your data">` authoring pattern — see `src/bundled-interactives/first-dashboard/content.json`), the heading entry collapses into the section entry so the outline doesn't show the same title twice in a row.
- `src/utils/slug.ts` — GitHub-style slugify (diacritics stripped via a `\p{M}` Unicode property match, non-alphanumerics collapsed to hyphens) plus a `-1`/`-2` collision suffixer seeded from ids already present in the container.

## Why

The two-tab reader's planned outline navigation needs id generation before any rail UI can jump to headings, since most guide headings ship without DOM ids (`renderMarkdown()` doesn't generate them for JSON-guide markdown, and only some fetched HTML docs have author-provided ids). Splitting extraction into its own PR keeps the rail UI PR (next in this stack) reviewable as pure layout/interaction work.

## How to verify

No manual verification surface in this PR — it adds a hook and a util with unit test coverage (`slug.test.ts`, `useDocumentOutline.test.ts`) but isn't wired into any rendered component yet.

## Breaking changes

None.

## Out of scope

The outline rail UI, scroll-spy, and keyboard nav land in the next two PRs in this stack.

### DIFF
--- a/src/hooks/index.ts
+++ b/src/hooks/index.ts
@@ -6,3 +6,4 @@ export {
 } from './useAutoLaunchTutorial';
 export { useStepProgressFromEvents } from './useStepProgressFromEvents';
 export { usePersistedBoolean, usePersistedLocalState, usePersistedString } from './usePersistedLocalState';
+export { useDocumentOutline, type OutlineItem } from './useDocumentOutline';
--- a/src/hooks/useDocumentOutline.test.ts
+++ b/src/hooks/useDocumentOutline.test.ts
@@ -1,0 +1,125 @@
+import { renderHook } from '@testing-library/react';
+import { useDocumentOutline } from './useDocumentOutline';
+
+function containerWith(html: string): React.RefObject<HTMLDivElement> {
+  const div = document.createElement('div');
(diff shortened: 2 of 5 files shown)


Jayclifford345:
Consolidating this stack into a single PR: #1226 now contains all three commits.

PR #1225, opened 2026-07-02
feat(guide-reader): add outline rail UI to the guide reader

Jayclifford345:
## Summary

Adds `OutlineRail`, a sticky jump-to-heading nav rendered beside the two-tab guide reader's content column, and wires it into `GuideReaderOverlay` via the `useDocumentOutline` hook from the previous PR in this stack.

## What to look at

- `src/components/guide-reader/GuideReaderOverlay.tsx` — restructures the backdrop's single content column into a `layoutRow` flex row (content column + rail), adds a `contentReady` flag flipped by `ContentRenderer`'s `onContentReady` callback (reset via the "adjust state during render when a prop changes" pattern rather than a `useEffect` + `setState`, since this repo's `react-hooks/set-state-in-effect` lint rule flags the latter), and a `backdropRef` that the next PR's scroll-spy will use as the `IntersectionObserver` root.
- `src/components/guide-reader/OutlineRail.tsx` — click-to-jump: `jumpToHeading` does a container-scoped `querySelector` + smooth `scrollIntoView({ block: 'start' })` + the existing `fragment-highlight` pulse (mirrors `scrollToFragment` in `content-renderer.tsx` without touching that function, since the URL hash is reserved for controller pairing and can't be reused for heading anchors).
- `src/components/guide-reader/guide-reader.styles.ts` — the rail is `position: sticky` inside the backdrop's own scroll container, hidden below `1280px` (v1: hidden, not collapsible).
- This entire surface is dark: `GuideReaderOverlay` only mounts when the `enableTwoTabController` admin setting is on (default off), so there's no user-visible change without that opt-in.

## Why

The two-tab reader is the only Pathfinder surface with enough horizontal room (a full browser tab, 1100px content column) for an outline rail; the narrow sidebar docs panel is explicitly out of scope. Building it here means it ships behind the existing admin toggle with no new gating code.

## How to verify

1. This PR's tests (`OutlineRail.test.tsx`, updated `GuideReaderOverlay.test.tsx`) cover render/click/empty-state behavior with `scrollIntoView` mocked (jsdom doesn't implement it).
2. Manual verification of the rail rendering, jumping, and hiding at narrow widths in a real Grafana instance was not performed in this session — flagging for reviewer or a follow-up manual pass. This surface's E2E coverage is already deferred elsewhere pending the `enableTwoTabController` flag flip.

## Breaking changes

None.

## Out of scope

Scroll-spy highlighting, keyboard nav, and `aria-current` land in the next PR in this stack.

### DIFF
--- a/src/components/guide-reader/GuideReaderOverlay.tsx
+++ b/src/components/guide-reader/GuideReaderOverlay.tsx
@@ -14,9 +14,11 @@ import { InteractiveModeContext, type InteractiveMode } from '../../global-state
 import { ControllerChannelProvider, useControllerConnected } from '../../global-state/controller-channel';
 import { PathfinderFeatureProvider } from '../OpenFeatureProvider';
 import { testIds } from '../../constants/testIds';
+import { useDocumentOutline } from '../../hooks';
 import type { RawContent } from '../../types/content.types';
 import type { ControllerPairingLaunch } from '../../lib/pairing-manager';
 import { getGuideReaderStyles } from './guide-reader.styles';
+import { OutlineRail } from './OutlineRail';
 
 interface GuideReaderOverlayProps {
   doc: string;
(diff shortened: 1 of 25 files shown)


Jayclifford345:
Consolidating this stack into a single PR: #1226 now contains all three commits.

PR #1226, opened 2026-07-02
feat(guide-reader): add document outline navigation (jump-to-heading rail)

### LINKED ISSUE #1004: feat: External browser tab mode for custom guides

Jayclifford345:
## Summary

Adds a GitHub-style document outline (jump-to-heading rail) to the two-tab guide reader: heading/section extraction with generated anchor ids, a sticky rail with click-to-jump, scroll-spy highlighting of the current section, and keyboard/a11y support.

https://github.com/user-attachments/assets/89ef663c-4d31-42b9-8c7f-240245ee2b31

## What to look at

- `src/utils/slug.ts` + `src/hooks/useDocumentOutline.ts` — GitHub-style slugify (diacritics stripped, `-1`/`-2` collision suffixer) and a hook that queries `h2, h3, h4, [data-interactive-section="true"]` in document order, assigning generated ids onto headings that don't already have one. When a plain heading is immediately followed by an interactive section sharing the same title text (the common `## Explore your data` + `<InteractiveSection title="Explore your data">` authoring pattern — see `src/bundled-interactives/first-dashboard/content.json`), the heading entry collapses into the section entry so the outline doesn't show the same title twice in a row.
- `src/components/guide-reader/GuideReaderOverlay.tsx` — restructures the backdrop's single content column into a `layoutRow` flex row (content column + rail), adds a `contentReady` flag flipped by `ContentRenderer`'s `onContentReady` callback (reset via the "adjust state during render when a prop changes" pattern rather than a `useEffect` + `setState`, since this repo's `react-hooks/set-state-in-effect` lint rule flags the latter), and a `backdropRef` used as the scroll-spy's `IntersectionObserver` root.
- `src/components/guide-reader/OutlineRail.tsx` — click-to-jump (`jumpToHeading`: container-scoped `querySelector` + smooth `scrollIntoView({ block: 'start' })` + the existing `fragment-highlight` pulse, mirroring `scrollToFragment` in `content-renderer.tsx` without touching that function, since the URL hash is reserved for controller pairing); `aria-current` + active styling on the current item; roving `ArrowUp`/`ArrowDown` focus between the list's native `<button>`s.
- `src/hooks/useActiveOutlineItem.ts` — one `IntersectionObserver` per outline render, `rootMargin` carving out the top 25% of the viewport as the "active" band; on every callback it recomputes from live `getBoundingClientRect()` and picks the *last* item whose top has crossed the band (ties break to the later item). `notifyJump(id)` sets `activeId` immediately and suppresses spy recomputes for 1s so a clicked item stays highlighted through its own smooth-scroll animation instead of flickering mid-scroll.
- The rail is `position: sticky` inside the backdrop's own scroll container and hides below `1280px` (v1: hidden, not collapsible).
- This entire surface is dark: `GuideReaderOverlay` only mounts when the `enableTwoTabController` admin setting is on (default off), so there's no user-visible change without that opt-in.

## Why

The two-tab reader is the only Pathfinder surface with enough horizontal room (a full browser tab, 1100px content column) for an outline rail; the narrow sidebar docs panel is explicitly out of scope. Building it here means it ships behind the existing admin toggle with no new gating code. The scroll-spy design (top-quarter band, last-crossed-wins, suppress-during-jump) matches GitHub's file outline behavior: the highlighted section tracks scroll position, and clicking doesn't fight the spy while the page is still animating to the target.

## How to verify

1. Unit/component tests cover the extraction rules, click/empty-state rendering (`scrollIntoView` mocked — jsdom doesn't implement it), and the scroll-spy band-crossing/tie-break/suppression logic (`IntersectionObserver` mocked — jsdom has none too).
2. Manual verification in a real Grafana instance (rail rendering, click-jump, scroll-spy, keyboard nav, narrow-viewport hide) was not performed in this session — flagging for reviewer or a follow-up manual pass. This surface's E2E coverage is already deferred elsewhere pending the `enableTwoTabController` flag flip.

### DIFF
--- a/src/hooks/useActiveOutlineItem.ts
+++ b/src/hooks/useActiveOutlineItem.ts
@@ -30,6 +30,7 @@ export function useActiveOutlineItem(
       .filter((entry): entry is { item: OutlineItem; el: HTMLElement } => entry.el !== null);
 
     if (tracked.length === 0) {
+      setActiveId(null);
       return;
     }
 
@@ -37,8 +38,9 @@ export function useActiveOutlineItem(
       if (Date.now() < suppressedUntilRef.current) {
         return;
       }
-      const rootTop = root.getBoundingClientRect().top;
-      const bandHeight = root.getBoundingClientRect().height * TOP_BAND_FRACTION;
+      const rootRect = root.getBoundingClientRect();
+      const rootTop = rootRect.top;
+      const bandHeight = rootRect.height * TOP_BAND_FRACTION;
 
       // The active item is the last one whose top has scrolled above the
       // top-quarter band — iterating in document order means a later match
--- a/src/hooks/useActiveOutlineItem.ts
+++ b/src/hooks/useActiveOutlineItem.ts
@@ -1,0 +1,71 @@
+import { useCallback, useEffect, useRef, useState } from 'react';
+import type { OutlineItem } from './useDocumentOutline';
+
+const SUPPRESS_AFTER_JUMP_MS = 1000;
+const TOP_BAND_FRACTION = 0.25;
+
+export interface ActiveOutlineItem {
+  activeId: string | null;
+  notifyJump: (id: string) => void;
+}
+
+export function useActiveOutlineItem(
+  items: OutlineItem[],
+  containerRef: React.RefObject<HTMLDivElement>,
+  scrollRootRef: React.RefObject<HTMLDivElement>
+): ActiveOutlineItem {
+  const [activeId, setActiveId] = useState<string | null>(null);
+  const suppressedUntilRef = useRef(0);
+
+  useEffect(() => {
+    const container = containerRef.current;
+    const root = scrollRootRef.current;
+    if (!container || !root || items.length === 0) {
+      setActiveId(null);
+      return;
+    }
+
+    const tracked = items
+      .map((item) => ({ item, el: container.querySelector<HTMLElement>('#' + CSS.escape(item.id)) }))
+      .filter((entry): entry is { item: OutlineItem; el: HTMLElement } => entry.el !== null);
+
+    if (tracked.length === 0) {
+      return;
+    }
+
+    const recompute = () => {
+      if (Date.now() < suppressedUntilRef.current) {
+        return;
+      }
+      const rootTop = root.getBoundingClientRect().top;
+      const bandHeight = root.getBoundingClientRect().height * TOP_BAND_FRACTION;
+
+      // The active item is the last one whose top has scrolled above the
+      // top-quarter band — iterating in document order means a later match
+      // naturally overrides an earlier one, which is the tie-break we want.
+      let current = tracked[0]!.item;
+      for (const { item, el } of tracked) {
+        if (el.getBoundingClientRect().top - rootTop <= bandHeight) {
+          current = item;
+        }
+      }
+      setActiveId(current.id);
+    };
+
+    const observer = new IntersectionObserver(recompute, {
+      root,
+      rootMargin: `0% 0% -${(1 - TOP_BAND_FRACTION) * 100}% 0%`,
+    });
+    tracked.forEach(({ el }) => observer.observe(el));
+    recompute();
+
+    return () => observer.disconnect();
+  }, [items, containerRef, scrollRootRef]);
+
+  const notifyJump = useCallback((id: string) => {
+    suppressedUntilRef.current = Date.now() + SUPPRESS_AFTER_JUMP_MS;
+    setActiveId(id);
+  }, []);
+
+  return { activeId, notifyJump };
+}
--- a/src/hooks/useDocumentOutline.ts
+++ b/src/hooks/useDocumentOutline.ts
@@ -35,7 +35,7 @@ function extractOutline(container: HTMLElement): OutlineItem[] {
 
     let id = el.id;
     if (!id) {
-      id = uniqueSlug(slugify(text), taken);
+      id = uniqueSlug(slugify(text) || 'section', taken);
       el.id = id;
       taken.add(id);
     }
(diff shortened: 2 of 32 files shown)


src/locales/cs-CZ/grafana-pathfinder-app.json
```diff
@@ -81,5 +81,8 @@
       "simplifiedView": "This is a simplified view. You can replace this with Grafana's full HelpModal component.",
       "toggleViewMode": "Toggle view mode"
     }
+  },
+  "guideReader": {
```
moxious: What's the point of so many repeated i18n entries for different languages you don't populate?  Seems like wasted extra lines/complexity for no benefit, falling back to nothing.

src/hooks/useDocumentOutline.ts
```diff
@@ -0,0 +1,76 @@
...
+  for (const el of elements) {
+    const isSection = el.getAttribute('data-interactive-section') === 'true';
+    const text = (isSection ? el.querySelector(SECTION_TITLE_SELECTOR)?.textContent : el.textContent)?.trim() ?? '';
+    if (!text) {
+      continue;
+    }
+
+    let id = el.id;
+    if (!id) {
+      id = uniqueSlug(slugify(text), taken);
```
moxious: **(blocking)** `slugify` only strips Latin diacritics (NFKD), so any heading with no ASCII alphanumerics slugifies to `""`. A non-empty heading with an empty slug still gets pushed, yielding an empty `el.id`; `useActiveOutlineItem` then runs `querySelector('#' + CSS.escape(''))` → selector `'#'` → uncaught `SyntaxError`, which blanks the reader. `OutlineRail.jumpToHeading` throws the same way on click.

Suggested fix — guarantee a non-empty base slug:
```suggestion
      id = uniqueSlug(slugify(text) || 'section', taken);
```
Please add a regression test with a CJK/Cyrillic/emoji heading asserting non-empty ids and that the spy hook doesn't throw.

src/hooks/useActiveOutlineItem.ts
```diff
@@ -0,0 +1,71 @@
...
+    if (tracked.length === 0) {
+      return;
+    }
+
+    const recompute = () => {
+      if (Date.now() < suppressedUntilRef.current) {
+        return;
+      }
+      const rootTop = root.getBoundingClientRect().top;
+      const bandHeight = root.getBoundingClientRect().height * TOP_BAND_FRACTION;
```
moxious: **(nit)** Two forced-layout reads on the same element per callback. Read the rect once:
```suggestion
      const rootRect = root.getBoundingClientRect();
      const rootTop = rootRect.top;
      const bandHeight = rootRect.height * TOP_BAND_FRACTION;
```

src/hooks/useActiveOutlineItem.ts
```diff
@@ -0,0 +1,71 @@
...
+      return;
+    }
+
+    const tracked = items
+      .map((item) => ({ item, el: container.querySelector<HTMLElement>('#' + CSS.escape(item.id)) }))
+      .filter((entry): entry is { item: OutlineItem; el: HTMLElement } => entry.el !== null);
+
+    if (tracked.length === 0) {
+      return;
+    }
```
moxious, line 35: **(nit)** When `items` is non-empty but no ids resolve (`tracked.length === 0`), this returns without resetting `activeId`, unlike the `items.length === 0` guard above — a stale highlight could linger across a content swap. Suggest:
```suggestion
    if (tracked.length === 0) {
      setActiveId(null);
      return;
    }
```

moxious (changes requested):
Nice feature — clean architecture, DOM work kept in effects, `CSS.escape` used throughout, and good test coverage (31/31 passing locally). One correctness issue to fix before merge, plus two cheap cleanups worth folding in.

**Blocking:** headings with no ASCII alphanumerics (CJK, Cyrillic, Greek, Arabic, Hebrew, emoji) slugify to an empty string, produce an empty DOM id, and then crash the scroll-spy with an uncaught `SyntaxError: '' is not a valid selector`. Reproduced locally: Japanese/Cyrillic/emoji headings → ids `["", "-1"]` → throw in the active-item effect. `OutlineRail.jumpToHeading` throws the same way on click. One-line fallback fixes it; details inline.

src/hooks/useDocumentOutline.ts
Jayclifford345: Fixed in b605fd0d with exactly the suggested fallback (`uniqueSlug(slugify(text) || 'section', taken)`), plus the regression test you asked for: a CJK/Cyrillic/emoji-only heading case in `useDocumentOutline.test.ts` asserting non-empty ids, and an end-to-end test in `useActiveOutlineItem.test.ts` that runs real extraction output through the spy hook and asserts it doesn't throw.

src/hooks/useActiveOutlineItem.ts
Jayclifford345: Fixed in b605fd0d — reads `root.getBoundingClientRect()` once into `rootRect` and derives both `rootTop` and `bandHeight` from it.

src/hooks/useActiveOutlineItem.ts
Jayclifford345: Fixed in b605fd0d — `setActiveId(null)` now runs before the early return when `tracked.length === 0`, matching the `items.length === 0` guard above. Added a regression test covering a content swap to items whose ids don't resolve in the container.

src/locales/cs-CZ/grafana-pathfinder-app.json
Jayclifford345: Fair point — on reflection it's not worth a 20-locale diff for one untranslated aria-label. Removed the `guideReader.outlineLabel` key entirely in cc6ee15a (all locale files + the `t()` call in `OutlineRail.tsx`), hardcoded the aria-label to `"Document outline"` instead. Can revisit with a real i18n pass in a follow-up if this surface grows more user-facing text.

moxious (approved):

PR #1230, opened 2026-07-02
feat(e2e): lease isolated cloud stacks through pool manager

### LINKED ISSUE #1188: Lease and replace hot-pool Cloud stacks for E2E batches
Parent: #1184

## Problem

Cold-provisioning a stack before every unsafe cloud guide chain adds minutes of setup latency to batch runs.

## Scope

- Discover pool stacks using labels:
  - `pathfinder-e2e-pool=true`
  - optional `pathfinder-e2e-pool-id=<pool id>`
  - optional `pathfinder-e2e-state=available`
- Hydrate stack detail responses when the list API omits labels.
- Lease a pool stack once per CLI invocation.
- Mint a short-lived runner service-account token for the leased stack.
- Destroy the leased stack after use.
- Immediately create a replacement stack so the pool stays warm during batch runs.

## Acceptance criteria

- Unsafe cloud chains prefer a pool stack before cold provisioning.
- A used pool stack is destroyed and replaced automatically.
- Pool diagnostics explain why no stack was leased when the pool is exhausted.

### LINKED ISSUE #1189: Add durable pool coordination for concurrent Cloud E2E runs
Parent: #1184

## Problem

Stack labels and process-local tracking are enough for manual or single-process testing, but concurrent runs using the same pool can still race and lease the same stack.

## Scope

- Decide where durable lease state lives: object storage, a small service, or another shared store.
- Add atomic lease acquisition so concurrent jobs cannot pick the same stack.
- Add stale lease recovery for crashed jobs.
- Add scheduled capacity maintenance if immediate replacement is not enough.
- Keep destroy-and-replace semantics; do not reset and reuse dirty stacks until cleanup can be proven complete.

## Acceptance criteria

- Concurrent E2E jobs cannot lease the same stack.
- Expired or abandoned leases are recoverable.
- Pool capacity can be reconciled independently of individual test runs.

leslie-heinzen:
## Summary

This PR moves cloud guide-chain isolation from CLI-owned Grafana Cloud provisioning to the independently deployed pool manager service. Unsafe cloud chains, and cloud chains without shared-stack auth, now request a hot stack lease from the manager, run against the returned `grafanaUrl`/`runnerToken`, and retire the lease on cleanup.

## What to look at

- [Cloud stack pool manager client](https://github.com/grafana/grafana-pathfinder-app/blob/4fba0184573daa02beefce5a41b0633346ec21e8/src/cli/e2e/cloud-stack-pool-manager.ts): traces the new HTTP boundary for `POST /v1/leases` and `POST /v1/leases/{leaseId}/retire`, including config validation, token redaction, default `nightly` pool selection, and initial `requiredPlugins: []` behavior.
- [E2E command routing](https://github.com/grafana/grafana-pathfinder-app/blob/4fba0184573daa02beefce5a41b0633346ec21e8/src/cli/commands/e2e.ts): replaces Grafana Cloud Access Policy and Terraform flags with pool-manager URL/token flags, then advertises isolated-stack capability to remote package resolution only when manager config is present.
- [Cloud target provisioning](https://github.com/grafana/grafana-pathfinder-app/blob/4fba0184573daa02beefce5a41b0633346ec21e8/src/cli/e2e/cloud-provisioning.ts): keeps shared-stack service-account auth for safe chains, but sends unsafe/no-auth cloud chains through the manager and surfaces `no_capacity` as a provisioning failure instead of falling back to CLI-managed cold provisioning.
- [Cloud chain lifecycle contract](https://github.com/grafana/grafana-pathfinder-app/blob/4fba0184573daa02beefce5a41b0633346ec21e8/src/cli/e2e/cloud-chain-environment.ts): adds teardown outcome context so normal completion, failures, and interrupts can report meaningful lease retire metadata without coupling callers to the manager implementation.

## Why

The pool manager is now the component that owns privileged Grafana Cloud lifecycle operations, durable lease coordination, runner-token minting, stack cleanup, and eventual cold fallback. Keeping that logic in the CLI would continue to expose broad Cloud Access Policy credentials to runners and would duplicate the manager's lifecycle responsibilities.

This intentionally rejects the previous direct hot-pool/Terraform approach. For this phase, the CLI is only an unprivileged lease consumer; if no hot stack is available, the manager returns `no_capacity` and the affected chain fails rather than provisioning a cold stack locally.

## How to verify

1. Run the focused unit coverage for the pool-manager client and cloud target planning:
   ```bash
   npm run test:ci -- cloud-stack-pool-manager cloud-provisioning preflight-targets manifest-preflight
   ```
2. Verify the deployed pool-manager path against the `develop` pool. Set the bearer token in an environment variable first, then run a cloud guide or remote cloud batch that requires isolated-stack execution:
   ```bash
   export POOL_MANAGER_TOKEN=<pool-manager caller token>
   pathfinder-cli e2e --package find-transformations --tier cloud \
     --cloud-stack-pool-manager-url <pool-manager-url> \
     --cloud-stack-pool-manager-token POOL_MANAGER_TOKEN \
     --cloud-stack-pool-id develop \
     --output results.json
   ```
   Confirm the CLI leases a stack from `develop`, runs the affected chain against the returned `grafanaUrl` with the returned `runnerToken`, and retires the lease during teardown.
3. Repeat the deployed-manager flow when the `develop` pool has no available capacity, or with a caller token that is not authorized for that pool. Confirm the chain reports a pool-manager provisioning failure and does not fall back to a shared Cloud stack or the local Terraform provisioning path.
4. Start a manager-backed cloud chain with the deployed URL and interrupt it with `SIGINT`. Confirm cleanup still sends `POST /v1/leases/{leaseId}/retire` with a cancellation-style outcome and the CLI exits through the normal signal path.
5. Run a readonly cloud chain with matching `--cloud-instance-admin-token` and no pool-manager flags. Confirm it still uses the shared-stack per-chain service-account path.
## Breaking changes

The CLI no longer supports direct Grafana Cloud stack provisioning for guide-chain isolation. Callers must replace `--cloud-stack-access-policy-token`, `--cloud-stack-region`, `--cloud-stack-slug-prefix`, and `--cloud-stack-plugin-version` usage with pool-manager configuration (`--cloud-stack-pool-manager-url`, `--cloud-stack-pool-manager-token`, and optionally `--cloud-stack-pool-id`).

## Reversibility

This is reversible by reverting the CLI changes, but any manager-backed run can create external lease state while it is active. The CLI attempts to retire leases on normal completion and signal cleanup; if a process is killed before cleanup, recovery depends on the pool manager's lease expiry and reconciliation behavior.

## Out of scope

- Guide-manifest plugin requirements are not sent yet; lease requests intentionally use `requiredPlugins: []` until plugin install support is added in both the CLI metadata flow and the manager.
- Manager-side cold provisioning, queueing, and wait policies are not implemented here. The CLI treats `no_capacity` as a provisioning failure for now.
- The pool manager service implementation and deployment live outside this repository.

Closes #1188 
Closes #1189


src/cli/e2e/cloud-stack-pool-manager.ts
```diff
@@ -0,0 +1,408 @@
...
+    };
+
+    if (this.verbose) {
+      console.log(`   ☁️ Leased ${lease.source} Cloud stack ${lease.stackSlug} from pool ${lease.poolId}`);
+    }
+
+    return new CloudStackPoolManagerLease(lease, this.config, this.runId, chainId, this.verbose, this.fetchImpl);
+  }
+
+  private async api<T>(method: string, path: string, body: unknown): Promise<T> {
```
tomglenn: This `api()` method and `CloudStackPoolManagerLease.api()` below (lines 374-407) are near-identical: same fetch/timeout/header setup, same error-response unwrapping, same redaction pattern. Worth extracting into one shared helper (e.g. a `postJson(managerUrl, token, path, body, extraSecrets)`) so a future change to headers/timeout/error-shape doesn't need to be made twice.

src/cli/e2e/cloud-stack-pool-manager.ts
```diff
@@ -0,0 +1,408 @@
...
+  region?: unknown;
+  source?: unknown;
+  pluginVersion?: unknown;
+  waitMs?: unknown;
+  provisioningMs?: unknown;
+  diagnostics?: unknown;
+}
+
+interface RetireLeaseRequest {
+  outcome: string;
```
tomglenn: `outcome` is typed as plain `string` here even though `CloudChainRetireOutcome` already exists in `cloud-chain-environment.ts` and is used for `CloudChainTeardownContext.outcome`. Typing this as `CloudChainRetireOutcome` would give compile-time checking on the retire payload.

src/cli/e2e/cloud-chain-environment.ts
```diff
@@ -0,0 +1,28 @@
+export type CloudChainEnvironmentKind = 'shared' | 'pool';
+
+export type CloudChainRetireOutcome = 'passed' | 'failed' | 'provisioning_failed' | 'cancelled' | 'runner_error';
```
tomglenn: `'provisioning_failed'` is documented in `E2E_TESTING.md` ("the affected chain fails with `provisioning_failed`") but no call site ever actually sets this outcome — the `no_capacity` path throws before a lease exists (nothing to retire), and `runChains` only ever passes `'passed'`/`'failed'`/`'cancelled'`. Either wire a real retire call for the provisioning-failure case or drop this variant/doc claim so it doesn't look like a real code path.

src/cli/e2e/cloud-stack-pool-manager.ts
```diff
@@ -0,0 +1,408 @@
...
+    try {
+      const response = await this.api<RetireLeaseResponse>('POST', `/v1/leases/${this.lease.leaseId}/retire`, request);
+      if (this.verbose) {
+        const status = typeof response.status === 'string' ? response.status : 'retire_requested';
+        console.log(`   🧹 Retired Cloud stack lease ${this.lease.leaseId} (${status})`);
+      }
+      return parseCleanupWarnings(response.cleanupWarnings);
+    } catch (err) {
+      return [
+        `Failed to retire Cloud stack lease ${this.lease.leaseId}: ${redact(errorMessage(err), [
```
tomglenn: Minor: the error thrown by `this.api(...)` on line 358 is already redacted inside that `api()` call, so this `redact(...)` re-redacts an already-redacted string. Harmless no-op today, but worth collapsing once the two `api()` methods are deduplicated (see comment on line 275).

tomglenn (commented):
Automated review pass — a few low-severity cleanup items, nothing blocking. Nice test coverage on the new pool-manager client (redaction, no_capacity propagation, malformed-response validation).

tomglenn (dismissed):
Approved subject to cleanups documented above

tomglenn (approved):

PR #1235, opened 2026-07-03
Fix duplicate guide title rendering

Jayclifford345:
## Summary
- Bundled guides rendered their title twice: a synthetic `<h1>` from guide metadata, immediately followed by an identical leading heading in the guide's own markdown body.
- Removes the redundant heading from the 6 affected bundled guides (`first-dashboard`, `first-dashboard-cloud`, `welcome-to-grafana`, `welcome-to-grafana-cloud`, `e2e-framework-test`, `json-guide-demo`).
- Adds a `validateGuide()` suggestion warning (surfaced via the CLI `validate` command and the MCP authoring tool) so future guides get flagged before shipping.
- Adds a render-time structural dedup in `parseJsonGuide()` so the fix holds even for guides that ship without heeding the warning.

## Test plan
- [x] `npm run test:ci -- validate-guide json-parser-title-dedup` — 97 tests pass (14 new)
- [x] `npm run build:cli && npm run validate` — all 10 bundled guides valid, zero warnings
- [x] `npm run check` — full pre-merge gate passes (315 suites, 4939 tests)
- [x] Confirmed `ContentProcessor` (`content-renderer.tsx:559-564`) routes native-JSON guide content through `parseJsonGuide()`, so the fix sits on the real render path


moxious (changes requested):
Thanks for the root-cause fix on the bundled guides and the validation warning — both are the right direction. I'd like to propose a small scope adjustment before merging.

Keep: the direct content edits to the 6 bundled guides, and the headingDuplicatesTitle validation logic + CLI warning.

Remove: the runtime dedup in json-parser.ts (dedupeLeadingTitleHeading, dropLeadingHeading, findLeadingHeadingText, elementToPlainText, and the dedupeLeadingTitleHeading call in parseJsonGuide).

Rationale: The runtime strip is a silent fix — guide authors get no feedback, and there's no way to distinguish "the plugin quietly fixed this" from "everything is fine". More importantly, the prefix matching in headingDuplicatesTitle is bidirectional: if the title is shorter than the heading, a heading that adds information (e.g. # Create your first dashboard in Grafana Cloud when the title is Create your first dashboard) would be silently dropped. A false-positive validator warning is annoying; a false-positive runtime strip deletes rendered content with no trace. The bundled guides are already fixed directly, and CI runs validate on them — the runtime safety net isn't buying much.

Extending to interactive-tutorials: This pattern is exactly what the validator CLI should own across both repos. The check should stay at suggestion/INFO severity (not promoted to error in strict mode for this rule) so it doesn't block the interactive-tutorials build — the intent is guidance for authors, not a hard gate.

Jayclifford345:
Done in 84dad30 — removed the runtime dedup from `json-parser.ts` (`dedupeLeadingTitleHeading`, `dropLeadingHeading`, `findLeadingHeadingText`, `elementToPlainText`, and the call in `parseJsonGuide`) along with its test file. The direct guide content fixes and the `headingDuplicatesTitle` validator suggestion stay.

Also took your strict-mode point: the duplicate-heading suggestion is no longer promoted to an error in strict mode, so it stays authoring guidance and can't gate a build when the check extends to interactive-tutorials.

moxious (approved):
I have some reservations about what the right level / --strict behavior is on the CLI for finding duplicate titles, but it's not a reason to hold up this net improvement PR.  Once merged I'm actually going to try this out on `interactive-tutorials` and see what it does with the total build chain, decide from there.  We might adjust the error level in a follow-up.

PR #1241, opened 2026-07-07
chore: clear old dependencies and unblock faro-react

Jayclifford345:
## Summary

- Grafana Cloud now runs React 19 in prod; older self-hosted OSS Grafana is still on React 18. Plugins run on the host's shared/externalized React, so this is a forward/backward-compat pass, not a React version bump — `react`/`react-dom` stay pinned at 18.3.1 (matching `@grafana/ui`/`@grafana/scenes`'s own `react@^18.0.0` peer).
- Updates `@grafana/create-plugin` scaffolding 6.1.8 → 7.8.0 (webpack nested-plugin fix, `setupTests.d.ts` for jest-dom types, `tsconfig` `baseUrl` → `paths` for TS6 compat, `ts-node` → `nodenext`/`es2022`, and a locked `@grafana/sign-plugin` devDependency instead of an inline `npx --yes ...@latest`). The scaffolding update's own migration list also includes a bundle-stats/bundle-size workflow permission hardening step, but it reported "no changes were made" here — this repo's workflow already matched, so no workflow files are touched by this PR.
- The `baseUrl` → `paths` change tripped `architecture.test.ts`'s path-alias tripwire (the import-graph tier-enforcement scanner only resolved relative specifiers, so `paths`-aliased bare imports could silently bypass tier enforcement). Extended `import-graph.ts` with `resolvePathAlias`/`loadTsconfigPaths` to resolve `tsconfig.paths` generically, with unit test coverage, and updated the tripwire to a `KNOWN_BASELINE_PATHS` allowlist mirroring the existing `KNOWN_BASELINE_BASE_URL` pattern.
- Bumps `@grafana/faro-react` 2.0.2 → 2.8.2 (`@grafana/faro-web-tracing` to match). `faro-react` 2.8.0+ widened its *optional* `react-router` peer to `^7.12.0 || ^8.0.0`; npm resolves that to `react-router@8.x`, which requires `react>=19.2.7` and conflicts with this plugin's `react@18.3.1`. Added a `react-router: ^7.12.0` override so the optional peer resolves to a React-18-compatible 7.x release instead.

### Update after review

- Dropped a stray blank line in `.config/webpack/webpack.config.ts`'s `ReplaceInFileWebpackPlugin` config that wasn't in the upstream create-plugin template.
[...]

PR #1242, opened 2026-07-07
chore(deps): dependency cleanup — safe bumps, overrides, pin ranges, js-yaml fix

Jayclifford345:
## Summary

Stacked on #1241 (needs the `react-router` override it introduces). First layer of a broader dependency-cleanup effort — see the followup plan for `@grafana/scenes`/`commander`/`typescript`/`eslint` bumps, which land as separate PRs off `main` once this one merges.

- **Safe patch/minor bumps** across ~28 dependencies/devDependencies (full list in commit `04f19fdf`). Explicit `pkg@version` installs, not a blanket `npm update`, so this can't accidentally pull in `@grafana/scenes` (still resolver-blocked, handled separately) or eslint's major.
  - Needed `--legacy-peer-deps` for this install: confirmed via `npm view <pkg> peerDependencies` diffs that several bumped packages (`@grafana/assistant` in particular) trip the same npm resolver false-positive as `@grafana/scenes` on the deep `@grafana/ui|data|runtime|i18n` peer-of-peer chain — not a real conflict. Verified `react`/`react-dom`/`react-router(-dom)`/`@grafana/ui|scenes|data|runtime` versions are unchanged after install.
  - `i18next-parser` 9.4.0 surfaces a new upstream deprecation notice ("use i18next-cli instead") — noted for a future follow-up, not actioned here.
- **Prettier 3.9.4 reformat**, isolated in its own commit so it's easy to skim/trust as purely mechanical (union-type line-breaking heuristic changed — this is why Renovate's own attempt at this bump, #1205, is failing its prettier-check).
- **Overrides cleanup** — verified each removal empirically (remove, reinstall, check resolution + `npm audit`) rather than trusting static analysis:
  - Removed `@grafana/ui.uuid` (no longer has a direct `uuid` dep) and `minimatch@9.brace-expansion` (minimatch 9.0.9 already natively requires `brace-expansion ^2.0.2`) — both confirmed redundant.
  - **Kept** `@grafana/scenes.uuid` — removing it was a real regression: `@grafana/scenes@8.2.6` resolves to the deprecated `uuid@9.0.1` without it, and `npm audit` went from 1 low to 3 (1 low, 2 moderate) vulnerabilities.
  - Also **investigated and rejected** dropping `react-router-dom` from `dependencies` + deleting the `react-router` override entirely (raised in review as a "cleaner" alternative to the override). Verified empirically: `@grafana/scenes` genuinely imports `react-router-dom` at runtime (`useUrlSync.js`, `SceneApp.js`, `SceneAppPage.js`, `utils.js`) via a live, reachable path (`module.tsx` → `App.tsx` → scenes' `SceneApp`). Removing it makes the production build hard-fail with 4 `Module not found: Can't resolve 'react-router-dom'` errors. Not a safe change here — reverted.
- **Pin→range conversion** for 12 packages with no documented pinning rationale (`@testing-library/*`, `identity-obj-proxy`, `prettier`, `typescript`, `sass`, `sass-loader`, `style-loader`, `@emotion/css`, `prismjs`, `rxjs`) — `renovate.json`'s `:preserveSemverRanges` re-pins exact forever once a package starts that way; this stops it from happening again. `npm install` afterward produced only a metadata-mirror diff in the lockfile (zero resolved-version changes), confirming it was purely syntactic. Deliberately **not** touched: `react`/`react-dom`, the `@grafana/data|runtime|ui|schema|i18n` family (exact-pinned on purpose for a `compatibilitycheck` CI job using `@grafana/levitate`), `@grafana/faro-react` (already range-ified in #1241).
- **`eslint --fix`** cleared 13 stale `eslint-disable`-directive warnings (dead directives whose underlying issue no longer applies). Left the 23 `@typescript-eslint/no-deprecated` warnings alone — real product-code migrations (`getDataSourceSrv`/`.get()`/`.getList()` etc.), different risk class, tracked as a separate follow-up.
- **js-yaml build-warning fix**: `GrotGuideBlockForm.tsx` imported js-yaml's default export, which doesn't exist in v5 (pure ESM, named exports only) — only compiled via a TS-only `esModuleInterop` shim webpack doesn't replicate, producing a build warning. Switched to `import { loadAll } from 'js-yaml'`; removed the now-dead `@types/js-yaml` devDependency.
- **Lockfile-only fix**: this repo's `packageManager` declares `npm@11.17.0`; a local npm client below that (11.6.1) silently drops entries from the lockfile's optional cross-platform dependency matrix (seen with `@unrs/resolver-binding-*`, a transitive dep of `jest-resolve`, and separately `@emnapi/*`). `npm ci` on the correct npm version (what CI runs) exposed the gap. Regenerated via `npx --yes npm@11.17.0 install --legacy-peer-deps`.

## Test plan
- [x] `npm run typecheck && npm run lint && npm run prettier-test` — clean (0 lint errors, 23 known pre-existing warnings)
- [x] `npm run test:ci` — 314 suites, 4937 tests passing
- [x] `npm run build && npm run build:backend` — clean (the js-yaml default-export warning is gone; only the pre-existing, out-of-scope asset-size-limit warning remains)
- [x] `npm audit` — 1 low-severity transitive advisory (`@babel/core`), unchanged from baseline; confirmed via `npm audit fix --dry-run` this specific one needs the `@grafana/scenes` bump (separate PR) to fully resolve
- [x] `npm run check` — full pre-merge gate, exit 0
- [x] `npx --yes npm@11.17.0 ci --legacy-peer-deps` — succeeds cleanly (matches CI's exact install path)
- [x] Manually verified `loadAll`'s behavior against a sample multi-document `---`-frontmatter YAML (the exact shape `convertYamlToBlock` parses) since there's no existing test coverage for that function


moxious (changes requested):
1. Stale comment at AssistantSelectionPopover.tsx:62 — references Promise.resolve().then() and a directive that was removed in this very PR. Should be deleted (trim-on-touch rule).
2. convertYamlToBlock (the one behavioral change) has zero test coverage. TypeCheck covers it, but an optional unit test would permanently close the gap.

Jayclifford345:
Rebased onto the latest #1241 (picked up its `resetCache`/tripwire-test fixes cleanly, no conflicts this time — the base commit only touched `.ts` source, not `package.json`/lock). CI is fully green after the rebase.

Also addressed the two findings from @moxious's review:

1. **Stale comment in `AssistantSelectionPopover.tsx`** — referenced `Promise.resolve().then()` and justified the `react-hooks/set-state-in-effect` eslint-disable directive right below it, but that directive was removed by this PR's `lint --fix` cleanup, and the `useLayoutEffect` body doesn't use `Promise.resolve().then()` at all. Deleted in 1a70e355.
2. **`convertYamlToBlock` had zero test coverage** — added `GrotGuideBlockForm.test.tsx` in 6e563732, following the sibling forms' established convention (render + `fireEvent` through the public UI, not exporting internal helpers). 5 tests: the happy path (multi-document `---` frontmatter parsing + snake_case field mapping, e.g. `screen_id`→`screenId`, `link_text`→`linkText`) and all four of the function's error branches.

Full suite: 315 test suites, 4947 tests passing.

moxious (approved):

PR #1275, opened 2026-07-08
feat(faro): re-enable Faro frontend telemetry (errors, sessions, logs, user actions, sourcemaps)

Jayclifford345:
## Summary

Re-enables Grafana Faro frontend telemetry for Pathfinder in Grafana Cloud — disabled since #389 pending a collector CORS fix — and builds out the full pipeline: errors, sessions, logs, analytics mirroring, resource timing, and view tracking. Everything is Cloud-only, sits behind a remote kill-switch (`pathfinder.frontend-telemetry`), and is filtered so only telemetry attributable to Pathfinder leaves the page. Faro initializes at plugin start (the pattern the Faro team recommends), but a `beforeSend` activity gate drops every item until Pathfinder is open in one of its surfaces — docked sidebar, floating panel, fullscreen, controller tab, or kiosk — read from localStorage and per-tab DOM roots, so sessions at the collector mean "used Pathfinder", not "loaded a Grafana page". Exceptions and error-level logs bypass the gate, so a session can also mean "Pathfinder broke". Real interactive work (steps, sections, guided steps, remote steps) runs inside real-duration user actions, so logs/errors/mirrored events emitted during a run are stamped with the action's `parentId`/`name` — causal correlation without tracing. The other historical blocker — OTel's fetch instrumentation dropping `response.url`, which the docs fetcher needs — is fixed upstream (otel-js #6343, in Faro ≥ 2.7.1).

Originally landed as a 5-PR stack (#1271–#1274 superseded this one); consolidated into a single PR per request.

## What to look at

- **Faro core** ([`src/lib/faro.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/d58f39e6ccb840e29825eccd3ecee5f81666b91e/src/lib/faro.ts)) — an isolated instance (`isolate: true`, own collector, own global object key) with a whitelist `beforeSend`: exceptions must carry a Pathfinder stack frame, logs must carry the `[pathfinder]` prefix, resource-timing entries must hit the docs/recommender/interactive-learning hosts, and page-wide navigation timing is always dropped. Sessions are **volatile (sessionStorage), not persistent** — this looks conservative but is load-bearing: Faro's persistent-session localStorage key is a fixed SDK constant that `isolate` does not namespace, and Grafana core's Faro writes the same key. Persistent sessions would resume core's session and inherit core's sampling decision, and could contaminate core's RUM sampling in return.
- **Activity-gated delivery** (`passesActivityGate` in `src/lib/faro.ts`) — init is eager per Faro team review (the previous engagement-gated init and its pre-init buffer are gone); send/drop is decided per transport item instead. Everything is dropped until Pathfinder is open in one of its surfaces. Open-state comes from localStorage — the panel-mode key (floating/fullscreen; closing the floating panel resets it to sidebar) and Grafana's `extensionSidebarDocked` key — plus per-tab DOM roots for the controller and kiosk overlays, which have no storage footprint. The gate latches for the rest of the page load once open, so the sidebar-close mirror (which fires after unmount, when the docked key is already gone) and teardown errors still send. Exceptions and error-level logs bypass the gate entirely — Pathfinder code runs with every panel closed (module init, link interception, cross-tab executor), and a session that exists only because Pathfinder broke is exactly what error telemetry is for.
- **Analytics mirroring** (`src/lib/analytics.ts` → `pushFaroUserAction`) — every `reportAppInteraction` event is mirrored into Faro as a User Action under the identical `pathfinder_*` name, so the RudderStack and Faro pipelines can be cross-checked. Mirrors carry an incrementing `seq` attribute because Faro's dedupe compares name + attributes but not timestamps; without it, identical rapid-fire events collapse and the counts drift.
- **Logging** ([`src/lib/logging.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/d58f39e6ccb840e29825eccd3ecee5f81666b91e/src/lib/logging.ts) + ~110 call-site files) — all ~420 `console.warn`/`console.error` call sites now route through the `logger` facade: console output is preserved (sanitized message + raw context object, so DevTools fidelity is intact) and a sanitized copy goes to Faro when it is active. A new `no-console` ESLint rule (log/info/debug still allowed) makes the migration provably complete and blocks regressions; exemptions are tests, `src/cli` (a Node CLI where console is the output), the facade itself, and three justified inline disables (tier-0 `constants.ts` sits inside the logger's own import chain; the two error boundaries already report via `pushFaroError`).
- **Real-duration user actions** (`withFaroUserAction` in `src/lib/faro.ts`, applied at five execution funnels: the interactive hook's two dispatchers, `handleDoSection`, `GuidedHandler.executeGuidedStep`, and the cross-tab `runStepCommand`) — opens a real action around actual work with an `outcome` attribute (ok/error/timeout) and a safety timeout. Two SDK gotchas are load-bearing here: `end(attributes)` is ignored by the 2.8.2 implementation (outcome is stamped by mutating the action's `attributes` field) and `end()` is not idempotent (a `settled` guard prevents duplicate action events). The mirror falls back to a plain skip-dedupe event while an action is open — previously those mirrors were silently dropped; now they parent into the action. Dashboards counting mirrors must union both shapes (`faro.user.action` events + plain events by name).
- **Views and resource timing** — the active docs tab sets Faro's view meta to the doc's hostname + pathname (via `useGlobalActiveTabExposure`), and PerformanceInstrumentation entries are kept only for the three domains the plugin actually fetches from, giving load metrics for docs/CDN/recommender requests.
- **Source maps** (`webpack.config.ts`) — `FaroSourceMapUploaderPlugin` is wired in but dormant: it only activates when all four `FARO_SOURCEMAP_*` env vars are present, and nothing in CI/CD sets them in this PR. The CI/CD wiring to actually drive it is deferred to a follow-up — see Known open items for why.
- **Test infrastructure** (`jest-setup.js`) — Prism is forced into manual mode: importing the logger pulls `@grafana/runtime` → `@grafana/ui` → `prismjs` into test module graphs, and Prism's document-wide auto-highlight rAF raced tests that mock `document.querySelectorAll`.

## Why

Pathfinder had no frontend observability: errors, guide failures, and docs-fetch problems in Cloud were invisible unless a user reported them. RudderStack measures product usage but cannot answer "is the plugin healthy", and there was no way to cross-check the analytics pipeline itself. This PR provides error/session/log/User-Action telemetry with per-signal attribution on a page shared with other Faro instances, remote controls to disable without shipping a release, and de-minified production stack traces.

Tracing was evaluated and deliberately excluded: a plugin's `TracingInstrumentation` collides with Grafana core's through OTel's page-global API — "attempted duplicate registration of API: trace" — which `isolate: true` cannot prevent (faro-web-sdk #1818, still open upstream). Our interesting fetches are cross-origin, so spans would not connect server-side anyway; client-only spans would largely duplicate the resource-timing events already collected here.

## How to verify

1. Dev-build QA: run `npm run dev`, set `localStorage['pathfinder.faro.local'] = 'true'`, reload, and open Pathfinder. The Network tab shows batched POSTs to the `faro-collector-ops-eu-south-0` collector; clicking around Pathfinder emits `faro.user.action` events named `pathfinder_*`, each with a `seq` attribute.
2. On a Cloud dev stack, confirm in Frontend Observability that exceptions carry Pathfinder stack frames only, logs are `[pathfinder]`-prefixed, and resource entries appear only for grafana.com / recommender / interactive-learning hosts.
3. Open two different docs tabs and switch between them — view-changed events fire and subsequent telemetry carries the doc's hostname + path as its view.
4. Flip `pathfinder.frontend-telemetry` to false in MTFF — no collector requests on next load.
5. Trigger a guide warning (e.g. run a step against a missing element) and confirm the DevTools console still shows the warning with an expandable context object.
6. ~~After the next main release, open a production error in Frontend Observability and confirm the stack trace is de-minified~~ — not yet possible; sourcemap upload has no CI/CD wiring in this PR (see Known open items).
7. Activity gating: on a dev stack, browse Grafana without touching Pathfinder — no collector traffic; dock the sidebar (or pop out the floating panel, or enter fullscreen) — telemetry begins, and it keeps flowing for the rest of the page load even after the panel closes.
8. Run "Do section" in a guide — one `pathfinder_section_run` user action with realistic duration, an `outcome` attribute, and the `DoSectionButtonClick` mirror parented inside it; individual "Do it" clicks appear as `pathfinder_step_do` actions.

## Breaking changes

None user-facing. Console output shape changes slightly at migrated call sites: `console.error('msg:', err)` becomes message + context object (`'msg', { error: err }`), with objects still expandable in DevTools.

## Reversibility

The telemetry stream can be stopped fleet-wide without a release via the kill-switch flag. Event and attribute names (`pathfinder_*` User Actions, the `[pathfinder]` log prefix, `seq`) become downstream query contracts once dashboards are built on them — rename with care. Data already collected persists in the ops stack.

## Known open items

- Collector CORS status (the original #389 blocker) needs reconfirming on a live Cloud stack before merge.
- **Sourcemap CI/CD wiring is deferred to a follow-up PR, blocked on an upstream fix.** `publish.yml`'s `cd:` job (the only build whose output is actually deployed) is a call into `grafana/plugin-ci-workflows`'s reusable `cd.yml` workflow — we don't own any steps inside that job, only its `frontend-secrets` input. Live-tested via a scoped CD dispatch: Vault successfully fetched all four `FARO_SOURCEMAP_*` secrets, but the actual `npm run build` step's env had none of them. Root cause, confirmed by reading `actions/internal/plugins/frontend/action.yml` at `ci-cd-workflows/v10.1.0` (the latest tag): its "Build" step has no `env:` block wiring the fetched secrets in at all — a plain oversight in their action, not intentional, and not something we can fix from this repo. Also confirmed separately: the Vault field for `FARO_SOURCEMAP_API_KEY` is named `FARO_SOURCEMAP_API_KEY`, not `value` (whoever revisits this will need that fix too). Wiring it into `ci.yml`'s push-to-main build instead was considered and rejected: that job's build artifact is archived for CI inspection only and never deployed, so even a working upload there would tag a bundle no real browser ever loads — just orphaned sourcemap sets for no benefit. A tag-based `release.yml` alternative was also rejected: its build runs inside a third-party composite action, so getting secrets into it would require exporting them via `$GITHUB_ENV`, visible to every later step in that job including any nested action in that chain.

## Out of scope

- Sourcemap CI/CD wiring — blocked upstream (see Known open items); `webpack.config.ts`'s `FaroSourceMapUploaderPlugin` is in place and dormant, ready for a follow-up PR once `grafana/plugin-ci-workflows` exposes fetched secrets to its build step's env.
- Developer docs — a telemetry pipeline doc plus `pathfinder.frontend-telemetry` entries in `FEATURE_FLAGS.md` follow in a separate PR.
- Tracing — blocked upstream (faro-web-sdk #1818); revisit when it lands, at which point stock `TracingInstrumentation` suffices (the `response.url` fix removes any need for a custom provider).
- `console.log`/`info`/`debug` call sites — intentionally local-only; only warn/error carry observability value.

### DIFF
--- a/src/interactive-engine/action-handlers/button-handler.ts
+++ b/src/interactive-engine/action-handlers/button-handler.ts
@@ -3,6 +3,7 @@ import { NavigationManager } from '../navigation-manager';
 import { InteractiveElementData } from '../../types/interactive.types';
 import { INTERACTIVE_CONFIG } from '../../constants/interactive-config';
 import { isElementVisible } from '../../lib/dom';
+import { logger } from '../../lib/logging';
 import { isCssSelector } from '../../lib/dom/selector-detector';
 import { resolveWithRetry } from '../../lib/dom/selector-retry';
 
@@ -63,7 +64,7 @@ export class ButtonHandler {
     for (const button of buttons) {
       // Validate visibility before interaction
       if (!isElementVisible(button)) {
-        console.warn('Target button is not visible:', button);
+        logger.warn('Target button is not visible', { button });
         // Continue anyway (non-breaking)
       }
 
@@ -81,7 +82,7 @@ export class ButtonHandler {
     for (const button of buttons) {
       // Validate visibility before interaction
       if (!isElementVisible(button)) {
-        console.warn('Target button is not visible:', button);
+        logger.warn('Target button is not visible', { button });
         // Continue anyway (non-breaking)
       }
 
--- a/src/interactive-engine/action-handlers/button-handler.ts
+++ b/src/interactive-engine/action-handlers/button-handler.ts
@@ -2,7 +2,7 @@ import { InteractiveStateManager } from '../interactive-state-manager';
 import { NavigationManager } from '../navigation-manager';
 import { InteractiveElementData } from '../../types/interactive.types';
 import { INTERACTIVE_CONFIG } from '../../constants/interactive-config';
-import { isElementVisible } from '../../lib/dom';
+import { describeElement, isElementVisible } from '../../lib/dom';
 import { logger } from '../../lib/logging';
 import { isCssSelector } from '../../lib/dom/selector-detector';
 import { resolveWithRetry } from '../../lib/dom/selector-retry';
@@ -64,7 +64,7 @@ export class ButtonHandler {
     for (const button of buttons) {
       // Validate visibility before interaction
       if (!isElementVisible(button)) {
-        logger.warn('Target button is not visible', { button });
+        logger.warn('Target button is not visible', { button: describeElement(button) });
         // Continue anyway (non-breaking)
       }
 
@@ -82,7 +82,7 @@ export class ButtonHandler {
     for (const button of buttons) {
       // Validate visibility before interaction
       if (!isElementVisible(button)) {
-        logger.warn('Target button is not visible', { button });
+        logger.warn('Target button is not visible', { button: describeElement(button) });
         // Continue anyway (non-breaking)
       }
 
--- a/src/lib/analytics.ts
+++ b/src/lib/analytics.ts
@@ -8,7 +8,7 @@
 import { reportInteraction } from '@grafana/runtime';
 import packageJson from '../../package.json';
 import { isInteractiveLearningUrl } from '../security/url-validator';
-import { pushFaroEvent } from './faro';
+import { pushFaroUserAction } from './faro';
 import type { ExperimentConfig, ExperimentAnalyticsEntry } from '../utils/openfeature';
 
 type GetActiveExperimentsFn = () => ExperimentAnalyticsEntry[];
@@ -203,10 +203,11 @@ export function reportAppInteraction(
     };
 
     reportInteraction(interactionName, enrichedProperties);
-    // Mirrors every analytics event into Faro (same name, same properties) so the
-    // two pipelines can be cross-checked against each other. pushFaroEvent never
-    // throws, so a mirror failure can't affect the reportInteraction call above.
-    pushFaroEvent(interactionName, enrichedProperties);
+    // Mirrors every analytics event into Faro as a User Action (same name, same
+    // properties) so the two pipelines can be cross-checked against each other.
+    // pushFaroUserAction never throws, so a mirror failure can't affect the
+    // reportInteraction call above.
+    pushFaroUserAction(interactionName, enrichedProperties);
   } catch (error) {
     console.warn('Analytics reporting failed:', error);
   }
--- a/src/lib/analytics.ts
+++ b/src/lib/analytics.ts
@@ -9,6 +9,7 @@ import { reportInteraction } from '@grafana/runtime';
 import packageJson from '../../package.json';
 import { isInteractiveLearningUrl } from '../security/url-validator';
 import { pushFaroUserAction } from './faro';
+import { logger } from './logging';
 import type { ExperimentConfig, ExperimentAnalyticsEntry } from '../utils/openfeature';
 
 type GetActiveExperimentsFn = () => ExperimentAnalyticsEntry[];
@@ -209,7 +210,7 @@ export function reportAppInteraction(
     // reportInteraction call above.
     pushFaroUserAction(interactionName, enrichedProperties);
   } catch (error) {
-    console.warn('Analytics reporting failed:', error);
+    logger.warn('Analytics reporting failed', { error });
   }
 }
 
--- a/src/lib/analytics.ts
+++ b/src/lib/analytics.ts
@@ -8,6 +8,7 @@
 import { reportInteraction } from '@grafana/runtime';
 import packageJson from '../../package.json';
 import { isInteractiveLearningUrl } from '../security/url-validator';
+import { pushFaroEvent } from './faro';
 import type { ExperimentConfig, ExperimentAnalyticsEntry } from '../utils/openfeature';
 
 type GetActiveExperimentsFn = () => ExperimentAnalyticsEntry[];
@@ -202,6 +203,10 @@ export function reportAppInteraction(
     };
 
     reportInteraction(interactionName, enrichedProperties);
+    // Mirrors every analytics event into Faro (same name, same properties) so the
+    // two pipelines can be cross-checked against each other. pushFaroEvent never
+    // throws, so a mirror failure can't affect the reportInteraction call above.
+    pushFaroEvent(interactionName, enrichedProperties);
   } catch (error) {
     console.warn('Analytics reporting failed:', error);
   }
--- a/src/lib/analytics.ts
+++ b/src/lib/analytics.ts
@@ -8,7 +8,7 @@
 import { reportInteraction } from '@grafana/runtime';
 import packageJson from '../../package.json';
 import { isInteractiveLearningUrl } from '../security/url-validator';
-import { pushFaroUserAction } from './faro';
+import { notifyFaroEngagement, pushFaroUserAction } from './faro';
 import { logger } from './logging';
 import type { ExperimentConfig, ExperimentAnalyticsEntry } from '../utils/openfeature';
 
@@ -204,6 +204,12 @@ export function reportAppInteraction(
     };
 
     reportInteraction(interactionName, enrichedProperties);
+    // Flag exposures fire for the whole experiment population on page load —
+    // they must not summon Faro (a session would then exist for every user).
+    // They ride the pre-init buffer until real engagement instead.
+    if (type !== UserInteraction.FeatureFlagEvaluated) {
+      void notifyFaroEngagement();
+    }
     // Mirrors every analytics event into Faro as a User Action (same name, same
     // properties) so the two pipelines can be cross-checked against each other.
     // pushFaroUserAction never throws, so a mirror failure can't affect the
--- a/src/lib/analytics.ts
+++ b/src/lib/analytics.ts
@@ -203,18 +203,22 @@ export function reportAppInteraction(
       ...(kioskSessionId && { kiosk_session_id: kioskSessionId }),
     };
 
-    reportInteraction(interactionName, enrichedProperties);
-    // Flag exposures fire for the whole experiment population on page load —
-    // they must not summon Faro (a session would then exist for every user).
-    // They ride the pre-init buffer until real engagement instead.
-    if (type !== UserInteraction.FeatureFlagEvaluated) {
-      void notifyFaroEngagement();
+    try {
+      reportInteraction(interactionName, enrichedProperties);
+    } finally {
+      // Flag exposures fire for the whole experiment population on page load —
+      // they must not summon Faro (a session would then exist for every user).
+      // They ride the pre-init buffer until real engagement instead.
+      if (type !== UserInteraction.FeatureFlagEvaluated) {
+        void notifyFaroEngagement();
+      }
+      // Mirrors every analytics event into Faro as a User Action (same name,
+      // copied properties) so the two pipelines can be cross-checked against
+      // each other. The finally keeps the mirror alive when reportInteraction
+      // itself throws — a lost RudderStack event is exactly the divergence the
+      // mirror exists to surface. pushFaroUserAction never throws.
+      pushFaroUserAction(interactionName, { ...enrichedProperties });
     }
-    // Mirrors every analytics event into Faro as a User Action (same name, same
-    // properties) so the two pipelines can be cross-checked against each other.
-    // pushFaroUserAction never throws, so a mirror failure can't affect the
-    // reportInteraction call above.
-    pushFaroUserAction(interactionName, enrichedProperties);
   } catch (error) {
     logger.warn('Analytics reporting failed', { error });
   }
--- a/src/lib/analytics.ts
+++ b/src/lib/analytics.ts
@@ -8,7 +8,7 @@
 import { reportInteraction } from '@grafana/runtime';
 import packageJson from '../../package.json';
 import { isInteractiveLearningUrl } from '../security/url-validator';
-import { notifyFaroEngagement, pushFaroUserAction } from './faro';
+import { pushFaroUserAction } from './faro';
 import { logger } from './logging';
 import type { ExperimentConfig, ExperimentAnalyticsEntry } from '../utils/openfeature';
 
@@ -206,12 +206,6 @@ export function reportAppInteraction(
     try {
       reportInteraction(interactionName, enrichedProperties);
     } finally {
-      // Flag exposures fire for the whole experiment population on page load —
-      // they must not summon Faro (a session would then exist for every user).
-      // They ride the pre-init buffer until real engagement instead.
-      if (type !== UserInteraction.FeatureFlagEvaluated) {
-        void notifyFaroEngagement();
-      }
       // Mirrors every analytics event into Faro as a User Action (same name,
       // copied properties) so the two pipelines can be cross-checked against
       // each other. The finally keeps the mirror alive when reportInteraction
--- a/src/lib/faro.ts
+++ b/src/lib/faro.ts
@@ -1,4 +1,12 @@
-import type { ExceptionEvent, Faro, LogEvent, LogLevel, TransportItem, APIEvent } from '@grafana/faro-web-sdk';
+import type {
+  ExceptionEvent,
+  Faro,
+  LogEvent,
+  LogLevel,
+  TransportItem,
+  APIEvent,
+  UserActionInternalInterface,
+} from '@grafana/faro-web-sdk';
 import { config } from '@grafana/runtime';
 import packageJson from '../../package.json';
 
@@ -164,16 +172,14 @@ export function stringifyAttributes(attributes: Record<string, unknown>): Record
   return result;
 }
 
-export function pushFaroEvent(name: string, attributes?: Record<string, unknown>): void {
+// Faro's public startUserAction() return type only exposes `name`/`parentId` —
+// ending it requires the internal interface. This isn't a hack: Faro's own
+// built-in click instrumentation (UserActionController) does the exact same
+// cast internally, since UserActionsAPI has no top-level `endUserAction()`.
+export function pushFaroUserAction(name: string, attributes?: Record<string, unknown>): void {
   try {
-    // Faro's pushEvent dedupes consecutive identical (name + attributes)
-    // pushes by default. That's wrong for an analytics mirror meant for
-    // exact cross-checking against Rudderstack, which doesn't dedupe — a
-    // legitimate repeated interaction (e.g. retrying the same button) would
-    // otherwise silently vanish from Faro's side only.
-    faroInstance?.api.pushEvent(name, attributes ? stringifyAttributes(attributes) : undefined, undefined, {
-      skipDedupe: true,
-    });
+    const action = faroInstance?.api.startUserAction(name, attributes ? stringifyAttributes(attributes) : undefined);
+    (action as UserActionInternalInterface | undefined)?.end();
   } catch {
     // Telemetry must never break the app it's observing.
   }
--- a/src/lib/faro.ts
+++ b/src/lib/faro.ts
@@ -142,6 +142,32 @@ export function pushFaroLog(level: FaroLogLevel, message: string, context?: Reco
   }
 }
 
+const MAX_ATTRIBUTE_LENGTH = 500;
+
+// Faro event attributes are string-only; analytics properties are not
+// (numbers, booleans, the experiments array). Stringify here so every
+// caller of pushFaroEvent gets the same coercion.
+export function stringifyAttributes(attributes: Record<string, unknown>): Record<string, string> {
+  const result: Record<string, string> = {};
+  for (const [key, value] of Object.entries(attributes)) {
+    if (value === null || value === undefined) {
+      continue;
+    }
+    const stringified =
+      typeof value === 'string' ? value : typeof value === 'object' ? JSON.stringify(value) : String(value);
+    result[key] = stringified.slice(0, MAX_ATTRIBUTE_LENGTH);
+  }
+  return result;
+}
+
+export function pushFaroEvent(name: string, attributes?: Record<string, unknown>): void {
+  try {
+    faroInstance?.api.pushEvent(name, attributes ? stringifyAttributes(attributes) : undefined);
+  } catch {
+    // Telemetry must never break the app it's observing.
+  }
+}
+
 export function pauseFaroBeforeReload(): void {
   try {
     faroInstance?.pause();
--- a/src/lib/faro.ts
+++ b/src/lib/faro.ts
@@ -1,4 +1,5 @@
 import type {
+  EventEvent,
   ExceptionEvent,
   Faro,
   LogEvent,
@@ -9,6 +10,11 @@ import type {
 } from '@grafana/faro-web-sdk';
 import { config } from '@grafana/runtime';
 import packageJson from '../../package.json';
+import {
+  ALLOWED_GRAFANA_DOCS_HOSTNAMES,
+  ALLOWED_INTERACTIVE_LEARNING_HOSTNAMES,
+  ALLOWED_RECOMMENDER_DOMAINS,
+} from '../constants';
 
 const COLLECTOR_URL = 'https://faro-collector-ops-eu-south-0.grafana-ops.net/collect/d6ec87b657b65de6e363de05623d9c57';
 const APP_NAME = packageJson.name;
@@ -68,10 +74,39 @@ function isLogItem(item: TransportItem<APIEvent>): item is TransportItem<LogEven
   return item.type === 'log';
 }
 
+function isEventItem(item: TransportItem<APIEvent>): item is TransportItem<EventEvent> {
+  return item.type === 'event';
+}
+
+// PerformanceInstrumentation reports every fetch/xhr resource load on the
+// page (Grafana core's own requests included) — these two are the only
+// domains this plugin itself ever fetches from, so anything else is core's
+// traffic, not ours.
+const TRACKED_RESOURCE_HOSTNAMES = new Set([
+  ...ALLOWED_GRAFANA_DOCS_HOSTNAMES,
+  ...ALLOWED_RECOMMENDER_DOMAINS,
+  ...ALLOWED_INTERACTIVE_LEARNING_HOSTNAMES,
+]);
+
+function isTrackedResourceUrl(resourceUrl: string | undefined): boolean {
+  if (typeof resourceUrl !== 'string') {
+    return false;
+  }
+  try {
+    return TRACKED_RESOURCE_HOSTNAMES.has(new URL(resourceUrl).hostname);
+  } catch {
+    return false;
+  }
+}
+
 // Whitelist, not blocklist: Grafana core and other app plugins run their own
 // Faro instances on the same page, so exceptions/logs must be attributable to
-// Pathfinder. Other item types (events, measurements) only ever originate
-// from this isolated instance's own API, so they pass through unfiltered.
+// Pathfinder, and PerformanceInstrumentation's resource entries must be
+// attributable to a domain we actually fetch from. Page-wide navigation
+// timing (the whole page's load, not a specific resource) is always dropped.
+// Everything else — our own pushed events/user-actions/measurements — only
+// ever originates from this isolated instance's own API, so it passes
+// through unfiltered.
 export function filterPathfinderTelemetry(item: TransportItem<APIEvent>): TransportItem<APIEvent> | null {
   if (isExceptionItem(item)) {
     const frames = item.payload.stacktrace?.frames ?? [];
@@ -79,6 +114,14 @@ export function filterPathfinderTelemetry(item: TransportItem<APIEvent>): Transp
   }
   if (isLogItem(item)) {
     return item.payload.message.startsWith(LOG_PREFIX) ? item : null;
+  }
+  if (isEventItem(item)) {
+    if (item.payload.name === 'faro.performance.navigation') {
+      return null;
+    }
+    if (item.payload.name === 'faro.performance.resource') {
+      return isTrackedResourceUrl(item.payload.attributes?.['name']) ? item : null;
+    }
   }
   return item;
 }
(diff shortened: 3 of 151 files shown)


src/lib/logging.ts
```diff
@@ -0,0 +1,39 @@
+import { sanitizeForLogging } from '../security/log-sanitizer';
+import { pushFaroError, pushFaroLog } from './faro';
+
+function sanitizeContext(context?: Record<string, unknown>): Record<string, string> | undefined {
```
moxious, line 4: **Serialization loses errors and can throw (must fix).** `sanitizeContext` runs every value through `sanitizeForLogging`, which does `JSON.stringify(value)` for non-strings. Two consequences:

- **`Error` → `"{}"`** (confirmed): `message`/`stack` are non-enumerable, so `JSON.stringify(new Error('boom')) === '{}'`. All ~240 `logger.error/warn(msg, { error })` sites send `context.error = "{}"` to Faro — the console still shows the real error, but the telemetry has zero detail.
- **Circular values throw**: `JSON.stringify` throws on a live React-managed DOM node (`__reactFiber$` circular). Since `sanitizeContext(context)` is evaluated *outside* `pushFaroLog`'s try/catch, the throw escapes `logger.warn` itself.

Fix: special-case `value instanceof Error` (serialize `name: message` + stack) and wrap the `JSON.stringify` in try/catch with a safe fallback. Add tests asserting an `Error` context doesn't become `{}` and a circular object doesn't throw.

src/lib/logging.ts
```diff
@@ -0,0 +1,39 @@
...
+    pushFaroLog('info', sanitized, sanitizeContext(context));
+  },
+
+  warn(message: string, context?: Record<string, unknown>): void {
+    const sanitized = sanitizeForLogging(message);
+    console.warn(sanitized, context ?? '');
+    pushFaroLog('warn', sanitized, sanitizeContext(context));
+  },
+
+  error(message: string, context?: Record<string, unknown>): void {
```
moxious: **Route real errors to `pushError`, not `pushLog` (must fix — this is the headline).** `logger.error` → `pushFaroLog('error', …)` produces a *log* signal. Per Faro/Frontend-Observability docs, the Errors/Exceptions view is populated by *exception* signals (`pushError` + `ErrorsInstrumentation`), not error-level logs. Right now ~118 sites use `logger.error` and only 1 uses `logger.exception`, so almost nothing lands in Error Tracking — no grouping, dedup, or symbolication.

Suggest an Error-aware overload: `error(message, errorOrContext?, context?)` that, when a throwable is present, calls `pushFaroError` (exception signal) as well as/instead of `pushLog`. Bonus: call sites become `logger.error('X', err)` instead of `logger.error('X', { error: err })` — shorter and correct.

src/interactive-engine/action-handlers/button-handler.ts
```diff
@@ -63,7 +64,7 @@ export class ButtonHandler {
     for (const button of buttons) {
       // Validate visibility before interaction
       if (!isElementVisible(button)) {
-        console.warn('Target button is not visible:', button);
+        logger.warn('Target button is not visible', { button });
```
moxious: **Passing a DOM element as context can throw in production (representative of ~8 sites).** `button` is a React-managed Grafana element; `{ button }` → `sanitizeContext` → `JSON.stringify(button)` throws `Converting circular structure to JSON` on the circular `__reactFiber$` props. jsdom has no fiber attached so `JSON.stringify` returns `"{}"` and the test suite is green — the throw only happens in the real app. Because this line is followed by "Continue anyway (non-breaking)," a throw here breaks a step that was explicitly meant not to.

Same pattern at focus-handler / form-fill / guided / hover-handler / navigation-manager (`{ element }`, `{ targetElement }`). Fixed centrally by the crash-proof serialization in `logging.ts`; also consider not putting raw nodes in context (e.g. `{ selector }` or `element.outerHTML.slice(0, N)`).

src/lib/faro.ts
```diff
@@ -1,152 +1,276 @@
...
+// Whitelist, not blocklist: Grafana core and other app plugins run their own
+// Faro instances on the same page, so exceptions/logs must be attributable to
+// Pathfinder, and PerformanceInstrumentation's resource entries must be
+// attributable to a domain we actually fetch from. Page-wide navigation
+// timing (the whole page's load, not a specific resource) is always dropped.
+// Everything else — our own pushed events/user-actions/measurements — only
+// ever originates from this isolated instance's own API, so it passes
+// through unfiltered.
+export function filterPathfinderTelemetry(item: TransportItem<APIEvent>): TransportItem<APIEvent> | null {
+  if (isExceptionItem(item)) {
```
moxious, line 141: **Fragile exception whitelist that fails silently to zero (must fix).** This drops any exception whose stack lacks a frame matching the literal `'grafana-pathfinder-app'`/`'/pathfinder/'`. Two problems:

- **No test pins the real production filename** — `faro.test.ts` uses hand-authored fixtures that contain the substring by construction. A future webpack/CDN/plugin-rename change makes the match fail and **every** exception is silently dropped (a `null` from `beforeSend` is a no-op, not an error), which reads as a healthy "zero errors" dashboard.
- **Explicitly-reported errors are filtered too**: the error boundary's `pushFaroError` and `logger.exception` are already attributed by originating in our code, yet a stackless error or a render error thrown in a shared `@grafana/*` chunk has no pathfinder frame and gets dropped — exactly the crash the boundary exists to report.

Suggest: let `pushFaroError`-originated exceptions bypass the frame filter, and add a post-build test that throws from a bundled function and asserts `isPathfinderStackFrame` matches the *actual* emitted filename.

src/lib/faro.ts
```diff
@@ -1,152 +1,276 @@
...
+      // (`com.grafana.faro.session`) that `isolate` does not namespace, and
+      // Grafana core's Faro uses it too — persistent:true would resume core's
+      // session, inherit its sampling decision (making samplingRate below a
+      // no-op), and could contaminate core's RUM sampling in return. Volatile
+      // sessions live in sessionStorage, which core doesn't touch.
+      persistent: false,
+      // A session not selected by the sample sends nothing for its entire
+      // lifetime. The local QA override always gets the real rate (1), not
+      // whatever the remote production-volume control happens to be set to.
+      samplingRate: isLocalOverride ? 1 : sampleRate,
```
moxious: **Clamp the sample rate (must fix).** `sampleRate` is forwarded straight into `samplingRate`. `values: [1]` on the flag is compile-time type metadata only — it does not clamp what MTFF returns, and `getNumberFlagValue` forwards `0`/`NaN`/`"1"` unchanged. A fat-fingered `0` (or a JSON-typed string) silences a session's entire telemetry for its lifetime, silently. Clamp to `(0, 1]` with a `logger.warn` on rejection.

src/utils/openfeature.ts
```diff
@@ -104,6 +105,32 @@ const pathfinderFeatureFlags = {
...
+  /**
+   * Fraction of sessions Faro actually sends telemetry for (Faro's own
+   * `sessionTracking.samplingRate`), remotely tunable without a release.
+   * A session not selected by the sample sends nothing — errors, events,
+   * logs — for its entire lifetime, not just a fraction of its signals.
+   * 1 = every session (default, current behavior unchanged); 0 = none.
+   */
+  'pathfinder.frontend-telemetry-sample-rate': {
+    valueType: 'number',
+    values: [1],
```
moxious: **`values: [1]` contradicts the flag's own intent.** The JSDoc above says `0 = none` and the PR sets 0/0.25 remotely, but `values: [1]` advertises `1` as the only legal value, and `FlagValue<T>` derives its type as `values[number]` → literal `1`. It doesn't clamp at runtime (consumed via `getNumberFlagValue`), so this is misleading metadata rather than a runtime bug. Use `values: [0, 0.25, 1]` or document as unconstrained. (See the clamp request in `faro.ts` for the runtime guard.)

src/lib/faro.ts
```diff
@@ -1,152 +1,276 @@
...
+  return result;
+}
+
+let userActionSeq = 0;
+
+// Faro's public startUserAction() return type only exposes `name`/`parentId` —
+// ending it requires the internal interface. This isn't a hack: Faro's own
+// built-in click instrumentation (UserActionController) does the exact same
+// cast internally, since UserActionsAPI has no top-level `endUserAction()`.
+export function pushFaroUserAction(name: string, attributes?: Record<string, unknown>): void {
```
moxious, line 314: **Prefer the documented `pushEvent` over the internal-interface cast.** The custom-signals docs model "track a user behavior" events with `pushEvent(name, attributes, domain)` and never document `startUserAction`/`endUserAction`. This synthetic zero-duration User Action relies on casting to `UserActionInternalInterface` because there's no public `endUserAction()` — which is also an upgrade hazard: a `@grafana/faro-web-sdk` bump (Renovate is active here) that renames `.end()` still compiles and silently kills the entire mirror stream (the `try/catch` swallows it, RudderStack keeps working, so nobody notices). `pushEvent` takes attributes directly and needs no cast.

src/lib/analytics.ts
```diff
@@ -202,8 +204,13 @@ export function reportAppInteraction(
     };
 
     reportInteraction(interactionName, enrichedProperties);
+    // Mirrors every analytics event into Faro as a User Action (same name, same
+    // properties) so the two pipelines can be cross-checked against each other.
+    // pushFaroUserAction never throws, so a mirror failure can't affect the
+    // reportInteraction call above.
+    pushFaroUserAction(interactionName, enrichedProperties);
```
moxious: **Mirror ordering + shared reference.** Two small things:

- `pushFaroUserAction` runs *after* `reportInteraction` in the same `try`, so if `reportInteraction` throws, the Faro mirror is skipped. The comment only reasons about the reverse direction. Reorder or give each its own try.
- `enrichedProperties` is passed by reference to both pipelines. It's safe today (`stringifyAttributes` copies, and RudderStack fires first), but a defensive copy — `pushFaroUserAction(interactionName, { ...enrichedProperties })` — makes the isolation explicit against future edits.

Also worth a code comment: the RudderStack↔Faro cross-check is only meaningful at `sampleRate == 1` with the kill-switch on — RudderStack has no sampling, so counts diverge multiplicatively at any lower rate.

src/lib/faro.ts
```diff
@@ -1,152 +1,276 @@
...
+function isLogItem(item: TransportItem<APIEvent>): item is TransportItem<LogEvent> {
+  return item.type === 'log';
+}
+
+function isEventItem(item: TransportItem<APIEvent>): item is TransportItem<EventEvent> {
+  return item.type === 'event';
+}
+
+// PerformanceInstrumentation reports every fetch/xhr resource load on the
+// page (Grafana core's own requests included) — these two are the only
```
moxious: **Stale comment + over-broad whitelist.** The comment says "these two are the only domains" but three hostname sets are spread into the Set below. More importantly, `ALLOWED_GRAFANA_DOCS_HOSTNAMES` contains the bare host `grafana.com`, and Grafana core itself fetches `grafana.com` (plugin catalog, news, usage-insights); `PerformanceInstrumentation` reports those and `isTrackedResourceUrl` matches the hostname exactly, so core's traffic is misattributed as Pathfinder resource telemetry — the opposite of the whitelist's intent. (Separately: the comment on `stringifyAttributes` references a nonexistent `pushFaroEvent` — the caller is `pushFaroUserAction`.)

moxious (changes requested):
Thanks for the thorough writeup — the SDK wiring, isolation reasoning, and the session-persistence decision are all sound, and this doesn't touch any existing RudderStack payload, so downstream analytics consumers are safe. I'm requesting changes because a set of findings converge on one conclusion: **as written, the error path won't deliver the error observability this PR exists to provide**, and two paths fail silently to zero. Priorities below; details inline.

## Must fix — the error path (one root, three symptoms)
1. **Most errors are logs, not exceptions.** ~118 `logger.error` sites route to `pushLog(level='error')`; only 1 routes to `pushError`. In Faro, the Errors/Exceptions view is fed by *exception* signals — error-level logs won't appear there (no grouping, dedup, or symbolication). Error Tracking will look nearly empty.
2. **The errors that are sent lose all detail.** `sanitizeForLogging` does `JSON.stringify(value)`, and `JSON.stringify(new Error()) === '{}'`. All ~240 `{ error }` context sites send `context.error = "{}"`.
3. **The same shape can throw in production.** 8 interactive-engine sites pass a live DOM element as context; `JSON.stringify` returns `"{}"` in jsdom (tests pass) but throws on React-managed elements in prod, aborting a path marked "non-breaking".

A single fix addresses all three: make serialization Error-aware and crash-proof, and route real throwables through `pushError`.

## Must fix — silent-to-zero paths
- **`beforeSend` filename coupling** (fragile whitelist, no test pins the real bundle path → a future build change silently drops every exception).
- **Sample-rate has no clamp** (`0`/`NaN` silences an entire session; `values: [1]` is type-only metadata and does not clamp).

## Should fix
- RudderStack↔Faro mirror ordering (a `reportInteraction` throw skips the mirror) + document that the cross-check is only valid at sampleRate=1.
- Prefer the documented `pushEvent` over the `UserActionInternalInterface` cast (upgrade hazard).
- `grafana.com` bare-host whitelist misattributes core traffic; two stale comments.

## Cleanup (optional, ~50 LOC)
Collapse the 5 `try{ faroInstance?.api.X }catch{}` wrappers into one `guardTelemetry(fn)` helper; make `logging.ts` info/warn/error table-driven; fold the three scalar flag getters into one generic.

## Verify (not code)
Confirm the publish workflow actually exposes the Vault secrets to the build step so sourcemaps upload (your own "known open items" flags this) — without them even genuine `pushError`s have unreadable stacks. Consider a follow-up for a collector-side volume-floor alert; nearly every silent-failure mode here manifests as "stream volume dropped and nothing told us."

src/module.tsx
```diff
@@ -58,12 +47,24 @@ const { createExperimentDebugger, initializeHighlightedGuideExperiment, setupHig
...
+// Import stays dynamic so the SDK never downloads when the flag is off.
+try {
+  if (getFeatureFlagValue('pathfinder.frontend-telemetry', true)) {
+    const sampleRate = getNumberFlagValue('pathfinder.frontend-telemetry-sample-rate', 1);
+    const { initFaro } = await import('./lib/faro');
+    await initFaro(sampleRate);
+  }
+} catch (e) {
+  logger.exception(e, { source: 'Faro init' });
+}
```
leslie-heinzen: Is this accurate? If the flag value is not explicitly set to false, we fall back to true, so this block is executed when no feature flag is present.

Answer to my question might be an obvious "yes, of course we want to fall back to initializing faro!" -- just wanted to flag it (no pun intended).

src/lib/logging.ts
Jayclifford345: Fixed in 75f81f7e. `sanitizeForLogging` now serializes `Error` as `name: message` + stack, and wraps `JSON.stringify` in a double fallback (`String()`, then `'[unserializable]'`), so circular values — React-managed DOM nodes included — can't throw. That also closes the escape you spotted where `sanitizeContext(context)` was evaluated as an argument, outside `pushFaroLog`'s try/catch. Tests: an `Error` context no longer becomes `"{}"`, a circular object doesn't throw, and a hostile circular null-prototype value falls back safely.

src/lib/logging.ts
Jayclifford345: Done in 75f81f7e. `logger.error(message, errorOrContext?, context?)` — a throwable as the second arg, or in the legacy `{ error }` context key, routes to `pushFaroError` (exception signal) with the message and remaining context attached, and skips the error-level `pushLog` so one failure isn't double-counted as both a log and an exception. All ~149 existing `logger.error('X', { error })` sites now produce exception signals with zero call-site changes; new code can write the shorter `logger.error('X', err)`.

One deliberate boundary: non-`Error` values in the `error` key stay logs (now properly serialized) — synthesizing an `Error` at log time would stamp every stack with `logging.ts` frames and hurt grouping.

src/interactive-engine/action-handlers/button-handler.ts
Jayclifford345: Fixed centrally by the crash-proof serialization in `log-sanitizer.ts`, and took the second suggestion too: added `describeElement()` (renders `tag#id.classes[data-testid="…"]`) and swapped all 8 sites — button/focus/form-fill/hover/guided handlers and navigation-manager — to log descriptors instead of raw nodes. The telemetry now says *which* element was invisible instead of `"{}"`. In 75f81f7e.

src/lib/faro.ts
Jayclifford345: Both parts in 75f81f7e:

1. **Explicit-report bypass**: `pushFaroError` stamps a `pathfinder_reported: 'true'` context marker, and `beforeSend` passes any exception carrying it — so error-boundary and `logger.exception` reports survive even when stackless or thrown entirely inside a shared `@grafana/*` chunk. Ambient `ErrorsInstrumentation` window errors still require a Pathfinder frame, so attribution isolation is intact. The marker doubles as an explicit-vs-ambient dimension in the data.
2. **No more literals**: the filter derives from `plugin.json` id (`/public/plugins/<id>/` is a Grafana platform invariant for served plugin assets) and `package.json` name (covers `webpack://` sourcemapped frames). The tests now construct their fixtures from `pluginJson.id` rather than hand-authored strings, so a plugin rename changes both sides together instead of silently zeroing the stream. Dropped the speculative `/pathfinder/` fallback — nothing is served from that path.

A true post-build throw-and-assert would need an e2e test; happy to add one as a follow-up if you want the extra pin.

src/lib/faro.ts
Jayclifford345: Resolved by deletion rather than a clamp: the sample-rate flag is removed entirely in 75f81f7e (see the openfeature.ts thread). A clamp can't distinguish a deliberate remote `0` from a fat-fingered one, and "off" already has a correct tool — the `pathfinder.frontend-telemetry` boolean kill-switch. `initFaro` no longer takes a rate at all (sessions sample at the SDK default of 1, volume-bounded by the engagement gating), so there's no unvalidated path left to guard.

src/utils/openfeature.ts
Jayclifford345: Flag deleted in 75f81f7e, along with `getNumberFlagValue` (it had no other consumers). Volume control is now: engagement gating (a session only starts on real Pathfinder use or an error) plus the `pathfinder.frontend-telemetry` kill-switch. One operational note: if a value was already set on this flag in MTFF, it's inert now and can be cleaned up server-side.

src/lib/faro.ts
Jayclifford345: Pushing back on this one, with history: 55dd84fc moved the mirror *from* `pushEvent` *to* User Actions deliberately — generic events have no dedicated page in Frontend Observability at all (confirmed by crawling the product nav at the time), and User Actions is the surface this mirror exists to populate. The cast isn't a hack we invented: Faro's own `UserActionController` does the identical cast internally, since `UserActionsAPI` has no top-level `endUserAction()`.

On the upgrade hazard: a *type-level* rename already fails `npm run check`, because we call `.end()` against the imported `UserActionInternalInterface` type. The residual risk is runtime-semantics drift with unchanged types — so 75f81f7e adds `faro-sdk-contract.test.ts`, which initializes the **installed** SDK (no mocks) against a capture transport and asserts `startUserAction(...).end()` actually emits a `faro.user.action` event. A Renovate bump that breaks the internal API now turns CI red instead of silently killing the mirror.

src/lib/analytics.ts
Jayclifford345: Done in 75f81f7e: `reportInteraction` is wrapped in try/finally with `notifyFaroEngagement` + the mirror in the `finally` — a RudderStack throw can't skip the mirror anymore, and Faro recording an event RudderStack lost is exactly the divergence the cross-check should surface (the outer catch still logs the failure). The mirror also gets a spread copy now, not the shared reference.

On the sampleRate==1 comment: moot — the sample-rate flag is removed in this same commit, so sampling is structurally 1 and the two pipelines stay count-comparable (modulo the engagement/Cloud gating, which was already documented).

src/lib/faro.ts
Jayclifford345: Fixed in 75f81f7e. Bare `grafana.com` now additionally requires a `/docs` or `/tutorials` path prefix — matching what `url-validator` permits this plugin to fetch — so core's plugin-catalog/news traffic no longer counts as ours. `docs.grafana.com`, `play.grafana.com`, and the recommender/CDN hosts stay hostname-matched (core doesn't fetch those). Added drop-tests for `grafana.com/api/...` and a `/docs-lookalike` path. Both stale comments rewritten (the counting one, and `stringifyAttributes` no longer references the nonexistent `pushFaroEvent`).

src/module.tsx
Jayclifford345: Yes — deliberate fail-open, matching `pathfinder.enabled`'s default-true convention: the flag is a kill-switch (flip to `false` to stop telemetry fleet-wide without a release), so a missing/unreachable flag means enabled. Good instinct to flag it though, and worth stating: the fallback can never enable telemetry outside Grafana Cloud, because `armFaro` → `resolveFaroEnvironment` independently gates on `isGrafanaCloud()` + `config.analytics.enabled` + the hostname allowlist — OSS/self-hosted never initializes regardless of the flag. Amended the comment in 75f81f7e to say this explicitly (default-on: a missing flag means enabled).

Jayclifford345:
All review threads addressed in 75f81f7e — details inline on each thread. The shape of the response, mapped to the review summary:

**Error path (the headline):** one root fix as suggested — crash-proof, Error-aware serialization in `sanitizeForLogging`, plus `logger.error` now routes throwables to `pushError` so the ~149 existing sites produce real exception signals. The 8 DOM-element sites log `describeElement()` descriptors instead of raw nodes.

**Silent-to-zero paths:** the `beforeSend` whitelist is derived from `plugin.json`/`package.json` (tests build fixtures from the same source), and explicitly-reported exceptions bypass the frame filter via a context marker. The sample-rate flag is **removed entirely** rather than clamped — a clamp can't tell a deliberate remote 0 from a fat-fingered one, and the boolean kill-switch is the correct off-switch. Every engaged session now samples at 1. (If a value was already set on the old flag in MTFF, it's inert and can be deleted server-side.)

**Should-fix:** mirror moved into a `finally` with a copied payload; `grafana.com` resource timing is path-constrained to `/docs`/`/tutorials`; stale comments rewritten. On the `UserActionInternalInterface` cast I've pushed back (see thread): `pushEvent` has no page in Frontend Observability, which is why 55dd84fc moved *to* User Actions — instead there's now an SDK-contract test that runs the installed SDK against a capture transport, so an SDK bump that breaks `end()` fails CI loudly.

**Cleanup:** took `guardTelemetry(fn)` (collapsed the 5 identical wrappers); skipped the table-driven logger (moot once `error()` diverges from `info`/`warn`) and the generic flag getter (the number getter is deleted anyway).

**Sourcemap secrets (verify item):** verified as far as the repo allows — `publish.yml` wires all four Vault secrets (main-only, correctly guarded against non-main publishes) and `webpack.config.ts` gates the uploader on all four env vars. What can't be confirmed locally is whether the shared plugins-CI workflow forwards `frontend-secrets` into the build step's env; needs a check of the uploader output in the next main publish log. Collector-side volume-floor alert noted as a follow-up — agreed that nearly every failure mode here manifests as "volume dropped silently."

moxious (changes requested):
## Blocking

### Assistant dev-mode prompt dumps now go to Faro logs

`src/integrations/assistant-integration/assistant-dev-mode.ts` now logs `prompt`, `context`, and `systemPrompt` through `logger.warn`. `logger.warn` forwards warnings to Faro via `src/lib/logging.ts` / `src/lib/faro.ts`, including buffered logs after engagement.

That changes these dev-mode dumps from local console output into remote telemetry. The sanitizer escapes/truncates strings, but it does not redact sensitive prompt/context data. Assistant dev mode is gated, but it is not local-build-only.

Suggested fix: keep these dumps local-only, for example with `logger.debug`, a non-Faro local logger, or a narrow `console.warn` lint exemption. Alternatively, redact/drop `prompt`, `context`, and `systemPrompt` before using Faro-backed logging.

## Medium

### First real-duration actions can be missed by engagement-gated init

`withFaroUserAction()` only starts a Faro user action when `faroInstance` already exists; otherwise it just runs `work()`. That means flows where the wrapper itself is the first interaction do not get the real-duration action envelope.

Concrete example: `pathfinder_section_run` starts before the section completion analytics event that triggers Faro engagement, so the first “Do section” in a fresh tab can run outside the real-duration Faro action. Remote step execution has a similar shape.

Suggested fix: make `withFaroUserAction()` initialize when Faro is armed before deciding to pass through, or trigger engagement before section/remote wrappers. Add tests proving a fresh armed tab captures the first `pathfinder_section_run` / `pathfinder_remote_step` action.

Jayclifford345:
Both items addressed in 3e8f6689.

**Assistant dev-mode prompt dumps (blocking):** all dev-mode dumps in `assistant-dev-mode.ts` moved from `logger.warn` to `logger.debug`, which is console-only (`console.log`) and never forwards to Faro — prompt/context/systemPrompt can no longer leave the browser, including via the pre-init buffer. The file's docstring now states the constraint, and a new `assistant-dev-mode.test.ts` pins it: both dump sites (openAssistant and the inline generate) assert `logger.debug` was used and that no Faro-backed level (`info`/`warn`/`error`) was called.

**First real-duration actions under engagement-gated init (medium):** went with your first suggestion — `withFaroUserAction()` now treats reaching an interactive funnel as engagement. When Faro is armed but uninitialized, it awaits `notifyFaroEngagement()` (completing init and flushing the buffer) before deciding whether to start or pass through, so the first `pathfinder_section_run` / `pathfinder_remote_step` in a fresh tab gets a real action envelope. The await is gated on `!faroInstance`: once initialized, the action still starts synchronously in the same tick as before (this also matters for the fake-timer timeout tests, which pin the synchronous start). Outside Grafana Cloud nothing arms, `notifyFaroEngagement()` resolves immediately, and the wrapper remains a pure passthrough. Tests added for all three shapes: fresh armed tab captures the first `pathfinder_section_run`, same for `pathfinder_remote_step`, and never-armed stays a passthrough with zero init.

moxious (approved):

leslie-heinzen (approved):

PR #1280, opened 2026-07-08
fix(ci): post PR comments via workflow_run so fork PRs stay green

moxious:
## Problem

CI comments on PRs in two places — the `publish-report` job (Playwright e2e results table) and the `bundle-size` step (bundle-size diff). Both run on the `pull_request` trigger.

For PRs opened from **forks**, GitHub forcibly downgrades the workflow's `GITHUB_TOKEN` to **read-only**, regardless of any `permissions:` block. The comment steps then fail with:

```
Resource not accessible by integration
```

So CI goes red on every PR from an external collaborator, even though nothing is actually broken. (Example: [run 28970282474](https://github.com/grafana/grafana-pathfinder-app/actions/runs/28970282474/job/85967229765?pr=1279).) Same-repo PRs are unaffected because they get a full write token.

The read-only-token restriction is deliberate on GitHub's side: it's the defense against **"pwn requests"** — untrusted fork code running in a job that also holds a write-capable token. The `permissions:` block is a *ceiling*, not a *grant*, and GitHub applies a lower ceiling to forks.

### `ci.yml` — Playwright results comment → `workflow_run` split
- Removed the `publish-report` job.
- Added a tiny `save-pr-number` job that persists `github.event.number` as an artifact (the `workflow_run` event doesn't reliably know the PR number for fork PRs).
- The matrix job continues to upload the per-version `gf-playwright-report-*` summary artifacts unchanged.

### New `ci-comment.yml` — the write-back, triggered by `workflow_run`
Runs when **CI** completes, downloads the report + PR-number artifacts from the triggering run, aggregates the table with our existing `build-playwright-comment.js`, and posts the comment.

### `bundle-stats.yml` — `continue-on-error` on the bundle-size step
The `grafana/plugin-actions/bundle-size` action couples build → compare-against-main → comment in a single third-party composite step, gated only on `pull_request`. It can't be split into a `workflow_run` flow without reimplementing its diff logic, and its comment code expects a `pull_request` context. So instead of splitting it, the step is marked `continue-on-error: true`: on fork PRs its comment failure no longer reds the build. Structure is otherwise unchanged; on same-repo PRs it still comments as before.

## Why the `workflow_run` model is safe (and avoids pwn attacks)

A pwn request requires **both** in one job: (1) execution of attacker-controlled fork input as *code*, and (2) a privileged context (write token / secrets). This design keeps those two apart:

1. **It runs only trusted code from the default branch.** A `workflow_run` workflow is defined by the version of `ci-comment.yml` on the base repo's default branch — a fork's edits to it are ignored. The `checkout` takes the **default ref**; it never checks out `github.event.workflow_run.head_sha`, so fork source is never placed on disk or executed.
2. **Everything crossing the trust boundary is inert data, never code.** The only things flowing from the fork-triggered run are downloaded artifacts (a results table + a PR number). They are consumed by *our* trusted script and rendered into a comment — never executed.
3. **Untrusted input is validated.** Artifact contents are treated as tainted: the PR number is accepted only if it matches `^[0-9]+$` before being used in any later step, so it can't inject into an expression.

Because the write-capable token in `ci-comment.yml` only ever touches trusted code + inert data, the fork author gains no execution and therefore can't abuse the token. The one rule that keeps this safe — **never check out the PR head ref or execute downloaded/fork-provided content in this workflow** — is documented in a header comment in `ci-comment.yml` to prevent future regressions.

`workflow_run` is preferred over `pull_request_target` here: both yield a write token on fork PRs, but `pull_request_target`'s natural shape tempts you to check out and build the PR (untrusted code + write token = the exact vulnerability), whereas `workflow_run`'s natural shape is "download results, act on them" — data-in, no code execution.

## Notes / trade-offs
- `workflow_run` workflows do **not** appear as PR status checks — they run in the Actions tab against the default branch. So even if commenting fails, it can never red or block a PR.
- **This workflow only takes effect once merged to the default branch** (that's how `workflow_run` resolves its definition), so it can't be fully exercised from this PR itself.
- Audited the change with `zizmor` — `ci-comment.yml` is clean; no new findings introduced.

### DIFF
--- a/.github/workflows/bundle-stats.yml
+++ b/.github/workflows/bundle-stats.yml
@@ -24,4 +24,10 @@ jobs:
         with:
           persist-credentials: false
 
+      # The bundle-size action builds, compares against main, and comments in a single
+      # composite step. On PRs from forks the GITHUB_TOKEN is read-only, so the comment step
+      # fails with "Resource not accessible by integration". This is informational reporting,
+      # not a gate, so we let it fail without reddening the build. The action cannot be split
+      # into a workflow_run flow (like ci.yml's comment) without reimplementing its diff logic.
       - uses: grafana/plugin-actions/bundle-size@4698961fa64137fcac207108397ebe8a738a33d1
+        continue-on-error: true
--- a/.github/workflows/ci-comment.yml
+++ b/.github/workflows/ci-comment.yml
@@ -1,0 +1,98 @@
+name: CI comment
+
+# Posts the aggregated Playwright e2e results table as a PR comment.
+#
+# WHY THIS IS A SEPARATE WORKFLOW (and why it is safe):
+# The CI workflow runs on `pull_request`. For PRs opened from forks, GitHub forcibly
+# downgrades that workflow's GITHUB_TOKEN to read-only regardless of any `permissions:`
+# block, so it cannot comment on the PR. This is GitHub's defense against "pwn requests"
+# (untrusted fork code running with a write-capable token).
+#
+# This workflow runs on `workflow_run`, which executes in the context of the BASE repo's
+# default branch, so its token is write-capable even for fork PRs. That is only safe
+# because of two invariants that MUST be preserved:
+#   1. It runs ONLY trusted code from the default branch. The checkout below takes the
+#      default ref - it never checks out `github.event.workflow_run.head_sha`, so fork
+#      source code is never placed on disk or executed here.
+#   2. Everything crossing the trust boundary from the fork-triggered run arrives as inert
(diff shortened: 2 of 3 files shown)


moxious:
Note this PR was raised because functionally, we can't have totally green pipelines on user-contributed PRs, as our pipeline requires commenting on the PRs, which requires in turn this permissions change.  I reviewed this 3 different ways and find the automatic evidence weak "most often used in an insecure way" but as our org has zizmor configured and we can't have a merge-able build unless zizmor is happy, this PR is effectively not going to be merge-able even if the security thesis is correct.

PR #1283, opened 2026-07-09
fix(companion): floating panel coexists with and dodges native modals

### LINKED ISSUE #1003: Bug: Pop out pathfinder z-axis hides modals

Jayclifford345:
Combines the two bug-fix layers of the former companion-mode stack (#1010 + #1011) into one PR, now that the upstream prerequisite has shipped.

## What this does

**Render the floating panel into `getPortalContainer()`** (instead of `document.body`). Grafana's `ModalBase` treats portal-container content as "inside" the modal, and as of grafana/grafana#126261 — **shipped in Grafana v13.1.0** — presses there no longer dismiss the modal. So on Grafana >=13.1, clicking or scrolling the popped-out panel next to an open native modal no longer closes it (#1007). On older Grafana the behavior is unchanged (no regression; the fix simply lights up when the runtime supports it). Falls back to `document.body` when the container is absent (e.g. tests). Also excludes `[data-pathfinder-content="true"]` from the devtools auto-detect filter and the highlight click-outside handler, so panel clicks don't record spurious interactions or clear an active highlight.

**Add `modal-watcher` + modal-aware dodging** — `detectModalActive()` (shared modal detection that excludes Pathfinder's own `role=dialog` panel) with a refcounted `startModalWatch`/`stopModalWatch` emitting `pathfinder-modal-state-changed`. `useHighlightDodge` folds the largest visible native modal rect into its dodge union, so the popped-out panel repositions beside open modals instead of covering them — version-independent. `GlobalInteractionBlocker.isModalActive` now delegates to `detectModalActive` (single source; also fixes the panel self-trigger).

Fixes #1003

## What happened to the rest of the stack

The authored `companion` section flag (#1009) and the auto-pop-out controller (#1012) are **deferred pending real user demand** — auto-moving the panel is a UX commitment we're not ready to make, and the flag would permanently commit the guide schema. With this PR, users pop the panel out manually exactly as today; once popped out it coexists with and dodges native modals. See the closing comments on #1009–#1012.

## Verification

- Full unit suite green locally (`npm run test:ci`, 320 suites); lint clean on changed files.
- The Playwright matrix now includes a 13.1.0 image where the coexistence behavior is live.
- Live check (Grafana >=13.1): pop out the panel → open a native modal → panel clicks/scrolls don't dismiss it, and the panel repositions beside it.
- Live check (12.x): dodge works; dismiss-on-click unchanged from today (pre-existing, fixed only on >=13.1).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,7 @@ ci/
 /blob-report/
 /playwright/.cache/
 /playwright/.auth/
+/.playwright-mcp/
 
 # Editor
 .idea
--- a/src/components/floating-panel/FloatingPanel.tsx
+++ b/src/components/floating-panel/FloatingPanel.tsx
@@ -1,8 +1,9 @@
 import React, { useState, useCallback, useEffect, useRef } from 'react';
 import { createPortal } from 'react-dom';
-import { IconButton, useStyles2 } from '@grafana/ui';
+import { IconButton, useStyles2, getPortalContainer } from '@grafana/ui';
 import { reportAppInteraction, UserInteraction } from '../../lib/analytics';
 import { buildPathfinderShareUrl } from '../../utils/pathfinder-search-params';
+import { startModalWatch, stopModalWatch } from '../../interactive-engine';
 import { getFloatingPanelStyles } from './floating-panel.styles';
 import { useDragResize } from './useDragResize';
 import { useHighlightDodge } from './useHighlightDodge';
@@ -61,8 +62,13 @@ export function FloatingPanel({
   const [isDodging, setIsDodging] = useState(false);
   const { geometry, setPosition, drag, resize } = useDragResize();
 
-  // Auto-reposition when interactive highlights overlap the panel
-  useHighlightDodge(geometry, panelState === 'minimized');
+  // Auto-reposition to dodge interactive highlights and any open native modal
+  useHighlightDodge(geometry, panelState === 'minimized', true);
+
+  useEffect(() => {
+    startModalWatch();
+    return () => stopModalWatch();
+  }, []);
 
   const handleMinimize = useCallback(() => {
     setPanelState('minimized');
@@ -265,5 +271,5 @@ export function FloatingPanel({
     </>
   );
 
-  return createPortal(panelContent, document.body);
+  return createPortal(panelContent, getPortalContainer());
 }
--- a/src/components/floating-panel/useDragResize.ts
+++ b/src/components/floating-panel/useDragResize.ts
@@ -90,9 +90,15 @@ function useDrag(
     if (!dragStartRef.current) {
       return;
     }
+    const { originX, originY } = dragStartRef.current;
     dragStartRef.current = null;
     setGeometry((prev) => {
       panelModeManager.setPanelGeometry(prev);
+      if (prev.x !== originX || prev.y !== originY) {
+        document.dispatchEvent(
+          new CustomEvent('pathfinder-floating-manual-move', { detail: { x: prev.x, y: prev.y } })
+        );
+      }
       reportAppInteraction(UserInteraction.FloatingPanelMoved, {
         trigger: 'manual_drag',
         x: prev.x,
--- a/src/components/floating-panel/useHighlightDodge.test.ts
+++ b/src/components/floating-panel/useHighlightDodge.test.ts
@@ -1,0 +1,87 @@
+import { renderHook } from '@testing-library/react';
+
+import { useHighlightDodge } from './useHighlightDodge';
+
+const GEOMETRY = { x: 50, y: 50, width: 400, height: 400 };
+
+function mockRect(el: HTMLElement, left: number, top: number, width: number, height: number) {
+  el.getBoundingClientRect = () =>
+    ({
+      left,
+      top,
+      right: left + width,
+      bottom: top + height,
+      width,
+      height,
+      x: left,
+      y: top,
+      toJSON: () => ({}),
+    }) as DOMRect;
+}
+
+function addModal(left: number, top: number, width: number, height: number, attrs: Record<string, string> = {}) {
+  const el = document.createElement('div');
+  el.setAttribute('role', 'dialog');
+  el.style.opacity = '1';
+  Object.entries(attrs).forEach(([k, v]) => el.setAttribute(k, v));
+  document.body.appendChild(el);
+  mockRect(el, left, top, width, height);
+  return el;
+}
+
+function captureEvents() {
+  const events: string[] = [];
+  const dodge = () => events.push('dodge');
+  const compact = () => events.push('compact');
+  document.addEventListener('pathfinder-floating-dodge', dodge);
+  document.addEventListener('pathfinder-floating-compact', compact);
+  return {
+    events,
+    cleanup: () => {
+      document.removeEventListener('pathfinder-floating-dodge', dodge);
+      document.removeEventListener('pathfinder-floating-compact', compact);
+    },
+  };
+}
+
+beforeEach(() => {
+  Object.defineProperty(window, 'innerWidth', { value: 1280, configurable: true });
+  Object.defineProperty(window, 'innerHeight', { value: 800, configurable: true });
+  document.body.innerHTML = '';
+});
+
+describe('useHighlightDodge — modal dodging', () => {
+  it('dodges to a clear corner when a modal overlaps the panel (dodgeModal=true)', () => {
+    addModal(0, 0, 300, 300); // top-left modal, overlaps the panel at 50,50
+    const cap = captureEvents();
+    renderHook(() => useHighlightDodge(GEOMETRY, false, true));
+    expect(cap.events).toContain('dodge');
+    expect(cap.events).not.toContain('compact');
+    cap.cleanup();
+  });
+
+  it('compacts when a full-viewport modal leaves no clear corner', () => {
+    addModal(0, 0, 1280, 800); // covers everything
+    const cap = captureEvents();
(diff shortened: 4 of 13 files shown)


moxious (changes requested):
## 🔴 Request changes

Nice work landing the portal fix — rendering into `getPortalContainer()` is the right mechanism for the v13.1 dismiss-on-click problem, and the refcounted watcher is clean. My concerns are all in the seam between the **modal detector** (fires the event) and the **dodge positioner** (moves the panel), plus one manual-drag interaction. These are behavioral, so I'd like to resolve them before merge.

### 1. Detector and positioner use different modal selectors — the panel can restore *onto* a modal it can't see

`detectModalActive()` (`modal-watcher.ts`) counts `.journey-image-modal` and `[data-overlay-container="true"]` as active modals, but `largestVisibleModalRect()` (`useHighlightDodge.ts`) only queries `[role="dialog"], [aria-modal="true"]`.

When a modal matched *only* by the detector opens:
- `pathfinder-modal-state-changed{isOpen:true}` fires → `checkAndDodge` runs → `largestVisibleModalRect()` returns `null`.
- With no highlights present, `obstacles.length === 0`, so the empty-obstacles branch runs. If `previousPositionRef` was set by an earlier dodge, it dispatches `restore-position`/`restore-full` — **moving the panel back onto the just-opened modal**, the opposite of the intent.

Confirmed affected: Pathfinder's own journey image lightbox (bare `<div class="journey-image-modal">`, no role/aria — `link-handler.hook.ts`) and any native modal exposed only via `data-overlay-container`. That the watcher's `attributeFilter` explicitly lists `data-overlay-container` suggests those modals are real targets, not hypothetical.

**Suggested fix:** single source of truth for "visible native modals" — have `largestVisibleModalRect` and `detectModalActive` share one scan/selector list so detection and positioning can never diverge.

### 2. Manual drag during an active dodge is silently discarded when the highlight clears

Repro:
1. Panel at **A**. Highlight appears and overlaps → `previousPositionRef` is set to **A**, panel dodges to corner **B**.
2. User drags the panel from **B** back to **C** (e.g. deliberately, to see it next to the highlight). Geometry **C** is persisted; `previousPositionRef` is untouched (drag doesn't run `checkAndDodge`).
3. Highlight is removed → `checkAndDodge` sees zero obstacles → dispatches `restore-position` with the **stale A**.
4. Panel teleports to **A**, discarding **C** — and the restore overwrites the persisted geometry too.

So the panel jumps back to a position from two moves ago and throws away the user's manual placement.

**Suggested fix:** when a manual drag occurs while a dodge is active, either clear `previousPositionRef` (treat the drag as the new baseline — user overrode the dodge) or update it to the dragged position, so restore doesn't fight the user.

### 3. Dodge quality: only the largest modal is modeled, and the compact fallback can be a no-op

Two related gaps in `findDodgePosition`/`largestVisibleModalRect`:

- **Largest-only:** `largestVisibleModalRect()` keeps only the single max-area modal (`if (area > bestArea)`). With two dialogs open — a large non-overlapping one and a small overlapping one — the union is the non-overlapping rect, `rectsOverlap` is false, and the panel stays buried under the small modal. Collecting *all* visible modal rects closes this.
- **Compact is shrink-in-place:** when no corner clears the obstacle union, we fall back to `pathfinder-floating-compact`, which only sets `panelState = 'compact'` (height → `auto`) without repositioning. If the occlusion is horizontal or behind the header, compacting the height uncovers nothing and the panel keeps occluding. Worth a comment on the intended limits here at minimum, ideally a reposition attempt in compact mode.

Also note the union-of-all-obstacles approach treats the empty space *between* a highlight and a modal as occupied, so two small far-apart obstacles can inflate the union to near-full-screen and force compact when there was plenty of room between them.

### 4. An already-open modal that resizes/repositions doesn't re-trigger the dodge

`check()` emits `pathfinder-modal-state-changed` only when the `detectModalActive()` boolean *flips*; the 1s poll inherits the same gate, and `useHighlightDodge`'s observer only reacts to highlight nodes. So a modal that stays open but grows/shrinks (async content load, validation expand, viewport re-center) never re-runs `checkAndDodge` — the panel stays in its prior dodged/compacted spot even when a clear corner opens up. Lower severity (native Grafana modals aren't user-resizable), but real for content-driven height changes.

---

## 🟡 Non-blocking (cleanup / follow-up)

- **Duplicated, already-divergent modal logic.** The visibility predicate and modal DOM scan now live in both `modal-watcher.ts` and `useHighlightDodge.ts`, and they already disagree (`opacity > 0` vs `opacity === 0`; watcher includes `data-overlay-container`, positioner doesn't). This divergence is the root cause of #1 — a shared helper fixes both at once.
- **1s `setInterval` poll alongside the attribute-filtered MutationObserver.** The observer already catches modal open/close; the unconditional per-second `querySelectorAll` + `getComputedStyle` scan runs for the panel's whole lifetime. If it's a backstop for changes the filter misses, a comment saying so (or a longer interval) would help; otherwise it looks redundant.
- **`data-pathfinder-content="true"` self-exclusion inlined in four places** (`largestVisibleModalRect`, `detectModalActive`, `navigation-manager` click-outside, `action-detector`). A shared `isPathfinderInternal(el)` predicate would keep a future marker change from silently re-including the panel at one site. (Existing `isPathfinderInternal` is Tier-3 `integrations/`; reuse means promoting it lower, not importing upward.)

---

Happy to pair on the shared-modal-scan helper — items #1 and the first cleanup bullet collapse into one change.

moxious:
Manual extra follow-up, this one is tricky.  Note that in general on screen size, users will be able to get you into situations where nothing you do is reasonable.  In the case of the FE team, they'll simply know better, but this PR is a source of bugs in the future; I'm not sure how best to note this (perhaps something somewhere relevant in the documentation).  The source of bugs and future user complaints is that the floating panel will be jumping around (for good reason) but sometimes it might occlude itself, or there might be "nowhere to jump to" that's reasonable.  

We know Grafana isn't best on mobile, but in the pathological case where someone is trying this on mobile or on a low-res iPad, there's literally nothing this code can do that would make the experience good.  We might consider a different change later that simply FULLY DISABLES floating panel mode on screens below a certain size.

Not blocking this PR but I think the functionality has some conceptual challenges with the rest of how the platform works.  Don't take this as "you have to change this PR to make it perfect" but put a bit more thought into it, and perhaps the leave-behind from this PR is an issue to address later (make it harder to get into the position where this feature will misbehave).

Jayclifford345:
Thanks for the thorough review — all four behavioral items are addressed in 46b5e2cc, plus two of the three cleanups.

**1. Detector/positioner divergence → one shared scan.** `getVisibleModalRects()` in `modal-watcher.ts` is now the single scan behind both `detectModalActive()` (`= rects.length > 0`) and the dodge positioner. One selector list (`role="dialog"`, `aria-modal`, `data-overlay-container`, `.journey-image-modal`), one visibility predicate, one self-exclusion. `largestVisibleModalRect` is deleted from `useHighlightDodge`. Regression test added: the journey image lightbox (class-only, no role/aria) now triggers a dodge instead of a restore-onto-the-modal.

**2. Manual drag vs restore.** `useDrag.onPointerUp` now dispatches `pathfinder-floating-manual-move` with the final position (only when the position actually changed — a header click without movement doesn't fire). While a dodge is active, the hook updates its restore target to the dragged position, so when obstacles clear the panel "restores" to where the user put it — no stale teleport, and the user's persisted geometry is never overwritten by an older position. Test covers dragged-during-dodge and drag-with-no-dodge-armed.

**3. Dodge quality.** Three changes: (a) every visible modal is an obstacle, not just the largest — test asserts the chosen corner clears both a large non-overlapping and a small overlapping dialog; (b) corners are tested against each obstacle individually instead of one union rect, so free space between far-apart obstacles is usable again (test: highlight top-left + small modal bottom-right no longer forces compact); (c) when no corner fits the full panel, we still compact but also reposition to a corner that fits the min-height panel (280px, the best known lower bound for the compacted size) when one exists — compact-in-place remains only the true last resort (full-viewport modal).

**4. Modal resize/move re-trigger.** `check()` now compares a rounded rect signature instead of a boolean, so an open modal that grows, shrinks, or re-centers re-emits `pathfinder-modal-state-changed` and the dodge re-runs. The 1s poll is now the documented backstop for geometry changes that mutate no watched attribute (e.g. re-centering on window resize). Only `useHighlightDodge` listens to this event (the blocker calls `detectModalActive()` on demand), so the extra emissions are safe. Test simulates a height change on an open dialog and asserts the second emission.

**Cleanups:** the duplicated modal logic is gone with the shared scan (your first bullet), and the four inline `data-pathfinder-content` exclusions now go through a shared `isPathfinderContent()` in `src/lib/dom/pathfinder-content.ts` (Tier 1, importable from everywhere that needs it). I left `integrations/assistant-integration`'s `isPathfinderInternal` alone — it has broader semantics (also matches `interactive-step-*` testids), so folding it in would change behavior there; happy to do that as a follow-up if you'd like it unified.

**On the small-screen concern:** agreed there's no good answer inside the dodge logic itself — opened #1284 to track disabling/limiting floating mode below a minimum viewport size, per your suggestion.

moxious (approved):

PR #1285, opened 2026-07-09
fix(recommendations): bound the online package index fetch so OSS boot isn't blocked

Jayclifford345:
## Summary

OSS users who haven't accepted T&C (recommender disabled — the OSS default) stared at the recommendations skeleton for 12–15s on first panel open (minutes on slow networks). The disabled branch of `ContextService.fetchRecommendations` awaits the backend `/package-recommendations` resource, which on a cold cache (every plugin restart) fetched `repository.json` and then synchronously downloaded **all 389** packages' `manifest.json` (8-way, 5s timeout each) before responding — the code's own comment admitted a "~50s manifest fan-out window". Meanwhile even locally-bundled recommendations were held back, and the frontend had no timeout on the call.

Three layered bounds fix it:

1. **Backend — enrich only targeted entries** (389 → 89 manifest fetches). The 300 untargeted entries exist solely for lazy by-ID resolution, and `OnlineCdnPackageResolver.loadFromCdn` already falls back to fetching `manifest.json` itself when none is inlined (now pinned by a test).
2. **Backend — 3s total enrichment budget.** Entries that miss the deadline ship manifest-less (the frontend already renders a stub); budget-expired responses are cached for 15min instead of 6h so the degraded tail self-heals. Durable against CDN growth and slow RTT.
3. **Frontend — 3s boot budget** racing `fetchOnlinePackageRecommendations()`. On timeout, bundled/static recommendations render immediately; the race abandons the await, not the request — the client's shared in-flight promise keeps populating its module cache and the sticky `unavailable` flag is untouched, so the next recommendations cycle (any navigation) picks the packages up instantly.

Also bumps manifest fan-out concurrency 8 → 16 (a one-constant change, independently revertable).

**Measured against the live CDN: cold-cache fetch 12–15s → 1.35s**, all 89 targeted entries fully enriched within budget (`partial=false`).

## What to look at

- `pkg/plugin/package_recommendations.go` — the budget context threading in `enrichPackagesWithManifests` (per-fetch contexts now derive from `budgetCtx`; the select stops queueing on expiry) and the `partial` flag → short-TTL path in `getCachedPackageRecommendations`.
- `src/context-engine/context.service.ts` — `getOnlinePackageRecommendations` timeout race semantics; the sentinel message distinguishes expected timeouts (info) from real failures (warn).
- The enrichment budget still derives from the `context.WithoutCancel` context, so a cancelled request can't shrink the budget for single-flight waiters.

## Why

Air-gapped installs improve too: a TCP-level hang now costs ≤3s of skeleton instead of the full backend timeout chain, and the one-attempt-per-session contract is unchanged.

## How to verify

1. Build backend + `npm run server`, leave T&C unaccepted, restart the Grafana container (clears the in-process cache), open the panel → recommendations paint in ~1.5–2s with CDN cards and milestone counts intact (previously ~12–15s of skeleton).
2. Slow-CDN sim (delay is backend-side, browser throttling won't work): `docker compose exec -u root grafana sh -c "apk add iproute2 && tc qdisc add dev eth0 root netem delay 1000ms"` → bundled/static recommendations paint at ≤ ~3s; navigate to another page → CDN cards appear (progressive pickup). Clean up with `tc qdisc del dev eth0 root netem`.
3. Lazy fallback: expand a path-type CDN guide card → milestones resolve via per-ID `manifest.json` fetches in the Network tab.
4. `go test ./pkg/plugin/...` and `npm run check` both pass (new tests: total-budget bound, partial short-TTL, untargeted-never-enriched, frontend timeout race, cache population after abandoned await, resolver manifest fallback).

## Breaking changes

None. Response shape is unchanged; untargeted entries simply arrive without inlined manifests, which the existing stub/fallback paths already handle.

## Reversibility

Fully reversible. Each bound is independent: the concurrency bump is one constant, the frontend budget is one `withTimeout` wrapper, and the backend budget/targeted-only enrichment are contained in `enrichPackagesWithManifests` + the `partial` cache flag.

### DIFF
--- a/pkg/plugin/package_recommendations.go
+++ b/pkg/plugin/package_recommendations.go
@@ -30,10 +30,10 @@ const (
 	packageRepositoryPartialCacheTTL = 15 * time.Minute
 
 	// Per-manifest fetch limits — keep manifests small and the fan-out bounded
-	// so a slow CDN can't stall the whole response or run us out of memory.
-	packageManifestFetchTimeout = 5 * time.Second
-	packageManifestMaxBytes     = 256 * 1024
-	packageManifestConcurrency  = 16
+	// so the enrichment can't run us out of memory. Latency is bounded by
+	// packageManifestEnrichTotalBudget, not per fetch.
+	packageManifestMaxBytes    = 256 * 1024
+	packageManifestConcurrency = 16
 )
 
 // packageManifestEnrichTotalBudget caps the whole manifest fan-out so the
@@ -229,9 +229,9 @@ func (a *App) getCachedPackageRecommendations(ctx context.Context) (*PackageReco
 
 	// Detach the upstream fetch from the request's cancellation: a canceled
 	// request (browser closed, panel collapsed mid-flight) must not poison
-	// the 6-hour cache with a "context canceled" error. The per-fetch
-	// timeouts inside fetchAndParsePackageRepository / enrichPackagesWithManifests
-	// still apply because they're added with their own context.WithTimeout.
+	// the 6-hour cache with a "context canceled" error. The index fetch
+	// timeout and the enrichment budget still apply because they're added
+	// with their own context.WithTimeout.
 	resp, partial, err := fetchAndParsePackageRepository(context.WithoutCancel(ctx), packageRepositoryURL)
 
 	packageCacheMu.Lock()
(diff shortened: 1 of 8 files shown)


moxious (changes requested):
Minor but worth changing before we cast it in.

pkg/plugin/package_recommendations.go:34,44,359 — packageManifestFetchTimeout (5s) is dead configuration: every per-fetch fetchCtx derives from budgetCtx (3s deadline via packageManifestEnrichTotalBudget, line 324), and context.WithTimeout always takes the earlier deadline, so the 5s per-fetch cap can never actually fire in production, the 3s budget always wins first. 

Not a crash, but misleading: a reader tuning per-fetch latency would change a constant that has no effect. Worth either removing the now-redundant inner timeout or asserting/documenting that the budget must stay ≤ the per-fetch timeout for it to matter.

Otherwise LGTM, patch that and it's gtg

Jayclifford345:
Fixed in 70e65f36 — went with removal rather than documenting the constraint. `packageManifestFetchTimeout` and the inner `context.WithTimeout` are gone; manifest fetches now run directly on the 3s `packageManifestEnrichTotalBudget` context, so the budget is visibly the only latency bound for enrichment. Also updated the two comments that referenced the per-fetch timeout (the constants block and the detach-from-cancellation note in `getCachedPackageRecommendations`). The index fetch keeps its own 5s cap inside `defaultPackageRepositoryFetcher`.

moxious (approved):

Issue #1232, opened 2026-07-02
feat: Keep context for current location in guide list


hedss:
When selecting a Pathfinder guide from the `Recommended` list, returning to the list after viewing the guide resets to the first entry. This means that users have to potentially scroll down the list until they reach the guide they need.

For example, in a recent enablement session we had 66 guides available for users. When using guides towards the end of the list, this meant they had to continually return to the `Recommended` list and then scroll all the way to the bottom of that list to pick the next guide in the series: 
<img width="960" height="1244" alt="Image" src="https://github.com/user-attachments/assets/f83d49ae-66c8-4fab-a84c-89e5bb1a30b2" />

Pathfinder should keep contextual information about the last guide selected from the `Recommended` list, and then return to that position when it is next rendered.

Issue #1002, opened 2026-06-11
FE event feedback


Jayclifford345:
This epic contains a collection of issues based on feedback from the Field Engineering team during their last event.

moxious:
https://github.com/grafana/grafana-pathfinder-app/issues/1133

moxious:
## 🧭 Implementation readiness

**Stakes:** This is not one implementation. It is a bundle of separate feedback items from the Field Engineering team.

**Sanity check:** The issue's own first line says "This epic contains a collection of issues based on feedback from the Field Engineering team during their last event." A single agent cannot ground or implement a bundle in one pass.

**Verdict:** ⚪ Excluded. Not an implementation issue. A human should split this into individual issues before any of them is ready for an agent. (tracking: `agent/excluded`)

_Posted by the `prep-issues` tool, not written by a person._

PR #1286, opened 2026-07-09
fix(docs-panel): keep "my learning" navigation in the sidebar panel

moxious:
## Summary

"Return to my learning" (the guide footer button) and the "My learning" tab-bar icon did a full-page navigation to the plugin home page, dropping users out of the Grafana page they were on and tearing down the sidebar. Multiple field-event attendees reported this context loss while working through guides alongside the Grafana UI. Both controls now switch to the in-panel recommendations tab instead, so users stay in the sidebar next to whatever they were doing.

## What to look at

- **[`DocsPanelContentArea.tsx`](https://github.com/grafana/grafana-pathfinder-app/blob/f333c8790af1317466fbab10ab0586b276f18359/src/components/docs-panel/components/DocsPanelContentArea.tsx)** — the footer button changed from `window.location.assign(PLUGIN_BASE_URL)` (a hard reload) to `model.setActiveTab('recommendations')`. `recommendations` is a permanent tab, so the switch is a pure `activeTabId` state change with no content fetch.
- **[`TabBarActions.tsx`](https://github.com/grafana/grafana-pathfinder-app/blob/f333c8790af1317466fbab10ab0586b276f18359/src/components/docs-panel/components/TabBarActions.tsx)** + **[`DocsPanelTabBar.tsx`](https://github.com/grafana/grafana-pathfinder-app/blob/f333c8790af1317466fbab10ab0586b276f18359/src/components/docs-panel/components/DocsPanelTabBar.tsx)** — new optional `onNavigateToRecommendations` prop, wired from the tab bar's existing `onSetActiveTab`. The old `locationService.push` is kept only as a fallback for when `TabBarActions` renders without the handler.
- **[`context-panel.tsx`](https://github.com/grafana/grafana-pathfinder-app/blob/f333c8790af1317466fbab10ab0586b276f18359/src/components/docs-panel/context-panel.tsx)** — the recommendations empty-state "My learning" button is deliberately left as a full-page navigation (now with a comment explaining why): it only renders when there are no recommendations to return to in-panel, so opening the full My Learning page is the correct escape hatch. It also closes the sidebar, which the two fixed buttons do not.
- **Scope confinement** — these components render only in the sidebar surface. The floating panel and the fullscreen page use `FloatingPanelContent`/`FullScreenLayout`, so no cross-mode behavior changes. The one shared case (the sidebar showing the fullscreen-mode notice) intentionally keeps the tab bar interactive, and switching to recommendations there is consistent with that notice's documented "queue the next active tab" behavior.

## How to verify

1. In the Pathfinder sidebar, open a guide while on a dashboard (or Explore) page. Scroll to the bottom of the guide and click **Return to my learning** — the sidebar should switch to the recommendations list, and the underlying Grafana page should **not** reload or navigate away.
2. With a guide tab open, click the **My learning** (book) icon in the tab bar — same in-panel switch, no navigation away.
3. Confirm the guide tab stays open and you can switch back to it from the tab bar.

## Breaking changes

None. Fully reversible — no storage, schema, telemetry, or URL contract changes.

## Out of scope

- **Scroll-position preservation in the guide list** — the second ask in #1051. The existing scroll machinery (`#inner-docs-content`) only wraps guide content, not the recommendations list, so preserving list scroll is a separate change tracked in #1232.

Addresses #1051. Refs #1232, #1002.

### DIFF
--- a/src/components/docs-panel/components/DocsPanelTabBar.tsx
+++ b/src/components/docs-panel/components/DocsPanelTabBar.tsx
@@ -260,6 +260,7 @@ export function DocsPanelTabBar({
         activeTab={activeTab}
         isDevMode={isDevMode}
         onReloadActiveTab={reloadActiveTab}
+        onNavigateToRecommendations={() => onSetActiveTab('recommendations')}
       />
     </div>
   );
--- a/src/components/docs-panel/components/DocsPanelContentArea.test.tsx
+++ b/src/components/docs-panel/components/DocsPanelContentArea.test.tsx
@@ -1,0 +1,118 @@
+/**
+ * Tests for DocsPanelContentArea.
+ *
+ * Focused on the "Return to my learning" footer button, which must switch the
+ * panel back to the recommendations tab in place (issue #1051) rather than
+ * navigating away from the Grafana UI.
+ */
+
+import React from 'react';
+import { render, screen, fireEvent } from '@testing-library/react';
+import { DocsPanelContentArea, type DocsPanelContentAreaProps } from './DocsPanelContentArea';
+
+jest.mock('@grafana/i18n', () => ({
+  t: (_key: string, fallback: string) => fallback,
+}));
+
+jest.mock('@grafana/data', () => ({
+  ...jest.requireActual('@grafana/data'),
+  usePluginContext: () => ({ meta: { jsonData: {} } }),
+}));
+
+jest.mock('../../../lib/analytics', () => ({
+  reportAppInteraction: jest.fn(),
+  getContentTypeForAnalytics: jest.fn(() => 'docs'),
+  UserInteraction: {
+    DocsPanelInteraction: 'docs_panel_interaction',
+    OpenExtraResource: 'open_extra_resource',
+  },
+}));
+
+jest.mock('../../../docs-retrieval', () => ({
+  getMilestoneSlug: jest.fn(),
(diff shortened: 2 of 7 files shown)


src/components/docs-panel/components/DocsPanelTabBar.tsx
```diff
@@ -260,6 +260,7 @@ export function DocsPanelTabBar({
         activeTab={activeTab}
         isDevMode={isDevMode}
         onReloadActiveTab={reloadActiveTab}
+        onNavigateToRecommendations={() => onSetActiveTab('recommendations')}
```
leslie-heinzen, line 263: Just so I'm clear: swapping to the recommendations tab here and in the footer means that the only ways to get to the plugin homepage are:

* Direct navigation to `/a/grafana-pathfinder-app` from outside of pathfinder. I vaguely remember the onboarding team using this link for some things, for example.
* The recommendation tab's empty state screen has a My Learning button.

Is that accurate or is there another way to get there from pathfinder?

leslie-heinzen (approved):
LGTM, think this is the right call.

Clarifying question about ways to access the plugin homepage after this change.

src/components/docs-panel/components/DocsPanelTabBar.tsx
moxious: that's accurate

PR #1287, opened 2026-07-09
fix(block-editor): only show Library action once the user has saved a guide

### LINKED ISSUE #1257: Only show "manage guides" icon if the user has created a guide before
## Context

This icon/action should be conditional on guide-creation history rather than always visible.

> "If the user has created a guide before, we have a separate icon appear for managing those"

## Code pointers

- `src/components/block-editor/BlockEditorHeader.tsx:372-379` — the "Library" `Menu.Item` (icon `book-open`) inside the header's kebab "More actions" dropdown, calling `onOpenGuideLibrary`
- `src/components/block-editor/BlockEditor.tsx:911` — `handleOpenGuideLibrary`, opens `GuideLibraryModal`
- `src/components/block-editor/GuideLibraryModal.tsx` — lists/loads/deletes backend guides

Current gating is only `{isBackendAvailable && (...)}` — there is no check tied to guide-creation history. No `hasCreatedGuide`-style flag or equivalent exists anywhere in `src` (confirmed no localStorage flag or guides-list-length check). `useGuideHistory.ts` exists but is unrelated — it's undo/redo history, not a record of past guide creation. This will need a new signal (e.g. checking whether the backend guides list is non-empty, or a persisted flag) added as part of the fix.

---

alliasgher:
## Summary

The block editor's more-actions menu showed the "Library" item whenever the Pathfinder backend was available, even for users who had never saved a guide. This gates the item on whether the user actually has saved guides, so it stays hidden until there is something to manage.

## Why

The Library modal lists, opens, and deletes backend guides. Surfacing it before the user has created any guide points at an empty modal and adds a control with nothing behind it. `useBackendGuides` already loads the guide list on mount, so the presence check adds no new API call or state.

## What to look at

- `BlockEditorHeader` gains a `hasBackendGuides` prop; the Library `Menu.Item` now renders on `isBackendAvailable && hasBackendGuides` instead of `isBackendAvailable` alone.
- `BlockEditor` passes `hasBackendGuides={backendGuides.guides.length > 0}` from the already-fetched list.

## How to verify

1. Open the block editor as a user with no saved guides and open the more-actions (ellipsis) menu: the Library item is absent.
2. Save a guide, reopen the menu: the Library item appears.

## Breaking changes

None.

Fixes #1257

### DIFF
--- a/src/components/block-editor/BlockEditor.tsx
+++ b/src/components/block-editor/BlockEditor.tsx
@@ -20,7 +20,7 @@ import { useRecordingActions } from './hooks/useRecordingActions';
 import { useJsonModeHandlers } from './hooks/useJsonModeHandlers';
 import { useBlockConversionHandlers } from './hooks/useBlockConversionHandlers';
 import { useGuideOperations } from './hooks/useGuideOperations';
-import { useBackendGuides } from './hooks/useBackendGuides';
+import { useBackendGuides, hasManageableBackendGuides } from './hooks/useBackendGuides';
 import { useGuidePreviewProgress } from './hooks/useGuidePreviewProgress';
 import { isBackendApiAvailable } from '../../utils/fetchBackendGuides';
 import { getBlockEditorStyles } from './block-editor.styles';
@@ -1055,6 +1055,7 @@ function BlockEditorInner({ initialGuide, onChange, onCopy, onDownload }: BlockE
         isPostingToBackend={backendGuides.isSaving}
         onNewGuide={handleNewGuideClick}
         isBackendAvailable={backendAvailable}
+        hasBackendGuides={hasManageableBackendGuides(backendGuides)}
         hasBlocks={hasBlocks}
         isSelectionMode={selection.isSelectionMode}
         onToggleSelectionMode={selection.toggleSelectionMode}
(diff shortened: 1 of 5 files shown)


moxious (changes requested):
Thank you for this contribution!  I'd like to get this in

Changes needed to merge:
- Distinguish "guides not loaded yet" from "confirmed zero guides" (use isLoading/hasInitiallyLoaded) so the Library item doesn't flash-hide for existing users and doesn't get stuck hidden after a failed fetch.
- Consider keeping Library visible whenever isBackendAvailable is true and letting GuideLibraryModal's existing empty-state/refresh/error UI handle the "no guides"/"fetch failed" cases, rather than hiding the entry point.
- Add a test for the loading/error transition, not just the static boolean.

alliasgher:
Thanks for the detailed review. Updated in c3ec66cf:

- useBackendGuides now exposes `hasLoaded`, and a new `hasManageableBackendGuides` helper keeps the Library entry visible while the list is still loading and after a failed fetch, hiding it only once an initial fetch has confirmed an empty list. That removes both the flash-hide for existing users and the stuck-hidden-on-error case.
- Added `useBackendGuides.helpers.test.ts` covering the loading, error, confirmed-empty, populated, and loading-to-empty transition cases (plus the existing static header gating tests).

I kept the hide (per the issue) rather than always-showing and leaning on the modal's empty state, since the helper now handles the loading/error edges. If you would still rather keep the entry always visible and let GuideLibraryModal handle it, that is a one-line change and I am happy to switch.

moxious (approved):
Thanks for the (third in one day) great contribution, and for addressing the feedback.  LGTM!

PR #1288, opened 2026-07-09
fix(docs-panel): move overflow-tab chevron to the right and flip it when open

### LINKED ISSUE #1252: Fix chevron side on "Show N more tabs" control
## Context

The chevron icon on the overflow tab button is on the wrong side.

> "the chevron is on the wrong side"

## Code pointers

- `src/components/docs-panel/components/DocsPanelTabBar.tsx:162-180` — "Show N more tabs" control, `<Icon name="angle-down" size="sm" />` next to `+{overflowGuideTabs.length}`
- `src/components/docs-panel/docs-panel.styles.ts:398` — `chevronTab` style

Also note: the icon does not currently rotate/swap (e.g. to `angle-up`) when the overflow dropdown is open — worth fixing alongside the side placement since both are in the same handful of lines.

---

alliasgher:
## Summary

The "Show N more tabs" overflow control in the docs panel tab bar rendered the chevron to the left of the count, and the chevron stayed `angle-down` even while the dropdown was open. This puts the count first so the chevron sits on the right, and flips it to `angle-up` while the dropdown is open so it reflects the open/closed state.

## What to look at

- `DocsPanelTabBar` overflow button: the `<span>+N</span>` and `<Icon>` are swapped so the chevron trails the count, and the icon name is now `isDropdownOpen ? 'angle-up' : 'angle-down'`.

## How to verify

1. Open enough guide tabs to trigger the overflow control: the "+N" count appears with the chevron on its right, pointing down.
2. Click the control to open the dropdown: the chevron flips to point up. Close it: it points down again.

## Breaking changes

None.

Fixes #1252

### DIFF
--- a/src/components/docs-panel/components/DocsPanelTabBar.test.tsx
+++ b/src/components/docs-panel/components/DocsPanelTabBar.test.tsx
@@ -88,4 +88,21 @@ describe('DocsPanelTabBar', () => {
     expect(props.onSetActiveTab).toHaveBeenCalledWith('recommendations');
     expect(mockPush).not.toHaveBeenCalled();
   });
+
+  describe('overflow chevron', () => {
+    const overflowTab: any = { id: 'overflow-1', title: 'overflow-1', baseUrl: '', currentUrl: '' };
+
+    it('renders the count before the chevron and points it down when the dropdown is closed', () => {
+      render(<DocsPanelTabBar {...makeProps({ overflowGuideTabs: [overflowTab], isDropdownOpen: false })} />);
+      const button = screen.getByTestId(testIds.docsPanel.tabOverflowButton);
+      expect(button.firstElementChild?.textContent).toBe('+1');
+      expect(screen.getByTestId('angle-down')).toBeInTheDocument();
+    });
+
+    it('points the chevron up when the dropdown is open', () => {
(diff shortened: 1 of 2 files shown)


moxious (dismissed):
Nice catch pulling both halves of #1252 together — the side swap and the open/closed flip are easy to treat as separate bugs, but you fixed them as one coherent change. The test coverage is exactly right: you didn't just assert the new order, you asserted the actual open vs. closed icon behavior, which is the part most PRs would skip. Genuinely minimal diff too.

Changes needed to merge:
- (Branch caught a merge conflict) -- Rebase fix/overflow-tab-chevron onto current main — main already has a DocsPanelTabBar.test.tsx from #1286 (a "My learning" navigation regression test), and this branch conflicts with it.
- When resolving, keep both test suites in one file — the existing "My learning" test plus the new overflow-chevron tests — don't drop either.

alliasgher:
Thanks! Rebased onto main and moved the two chevron tests into the existing DocsPanelTabBar.test.tsx from #1286 (using its makeProps harness) instead of adding a second file. In d19bccee.

Jayclifford345 (approved):
LGTM! Huge thank your the contribution!

PR #1295, opened 2026-07-10
fix(selector-generator): reject React useId ids and prefer testid-scoped bare tags

Jayclifford345:
## Summary

Record mode was picking unstable selectors — sometimes embedding auto-generated fragments, sometimes missing an obviously better selector nearby. Both root causes were confirmed with runtime tracing of the candidate pipeline against a live Grafana instance (panel editor transformations, explore toolbar, Monaco editor, select dropdowns).

- **Reject React `useId()` ids.** `useId()` output sanitized for HTML ids (`_r_15p_`, or prefixed forms like `start-panel-_r_15p_`) passed `isAutoGeneratedId()`, so render-position-dependent ids won as `#id` candidates (score 50 — beating real `aria-label`/button-text candidates) and were also used as ancestor scopes. Both `isAutoGeneratedId()` and the `session-unstable` stability flag now recognize the underscore-sanitized form; the raw `:r15p:` colon form was already covered.
- **New `scoped-bare-tag` candidate.** No candidate combined a meaningful testid ancestor scope with a plain tag, so targets like the "Filter data by values" transform card fell back to i18n-sensitive text matching or hashed emotion classes. The new candidate emits selectors like `div[data-testid='data-testid New transform Filter data by values'] button`, guarded so it only wins via clean descendant scoping — never through positional `nth-child`/`nth-match` fallbacks or accidental global uniqueness.

## What to look at

- `src/lib/dom/selector-generator.ts` — the `_r_` regex in `isAutoGeneratedId()`/`computeStabilityFlags()`, the `candidateBareTag()` generator, and its three guards (no unscoped wins in `rankAndSelect()`, no nth-child/nth-match paths in `disambiguate()`).
- One test expectation updated: a testid-scoped bare tag now outranks `input[name=…]`, consistent with the existing ordering where aria-label/button-text already outrank `name`.

## How to verify

- `npm run test:ci` — all selector-generator tests pass (42/42).
- In record mode, click the "Filter data by values" card in the panel editor transformations tab: the recorded selector should be `div[data-testid='data-testid New transform Filter data by values'] button` with no `_r_…_` fragments.

## Breaking changes

None. Selector output changes only where the previous output embedded unstable React ids or fragile text/class fallbacks.


src/lib/dom/selector-generator.ts
```diff
@@ -709,6 +732,7 @@ function generateCandidates(element: HTMLElement): Candidate[] {
   push(candidateName(element));
   push(candidateHref(element));
   push(candidateCompound(element));
+  push(candidateBareTag(element));
```
leslie-heinzen, line 735: @Jayclifford345 bot review feedback. Categorized as blocking:

rankAndSelect() only accepts scoped-bare-tag after stable ancestor scoping (src/lib/dom/selector-generator.ts:925). However, generateFallbackSelectors() accepts globally unique candidates unchanged (src/lib/dom/selector-generator.ts:1108). This can return bare alternatives such as input or button. InteractiveBlockForm allows these alternatives to be promoted directly into the persisted reftarget.

**Suggested action:** Share candidate-admissibility logic between primary and fallback generation, ensuring bare-tag alternatives are always stably scoped. Add regression tests covering:

leslie-heinzen (changes requested):

src/lib/dom/selector-generator.ts
Jayclifford345: Fixed in 1476dc01. Extracted the per-candidate admission logic from `rankAndSelect()` into a shared `admitCandidate()` helper, now used by both `rankAndSelect()` and `generateFallbackSelectors()`. Fallback generation gets the same rules as the primary: unscoped bare-tag candidates are rejected even when globally unique, unique candidates get the same ancestor-scope upgrade, and overlay visibility matching is applied consistently. Added two regression tests: a globally-unique bare `button` with no stable ancestor never appears in fallbacks, and a bare `input` alternative is only admitted in its testid-scoped form (`div[data-testid='env-field'] input`).

leslie-heinzen (approved):

PR #1296, opened 2026-07-10
fix(recommendations): restore scroll position when returning to the list

### LINKED ISSUE #1232: feat: Keep context for current location in guide list

moxious:
## Summary

Selecting a guide from the Recommended list and returning to it previously reset the list to the top, which made long recommendation sets painful to navigate. This PR persists the Recommended list scroll offset in tab-scoped `sessionStorage` and restores it when the recommendations tab is mounted again after its content has actually settled.

The restore logic now avoids the two lifecycle traps called out in review: it no longer saves during unmount, and it waits for context, recommendations, and custom guides to finish loading before applying the saved offset.

## What to look at

- [`useRecommendationsScrollPosition`](https://github.com/grafana/grafana-pathfinder-app/blob/0e6604e85b90bfa1679c6d9b05fe43364313fff3/src/components/docs-panel/hooks/useRecommendationsScrollPosition.ts) owns the save/restore mechanics. The scroll listener saves only on real scroll events, the restore runs in `requestAnimationFrame`, and the one-shot guard latches only after the saved offset can fit in the rendered container.
- [`ContextPanelRenderer`](https://github.com/grafana/grafana-pathfinder-app/blob/0e6604e85b90bfa1679c6d9b05fe43364313fff3/src/components/docs-panel/context-panel.tsx) attaches the hook to the outer scrollable `.content` element and gates readiness on context data, recommendation fetch completion, and custom-guide completion.
- [`useContextPanel`](https://github.com/grafana/grafana-pathfinder-app/blob/0e6604e85b90bfa1679c6d9b05fe43364313fff3/src/context-engine/context.hook.ts) and [`usePublishedGuides`](https://github.com/grafana/grafana-pathfinder-app/blob/0e6604e85b90bfa1679c6d9b05fe43364313fff3/src/utils/usePublishedGuides.ts) expose settled-state flags so the scroll restore does not burn on an intermediate skeleton render.
- [`StorageKeys`](https://github.com/grafana/grafana-pathfinder-app/blob/0e6604e85b90bfa1679c6d9b05fe43364313fff3/src/lib/storage-keys.ts) adds the tab-scoped `RECOMMENDATIONS_SCROLL_POSITION` key, with the stable string contract updated in the matching test.

## Why

The Recommended tab fully unmounts while a guide tab is active, so an in-memory ref cannot preserve the list position. `sessionStorage` is intentionally scoped to the browser tab/session, which preserves the return trip without making the offset durable across unrelated browser sessions.

## How to verify

1. Open the Pathfinder sidebar with enough recommended guides to require scrolling, or temporarily reduce the panel height.
2. Scroll partway down the Recommended list and open a guide from the list.
3. Return by clicking the Recommended / My learning tab in the panel. Confirm the list restores to the saved position instead of jumping to the top.
4. Repeat with custom guides available above the suggested guides section. Confirm the restored position lands after custom guides finish loading and does not shift noticeably afterward.
5. Reload the browser tab after scrolling the Recommended list. Confirm the saved position is scoped to the current browser tab/session and is not treated as a durable cross-session preference.

## Breaking changes

## Reversibility

The only persisted state added here is `grafana-pathfinder-app-recommendations-scroll-position` in `sessionStorage`. Reverting the PR stops reading and writing that key; any leftover value is inert and cleared with the browser tab/session.

## Out of scope

- Changing which tab becomes active when a guide tab is closed. Existing tab-order behavior can land on the guide editor if the editor tab sits before the closed guide tab; this PR focuses on restoring scroll when the user returns to the Recommended tab.

Closes #1232

### DIFF
--- a/src/components/docs-panel/context-panel.tsx
+++ b/src/components/docs-panel/context-panel.tsx
@@ -24,6 +24,7 @@ import { CustomGuidesSection } from './CustomGuidesSection';
 import { usePublishedGuides, PublishedGuide } from '../../utils/usePublishedGuides';
 import { ContextPanelState, PackageOpenInfo } from '../../types/content-panel.types';
 import { getPackageRenderType } from '../../types/package.types';
+import { useRecommendationsScrollPosition } from './hooks';
 
 /**
  * Resolve the effective display type for a recommendation.
@@ -1108,9 +1109,11 @@ function ContextPanelRenderer({ model }: SceneComponentProps<ContextPanel>) {
     recommendations.length > 0 &&
     !configWithDefaults.acceptedTermsAndConditions;
 
+  const scrollContainerRef = useRecommendationsScrollPosition(!isLoadingRecommendations && !contextData.isLoading);
+
   return (
     <div className={styles.container} data-testid={testIds.contextPanel.container}>
-      <div className={styles.content}>
+      <div className={styles.content} ref={scrollContainerRef} data-testid={testIds.contextPanel.scrollContainer}>
         <div className={styles.contextSections}>
           {/* User profile bar with learning stats and next action */}
           <UserProfileBar onOpenGuide={openLearningJourney} />
--- a/src/components/docs-panel/context-panel.tsx
+++ b/src/components/docs-panel/context-panel.tsx
@@ -1081,6 +1081,7 @@ function ContextPanelRenderer({ model }: SceneComponentProps<ContextPanel>) {
   const {
     contextData,
     isLoadingRecommendations,
+    hasFetchedRecommendations,
     otherDocsExpanded,
     openLearningJourney,
     openDocsPage,
@@ -1090,7 +1091,11 @@ function ContextPanelRenderer({ model }: SceneComponentProps<ContextPanel>) {
     onOpenLearningJourney: model.state.onOpenLearningJourney,
     onOpenDocsPage: model.state.onOpenDocsPage,
   });
-  const { guides: customGuides, isLoading: isLoadingCustomGuides } = usePublishedGuides();
+  const {
+    guides: customGuides,
+    isLoading: isLoadingCustomGuides,
+    hasLoaded: hasLoadedCustomGuides,
+  } = usePublishedGuides();
   const [customGuidesExpanded, setCustomGuidesExpanded] = useState(true);
   const [suggestedGuidesExpanded, setSuggestedGuidesExpanded] = useState(true);
 
@@ -1109,7 +1114,9 @@ function ContextPanelRenderer({ model }: SceneComponentProps<ContextPanel>) {
     recommendations.length > 0 &&
     !configWithDefaults.acceptedTermsAndConditions;
 
-  const scrollContainerRef = useRecommendationsScrollPosition(!isLoadingRecommendations && !contextData.isLoading);
+  const recommendationsScrollReady =
+    !contextData.isLoading && !isLoadingRecommendations && hasFetchedRecommendations && hasLoadedCustomGuides;
+  const scrollContainerRef = useRecommendationsScrollPosition(recommendationsScrollReady);
 
   return (
     <div className={styles.container} data-testid={testIds.contextPanel.container}>
--- a/src/components/docs-panel/hooks/useRecommendationsScrollPosition.test.tsx
+++ b/src/components/docs-panel/hooks/useRecommendationsScrollPosition.test.tsx
@@ -1,0 +1,78 @@
+import React from 'react';
+import { render, act } from '@testing-library/react';
+import { useRecommendationsScrollPosition } from './useRecommendationsScrollPosition';
+import { StorageKeys } from '../../../lib/storage-keys';
+
+const flushAnimationFrame = () => act(() => new Promise((resolve) => requestAnimationFrame(resolve)));
+
+function TestContainer({ isReady }: { isReady: boolean }) {
+  const ref = useRecommendationsScrollPosition(isReady);
+  return (
+    <div ref={ref} data-testid="scroll-container" style={{ height: '100px', overflow: 'auto' }}>
+      <div style={{ height: '2000px' }} />
+    </div>
+  );
+}
+
+describe('useRecommendationsScrollPosition', () => {
+  beforeEach(() => {
+    sessionStorage.clear();
+  });
+
+  it('restores the saved scroll position once content is ready', async () => {
+    sessionStorage.setItem(StorageKeys.RECOMMENDATIONS_SCROLL_POSITION, '450');
+
+    const { getByTestId } = render(<TestContainer isReady={true} />);
+    const container = getByTestId('scroll-container') as HTMLDivElement;
+    Object.defineProperty(container, 'scrollTop', { value: 0, writable: true });
+
+    await flushAnimationFrame();
+
+    expect(container.scrollTop).toBe(450);
+  });
+
+  it('does not restore before content is ready', () => {
+    sessionStorage.setItem(StorageKeys.RECOMMENDATIONS_SCROLL_POSITION, '450');
+
+    const { getByTestId, rerender } = render(<TestContainer isReady={false} />);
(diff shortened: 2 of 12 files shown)


src/components/docs-panel/hooks/useRecommendationsScrollPosition.ts
```diff
@@ -0,0 +1,64 @@
...
+      try {
+        sessionStorage.setItem(StorageKeys.RECOMMENDATIONS_SCROLL_POSITION, String(container.scrollTop));
+      } catch {
+        // Ignore storage errors (quota exceeded, private browsing, etc.)
+      }
+    };
+
+    container.addEventListener('scroll', handleScroll, { passive: true });
+    return () => {
+      handleScroll();
```
Jayclifford345: **[blocking][react]** This unmount-time save runs after React has already removed the node from the document — passive-effect cleanup on unmount fires post-removal, and a detached element has no box, so `container.scrollTop` reads `0` (CSSOM). Net effect: every navigation away from the Recommended tab overwrites the value the scroll listener just saved.

Verified empirically (React 18 + Chromium, same two-effect shape): scroll listener saves `450` → unmount → cleanup observes `{ isConnected: false, scrollTop: 0 }` and writes `'0'`. The narrower races land in the same place (unmount while `isReady` is still false, or between the restore effect and its pending rAF — cleanup runs before the next frame).

Suggested fix: drop `handleScroll()` from the cleanup entirely — the passive scroll listener already persists every scroll, so the unmount save contributes only the clobber. (`useScrollPositionPreservation` guards its cleanup save with `lastContentUrlRef` for the same reason.) A regression test that fails today: seed `'450'`, mount with `isReady={false}`, unmount, assert sessionStorage still reads `'450'`. Note jsdom can't reproduce the detached read (no layout), which is why the current suite stays green despite this.

src/components/docs-panel/context-panel.tsx
```diff
@@ -1108,9 +1109,11 @@ function ContextPanelRenderer({ model }: SceneComponentProps<ContextPanel>) {
     recommendations.length > 0 &&
     !configWithDefaults.acceptedTermsAndConditions;
 
+  const scrollContainerRef = useRecommendationsScrollPosition(!isLoadingRecommendations && !contextData.isLoading);
```
Jayclifford345: **[blocking]** `isReady` flips true one commit before the recommendations fetch starts, and the hook's one-shot guard burns on that commit:

1. Mount: `contextData.isLoading = true`, `isLoadingRecommendations = false` → not ready.
2. `getContextData()` resolves → a single `setContextData({ …, isLoading: false, recommendations: [] })` → that commit renders with **both flags false**: the fetch hasn't started, because `setIsLoadingRecommendations(true)` only runs inside the post-commit effect (`context.hook.ts:243` → `fetchRecommendations` → `:91`).
3. The restore effect observes that committed `isReady === true`, latches `hasRestoredRef`, and its rAF fires a frame later against UserProfileBar + the single-bar skeleton — the container isn't scrollable yet, so the offset clamps to ~0.
4. The genuine ready transition after the fetch finds `hasRestoredRef` already true and never retries.

This is structural, not a race (a network fetch can't beat one frame), so the restore essentially never lands at real list height. Suggested fix: gate on the fetch actually having completed — e.g. expose a `hasFetchedRecommendations` (set in `fetchRecommendations`'s finally) from `useContextPanel` — or don't latch `hasRestoredRef` until the container's `scrollHeight` can accommodate the saved offset.

**[suggestion]** While here: `isLoadingCustomGuides` is excluded from readiness, and `CustomGuidesSection` renders above the list inside the same scroll container, so a late custom-guides load shifts the restored viewport. Folding `!isLoadingCustomGuides` in avoids that.

Jayclifford345 (commented):

Routed per `docs/design/CONCERNS.md`: six concern reviewers ran independently (security, correctness, testing, reversibility, cross-cutting architecture, context-engine); every medium+ finding was then attacked by three independent skeptic agents, plus direct verification — a live React 18 + Chromium experiment for the unmount behavior and mutation tests against the jest suite.

**Net: two confirmed blocking findings, and they conspire — either one alone defeats the feature on its primary flow, so both need fixing for it to work at all:**

1. **Leaving the Recommended tab overwrites the saved offset with `0`** — the unmount-time save reads a detached node (inline comment on the hook cleanup).
2. **Returning burns the one-shot restore on a gap render** before the recommendations fetch starts, so the restore clamps against skeleton-height content and never retries (inline comment on the `isReady` expression).

Given both mechanisms, step 3 of the PR's "How to verify" shouldn't currently pass as described — worth re-running manually with devtools watching the `grafana-pathfinder-app-recommendations-scroll-position` sessionStorage key while navigating away and back.

**What's clean:**
- **Security**: the only new sink is `element.scrollTop = Number(saved)` behind a `Number.isFinite` check, and the browser clamps it — no F1–F6 surface.
- **Reversibility**: fully reversible. Tab-scoped sessionStorage; the raw-string write matches the `SUGGESTIONS` key convention; correctly outside the server-sync migration list and `clear()` sweeps; the locked contract test reverts atomically with the key.
- **Contracts**: storage-keys contract test updated correctly; the new `data-testid` is jest-scoped and doesn't touch the `data-test-*` E2E contract.
- **Architecture**: tier imports legal; the split from `useScrollPositionPreservation` is principled (different container and lifecycle); nothing else scrolls this container programmatically; dual-mount key contention isn't reachable (panel-mode closes the sidebar on floating/fullscreen).
- **Doc drift**: none.

**Tech debt (non-blocking)**: the new hook reimplements the sibling hook's save/restore shape but drops its "only save once tracking is established" cleanup guard — that dropped guard is exactly finding 1. When fixing, consider whether the shared mechanics of the two hooks deserve one parameterized helper.

Tech debt scan and doc-drift check otherwise clean. Coverage confidence: high — this area is well modeled by the concern registry.

src/components/docs-panel/hooks/useRecommendationsScrollPosition.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  it('does not restore before content is ready', () => {
+    sessionStorage.setItem(StorageKeys.RECOMMENDATIONS_SCROLL_POSITION, '450');
+
+    const { getByTestId, rerender } = render(<TestContainer isReady={false} />);
+    const container = getByTestId('scroll-container') as HTMLDivElement;
+    Object.defineProperty(container, 'scrollTop', { value: 0, writable: true });
+
+    rerender(<TestContainer isReady={false} />);
+    expect(container.scrollTop).toBe(0);
```
Jayclifford345, line 55: **[suggestion]** This test can't fail: the hook's restore happens inside a rAF and there's no `await flushAnimationFrame()` before this assertion, so a scheduled-but-unfired restore also leaves `scrollTop` at 0. Mutation-tested to confirm: deleting `!isReady` from the hook's guard leaves all 5 tests green; adding the flush before this assertion makes the test correctly fail against that mutant. One-line fix: `await flushAnimationFrame()` (keeping `isReady={false}`).

moxious:
Post review notes:  comment from slack:

> I tried to run through the feature manually and  find at the moment that if I close the tab we jump to block editor rather than the context tab (not an error you introduced just seems to be somthing that has been round a while). Instead i tried to just click back on the context tab but it doesn't seem to activate the scroll functionality

I've amended/expanded the scope of this PR to fix that bug to make the manual testing easier to reach.

Jayclifford345 (approved):

PR #1297, opened 2026-07-10
fix(floating-panel): preserve scroll position across highlight-dodge compact/restore cycles

### LINKED ISSUE #1291: Floating panel: guide scroll resets to top when highlight dodge compact/restore cycles
### Description

When testing a guide in the **floating panel** with **Show me** active on a wide target (e.g. a whole form row), the guide content **scroll position jumps to the top**. Step completion is preserved; the user must manually scroll back to the current step.

Observed on live testing of query-variable modal steps (**Refresh**, **Preview**) while the guide is popped out. The orange highlight box covers a large area; the floating panel overlaps the highlight.

### Environment

- Floating panel (undocked guide)
- Interactive step with `action: "highlight"` and **Show me**
- Wide `reftarget` (e.g. `[data-testid='data-testid Variable editor Form Query Refresh select']` wrapping an entire field row)
- Grafana modal open (Query Variable editor)

### Steps to reproduce

1. Open a guide in the **floating panel** (pop out).
2. Scroll down to a later step (e.g. Refresh or Preview in the query variable modal flow).
3. Click **Show me** on a step whose highlight target is a **wide container** (whole Refresh row or similar).
4. Observe the large `.interactive-highlight-outline` overlapping the floating panel.
5. Either:
   - The panel **repositions** (highlight dodge), or
   - The panel enters **compact** mode (`height: auto`) when dodge cannot find clearance.
6. Interact with the highlighted Grafana control (e.g. select **On dashboard load**, or click **Preview**).
7. When the highlight clears or updates, the floating panel restores to **full** mode.

**Actual:** Guide scroll inside the floating panel resets to the **top**.

**Expected:** Scroll position is preserved across dodge / compact ↔ full transitions so the user stays on the current step.

### Root cause (suspected)

`useHighlightDodge` watches `.interactive-highlight-outline` and `.interactive-comment-box` via `MutationObserver`. When the panel overlaps a highlight:

- Dispatches `pathfinder-floating-dodge` (reposition), or
- Dispatches `pathfinder-floating-compact` → `FloatingPanel` sets `panelState = 'compact'` and `height: 'auto'`

When highlights are removed, it dispatches `pathfinder-floating-restore-full` and `pathfinder-floating-restore-position`.

The floating panel content area (`.content` with `overflow: auto`) likely **loses `scrollTop`** when height collapses to `auto` in compact mode and expands again on restore-full. There is no scroll preservation hook for floating panel content (unlike sidebar `useScrollPositionPreservation` on `#inner-docs-content`).

**Related code:**

- `src/components/floating-panel/useHighlightDodge.ts`
- `src/components/floating-panel/FloatingPanel.tsx` — `compact` / `full` state, `pathfinder-floating-compact`, `pathfinder-floating-restore-full`
- `src/components/floating-panel/floating-panel.styles.ts` — scrollable `.content` area

### Impact

| Audience | Affected? |
|----------|-----------|
| Learners using floating panel + Show me on wide targets | **Yes** — disorienting scroll jump |
| Authors testing guides in floating panel | **Yes** — must scroll back after Refresh/Preview |
| Docked sidebar only | **Less likely** — dodge logic is floating-specific |

### Suggested fix

1. **Preserve scroll position** on the floating panel `.content` container across `compact` ↔ `full` transitions (save `scrollTop` before compact, restore after restore-full).
2. Optionally **avoid compact** when only repositioning would suffice, or shrink highlight bounds for radio/toggle groups.
3. Consider **scroll preservation** in `FloatingPanelContent` similar to sidebar `useScrollPositionPreservation`.

### Workaround

Skip **Show me** on wide modal targets; read the instruction and act manually. Step progress is not lost.

### Related

- May compound with #1290 (Block Editor Preview → Edit on pop out/dock) during author testing, but this issue is distinct: scroll reset during highlight dodge while already in floating mode.

moxious:
## Summary

When `useHighlightDodge` compacts the floating panel to dodge a wide `.interactive-highlight-outline` (or restores it), the guide's scroll position resets to the top. Root cause: `.content` is `flex: 1; overflow: auto`; when the panel's own inline `height` becomes `'auto'` in compact mode, that flex item's height collapses to fit its content, leaving nothing to scroll — `scrollTop` gets forced to `0` and stays there once the fixed height (and scroll container) return on restore-full. This saves `.content`'s `scrollTop` right before compacting and restores it, via `requestAnimationFrame` after the fixed height comes back, on `pathfinder-floating-restore-full`.

## How to verify

1. Open a guide with a step that highlights a wide target (e.g. a query-variable modal's Refresh/Preview row) in the floating panel.
2. Scroll the guide content down to a later step, then trigger **Show me** on that step so the panel dodges/compacts.
3. Interact with the highlighted control so the highlight clears and the panel restores to full.
4. Confirm the guide content is still scrolled to the same step instead of jumping back to the top.
5. (Regression) `src/components/floating-panel/FloatingPanel.test.tsx` dispatches the `pathfinder-floating-compact` / `pathfinder-floating-restore-full` events directly and asserts `.content`'s `scrollTop` round-trips, including a second cycle starting from `scrollTop: 0` to guard against resurrecting a stale saved position.

## Breaking changes

None. This change is additive and fully reversible — it only affects a ref-based scrollTop save/restore local to `FloatingPanel`.

Fixes #1291


moxious:
Follow-up issue filed from the repeated review findings on this PR: #1298.

It captures the underlying pattern we found while fixing #1291 here: floating panel dodge/restore is an implicit state machine spread across `FloatingPanel`, `useHighlightDodge`, `useDragResize`, document events, and browser-owned layout state like `scrollTop`. This PR keeps the fix scoped to the immediate scroll race, while #1298 tracks the bounded refactor to model dodge sessions explicitly.

src/components/floating-panel/FloatingPanel.tsx
```diff
@@ -70,13 +74,31 @@ export function FloatingPanel({
...
+        return;
+      }
+      savedScrollTopRef.current = null;
+      if (contentRef.current) {
+        contentRef.current.scrollTop = savedScrollTop;
+      }
+    });
+  }, []);
+
   const handleMinimize = useCallback(() => {
```
Jayclifford345, line 94: **[suggestion]** `handleMinimize` is the only panel-state transition that doesn't participate in the save/restore protocol, which leaves one narrow losing race: compact saves `800` → highlight clears, restore-full schedules the deferred write → user minimizes within the ~2-frame window → the `.then` fires with its token still valid, nulls `savedScrollTopRef`, and the `scrollTop` write lands on a `display:none` element — a spec no-op (verified in Chromium: writes while hidden are silently dropped). The pill restore then finds `saved === null` and the position is gone; the engine's own retained offset is the compact-clamped `0`, so it doesn't rescue it. Same race exists via pill-restore → re-minimize within two frames.

One-line fix mirroring `handleCompact`: bump `restoreTokenRef.current++` in `handleMinimize`, so the in-flight write is cancelled and the saved value stays armed for the next `handleRestore` (harmless when nothing is saved). Test 5 covers minimize *before* restore-full but not minimize *after* it while the write is in flight.

For what it's worth, the plain full → minimize → restore path (no compact) needs no handling at all — Chromium retains the scroll offset across `display:none` and restores it on re-show, verified empirically.

src/components/floating-panel/FloatingPanel.test.tsx
```diff
@@ -0,0 +1,155 @@
...
+    content.scrollTop = scrollTopDuringCompact;
+
+    await act(async () => {
+      document.dispatchEvent(new CustomEvent('pathfinder-floating-restore-full'));
+      await flushReactUpdates();
+    });
+    expect(screen.getByRole('dialog')).toHaveAttribute('data-panel-state', 'full');
+  }
+
+  it('restores scrollTop after a compact -> restore-full cycle', async () => {
```
Jayclifford345, line 54: **[suggestion]** Three small additions would pin the remaining semantics (the existing five are strong — each kills a distinct mutant):

1. **Restore-full with no prior compact is a scroll no-op** — the most common production path through the new code (highlights clearing while the panel is already full). Set `scrollTop = 500`, dispatch `pathfinder-floating-restore-full`, flush, expect `500`. Nothing currently exercises the `savedScrollTop === null` early return.
2. **Token-check-before-null ordering** — a variant of the interrupted-restore test with `await flushReactUpdates()` *between* the second compact and the second restore-full, expect `800`. As written, both restores capture `800` at invocation time, so a mutant that nulls the ref before checking the token still passes; real dodge cycles are separated by frames, making this the realistic timing.
3. **Unmount with a pending deferred restore** — compact, restore-full, `unmount()`, flush; assert no error. The `if (contentRef.current)` guard is the only thing preventing a TypeError inside a floating promise, and no test fails if it's removed.

Jayclifford345 (commented):

Same process as #1296: routed per `docs/design/CONCERNS.md` — five always-on reviewers plus a dedicated floating-panel/dodge-event-contract reviewer (this subsystem has no concern entry yet, see the coverage note), tech-debt scan, doc-drift check, and empirical browser verification of the timing claims.

**Verdict: solid.** The token/save-if-null state machine holds up under every event sequence the dodge dispatcher can actually emit (traced from `useHighlightDodge`, not assumed), the save point reads the pre-collapse `scrollTop`, no CSS transition touches height so the double-rAF write can't clamp, and the test suite is genuinely strong — each of the five tests kills a distinct mutant of the state machine. No blocking findings; two `[suggestion]` inline comments (a narrow minimize race, and three small test additions).

**Empirically verified along the way (Chromium):**
- Scroll offset **survives** a `display:none` round-trip natively (`800 → hidden → 800` after re-show) — so the plain full → minimize → restore path needs no save/restore of its own; one reviewer's concern there was refuted by the experiment.
- A `scrollTop` **write** while `display:none` is a silent no-op — which is what makes the minimize-during-pending-restore race in the inline comment real.

**What's clean:** security (handlers consume no event payload; nothing a same-origin dispatcher couldn't already do), reversibility (in-memory refs only; plain revert restores everything; `data-panel-state` pre-exists on main and is in no E2E contract), the `bugfix-patterns.md` row (column-consistent, accurate, correctly dated), comment hygiene (the new comments document hidden invariants — keep-list), tier imports, doc drift (none).

**Non-blocking follow-ups (not this PR):**
- **Coverage note**: `src/components/floating-panel/**` appears in no `CONCERNS.md` trigger paths, so this growing subsystem (dodge events, panel-mode, MinimizedPill, drag/resize) is reviewed only by always-on catch-alls. Its core risk — the implicit `pathfinder-floating-*` event ordering contract — is exactly what a subsystem entry would model. Worth a `companion-mode` concern row or an adjacent concern's trigger-path expansion.
- **Tech debt (A2, high confidence)**: `FloatingPanel` is now ~305 lines with 20 hook calls owning five concerns; the new scroll-preservation machine is a natural extraction into a sibling hook (e.g. `useCompactScrollPreservation`), matching the directory's `useDragResize`/`useHighlightDodge` pattern — and it would give the minimize path from the inline comment a single owner.

Jayclifford345 (approved):
No true blockers LGTM

moxious:
Thanks for the approval; as I'm trying to do something specific here, I've carried over the suggestions and the other items to the follow-on ticket already on this PR.  I don't want to expand the scope of this PR but those suggestions should get swept into the fix for the other.

PR #1300, opened 2026-07-10
refactor(floating-panel): model dodge sessions explicitly

### LINKED ISSUE #1298: Model floating panel dodge sessions explicitly
## Summary

Floating panel dodge/restore behavior has become an implicit state machine spread across `FloatingPanel`, `useHighlightDodge`, `useDragResize`, browser layout state, and document-level custom events. Recent fixes around highlight dodging, compact mode, scroll preservation, and restore ordering have repeatedly exposed edge cases where one part of the panel restores correctly while another part loses its authoritative state.

We should introduce a small explicit "dodge session" model, likely as `useFloatingPanelDodgeSession` or a reducer, to own the transactional state for compact/restore cycles.

## How we discovered this

This came out of repeated review rounds on PR #1297 while fixing #1291.

PR #1297 initially fixed the obvious scroll reset when a wide highlight forced the floating panel into compact mode. Follow-up review then found additional blocking interleavings:

- compact saved the current guide scroll position
- restore scheduled an async scroll write after React/layout settled
- another compact could happen before that write landed
- the browser-clamped compact-mode `scrollTop` value could become authoritative
- the final restore could still land at the top of the guide

The PR now has a focused fix and regression coverage for that immediate race, but the review made clear that the underlying behavior is a state machine that is only implied by refs, event ordering, and browser layout side effects.

## Context

In #1291, the floating panel guide scroll reset to the top when highlight dodge compacted and restored the panel.

The root cause was subtle:

- `useHighlightDodge` emits `pathfinder-floating-compact` when a wide highlight/modal leaves no room for the full panel.
- `FloatingPanel` switches to `compact`, setting panel height to `auto`.
- The `.content` element is `flex: 1; overflow: auto`.
- In compact mode, browser layout can collapse the scroll container, forcing `scrollTop` to `0`.
- That `0` is not user intent; it is a layout artifact.
- Restore happens asynchronously after React/layout settles.
- If another compact/restore cycle interleaves before the deferred restore lands, stale or clamped scroll values can become authoritative.

The PR #1297 fix preserves the saved scroll value until the deferred restore actually lands, guarded by a restore token. That is appropriate for the immediate bug, but it is still a local patch over an implicit lifecycle.

## Problem

Floating mode currently has several independent sources of truth:

- `FloatingPanel` owns `panelState`: `full | compact | minimized`.
- `useHighlightDodge` observes DOM highlights/modals and emits document events.
- `useDragResize` owns geometry and persisted user position/size.
- The browser owns layout-derived state like `scrollTop`.
- The interactive engine owns highlight lifecycle.
- `FloatingPanelManager` owns surface handoff and guide content lifecycle.

These pieces communicate through events such as:

- `pathfinder-floating-compact`
- `pathfinder-floating-restore-full`
- `pathfinder-floating-restore-position`
- `pathfinder-floating-manual-move`

Together these form a state machine, but the code does not model it as one. There is no single state object for a dodge cycle, no transition table, and no explicit contract for which values are authoritative during compact/restore.

That makes bugs likely in transitions such as:

- compact -> restore-full
- compact -> restore-full scheduled -> compact again
- compact -> minimize -> restore from pill
- dodge -> manual drag -> obstacle clears
- obstacle changes while restore is pending
- restore position succeeds but scroll restore is stale, or vice versa

## Proposal

Add a bounded dodge-session abstraction rather than a full floating-panel rewrite.

Possible shape:

```ts
type FloatingPanelView = 'full' | 'compact' | 'minimized';

type DodgeSession = {
  id: number;
  savedScrollTop: number | null;
  previousPosition: { x: number; y: number } | null;
  restoreScroll: 'idle' | 'scheduled';
};
```

This could live behind a hook or reducer, for example:

```ts
useFloatingPanelDodgeSession({
  contentRef,
  setPanelState,
  setPosition,
});
```

or:

```ts
floatingPanelDodgeReducer(state, action);
```

The key contract should be:

> During compact/dodge, browser layout-derived values like `scrollTop` are not authoritative unless the dodge session explicitly marks them valid.

## Scope

This should focus only on dodge/compact/restore transactional state.

Include:

- Saving scroll before compact.
- Preserving saved scroll while compacted.
- Deferring scroll restore until layout settles.
- Invalidating stale deferred restores.
- Preserving or replacing previous panel position according to manual-drag behavior.
- Handling minimize/restore while a dodge session is active.
- Consolidating the custom event handlers in `FloatingPanel` around this model.

Do not include:

- Rewriting `FloatingPanelManager`.
- Redesigning sidebar/fullscreen/floating handoff.
- Reworking tab persistence.
- Replacing `useDragResize`.
- Changing visual behavior unless required to preserve existing semantics.

## Acceptance criteria

- Dodge compact/restore behavior is represented by one explicit session object or reducer state.
- The code defines which state is authoritative during compact mode.
- Existing PR #1297 regressions remain covered:
  - scroll restores after compact -> restore-full
  - second cycle from scrollTop `0` does not resurrect stale scroll
  - stale deferred restore does not write while compact
  - interrupted restore followed by another restore preserves the original scroll
  - minimize during compact still restores saved scroll
- Manual drag during dodge still updates the restore target and does not teleport to stale position.
- Tests cover the main transition table, not just the original bug.
- No broad floating-panel rewrite is required.

## Why this is the right next step

A full floating-panel controller would probably be too large and risky right now. The broader floating mode includes geometry persistence, docking, fullscreen handoff, content restoration, guide lifecycle, and interactive engine coordination.

But the recurring bug class is narrower: temporary dodge sessions with rollback. Modeling that bounded lifecycle should remove the sharpest implicit state without turning this into a large component rewrite.

The expected size is roughly:

- 100-180 production LOC
- 80-150 test LOC

That is large enough to make the contract explicit, but small enough to review safely.

moxious:
## Summary

Models the floating panel's highlight-dodge compact/restore lifecycle as an explicit state machine instead of an implicit combination of refs, custom-event ordering, and browser layout side effects. A pure reducer ([dodge-session.ts](https://github.com/grafana/grafana-pathfinder-app/blob/9ebf9a5d983b44a117b2fbf4d97615a4e650c43e/src/components/floating-panel/dodge-session.ts)) now owns the transactional session state (view, saved scroll, restore phase, staleness token), and a thin hook adapter ([useDodgeSession.ts](https://github.com/grafana/grafana-pathfinder-app/blob/9ebf9a5d983b44a117b2fbf4d97615a4e650c43e/src/components/floating-panel/useDodgeSession.ts)) wires it to the DOM. One intentional behavioral change rides along: minimizing now invalidates a scheduled scroll restore, fixing a latent race where the deferred write could land against a `display:none` container.

## Why

PR #1297 (fixing #1291) went through repeated review rounds because each fix exposed another interleaving: compact saved scroll, restore scheduled an async write, another compact landed first, and the browser-clamped `scrollTop` of a collapsed container became authoritative. The behavior was a state machine that no code modeled as one. Issue #1298 asks for a bounded "dodge session" abstraction with the contract: **during compact/dodge, layout-derived values like `scrollTop` are not authoritative unless the session explicitly marks them valid** — without a broader floating-panel rewrite.

A full floating-panel controller was rejected in the issue as too large and risky; this PR deliberately keeps geometry persistence, mode handoff, and guide lifecycle untouched.

## What to look at

- `src/components/floating-panel/dodge-session.ts` — the transition table. The load-bearing rules: `COMPACT` never overwrites a non-null `savedScrollTop` with a measured (layout-derived) read, and `SCROLL_RESTORE_LANDED` only commits when its token is still current. Every other action that could stale a pending write bumps the token.
- `src/components/floating-panel/useDodgeSession.ts` — the DOM adapter. Note the synchronous dispatch wrapper (reducer applied to a ref before `setState`): it looks redundant but is required so the restore-full handler knows immediately whether a write was scheduled and under which token. It must only be called from event/promise callbacks, never during render.
- `src/components/floating-panel/FloatingPanel.tsx` — shrinks 305 → 207 lines; now renders purely from session state. The dodge flash/analytics timers moved verbatim into the hook. The pre-dodge *position* restore target intentionally stays in `useHighlightDodge` — its existing tests pin that the detector alone owns manual-drag retargeting, and duplicating position into the session would recreate the split-brain this PR removes.
- `src/lib/event-names.ts` — the five `pathfinder-floating-*` event names centralized as `FloatingPanelEvents`. Test files deliberately keep raw string literals so a typo in the constants fails CI.

## How to verify

CI covers the transition table (`dodge-session.test.ts`), the new interleaving tests (`FloatingPanel.dodge-session.test.tsx`), and the unmodified #1297 regressions. Manual verification in floating mode (`npm run server`, open a guide, switch to floating panel):

1. Scroll partway down a guide, then run an interactive step whose highlight covers most of the viewport — the panel compacts and moves. Dismiss the highlight: the panel returns to its previous position with the scroll position intact.
2. While a wide highlight is active (panel dodged), drag the panel somewhere else. Dismiss the highlight: the panel stays where you put it (no teleport to the pre-dodge spot).
3. Scroll down, trigger a compact, then press Escape to minimize before the panel restores. Reopen from the pill: the guide is at the saved scroll position.
4. Trigger several rapid compact/restore cycles (open/close a modal repeatedly): the guide scroll never resets to the top.

## Breaking changes

None. Event names, `data-panel-state` values, and the DOM structure are byte-compatible; all existing tests pass unmodified.

## Out of scope

- `FloatingPanelManager`, sidebar/fullscreen/floating handoff, and tab persistence — excluded by #1298.
- `useDragResize` logic and geometry persistence — unchanged (only the event-name literal was swapped for the constant).
- `pathfinder-modal-state-changed` — owned by the interactive engine, not part of the dodge session surface.

Closes #1298


Jayclifford345 (approved):

Jayclifford345 (commented):

Same harness as #1296/#1297: seven reviewers this time (five always-on + the dodge-contract lens + an ai-subsystem reviewer for the `CONCERNS.md` row), adversarial skeptic verification on medium+ findings, tech-debt scan, doc-drift check.

**Verdict: strong PR — no blocking findings.** The things that matter were verified, not assumed:

- **The #1297 minimize race is genuinely fixed.** An independent walk of the full event grammar (from `useHighlightDodge`'s actual dispatch code) confirmed `MINIMIZE`'s token bump closes the old loss path, and found no new stale-restore, save-loss, stuck-save, or view-divergence sequence. Pill-restore-while-highlight-active re-compacts with exact parity to main; the only residual hazard (a straggler rAF `checkAndDodge` in the one-frame window after minimize) is byte-for-byte parity with main, not introduced here.
- **The reducer's subtlest choice is sound**: `RESTORE_FULL` not bumping the token when `savedScrollTop === null` is safe because a null save implies the current-token write already resolved — no live write can exist.
- **The `sessionRef` → `setView` mirror was adjudicated against this repo's parallel-state ban and passes**: single-writer choke point (`setView` is called exactly once, inside `dispatch`), and plain `useReducer` genuinely can't replace it since dispatch doesn't return the next state, which `restoreFull` needs synchronously for the token. Suggested tripwire: if the session ever grows a second render-relevant field or `dispatch` is ever exposed outside the hook, migrate to `useSyncExternalStore` rather than adding a second mirror.
- **Refactor hygiene checks out**: old implementation fully deleted (no orphaned refs/handlers), `dodge-session.ts` verified import-free pure, `copyTimer` cleanup actually improved, event-name strings byte-identical to main's literals, old #1297 suite unmodified and green (41/41 across the floating-panel suites).
- Security, reversibility (plain revert, no residue), and doc-drift all clean.

**Adversarial-verification disclosure:** the `CONCERNS.md` inbound-coupling finding (inline comment) survived 2:1 but was materially narrowed by the skeptics: `.interactive-comment-box` is already owned end-to-end by the E2E contract + `comment-box.contract.test.ts`, and `getVisibleModalRects` is a typed import that fails typecheck on rename. The surviving kernel is exactly two unowned string couplings: `interactive-highlight-outline` and `pathfinder-modal-state-changed`.

**Tech debt:** the A2 (god component) flagged on #1297 is **resolved** — `FloatingPanel.tsx` is down to 207 lines / 12 hooks, below the two-criterion threshold. One minor C2 remains: `handleSwitchToSidebar` and `handleRestoreFull` are zero-arg pass-throughs that could be inlined (`restoreFull` can be registered directly as the listener). Non-blocking.

docs/design/CONCERNS.md
```diff
@@ -63,6 +63,7 @@ Column key — **cat**: `AO`=always-on · `sub`=subsystem · `xcut`=cross-cuttin
...
+| `floating-panel`                 | sub  | N   | strong | 2   | 8   | `src/components/floating-panel/**`, `src/global-state/panel-mode.ts`, `src/constants/floating-panel.ts`, `src/lib/event-names.ts`                                                                                                                                                                                                                                                                                   | `pathfinder-floating-`, `FloatingPanelEvents`, `dodgeSessionReducer`, `useDodgeSession`, `useHighlightDodge`, `useDragResize`, `panelModeManager`, `data-panel-state`, `MinimizedPill`, `savedScrollTop`, `restoreToken`, `FloatingPanelGeometry`                                                                                                                                                 |
```
Jayclifford345, line 66: **[suggestion]** This row models the *outbound* contract well (event ordering, token staling, `data-panel-state`) but leaves the two *inbound* couplings dodging depends on unowned — verified by adversarial cross-check:

- `interactive-highlight-outline` (`HIGHLIGHT_SELECTOR`, `useHighlightDodge.ts:11`; produced in `navigation-manager.ts` / the overlay styles) — appears in no concern's keywords, and a producer-side rename PR updates the producer's own tests in the same PR while `useHighlightDodge.test.ts` fabricates its own elements with the stale literal, so CI stays green and dodging silently dies.
- `pathfinder-modal-state-changed` (raw literal on both the `modal-watcher.ts` dispatch side and the `useHighlightDodge.ts:213` listen side) — same silent-break shape.

(For completeness: `.interactive-comment-box` is already owned by the E2E contract + `comment-box.contract.test.ts`, and `getVisibleModalRects` fails typecheck on rename — so those two need nothing.)

Suggested row edits: add `interactive-highlight-outline` and `pathfinder-modal-state-changed` to trigger_keywords, plus a review question like "Does this change the highlight class names or the modal-state event the dodge observer watches?". Two smaller cell tweaks while here: `load_docs` currently points at generic rules that barely mention floating-panel — `docs/developer/bugfix-patterns.md` (the #1291 row) is the canonical statement of the scroll invariant; and `load_code` omits `src/constants/floating-panel.ts` even though it's in trigger_paths and defines `FloatingPanelGeometry` + the dodge margins.

src/components/floating-panel/useHighlightDodge.ts
```diff
@@ -204,19 +205,19 @@ export function useHighlightDodge(geometry: FloatingPanelGeometry, isMinimized:
     // discarding the user's placement.
     const handleManualMove = (e: Event) => {
       if (previousPositionRef.current) {
-        previousPositionRef.current = (e as CustomEvent<{ x: number; y: number }>).detail;
+        previousPositionRef.current = (e as CustomEvent<FloatingPanelMoveDetail>).detail;
       }
     };
 
     observer.observe(document.body, { childList: true, subtree: true });
     document.addEventListener('pathfinder-modal-state-changed', handleModalStateChange);
```
Jayclifford345, line 213: **[suggestion]** The centralization stops one event short in the effect it refactored: the four `pathfinder-floating-*` names now come from `FloatingPanelEvents`, but `'pathfinder-modal-state-changed'` here (and on the dispatch side in `modal-watcher.ts:44`) stays a raw literal — and it's the only event in this file whose producer and consumer live in different subsystems, i.e. the one the registry helps most. I saw the "Out of scope" note in the description, so treat this as a follow-up nomination rather than a request for this PR: add it to `event-names.ts` (e.g. `InteractiveEngineEvents`) and import it on both sides. Same seam, same follow-up: `HIGHLIGHT_SELECTOR` and the `MutationObserver`'s `classList.contains` filter duplicate the overlay class names within this file (plus the producer side) — deriving both from one shared array would make "add an overlay class" a one-place change instead of a silent dodge-disabler.

src/components/floating-panel/FloatingPanel.dodge-session.test.tsx
```diff
@@ -0,0 +1,182 @@
...
+      act(() => {
+        jest.advanceTimersByTime(1000);
+      });
+      expect(dialog).toHaveStyle({ left: '80px', top: '90px' });
+      expect(dialog.className).toBe(restingClassName);
+    } finally {
+      jest.useRealTimers();
+    }
+  });
+});
```
Jayclifford345, line 182: **[suggestion]** Three small additions would close the remaining mutant escapes (the suites are otherwise strong — 41/41, and the token/interleaving semantics are well pinned):

1. **Unmount with a pending deferred restore** (carried over from the #1297 review — still uncovered): compact, restore-full, `unmount()` before flushing frames, then flush; assert no error. Removing the `if (contentRef.current)` guard in `useDodgeSession.ts` currently fails no test, and a regression surfaces as an unhandled rejection in production.
2. **Hook-level restore-full with no prior compact**: the no-op path is only pinned at the reducer level. Surviving mutant: replace the early return with `contentRef.current.scrollTop = next.savedScrollTop ?? 0` — compiles, all 41 tests pass, but a bare restore-full event (routine when obstacles clear without a compact) would stomp the user's scroll to 0. Set scrollTop, dispatch restore-full alone, flush, assert unchanged.
3. **`savedScrollTop === 0` is valid session state**: mutating `state.savedScrollTop ?? action.measuredScrollTop` to `||` in `COMPACT`/`MINIMIZE` survives the suite because a saved value of exactly 0 is never exercised — and that mutant reintroduces precisely the layout-derived-overwrite class this reducer exists to prevent. One reducer case pins it: compact at 0, restore, compact again with a nonzero measured value, assert saved stays 0.

PR #1302, opened 2026-07-10
fix(block-editor): reject invalid regex patterns in InputBlockForm

### LINKED ISSUE #822: InputBlockForm: validation pattern field accepts invalid regex silently

dgehring-grafana:
## Summary

The input block authoring form allowed saving an optional Validation pattern that was not a valid JavaScript regex, which made it easy for guide authors to persist mistakes that only show up later. This change compiles the pattern with `new RegExp` before submit, disables save when it fails for text inputs, and shows an inline error after blur (or immediately when editing a block that already has a bad pattern).

## What to look at

- [InputBlockForm](https://github.com/grafana/grafana-pathfinder-app/blob/602237560667cf61f5affc09c2e7eed8b55e6bbf/src/components/block-editor/forms/InputBlockForm.tsx) — submit/`isValid` gate vs blur-gated error display: invalid patterns block save immediately, but the inline error waits for blur so mid-typing noise is avoided. Validation applies only when `inputType === 'text'` so leftover pattern state does not brick Checkbox/Datasource after a type switch. New unit coverage lives alongside the form; keep or trim if the team prefers thinner form tests.

## Why

Authors could previously submit a broken pattern from the form builder and only discover the problem downstream. A keystroke-level error (matching other fields more closely) was considered, but blur-gated messaging was chosen for this field because regex is often temporarily invalid while typing.

## How to verify

1. Open the block editor, add an Input block, leave type as Text, fill Prompt and Variable name.
2. Enter an invalid pattern such as `[unclosed`, blur the field — confirm "Invalid regex pattern" appears and Add block stays disabled.
3. Switch Input type to Checkbox with that invalid pattern still in state — confirm Add block enables and saving produces a boolean input without a `pattern` field.
4. Edit an existing text input whose stored `pattern` is already invalid — confirm the error shows on open and Update block is disabled until the pattern is fixed or cleared.

## Breaking changes

Fixes #822

### DIFF
--- a/src/components/block-editor/forms/InputBlockForm.test.tsx
+++ b/src/components/block-editor/forms/InputBlockForm.test.tsx
@@ -1,0 +1,118 @@
+import React from 'react';
+import { fireEvent, render, screen } from '@testing-library/react';
+
+import { InputBlockForm } from './InputBlockForm';
+import type { JsonBlock } from '../types';
+
+function renderForm(onSubmit: (block: JsonBlock) => void = jest.fn()) {
+  render(<InputBlockForm onSubmit={onSubmit} onCancel={jest.fn()} />);
+  return {
+    promptInput: screen.getByPlaceholderText('e.g., What is the name of your Prometheus data source?'),
+    variableNameInput: screen.getByPlaceholderText('e.g., prometheusDataSource'),
+    submitButton: screen.getByRole('button', { name: 'Add block' }),
+  };
+}
+
+describe('InputBlockForm', () => {
+  it('keeps submit disabled until required fields are filled', () => {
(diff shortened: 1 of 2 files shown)


moxious (approved):

moxious (approved):
no comments / nothing to change, but waiting on merge, I think darrell should sign the CLA since his account is new.

Issue #1290, opened 2026-07-09
Block Editor: pop out / dock resets viewMode from Preview to Edit


bonnywelsford-source:
### Description

When testing a guide in Block Editor dev mode (`?dev=true`), popping the panel out (sidebar → floating) or docking it back (floating → sidebar) resets the editor from **Preview** to **Edit**. The floating/sidebar surface remounts `BlockEditor` with the default `viewMode: 'edit'`, even though the author was actively testing in Preview.

This makes live guide testing awkward: authors must re-select Preview after every pop out/dock, including when completing popout guide steps.

### Environment

- **Surface:** Block Editor dev mode (`/plugins/grafana-pathfinder-app?dev=true`)
- **Active tab:** Editor tab (`type: 'editor'`)
- **Trigger:** Toolbar Pop out/Dock, or interactive `popout` guide steps (`targetvalue: 'floating'` / `'sidebar'`)

### Steps to reproduce

1. Open Block Editor at `?dev=true`.
2. Load or paste guide JSON.
3. Switch to **Preview** (eyeball icon).
4. Click **Pop out** (toolbar) or complete a guide step with `action: "popout"`, `targetvalue: "floating"`.
5. Observe the floating panel opens in **Edit** mode (block list, Draft, Save as draft) instead of Preview.
6. Click **Dock** to return to sidebar.
7. Observe the sidebar panel is also back in **Edit** mode.

### Expected behavior

Pop out and dock should **preserve the active Block Editor `viewMode`** (Preview, Edit, or JSON) across panel-mode handoffs. If the author was in Preview before pop out, the floating panel should open in Preview.

### Actual behavior

Each handoff remounts `BlockEditor` in the target surface with `viewMode` reset to `'edit'`.

### Root cause

1. **Separate React mounts:** Sidebar (`DocsPanelContentArea`) and floating (`FloatingPanelManager`) each lazy-load their own `BlockEditor` instance when `activeTab.type === 'editor'`.

2. **Default view mode on init:** `useBlockEditor` always initializes with `viewMode: 'edit'`:

   ```ts
   // src/components/block-editor/hooks/useBlockEditor.ts
   viewMode: 'edit' as ViewMode,
   ```

3. **`viewMode` not persisted:** `useBlockPersistence` saves guide JSON and `blockIds` to `pathfinder-block-editor-state`, but not `viewMode`.

[...]

PR #1314, opened 2026-07-13
fix(block-editor): preserve viewMode and JSON drafts across remounts

moxious:
## Intent

Fix all review findings on PR #1314 and push the existing branch. Preserve Edit, Preview, and JSON mode across sidebar/floating remounts without resetting a fresh mode toggle; preserve unapplied valid or invalid JSON drafts and derive validation after restore; reject malformed persisted viewMode values by falling back to Edit. Keep the additive localStorage shape backward compatible, use one restore owner, add deterministic regression coverage, and retain the existing branch and PR.

## What Changed

- Extended block editor persistence (`useBlockPersistence`) to store and restore `viewMode` and `jsonModeState` alongside the guide/blockIds in localStorage, with defensive parsing that rejects malformed persisted values (falling back to `edit` mode) and stays backward compatible with the existing storage shape.
- Consolidated guide restoration in `BlockEditor.tsx` into a single `onLoad` callback (removing the separate mount-time load effect) so `loadGuide` and the new `restoreJsonMode` helper (`useJsonModeHandlers.ts`) run from one owner, restoring Preview/Edit/JSON mode and unapplied JSON drafts with validation re-derived on restore.
- Added a dedicated effect in `useBlockPersistence` to save on `viewMode`/`jsonModeState` changes independent of the debounced guide-content save, since panel pop-out/dock remounts can occur before that debounce completes.
- Added regression coverage across `BlockEditor.persistence.test.tsx`, `useBlockEditor.test.ts`, `useBlockPersistence.test.ts`, and `useJsonModeHandlers.test.ts`, plus doc updates noting the new view-state persistence behavior.

## Risk Assessment

✅ Low: The change is a well-bounded, additive persistence fix: it consolidates to a single restore owner, extends the stored shape backward-compatibly (optional viewMode/jsonModeState fields), validates/falls back malformed persisted viewMode to 'edit', and adds an immediate-save effect whose dependency chain (save() recreated when viewMode/jsonModeState change, which cascades into canceling any stale pending debounce timeout) correctly avoids clobbering fresh state with stale data; new tests cover the restore and immediate-save paths.

## Testing

Ran the fix's full unit-test suite (35 targeted tests + 529 block-editor tests) against the target commit — all pass — and proved the new tests are non-vacuous by rerunning them unchanged against the pre-fix base commit, where 9 failed exactly as expected. Additionally wrote and ran (then removed) a manual verification test that mounts the real BlockEditor, switches to Preview/JSON mode, and fully unmounts/remounts to reproduce the actual pop-out/dock user scenario from issue #1290, confirming the selected mode and its content survive the remount; a pixel-level browser screenshot was skipped as disproportionate since this is a pure state/localStorage fix with no visual change, and the component-remount reproduction is the closest available end-user-scenario evidence.

<details>
<summary>Evidence: Manual pop-out/dock remount verification output</summary>

<code>[manual-verify] Preview button class after remount: css-xf7q69-button&#10;[manual-verify] Edit button class after remount: css-g0dahz-button&#10;[manual-verify] palette present after remount: false&#10;[manual-verify] JSON button class after remount: css-xf7q69-button&#10;[manual-verify] Edit button class after remount: css-g0dahz-button&#10;[manual-verify] json editor present after remount: true&#10;Tests: 2 passed, 2 total</code>

```text
    [manual-verify] Preview button class after remount: css-xf7q69-button
    [manual-verify] Edit button class after remount: css-g0dahz-button
    [manual-verify] palette present after remount: false
    [manual-verify] JSON button class after remount: css-xf7q69-button
    [manual-verify] Edit button class after remount: css-g0dahz-button
    [manual-verify] json editor present after remount: true
Tests:       2 passed, 2 total
```
</details>

## Pipeline

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.
- <code>`npx jest src/components/block-editor/BlockEditor.persistence.test.tsx src/components/block-editor/hooks/useBlockEditor.test.ts src/components/block-editor/hooks/useBlockPersistence.test.ts src/components/block-editor/hooks/useJsonModeHandlers.test.ts` (target commit, all 35 pass)</code>
- <code>`npx jest src/components/block-editor` (full block-editor suite, 529 tests pass, no regressions)</code>
- `Same 4 test files re-run against base commit 342121e in a temp worktree with target's test files copied over — 9 tests fail, confirming they genuinely detect the pre-fix regression`
- `Manual verification test: rendered <BlockEditor/>, switched to Preview mode, fully unmounted, remounted fresh instance, confirmed Preview state (not Edit) survives — reproduces issue #1290's pop-out/dock scenario`
- `Manual verification test: same remount cycle for JSON mode, confirmed JSON editor and active-JSON button state survive remount`

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>


Jayclifford345 (commented):
# Review — multi-agent pass (5 concern reviewers + adversarial verification)

**Verdict: approve with one fix requested before merge.** The fix itself is correct, architecturally sound, revert-safe, and verified working — but the committed test suite does not actually lock the regression it fixes. Every finding below survived an adversarial verification pass (three independent skeptics per medium finding, mutation testing included). Security, doc-drift, and rollback checks came back clean: the restored draft only reaches text-safe sinks (Monaco `value`, JSX text), the restore validators build fresh whitelisted objects, and the additive storage shape is safe to read in both directions across a revert.

## Findings

### 1. Medium — no committed test locks the #1290 regression (testing-and-verification, confirmed)

The restore glue in `BlockEditor.tsx` (`onLoad` passing `savedViewMode` into `editor.loadGuide(...)` and calling `jsonMode.restoreJsonMode(...)` when the saved mode is `json`) has zero failing-test coverage. Three independent mutation runs confirmed it: with that glue deleted (fully reintroducing #1290), **tsc, eslint, and the entire 5,231-test frontend suite all stay green** (`loadGuide`'s third param is optional, so TypeScript can't catch it either). The new `BlockEditor.persistence.test.tsx` test also **passes on pre-fix code** — with real timers and a synchronous assertion, the pre-fix debounced save never fires, so the old competing load path never triggers inside the test. The hook-level tests are good and do fail pre-fix, but they pin each link in isolation, not the chain; #1290 was a composition bug, and the composition point is exactly what's unpinned. The manual remount verification test mentioned in the PR description was the load-bearing evidence, and it was removed before commit.

**Suggested action (proven feasible during review — ~2.3s runtime):**

1. Render `BlockEditor` → click Preview → `unmount()` → render again → assert palette absent.
2. Seed `StorageKeys.BLOCK_EDITOR_STATE` with a v2 payload (`viewMode: 'json'` + unapplied draft) → render → assert the JSON editor shows the draft. Needs a one-line `jest.mock('./BlockJsonEditor', ...)` stub because `@grafana/ui`'s CodeEditor pulls unmocked `monaco-editor` into jsdom.

Keep the existing test (it does guard reintroduction of the old competing load path alongside the new immediate save — verified) but retitle it, since no "persistence write" occurs within it.

### 2. Low — "New guide" re-persists the discarded guide's text as a dormant draft (reversibility)

If the user is in JSON mode and clicks "New guide", `handleNewGuide` calls `persistence.clear()` then `resetGuide()`, but nothing clears the in-memory `jsonModeState`. The resulting `json→edit` mode change triggers the new immediate save, which writes the full serialized text of the guide the user just discarded right back into the key `clear()` just wiped. Same for import/template while in JSON mode. It's never restored to the UI and self-heals on remount, but content the user explicitly discarded lingers in localStorage. Fix at the choke point: in `save()`, persist the draft only when it's restorable — `jsonModeState: viewMode === 'json' ? jsonModeState ?? undefined : undefined`.

### 3. Low — cross-session scope is broader than documented (architecture)

The restore fires on any mount with stored state, so viewMode and invalid JSON drafts now survive **browser restarts**, not just pop-out/dock — a user who closed the browser mid-draft reopens days later landing in JSON mode with an invalid draft. Defensible (guide content already persists across sessions by design), but the README bullet only says "survive remounts". Confirm it's intended and say so in the README; while there, the new bullet "View State" should be sentence case ("View state").

### 4. Low — malformed `jsonModeState` shapes untested (testing)

`restoreJsonModeState` correctly rejects wrong-typed fields, but only absent-field payloads are tested. One test seeding e.g. `jsonModeState: { json: 5 }` and asserting `onLoad` gets `undefined` would pin the validator against a future "simplification".

## Dropped after adversarial verification

**Guide-content edits <1s before pop-out still lost (was medium, refuted 3/3).** The mechanism is real — the debounce cleanup clears the pending save with no flush, and pop-out genuinely remounts (separate React roots) — but the gap is pre-existing and byte-identical to main, exists on every unmount path (tab switch, refresh), is outside #1290's explicitly viewMode-scoped ask, and this PR strictly *narrows* it (every view-state change now flushes the full guide). **Recommended: file as a follow-up issue** (an unmount-only flush effect in `useBlockPersistence`), not a blocker here.

Also verified: JSON drafts do **not** save per keystroke — `BlockJsonEditor.tsx` commits on blur, so the synchronous save fires once per blur/mode-toggle. No perf concern.

## Tech debt (non-blocking)

- `BlockEditorInner` is ~915 lines / 40+ hooks and the repo's hottest file (15 commits/90d); the ~200-line backend save/publish orchestration would extract cleanly into a `useBackendSaveFlow` hook.
- The PR trimmed narrating comments in the files it rewrote but touched `loadGuide` in `useBlockEditor.ts` without applying trim-on-touch there (~20 identifier-restating one-liners remain).
- After this PR, `persistence.load()`, `hasSavedGuide()`, and `getLastSaveTime()` have only test callers — prunable dead API surface in a follow-up.

**Doc drift:** none — both relevant docs were updated in the PR, and no agent-guidance file references the storage shape.

**Coverage note:** `src/components/block-editor/**` is not modeled by any subsystem concern in `docs/design/CONCERNS.md` — this review ran on always-on concerns only. Given it's the highest-churn area in the repo right now, a block-editor concern entry may be worth adding.

moxious:
Jay's review findings are addressed in `12922ea4` and `bc9ab395`:

- Added component-level persistence coverage that performs a real Preview unmount/remount and verifies exact restoration of an unapplied invalid JSON draft, including validation and undo state.
- Added an explicit assertion for the immediate Preview persistence write.
- Prevented dormant JSON drafts from being serialized when the editor is no longer in JSON mode, covering guide replacement paths through the shared persistence boundary.
- Added table-driven coverage for malformed persisted `jsonModeState` field types.
- Clarified that view state and unapplied drafts survive reloads/browser restarts, documented their discard boundary, and fixed the README heading to sentence case.
- Added the `block-editor-authoring` concern to `docs/design/CONCERNS.md`, covering restore ownership, Edit/Preview/JSON invariants, guide replacement, storage compatibility, backend tracking, and component-level remount verification.

Validation: `npm run check` passes, including typecheck, lint, Prettier, docs sync, Go lint/tests, and 332 frontend suites (5,216 passing tests).

The larger component extraction and dead-API cleanup noted as non-blocking tech debt remain follow-up work rather than expanding this bugfix PR.

Jayclifford345 (approved):

PR #1334, opened 2026-07-14
feat(telemetry): enrich Faro with identity, session attributes, and funnel instrumentation

Jayclifford345:
## Summary

Enriches Pathfinder's Faro telemetry so Frontend Observability data becomes joinable, filterable, and alertable. None of the existing isolation mechanisms (activity gate, attribution filter, `guardTelemetry`) are weakened.

- **Identity**: `setUser` with the same SHA-256 hashes the recommender payload uses (`hashUserData`) — the email hash is `meta.user.id` (primary join key), with the hashed analytics identifier as a `user_id_hash` attribute. Sessions also carry an `instance` attribute (stack hostname) so they are attributable to a Grafana instance. No raw PII.
- **Session attributes**: `surface` (sidebar/floating/fullscreen/kiosk/controller), `edition`, `language`, and `experiments` are stamped once per session via a new merge-style `setFaroSessionAttributes`, instead of repeating the ~300-char experiments JSON on every mirrored user action.
- **Resource noise**: an inverted-allowlist `ignoreUrls` regex drops Grafana core's resource-timing entries inside the `PerformanceObserver`, before payloads are built (~6,000 discarded payloads per session previously). The `beforeSend` path filter stays as the belt-and-braces layer.
- **Funnel measurements** (`pushMeasurement`): recommender latency (measured at the response, so ok/error durations are comparable), content-fetch ladder tiers, step execution, requirements-retry exhaustion, and panel time-to-content (gated on the recommendations list actually rendering via a `pathfinder-recommendations-ready` event, not just the tab being active).
- **Funnel failure events**: `recommender_fallback`, `content_fetch_fallback`, `requirements_exhausted` — pushed as plain Faro events (new `pushFaroEvent`) for alerting only; they do not reach RudderStack, since they are operational signals rather than product analytics.
- **Real durations**: guide-open, section-run, and "do it" flows wrapped in `withFaroUserAction` with `importance: critical`, so the Actions tab's avg/P95 stop reading ~0ms.
- **Explicit error capture**: interactive handlers failing against Grafana core's DOM (no plugin stack frame, previously invisible) now report Faro exceptions with step/action context, deduped per Error instance so handler+caller double-reporting counts once.
- **Cold-start view fix**: `setFaroViewName('recommendations')` when the active tab has no URL.
- **Ride-along cleanups**: `pathfinder-panel-mode-change` hoisted to a shared constant in `lib/event-names.ts` (was a raw literal in 8 files), `isPathfinderOpen` collapsed onto the new `readPathfinderSurface`, and structured context added to funnel-path `logger.warn/error` calls.

## What to look at

- `src/lib/faro.ts` — the new API surface (`setFaroSessionAttributes`, `pushFaroMeasurement`, `setFaroViewName`, `buildResourceIgnorePattern`, critical-importance option) and the user/surface stamping in `initFaro`.
- `src/components/docs-panel/docs-panel.tsx` + `context-panel.tsx` — the `panel_lcp_ms` ready-event handshake.
- `src/interactive-engine/interactive-state-manager.ts` — the `WeakSet` exception dedupe.

## How to verify

- `npm run test:ci` — 333 suites / 5,248 tests green (includes ~40 new tests across faro, analytics, state-manager, sequence-manager, content-fetcher, recommender, and view-exposure suites).
- `npm run lint` + typecheck — clean (pre-existing warnings only).
- Manual (optional, Grafana Cloud dev): set `localStorage['pathfinder.faro.local']='true'`, open Pathfinder, confirm collector POSTs carry `meta.user.id` (64-char hex), session attributes with `surface`, and no `faro.performance.resource` items for core hosts.

## Breaking changes

None. Existing `pathfinder_*` event names are unchanged (RudderStack parity preserved); `experiments` still reaches RudderStack in full and is only stripped from the Faro mirror.

### DIFF
--- a/src/components/App/ContextPanel.tsx
+++ b/src/components/App/ContextPanel.tsx
@@ -4,6 +4,7 @@ import { CombinedLearningJourneyPanel } from 'components/docs-panel/docs-panel';
 import { getConfigWithDefaults } from '../../constants';
 import { PathfinderFeatureProvider } from '../OpenFeatureProvider';
 import { panelModeManager, type PanelMode } from '../../global-state/panel-mode';
+import { PANEL_MODE_CHANGE_EVENT } from '../../lib/event-names';
 
 export default function MemoizedContextPanel() {
   const pluginContext = usePluginContext();
@@ -14,9 +15,9 @@ export default function MemoizedContextPanel() {
     const handleModeChange = (e: CustomEvent<{ mode: PanelMode }>) => {
       setMode(e.detail.mode);
     };
-    document.addEventListener('pathfinder-panel-mode-change', handleModeChange as EventListener);
+    document.addEventListener(PANEL_MODE_CHANGE_EVENT, handleModeChange as EventListener);
     return () => {
-      document.removeEventListener('pathfinder-panel-mode-change', handleModeChange as EventListener);
+      document.removeEventListener(PANEL_MODE_CHANGE_EVENT, handleModeChange as EventListener);
     };
   }, []);
 
--- a/src/components/block-editor/BlockEditorHeader.tsx
+++ b/src/components/block-editor/BlockEditorHeader.tsx
@@ -15,6 +15,7 @@ import { GrafanaTheme2 } from '@grafana/data';
 import type { ViewMode } from './types';
 import { testIds } from '../../constants/testIds';
 import { panelModeManager, type PanelMode } from '../../global-state/panel-mode';
+import { PANEL_MODE_CHANGE_EVENT } from '../../lib/event-names';
 
 export interface BlockEditorHeaderProps {
   /** Guide title to display */
@@ -286,9 +287,9 @@ export function BlockEditorHeader({
     const handleModeChange = (e: CustomEvent<{ mode: PanelMode }>) => {
       setPanelMode(e.detail.mode);
     };
-    document.addEventListener('pathfinder-panel-mode-change', handleModeChange as EventListener);
+    document.addEventListener(PANEL_MODE_CHANGE_EVENT, handleModeChange as EventListener);
     return () => {
-      document.removeEventListener('pathfinder-panel-mode-change', handleModeChange as EventListener);
+      document.removeEventListener(PANEL_MODE_CHANGE_EVENT, handleModeChange as EventListener);
     };
   }, []);
 
--- a/src/components/docs-panel/context-panel.tsx
+++ b/src/components/docs-panel/context-panel.tsx
@@ -18,6 +18,7 @@ import { useContextPanel, Recommendation } from '../../context-engine';
 import type { ResolvedNavLink } from '../../types/context.types';
 import { reportAppInteraction, UserInteraction, getContentTypeForAnalytics } from '../../lib/analytics';
 import { logger } from '../../lib/logging';
+import { RECOMMENDATIONS_READY_EVENT } from '../../lib/event-names';
 import { getConfigWithDefaults, PLUGIN_BASE_URL } from '../../constants';
 import { isDevModeEnabled } from '../../utils/dev-mode';
 import { testIds } from '../../constants/testIds';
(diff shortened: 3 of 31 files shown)


moxious (changes requested):
## Request changes

I found five blocking issue clusters and several additional telemetry-correctness problems. CI is green, but the current tests miss the load-bearing integration seams.

### Blocking findings

1. **Execution funnels discard typed outcomes and report failures as success.**

   - Guide loaders store errors with `failTab()` and resolve, so `withFaroUserAction` records failed opens as `ok`.
   - `executeStep()` returns `false` on execution or verification failure, but its caller discards that boolean and records `step_exec_ms` as `ok`.
   - Guided timeout, cancellation, skip, and caught errors resolve normally, so critical guided actions are stamped `ok`.
   - Sequence exhaustion resolves `Promise<void>`, after which the hook emits `interactive-action-completed`. Persistent action exceptions are also mislabeled `requirements_exhausted`.

   Please propagate a typed result such as `completed | cancelled | timeout | requirements_exhausted | action_error` through loaders, lazy scrolling, guided execution, and sequence retries. Only emit completion for `completed`.

2. **One recommender request can emit both success and failure.** `context.service.ts` records `outcome: ok` immediately after JSON parsing. Later mapping, sanitization, or journey processing can still throw into `fail()`, producing a second error measurement and fallback event for the same request. Validate/process first and settle telemetry exactly once; add a malformed-response test asserting one outcome.

3. **Raw content URLs cross the Faro boundary.** Guide actions, fetch measurements, and fallback events use the original URL, including query strings and fragments. Faro stringification/truncation does not remove sensitive or high-cardinality components. Centralize a telemetry URL normalizer that emits a bounded hostname/path or controlled content identifier and strips userinfo, query, and fragment before every telemetry call.

4. **Identity enrichment can merge unrelated users and needs an explicit privacy decision.** `stampFaroUser` hashes `unknown@example.com` as the primary Faro ID for every email-less Cloud user, even when a distinct analytics identifier is available. The Cloud/OSS identity and fallback policy is also duplicated between Faro and the recommender. Extract one typed identity builder, fall back to the hashed analytics identifier when email is absent, and explicitly align/document the intended data-use gate for Faro identity, raw instance hostname, role, language, and experiments.

5. **Experiment enrichment is an unversioned, distributed schema migration.** The PR removes `experiments` from mirrored actions and stamps it later on the session through bootstrap code. There is no compatibility window or schema version. The serialized JSON also passes through `stringifyAttributes`, which slices strings at 500 characters and can produce invalid JSON. Define a compact, versioned cohort schema; avoid truncating encoded JSON; preserve compatibility for downstream consumers; and make one bootstrap API own collection, serialization, and session stamping.

### Additional correctness findings

- **Surface attribution becomes stale.** Faro only restamps on panel-mode changes. Sidebar, kiosk, and controller mounts do not emit that event; floating-to-sidebar stamps `closed` before the sidebar opens and never corrects it.
- **Panel LCP starts too late.** Its clock is initialized inside the passive measurement effect, so already-ready content measures near zero and render/commit time is omitted.
- **Terminal content-ladder failures lose their classification.** When both `content.json` and `unstyled.html` fail, the measurement remains `tier: other` and no fallback event is emitted.
- **Behavioral seam tests are missing.** Tests cover generic Faro helpers, but not experiment bootstrap, production `ignoreUrls` wiring, panel readiness/LCP, guide-open outcomes, or step-execution outcomes.

### Tech debt and modularization

The Faro affordance is becoming a fractured cross-tier API:

- `faro.ts` grew from 415 to 538 lines in four days and now exposes 21 exports, including roughly 18 callable primitives.
- It combines environment policy, SDK lifecycle, attribution filtering, activity gating, surface detection, identity, session mutation, actions, measurements, events, views, and resource filtering.
- Context, retrieval, interactive, integration, and UI tiers directly import vendor-specific `pushFaro*` functions and define raw event names and payload shapes locally.
- The pending OTel branch modifies the same initialization and action hotspot.
- `module.tsx` is now a 521-line bootstrap hub; this PR adds another ordering-sensitive initialization sequence there.

Please establish the boundary in this PR: keep `lib/faro.ts` temporarily as a compatibility barrel; add a neutral, typed telemetry facade with domain operations; keep Faro and OTel as internal adapters; split bootstrap/session enrichment, filtering, identity, actions, and view/resource concerns; and introduce one typed surface-state owner plus one shared hashed-identity builder.

The developer documentation follow-up promised by #1275 is also still absent. The collector, gates, identity fields, schemas, rollback behavior, and `pathfinder.frontend-telemetry` flag should be documented as this contract expands.

The adversarial verification pass dropped one proposed finding: the recommendations-ready listener race was refuted because `hasFetchedRecommendations` initializes `false`. The findings above survived independent verification.

Jayclifford345:
Closed PR out for #1338

PR #1335, opened 2026-07-14
fix(bootstrap): survive chunk-load failures instead of killing the plugin

Jayclifford345:
## Summary

A production `ChunkLoadError` (observed on `plugins-cdn.grafana-dev.net` for the 2.14.0 build) currently kills Pathfinder entirely: the experiments / sidebar-auto-open / openfeature bootstrap chunks and `initPluginTranslations` are loaded with **unguarded top-level awaits** in `module.tsx`, so one failed chunk fetch rejects module evaluation and Grafana reports `Could not load plugin`.

```
Error: Could not load plugin
Caused by: ChunkLoadError: Loading chunk utils_openfeature-tracking_ts-utils_openfeature_ts failed.
```

This happens whenever a chunk fetch fails — most commonly when a redeploy replaces hashed chunks while a browser still holds a cached `module.js` referencing the old names, or on transient CDN/network failures.

## What changed

- The three bootstrap dynamic imports are loaded together under one try/catch into a nullable `bootstrap` handle.
- On failure the plugin **degrades instead of dying**: feature flags fall back to their defaults (`pathfinder.enabled` → true, so the sidebar still registers), the highlighted-guide experiment and auto-open are skipped, and the failure is reported via `logger.exception`.
- `initPluginTranslations` is guarded the same way — `t()` falls back to default messages.
- The already-guarded OpenFeature init and Faro init blocks are unchanged.

## How to verify

- `npm run test:ci`, typecheck, and lint all pass.
- Manual: block one of the bootstrap chunk URLs in devtools (network request blocking) and reload — the plugin loads with default flags and logs the exception instead of failing preload.

## Breaking changes

None. When all chunks load (the normal case) behavior is identical.

### DIFF
--- a/docs/developer/FEATURE_FLAGS.md
+++ b/docs/developer/FEATURE_FLAGS.md
@@ -29,6 +29,8 @@ The plugin uses the [OpenFeature](https://openfeature.dev/) standard with the OF
 
 **Important**: This is the only gate on whether Pathfinder mounts. It is evaluated independently of the highlighted-guide experiment: setting `pathfinder.enabled` to `false` dismounts the plugin regardless; when `true`, the experiment applies as normal.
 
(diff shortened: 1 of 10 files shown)


Jayclifford345:
@claude review this PR

PR #1346, opened 2026-07-15
feat(e2e): send requiredPlugins to pool manager when leasing a stack

### LINKED ISSUE #1190: Install manifest-declared plugins on isolated Cloud E2E stacks
Parent: #1184

## Problem

Guides can declare required plugins in `testEnvironment.plugins`, but the E2E runner currently checks for those plugins instead of installing them.

## Scope

- Carry `testEnvironment.plugins` through remote package resolution.
- Install declared plugins on isolated stacks before running a guide.
- Keep shared stacks read-only; do not install plugins into shared stacks automatically.
- Decide whether versioned plugin requirements are needed later. The current manifest shape only declares plugin IDs.

## Acceptance criteria

- Isolated stack provisioning installs required manifest plugins before Playwright preflight.
- Shared-stack runs still only check plugin availability.
- Missing or un-installable plugins fail setup with a clear message.

leslie-heinzen:
## Summary

Guides that declare `testEnvironment.plugins` in their manifest now have those plugins installed on the isolated Cloud stack before Playwright runs. Previously, `leaseForChain` always sent `requiredPlugins: []` to the pool manager, so the pool manager never installed them and guides requiring extra plugins would fail at runtime with a missing-plugin error.

## What to look at

- **[`src/cli/e2e/e2e-package.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/890b5f812562eb3e44896b95d7a65ac285b4163e/src/cli/e2e/e2e-package.ts)** — `buildGuideOrSkip` now spreads `testEnvironment.plugins` onto the returned `ResolvedRemoteGuide` only when the array is non-empty, keeping the field absent for guides that declare no plugins.

- **[`src/cli/e2e/e2e-results.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/890b5f812562eb3e44896b95d7a65ac285b4163e/src/cli/e2e/e2e-results.ts)** — `PackageMeta` gains an optional `plugins` field; `buildPackageMetaMap` carries it through from the resolved guide so downstream code can reach it from the map.

- **[`src/cli/e2e/cloud-stack-pool-manager.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/890b5f812562eb3e44896b95d7a65ac285b4163e/src/cli/e2e/cloud-stack-pool-manager.ts)** — `collectRequiredPlugins` walks the chain, reads `plugins` from each guide's `PackageMeta`, and deduplicates across guides. `leaseForChain` now calls it instead of hardcoding `[]`. The `CreateLeaseRequest.requiredPlugins` type is corrected from the `[]` empty-tuple literal to `PluginRequirement[]`.

## Why

Guides can declare `testEnvironment.plugins` in their manifest to flag Grafana plugins they depend on. The pool manager already supported `requiredPlugins` in the lease request — it validates each slug against the pool's `allowedExtraPlugins` allowlist and installs them on the leased stack before returning. The CLI side had never been wired up, so the pool manager never saw these requests.

The fix threads the information through three layers (`ResolvedRemoteGuide` → `PackageMeta` → `leaseForChain`) without changing any behavior for shared-stack runs, which still check plugin availability through the existing preflight path.

## How to verify

1. Run a remote package that declares `testEnvironment.plugins` against a pool that allows those plugins:
   ```
   pathfinder-cli e2e --package <guide-id> --tier cloud \
     --cloud-stack-pool-manager-url <url> \
     --cloud-stack-pool-manager-token <env-var> \
     --cloud-stack-pool-id <pool-id>
   ```
   Confirm the lease succeeds and the guide runs without a missing-plugin error. Previously the plugin would not be installed and the guide would fail at runtime.

2. Run the same command for a guide with no `testEnvironment.plugins`. Confirm it still leases and runs normally (empty `requiredPlugins` path is unchanged).

3. Run a guide whose `testEnvironment.plugins` entry is not in the pool's `allowedExtraPlugins`. Confirm the CLI fails at lease time with a clear "required plugin is not allowed for pool" message rather than a runtime test failure.

## Breaking changes

None. The `plugins` field is optional on both `ResolvedRemoteGuide` and `PackageMeta`. Shared-stack runs and guides with no declared plugins are unaffected. The pool manager's `requiredPlugins` field has always accepted an empty array; the only change is that it now receives non-empty arrays when guides declare plugins.

Closes #1190


moxious:
@cursoragent review

src/cli/e2e/cloud-stack-pool-manager.ts
```diff
@@ -37,13 +37,18 @@ interface LeaseForChainOptions {
   packageMetaById: Map<string, PackageMeta>;
 }
 
+interface PluginRequirement {
+  slug: string;
+  version?: string;
```
moxious, line 42: Nit.  You don't appear to need or actually use the `version` string.  Seems unnecessary as it isn't referred to elsewhere in the PR.  I can see why you'd want it for forward compat and it's not a huge deal so that's not really a reason to stop the PR.

moxious (approved):

PR #1348, opened 2026-07-15
perf(telemetry): latch the cold-path surface read until the first report

### LINKED ISSUE #1344: telemetry: latch cold-path surface state so beforeSend stops re-reading DOM/localStorage per event (follow-on to #1338)
Follow-on from #1338. Not a regression — the pre-#1338 `faro.ts` had the identical cost — but the new module split makes the fix easy to place, and the `reportedSurface` cache the PR added already solves the post-open half of it.

## Problem

The Faro `beforeSend` hook runs `passesActivityGate` on every telemetry event (`src/lib/telemetry/faro-adapter.ts` ~line 98 → `src/lib/telemetry/filtering.ts` ~line 179). Until the first `reportPathfinderSurface()` call lands, `getPathfinderSurface()` falls through to `readPathfinderSurface()` (`src/lib/telemetry/surface.ts` ~lines 21–46), which does per event:

- `localStorage.getItem(StorageKeys.PANEL_MODE)`
- `isExtensionSidebarOwnedByPathfinder()` (DOM traversal)
- `document.querySelector` (kiosk overlay) + `document.getElementById` (controller root)

`reportedSurface` is only stamped when a surface owner mounts (module init for controller mode, panel mount for floating/fullscreen, async import completion for kiosk). Before that — i.e. the entire time the panel is closed — every error/web-vitals/instrumentation event pays the full DOM + localStorage read chain.

## Fix

Latch the cold-path answer instead of re-deriving it per event. Options, roughly in order of preference:

1. Stamp an initial surface value once at adapter init (`reportPathfinderSurface(readPathfinderSurface())` or an explicit `'closed'` default), so `reportedSurface` is never null and the per-event fallback disappears. Surface owners already overwrite it on mount and the write path dedups unchanged values.
2. Or cache `readPathfinderSurface()`'s result with invalidation on the events that can change it (panel-mode storage writes, mount/unmount reports).

Keep the existing behavior that the gate re-opens correctly when Pathfinder opens — the surface listener ordering note in `faro-adapter.ts` documents the invariant to preserve.

moxious:
## Summary

While Pathfinder is closed, every Faro telemetry event (errors, web vitals, instrumentation) pays a full DOM + localStorage read chain: the `beforeSend` activity gate calls `getPathfinderSurface()`, which falls through to `readPathfinderSurface()` — a `localStorage.getItem` for panel mode, a docked-sidebar ownership check, a `document.querySelector` for the kiosk overlay, and a `document.getElementById` for the controller root — on every event until the first `reportPathfinderSurface()` lands. This PR latches that cold read once in a module-level cache, so the closed-state cost is a single read instead of per-event.

Follow-on to #1338, not a regression from it: the pre-#1338 `faro.ts` had identical per-event cost. #1338's `reportedSurface` cache already fixed the post-open half; this fixes the cold (pre-open) half.

## What to look at

- [`src/lib/telemetry/surface.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/0a46c12519f9eac89096429d76a014d2b185648a/src/lib/telemetry/surface.ts) — the latch/supersede interaction. Two things that look subtle but are deliberate:
  - A first report equal to the latched cold value **still notifies listeners**, because the dedup compares against `reportedSurface` (null until the first report), not the latch. `faro-adapter.ts`'s `stampSurface` listener depends on seeing that first transition.
  - An out-of-band localStorage write after the latch (e.g. a cross-tab `setMode`) no longer re-opens the activity gate via re-read — by design. Every same-tab open path reports on mount/open (`setMode` reports synchronously after its localStorage write, sidebar mount reports, kiosk `handleOpen` reports synchronously, controller reports at module init), so the report path alone preserves the gate-opens-when-Pathfinder-opens invariant. The decision is encoded in tests, not comments.
- Tests: pre-existing cases asserting per-event re-read semantics were rewritten to report-driven semantics (`surface.test.ts` fallback + kiosk-overlay tests; `faro.test.ts` "closed does not latch" gate test and the `beforeSend` wiring test), with original intent preserved via fresh-module cold reads or explicit reports. The second commit implements a mutation-verified self-review finding: the "latches open for the rest of the page load" gate test closed via an out-of-band `localStorage` removal the latch no longer observes, which made the `pathfinderWasOpen` coverage vacuous — it now closes via `reportPathfinderSurfaceClosed`, which kills that mutant.

## How to verify

1. In a dev Grafana with the plugin loaded and Pathfinder closed, instrument reads in DevTools (`const g = Storage.prototype.getItem; Storage.prototype.getItem = function (...a) { console.count('getItem'); return g.apply(this, a); }`) and generate telemetry events (navigate, trigger web vitals) — confirm no per-event panel-mode reads after the first.
2. Open Pathfinder (sidebar, floating, or kiosk) — confirm subsequent events pass the activity gate and the session's `surface` attribute stamps to the opened surface.
3. The pipeline's read-count evidence below (100 closed-state events through the real `beforeSend` gate): constant 2/1/1 reads with the latch vs linear 200/100/100 on the base commit.

## Breaking changes

None. No event names, payload shapes, or telemetry routing change — only when the surface value is derived.

## Reversibility

Fully reversible; no persisted state or external contract is touched. Revert restores the per-event read chain. The only observable behavior delta is the edge case above: an out-of-band localStorage write between the first closed-state event and the first surface report no longer opens the gate by itself — same-tab opens are unaffected because they always report.

Fixes #1344

<details>
<summary>Evidence: Before/after per-event surface read counts (issue #1344)</summary>

```text
Issue #1344 — latch cold-path surface read so Faro beforeSend stops re-reading
DOM/localStorage per telemetry event while Pathfinder is closed.

Measurement: 100 telemetry events pushed through the REAL Faro activity gate
(passesActivityGate -> isPathfinderOpen -> getPathfinderSurface) with Pathfinder
CLOSED the entire time. Spies count actual localStorage/DOM reads.

                          BASE (7810159, no latch)   TARGET (0a46c12, latched)
  localStorage.getItem      200  (2.00 / event)         2  (0.02 / event)
  document.querySelector    100  (1.00 / event)         1  (0.01 / event)
  document.getElementById   100  (1.00 / event)         1  (0.01 / event)

BASE:   reads scale linearly with event volume (the bug in #1344).
TARGET: one-time cold read, then 99/100 events answer from the module latch.
        A real open transition (reportPathfinderSurface) still flips the gate
        open exactly once, so no telemetry is lost.
```
</details>

## Pipeline

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.
- <code>`npx jest src/lib/telemetry/surface.test.ts src/lib/faro.test.ts` — 137 committed tests pass</code>
- <code>Temporary evidence test pushing 100 closed-state events through the real `passesActivityGate` path, counting `localStorage.getItem`/`document.querySelector`/`document.getElementById` calls on the target commit (constant one-time cold read)</code>
- `Same measurement against base commit 7810159's surface.ts (reads scale linearly: 200/100/100) to establish before/after contrast`
- <code>Mutation check: removed the `pathfinderWasOpen` latch in filtering.ts and ran the "latches open for the rest of the page load" test — it correctly failed, confirming the rewritten test kills the mutant</code>

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

### DIFF
--- a/src/lib/faro.test.ts
+++ b/src/lib/faro.test.ts
@@ -1174,10 +1174,11 @@ describe('passesActivityGate', () => {
   });
 
   it('latches open for the rest of the page load once Pathfinder was open', () => {
-    localStorage.setItem(PANEL_MODE_KEY, 'floating');
     const faro = freshFaro();
+    const surface: typeof import('./telemetry/surface') = require('./telemetry/surface');
+    surface.reportPathfinderSurface('floating');
     expect(faro.passesActivityGate(eventItem())).toBe(true);
-    localStorage.removeItem(PANEL_MODE_KEY);
+    surface.reportPathfinderSurfaceClosed('floating');
     expect(faro.passesActivityGate(eventItem())).toBe(true);
   });
 
--- a/src/lib/faro.test.ts
+++ b/src/lib/faro.test.ts
@@ -1183,8 +1183,19 @@ describe('passesActivityGate', () => {
 
   it('opens on a later check after starting closed — closed does not latch', () => {
     const faro = freshFaro();
+    const surface: typeof import('./telemetry/surface') = require('./telemetry/surface');
+    expect(faro.passesActivityGate(eventItem())).toBe(false);
+    surface.reportPathfinderSurface('floating');
+    expect(faro.passesActivityGate(eventItem())).toBe(true);
+  });
+
+  it('an out-of-band localStorage write after a closed check does not open the gate — only a surface report does', () => {
+    const faro = freshFaro();
+    const surface: typeof import('./telemetry/surface') = require('./telemetry/surface');
     expect(faro.passesActivityGate(eventItem())).toBe(false);
     localStorage.setItem(PANEL_MODE_KEY, 'floating');
+    expect(faro.passesActivityGate(eventItem())).toBe(false);
+    surface.reportPathfinderSurface('floating');
     expect(faro.passesActivityGate(eventItem())).toBe(true);
   });
 
@@ -1216,7 +1227,8 @@ describe('beforeSend wiring', () => {
     expect(beforeSend(exceptionItem(['webpack://grafana-pathfinder-app/./src/lib/faro.ts']))).not.toBeNull();
     expect(beforeSend(exceptionItem(['webpack://grafana-core/./src/app.ts']))).toBeNull();
 
-    localStorage.setItem('grafana-pathfinder-app-panel-mode', 'floating');
+    const surface: typeof import('./telemetry/surface') = require('./telemetry/surface');
+    surface.reportPathfinderSurface('floating');
     expect(beforeSend(eventItem())).not.toBeNull();
   });
 });
--- a/src/lib/telemetry/surface.test.ts
+++ b/src/lib/telemetry/surface.test.ts
@@ -63,7 +63,7 @@ describe('surface owner', () => {
     expect(listener).toHaveBeenCalledWith('floating');
   });
 
-  it('double init: a second consumer cold-reading after the latch sees the latched value, not the changed DOM', () => {
+  it('a later cold-path caller sees the latched value, not the changed DOM', () => {
     const surface = freshSurface();
     expect(surface.isPathfinderOpen()).toBe(false);
 
--- a/src/lib/telemetry/surface.test.ts
+++ b/src/lib/telemetry/surface.test.ts
@@ -10,28 +10,84 @@ describe('surface owner', () => {
     document.body.innerHTML = '';
   });
 
-  it('falls back to the DOM/localStorage read until a surface is reported', () => {
-    const surface = freshSurface();
-    expect(surface.getPathfinderSurface()).toBe('closed');
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
 
+  it('falls back to the DOM/localStorage read until a surface is reported', () => {
     localStorage.setItem('grafana-pathfinder-app-panel-mode', 'floating');
+    const surface = freshSurface();
     expect(surface.getPathfinderSurface()).toBe('floating');
   });
 
-  it('cold-read fallback requires the visible kiosk overlay, not just its persistent mount root', () => {
+  it('performs the cold read once and answers subsequent calls from the latch', () => {
     const surface = freshSurface();
+    const getItem = jest.spyOn(Storage.prototype, 'getItem');
+    const querySelector = jest.spyOn(document, 'querySelector');
 
+    expect(surface.getPathfinderSurface()).toBe('closed');
+    const storageReads = getItem.mock.calls.length;
+    const domQueries = querySelector.mock.calls.length;
+    expect(storageReads).toBeGreaterThan(0);
+
+    surface.getPathfinderSurface();
+    surface.isPathfinderOpen();
+    expect(getItem).toHaveBeenCalledTimes(storageReads);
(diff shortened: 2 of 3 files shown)


moxious:
## Self-review (`/review` per `.cursor/skills/review/SKILL.md`)

**Classification:** `product-runtime` (telemetry gating path) + test changes.
**Routing:** always-on `security`, `correctness-and-reliability`, `testing-and-verification`, `reversibility-and-one-way-door`, plus activated `analytics-and-telemetry` (trigger path `src/lib/telemetry/**`, keywords `beforeSend`/`Faro`). `performance-and-bundle` considered: no bundle-boundary change (no new imports; `surface.ts` stays entry-eager-safe).

### Reviewer verdicts

| Concern | Verdict |
| --- | --- |
| security | `reviewed_clean` — the latch narrows the same-origin localStorage spoofing window (an out-of-band write could previously flip the gate on any later event; now only the single pre-report cold read), no new data reaches the collector |
| correctness-and-reliability | `reviewed_clean` — `reportedSurface` is monotonically non-null after the first report, so clearing the latch on report can never cause a later cold re-read; dedup compares against `reportedSurface`, never the latch, so the first report always notifies listeners |
| analytics-and-telemetry | `reviewed_clean` — the only collector-visible delta is dropping cross-tab sessions whose gate opened without Pathfinder ever being used in that tab, which the gate's documented semantics ("used Pathfinder or Pathfinder errored") say should never have passed; also fixes a pre-existing inconsistency where the gate could open while the stamped `surface` session attribute stayed `closed` |
| reversibility-and-one-way-door | `reviewed_clean` — module-local state only, revert restores old behavior byte-for-byte |
| testing-and-verification | **1 finding (medium, mutation-verified)** — see below |

### Finding addressed (commit 0a46c125)

The pre-existing `faro.test.ts` test "latches open for the rest of the page load once Pathfinder was open" became vacuous under the latch: it closed the surface via an out-of-band `localStorage.removeItem`, which the latched cold read no longer re-reads — so the surface latch alone made the second gate check pass, and deleting the `pathfinderWasOpen` gate latch in `filtering.ts` shipped green (verified by mutation: 166/166 tests passed with the latch deleted). The gate latch is still load-bearing in production (surfaces genuinely report `closed` on unmount, and the latch keeps post-close telemetry like the sidebar-close mirror flowing). Fixed by closing via the surface owner (`reportPathfinderSurfaceClosed('floating')`), which the latch cannot mask; re-verified that the rewritten test kills the mutant. Also renamed a surface test whose "double init" name overstated what it exercises.

### Adversarial verification

The single medium finding was verified by running the mutation locally before and after the fix (mutant survived pre-fix, killed post-fix) rather than accepted on the reviewer's word.

### Tech debt (non-blocking)

E4 (module-level mutable registry): `surface.ts` now holds two module-level mutable variables plus a listener set with no exported reset, which is why tests need the `jest.resetModules()` / inline-`require` choreography. Pre-existing pattern from #1338's singleton design, deepened by one variable here; churn is low (1 commit/90d). Worth a `createSurfaceStore()` or test-reset export if this file keeps growing — deliberately not done in this PR to keep the footprint inside `surface.ts` semantics-only.

### Doc drift

None — no new subsystem, script, skill, route, or flag; `src/lib/telemetry/**` is already routed in `docs/design/CONCERNS.md` and the tier map is unchanged.

Jayclifford345 (approved):

PR #1347, opened 2026-07-15
fix(telemetry): replace raw error_message with low-cardinality classification in sequence action errors

moxious:
## Summary

Follow-on to #1338. The `sequence_action_error` telemetry event no longer ships the raw JavaScript `error.message` to Faro — it now records a bounded `error_name` plus a coarse `error_category` (`timeout | not_found | dispatch_failed | other`), so free-text failure content never leaves the client.

## Why

The #1338 review found that `recordSequenceActionError` forwarded `error.message` with only length truncation (200 chars at the call site, 500 in the Faro adapter), and nothing downstream scrubs regular event attributes — `filterPathfinderTelemetry` passes them through unchanged. DOM/action failure messages routinely embed URLs with query strings, CSS selectors, fetch response text, or echoed user input. Faro data is Grafana-internal, so this is privacy hygiene rather than an external leak, but it bypassed the content scrubbing the logging path gets. Of the two fixes the issue proposed (sanitize the string vs. classify), classification is stronger and matches the facade's own direction of typed, low-cardinality outcomes over free-form strings.

## What to look at

- [`src/interactive-engine/sequence-manager.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/179308e0c71e556a1c52ce5e6be1c4cd81aaccf1/src/interactive-engine/sequence-manager.ts) — new `classifySequenceError` at the catch site: timeout and not-found keyword matches first, other `Error` instances map to `dispatch_failed` (the throw source in this retry path is the action dispatch), non-`Error` throws map to `other`. `verify_failed` from the issue's example set was deliberately omitted — post-action verification failures return `false` and flow to `requirements_exhausted`, never through this throw path, so the category would be dead.
- [`src/lib/telemetry/facade.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/179308e0c71e556a1c52ce5e6be1c4cd81aaccf1/src/lib/telemetry/facade.ts) + [`types.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/179308e0c71e556a1c52ce5e6be1c4cd81aaccf1/src/lib/telemetry/types.ts) — `recordSequenceActionError` now takes a typed `SequenceErrorClassification` instead of a string, making the low-cardinality contract explicit at the boundary.

## How to verify

1. Run a guide with an interactive step whose selector cannot match (or point a step at a URL that will time out) and let retries exhaust.
[...]

PR #1350, opened 2026-07-15
fix(telemetry): redact embedded URLs from Faro exception/log content

### LINKED ISSUE #1345: telemetry: sanitize error_message in recordSequenceActionError (follow-on to #1338)
Follow-on from #1338, which added `recordSequenceActionError` to classify terminal sequence failures instead of leaving them as silent `tier: other`.

## Problem

The new event ships the raw JavaScript error message to Faro with only length truncation:

- `src/interactive-engine/sequence-manager.ts` (~line 66) captures `error.message` from a failed interactive action; ~line 79 passes it as `lastErrorMessage.slice(0, MAX_ERROR_CONTEXT_LENGTH)` (200 chars).
- `src/lib/telemetry/facade.ts` (~line 81) pushes it as the `error_message` attribute on the `sequenceActionError` event.
- Nothing downstream scrubs it: `filterPathfinderTelemetry` in `src/lib/telemetry/filtering.ts` returns regular events' attributes unchanged, and `stringifyAttributes` in the adapter only truncates to 500 chars.

Error messages from DOM/action failures routinely embed high-cardinality or sensitive fragments: URLs (with query strings), CSS selectors, fetch response text, and occasionally echoed user input. The exposure is bounded — Faro data is Grafana-internal and 200 chars is a hard cap — so this is privacy hygiene rather than an external leak, but it bypasses the content-based scrubbing that the logging path gets via `sanitizeForLogging`.

## Fix

Apply content sanitization before the string reaches the event, e.g.:

- run `lastErrorMessage` through the existing `src/security/log-sanitizer.ts` (and add URL normalization for embedded URLs via `normalizeTelemetryUrl`), or
- record a low-cardinality classification instead of free text: error `name` plus a coarse category (`timeout | not_found | dispatch_failed | verify_failed`), keeping the free-text message out of telemetry entirely.

The second option is stronger and matches the PR's own direction of typed, low-cardinality outcomes over free-form strings.

Jayclifford345:
## Summary

Follow-on to #1347, found while doing a final review pass on it. The `sequence_action_error` event fix in #1347 narrowed what reaches that one *custom event* — but the exact same secret-laden `Error` still leaks through an untouched sibling path: Faro *exceptions* and *logs*.

## Problem

`SequenceManager`'s retry loop calls `this.stateManager.logError(errorContext, error as Error, context.data)` on every retry, before `classifySequenceError` even runs. That raw `Error` flows through `logger.error` → `pushFaroError` → `faroInstance.api.pushError(error, ...)` untouched — only the wrapper message string gets `sanitizeForLogging()` (control-char stripping + length cap, no URL/secret redaction).

`filterPathfinderTelemetry`'s `beforeSend` exception branch only decides keep-vs-drop by stack-frame attribution; it never scrubs `payload.value`. So a DOM-selector-not-found message like:

```
No elements found matching selector: a[href="https://grafana.example/d/abc?token=S:sk-live-9f83b2&user=jdoe@corp.com"]
```

reaches Faro in full — full message, uncapped, on every retry. The same gap exists for log items when the thrown value isn't an `Error` instance (`logger.error` falls back to `pushFaroLog`).

## Fix

Redact at the shared choke point instead of patching the one call site: `filterPathfinderTelemetry` now runs any `https?://` substring found in an exception's `.value` or a kept log's `.message` through the existing `normalizeTelemetryUrl` (bounded `hostname/path`; userinfo, query, and fragment stripped) — the same rule already applied to dedicated URL-carrying telemetry attributes elsewhere in this codebase. This covers every current and future exception/log producer, not just this one retry loop, and it's a blanket rule (all query params dropped, not just ones that look sensitive), matching the existing `normalizeTelemetryUrl` policy.

Reference identity is preserved when nothing changes (no embedded URL), so existing `toBe(item)` assertions in `faro.test.ts` are untouched.

## What to look at

- `src/lib/telemetry/filtering.ts` — `redactEmbeddedUrls` + its two call sites in `filterPathfinderTelemetry`
- `src/lib/faro.test.ts` — new cases for attributed/explicit-report exceptions and kept logs with embedded URLs, plus a same-reference case confirming no-URL content is untouched

## How to verify

1. Run a guide with an interactive step whose selector cannot match (e.g. pointed at a URL with a query string) and let retries exhaust.
2. Inspect the outgoing Faro exception payload: `payload.value` should show `hostname/path` only, with no query string, token, or email.

## Breaking changes

None. This only changes what's redacted before send; no event/attribute shape changes.

## Reversibility

Fully reversible by revert — pure `beforeSend` content transform, no persisted state or contract dependency.

## Out of scope

- The `sequence_action_error` custom event itself — already fixed by #1347.
- General secret-pattern redaction (API keys not embedded in a URL) — this fix specifically closes the URL/query-string vector identified in review; a broader secret-scrubber would be new scope.

Fixes #1345 (exception/log path — the event-attribute path was already closed by #1347)

### DIFF
--- a/src/lib/faro.test.ts
+++ b/src/lib/faro.test.ts
@@ -196,13 +196,14 @@ describe('isGrafanaCloud', () => {
 
 function exceptionItem(
   filenames: Array<string | undefined>,
-  context?: Record<string, string>
+  context?: Record<string, string>,
+  value = 'boom'
 ): TransportItem<APIEvent> {
   return {
     type: 'exception',
     payload: {
       type: 'Error',
-      value: 'boom',
+      value,
       timestamp: new Date().toISOString(),
       stacktrace: { frames: filenames.map((filename) => ({ filename, function: 'fn' })) },
       ...(context && { context }),
@@ -211,10 +212,10 @@ function exceptionItem(
   } as unknown as TransportItem<APIEvent>;
 }
 
-function exceptionItemWithoutStacktrace(context?: Record<string, string>): TransportItem<APIEvent> {
+function exceptionItemWithoutStacktrace(context?: Record<string, string>, value = 'boom'): TransportItem<APIEvent> {
   return {
     type: 'exception',
-    payload: { type: 'Error', value: 'boom', timestamp: new Date().toISOString(), ...(context && { context }) },
+    payload: { type: 'Error', value, timestamp: new Date().toISOString(), ...(context && { context }) },
     meta: {},
   } as unknown as TransportItem<APIEvent>;
 }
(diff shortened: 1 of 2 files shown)


moxious:
@cursoragent review

moxious (approved):

Issue #1351, opened 2026-07-15
bug: "Full screen" button redirects to home instead of opening full screen (deep-link handler hijacks the handoff ?doc=)


moxious:
## Summary

Clicking **"Full screen"** in the sidebar docs panel does **not** open the guide in full screen. Instead the Pathfinder panel vanishes and the user is bounced back to the Grafana home page, with the plugin's internal mode left stuck on `fullscreen` (sidebar closed, no panel rendered anywhere).

This affects **every bundled interactive guide**, so the Full screen affordance is effectively broken for the default content.

## Steps to reproduce

1. Open Grafana with Pathfinder docked in the sidebar (e.g. the bundled **"Welcome to Grafana"** interactive guide).
2. Click the **"Full screen"** button in the panel header (`aria-label="Open in full screen"`).
3. Observe: the panel closes and the main view returns to the Grafana home page. Full screen never appears.

## Expected vs. actual

- **Expected:** the guide opens on the full-screen route (`/a/grafana-pathfinder-app/fullscreen?doc=…`).
- **Actual:** the app briefly navigates to the full-screen route, then immediately redirects to `/` (home). `grafana-pathfinder-app-panel-mode` in `localStorage` is left as `fullscreen` even though no full-screen surface is mounted — a stuck/broken state.

## Investigation — captured navigation trace

Instrumented `history.pushState`/`replaceState` and clicked "Full screen". Three history operations fire within ~30ms:

| # | op | target | origin |
|---|-----|--------|--------|
| 1 | `pushState` | `/a/grafana-pathfinder-app/fullscreen?doc=bundled:welcome-to-grafana&type=docs` | button `onClick` handler — **correct** |
| 2 | `replaceState` | **`/`** | **`module.js` → `locationService.replace`** — **the culprit** |
| 3 | `replaceState` | `/?orgId=1&from=now-6h&to=now&timezone=browser` | Grafana core re-normalizing home |

Step 1 is the correct handoff to the full-screen route. Step 2, ~15ms later, yanks the user off it and onto `/`.

## Root cause

The `?doc=` query param collides between two subsystems:

[...]

PR #1353, opened 2026-07-15
fix: resolve full-screen deep-link routing under Grafana subpaths

moxious:
## Intent

Apply the requested fix to PR #1353 and push it to the existing branch. Make the full-screen deep-link guard and adjacent panel-mode self-heal work when Grafana is served under config.appSubUrl by using Grafana's basename-normalized router pathname, while preserving exact full-screen route semantics and the existing root-deployment behavior. Add focused test-first coverage for direct handler, history-listener, and panel-mode paths; keep the change minimal, reversible, and compatible with supported Grafana versions.

## What Changed

- Reworked `handlePathfinderDeepLink` to check the full-screen route via `locationService.getLocation().pathname` (Grafana's basename-normalized router path) instead of raw `window.location.pathname`, and to build the full-screen redirect target from the shared `PLUGIN_BASE_URL`/`ROUTES.FullScreen` constants rather than `plugin.json` — so the sidebar→full-screen handoff `?doc=` is no longer hijacked and redirected to Home under `config.appSubUrl`.
- Narrowed the full-screen guard so it bails only when no pathfinder activation param needs handling (`panelMode`/`kiosk_session` still process), and moved that early return ahead of the dedup gate so a full-screen search string isn't latched against a later normal route.
- Applied the same router-normalized pathname check to the `usePanelMode` self-heal so persisted full-screen mode is corrected correctly on subpath deployments; added focused coverage in `pathfinder-deep-link-handler.test.ts` and `usePanelMode.test.ts`, plus a CHANGELOG entry and a bugfix-patterns note.

## Risk Assessment

✅ Low: Well-bounded, heavily test-covered bugfix that correctly normalizes route identity for subpath deployments and preserves prior root-deployment semantics; the only observation is a very narrow pre-existing dedup-latch edge case.

## Testing

Ran the 28 focused unit tests (all pass) and validated the fix end-to-end against a real Grafana 13.1.0 instance served under a `/grafana` subpath. Captured reviewer-visible screenshots showing the fixed build opening the full-screen guide at the subpath deep link, versus a pre-subpath-fix build (guard reverted to `window.location.pathname`) redirecting to Grafana Home — the exact issue #1351 symptom. Restored the fixed build afterward and confirmed recovery; root-deployment behavior is unchanged (a no-op at root since the two path sources are equal) and is covered by the existing unit tests. Test container and volume were removed and the worktree left clean.

- Evidence: Fixed build: full-screen guide opens under /grafana subpath (local file: <code>/var/folders/hg/34vhzxxj0xd92_m7gnjz5zrw0000gn/T/no-mistakes-evidence/01KXKT6QYRMGAJX5V2RZ1BGG7A/fixed-01-fullscreen-subpath.png</code>)
- Evidence: Pre-fix build: same deep link redirects to Grafana Home (issue #1351) (local file: <code>/var/folders/hg/34vhzxxj0xd92_m7gnjz5zrw0000gn/T/no-mistakes-evidence/01KXKT6QYRMGAJX5V2RZ1BGG7A/buggy-01-redirected-home-subpath.png</code>)

## Pipeline

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>
<summary>⚠️ **Review** - 1 info</summary>

- ⚠️ `src/utils/pathfinder-deep-link-handler.ts:62` - The new early-return in handlePathfinderDeepLink bails on the full-screen route for ANY pathfinder activation param (doc, panelMode, kiosk_session), but the comment says only &#34;FullScreenPanel owns ?doc= on this route.&#34; A URL arriving on the full-screen route with ?kiosk_session= (which sets window.__pathfinderKioskSessionId) or ?panelMode= is now silently dropped, whereas before it would have been processed. If that&#39;s intended (nothing should process while on the full-screen route), fine — but the guard is broader than its stated rationale and could regress kiosk-session capture on that route.
- ℹ️ `src/utils/pathfinder-deep-link-handler.ts:144` - needsRedirect still compares the basename-relative redirectTarget against raw window.location.pathname, which includes appSubUrl under a subpath deployment. This is the same raw-vs-normalized pathname mismatch this PR fixes elsewhere: under config.appSubUrl the comparison never matches even when already on the target page, causing a redundant locationService.replace to the current location. Pre-existing and not in the changed lines, but a residual gap in the file the PR is hardening for subpath deployments.

🔧 Fix: narrow full-screen guard to preserve kiosk_session/panelMode processing
1 warning still open:
- ⚠️ `package-lock.json:191` - This bugfix PR includes ~99 lines of package-lock.json churn (62 added &#34;peer&#34;: true metadata markers and removal of the @emnapi/core and @emnapi/runtime optional dependency entries) while package.json is unchanged. This is incidental lockfile regeneration from running npm install under a different npm version, not part of the deep-link fix. It mixes dependency-resolution drift into a focused fix, obscuring the diff and risking unrelated install-behavior changes; consider reverting the lockfile to keep the PR scoped, or split it into its own maintenance commit.

🔧 Fix: revert incidental package-lock.json drift to base
1 info still open:
- ℹ️ `src/utils/pathfinder-deep-link-handler.ts:71` - The early bail before the dedup gate only fires when BOTH panelMode and kiosk_session are absent. A full-screen-route URL carrying `?doc=X&amp;kiosk_session=Y` (kiosk present, no panelMode) skips the bail, latches `lastProcessedSearch` at line 78, and is NOT rewritten (doc/kiosk left intact for FullScreenPanel). If the user then lands on a normal route with the identical search string, the dedup gate at line 75 returns false and the `?doc=` is never dispatched there. The panelMode branches avoid this because rewriteCurrentUrl resets the latch to null; the kiosk-without-panelMode path does not. The scenario requires an exact search-string match across a route change, so it is very narrow — the comment at lines 67-69 reasoned about the latch only for the doc-only case. No action needed unless kiosk+doc handoffs are expected to survive onto a subsequent identical-query normal route.

</details>

<details>

✅ No issues found.
- <code>`CI=true npx jest src/utils/pathfinder-deep-link-handler.test.ts src/components/docs-panel/hooks/usePanelMode.test.ts` — 28 passed</code>
- <code>Built plugin (frontend + linux arm64/amd64 Go backend) and ran Grafana 13.1.0 in Docker under a subpath (`GF_SERVER_ROOT_URL=http://localhost:3000/grafana/`, `GF_SERVER_SERVE_FROM_SUB_PATH=true`)</code>
- <code>Fixed build: navigated to `/grafana/a/grafana-pathfinder-app/fullscreen?doc=bundled:welcome-to-grafana&amp;guideType=docs` → full-screen guide renders, URL + `?doc=` preserved (screenshot)</code>
- <code>Contrast: reverted only commit 9746d9d (guard on `window.location.pathname`), rebuilt, restarted → same URL redirects to `/grafana/?orgId=1...` (Grafana Home), reproducing issue #1351 (screenshot)</code>
- `Restored fixed sources + dist and re-verified the full-screen route holds; confirmed clean worktree and untouched foreign containers`

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>


moxious:
@cursoragent review

leslie-heinzen (dismissed):

Jayclifford345 (approved):

PR #1355, opened 2026-07-15
feat(video): self-hosted video assets + first-class Vimeo provider

### LINKED ISSUE #1352: feat(video): self-hosted video assets (relative/absolute URLs) + first-class Vimeo provider
## Summary

Pathfinder video blocks today assume public YouTube embeds. As we move toward private CDNs where guides host their own assets, authors need:

1. A **native `<video>` block** whose `src` can be either a fully-qualified URL (`https://foo.com/video.mp4`) or a path **relative to the guide's package root** (`assets/my-video.mp4`), resolved against the URL the guide was fetched from — mirroring how image blocks already work.
2. **First-class Vimeo support** as a hosted provider, on par with YouTube.

**Explicitly out of scope:** credential / auth passing for protected assets. This issue assumes the asset is reachable at a stable URL; auth is handled separately.

## Motivation

- YouTube embeds require assets to be public. Private / self-hosted content can't use them.
- Image blocks already resolve package-relative paths against the guide base URL; video is inconsistent — the same author mental model should apply to both.
- Field-eng custom guides and future private-CDN packages need to co-locate video with their `content.json`.
- Vimeo is already a common hosting choice and the security layer half-supports it (see below), but authors can't select it as a provider without hand-writing raw iframe HTML.

## Current behavior (as of `main`)

**Native / self-hosted video**

- `JsonVideoBlock` supports `provider: 'youtube' | 'native'`. Non-YouTube URLs already route to the native `<video>` renderer.
- `SafeUrlSchema` already accepts relative `src` values (protocol-checked against a dummy base), so no schema change is needed for authoring relative paths.
- **Gap 1 — `baseUrl` is dropped for video.** The JSON parser calls `convertImageBlock(block, path, baseUrl)` but `convertVideoBlock(block, path)` — omitting the `baseUrl` argument (`src/docs-retrieval/json-parser.ts:292` vs `:294`), so the rendered `video` element never receives a base to resolve against.
- **Gap 2 — `VideoRenderer` under-resolves.** It only rewrites `src` values beginning with `/` and has no synthetic-base fallback (`src/docs-retrieval/components/docs/video-renderer.tsx:51-68`), unlike `ImageRenderer` (`src/docs-retrieval/components/docs/image-renderer.tsx:29-63`), which already handles bare-relative (`assets/x.mp4`), root-absolute (`/media/x.mp4`), protocol-relative (`//`), `data:`, absolute, and synthetic bases (`block-editor://`, `bundled:`, falling back to `https://grafana.com/`).

Net: a fully-qualified `https://foo.com/video.mp4` on the native provider already renders today; package-relative paths do **not**.

**Vimeo**

- `isVimeoDomain` exists (`src/security/url-validator.ts:220`) and the HTML sanitizer already permits Vimeo iframes (`player.vimeo.com`, `vimeo.com`, `www.vimeo.com`, `f.vimeocdn.com`) with the same non-sandboxed treatment as YouTube (`src/security/html-sanitizer.ts`).
- **But there is no first-class Vimeo provider.** The `provider` enum is only `youtube | native`; `convertVideoBlock` only detects YouTube, so a Vimeo URL falls through to native `<video>` and breaks (a `vimeo.com/<id>` URL is a page, not a media file). There is no URL normalization (`vimeo.com/<id>` → `player.vimeo.com/video/<id>`) and no editor provider option. Vimeo works only if an author hand-writes a raw iframe HTML block.

## Base URL plumbing already exists

`RawContent.url` holds the fetched guide URL and flows to the parser as `baseUrl` — already consumed by images. For a CDN package, `new URL('assets/my-video.mp4', '.../pkg/content.json')` resolves to `.../pkg/assets/my-video.mp4`, exactly the package-relative behavior we want.

## Proposed change

**A. Self-hosted / native video (relative + absolute URLs)**

1. Thread `baseUrl` into the native branch of `convertVideoBlock` (`json-parser.ts`), matching `convertImageBlock`. Leave the YouTube branch unchanged (embed URLs are always absolute).
2. Refactor `VideoRenderer`'s `resolvedSrc` to reuse the same resolution logic as `ImageRenderer` — ideally extract a shared `resolveAssetUrl(src, baseUrl)` helper so image and video can't drift:
   - absolute (`http(s)://`, `//`, `data:`) → pass through unchanged
   - relative (`assets/x.mp4`, `/media/x.mp4`) → `new URL(src, effectiveBaseUrl)`
   - synthetic base (`block-editor://`, `bundled:`) → fall back to `https://grafana.com/` as images do
3. Editor / CLI parity: ensure `VideoBlockForm` and CLI input normalizers accept relative paths for the native provider without forcing YouTube normalization (`src/cli/utils/input-normalizers.ts`, `src/components/block-editor/forms/VideoBlockForm.tsx`).

**B. First-class Vimeo provider**

4. Add `'vimeo'` to the `provider` enum (`JsonVideoBlock` type + `JsonVideoBlockSchema`).
5. Route Vimeo in `convertVideoBlock`: detect `provider === 'vimeo'` (or a Vimeo host) and produce a Vimeo embed element. Reuse or generalize the YouTube iframe renderer into a shared embed renderer rather than duplicating (`src/docs-retrieval/components/docs/youtube-video-renderer.tsx`).
6. Normalize Vimeo input URLs to the canonical `player.vimeo.com/video/<id>` embed form in the CLI normalizer, matching how YouTube watch/short/shorts URLs are canonicalized today.
7. Add Vimeo to the editor provider options (`VideoBlockForm` / block-editor constants) with the same preview affordance YouTube has.
8. Confirm the sanitizer / `isVimeoDomain` allowlist covers the embed form produced (it already permits `player.vimeo.com`).

## Security / open questions (open for discussion — not yet decided)

- **CSP `media-src`.** A fully-qualified external `src` makes the browser fetch media from an arbitrary origin. Native `<video>` doesn't execute scripts (lower risk than the iframe path), but plugin CSP may block off-origin `media-src`. Do we allow arbitrary absolute media URLs, or restrict to same-origin / an allowlist (as we do for YouTube/Vimeo iframes)?
- **Mixed content & privacy.** Off-origin media leaks a referrer and can be a tracking vector; consider `crossorigin` / referrer handling for parity with our iframe hardening.
- **Possible posture.** One option: relative paths (resolved against the trusted guide base) are unrestricted; fully-qualified cross-origin media URLs go through a host allowlist. Left open here for discussion.

## Acceptance criteria

- A native video block with `src: "assets/my-video.mp4"` renders and plays, resolved against the guide's fetched base URL (CDN package, bundled, and block-editor preview cases).
- A native video block with `src: "https://foo.com/video.mp4"` renders and plays (subject to the CSP decision above).
- A video block with `provider: 'vimeo'` (or a Vimeo URL) renders a working Vimeo embed, with input URL normalization to the canonical embed form.
- `start` / `end` continue to work for native video.
- Image and video share one resolution code path (regression-tested) so they can't diverge.
- Unit tests mirror `image-renderer.test.tsx` for the video resolver; e2e coverage for a package-relative video and a Vimeo block.

## Non-goals

- Authenticated / credentialed asset fetching.
- HLS/DASH adaptive streaming, poster images, captions / subtitle tracks — trackable as follow-ups.

moxious:
## Summary

Video blocks can now point at self-hosted assets instead of only public YouTube embeds, and Vimeo becomes a first-class provider alongside YouTube:

- **Native video** `src` may be a fully-qualified URL (`https://foo.com/video.mp4`) **or** a path relative to the guide's package root (`assets/demo.mp4`), resolved against the URL the guide was fetched from — the same model image blocks already use.
- **Vimeo** is now a real provider (`provider: 'vimeo'`, or any `vimeo.com` URL): it renders through a dedicated player embed instead of only working when an author hand-writes a raw iframe.

## What to look at

- **`src/docs-retrieval/components/docs/resolve-asset-url.ts`** — new shared `resolveAssetUrl(src, baseUrl)` helper. Both [`image-renderer.tsx`](https://github.com/grafana/grafana-pathfinder-app/blob/af59c55c37b57f0b799103e1e24ece77a5ff20c8/src/docs-retrieval/components/docs/image-renderer.tsx) and [`video-renderer.tsx`](https://github.com/grafana/grafana-pathfinder-app/blob/af59c55c37b57f0b799103e1e24ece77a5ff20c8/src/docs-retrieval/components/docs/video-renderer.tsx) now call it, so image and video resolution can't drift. Note the behavior change for video: it previously resolved **only** `/`-prefixed paths; it now also handles bare-relative (`assets/x.mp4`) and synthetic bases (`block-editor://`, `bundled:` → fall back to `https://grafana.com/`), matching images.
- **[`src/docs-retrieval/json-parser.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/af59c55c37b57f0b799103e1e24ece77a5ff20c8/src/docs-retrieval/json-parser.ts)** — `convertVideoBlock` now receives `baseUrl` (it was the one media branch that dropped it) and routes to three element types: `youtube-video`, the new `vimeo-video`, or native `video` (which carries `baseUrl`).
- **[`src/docs-retrieval/components/docs/vimeo-video-renderer.tsx`](https://github.com/grafana/grafana-pathfinder-app/blob/af59c55c37b57f0b799103e1e24ece77a5ff20c8/src/docs-retrieval/components/docs/vimeo-video-renderer.tsx)** — new renderer. It builds a canonical `player.vimeo.com/video/<id>` embed, preserves the unlisted `h` hash, and applies a start offset via a media fragment. **Security-relevant:** this iframe is created directly in React (it does not pass through the HTML sanitizer), so the renderer gates on the existing `isVimeoDomain` allowlist and returns a fallback message for anything else — worth a close look.
- **[`src/docs-retrieval/html-parser.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/af59c55c37b57f0b799103e1e24ece77a5ff20c8/src/docs-retrieval/html-parser.ts)** — `<iframe>` with a Vimeo src now maps to `vimeo-video` (previously only YouTube was special-cased; Vimeo fell through to a raw iframe).
- **[`src/cli/utils/input-normalizers.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/af59c55c37b57f0b799103e1e24ece77a5ff20c8/src/cli/utils/input-normalizers.ts)** — Vimeo watch URLs normalize to the player embed form, mirroring the existing YouTube normalization (soft `INPUT_NORMALIZED` warning, never a hard fail).
- **Type/schema + editor** — `provider` enum gains `'vimeo'` in [`json-guide.types.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/af59c55c37b57f0b799103e1e24ece77a5ff20c8/src/types/json-guide.types.ts) / [`json-guide.schema.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/af59c55c37b57f0b799103e1e24ece77a5ff20c8/src/types/json-guide.schema.ts); the block-editor video form gains a Vimeo option and a relative-path hint.
- **`src/docs-retrieval/__fixtures__/video-showcase.content.json` + `video-showcase.fixture.test.tsx`** — a test-only raw content.json exercising all four forms (YouTube, Vimeo, native-absolute, native-relative). The test parses it with a simulated CDN base URL and asserts routing plus end-to-end package-relative resolution.

## Why

YouTube/Vimeo embeds require assets to be publicly hosted. As Pathfinder moves toward private CDNs where guides host their own assets, authors need a video block that can reference a self-hosted file by stable URL or by a path relative to the guide package — without the asset being public. Images already resolve package-relative paths; video was the inconsistent case (the parser silently dropped `baseUrl` for video, and the renderer only handled root-absolute paths). This PR closes that gap and, while in the video path, promotes Vimeo from "works only via raw iframe HTML" to a first-class provider. See #1352 for the full motivation.

**Reviewer gotcha:** `SafeUrlSchema` on `src` looks like it would reject a relative path, but it validates the protocol via `new URL(url, 'https://example.com')`, so `assets/demo.mp4` resolves to an `https:` URL and passes — no schema change was needed to accept relative native sources.

## How to verify

1. Author a native video block with `src: "assets/demo.mp4"` in a guide served from a package (or block-editor preview) and confirm the rendered `<video>` points at the package-relative URL, not the bare string.
2. Author a native video block with a fully-qualified `https://…/video.mp4` and confirm it plays. If your Grafana instance sets a restrictive CSP `media-src`, expect the browser to block off-origin media — this is the open security question below.
3. Author a Vimeo block three ways — `provider: 'vimeo'` with a raw `vimeo.com/<id>` URL, a bare `vimeo.com/<id>` URL with no provider, and an unlisted `vimeo.com/<id>/<hash>` URL — and confirm each renders a working player and that the unlisted hash reaches the embed.
4. In the block editor, add a video block, switch the provider dropdown to Vimeo, and confirm the URL hint/placeholder update.
5. Confirm a bogus Vimeo-looking URL (e.g. `https://evil.com/vimeo.com/123`) renders the "Invalid video URL provided" fallback rather than an iframe.

## Breaking changes

None. The change is additive — the `provider` enum only gains a value, and existing YouTube/native blocks render exactly as before. The one behavior change is that native video now resolves relative paths it previously left untouched; a guide relying on an unresolved relative `src` would have been broken already.

## Out of scope

- **Credentialed/authenticated asset fetching** — deferred; this PR assumes the asset is reachable at a stable URL.
- **Cross-origin media security posture** — whether fully-qualified external media URLs should be unrestricted or run through a host allowlist (and CSP `media-src` implications) is intentionally left **open for discussion on #1352**. Native `<video>` does not execute scripts, so it is lower-risk than the iframe path, but the decision is deliberately not baked in here.
- **Vimeo playback analytics** — the YouTube renderer tracks play/view-length via the YT iframe API; the Vimeo renderer does not yet. Trackable as a follow-up.
- **Poster images, captions/subtitle tracks, HLS/DASH streaming** — follow-ups.
- **Local `assets/` in *bundled* guides** — deliberately not supported. Bundled guides receive a synthetic `bundled:` base URL (so relative paths fall back to `https://grafana.com/`), and the webpack pipeline emits no `.mp4`/`.webm` (`asset/resource` is images + fonts only). Package-relative assets resolve only for **remote/CDN** guides, whose base URL is the real content.json URL. This is why the fixture is test-only and driven with an http base rather than shipped as a `bundled:` guide.

Closes #1352


moxious:
@cursoragent review

tomglenn (approved):

PR #1356, opened 2026-07-15
fix(telemetry): derive panel-ready readiness from scene state, not a one-shot event

### LINKED ISSUE #1342: telemetry: panel-ready measurement can lose its readiness event — derive from state instead (follow-on to #1338)
Follow-on from #1338, which anchored the panel LCP clock at first render and completes the panel-ready measurement when recommendations are ready.

## Problem

The readiness signal is a two-phase CustomEvent with no replay:

- Dispatch: `src/components/docs-panel/context-panel.tsx` (~line 1128) — `ContextPanelRenderer` dispatches `RECOMMENDATIONS_READY_EVENT` via `queueMicrotask` when `recommendationsScrollReady` becomes true.
- Listen: `src/components/docs-panel/hooks/usePanelReadyMeasurement.ts` (~line 18) — the hook (mounted in the parent `CombinedPanelRendererInner`) registers a window listener in a mount-once effect, initializes `recommendationsReady = false`, and only flips it in the listener callback. There is no mount-time check for an already-fired event.

If the dispatch ever beats listener registration (recommendations resolving synchronously from cache is the risky path), the event is lost and the panel-ready measurement silently never records for that session. The `queueMicrotask` deferral mitigates this — React flushes all synchronous effects before microtasks run — and the code comment at the dispatch site acknowledges the race it's defending against. But the correctness of the measurement rests on effect-flush timing rather than on state, and `ContextPanelRenderer` is known to remount on tab changes (the reason the auto-open listener was previously moved to the parent), so the timing assumption is fragile under refactoring.

## Fix

Drop the CustomEvent indirection. The hook already receives `hasContent` and `isRecommendationsTab`; pass (or derive) recommendations readiness as a prop/state so the measurement completes from observable state on any render, regardless of mount order. State can't be "missed" the way a one-shot event can.

Impact if unfixed: telemetry-only (lost `panel_ready` measurements, biased toward cache-hit sessions — the fastest ones, skewing the metric). No user-facing behavior is affected.

moxious:
## Summary

Follow-on to #1338. The panel-ready (`panel_lcp_ms`) measurement's readiness signal was a two-phase `RECOMMENDATIONS_READY_EVENT` CustomEvent with no replay. If the dispatch ever beat listener registration — the risky path when recommendations resolve synchronously from cache — the event was lost and `panel_ready` silently never recorded for that session, biasing the metric toward the fastest, cache-hit sessions. The `queueMicrotask` deferral only mitigated this by leaning on effect-flush timing, which is fragile given `ContextPanelRenderer` remounts on tab changes.

## Change

Drop the CustomEvent indirection in favor of replayable observable state:

- `ContextPanelRenderer` writes `recommendationsReady: true` onto the `ContextPanel` scene object's state when `recommendationsScrollReady` becomes true (idempotent).
- `CombinedPanelRendererInner` reads it via `contextPanel.useState()` and passes `recommendationsReady` into `usePanelReadyMeasurement`.
- `usePanelReadyMeasurement` accepts `recommendationsReady` as a prop; the window-listener effect and the local event state are removed.
- `RECOMMENDATIONS_READY_EVENT` is deleted from `event-names.ts` (no other consumers — grep-confirmed).

Scene state is observable on any render regardless of mount order, so it survives `ContextPanelRenderer` remounting and cannot be "missed" the way a one-shot event can. The LCP-anchoring semantics from #1338 are preserved — the measurement still completes exactly once when recommendations become ready.

Telemetry-only; no user-facing behavior changes.

## Tests

`usePanelReadyMeasurement.test.ts` updated to drive readiness via the prop, adding explicit coverage for the previously-racy path (recommendations already ready at mount / synchronous cache hit) and for readiness being ignored when the recommendations tab is not active. `npm run typecheck` and the docs-panel test suites pass.

Closes #1342

### DIFF
--- a/src/components/docs-panel/context-panel.tsx
+++ b/src/components/docs-panel/context-panel.tsx
@@ -18,7 +18,6 @@ import { useContextPanel, Recommendation } from '../../context-engine';
 import type { ResolvedNavLink } from '../../types/context.types';
 import { reportAppInteraction, UserInteraction, getContentTypeForAnalytics } from '../../lib/analytics';
 import { logger } from '../../lib/logging';
-import { RECOMMENDATIONS_READY_EVENT } from '../../lib/event-names';
 import { getConfigWithDefaults, PLUGIN_BASE_URL } from '../../constants';
 import { isDevModeEnabled } from '../../utils/dev-mode';
 import { testIds } from '../../constants/testIds';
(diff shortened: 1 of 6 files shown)


moxious:
@cursoragent review

Jayclifford345 (approved):

PR #1358, opened 2026-07-15
test(architecture): file-level circular-dependency ratchet

moxious:
## What this does

Extends the architecture invariant tests (`src/validation/architecture.test.ts`) with **general, file-level circular-dependency detection** over the whole import graph — complementing the existing Tier-2-only lateral-cycle check — plus **two sample paydown refactors** and **anti-gaming safeguards** on the allowlist.

## Mechanism

- `buildModuleGraph()` reuses the existing AST import scanner to build a file-level graph over production source (test files, `.d.ts`, and excluded dirs dropped; edges closed over real nodes).
- `findStronglyConnectedComponents()` — iterative Tarjan (stack-safe on the ~600-node graph). `findCycles()` returns SCCs of size ≥ 2.
- New ratchet test with an `ALLOWED_CYCLES` allowlist — the same shrink-only mechanism the tier/lateral/barrel checks already use. Counts **all** imports (including `import type`). An informational test logs the full footprint on every run.

## Footprint: 7 → 5 clusters

Baseline was `clusters=7 filesInCycles=33 largestCluster=9`. This PR pays down two, leaving **5** grandfathered in `ALLOWED_CYCLES`, tracked for opportunistic paydown in #1359.

## Sample paydown patterns (copy these)

**1. Extract the shared type** — `global-state/link-interception` ↔ `utils.link-interception`. The closing edge was a shared `QueuedDocsLink` type; moved to `src/types/link-interception.types.ts` (Tier 0). *Use when the back-edge is a type/interface.*

**2. Dependency injection** — `lib/storage/bounded-record-storage` ↔ `lib/user-storage`. The low-level building block imported the higher-level `user-storage` for its backend + quota notifier; now `createBoundedRecordStorage` receives them by injection from its two callers. *Use when a lower module reaches up to a higher one.* (A re-export shim can't break this cycle — it re-forms through `bounded-record-storage`; and moving the completion instances would churn 8 consumers + ~16 test mocks. Injection touches only the 2 files in the cycle.)

## Anti-gaming safeguards on the allowlist

The failure hands agents an escape hatch ("add its key to ALLOWED_CYCLES"), and the shrink-only ratchet doesn't stop *additions*. Two guards make silent rubber-stamping hard:

**A. Enforced justification + tracking.** `ALLOWED_CYCLES` entries are now `{ cycle, reason, tracking }`. A sibling test (`validateAllowedCycleEntries`, pure + unit-tested) fails unless every entry has a substantive `reason` and a `tracking` issue (`#1234` or a GitHub issues URL) — so a new cycle can't be silenced with an empty comment.

**B. Guidance toward the fix.** The new-cycle failure message now points at the two worked examples above and notes the enforced fields, steering agents to break the cycle rather than allowlist it.

> These raise the bar; they don't make the allowlist unforgeable (an agent can still write a plausible fake reason). The stronger mechanism-level lock — CODEOWNERS on `architecture.test.ts` or a frozen grandfathered-vs-new split — is deferred for separate discussion.

## Testing

- SCC algorithm: unit tests for simple/long/nested/independent cycles + deep-chain stack-safety.
- `validateAllowedCycleEntries`: unit tests for short/blank reason, bad tracking (incl. rejecting a PR URL), duplicate keys, valid `#`/URL forms.
- Building-block test wires the real backend via a `makeStore` helper — behavior unchanged.
- Full suite green (347 suites / 5416 tests); lint, prettier, typecheck clean.

### DIFF
--- a/src/global-state/link-interception.ts
+++ b/src/global-state/link-interception.ts
@@ -1,12 +1,7 @@
 import { getDocsLinkFromEvent } from 'global-state/utils.link-interception';
 import { sidebarState } from 'global-state/sidebar';
 import { reportAppInteraction, UserInteraction } from 'lib/analytics';
-
-export interface QueuedDocsLink {
-  url: string;
-  title: string;
-  timestamp: number;
-}
+import type { QueuedDocsLink } from 'types/link-interception.types';
 
 /**
  * Global state manager for the Pathfinder plugin's link interception.
--- a/src/global-state/utils.link-interception.ts
+++ b/src/global-state/utils.link-interception.ts
@@ -1,4 +1,4 @@
-import { QueuedDocsLink } from 'global-state/link-interception';
+import type { QueuedDocsLink } from 'types/link-interception.types';
 import { isAllowedContentUrl, isLocalhostUrl, isGitHubRawUrl } from 'security/url-validator';
 import { isDevModeEnabledGlobal } from '../utils/dev-mode';
 
--- a/src/lib/storage/bounded-record-storage.test.ts
+++ b/src/lib/storage/bounded-record-storage.test.ts
@@ -1,12 +1,18 @@
 import { getAppEvents } from '@grafana/runtime';
 
-import { __resetQuotaWarningForTests } from '../user-storage';
-import { createBoundedRecordStorage } from './bounded-record-storage';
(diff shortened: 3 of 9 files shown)


moxious:
@cursoragent review

leslie-heinzen (approved):

PR #1360, opened 2026-07-16
fix(dom): escape CSS attribute values and split grafana: params at first colon

jackw:
## What

Fixes the two selector bugs deferred during the #1156 review, plus the wider unescaped-interpolation class (scope confirmed in plan review):

1. **Colon-split**: `resolveSelector` split `grafana:path:param` at the *last* colon, so parameters containing colons (e.g. a breadcrumb `Prod: Overview`) mis-split, threw inside `toGrafanaSelector`, and silently resolved to nothing at replay. Selector paths are dot-separated and never contain colons, so the first colon is unambiguously the separator.

2. **CSS attribute-value escaping**: live-DOM attribute values were interpolated unescaped into quoted CSS attribute selectors at ~20 sites (`toGrafanaSelector`, `resolvePanelSelector`, every candidate/scope builder in `selector-generator.ts`, `enhanced-selector.ts`'s prefix fallback, `action-detector.ts`, `navigation-manager.ts`). A value containing a quote produced malformed CSS that silently matched nothing — the picker degraded quote-bearing titles to positional `nth-of-type` selectors, and a crafted value like `x'], [data-foo='y` could rewrite selector structure. Fixed with a new shared `escapeCssAttributeValue` (escapes backslash + the delimiting quote, per CSS string syntax).

Out of scope: `:text('…')`/`:contains('…')` pseudo arguments stay raw — their parser in `enhanced-selector.ts` reads them with a regex that cannot handle backslash escapes.

## Notes for review

- The `jest.mock` in `selector-resolver.test.ts` was removed: it demonstrably never applied (its default branch *returns* instead of throwing, which the old multi-colon test's pass depended on), so the suite already exercised the real `toGrafanaSelector`. The multi-colon test that encoded the last-colon bug is rewritten against the real parameterized `components.Breadcrumbs.breadcrumb` selector.
- First test coverage for `resolvePanelSelector` (basic, child selector, quote-bearing title).
- New end-to-end picker test showed the escaper restoring a `grafana:` selector that previously died at the uniqueness gate: an aria-label `Delete Mark's panel` now correctly reverse-matches the legacy `${title} panel` template and survives resolution.

## Verification

- `npm run test:ci` — 338 suites / 5278 tests green
- `npx tsc --noEmit` clean; lint 0 errors (23 pre-existing warnings in unrelated files)

Prerequisite for the upcoming `{grafana:...}` token-grammar work (making #1156 + #1157 compose).

### DIFF
--- a/src/integrations/workshop/action-replay.ts
+++ b/src/integrations/workshop/action-replay.ts
@@ -15,6 +15,7 @@ import type {
   InteractiveAction,
 } from '../../types/collaboration.types';
 import { isCssSelector } from '../../lib/dom/selector-detector';
+import { escapeCssAttributeValue } from '../../lib/dom/css-escape';
 import { querySelectorAllEnhanced } from '../../lib/dom/enhanced-selector';
 import { logger } from '../../lib/logging';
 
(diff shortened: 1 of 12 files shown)


Jayclifford345:
@cursoragent, review this PR.

Jayclifford345 (approved):

PR #1361, opened 2026-07-16
feat(dom): compose grafana: selectors with ancestor/descendant anchoring

jackw:
## What

Makes #1156 (version-stable `grafana:` selectors) and #1157 (stable-ancestor / `:has()` descendant anchoring) compose. Before this, the anchoring machinery pinned raw testid literals — the exact values #1156 exists to avoid — and non-unique `grafana:` candidates were dropped entirely. Plan reviewed via crit (5 rounds); built on the merged escaping/colon-split PR #1360.

One commit per step, reviewable independently:

1. **`:is()` keystone** — `toGrafanaSelector` returns `:is([data-testid=V], [aria-label=V])` instead of a top-level comma list, so the resolved form can be scoped, prefixed, and embedded. Values are quoted via `toCssAttributeString`, which picks the quote style avoiding escapes: nwsapi (jsdom) silently drops matches for backslash-escaped quotes inside `:is()`, and the codebase's selector-parsing regexes (`[^'"]+`) mis-capture them.
2. **`{grafana:path[:param]}` token grammar** — a pre-pass in `resolveSelector` resolves embedded tokens anywhere in a CSS string. All replay consumers already funnel through `resolveSelector`, so stored token reftargets version-resolve with no consumer changes. **Forward-compat:** plugin versions older than this change treat tokens as literal CSS (match nothing) when replaying newly-authored guides.
3. **Scoped grafana candidates** — a non-unique `grafana:` candidate re-emits as `div[data-testid='panel-B'] {grafana:components.RefreshPicker.runButtonV2}` (method `scoped-grafana`, score 12) instead of being dropped; no positional hybrids. Raw testid stays available in fallbacks.
4. **Version-stable anchors** — `findAllAncestorScopes` emits `{grafana:...}` token scopes when the ancestor reverse-maps (penalty 45, just under `parentTestId`), and `bestStableDescendant` embeds the grafana identity it previously computed and discarded: `div:has({grafana:components.Breadcrumbs.breadcrumb:Home})`.
5. **Admission-time `:has()` rescue** — identity candidates that die at admission (e.g. a solitary non-unique grafana candidate with no scoping ancestor) now retry descendant anchoring instead of degrading to positional `nth-of-type`.
6. **Badge resolvability validation** — `analyzeSelectorString` actually resolves grafana identities; a typo'd path now reads poor/0 with "Unknown grafana selector path for this Grafana version" instead of good/100. Parameterized grafana/`panel:` titles get the `environment-unstable` flag (free-text instance content). Testid-subject `:has()` selectors classify by subject. This check immediately caught the old pinned test and two docstrings referencing `components.RefreshPicker.runButton` — a path that doesn't exist in the resolved tree (only `runButtonV2` does).

## Verification

- `npm run check` fully green (5463 tests, 349 suites, Go, lint, prettier, coverage thresholds).
- New tests pin: token round-trip (params with colons, quotes), scoped-grafana emission with duplicated components, drop-when-unscopable, token ancestor scopes, token `:has()` anchors, admission-time rescue, badge poor-rating for unresolvable paths/tokens, instance-fragility flags.
- Suggested manual pass (`npm run server`): pick a Select/RefreshPicker inside one of several identical panels — primary selector should keep a `{grafana:...}` identity; type a bogus `grafana:` path in the block editor — badge should go red; replay a token selector with "Do it".

### DIFF
--- a/src/lib/dom/css-escape.test.ts
+++ b/src/lib/dom/css-escape.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from '@jest/globals';
-import { escapeCssAttributeValue } from './css-escape';
+import { escapeCssAttributeValue, toCssAttributeString } from './css-escape';
 
 describe('escapeCssAttributeValue', () => {
   it('returns plain values unchanged', () => {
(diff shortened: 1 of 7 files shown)


Jayclifford345:
@cursoragent review this pr

Jayclifford345 (approved):

PR #1362, opened 2026-07-16
fix(telemetry): align Faro event/action names with the analytics.ts convention

Jayclifford345:
## Summary

Two distinct naming problems in the Faro-only telemetry path, both traced back to the `analytics.ts` convention (`pathfinder_<snake_case>`, applied via `createInteractionName`):

- **Missing prefix**: `TELEMETRY_EVENTS` in `src/lib/telemetry/types.ts` defined four Faro-only event names (`recommender_fallback`, `content_fetch_fallback`, `requirements_exhausted`, `sequence_action_error`) with no `pathfinder_` prefix — dropped when PR #1338 split them off the RudderStack pipeline into raw `pushFaroEvent(...)` calls. These stay Faro-only (that split was a deliberate, documented decision); this PR only adds the missing prefix.
- **Invented duplicate ("ghost") names**: five `withFaroUserAction(...)` call sites invented their own `pathfinder_`-prefixed name from scratch, even though the *same click* is already reported through `reportAppInteraction`/`UserInteraction` a few lines away:
  - `pathfinder_step_do` / `pathfinder_step_show` (`interactive.hook.ts`) → real name is `UserInteraction.DoItButtonClick` / `ShowMeButtonClick`
  - `pathfinder_section_run` (`interactive-section.tsx`) → real name is `UserInteraction.DoSectionButtonClick`
  - `pathfinder_guided_step` (`guided-handler.ts`) → real name is `UserInteraction.DoItButtonClick` (fired by `interactive-guided.tsx`'s `handleDoAction`)
  - `pathfinder_guide_open` (`facade.ts`'s `TELEMETRY_ACTIONS.guideOpen`) → the generic `UserInteraction.DocsPanelInteraction` action-bucket pattern (`{ action, source, ... }`) already used across the panel is the right fit; retargeted with `action: 'open_guide'`

`createInteractionName` is now exported from `analytics.ts` so these call sites can reuse the real name instead of inventing one.

One name investigated and found **not** to be a duplicate: `pathfinder_remote_step` (`live-tab-executor.ts`) times the two-tab controller's live-tab relay execution — the click is already reported as a real interaction on the *control* tab; this measures relay/execution latency on the *live* tab, a genuinely different signal. Moved into the `TELEMETRY_ACTIONS` registry instead of staying an inline literal, same treatment as the `TELEMETRY_EVENTS` funnel signals.

Sharing the real interaction name also means the fully-attributed `analytics.ts` Faro mirror (`step_id`, `section_id`, `current_step`, `completion_percentage`, etc. — flattened key-by-key by `stringifyAttributes`, not a nested JSON blob) is now discoverable under the same name as the engine's duration-tracked span, so these actions are searchable in Faro the same way they already are in RudderStack — no attribute mapping/threading needed, since that mirror already sends the full properties object.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run check` (348 suites, 5426 tests, all passing)
- [x] Grep sweep confirms no remaining un-prefixed/invented literal outside test fixtures and the unrelated `LoopExitReason`/`outcome-classifier.ts` vocabulary
- [ ] Manual sanity in a dev Grafana: trigger a "Show me"/"Do it" step, a guided step, a section run, and a guide open; confirm the Faro Frontend Observability "User actions" view shows the renamed actions

## Breaking changes

None for product behavior. Grafana-internal Faro event names change for 5 actions (`pathfinder_guide_open`→`pathfinder_docs_panel_interaction`, `pathfinder_step_do`→`pathfinder_do_it_button_click`, `pathfinder_step_show`→`pathfinder_show_me_button_click`, `pathfinder_guided_step`→`pathfinder_do_it_button_click`, `pathfinder_section_run`→`pathfinder_do_section_button_click`) plus a prefix added to 4 more (`pathfinder_recommender_fallback` etc.). All of this telemetry is very recent (PR #1338, merged 2026-07-15) — low blast radius for any existing dashboard/alert.

### DIFF
--- a/src/components/docs-panel/docs-panel.load-tab-content.test.ts
+++ b/src/components/docs-panel/docs-panel.load-tab-content.test.ts
@@ -89,7 +89,8 @@ jest.mock('../../lib/user-storage', () => ({
 jest.mock('../../lib/analytics', () => ({
   setupScrollTracking: jest.fn(),
   reportAppInteraction: jest.fn(),
-  UserInteraction: {},
+  createInteractionName: jest.fn((type: string) => `pathfinder_${type}`),
+  UserInteraction: { DocsPanelInteraction: 'docs_panel_interaction' },
   getContentTypeForAnalytics: jest.fn(),
 }));
 
--- a/src/components/docs-panel/docs-panel.tab-restore-guard.test.ts
+++ b/src/components/docs-panel/docs-panel.tab-restore-guard.test.ts
@@ -94,7 +94,8 @@ jest.mock('../../lib/user-storage', () => ({
 jest.mock('../../lib/analytics', () => ({
   setupScrollTracking: jest.fn(),
   reportAppInteraction: jest.fn(),
-  UserInteraction: {},
+  createInteractionName: jest.fn((type: string) => `pathfinder_${type}`),
+  UserInteraction: { DocsPanelInteraction: 'docs_panel_interaction' },
   getContentTypeForAnalytics: jest.fn(),
 }));
 
--- a/src/components/interactive-tutorial/interactive-section.runner.tripwire.test.tsx
+++ b/src/components/interactive-tutorial/interactive-section.runner.tripwire.test.tsx
@@ -198,7 +198,7 @@ describe('handleDoSection — Phase 0 tripwire (Tier C gate)', () => {
     });
   });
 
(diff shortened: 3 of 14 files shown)


Jayclifford345:
@cursoragent review

moxious (approved):

PR #1363, opened 2026-07-16
fix(analytics): standardize content_type/link_type attribute values

Jayclifford345:
## Summary

Analytics attributes like `content_type` and `link_type` had no shared vocabulary — every `reportAppInteraction` call site wrote its own literal string, so the same concept (e.g. "this is an interactive guide") ended up with multiple, diverging values depending on which button fired the event.

- Adds `AnalyticsContentType` and `AnalyticsLinkType` enums to `src/lib/analytics.ts`, plus a `tabTypeToContentType()` helper that centralizes the tab-type → content-type mapping.
- Migrates every `reportAppInteraction` call site off raw `content_type`/`link_type` string literals onto the new enums.
- Fixes three real bugs surfaced while auditing this (not just style):
  - `content_type` diverged between `'interactive'` and the canonical interactive-guide value for the *same* guide kind, depending on which button opened it. Retyping `getContentTypeForAnalytics` forced the compiler to surface **8 call sites** with this bug (2 found by manual audit, 6 more sibling sites in `context-panel.tsx` sharing the same `Recommendation['type']` pattern).
  - `source_page` held a UI-location string instead of a URL at one call site in `DocsPanelContentArea.tsx` (every sibling call site holds a URL).
  - Four dead `blockType === 'code'` comparisons in `AssistantBlockWrapper.tsx` — the canonical `BlockType` union only has `'code-block'`, so this branch could never match.

## Explicit call-outs

- **Breaking analytics-value change:** the interactive-guide `content_type` value is renamed from the pre-existing production value `interactive_guide` (underscore) to `interactive-guide` (hyphen), to match the separator convention already used by `learning-journey`/`package-nav-link`. This changes what gets sent to RudderStack/Faro going forward for that value — any saved dashboard/query filtering on the old underscore value needs updating. Every other pre-existing value was left as-is to avoid forking historical data.
- **Deferred structural refactor:** `analytics.ts` is still a flat file of exports, unlike `telemetry/` which has a proper folder + class structure (bridge/adapter/session/url). We should probably do the same for analytics eventually — deferring that to a separate PR since it's independent of this value-standardization work.
- **Deferred attribute cleanup (follow-up PR):** unifying `interaction_location`/`source`(menu-item)/`close_location` into one key name; splitting the overloaded `target_action` field (DOM verb vs. composite block-kind) and reconciling its two divergent TS unions; `action` key sharing unrelated vocabularies; `panel_type`, `trigger`/`trigger_source`, `ref_target`/`reftarget` naming drift. Scope was bounded to the analytics-value boundary here, not the underlying app-wide type unions that already disagree with each other.

## Test plan

- [x] `npx tsc --noEmit` — clean (the `getContentTypeForAnalytics` signature change forced every non-enum call site to surface)
- [x] `npm run lint` — no new errors (pre-existing `getDataSourceSrv` deprecation warnings only)
- [x] `npx prettier --check` on all touched files
- [x] `npm run test:ci` — 349/349 suites, 5441 passed
- [ ] Manual smoke test in a running Grafana instance not done here (no live dev server); this change only affects analytics payload values, and the fixed call sites are covered by component tests asserting the exact `reportAppInteraction` arguments

### DIFF
--- a/src/components/FeedbackButton/FeedbackButton.tsx
+++ b/src/components/FeedbackButton/FeedbackButton.tsx
@@ -1,5 +1,10 @@
 import React from 'react';
-import { reportAppInteraction, UserInteraction, calculateJourneyProgress } from '../../lib/analytics';
+import {
+  reportAppInteraction,
+  UserInteraction,
+  calculateJourneyProgress,
+  AnalyticsContentType,
+} from '../../lib/analytics';
 import { getFeedbackButtonStyles } from '../../styles/feedback-button.styles';
 import { useTheme2 } from '@grafana/ui';
 import { t } from '@grafana/i18n';
@@ -9,7 +14,7 @@ interface FeedbackButtonProps {
   className?: string;
   variant?: 'primary' | 'secondary';
   contentUrl?: string;
-  contentType?: string;
+  contentType?: AnalyticsContentType;
   interactionLocation?: string; // Specific location identifier for analytics
   currentMilestone?: number; // For learning journeys - current milestone user is viewing
   totalMilestones?: number; // For learning journeys - total milestones in journey
@@ -19,7 +24,7 @@ export const FeedbackButton: React.FC<FeedbackButtonProps> = ({
   className,
   variant = 'primary',
   contentUrl = '',
-  contentType = '',
+  contentType,
   interactionLocation = 'feedback_button', // Default fallback
   currentMilestone,
   totalMilestones,
--- a/src/components/LearningPaths/MyLearningTab.tsx
+++ b/src/components/LearningPaths/MyLearningTab.tsx
@@ -15,7 +15,7 @@ import { LearningPathCard } from './LearningPathCard';
 import { BadgeIcon } from './BadgeIcon';
 import { SkeletonLoader } from '../SkeletonLoader';
 import { FeedbackButton } from '../FeedbackButton/FeedbackButton';
-import { reportAppInteraction, UserInteraction } from '../../lib/analytics';
+import { reportAppInteraction, UserInteraction, AnalyticsContentType } from '../../lib/analytics';
 import { StorageEvents } from '../../lib/event-names';
 import {
   learningProgressStorage,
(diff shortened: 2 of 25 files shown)


Jayclifford345:
@cursoragent review

moxious (dismissed):
Non-blocking findings, prioritized:

1. ⚠️ Incomplete standardization — worth fixing in-PR (one line). Scroll tracking at src/lib/analytics.ts:420 (buildScrollEventProperties) still emits raw activeTab?.type, so PanelScroll events on an interactive tab keep emitting the deprecated content_type: 'interactive' — the exact value the PR eliminates everywhere else. The PR edited this line (swapped the fallback to the enum) but left activeTab?.type raw. Fix: isRecommendationsTab ? '' : tabTypeToContentType(activeTab?.type).
2. Scope creep + untested behavior change. AssistantBlockWrapper.tsx 'code' → 'code-block' (4 sites). The 'code' value was genuinely dead so it's directionally correct, but blockType is typed string — the compiler did not enforce it (unlike the content_type work), and it's a live behavior change (code blocks now hit extractQueryFromResponse + interactive render path) bundled into an analytics-values PR with no test. Recommend a focused test and ideally splitting it out.
3. Fallback semantics shifted at 3 sites (edge case). tabTypeToContentType(undefined) returns docs; previously two sites defaulted to learning-journey and one to interactive. Likely unreachable in practice (tab type is set), but worth an author gut-check.
4. source_page carries a raw URL to Faro (DocsPanelContentArea.tsx:343). Intended fix, matches siblings — but source_page doesn't match the /url$/i normalization key pattern, so the URL reaches Faro un-redacted. Pre-existing gap shared by other sites, low-sensitivity internal URLs → non-blocking, but a good follow-up.

One-way door (disclosed): interactive_guide → interactive-guide is irreversible downstream — historical Faro/RudderStack data keeps the underscore. PR calls this out; Faro is Grafana-internal so affected dashboards are owned internally. Author should confirm saved dashboard/alert queries filtering on interactive_guide are updated (and note finding #1 means scroll events still emit a third value, interactive, until fixed).

moxious (approved):

PR #1372, opened 2026-07-16
feat(e2e): add a new runner image, schema-backed report contract, and capacity retries

leslie-heinzen:
## Summary

Packages E2E execution in a dedicated non-root Playwright runner image for orchestration systems, with Chromium, the CLI, and its matching versioned report schema in one immutable artifact.

The Zod-backed report contract is exported as JSON Schema and validated before every write. If validation fails, the E2E command preserves the original report for diagnosis and exits with configuration error 2. Catchable CLI failures produce structured zero-step reports, and bounded retries handle temporary pool-capacity pressure.

## What to look at

- **[Report schema and writer](https://github.com/grafana/grafana-pathfinder-app/blob/c72f63160c2946f04a9fa679249f87678edebaa2/src/cli/e2e/schemas/e2e-report.schema.ts)** — `E2EExecutionOutcome`, `E2EErrorCode`, and `RunnerProvenance` form the cross-repository contract. Successful validation strips unknown fields; failed validation writes the original diagnostic object and returns an explicit failure status.
- **[Report finalization and exit-code selection](https://github.com/grafana/grafana-pathfinder-app/blob/c72f63160c2946f04a9fa679249f87678edebaa2/src/cli/e2e/e2e-console-reporter.ts)** — schema validity is propagated from single- and multi-guide report writers into the [E2E command](https://github.com/grafana/grafana-pathfinder-app/blob/c72f63160c2946f04a9fa679249f87678edebaa2/src/cli/commands/e2e.ts). An invalid report takes precedence over test and authentication outcomes and exits with code 2 without discarding the artifact.
- **[Schema command](https://github.com/grafana/grafana-pathfinder-app/blob/c72f63160c2946f04a9fa679249f87678edebaa2/src/cli/commands/schema.ts)** — `e2e-report` and `e2e-multi-report` are registered alongside the guide and package schemas, emit stable `$id` values, and include `x-schema-version` when requested. Reference reuse prevents recursive guide schemas from expanding into an unnecessarily large inline schema.
- **[CLI error boundary](https://github.com/grafana/grafana-pathfinder-app/blob/c72f63160c2946f04a9fa679249f87678edebaa2/src/cli/commands/e2e.ts)** — catchable setup and execution failures map to structured outcomes and produce conforming zero-step reports. Tier skips remain distinct from configuration and infrastructure failures.
- **[Pool-manager client](https://github.com/grafana/grafana-pathfinder-app/blob/c72f63160c2946f04a9fa679249f87678edebaa2/src/cli/e2e/cloud-stack-pool-manager.ts)** — `leaseForChain` retries `no_capacity` responses with bounded exponential backoff. A single injected clock tracks elapsed time so the maximum wait budget is not double-counted.
- **[Runner image](https://github.com/grafana/grafana-pathfinder-app/blob/c72f63160c2946f04a9fa679249f87678edebaa2/Dockerfile.e2e-runner) and [publication workflow](https://github.com/grafana/grafana-pathfinder-app/blob/c72f63160c2946f04a9fa679249f87678edebaa2/.github/workflows/e2e-runner-publish.yml)** — the image includes Playwright and Chromium, runs as the non-root `node` user, and exposes the same schema as the packaged CLI. The workflow verifies the container contract before publishing signed immutable `commit-<SHA>` tags.
- **[E2E testing documentation](https://github.com/grafana/grafana-pathfinder-app/blob/c72f63160c2946f04a9fa679249f87678edebaa2/docs/developer/E2E_TESTING.md)** — documents report fields, schema extraction, failure semantics, runner image usage, and the boundary between runner outcomes and downstream health telemetry.

## Why

The previous report had no schema version or language-agnostic contract, forcing consumers to parse human-readable output or depend on internal TypeScript types. Several pre-execution paths also exited without a report, preventing consumers from reliably distinguishing skips, configuration errors, capacity pressure, and infrastructure failures.

A producer must not return success for output it knows violates its versioned contract. Exiting with configuration error 2 makes the contract enforceable while preserving the malformed report for diagnosis.

The existing slim CLI image also lacks Playwright and Chromium. Continuous E2E orchestration needs a purpose-built image containing the browser runtime, CLI, and exact schema version used to produce its reports.

## How to verify

1. Export the report schema with `npx pathfinder-cli@... schema e2e-report --include-version`; confirm it includes an `$id` containing `e2e-test-report-1.0.0` and `x-schema-version: "1.0.0"`.
2. Build the runner image with `docker build --file Dockerfile.e2e-runner --tag pathfinder-e2e-runner:local --build-arg RUNNER_VERSION=local .`.
3. Run `docker run --rm --entrypoint id pathfinder-e2e-runner:local -u`; confirm it prints a non-zero UID.
4. Export the schema from the image with `docker run --rm --entrypoint node pathfinder-e2e-runner:local /app/dist/cli/cli/index.js schema e2e-report --include-version`; confirm it matches the checkout schema version and `$id`.
5. Run a cloud-tier guide with `--tier local`, then inspect `<artifacts>/report.json`; confirm the report distinguishes the skip from configuration and infrastructure failures.
6. Review `.github/workflows/e2e-runner-publish.yml`; confirm package and identity-token write permissions are scoped to the publish job while the build job remains read-only.

## Breaking changes

Existing CLI consumers are unaffected as long as generated reports conform to the new schema. If the runner emits an invalid report, the command now preserves the malformed artifact for diagnosis but exits with configuration error 2, taking precedence over test and authentication outcomes. This fail-closed behavior prevents automation from accepting output that claims a versioned contract but does not satisfy it.

The new report fields are additive for consumers that do not yet validate against the published schema.

## Reversibility

The validation behavior can be reverted without modifying stored data, but the report schema is a published cross-repository contract. Reverting before downstream consumers pin to it is safe. Once an orchestrator consumes the versioned outcome and error-code fields, renaming or removing them requires a coordinated schema-version change. Published runner images are immutable, so operational rollback means pinning a prior image digest rather than replacing an existing tag.


src/cli/e2e/schemas/e2e-report.schema.ts
```diff
@@ -0,0 +1,231 @@
...
+    summary: ReportSummarySchema,
+    steps: z.array(ReportStepResultSchema),
+    aborted: z.boolean().optional(),
+    abortReason: AbortReasonSchema.optional(),
+    abortMessage: z.string().optional(),
+    preRunSkipped: z.array(PreRunSkipSchema).optional(),
+    cleanupWarnings: z.array(z.string()).optional(),
+  })
+  .meta({
+    id: E2E_REPORT_SCHEMA_ID,
```
moxious: **High** — using the full URL as `.meta({ id })` makes zod key this schema's `$defs` entry by that URL when it is nested inside the multi-guide schema. The `e2e-multi-report` export then contains `"$ref": "#/$defs/https://grafana.com/schemas/pathfinder/e2e-test-report-1.0.0.json"` — an invalid JSON Pointer per RFC 6901 (unescaped `/` must be `~1`), and the def carries no internal `$id` for URI-based resolution. ajv fails to compile the export (reproduced against this repo's zod 4.4.3), so downstream consumers cannot use the multi-guide contract at all. The single-report export is fine.

Fix direction: keep the URL out of `.meta({ id })` (the manual root-`$id` injection in `schema.ts` already handles the root correctly) or use a short non-URL id so nested refs stay `#/$defs/<safe-name>`.

src/cli/e2e/e2e-console-reporter.ts
```diff
@@ -37,9 +37,21 @@ function skipOnlyReport(
...
+  const timestamp = new Date().toISOString();
+  const runner: RunnerProvenance = {
+    name: 'pathfinder-e2e-runner',
+    version: process.env.PATHFINDER_E2E_RUNNER_VERSION ?? 'source',
+    nodeVersion: process.version,
+    playwrightVersion: process.env.PLAYWRIGHT_VERSION ?? 'unknown',
+  };
   return {
+    schemaVersion: E2E_REPORT_SCHEMA_VERSION,
+    outcome: 'skipped',
```
moxious: **Medium** — `validation_failed`, `fetch_failed`, and `resolution_failed` are in both `REMOTE_SKIP_REASONS` (so they land in `preRunSkipped` with no results data) and `FAILURE_STATUSES` (non-zero exit). A run where all packages fail validation takes the skip-only branch and reports `outcome: "skipped"` while the process exits 1 — and when other guides passed, `generateMultiGuideReport` derives `outcome` only from executed guides, so the report can even say `passed`. Consumers gating on the new `outcome` field misread a failing run as clean — the exact drift this contract exists to prevent.

src/cli/e2e/playwright-runner.ts
```diff
@@ -5,13 +5,37 @@
...
 import { E2E_ENV, encodeEnvFlag } from './e2e-runner-contract';
 import type { LoadedGuide } from '../utils/file-loader';
-import type { TestResultsData } from './e2e-reporter';
+import type { E2EErrorCode } from './schemas/e2e-report.schema';
+import { contentDigest, createMinimalResultsData, type TestResultsData } from './e2e-reporter';
+
+const PLAYWRIGHT_CONFIG_PATH = join('tests', 'e2e-runner', 'playwright.config.ts');
+const GUIDE_RUNNER_SPEC_PATH = join('tests', 'e2e-runner', 'guide-runner.spec.ts');
+
+export function findRunnerRoot(startDir: string): string {
```
moxious: **Medium** — `findRunnerRoot(__dirname)` has no cwd fallback, and the published package ships only `dist/` (`files: ["dist/"]` in package.json), so a registry-installed CLI can never find `tests/e2e-runner`: the walk-up from the install location throws before anything runs, where the old cwd-relative paths worked from a checkout. The Docker image (`COPY . .` into `/app`) and in-repo `npx` both resolve correctly, and no npm publish exists today — but the published-CLI deployment mode is structurally unsatisfiable as shipped. Consider falling back to `process.cwd()` when the walk-up fails. If registry publishing is planned soon, treat this as high severity.

.github/workflows/e2e-runner-publish.yml
```diff
@@ -0,0 +1,155 @@
...
+            echo "::error::Tag already exists: ${{ steps.tag.outputs.value }}"
+            exit 1
+          fi
+      - name: Build and push immutable image
+        id: push
+        uses: docker/build-push-action@53b7df96c91f9c12dcc8a07bcb9ccacbed38856a # v7
+        with:
+          context: .
+          file: Dockerfile.e2e-runner
+          push: true
```
moxious: **Medium** — the publish job rebuilds the image from scratch on a fresh runner (no `cache-from`/`cache-to`, no artifact handoff from the build job) and signs that new digest. `npx playwright install --with-deps` pulls apt packages at build time, so the rebuild is not even functionally pinned — the signed `commit-<SHA>` tag is not guaranteed to be the artifact that passed the schema and Playwright contract checks.

Fix options: build + `load` + test + push in a single job (same builder, same layers); `docker save`/upload-artifact the tested image into publish; or push a candidate digest first and only tag/sign after in-registry verification.

src/cli/e2e/e2e-console-reporter.ts
```diff
@@ -223,14 +235,17 @@ export function writeJsonReport(
   if (resultsWithData.length === 0 && preRunSkipped.length > 0) {
     try {
       const report = skipOnlyReport(preRunSkipped, cleanupWarnings);
-      writeMultiGuideReport(report, outputPath);
+      const schemaValid = writeMultiGuideReport(report, outputPath);
       console.log(`\n📄 Multi-guide JSON report written to: ${outputPath}`);
       console.log(`   ${formatMultiGuideSummary(report)}`);
+      return schemaValid;
     } catch (err) {
       console.warn(`   ⚠ Failed to write JSON report: ${err instanceof Error ? err.message : 'Unknown error'}`);
```
moxious, line 244: **Medium** — every write-failure `catch` in `writeJsonReport` (here and the two branches below) returns `true`, meaning "schema valid". An unwritable artifacts dir therefore produces no report on disk while the exit path proceeds as if a conforming report was written — a run can exit 0 with no report, and a schema-validation failure detected before the throw is masked. That undercuts the fail-closed exit-2 guarantee exactly in the corner (broken artifacts path) where diagnosis matters. Consider returning `false` or a distinct status from these branches.

Dockerfile.e2e-runner
```diff
@@ -0,0 +1,27 @@
...
+
+COPY package.json package-lock.json .npmrc ./
+RUN npm ci --no-audit --no-fund \
+    && npx playwright install --with-deps chromium \
+    && npm cache clean --force
+
+COPY . .
+RUN npm run build:cli \
+    && chmod +x scripts/e2e-runner-entrypoint.sh \
+    && chown -R node:node /app /ms-playwright
```
moxious: **Medium** — `chown -R node:node /app /ms-playwright` in a RUN layer after `npm ci` and `COPY . .` copies every file it touches into a new layer, roughly doubling the image (full node_modules including dev deps, Chromium, and browser OS deps duplicated). That is a multi-GB pull penalty on every orchestrated cold start. Use `COPY --chown=node:node . .` and install browsers for the target user instead, or chown only the specific paths that need it.

src/cli/e2e/playwright-runner.ts
```diff
@@ -131,6 +168,7 @@ function processPlaywrightResults(
     exitCode: success ? ExitCode.SUCCESS : ExitCode.TEST_FAILURE,
     traceFile,
     resultsData,
+    ...(!resultsData ? { errorCode: 'REPORT_MISSING' as const } : {}),
```
moxious: **Medium** — when Playwright exits 0 but writes no results file, this path returns `success: true` / `ExitCode.SUCCESS` alongside `errorCode: "REPORT_MISSING"`, and the synthesized results data records `outcome: "infrastructure_error"`. The caller then marks the guide `passed` and the process exits 0 — a silently missing results file counts as a pass while the versioned report contradicts itself (per-guide `infrastructure_error` inside a passing run).

src/cli/commands/e2e.ts
```diff
@@ -477,7 +555,11 @@ async function runPreflightChecks(
         const tierMsg = packageManifest.testEnvironment?.tier ?? 'unknown';
         console.log(`\n⊘ Guide skipped: requires tier "${tierMsg}" but current environment is "${options.tier}".`);
         console.log(`   Use --tier ${tierMsg} to run this guide against a matching environment.`);
-        process.exit(ExitCode.SUCCESS);
+        failCommand(
+          `Guide skipped: requires tier "${tierMsg}" but current environment is "${options.tier}".`,
+          ExitCode.SUCCESS,
+          'SKIPPED_PREREQ'
+        );
```
moxious, line 563: **Low** — tier mismatch exits through `failCommand(..., "SKIPPED_PREREQ")` (here and at the multi-guide site below), the same code — and back-filled `abortReason` — used when a prerequisite guide fails. Outcome and exit code stay correct (`skipped`, exit 0), but consumers lose the skip reason: "environment tier mismatch" and "prerequisite guide did not pass" become indistinguishable. A dedicated `TIER_MISMATCH` code is a one-way addition best made before the v1.0.0 enum freezes.

src/cli/e2e/e2e-reporter.ts
```diff
@@ -587,12 +447,39 @@ export function generateMultiGuideReport(resultsArray: TestResultsData[], grafan
...
+    reports.map(({ outcome }) => (outcome === 'aborted' ? 'failed' : outcome))
+  );
+  const outcomePriority: E2EExecutionOutcome[] = [
+    'infrastructure_error',
+    'configuration_error',
+    'failed',
+    'skipped',
+    'passed',
+  ];
+  const outcome = outcomePriority.find((candidate) => reportOutcomes.has(candidate)) ?? 'passed';
```
moxious, line 473: **Low** — this priority ranks `skipped` above `passed`, so a run with N passed guides plus one tier-skipped guide yields top-level `outcome: "skipped"` on an exit-0 run, while the single-guide path reports `passed` with `preRunSkipped` entries for the same situation. Consumers checking `outcome === "passed"` treat a fully successful run as not-passed, and the two report shapes disagree.

src/cli/e2e/schemas/e2e-report.schema.ts
```diff
@@ -0,0 +1,231 @@
...
+
+export const E2ETestReportSchema = z
+  .object({
+    schemaVersion: z.literal(E2E_REPORT_SCHEMA_VERSION),
+    outcome: E2EExecutionOutcomeSchema,
+    errorCode: E2EErrorCodeSchema.optional(),
+    errorMessage: z.string().optional(),
+    runner: RunnerProvenanceSchema,
+    startedAt: z.string(),
+    endedAt: z.string(),
```
moxious: **Low** — contract-tightness decisions worth making before 1.0.0 freezes: timestamps here (and `config.timestamp`) are bare `z.string()` with no `format: date-time` even though producers always emit `toISOString()`; summary counts and step `index`/`duration` are `z.number()`, so negative and fractional values validate; and every object exports `additionalProperties: false` alongside `schemaVersion: z.literal("1.0.0")`, making any additive optional field a breaking version bump. Deliberate strictness may be intended, but the evolution policy deserves an explicit decision now — tightening any of this later is a breaking change.

src/cli/__tests__/schema-e2e-report.test.ts
```diff
@@ -0,0 +1,27 @@
...
+    const schema = exportSchema('e2e-report', true);
+
+    expect(schema).not.toBeNull();
+    expect(String(schema?.$id)).toContain('e2e-test-report-1.0.0');
+    expect(schema?.['x-schema-version']).toBe('1.0.0');
+  });
+
+  it('exports the multi-guide report schema without throwing', () => {
+    expect(() => exportSchema('e2e-multi-report', false)).not.toThrow();
+    expect(exportSchema('e2e-multi-report', false)).not.toBeNull();
```
moxious, line 26: **Low** — these assertions only check no-throw/non-null; nothing verifies the export is a compilable, resolvable JSON Schema, which is exactly how the unresolvable `$ref` in the `e2e-multi-report` export shipped green. An ajv compile assertion per registered e2e schema would catch that whole class of regression (ref shape, `$defs` keying, zod version bumps).

Dockerfile.e2e-runner
```diff
@@ -0,0 +1,27 @@
+# Dedicated Pathfinder E2E runner image.
+# This image is separate from Dockerfile.cli because it includes Playwright,
+# Chromium, and the browser runner.
+
+FROM node:24-bookworm@sha256:5711a0d445a1af54af9589066c646df387d1831a608226f4cd694fc59e745059
+
+ARG RUNNER_VERSION=source
+ENV PATHFINDER_E2E_RUNNER_VERSION=${RUNNER_VERSION}
+ENV PLAYWRIGHT_BROWSERS_PATH=/ms-playwright
```
moxious, line 9: **Low** — the reporters read `process.env.PLAYWRIGHT_VERSION ?? "unknown"` for `runner.playwrightVersion`, but nothing in the image, entrypoint, or workflow sets it, so every image-produced report carries `playwrightVersion: "unknown"` while the docs show a concrete version. Rather than adding an ENV that drifts from the lockfile, consider resolving `require("playwright/package.json").version` at runtime in the reporter.

Dockerfile.e2e-runner.dockerignore
```diff
@@ -0,0 +1,12 @@
+.git
```
moxious, line 1: **Low** — `.env*` and `playwright/.auth/` (both gitignored precisely because they hold secrets) are not excluded here, so a local `docker build` bakes live Grafana session state and env secrets into distributable layers via `COPY . .`. CI builds from a clean checkout are unaffected; adding those patterns closes the local-build leak.

.github/workflows/e2e-runner-publish.yml
```diff
@@ -0,0 +1,155 @@
...
+      - 'tests/e2e-runner/**'
+      - 'tests/fixtures.ts'
+      - 'src/cli/e2e/schemas/**'
+      - 'scripts/e2e-runner-entrypoint.sh'
+      - 'Dockerfile.e2e-runner'
+      - 'Dockerfile.e2e-runner.dockerignore'
+      - 'package.json'
+      - 'package-lock.json'
+      - '.npmrc'
+      - '.github/workflows/e2e-runner-publish.yml'
```
moxious, line 21: **Low** — the `paths` filters omit `tsconfig.cli.json`, which `build:cli` uses as a direct build input. A compiler-options change passes `npm run check` CI but never triggers a runner-image rebuild/republish, so the published image silently drifts from source until an unrelated path changes.

.github/workflows/e2e-runner-publish.yml
```diff
@@ -0,0 +1,155 @@
...
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - uses: sigstore/cosign-installer@6f9f17788090df1f26f669e9d70d6ae9567deba6 # v4.1.2
+      - name: Compute immutable tag
+        id: tag
+        run: echo "value=commit-${GITHUB_SHA}" >> "$GITHUB_OUTPUT"
+      - name: Verify tag does not exist
+        run: |
+          if docker manifest inspect "${{ env.IMAGE_NAME }}:${{ steps.tag.outputs.value }}" >/dev/null 2>&1; then
```
moxious: **Low** — this gate runs before build+push+sign as a unit: if a run fails between `push` and `cosign sign`, the re-run hard-fails here, leaving a published-but-unsigned `commit-<SHA>` tag with no recovery path short of manual package deletion. Also worth noting GHCR tags are not actually immutable, so this check is best-effort self-discipline — the docs correctly tell consumers to pin digests, which is the real invariant.

src/cli/e2e/cloud-stack-pool-manager.ts
```diff
@@ -20,6 +22,15 @@ export interface CloudStackPoolManagerConfigInput {
   env?: NodeJS.ProcessEnv;
 }
 
+function retryDelayMs(attempt: number, random: () => number): number {
+  const exponential = Math.min(CAPACITY_RETRY_MAX_DELAY_MS, CAPACITY_RETRY_BASE_DELAY_MS * 2 ** attempt);
+  return Math.round(exponential * (0.5 + random()));
```
moxious: **Low** — jitter (`0.5 + random()`, up to 1.5×) is applied after the clamp to `CAPACITY_RETRY_MAX_DELAY_MS`, so the effective per-retry maximum is ~7,500 ms, not the named 5,000 ms cap. Harmless in practice because the remaining-budget clamp still applies, but the constant lies — clamp after applying jitter.

(While verifying this area: the budget accounting itself is correct — single injected clock, no double-counting, sleeps clamped to the remaining budget, and retries cannot double-lease.)

src/cli/commands/e2e.ts
```diff
@@ -951,6 +1048,6 @@ export const e2eCommand = new Command('e2e')
       reportResults([...inputs.preRunSkipped, ...outcome.results], options, outcome.cleanupWarnings);
     } catch (error) {
       console.error('❌ Error:', error instanceof Error ? error.message : 'Unknown error');
```
moxious: **Low** — converting the tier-skip `process.exit(SUCCESS)` into a thrown `E2ECommandError` routes it through this generic catch, which prints `❌ Error: Guide skipped: ...` on stderr immediately after the intentional `⊘ Guide skipped` message, on a run that exits 0. A successful skip now looks like an error in CI logs. Also note `writeCommandFailureReport` calls `writeReport` with no try/catch inside this handler — if the artifacts dir is unwritable, the exception escapes the catch as an unhandled rejection (exit 1, no report) on the one path meant to guarantee a conforming failure report.

Dockerfile.e2e-runner
```diff
@@ -0,0 +1,27 @@
...
+
+COPY . .
+RUN npm run build:cli \
+    && chmod +x scripts/e2e-runner-entrypoint.sh \
+    && chown -R node:node /app /ms-playwright
+
+USER node
+WORKDIR /workspace
+
+ENTRYPOINT ["/app/scripts/e2e-runner-entrypoint.sh"]
```
moxious, line 29: **Low** — Chromium-in-Docker generally needs `--shm-size`/`--ipc=host` per Playwright's own container guidance, and nothing here or in the docs mentions it; long-lived runs can also accumulate browser zombies since there is no init reaper (the entrypoint's `exec` correctly makes node PID 1 with signal handlers, which mitigates but does not reap). Mostly latent for one-shot jobs, but a known flake source — worth documenting the recommended `docker run` flags or adding `--init`/tini guidance.

moxious (changes requested):
Requesting changes: 1 high and 6 medium findings, plus a set of low-severity items worth deciding before the v1.0.0 schema freezes. Findings were cross-verified; the high-severity one was reproduced empirically with ajv against the repo's zod 4.4.3.

**Must fix**
- `src/cli/e2e/schemas/e2e-report.schema.ts:160` — the exported `e2e-multi-report` JSON Schema is unresolvable by standard validators (invalid `$ref` JSON Pointer), so the multi-guide half of the cross-repo contract cannot be consumed.

**Should fix**
- `e2e-console-reporter.ts:49` — validation-failed runs exit 1 but report `outcome: "skipped"` (or `"passed"`).
- `playwright-runner.ts:20` — the registry-published CLI mode structurally cannot locate the runner root.
- `e2e-runner-publish.yml:148` — the signed `commit-<SHA>` image is a fresh rebuild, not the artifact that passed the contract checks.
- `e2e-console-reporter.ts:243` — report write failures count as "schema valid"; a run can exit 0 with no report.
- `Dockerfile.e2e-runner:21` — `chown -R` in a separate layer roughly doubles the image.
- `playwright-runner.ts:171` — Playwright exit 0 with a missing results file counts as a pass.

**Before the contract freezes (low severity)** — tier-skip vs prerequisite-skip code granularity, `skipped` outranking `passed` in the multi-guide outcome, loose timestamp/count types plus the closed-world `additionalProperties: false` + literal-version evolution policy, and a schema compile assertion in tests (which would have caught the high finding).

On the existing Bugbot comments: "provisioning misclassified" and "abort fallback forces infra outcome" are refuted at this head (explicit mappings exist and tests cover both scenarios); the runner-root and tier-skip findings are partially right (see inline); the workflow rebuild and `PLAYWRIGHT_VERSION` findings are confirmed.

Verified sound along the way: pool-manager backoff budget accounting (single injected clock, no double-counting, no double-lease risk), all action pins resolve to genuine upstream commits, per-job permissions scoping matches the PR claims, entrypoint signal handling, and producer output round-trips `safeParse` on every path checked.

moxious:
cli-and-e2e-runner — coherent_extension. The machine-readable output contract originated in PR #532 (hand-written TS interfaces + exit codes 0–4) and was reshaped by #1152, #1230, #1346 — so this PR is third_or_later on the surface. But it converges rather than branches: the scan verified at head that e2e-reporter.ts deletes its hand-written report interfaces and derives every type from the new Zod schema via z.infer, and the schema export reuses the pre-existing SCHEMA_REGISTRY rather than inventing a parallel path. After the PR there are fewer competing representations than before, and no branching condition applies — so the "third-or-later + branching condition → blocking" row never fires. It also checked that E2E_TESTING_CONTRACT.md isn't a competing document (it covers DOM data-test-* attributes, a different surface).

build-and-ci — coherent_extension. The new e2e-runner-publish.yml/Dockerfile.e2e-runner pair replicates the established cli-publish.yml pattern byte-for-byte where it matters (identical action pins, same permissions posture, cosign digest signing) and strengthens the tag contract (immutable commit-<sha>, pre-push verification) with documented rationale. The gate reported partial history; the agent legitimately upgraded it to complete by inspecting the single unclassified commit (#1102, a Playwright bump touching only package files) and recording why it's irrelevant — the exact upgrade procedure the skill mandates.

src/cli/e2e/schemas/e2e-report.schema.ts
leslie-heinzen: I've thought about the breaking version bump caused by purely additive changes for a bit now and I think in our case I prefer an open-world policy. I think this will ease up on coordination between services for us.

I've pushed a change in that vein. The fix is at the export layer only (`schema.ts`). The Zod schemas themselves are unchanged, so the runner's own internal validation remains strict. Only the JSON Schema surface exposed to cross-service consumers becomes open-world. A regression test locks this in.

moxious (changes requested):
Requesting changes on current head `5563a02f`. The earlier review fixes materially improved this PR, but two production defects still make the new runner contract unsafe to publish:

1. **[blocking] The “immutable” commit tag can be overwritten.** In `.github/workflows/e2e-runner-publish.yml:148-157`, an existing `commit-${GITHUB_SHA}` tag only produces a warning; the next step unconditionally pushes to the same tag and signs the replacement digest. A workflow rerun can therefore move the advertised immutable reference, and reverting the workflow cannot restore the old registry mapping. Please make reruns verify/sign the existing digest without repushing, or fail on a conflicting tag.

2. **[blocking] Relative artifact paths are resolved from two different working directories.** `runPlaywrightTests` changes the child cwd to `runnerRoot` (`/app` in the image) but forwards relative `--artifacts` unchanged. Playwright writes screenshots/DOM under `/app`, while the parent writes `report.json` relative to `/workspace`; `docker run --rm` then discards the browser artifacts. Resolve the artifacts path against the caller cwd before spawning and pin this with a test.

Two smaller contract issues should be corrected in the same pass:

- A `PLAYWRIGHT_SPAWN_FAILED` result is created with setup/configuration exit code 2, but final result aggregation maps its generic `failed` status to exit code 1. Preserve setup-error precedence so the report and process exit agree.
- HEAD fixes `aborted` outcome precedence when failed `preRunSkipped` entries are attached, but no test exercises that exact mixed state. Add a `writeJsonReport` regression test with an aborted runnable result plus a failed pre-run skip.

These findings survived three independent refutation passes. The rest of the evidence is strong: 32 checks pass, 102 focused tests pass, typecheck and diff checks pass, the PR merges cleanly with current `main`, and the production dependency audit is clean.

moxious (approved):

PR #1377, opened 2026-07-17
fix(e2e): isolate selector resolution from browser runtime

leslie-heinzen:
## Summary

A recent change made Grafana selector resolution version-aware by importing `@grafana/runtime` from the shared selector module. That change was breaking for local E2E runs: Playwright loads guide runner modules in Node before launching Chromium, so `@grafana/runtime` accessed `window` during test discovery and aborted with `ReferenceError: window is not defined`.

This PR separates selector logic into environment-neutral cores and thin browser/Node adapters, restoring local CLI E2E runs while preserving browser version-awareness and the runner's existing latest-selector behavior.

## What to look at

- **Selector boundary**: [`grafana-selector-core.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/e2706f70d36846807ece1cfd7fa6c499c4cbc14d/src/lib/dom/grafana-selector-core.ts) and [`selector-resolver-core.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/e2706f70d36846807ece1cfd7fa6c499c4cbc14d/src/lib/dom/selector-resolver-core.ts) own environment-neutral catalog and reference resolution. The browser-facing adapters continue to supply `config.buildInfo.version` from Grafana runtime.
- **E2E runner boundary**: [`selector-resolver.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/e2706f70d36846807ece1cfd7fa6c499c4cbc14d/tests/e2e-runner/utils/selector-resolver.ts) owns the runner's existing `latest` selector policy, and [`execution.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/e2706f70d36846807ece1cfd7fa6c499c4cbc14d/tests/e2e-runner/utils/guide-runner/execution.ts) imports that narrow Node-safe adapter instead of the browser-oriented DOM barrel.

## How to verify

1. Start Grafana with the Pathfinder plugin available at `http://localhost:3000`.
2. Run `npx playwright test tests/e2e-runner/guide-runner.spec.ts --config=tests/e2e-runner/playwright.config.ts --project=chromium --list`. Confirm Playwright lists the auth and guide tests without a `window is not defined` error.
3. Run `pathfinder-cli e2e bundled:welcome-to-grafana`. Confirm the command proceeds through Playwright discovery and executes the guide.

## Breaking changes

None. Browser selector behavior remains version-aware, and the E2E runner retains its existing latest-selector policy. The change is fully reversible.

## Out of scope

[...]

PR #1378, opened 2026-07-18
test(architecture): environment-reachability check for Node contexts

moxious:
## Summary

Adds an environment-reachability invariant to the architecture tests: everything transitively reachable from Node-executed code (`src/cli/`, `tests/`, `playwright.config.ts`) must load in plain Node. This closes the blindspot behind the e2e breakage fixed in #1377 — tier enforcement models which of our directories may import which, but it cannot see that an external package like `@grafana/runtime` requires a browser at module-evaluation time.

## What to look at

- **Scan mechanism** ([`src/validation/import-graph.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/e20153a0829e18350f28e1de5cb0efc066c164c9/src/validation/import-graph.ts)): the AST walker now emits full import records (specifier + type-only + dynamic + relative/external) instead of relative specifiers only; `extractRelativeImports` is a thin wrapper, so tier/cycle enforcement is unchanged. `scanNodeEnvReachability` BFS-walks the value-import closure from `NODE_CONTEXT_ROOTS` and returns violations with witness chains. Two deliberate semantics to trace: type-only imports are exempt (erased at compile time — the adapter pattern legitimately does `import type` from `@grafana/runtime`), while dynamic `import()` edges ARE followed (the CLI lazy-loads its commands, so lazy modules still execute in Node eventually).
- **Policy + failure message** ([`src/validation/architecture.test.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/e20153a0829e18350f28e1de5cb0efc066c164c9/src/validation/architecture.test.ts)): `NODE_SAFE_EXTERNALS` is an allowlist with the opposite polarity from the ratchets — growing it with a proven Node-safe dependency is expected maintenance, not degradation, and both the constant's comment and the failure message say so explicitly. The failure message is written for the agent/developer who trips it while not looking at the CLI: witness chain, the `*-core.ts` adapter fix pattern, the `import type` escape, and the exact procedure to prove a new dependency Node-safe before allowlisting it. A stale-entry test keeps the list honest.
- **Root selection**: jest-managed test files are excluded from the roots — they run under `jest-environment-jsdom` with fake browser globals, so they are not evidence of plain-Node execution. The effective jest testMatch is the root `jest.config.js` (src/ patterns plus `tests/e2e-runner/utils/**/*.test.*`); Playwright loads only `*.spec.ts` files, which are included along with helpers, fixtures, and `playwright.config.ts` itself. Covering all of `tests/` (not just the e2e-runner) closes the same class of blindspot for the main Playwright suite, whose specs also value-import `src/` modules during Node-side discovery.
- **Docs** ([`AGENTS.md`](https://github.com/grafana/grafana-pathfinder-app/blob/e20153a0829e18350f28e1de5cb0efc066c164c9/AGENTS.md), [`.cursor/rules/systemPatterns.mdc`](https://github.com/grafana/grafana-pathfinder-app/blob/e20153a0829e18350f28e1de5cb0efc066c164c9/.cursor/rules/systemPatterns.mdc), [`docs/design/CONCERNS.md`](https://github.com/grafana/grafana-pathfinder-app/blob/e20153a0829e18350f28e1de5cb0efc066c164c9/docs/design/CONCERNS.md)): environment reachability documented as an axis orthogonal to tiers; `cli-and-e2e-runner` concern gains trigger keywords and a review question. The AGENTS.md/systemPatterns paragraphs sit inside the section parsed by the tier-doc-sync test and were written to stay outside its bullet/excluded-line patterns.

## Why

The #1377 incident: a shared selector module gained an `@grafana/runtime` import, Playwright loaded it in Node during test discovery, and `pathfinder-cli e2e` died with `ReferenceError: window is not defined` — with zero CI signal. The existing tier test could not have caught it because external packages are invisible to it and the Node contexts (`cli/` is tier-excluded; `tests/e2e-runner/` is outside `src/`) aren't in its graph at all.

Retro-validation: running this scan at the pre-fix commit (8ef83721) catches the exact breakage plus two more latent browser-reachable chains (`lib/logging.ts → lib/faro.ts → @grafana/runtime` and onward into `src/constants.ts`) that #1377 incidentally cured. At current `main` the closure is clean (137 files, 7 external packages, 0 violations), so this ships as a hard invariant with no grandfathered violations.

Allowlist polarity was chosen over a browser-package denylist deliberately: the original failure mode was "we couldn't statically tell it required a browser," and a denylist reproduces that blindspot for the next unknown package. A `lib/shared` directory boundary was also considered and rejected — the reachability closure itself defines what is shared, so a directory convention would add a second thing to keep in sync without adding enforcement power.

## How to verify

1. Trip the check: add `import '@grafana/runtime';` (a value import) to any file under `src/cli/` or to `src/lib/dom/selector-resolver-core.ts`, run `npx jest src/validation/architecture.test.ts`, and confirm the failure names the witness chain from a Node entrypoint and walks through the three resolution options (break the chain / `import type` / prove-then-allowlist). Revert.
2. Trip the stale check: add a fake package name to `NODE_SAFE_EXTERNALS` and confirm the stale-entry test fails, telling you to remove it. Revert.
3. Confirm the e2e surface still discovers: `npx playwright test tests/e2e-runner/guide-runner.spec.ts --config=tests/e2e-runner/playwright.config.ts --project=chromium --list` lists tests without a `window is not defined` error.

## Breaking changes

None. Test-and-docs-only change; no production code paths are touched, and `extractRelativeImports` output is unchanged for existing callers.

## Out of scope

- An ESLint `no-restricted-imports` mirror for editor-time feedback on direct browser imports from Node contexts — the architecture test catches the transitive case, which is the one that bit us; a lint mirror can follow if direct-import friction shows up.
- Static detection of browser-global usage (`window`/`document` at module scope) in our own first-party code — the external-package allowlist plus the `*-core.ts` convention covers the observed failure mode.

Refs #1377

Updated after an independent review pass; see the second commit for the seven findings addressed (root-selection premise, main-suite coverage, unresolvable-edge pruning, record-merge semantics, dead dedupe, chain-depth test, builtin check).


src/validation/import-graph.ts
```diff
@@ -556,3 +617,191 @@ export function assertRatchet(
...
+export const NODE_CONTEXT_ROOTS = ['src/cli', 'tests', 'playwright.config.ts'];
+
+/**
+ * Jest-managed test files (jsdom environment by default) are excluded from
+ * the Node-context roots: they get emulated browser globals and jest.mock
+ * protection, so they are not evidence of plain-Node execution. The effective
+ * testMatch is the root jest.config.js — src/ patterns from the scaffolded
+ * .config/jest.config.js plus tests/e2e-runner/utils/**\/*.test.*. Playwright
+ * loads only *.spec.ts files (see playwright.config.ts testMatch), which jest
+ * never matches outside src/.
```
leslie-heinzen, line 644: 👍🏻

leslie-heinzen (approved):
LGTM. Gave it a whirl by attempting to break things in a few ways. Thanks!

PR #1383, opened 2026-07-20
feat(guides): convert first-dashboard-cloud interactive steps to e2e selectors

Jayclifford345:
## What this does

Applies the same e2e-selector conversion from #1382 (first-dashboard, OSS) to the Grafana Cloud variant of the guide, `first-dashboard-cloud`. Verified live against `pathfinder.grafana-dev.net` (Grafana Cloud 13.2.0), and re-verified in the DOM against `learn.grafana.net` (`13.2.0-31683987170`).

- Converts the cloud-only `datasource-setup` section: nav to Add new connection, TestData search, TestData connection link (left as-is — plugin-id-based href is already stable), and the "Add new data source" button (previously plain text, no testid) → `grafana:pages.DataSources.dataSourceAddButton`.
- Carries over the same Explore/dashboard/panel-editor fixes as #1382 (scenario select, labels A/B, Add query, Add/Add to dashboard/Open dashboard, unit picker, save dashboard flow) — all verified to behave identically on Cloud's newer Grafana version.
- The cloud guide's original unit-input selector (`div[data-testid='input-wrapper'] #unit`) is replaced with the same `grafana:components.UnitPicker.container` used in the OSS guide.
- Raises `testEnvironment.minVersion` from `12.2.0` to `13.2.0` (and regenerates `repository.json`) so the manifest states the floor the converted selectors actually need — see below.

## Dependency: satisfied

The blocking review finding was that three of the new selector paths didn't exist in the pinned selectors package. That's resolved:

- grafana/grafana#128750 merged 2026-07-30 (`93044b13`, milestone 13.2.x).
- `main` pins `@grafana/e2e-selectors@13.2.0` as of #1674 (`09b69765`); this branch merged `main` at `4e7c1ac0`.
- All **15** `grafana:` reftargets in the guide — the 12 this PR changes plus the 3 pre-existing ones — resolve through this repo's own `toGrafanaSelectorForVersion` with **0 failures**, checked at `12.3.9`, `13.1.1`, `13.2.0`, `13.2.0-<build>` and `latest`. `pages.Explore.General.addQueryRow` resolves (contra the Bugbot comment suggesting `addQueryRowButton`); `components.DataSource.TestData.QueryTab.labelsInput` and `components.AddToDashboard.confirmButton` now exist.

## Minimum Grafana version

Taking the explicit call the review asked for: **accept the pre-13.2 gap for this guide rather than gating its content.** Five steps depend on DOM that only exists from 13.2:

| Reftarget | Floor |
| --- | --- |
| `components.DataSource.TestData.QueryTab.labelsInput:A` / `:B` | 13.2.0 |
| `pages.Explore.General.addQueryRow` | 13.2.0 |
| `components.AddToDashboard.confirmButton` | 13.2.0 |
| `pages.DataSources.dataSourceAddButton` | constant since 12.4.0, but wired onto `GetStartedWithDataSource.tsx` only in 13.2 |
| `components.UnitPicker.container` | 13.1.0 |

Why accepting is reasonable here: the guide is `targetPlatform: cloud`-gated, and Cloud is on 13.2 (confirmed on a live stack, `13.2.0-31683987170`). The resolver's fallback is fail-forward, so a below-floor stack resolves to the 13.2 testid and the step blocks with no error and no log line — which is exactly why leaving `testEnvironment.minVersion` at `12.2.0` was wrong. It now says `13.2.0`, so the e2e runner's manifest preflight fails with "below the required minimum" instead of stalling mid-guide.

Not added: per-step `min-version:13.2.0` runtime requirements. They would turn a silent block into an explicit message for a user on a pre-13.2 stack, but the cloud targeting already makes that population near-empty. Easy to add later if we'd rather be loud.

## Test plan

- [x] Manually walked the full guide (datasource setup through save dashboard) against `pathfinder.grafana-dev.net` via Playwright; created and cleaned up a test TestData datasource + dashboard used for verification.
- [x] Live DOM verification of every converted reftarget on `learn.grafana.net` (Grafana `13.2.0-31683987170`) — each one lands on exactly the element the old CSS selector targeted.
- [x] Re-verified after the `@grafana/e2e-selectors` 13.2.0 bump: 15/15 reftargets resolve, 0 failures.
- [x] `npm run validate:strict` (10/10 guides valid), `npm run validate:packages` (10/10 valid), `npm run repository:check` clean, `jest src/validation src/lib/dom` (30 suites, 1168 tests).

## Follow-up

No CI guard exists yet that resolves `grafana:` reftargets in bundled guides against the pinned selectors package — that gap is what let this PR sit mergeable-by-accident with three unresolvable paths. Worth landing on its own, independently of this PR.


src/bundled-interactives/first-dashboard-cloud/content.json
```diff
@@ -108,7 +108,7 @@
         {
           "type": "interactive",
           "action": "formfill",
-          "reftarget": "input[id=\"labels-A\"]",
+          "reftarget": "grafana:components.DataSource.TestData.QueryTab.labelsInput:A",
```
moxious, line 112: **F1 (blocking)** — `components.DataSource.TestData.QueryTab.labelsInput` does not exist in the pinned `@grafana/e2e-selectors@13.1.1`. The `QueryTab` node there contains only: `scenarioSelectContainer`, `scenarioSelect`, `max`, `min`, `noise`, `seriesCount`, `spread`, `startValue`, `drop`.

It exists only in `13.2.0-*` prereleases:

```js
labelsInput: { "13.2.0": (refId) => `data-testid Test Data Query labels input ${refId}` }
```

npm `latest` is 13.1.2 and also lacks it, so `npm update` won't fix this — it needs a 13.2.x bump.

Until then `toGrafanaSelectorForVersion` throws, the raw `grafana:…` string is returned, `querySelectorAllEnhanced` swallows the `SyntaxError`, and `exists-reftarget` blocks this step. Same applies to `labelsInput:B` on line 128.

The param values themselves are right — `A`/`B` are correct Grafana query refIds.

src/bundled-interactives/first-dashboard-cloud/content.json
```diff
@@ -117,15 +117,15 @@
         {
           "type": "interactive",
           "action": "highlight",
-          "reftarget": "button:contains(\"Add query\")",
+          "reftarget": "grafana:pages.Explore.General.addQueryRow",
```
moxious, line 121: **F1 (blocking)** — `pages.Explore.General.addQueryRow` does not exist in the pinned 13.1.1. That node contains only `container`, `graph`, `table`, `scrollView`, `addFromQueryLibrary`.

Added in 13.2.0 as `{ "13.2.0": "data-testid explore add query row button" }`. This step blocks until the package is bumped.

src/bundled-interactives/first-dashboard-cloud/content.json
```diff
@@ -149,17 +149,17 @@
...
             {
               "action": "highlight",
-              "reftarget": "button:contains(\"Add to dashboard\")",
+              "reftarget": "grafana:pages.Explore.toolbar.add:Add to dashboard",
               "requirements": ["exists-reftarget"]
             },
             {
               "action": "highlight",
-              "reftarget": "button:contains(\"Open dashboard\")",
+              "reftarget": "grafana:components.AddToDashboard.confirmButton",
```
moxious, line 163: **F1 (blocking)** — `components.AddToDashboard` does not exist as a node in the pinned 13.1.1 at all (the only `Dashboard`-ish component keys are `DashboardEditPaneSplitter`, `DashboardLinks`, `DashboardImportPage`, `ImportDashboardForm`, `DashboardRow`).

Added in 13.2.0 as `AddToDashboard.confirmButton: { "13.2.0": "data-testid Add to dashboard confirm button" }`. This step blocks until the package is bumped.

src/bundled-interactives/first-dashboard-cloud/content.json
```diff
@@ -44,7 +44,7 @@
         {
           "type": "interactive",
           "action": "highlight",
-          "reftarget": "button:contains(\"Add new data source\")",
+          "reftarget": "grafana:pages.DataSources.dataSourceAddButton",
```
moxious, line 48: **F2** — this one is the sneaky case: it *resolves* fine today, so there's no error to alert you, but it won't match pre-13.2.

I diffed the definition across 13.1.1 and the 13.2.0 nightly and it's byte-identical:

```js
dataSourceAddButton: { "12.4.0": "data-testid data-source-add-button" }
```

So #128750 didn't change the constant — it wired the existing testid onto `GetStartedWithDataSource.tsx`. On Grafana < 13.2 the plugin-catalog detail page (which `on-page:/connections/datasources/grafana-testdata-datasource` pins this step to) still renders that button without the attribute, so `exists-reftarget` fails with no console error and no log line.

The `button:contains("Add new data source")` this replaces works on those versions today.

src/bundled-interactives/first-dashboard-cloud/content.json
```diff
@@ -99,7 +99,7 @@
         {
           "type": "interactive",
           "action": "formfill",
-          "reftarget": "input[id=\"test-data-scenario-select-A\"]",
+          "reftarget": "grafana:components.DataSource.TestData.QueryTab.scenarioSelect",
```
moxious, line 103: **F4 (low, no action required)** — the old `input[id="test-data-scenario-select-A"]` was refId-scoped; `scenarioSelect` resolves to an unparameterized aria-label (`Test Data Query scenario select`).

Correct as written: this step runs before **Add query** creates row B, so there's exactly one match. Just fragile under step reordering or re-entry, and the asymmetry stands out given `labelsInput` two lines down *is* parameterized. Grafana offers no parameterized variant here, so there's nothing to change — flagging so it's a known property rather than a surprise later.

Picking `scenarioSelect` over `scenarioSelectContainer` is the right call for a `formfill`.

moxious (changes requested):
## Review — request changes

Nice work on the selector picks — the live verification against Cloud 13.2 shows in the quality of the choices (`scenarioSelect` over `scenarioSelectContainer`, refId-parameterized `labelsInput`, `UnitPicker.container` for a highlight-only step). The problem here is entirely sequencing, not judgement.

I resolved all 12 changed reftargets against the pinned `@grafana/e2e-selectors@13.1.1` using this repo's own `toGrafanaSelectorForVersion` logic. **8 resolve, 4 do not** (3 distinct paths). Inline comments mark each one. The 3 pre-existing `grafana:` selectors in this file (lines 86, 189, 213) all resolve fine, so the gap is isolated to this PR's additions.

---

### 🔴 F1 — Three new selector paths don't exist in the pinned selectors package (blocking)

| Path | Steps | Status in 13.1.1 |
|---|---|---|
| `components.DataSource.TestData.QueryTab.labelsInput` | 2 (`:A`, `:B`) | missing |
| `pages.Explore.General.addQueryRow` | 1 | missing |
| `components.AddToDashboard.confirmButton` | 1 | missing |

`resolveSelectorForVersion` throws `Selector not found`, logs via `logger.error`, and returns the raw `grafana:…` string (`selector-resolver-core.ts:19-27`). That reaches `querySelectorAllEnhanced`, which catches the resulting `SyntaxError` and returns nothing — so `exists-reftarget` fails and those **4 steps are permanently blocked**. Graceful, not a crash, but the guide can't be completed.

Where the fix actually lives: these keys exist **only** in `13.2.0-*` prereleases. npm `latest` for `@grafana/e2e-selectors` is **13.1.2**, which also lacks all three — so `npm update` won't fix it. The entries are gated at exactly `"13.2.0"`:

```js
labelsInput:   { "13.2.0": (refId) => `data-testid Test Data Query labels input ${refId}` }
addQueryRow:   { "13.2.0": "data-testid explore add query row button" }
confirmButton: { "13.2.0": "data-testid Add to dashboard confirm button" }
```

grafana/grafana#128750 **is merged** (2026-07-30, milestone 13.2.x), so the only blocker is the package bump here — exactly what your unchecked test-plan box says. The issue is that **this PR isn't marked draft while #1382 is**, and both carry the same unmet dependency, so #1383 is mergeable by accident. Suggest marking it draft to match, or adding a blocking checklist item.

### 🟠 F2 — After the bump, 5 steps still fail *silently* on Cloud stacks below 13.2

The resolver's version fallback is fail-**forward**, not fail-closed (`resolver.mjs`):

```js
if (!versionToUse) { versionToUse = versions[versions.length - 1]; }
```

On Grafana 12.3–13.1 no key satisfies `gte`, so it falls back to the newest and returns the 13.2-only testid. Resolution *succeeds*, the DOM attribute is absent, `exists-reftarget` fails, and the step blocks with **no console error and no log line** — unlike F1, which is at least visible in Faro.

This also covers `pages.DataSources.dataSourceAddButton` (line 47). I diffed its definition across both packages: byte-identical (`{"12.4.0": "data-testid data-source-add-button"}`), unchanged by #128750 — so the constant resolves today, but the DOM wiring onto `GetStartedWithDataSource.tsx` is new in 13.2, as your description notes.

Why it's a regression rather than just a wait: the CSS selectors being replaced work on 12.3–13.1 **today**. `targeting.match` limits this guide to `targetPlatform: cloud`, which narrows blast radius, but Cloud stacks do run below 13.2 during progressive rollout and on pinned versions. Neither `testEnvironment.minVersion` (`12.2.0`) nor `grafanaDependency` (`>=12.3.0-0`) gates this away.

**Ask:** make the call explicitly in the description rather than leaving it implicit — either accept the pre-13.2 gap and say so, or gate the guide. The next person converting a guide will hit this.

### 🟠 F3 — No CI guard resolves `grafana:` reftargets in bundled guides

`src/validation/bundled-guides.test.ts` passes **31/31 with three unresolvable paths in the tree**. It checks JSON parse + schema + strict mode only; `validate-guide.ts` has zero references to `reftarget`, and nothing in `src/validation/**` imports the selector resolver.

That gap is why F1 sits behind green CI, and why this PR reads as ready to merge.

Worth its own PR ahead of the guide conversions: walk the bundled guides, extract every `grafana:` reftarget (including embedded `{grafana:…}` tokens), and assert each resolves through `toGrafanaSelectorForVersion`. ~20 lines against machinery that already exists, and it protects #1382 and every future conversion.

### 🟡 F4 — `scenarioSelect` loses per-query-row scoping

See inline on line 102. Correct as written, worth knowing about.

---

### Reviewed clean

- **security** — static authored strings, funnelled through `toCssAttributeString` escaping into attribute selectors. No new sink, no trust-boundary change.
- **reversibility** — `reversible`. Content-only; no persisted state, schema, `data-test-*`, or telemetry-payload change. Revert restores the working CSS selectors exactly.
- **cross-cutting architecture** — respects the intended `grafana:` indirection; no engine-boundary violation; no new UI text.
- **Neutral:** `pages.Explore.toolbar.add:Add to dashboard` bakes an English label into a testid and so breaks on a localized Grafana — but it replaces `button:contains("Add to dashboard")`, which had the identical dependency. Not a regression.
- Pre-existing, not introduced here: `manifest.json` declares `testEnvironment.minVersion: "12.2.0"` while `src/plugin.json` requires `grafanaDependency: ">=12.3.0-0"`.

### Suggested order

1. Land the F3 validation test — converts this whole class of bug from "caught by a careful reviewer" to "caught by CI".
2. Bump `@grafana/e2e-selectors` to a 13.2.x line.
3. Land #1382 and #1383 together, with an explicit answer to F2.

🤖 Review generated with [Claude Code](https://claude.com/claude-code)

Jayclifford345:
## Live DOM verification on Cloud 13.2 — every selector pick is correct; F1 is unchanged

Followed up on the "Cloud is already on 13.2" question by testing against a real Cloud stack: `learn.grafana.net`, Grafana **`13.2.0-31683987170`**, authenticated session.

**Result: all seven `grafana:` reftargets this PR touches resolve to exactly the element the old CSS selector was targeting.** The DOM side of this PR is fully validated.

| Reftarget | Present | Element it lands on | Replaces |
|---|---|---|---|
| `pages.Explore.General.addQueryRow` | ✅ 1 | add query row button | `button:contains("Add query")` |
| `components.DataSource.TestData.QueryTab.labelsInput:A` | ✅ 1 | `input#labels-A` | `input[id="labels-A"]` |
| `components.DataSource.TestData.QueryTab.labelsInput:B` | ✅ 1 | `input#labels-B` | `input[id="labels-B"]` |
| `components.AddToDashboard.confirmButton` | ✅ 1 | **Open dashboard** button | `button:contains("Open dashboard")` |
| `pages.DataSources.dataSourceAddButton` | ✅ 1 | **Add new data source** | `button:contains("Add new data source")` |
| `pages.Explore.toolbar.add:Add to dashboard` | ✅ 1 | menu item | `button:contains("Add to dashboard")` |
| `components.DataSource.TestData.QueryTab.scenarioSelect` | ✅ 1 | `input#test-data-scenario-select-A` | `input[id="test-data-scenario-select-A"]` |

Notably `dataSourceAddButton` — the #128750 DOM wiring onto `GetStartedWithDataSource.tsx` — is confirmed live: the **Add new data source** button on `/connections/datasources/grafana-testdata-datasource` now carries `data-testid data-source-add-button`.

### F4 reproduced exactly

Counted `aria-label="Test Data Query scenario select"` matches either side of clicking **Add query**:

```
before Add query: 1
after  Add query: 2
```

The original read was precisely right — correct as written because the step runs before row B exists, fragile only under step reordering or re-entry. No change needed.

### F1 is unchanged by any of this

Worth stating plainly, because "Cloud is on 13.2" reads like it should fix F1 and it doesn't.

Selector **constants ship in the npm package**; the runtime Grafana version is only the *lookup key* into that package's tree. `getGrafanaVersion()` (`src/lib/dom/grafana-selector.ts`) reads `config.buildInfo.version`, but `toGrafanaSelectorForVersion` walks the tree built from the installed `@grafana/e2e-selectors`. With 13.1.3 installed, `labelsInput` / `addQueryRow` / `AddToDashboard` are not keys in that object at all — so the path-walk throws regardless of what the stack reports. Re-ran the resolver using this stack's exact version string to confirm:

```
installed @grafana/e2e-selectors = 13.1.3
===== grafanaVersion 13.2.0-31683987170 =====
  FAIL  grafana:components.AddToDashboard.confirmButton   (AddToDashboard is undefined)
  FAIL  grafana:...QueryTab.labelsInput:A                 (labelsInput is undefined)
  FAIL  grafana:...QueryTab.labelsInput:B                 (labelsInput is undefined)
  FAIL  grafana:pages.Explore.General.addQueryRow         (addQueryRow is undefined)
```

So there is **no resolver logic change that unblocks this** — there is no value to resolve to. The package bump is the only lever, and npm still has no stable 13.2.x:

```
latest:  13.1.3               ← main's pin
nightly: 13.2.0-31059932060
stable 13.2.x: NONE
```

There's an asymmetry worth flagging: Cloud is running Grafana build `31683987170`, while the newest *published* selectors package is build `31059932060`. Cloud is ahead of anything on npm.

### One thing that is already correct

`resolveSelectors` strips the prerelease suffix (`grafanaVersion.replace(/\-.*/, "")`, `resolver.mjs:5`), so Cloud's build-suffixed `13.2.0-<runId>` is treated as plain `13.2.0` and maps to the right tier. Verified end-to-end: `saveAsTitleInput` resolves to the 13.2 `data-testid Save dashboard title field` at the prerelease string rather than falling back to the older aria-label form. No workaround needed there.

### Where this leaves F2

This materially weakens F2 for **this** guide. The finding was about pre-13.2 Cloud stacks failing silently; the guide is `targetPlatform: cloud`-gated, and Cloud is on 13.2. That makes "accept the pre-13.2 gap" an evidence-backed call rather than a guess — which is exactly the explicit answer the review asked for.

Caveat to keep honest: this is one Grafana Labs learning stack, likely on a faster release channel than customer stacks. It supports "Cloud is on 13.2", not "no Cloud stack is below 13.2". If the fleet is known to be fully on 13.2, that's the sentence for the description.

The general hazard behind F2 doesn't go away for future conversions — a too-new selector still resolves successfully via fail-forward and fails with no error.

### Updated state of the four findings

| Finding | Status |
|---|---|
| 🔴 F1 — 3 paths missing from the pinned package | **still blocking**; needs the bump, no stable 13.2.x exists yet |
| 🟠 F2 — silent pre-13.2 failure | **materially weakened** for this cloud-gated guide; still worth one explicit sentence |
| 🟠 F3 — no CI guard | **still open, and now the clear top priority** — blocked on nothing, would have caught all of this |
| 🟡 F4 — `scenarioSelect` scoping | **confirmed live** (1 → 2 matches); no action |

Two open decisions: whether to pin `13.2.0-31059932060` now versus waiting for a stable 13.2.x, and landing F3's guard independently of either.

Verified with no side effects on the stack — no dashboard created, no datasource added.

🤖 Live verification generated with [Claude Code](https://claude.com/claude-code)

moxious (changes requested):
Independent review run against this repo's own review orchestrator. Requesting changes on three issues that stop the guide working for real users.

**[blocking] The retained TestData link matches nothing.** `content.json:39` keeps `a[href="/connections/datasources/grafana-testdata-datasource/"]`, but Grafana builds that URL without the trailing slash: `connections/constants.ts` defines `DataSourcesDetails: '/connections/datasources/:id'`, `usePluginFiltering.ts` substitutes the plugin id without appending `/`, and `CardGrid.tsx` passes the result straight to `href`. Exact attribute matching does not normalise the two forms, so this step highlights nothing — confirmed with a live CSS-semantics check against the rendered card. The peer Loki and Prometheus guide selectors already use the slashless form. Flagging it specifically because the PR body says this link was assessed and deliberately retained as stable.

```suggestion
"reftarget": "a[href=\"/connections/datasources/grafana-testdata-datasource\"]",
```

**[blocking] Four converted steps cannot resolve at this commit, and it is not a type-only problem.** The branch pins `@grafana/e2e-selectors@13.1.1`. That version and current stable 13.1.3 both lack `components.DataSource.TestData.QueryTab.labelsInput` (A and B), `pages.Explore.General.addQueryRow`, and `components.AddToDashboard.confirmButton`; only nightly `13.2.0-31059932060` carries them. On a missing path `resolveSelectorForVersion` catches the exception and returns the raw `grafana:...` token (`src/lib/dom/selector-resolver-core.ts:19-26`); `querySelectorAllEnhanced` then treats that token as an invalid selector and returns an empty list (`src/lib/dom/enhanced-selector.ts:30-74`); and these steps all require `exists-reftarget`. Lines 111, 120, 128 and 162 therefore block the guide even on a Grafana 13.2 UI. Updating the external nightly does not change this application bundle. Either retain the previous raw targets until a stable release is consumable, or wait for that release, pin it here, and re-run the complete Cloud guide.

**[blocking] The conversion drops compatibility with older supported Cloud stacks.** With the nightly tree installed the paths resolve, but the target DOM is versioned later than versions this plugin admits: `pages.Explore.toolbar.addTo` and `.add` at 12.4.0, `components.UnitPicker.container` at 13.1.0, and the TestData label inputs, `addQueryRow` and `AddToDashboard.confirmButton` at 13.2.0. `src/plugin.json:76` declares `>=12.3.0-0` and the guide has no runtime version gate. `manifest.json:21`'s `testEnvironment.minVersion` only drives E2E preflight (`docs/developer/package-authoring.md:168-175`), and the versioned resolver fail-forwards when the running version predates the first selector entry, emitting a selector whose DOM attribute is absent. Verifying one 13.2 stack does not establish a fleet-wide floor. Either preserve version-compatible fallbacks, or provide an enforceable invariant that every environment eligible to receive this guide is 13.2 or later.

Also worth noting: upstream PRs #128758 and #128759 are required by this guide alongside #128750, which is the only one the description names.

**[question] Two items look outside this PR's scope to me. Your call whether to file them, defer them, or drop them — I have not filed anything.**

- No stable 13.2 `@grafana/e2e-selectors` has been published. The definitions exist in upstream source and in the nightly, so this is a publication lag rather than a wrong path. Related: the automated suggestion of `pages.Explore.General.addQueryRowButton` earlier in this thread is incorrect — upstream source and the nightly both use `addQueryRow`.
- Guide validation does not exercise selector semantics. `npm run build:cli`, `npm run validate:strict` (10/10 valid) and the selector test suites (79 passing) are all green at this head while a `grafana:` path is absent from the pinned package. `.github/workflows/validate-guides.yml:50-54` checks structure only, and `playwright.config.ts:20-21` excludes the e2e-runner. A Node-safe check walking every `grafana:` path against the pinned package would have caught the second finding above in CI.

Jayclifford345:
## F1 cleared by the package bump; F2 answered explicitly

`main` now pins `@grafana/e2e-selectors@13.2.0` (#1674, `09b69765`), and this branch merged `main` at `4e7c1ac0`.

**F1 — resolved.** Re-ran the resolution over every `grafana:` reftarget in the guide (12 changed + 3 pre-existing = 15), splitting `path:param` the same way `resolveSelectorForVersion` does and walking the tree built from the installed package:

```
15 distinct grafana: reftargets — total failures: 0

components.DataSource.TestData.QueryTab.labelsInput:A -> "data-testid Test Data Query labels input A"
components.DataSource.TestData.QueryTab.labelsInput:B -> "data-testid Test Data Query labels input B"
pages.Explore.General.addQueryRow                     -> "data-testid explore add query row button"
components.AddToDashboard.confirmButton               -> "data-testid Add to dashboard confirm button"
```

All 15 resolve at `12.3.9`, `13.1.1`, `13.2.0`, `13.2.0-31683987170` and `latest`. The three formerly-missing paths are the ones quoted above. This also settles the Bugbot comment on line 120: `pages.Explore.General.addQueryRow` is the real key — there is no `addQueryRowButton`.

**F2 — accepting the pre-13.2 gap, and saying so.** The description now carries the per-selector floor table and the reasoning. One concrete change came out of your observation that neither `testEnvironment.minVersion` nor `grafanaDependency` gates this away: `minVersion` was `12.2.0`, which is simply no longer true of this guide, so it's now `13.2.0` (`repository.json` regenerated to match, `repository:check` clean). Below the floor the e2e runner's manifest preflight now fails with "below the required minimum" instead of the guide stalling mid-run with no log line. Per-step `min-version:13.2.0` runtime requirements were considered and left out — the `targetPlatform: cloud` targeting already makes the affected population near-empty — but they're a one-line-per-step addition if you'd rather be loud about it.

**F3 — agreed, and still open.** Nothing in `src/validation/` resolves `grafana:` reftargets against the pinned package today; that's exactly the guard that would have caught F1 before review. Noted as a follow-up in the description rather than bundled here.

**F4 — no action**, as previously confirmed live (1 → 2 `scenarioSelect` matches across **Add query**).

Local checks on the merged branch: `validate:strict` 10/10, `validate:packages` 10/10, `repository:check` clean, `jest src/validation src/lib/dom` 30 suites / 1168 tests passing.

moxious (changes requested):
**[blocking] The 13.2 floor is not enforced at runtime.**

`testEnvironment.minVersion` only controls E2E routing. Runtime targeting still admits any Cloud stack supported by the plugin (`>=12.3`), while several new selectors require 13.2. Users on 12.3–13.1 can launch the guide and become blocked.

Add a runtime `min-version:13.2.0` requirement or retain compatible selectors below 13.2.

**[blocking] The changed selector contract lacks automated coverage.**

Twelve executable selectors changed, but CI only validates JSON/schema structure. Add a data-driven test that loads this guide, resolves every `grafana:` reference at 13.2, rejects unresolved-token fallback, and ideally exercises the changed flow in E2E.

Jayclifford345:
Pushed `598dc5d9`, which resolves both blocking findings.

**The 13.2 floor is now enforced at runtime.** I checked each converted path against the versioned tree in `@grafana/e2e-selectors` 13.2.0. Seven of the twelve postdate this plugin's own `>=12.3.0-0` floor:

| introduced | path |
| --- | --- |
| 12.4.0 | `pages.DataSources.dataSourceAddButton`, `pages.Explore.toolbar.addTo`, `pages.Explore.toolbar.add` |
| 13.1.0 | `components.UnitPicker.container` |
| 13.2.0 | `…TestData.QueryTab.labelsInput`, `pages.Explore.General.addQueryRow`, `components.AddToDashboard.confirmButton` |

The failure mode is quieter than "not found": `resolveSelectors` floors an out-of-range target to the *lowest* entry in a selector's version map, so on 12.3 these resolve to their 13.2 value and match nothing. No throw, no fallback, just `exists-reftarget` that never passes.

All three sections now carry `min-version:13.2.0`, so a reader below the floor gets "This feature requires Grafana version 13.2.0 or higher" instead of a step that silently never unblocks. I gated at guide level rather than reverting the four 13.2-only steps — the guide is unfinishable below 13.2 either way, and reverting would leave the conversion half-done. One limitation worth knowing: `targeting.match` supports only `urlPrefix`/`urlPrefixIn`/`targetPlatform`, so the guide can still be *recommended* below 13.2; what changed is that it now says why.

**The selector contract has automated coverage.** `src/validation/bundled-guide-selectors.test.ts` is data-driven over every bundled guide (22 tests, 7 guides with tokens). Per guide it asserts that every `grafana:` reftarget resolves at the declared floor rather than falling back to the raw token string, that `testEnvironment.minVersion` covers every token used, and that any token above the plugin floor sits under a `min-version:` gate that covers it. Mutation-checked all three: reverting the manifest to 12.2.0 names all eight offending uses with their introduction versions, dropping the section gates fails the third, and a typo'd path fails the first.

The E2E half of that finding I did not do, and it is not an oversight I can fix here: this guide is `tier: cloud`, and `checkTier` skips cloud guides in a local-tier run, so CI's Playwright matrix (local Grafana images) cannot exercise it. It is runnable through the e2e CLI against a Cloud 13.2 stack.

Also added a short section to `docs/developer/interactive-examples/selectors-reference.md` on the version-floor trap, since "version-aware" reads as "works everywhere" — it means the *value* tracks the version, not that the selector exists.

`npm run check` clean; full frontend suite green.

---

**Follow-up, and the reason the section gates are the tactical fix rather than the permanent one:** hand-gating every section of every guide does not scale, and it does nothing for guides whose version dependency isn't a selector at all. [#1714](https://github.com/grafana/grafana-pathfinder-app/pull/1714) adds a `minGrafanaVersion` field to the package manifest and a warn-only notice at the top of the guide, driven by the running Grafana rather than by an e2e check. Once that merges, this guide should declare `"minGrafanaVersion": "13.2.0"` in its manifest, and the floor source in `bundled-guide-selectors.test.ts` should prefer it over `testEnvironment.minVersion`.

moxious (approved):

Issue #1359, opened 2026-07-15
Pay down remaining import cycles caught by the architecture ratchet


moxious:
Tracking issue for the circular-dependency clusters baselined into `ALLOWED_CYCLES` when file-level cycle detection was added (PR #1358). These are grandfathered so CI is green; each should be paid down opportunistically using the patterns demonstrated on that PR.

Remaining clusters (as of the #1358 baseline, 5 total):

- [ ] **telemetry + OpenFeature** across `lib`/`security`/`utils` (9 files) — the largest; needs a dedicated extraction pass
- [ ] **`requirements-manager/checks/*`** + shared utils (7 files)
- [ ] **`interactive-engine` ↔ `requirements-manager`** (5 files) — also tracked in `ALLOWED_LATERAL_VIOLATIONS`
- [ ] **`components/interactive-tutorial/*`** (4 files)
- [ ] **`components/docs-panel/components/*`** (4 files) — barrel-routed; likely a shared-type extraction

Worked-example patterns to copy (both on #1358):
- Extract the shared type to a Tier-0 leaf — `src/types/link-interception.types.ts`
- Inject the dependency instead of importing upward — `createBoundedRecordStorage` in `src/lib/storage/bounded-record-storage.ts`

As each cluster is broken, remove its entry from `ALLOWED_CYCLES` (the ratchet's stale-entry check enforces this).

moxious:
## 🧭 Implementation readiness

**Stakes:** Nothing changes for the user. This is internal work. Two circular-dependency clusters remain grandfathered in the architecture ratchet, so a new cycle inside either one cannot be told apart from the old ones.

**In plain words:** A tracking issue for the import cycles that were baselined when file-level cycle detection was added.

**Sanity check:** This is a tracking issue for a list of clusters rather than one change. Its checklist is also out of date: three of the five clusters have been paid down and only two remain in the allowlist.

**Severity:** 🟢 Low — internal architecture only. Continuous integration stays green either way.

**Code basis:**
- `src/validation/architecture.test.ts:228` — The cycle allowlist. Two entries remain, each tagged to this issue, so the current state is readable here rather than from the checklist.

**Test coverage:** src/validation/architecture.test.ts is itself the enforcing test. A cluster is paid down when its entry can be deleted and the suite still passes.

**Labels:** Already labeled correctly.

**Open judgment calls:**
None. The implementation approach is unambiguous.

**Traps for the implementer:**
[...]

PR #1391, opened 2026-07-20
refactor(architecture): dissolve three import-cycle clusters via leaf extractions

moxious:
## What

Pays down three of the five grandfathered `ALLOWED_CYCLES` clusters from the #1358 baseline, using the two patterns demonstrated on that PR (extract the shared symbol to a leaf; don't import upward for identity/type access). Their allowlist entries are removed; the ratchet's stale-entry check confirms the SCCs are gone.

**Part of #1359 — does not close it.** The two remaining clusters (telemetry + OpenFeature across `lib`/`security`/`utils`, and the `interactive-engine <-> requirements-manager` lateral) need runtime inversions and are left for follow-up PRs.

### `requirements-manager/checks/*` (7 files)

Every `checks/* -> requirements-checker.utils.ts` back-edge was the same single `import type { CheckResultError }`. The type moves to the Tier-0 `types/requirements.types.ts` leaf; the `requirements-manager` barrel exports it from there, so the public surface is unchanged.

### `components/interactive-tutorial` (4 files)

Two cuts:

- `use-section-requirements.ts` imported only `DEFAULT_INTERACTIVE_SECTION_TITLE` from `interactive-section.tsx`. The title constants move to a new `section-titles.ts` leaf; the barrel and tests import it directly.
- `section-numbering.tsx` imported `InteractiveConditional` solely for a component-identity check. The conditional now tags itself via a shared marker leaf (`skip-section-numbering.ts`) and the numbering helper checks the marker. This also removes the conditional from the cycle-load-order hazard documented in `section-numbering.tsx`'s header. The genuine render recursion (conditional renders `InteractiveSection`) is untouched and is now a one-way edge.

### `components/docs-panel` (4 files)

`DocsPanelContentArea` and `LearningJourneyMilestoneToolbar` imported the `CombinedLearningJourneyPanel` class from `docs-panel.tsx` as a prop type. Both now type against the existing `DocsPanelModelOperations` interface in `components/docs-panel/types.ts` (extended with `confirmAlignment`/`dismissAlignment`, which the class already exposes publicly — this was the interface's stated purpose, "Pattern E: interface-first decoupling").

## Verification

- `src/validation/architecture.test.ts`: 23/23 pass; ratchet log now reports `cycles: clusters=2` (down from 5)
- `tsc --noEmit` clean, lint 0 errors, prettier clean
- Full suite: 356 suites / 5,530 tests pass

## Review pass

An independent tech-debt-focused review (per `.cursor/skills/review/SKILL.md` + the techdebt lens) found 0 blockers and confirmed all three clusters are genuine graph surgery rather than allowlist appeasement. Its 7 minor/nit findings are addressed in the second commit: marker survives `React.memo`/`forwardRef` wrapping, the test-harness conditional mock self-tags like production, the `RemoteRequirementError` wire mirror got an honest rationale + a compile-time assignability guard, both single-consumer back-compat re-exports were dropped in favor of direct leaf imports, two QC8 narration comments removed, and the toolbar test mock now types against the operations interface.

### DIFF
--- a/src/components/interactive-tutorial/skip-section-numbering.ts
+++ b/src/components/interactive-tutorial/skip-section-numbering.ts
@@ -1,0 +1,22 @@
+/**
+ * Opt-out marker for section child numbering.
+ *
+ * `section-numbering.tsx` excludes wrapper components from the `1. 2. 3.`
+ * sequence. Components tag themselves at definition time instead of
+ * `section-numbering` importing them for identity checks — importing
+ * `InteractiveConditional` there closed an import cycle through
+ * `interactive-section.tsx` (#1359).
+ */
+
+import type React from 'react';
+
+const SKIP_SECTION_NUMBERING = Symbol.for('pathfinder.skipSectionNumbering');
+
+export function markSkipsSectionNumbering<T extends React.ComponentType<any>>(component: T): T {
+  (component as unknown as Record<symbol, boolean>)[SKIP_SECTION_NUMBERING] = true;
+  return component;
+}
+
+export function skipsSectionNumbering(type: unknown): boolean {
+  return typeof type === 'function' && (type as unknown as Record<symbol, boolean>)[SKIP_SECTION_NUMBERING] === true;
+}
--- a/src/components/interactive-tutorial/skip-section-numbering.ts
+++ b/src/components/interactive-tutorial/skip-section-numbering.ts
@@ -18,5 +18,8 @@ export function markSkipsSectionNumbering<T extends React.ComponentType<any>>(co
 }
 
 export function skipsSectionNumbering(type: unknown): boolean {
-  return typeof type === 'function' && (type as unknown as Record<symbol, boolean>)[SKIP_SECTION_NUMBERING] === true;
+  // memo/forwardRef components are exotic objects, not functions — the marker
+  // survives wrapping only if we read it off objects too.
+  const canCarryMarker = typeof type === 'function' || (typeof type === 'object' && type !== null);
+  return canCarryMarker && (type as unknown as Record<symbol, boolean>)[SKIP_SECTION_NUMBERING] === true;
 }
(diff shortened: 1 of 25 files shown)


moxious:
@cursoragent review

src/components/interactive-tutorial/skip-section-numbering.ts
```diff
@@ -0,0 +1,25 @@
...
+  (component as unknown as Record<symbol, boolean>)[SKIP_SECTION_NUMBERING] = true;
+  return component;
+}
+
+export function skipsSectionNumbering(type: unknown): boolean {
+  // memo/forwardRef components are exotic objects, not functions — the marker
+  // survives wrapping only if we read it off objects too.
+  const canCarryMarker = typeof type === 'function' || (typeof type === 'object' && type !== null);
+  return canCarryMarker && (type as unknown as Record<symbol, boolean>)[SKIP_SECTION_NUMBERING] === true;
+}
```
leslie-heinzen, line 25: Nice.

nit/food for thought: This works, but the intent is kind of hard to grok, and the double-casting is a bit of a code smell that there's a clearer design available, IMO.

A map to register components has a more obvious intent I think; also allows for stricter typing:

```ts
const SKIP_REGISTRY = new WeakMap<object, boolean>();

export function markSkipsSectionNumbering<T extends React.ComponentType<any>>(component: T): T {
  SKIP_REGISTRY.set(component, true);
  return component;
}

export function skipsSectionNumbering(type: unknown): boolean {
  if (typeof type !== 'function' && (typeof type !== 'object' || type === null)) {
    return false;
  }
  return SKIP_REGISTRY.has(type as object);
}
```

If we end up using this pattern of marking components for opt-out for other properties, I might consider moving to something like the above.

leslie-heinzen (approved):

src/components/interactive-tutorial/skip-section-numbering.ts
moxious: The double cast doesn't read great, admitted.

The reason I like the symbol for now is that the marker travels on the component object itself, and Symbol.for() is globally interned and keeps working even if this module ends up duplicated across chunks/test envs. A module-level WeakMap is per-instance state and could fail silently. Symbols also shows up inspecting a component in debugging.

If we grow a second opt-out property, I agree this should generalize

PR #1401, opened 2026-07-22
docs(design): add the backend App Platform proxy pattern

moxious:
## Summary

Adds [`docs/design/BACKEND_PROXY_PATTERN.md`](https://github.com/grafana/grafana-pathfinder-app/blob/81ec6c0aa924805021c6f3633fed95642aee34ec/docs/design/BACKEND_PROXY_PATTERN.md) — the canonical pattern for plugin-backend routes that proxy the App Platform aggregator (paginated LIST → shaped in-process cache → availability-first serving, riding the caller's own identity). It is the synthesis of two independent design reviews of PR #1398 (completion records) and PR #1400 (custom guide catalogue), with every contested claim verified against the PR diffs, the baseline `package_recommendations.go`, and repo history. Future proxies of this shape should implement the pattern rather than re-derive it.

## What to look at

- [`docs/design/BACKEND_PROXY_PATTERN.md`](https://github.com/grafana/grafana-pathfinder-app/blob/81ec6c0aa924805021c6f3633fed95642aee34ec/docs/design/BACKEND_PROXY_PATTERN.md) — the load-bearing section is §3 (caller identity), where the two source reviews disagreed and runtime evidence decided: Grafana strips the inbound `Authorization` header before plugin resource handlers, the only runtime-verified outbound credential shape is `Authorization: Bearer <id-token>` plus `X-Grafana-Id` (dev-stack smoke, commit `89d6bd5e` on `feat/external-import-api`), and the session `Cookie` has never been needed. Also note the dated appendix of conformance gaps in #1398/#1400 — it is intentionally deletable once both PRs conform, so the durable pattern doesn't carry stale per-PR state.
- [`docs/developer/CONTEXT_INDEX.md`](https://github.com/grafana/grafana-pathfinder-app/blob/81ec6c0aa924805021c6f3633fed95642aee34ec/docs/developer/CONTEXT_INDEX.md) — one new row under "Go backend" routing the doc on `pkg/plugin/*.go`; the rest of the hunk is prettier realigning the table columns.

## Why

PRs #1398 and #1400 built the same proxy shape contemporaneously and diverged on nearly every load-bearing decision — namespace source (trusted context vs query param), outbound credential set, failure semantics (stale-serve + cooldown vs flat 503), pagination, and testability seams. Rather than resolving each divergence ad hoc in review threads, the two reviews were cross-checked and merged into one normative document, so the decisions are made once, backed by verified evidence, and cited by path in future PRs.

## How to verify

Documentation only — CI has nothing to exercise. Reviewer spot-checks:

1. Verify the §3 runtime-evidence claims against commits `a60cd778` and `89d6bd5e` on the `feat/external-import-api` branch (aggregator 403 with the plugin SA token; 502 "authorization header missing" until the ID-token switch).
2. Cross-check the appendix gap lists against the current diffs of #1398 and #1400 — each bullet should correspond to real code in those branches as of 2026-07-22.
3. Confirm the new CONTEXT_INDEX row renders correctly in the "Go backend" table.

## Breaking changes

None. Documentation only; the pattern binds future code, not existing behavior.

## Out of scope

- Applying the appendix's conformance changes to #1398 and #1400 — those land in the PRs themselves.
- The unsigned-JWT trust-boundary write-up in `docs/developer/CODA.md` called for by §3 — it should land with whichever PR ships the shared identity helper.

Refs #1398, refs #1400.

### DIFF
--- a/docs/design/BACKEND_PROXY_PATTERN.md
+++ b/docs/design/BACKEND_PROXY_PATTERN.md
@@ -59,8 +59,10 @@ the fixed internal aggregator.
 
 ### Inbound (browser → plugin)
 
-- Fail closed: absent or structurally invalid identity → refuse cleanly. Never guess, never fall
-  back to `X-Grafana-User` or a numeric id, never use a service account.
+- Fail closed: absent or structurally invalid identity → serve no data. Never guess, never fall
+  back to `X-Grafana-User` or a numeric id, never use a service account. On GET reads the refusal
+  is expressed as the §7 capability envelope (soft-200), not a 401 — "fail closed" constrains
+  _what_ is served (nothing), not the status code.
 - **Every proxy structurally validates the ID token** before spending an upstream call:
   well-formed JWT, `exp` **present and** unexpired. **Reject `exp == 0`** — a forwarded Grafana
   ID token always carries `exp`, and accepting its absence weakens the one structural check we
@@ -81,7 +83,10 @@ the fixed internal aggregator.
 - The only runtime-verified configuration (dev-stack smoke, commit `89d6bd5e` on
   `feat/external-import-api`) is `Authorization: Bearer <id-token>` **+** `X-Grafana-Id`, both
   synthesized from the inbound ID token, with the `idForwarding` toggle on (standard on Cloud).
-  Start there; if a live smoke proves `X-Grafana-Id` alone suffices, narrow to that and record it.
+  That evidence is from the sibling guides-import proxy — same aggregator, different kind — not
+  from these routes themselves; extrapolating is reasonable, but each new proxy confirms the
+  header set via its own §10 runtime smoke. Start with both headers; if a live smoke proves
+  `X-Grafana-Id` alone suffices, narrow to that and record it.
 - **Never forward `Cookie`.** No branch in this repo's history has ever needed it against the
   aggregator; the caller's full session is the broadest possible ambient grant and the classic
   confused-deputy shape.
@@ -156,9 +161,12 @@ The baseline's model — error cached sticky for the full 6 h TTL, no stale-serv
   **transient hiccup**.
 - Structural unavailability is signaled **in-band**: HTTP 200 with
   `capability: { available: false, reason: "identity-unavailable" | "backend-unavailable" }`.
-  A bare 503 conflates "never works here" with "blip", and the front-end's session-latching 503
-  heuristics will turn a hiccup into a feature that stays dark. This is also why missing identity
-  on a GET read is soft-200, not 401: these routes gate whether a feature renders at all.
+  A bare 503 conflates "never works here" with "blip": the front-end already lumps 503 into its
+  not-rolled-out status set (`UNAVAILABLE_STATUSES` in `src/utils/fetchBackendGuides.ts`, mirrored
+  in `src/context-engine/context.init.ts`) and silently renders empty with no retry, so a
+  transient 503 darkens the feature for that load exactly as if it were structurally absent. This
+  is also why missing identity on a GET read is soft-200, not 401: these routes gate whether a
+  feature renders at all.
 - **"Unavailable" ≠ "empty result."** `{items: []}` must mean the user genuinely has none.
 - A capability probe route makes the same transient/terminal distinction as the data route — a
   probe that flips `false` during a 30-second blip greys out UI for everyone.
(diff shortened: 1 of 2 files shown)


tomglenn (commented):
## Review: Conditional GO — one blocking appendix fix

I cross-checked this against the two source PRs (#1398, #1400), the baseline `package_recommendations.go`, and the referenced runtime-evidence commits. The normative pattern (§1–§10 + checklist) is sound and, unusually, every empirical claim I could check verifies against real code and history. One material gap in the appendix needs fixing before this becomes the alignment source of truth, plus a few minor items.

### Verified
- **§3 runtime evidence is real and faithfully characterized.** `a60cd778` (plugin-SA token → aggregator 403) and `89d6bd5e` ("Grafana strips Authorization before the request reaches plugin resource handlers", 502 "authorization header missing", switched to forwarding the value as **both** `Authorization: Bearer <id>` and `X-Grafana-Id`; `idForwarding` dependency; `grafana-slo-app` precedent) both exist and say exactly what §3 summarizes.
- **"No toggle constant on main"** — correct; #1398 (`pathfinderBackendAggregationToggle`) and #1400 (`customGuideBackendFeatureToggle`) each define their own const with the same string.
- **#1398 replays inbound `Authorization` + `Cookie`** (`forwardedAuthHeaderNames`) and **accepts `exp == 0`** — both confirmed; the appendix's "forwards nothing" call on the stripped header is right.
- Baseline (sticky 6h error cache, no stale-serve), the `context.WithTimeout(context.WithoutCancel(ctx), deadline)` idiom, and the `timeNow` seam all match the code.

### 🔴 Blocking — appendix omits #1400's outbound-header fix
**#1400 currently sets `X-Grafana-Id` only** (`custom_guide_repository.go:285`), with no `Authorization: Bearer <id-token>`. §3 states the only runtime-verified shape is **both** headers. The appendix tells **#1398** to converge on `Bearer <id-token>` + `X-Grafana-Id`, but the **#1400** section says nothing about outbound headers.

Consequences:
1. If both headers are required (as the §3 evidence indicates), **#1400 may not authenticate against the aggregator on a real stack** — and nothing in the appendix flags it.
2. Followed literally, the two PRs **drift** to different outbound header sets — the exact failure the shared `forwardIdentityHeaders` helper in §3 exists to prevent. An alignment doc would itself encode a misalignment.

**Fix:** add to the #1400 appendix section, e.g.:
> - Outbound: add `Authorization: Bearer <id-token>` alongside `X-Grafana-Id` via the shared `forwardIdentityHeaders` helper — today it sends `X-Grafana-Id` only, half the §3 runtime-verified shape.

Both PRs must terminate at the same helper output.

### 🟡 Minor (non-blocking)
1. **Evidence is from a sibling route.** `a60cd778`/`89d6bd5e` are on the guides-import proxy (`feat/external-import-api`) — same aggregator, different kind. Reasonable to extrapolate to `completionrecords`/`interactiveguides`, and the mandated per-PR smoke covers the residual risk, but §3 reads as if verified for *these* proxies. A clause ("verified on the sibling import route; confirm per-proxy via smoke") would make the provenance honest.
2. **Appendix #1398 "flip the test that currently locks acceptance in."** The code accepts `exp==0`, but I couldn't find a test that *asserts* acceptance (the cases test expiry, not exp-absence). If none exists, the instruction is "add a rejection test," not "flip" one.
3. **§7 "front-end's session-latching 503 heuristics"** is stated as fact but I didn't verify it against frontend code. It's only a justification for soft-200 (correct regardless), so not load-bearing — but if it's an assumption, mark it as one.

### Nit
- "Fail closed → refuse cleanly" (§3 inbound) then "soft-200, not 401" (§3/§7) is correct but reads as briefly contradictory; one clause tying "fail closed = serve no data via the capability envelope" removes the double-take.

Net: the pattern is right and well-evidenced — the §3 identity conclusion (`Bearer <id-token>` + `X-Grafana-Id`, not `X-Grafana-Id` alone) is justified by real runtime data. Add the #1400 outbound-header bullet and this is a clean merge.

moxious:
@tomglenn all five points addressed in 7c2d717b (https://github.com/grafana/grafana-pathfinder-app/pull/1401/commits/7c2d717b):

- **Blocking**: #1400 appendix now has the outbound-header bullet (`Bearer <id-token>` + `X-Grafana-Id` via the shared helper)
- §3 evidence marked as sibling-route provenance, confirmed per-proxy via §10 smoke
- "Flip the test" is real, one correction: a test *does* lock `exp==0` acceptance in — the `typed prefix preserved verbatim` case builds its token with no `exp` claim and asserts `wantOK: true`. Doc now names it
- §7 "session-latching" replaced with the verified mechanism: `UNAVAILABLE_STATUSES` in `fetchBackendGuides.ts`/`context.init.ts` treats 503 as not-rolled-out, silent empty, no retry
- Fail-closed/soft-200 nit: §3 now says fail closed constrains *what* is served, not the status code

tomglenn (approved):

Issue #72, opened 2025-07-21
Architecture Refactoring Issue: Simplify Content Fetching & Rendering


Jayclifford345:
# Architecture Refactoring Issue: Simplify Content Fetching & Rendering

## Problem Statement

The current documentation fetching and rendering architecture has become complex and difficult to maintain, presenting several critical issues:

### Current Architecture Problems

1. **Tangled Responsibilities**: `single-docs-fetcher.ts` handles both content fetching AND complex DOM pre-processing
2. **Untestable Code**: Business logic is deeply intertwined with DOM manipulation, making unit testing nearly impossible
3. **Performance Issues**: Extensive DOM manipulation in React effects/hooks to reconstitute semantic content
4. **Development Bottlenecks**: Adding new interactive components (like multistep) requires navigating extremely complex, coupled code
5. **Maintenance Burden**: Changes to interactive behavior require modifications across multiple tightly-coupled files

### Specific Technical Debt

- **content-processing.hook.ts**: 915 lines of complex DOM manipulation in React effects
- **interactive.hook.ts**: 1019 lines mixing interactive logic with DOM processing
- **single-docs-fetcher.ts**: Handles both fetching AND pre-processing (violates single responsibility)
- **docs-fetcher.ts**: Separate but similar logic for learning journeys vs single docs

#### 1. **Unified Content Fetcher**
```typescript
// New unified fetcher approach
interface UnifiedContentFetcher {
  fetchContent(url: string): Promise<RawContent>;
  // No more DOM processing here - just raw content retrieval
}

interface RawContent {
  html: string;
  metadata?: LearningJourneyMetadata;
  type: 'learning-journey' | 'single-doc';
}
```

#### 2. **React-First Content Renderer**
```typescript
// New React component approach
<DocsContentRenderer 
  rawContent={content} 
  onInteractionComplete={handleComplete}
/>

// Instead of dangerouslySetInnerHTML + DOM manipulation
```

#### 3. **Component-Based Interactive Elements**
```typescript
// Replace DOM processing with React components
<InteractiveSection requirements="is-admin" outcome="has-plugin-volkslabs">
  <InteractiveStep targetAction="button" refTarget="Install Plugin">
    Install the Volkslabs plugin
  </InteractiveStep>
</InteractiveSection>
```

#### Phase 1: Unified Content Fetching
- [ ] Combine `docs-fetcher.ts` and `single-docs-fetcher.ts` into unified `content-fetcher.ts`
[...]

PR #1386, opened 2026-07-20
refactor(completion-records): route completions through single recorder boundary

moxious:
## Intent

Implement the completion-recorder boundary — PR 2 of the Learning Records epic (Completion Records RFC https://github.com/grafana/pathfinder-rfcs/pull/6, design-contract invariant 1, §5.4/§7.2/§7.3). Goal: route EVERY terminal guide/journey completion through ONE recorder so the two future write tracks (Track 1 analytics events, Track 2 durable CompletionRecord CRD writes) attach at a single seam and double-counting is structurally impossible. This PR is the boundary refactor + the new journey_completed trigger ONLY; it ships neither write track.

Key decisions/tradeoffs a reviewer won't see from the diff: (1) The recorder is a new Tier-1 module src/completion-records/ exposing recordGuideCompletion/recordJourneyCompletion + an onCompletionRecorded emitter seam with ZERO subscribers, making it deliberately behavior-neutral and BACKEND-AGNOSTIC — no HTTP, no backend client types, no availability-flag checks (captain directive: in-preparation refactoring PR, separately mergeable ahead of the backend). (2) It is keyed on (guideSource, guideId) = (manifest.repository, manifest.id) resolved off the package manifest, never a loader URL — the joint contract shared with the Custom Guide Packages RFC https://github.com/grafana/pathfinder-rfcs/pull/11. (3) markGuideCompleted and markMilestoneDone intentionally KEEP their local-cache/UX duties (badges, streaks, toasts, localStorage/user-storage progress) unchanged; only completion EMISSION moves to the recorder — so this is a refactor with one addition, and behavior parity is a hard requirement proven by tests. (4) I deliberately did NOT re-key or migrate the local progress store (URL/slug-keyed) — that is #11's F13 mechanism; keying only the emitted fact avoids colliding with #11 on the shared call sites and preserves parity. (5) The journey_completed trigger did not exist before; it now fires once when the final milestone crosses the all-milestones-complete threshold inside markMilestoneDone. (6) Exactly-once emission is deduped in the recorder; regression tests cover the double-fire hazards including PR #689's partial-progress duplicate race. (7) Adding a Tier-1 dir required synced edits to TIER_MAP, AGENTS.md, systemPatterns.mdc, and eslint.config.mjs (enforced by architecture.test.ts). Threaded an optional CompletionContext (packageManifest/title) through setJourneyCompletionPercentage and markMilestoneDone to all six call sites; two existing tests that pinned markMilestoneDone's exact args were updated for the added context arg.

[...]

PR #1408, opened 2026-07-23
feat: custom guide packages — private catalogue on GAP + on-behalf-of proxy auth

tomglenn:
## Summary

Ships the **private custom-guide catalogue** — the App Platform analogue of the CDN `repository.json` — on the new **Grafana App Platform (GAP)** group, end to end:

- **Frontend surface** — a Custom Guides section (path/journey cards + loose "orphan" guides) in the context panel, and ingestion of private path/journey packages into My Learning, with milestone resolution and `backend-guide:` launch.
- **GAP repoint** — the plugin's App Platform reads/writes and the catalogue proxy now target `pathfinderbackend.ext.grafana.app/v1alpha1`, gated on the `aggregation.pathfinderbackend-ext-grafana-app.enabled` toggle.
- **Backend on-behalf-of (OBO) auth** — a shared `pkg/plugin/auth` exchanger so the plugin backend can call its own stack's aggregated API **as the calling user**, which unblocks every App Platform proxy route (custom-guide catalogue *and* completion-records).

## What to look at

- **`pkg/plugin/auth` (new) + `app_platform_client.go` + `app_platform_identity.go`** — the OBO auth. The inbound `X-Grafana-Id` is exchanged (authlib, at auth-api) for a short-lived access token minted from the provisioned CAP token (`secureJsonData.accessToken` + `jsonData.stackId`), sent as `X-Access-Token` (aud `grafana`); the ID token is no longer forwarded as a credential. Mirrors grafana-dbo11y-app's production flow. **@konsalex authored this — please review the auth path closely.**
- **`completion_records.go` / `completion_records_client.go`** — updated to the shared minter (the LIST-client signature changed). **This overlaps the completion-records PRs (#1433/#1434) — coordinate merge order of `pkg/plugin/auth` with @moxious.**
- **`custom_guide_repository*.go`** — GAP group + `.app` aggregation toggle + `obo-unavailable` capability reason (clean degrade when unprovisioned).
- **Frontend** — `CustomGuidesSection`, `usePublishedGuides`, `learning-paths/app-platform-paths.ts`, `package-engine/app-platform-resolver.ts`, `useCustomGuideCatalogueOnOpen`.

## Why

Implements the private guide repository (RFC `CUSTOM-GUIDE-PACKAGES.md`) against GAP after the API group migration off the retiring CAP. The proxy initially 401'd because it forwarded the ID token as a credential — an ID token is an identity attestation, not a credential — so nothing on the outbound path accepted it. The fix is real OBO token exchange (above). A temporary frontend-only LIST shim was used to unblock surface testing during the investigation; it has been **dropped** in favour of the backend approach (server-side collation/block-stripping, per-user RBAC, shared auth for all proxy routes).

## How to verify

- **Requires a dev deploy** (backend auth — not Graft-testable). pathfinder-backend GAP (grafana/grafana-pathfinder-backend#72) must be merged + deployed.
- Publish a guide via the block editor; author a path **package** by POSTing an `InteractiveGuide` with `spec.manifest` to the GAP `interactiveguides` collection.
- Confirm the Custom Guides section renders the path card + My Learning ingests it; `…/resources/custom-guide-repository` returns `capability.available: true`.
- Gate: `npm run check` green (typecheck, lint, prettier, `lint:go`, `test:go` incl. `pkg/plugin/auth`, frontend suite).

## Breaking changes

None user-facing. The feature is gated: when the GAP aggregation toggle is off or OBO credentials aren't provisioned, the surface cleanly hides (`obo-unavailable`) and the rest of the plugin loads untouched.

## Reversibility

Fully feature-gated; no data migration. Degrades to hidden on any unavailability. The proxy/auth can be disabled by the aggregation toggle without a code change.

## Merge gates / dependencies

- pathfinder-backend GAP (#72) merged + deployed to the target envs.
- **Prod OBO provisioning** (`stack-state-service`: `secureJsonData.accessToken` + `jsonData.stackId`) confirmed before prod, else prod shows `obo-unavailable`. Verified on dev.
- Coordinate shared `pkg/plugin/auth` landing with completion-records #1433/#1434.


tomglenn:
## ⏸️ Parked — blocked on the Grafana App Platform (GAP) migration

This work is complete and verified against the **current** backend, but it can't ship yet: we're still on the **old Cloud App Platform**, and the `InteractiveGuide` CRD/API must migrate to the **new Grafana App Platform** (work in `pathfinder-backend` + `deployment_tools`) before the catalogue proxy can serve on a real stack.

### State of this branch (durable — just force-pushed)
- Rebased onto current `main`. The branch's own stale backend proxy was dropped in favour of the merged one (#1400); the stale client-side completion-keying commit was dropped in favour of main's recorder (#1386). Completion tracking for private guides is deferred — see #1423.
- Frontend catalogue client aligned to the proxy's `{capability, guides, asOf}` envelope (namespace derived server-side, not a query param).
- Backend proxy: the single `backend-unavailable` reason split into granular capability reasons + `upstream-<status>`, so the failure mode is diagnosable from the client without backend log access.
- Verified: typecheck, `go build`, `go vet`, `go test ./pkg/...`, and the catalogue frontend suites pass. (One repo-wide governance test fails locally only — a Node-22-vs-required-24 artifact; passes on CI.)

### Confirmed blocker (why it doesn't work on dev today)
A guide publishes and lists fine via a direct browser call to the old-GAP `interactiveguides` API, but the proxy returns `capability.available=false`. Expected — we're on old GAP; this is not a proxy bug to patch.

### Rework needed once GAP migration lands
- Point the proxy's outbound LIST at the new GAP API group/version/resource.
- Re-verify the proxy's credential (forwarded ID token) + namespace model against new GAP.
- Re-run the catalogue end-to-end, then pick up completion (#1423).

Leaving as **draft** until then.

src/components/docs-panel/hooks/useCustomGuideCatalogueOnOpen.ts
```diff
@@ -0,0 +1,30 @@
...
+  const hasFetchedRef = React.useRef(false);
+
+  React.useEffect(() => {
+    if (hasFetchedRef.current) {
+      return;
+    }
+    hasFetchedRef.current = true;
+    // The client resolves to [] on every failure path, so this catch only
+    // guards against a future contract change becoming an unhandled rejection.
+    fetchCustomGuideRepository(config.namespace).catch(() => {});
```
Jayclifford345, line 28: **Three identical catalogue fetches per panel open.**

`fetchCustomGuideRepository` now has three callers that all mount when the panel opens: this hook (wired in `docs-panel.tsx`), `usePublishedGuides` (context-panel), and the new mount effect inside `useLearningPaths` — which `UserProfileBar` pulls in via `useNextLearningAction`, and `UserProfileBar` renders at `context-panel.tsx:1124`. `useGuideCompletion()` mounts a fourth `useLearningPaths` instance whenever a guide renders.

The client has no dedupe or TTL, and the proxy deliberately has neither a cross-request cache nor single-flight (`custom_guide_repository.go:41-62`, and that reasoning is sound — a namespace-global blob shouldn't be shared across identities). So each of those callers drives a full paginated upstream drain (up to 5 × 500) plus its own `X-Access-Token` round trip, and they fire concurrently on open.

A single-flight + short-TTL wrapper inside `fetchCustomGuideRepository` collapses all of it without touching a single call site, and it doesn't cost this hook its purpose: the probe still guarantees a deterministic hit on panel open, it just rides the shared in-flight request instead of adding a third one.

src/package-engine/composite-resolver.ts
```diff
@@ -36,6 +43,21 @@ export class CompositePackageResolver implements PackageResolver {
...
+
+    // App Platform is a mutable repository (§6.8) — a successful resolution
+    // must never be served stale on a later call. Evict right after it
+    // settles: truly concurrent callers still share this in-flight promise
+    // (dedup preserved), but the next call re-fetches fresh. The `.catch()`
+    // only silences this derived promise — a thrown/rejected `promise` still
+    // propagates normally to whoever awaits the `resolve()` call itself.
+    promise
+      .then((result) => {
+        if (result.ok && UNCACHEABLE_REPOSITORIES.has(result.repository)) {
```
Jayclifford345: **Failed app-platform resolutions are memoized for the session, which defeats the re-resolve affordance this PR adds.**

The eviction is gated on `result.ok`, and `PackageResolutionFailure` (`package.types.ts:233`) has no `repository` field — so no failure can ever match `UNCACHEABLE_REPOSITORIES`. Negative results stay in `this.cache` for the resolver's whole lifetime, which is the module singleton `setPackageResolver(createCompositeResolver(pluginConfig))` installs once at docs-panel mount.

Concretely: `fe-alerting-path` lists `fe-alerting-02`, not published yet. A user expands the path, `fe-alerting-02` resolves `not-found`, it renders locked — all correct. The author publishes it. The user collapses and re-expands, which `CustomGuidesSection.toggleMembers` deliberately clears state for so a re-expand "re-resolves fresh (paths are mutable, §6.8)" — and `resolve('fe-alerting-02', { loadContent: 'metadata-only' })` hands back the cached failure. The member stays locked until the panel remounts. My Learning ingestion inherits the same blind spot.

Evicting on `!result.ok` as well is the cheap fix (the in-flight dedup you're preserving still works). If you'd rather keep negative caching for the static tiers, failures need to carry the repository that produced them.

pkg/plugin/auth/auth.go
```diff
@@ -0,0 +1,93 @@
...
+// Mint exchanges the CAP token for a short-lived access token scoped to the
+// user identified by idToken. The audience is the stack's own front door, which
+// is where the proxy routes send it; the aggregator re-signs from there.
+func (e *Exchanger) Mint(ctx context.Context, idToken string) (string, error) {
+	if idToken == "" {
+		return "", errors.New("cannot mint an access token without a caller id token")
+	}
+	ttl := tokenTTLSeconds
+	resp, err := e.client.Exchange(ctx, authn.TokenExchangeRequest{
+		Namespace:    "stacks-" + e.stackID,
```
Jayclifford345: **The exchange namespace and the LIST namespace come from two different sources, with nothing asserting they agree.**

`Mint` signs for `stacks-<jsonData.stackId>`, baked in once at `NewApp`. Every proxy route targets `backend.PluginConfigFromContext(r.Context()).Namespace` instead (`custom_guide_repository.go:225`, same in `resolveCompletionBackend`). Two independent sources of truth for the same value.

If stack-state-service ever writes a stale or wrong `stackId` — restored or cloned stack, org id instead of stack id — every request mints a token valid for one namespace and LISTs another. The aggregator 403s, `handleCustomGuideRepository` maps that to `upstream-403`, and the surface hides permanently for every caller on the stack. From the capability envelope that's indistinguishable from "this caller lacks list permission", which is precisely the ambiguity the granular reasons were introduced to kill. Backend logs would show the 403 and no reason to suspect the namespace.

Since the request namespace is already resolved on the hot path, passing it into `Mint` and dropping `stackID` from the `Exchanger` removes the divergence entirely. Failing that, compare the two and emit a distinct reason on mismatch — that's a one-line diagnosis instead of a provisioning archaeology dig.

src/components/docs-panel/CustomGuidesSection.tsx
```diff
@@ -10,43 +10,104 @@ import {
...
+        delete next[path.id];
+        return next;
+      });
+      return;
+    }
+
+    setExpandedMembers((prev) => ({ ...prev, [path.id]: 'loading' }));
+    const milestoneIds = path.manifest?.milestones ?? [];
+    const resolved = await resolvePackageMilestones(milestoneIds);
+    setExpandedMembers((prev) => ({ ...prev, [path.id]: resolved }));
```
Jayclifford345: **Collapsing while members are loading re-expands the list.**

`toggleMembers` collapses by deleting the key, but the in-flight resolve has no generation check, so this write puts the entry straight back.

Repro: click "View members", then click it again before it finishes. The list collapses as expected, then pops back open a moment later with the resolved members — the user's second click is silently undone.

Writing only when the entry is still `'loading'` is enough:

```ts
setExpandedMembers((prev) => (prev[path.id] === 'loading' ? { ...prev, [path.id]: resolved } : prev));
```

Worth a test too — `CustomGuidesSection.test.tsx` covers the happy-path drill-in and the locked member, but nothing exercises a toggle during the resolve.

src/package-engine/app-platform-resolver.ts
```diff
@@ -0,0 +1,163 @@
...
+      const response = await lastValueFrom(
+        getBackendSrv().fetch<InteractiveGuideResource>({ url, method: 'GET', showErrorAlert: false })
+      );
+      const resource = response.data;
+
+      if (!resource?.spec) {
+        return failure(packageId, 'not-found', `App platform guide "${packageId}" has no spec`);
+      }
+
+      resolution.manifest = buildManifest(packageId, resource.spec);
```
Jayclifford345, line 182: **Draft guides render as available path members.**

Every catalogue surface gates on published: `usePublishedGuides` filters `status === 'published'`, `fetchAppPlatformLearningPaths` does the same, and the read path this replaces (`fetchBackendGuides(ns, true)`) excluded drafts and explicitly treated a missing status as draft. This resolver ignores `spec.status` entirely, and so does the `backend-guide:` content fetch it hands off to.

So a published path whose manifest lists a member that's still a draft shows that member unlocked in the drill-in, and opening it renders the draft — for every viewer in the namespace, not just its author. Not an authorization problem (namespace RBAC governs both reads equally), but it does contradict the published-only rule applied everywhere else, and drafts are by definition the content least ready to be read.

Either fail resolution for a non-published spec, or if drafts-in-paths is deliberate for author preview, say so here — as written it reads like the status gate was just missed on this path.

src/learning-paths/app-platform-paths.ts
```diff
@@ -0,0 +1,68 @@
...
+      estimatedMinutes: 5,
+      url: `backend-guide:${entry.id}`,
+    };
+  }
+
+  const paths: LearningPath[] = published
+    .filter((entry) => entry.manifest?.type === 'path' || entry.manifest?.type === 'journey')
+    .map((entry) => ({
+      id: entry.id,
+      title: entry.manifest?.description || entry.title || entry.id,
```
Jayclifford345: **`manifest.description` beating `entry.title` for the card heading.**

Same chain in `CustomGuidesSection.pathTitle` (line 39). For guides where the resolver synthesizes the manifest it's right by construction — `buildManifest` maps `spec.title` into `description`, so description *is* the human label.

For a path published with both fields genuinely populated it inverts: title "Alerting enablement", description "Learn how to build alert rules, contact points, and notification policies." renders the whole sentence as the `<h3>` on both the My Learning card and the sidebar card, with the same text repeated in `description` on the line below.

Preferring `entry.title` and falling back to `manifest.description` gets both cases right, since the synthesized-manifest case has no competing title. Is there a case where you know the title is the worse label?

src/lib/custom-guide-repository-client.ts
```diff
@@ -0,0 +1,81 @@
...
+export interface CustomGuideRepositoryEntry {
+  id: string;
+  title?: string;
+  status?: string;
+  manifest?: CustomGuideManifest;
+}
+
+/**
+ * Availability signal the catalogue surfaces gate on. `available` is false with
+ * a machine `reason` (`identity-unavailable` / `backend-unavailable`) when the
```
Jayclifford345: Nit: this comment is the frontend-side record of the capability contract, and it lists the pre-split token set. As of this PR the backend can return `identity-unavailable`, `grafana-config-unavailable`, `feature-toggle-disabled`, `namespace-unavailable`, `app-url-unavailable`, `obo-unavailable`, `backend-unavailable`, or `upstream-<status>`.

`obo-unavailable` in particular is the one a reader debugging an empty surface most needs to know exists.

src/docs-retrieval/content-fetcher/package-content.ts
```diff
@@ -170,6 +178,37 @@ function getManifestMilestoneIds(manifest?: Record<string, unknown>): string[] {
...
+export function ensureNonEmptyCoverContent(jsonContent: string): string {
+  try {
+    const parsed = JSON.parse(jsonContent) as { blocks?: unknown[]; [key: string]: unknown };
+    if (Array.isArray(parsed.blocks) && parsed.blocks.length === 0) {
+      return JSON.stringify({
+        ...parsed,
+        blocks: [
+          {
+            type: 'markdown',
+            content: 'Cover content is missing for this path. Check back soon, or contact whoever published it.',
```
Jayclifford345, line 204: Nit: this is the first user-facing prose originating in `docs-retrieval` — everything else in this tier transforms authored content, and every user-visible string in the plugin goes through `t()` up in `components/` (i18n imports appear in 15 component files and `module.tsx`, nowhere else).

So this sentence is the one piece of copy a user with a broken path sees, and it's the one piece that will never translate. Threading the string in from the component layer keeps the tier boundary and the locale coverage; alternatively it's a small, deliberate gap — just worth being deliberate about it rather than inheriting it.

Jayclifford345 (commented):
## Review — routed via `docs/design/CONCERNS.md`

**Classification:** `mixed` (product-runtime + contracts-and-schemas). **Concerns activated:** `security`, `go-backend`, `correctness-and-reliability`, `package-engine`, `docs-retrieval-and-rendering`, `reversibility-and-one-way-door`, `performance-and-bundle`, `testing-and-verification`, `cross-cutting-architecture`.

**Gates run on `pr-1408`:** `go build ./...` clean; `go test ./pkg/...` green (`pkg/plugin` and the new `pkg/plugin/auth`). Frontend suite not re-run here — taking CI + your `npm run check` at face value.

### The auth path checks out

I went into the pinned `authlib` (`v0.0.0-20260707082819`) to check the claim that the minted token alone conveys the caller. `AccessTokenClaims.IsOnBehalfOfUser()` is `getIdentityActor() != nil`, which walks the `act` chain looking for a `user`/`service-account` identity type — so the user does ride in the actor chain of an OBO token, `X-Access-Token` on its own is a complete credential, and dropping `X-Grafana-Id` is correct rather than a downgrade to service-identity. `TestAppPlatformListClient_OutboundIdentityAndPagination` asserting that *neither* `Authorization` nor `X-Grafana-Id` goes out is a good contract pin, and `TestAppPlatformListClient_MintFailureAbortsRequest` closes the "fall back to an unauthenticated call" hole. No security findings.

Degradation is also genuinely clean: `auth.New` returning `(nil, nil)` when unprovisioned, `obo-unavailable` as a structural (not transient) reason, and the mint failure staying status-less so the classifier retries — those three fit together correctly.

### Findings

Eight inline, roughly in severity order. The three I'd actually want resolved before this unparks:

1. **Three catalogue fetches per panel open** — the new probe hook is the third unconditional caller, and the proxy deliberately has no cache and no single-flight.
2. **Failed app-platform resolutions are cached for the session** — `PackageResolutionFailure` carries no `repository`, so the new eviction can never fire on a failure, which silently defeats the collapse-and-re-expand refresh this PR builds.
3. **The OBO namespace comes from `jsonData.stackId`, the LIST namespace from the plugin context** — nothing checks they agree, and a mismatch presents as a permanent `upstream-403`.

Then: a collapse-during-load race in the member drill-in, draft guides surfacing as available path members, `description` beating `title` for card headings, and two nits (a stale reason list in the client's doc comment, untranslated placeholder copy).

### Reversibility

Reversible. Nothing here changes a persisted shape: the `isLocked` addition is optional and additive, App Platform paths are appended to `paths` rather than merged into stored progress, and `badgeId: ''` is inert because `getBadgesToAward` iterates `BADGES` against `getPathsData().paths` — the bundled set — so a synthesized path can never award or pollute a badge record. Milestone renumbering (`sequenceNumber++` → `i + 1`) is a real semantics change for counters, but it only affects paths that have locked members, which can only exist on the new App Platform repository.

One thing to watch on the reason-token split: `upstream-<status>` moves the capability `reason` from a closed set to an open-ish one. Fine as long as it never becomes an analytics attribute — it would be unbounded there.

### Coverage note

The new tests are well aimed (mint failure, unprovisioned stack, locked-member traversal, orphan fallback, flat-list preservation). The gaps that map to findings above: no test collapses the drill-in while a resolve is in flight, and nothing pins that a re-expand actually re-resolves after a previously-failed member publishes.

Doc drift: the `pkg/plugin/auth` subsystem, the second aggregation toggle, and the `stackId`/`accessToken` provisioning contract are all new load-bearing wiring that `.cursor/rules/systemPatterns.mdc` and `AGENTS.md` don't mention yet — `BACKEND_PROXY_PATTERN.md` and `CODA.md` carry it, which may be enough, your call.

Jayclifford345 (changes requested):
I saldy think we are blocked on this one due to the settings bug. I connot work at why its happening at the moment but my assumpution is the old method of saving settings in a plugin is causing some form or recreation in the backend which throws out the other settings SSS creates. When this happens we loose the token thats needed to speak to the private guide service. Here is an agent led invesitgation below, its not great but might help with a lead: 

## Bug Report: Private guides vanish after saving plugin settings

### Summary
Private guides disappear from the Custom Guides section after saving plugin settings and **do not automatically recover**. Users must manually navigate away and return (or perform some other action) to eventually see guides again.

### Root Cause
After plugin settings are saved, Grafana Cloud's OBO (On-Behalf-Of) token becomes unavailable. The backend returns `capability.available: false` with `reason: "obo-unavailable"`.

### Investigation Findings

1. **Frontend is NOT the cause**: The frontend does not send `secureJsonData` when saving basic settings - the provisioned `accessToken` should be preserved by Grafana.

2. **Backend receives empty credentials**: When `NewApp()` is called after settings save, `appSettings.DecryptedSecureJSONData["accessToken"]` appears to be empty, causing `oboExchanger` to be `nil`.

3. **No automatic recovery observed**: The OBO token remains unavailable indefinitely after the settings save. Recovery only occurs through some external trigger (navigation, time, or re-provisioning).

### Affected Component
This appears to be a Grafana Cloud platform issue where plugin instance recreation loses access to provisioned credentials (specifically the `accessToken` used for OBO authentication).

### Recommendation
Investigate at the Grafana Cloud / stack-state-service level to understand why provisioned credentials are lost during plugin settings updates.

Jayclifford345:
Fix was made to the above issue #1514, merged into main and will redeploy

Jayclifford345 (dismissed):

moxious (changes requested):
## Review — routed via `docs/design/CONCERNS.md`

**Classification:** `mixed` (product-runtime + contracts-and-schemas). **Concerns activated:** `security` (+ the standalone deep-security lens), `correctness-and-reliability`, `testing-and-verification`, `reversibility-and-one-way-door`, `cross-cutting-architecture`, `go-backend`, `package-engine`, `docs-retrieval-and-rendering`. **Coverage confidence:** high. Every medium-or-higher finding below went through an adversarial verification pass; three claims died there and are recorded at the bottom rather than dropped silently.

### Gates

`go build ./...`, `go vet ./pkg/...`, `go test ./pkg/... -count=1` and `go test -race ./pkg/plugin/...` all pass on `13e7ab8f`. Frontend suite green in a clean worktree: 408 suites, 6431 passed. CI 34/34 including E2E on six Grafana versions, semgrep, TruffleHog, zizmor. `govulncheck ./pkg/...`: 0 called vulnerabilities. Branch is one commit behind main (a CI action bump).

### The settings blocker is genuinely fixed

Root cause was `getConfigWithDefaults()` returning only `DocsPluginConfig` fields while Grafana's plugin-settings API full-overwrites `jsonData` — so any save dropped the provisioned `stackId`, `auth.New` failed, and the surface reported `obo-unavailable`. #1514 (`ac57719c`) is an ancestor of this branch. I audited all eight `updatePluginSettings` call sites: five spread `...(jsonData || {})` first, and the three in `dev-mode.ts` GET the live settings before writing. No full-overwrite path remains. `secureJsonData.accessToken` was never at risk — Grafana merges secure keys individually (`pluginsettings/service.go:207-209`) versus the full replace of `JsonData` at `:223`.

### The auth path checks out

Two independent reviewers went into the pinned `authlib`, and the one alarming security claim got three skeptics. The OBO design is correct: the minted token carries the caller in its `act` chain, `GetTokenPermissions()` returns `[]` for a user identity so authorization falls to `delegatedPermissions` — the OBO token is **narrower** than the CAP, not broader. Dropping `X-Grafana-Id` outbound is right, not a downgrade. Audience is a compile-time literal. Mint failure aborts rather than falling back unauthenticated, and a test pins it. No credential is logged, returned, or reflected in an error body. Bounds are real: 8 MB/page, 500/page, 50,000 entries with truncation logging, 15 s/page, 60 s/drain. @konsalex — nice work; the findings below are almost all on the frontend ingestion side.

---

## 1. [blocking] High — private path progress is written under one key and read under another

_Concern: `reversibility-and-one-way-door`. Confirmed by every verifier, one of them by running it._

`app-platform-paths.ts:66` puts **bare ids** in `LearningPath.guides`. But `app-platform-resolver.ts:112` sets `contentUrl = 'backend-guide:' + id`, that becomes `Milestone.url`, and `getMilestoneSlug` (`lib/learning-journey-url.ts:18-21`) has no `/` to split on so it returns the whole `backend-guide:fe-alerting-01`. That slug reaches `markGuideCompleted` → `badge-coordinator.ts:51` → `completedGuides`. The read side is `completedGuides.includes(bareId)` (`learning-paths.hook.ts:75, 321`). Nothing strips the prefix anywhere — repo-wide grep confirms, and `user-storage.ts` is untouched by this diff.

So a user completes every member of a private path and the My Learning ring stays at 0% forever. The comment at `app-platform-paths.ts:13-17` asserts these bare ids are "the same identifier the completion-keying fix (Appendix A F13) now records completions under" — that fix is not in this diff.

A verifier reproduced this against the real code rather than reasoning about it: calling `markMilestoneDone` in the toolbar's exact shape for **both** members of a 2-member private path leaves `learningProgressStorage` holding `["backend-guide:fe-alerting-01","backend-guide:fe-alerting-02"]`, and rendering `MyLearningTab` with the real `useLearningPaths` over that catalogue still shows **`0%` / `0/2 guides`**. The write path is live — all three call sites pass `getMilestoneSlug(activeTab.currentUrl)` (`link-handler.hook.ts:532`, `DocsPanelContentArea.tsx:434`, `useLastMilestoneAutoComplete.ts:54`) and `currentUrl` for a private member is `backend-guide:<id>`. No other writer records the bare id either: `setJourneyCompletionPercentage` only calls `markGuideCompleted` when the key `startsWith('bundled:')` (`learning-journey-helpers.ts:389`).

**Why this is the one that can't wait:** from the moment it ships, `learningProgressStorage.completedGuides` accretes rows under a key shape the PR's own design says is wrong, and they inflate `totalGuidesCompleted` (`MyLearningTab.tsx:457`). A code revert stops the drift but leaves the rows; a later keying fix needs a read-time migration. **`partially_reversible`.**

**Suggested fix:** normalize in `getMilestoneSlug`/`markMilestoneDone` (strip the scheme before it reaches `completedGuides`), or land the migration the comment already claims exists. Add a test that completes a synthesized member and asserts `getPathProgress` moves off 0. Fix the comment either way.

## 2. [blocking] High — locked milestone placeholders were taught to the traversal helpers and to nobody else

_Concern: `docs-retrieval-and-rendering`. Four surfaces, reported independently by three reviewers and confirmed by three verifiers, one empirically._

The PR changed `resolvePackageMilestones` from _dropping_ unresolvable milestones to _keeping_ them as `{url: '', isLocked: true}`. `getNextMilestoneUrl`/`getPreviousMilestoneUrl` learned about `isLocked`. Four other consumers of the same array did not.

**(a) The cover page's primary CTA is silently dead.** `addReadyToBeginButton` (`learning-journey-helpers.ts:214`) takes `milestones.find(m => m.number === 1)` with no lock check and emits `data-milestone-url=""`. The click handler's first branch fails on the empty string, its fallback reads `milestones[0].url` — also `''` — and the `logger.warn` sits in the unreached third branch. A verifier mounted the real `useLinkClickHandler` over the real generated cover HTML and clicked it: **`loadTab` 0 calls, `openDocsPage` 0 calls, zero `console.warn`.** A prominent, enabled "Ready to Begin" that does nothing and leaves no trace. Reachable: the MCP flow publishes one resource at a time and `validate-package.ts:238` only checks that `milestones` is non-empty. This is the scenario the PR cites as its own motivation.

**(b) Journey completion becomes unreachable.** `totalMilestones` now counts locked entries (`package-content.ts:279`), while the gate is `completed.size >= totalMilestones` (`:505`) — which also guards `awardBadge` and `recordJourneyCompletion`. Demonstrated as a clean before/after: completing both reachable members of a 3-member path with one locked fires `recordJourneyCompletion` **0 times**; the identical completions against the pre-PR denominator of 2 fire it **once**. `isLastMilestone` (`current === total`) also never becomes true, so `useLastMilestoneAutoComplete` is dead on that path too. This one reaches **existing CDN content**: the verifier produced the locked placeholder from `online-cdn-resolver`'s own `network-error` shape on a transient index failure, where the entry used to be dropped. Bundled paths are safe — `build-graph.ts:140-146` fails CI on a dangling `milestones` entry.

**(c) The context-panel milestone lists render locked entries as enabled buttons** (`context-panel.tsx:563-591`, `:833-878`) — no `disabled`, no lock icon, no guard in `onClick`. Rendering the real `RecommendationsSection`, the locked row comes out as a `<button>` with **`disabled === false`**, labelled with the raw package id; clicking it calls `openDocsPage('', title, packageInfo)`, and feeding those exact args to the real `loadDocsTabContentResult` calls `fetchPackageById('the-path', manifest)` — so the user lands on the path **cover** instead of the module they clicked. Both lists behave identically. Same transient-CDN route as (b), so not confined to the new surface.

**(d)** The bottom "Next →" renders whenever `currentMilestone !== totalMilestones` and is inert — a dead control rather than a wrong action. Lowest of the four.

**Suggested fix:** pick the first _unlocked_ milestone in `addReadyToBeginButton`; compute the completion threshold from unlocked entries (or state explicitly that a path can't complete until fully published, and surface that on the cover); apply `CustomGuidesSection`'s existing treatment — `disabled` + lock icon + `onClick` guard — to the two context-panel lists. Extracting one shared milestone-row component would stop the next surface drifting.

---

## 3. Medium — unpublished members leak into My Learning

_Concern: `correctness-and-reliability`. The resolver got the `spec.status` gate; `app-platform-paths.ts` did not._

`guides: entry.manifest?.milestones ?? []` copies every authored member id while `guideMetadata` is built only from published entries. A draft member falls through `resolveGuideMetadata` to `{title: guideId, estimatedMinutes: 5}` with no url, and `MyLearningTab.tsx:399` computes `bundled:<draftId>`.

A verifier ran the real `fetchAppPlatformLearningPaths` against a fixture where a published path lists a draft member: `guides = ['fe-alerting-01','fe-alerting-02-draft']` while `guideMetadata` holds only `['fe-alerting-path','fe-alerting-01']` — no upstream filter exists. Rendering `MyLearningTab` over that: the card reads **`0% … 0/2 guides`** with a row titled by the raw id, and clicking Start calls `prepareGuideLaunch('bundled:fe-alerting-00-draft', …)`.

Verifiers did correct the severity down from what three reviewers reported: member rows are non-interactive `<div>`s (`GuideList.tsx:26-39`), so the raw-id title is cosmetic; the bad launch comes from the card's Start/Continue button targeting the first incomplete guide — always `guides[0]` given finding 1 — and it surfaces as a "Could not open the guide" alert rather than silently. Denominator inflation is real but moot while the numerator is stuck at 0.

Meanwhile `CustomGuidesSection` gates correctly (`disabled={member.isLocked}` + lock + "(not yet available)"), so the two surfaces disagree by construction.

**Suggested fix, one line:** `guides: (entry.manifest?.milestones ?? []).filter((id) => id in guideMetadata)`.

## 4. Medium — negative caching is silently repealed for every resolver tier

_Concern: `package-engine`. Empirically reproduced._

App Platform is always appended last (`composite-resolver.ts:111`), `resolveUncached` returns `lastFailure`, and `app-platform-resolver.ts:63` tags **every** failure with `repository: 'app-platform'` — including the cheap GAP-off and no-namespace declines that never issue a request. So every all-miss is evicted. A skeptic built a real `createCompositeResolver` and resolved a missing id three times: `bundled:3 cdn:3 appPlatform:3`, cache size **0**. The comment at `:50-52` ("static-tier failures don't carry `repository`, so their negative caching is preserved") is false as written. Both existing failure-cache tests build one-resolver composites, which is why nothing catches it.

Cost is smaller than first assessed — the CDN index is process-memoized, bundled is an `Object.hasOwn` — but on recommender-enabled stacks it's one uncached `GET /api/v1/packages/{id}` per re-resolve, and `resolvePackageMilestones` re-runs on every path-card expand and every recommendation refresh. Sharpest case: a **GAP-off** stack pays the full cost for zero benefit.

**Suggested fix:** decide eviction from the tier that produced the decision rather than from `lastFailure`, or stop tagging the short-circuit declines. Add one test built from `createCompositeResolver`.

## 5. [security] Medium — mint namespace and LIST namespace still diverge, with nothing asserting agreement

_Concerns: `security` + `go-backend`. Unchanged since the previous review; Cursor Bugbot re-raised it against this head._

`auth.go:84` signs for `"stacks-" + jsonData.stackId`; `custom_guide_repository.go:224` and `completion_records.go:560` LIST against `PluginConfigFromContext().Namespace`. `grep '"stacks-"'` returns exactly one hit — no comparison, no log, no distinct reason token.

Two angles converge on one fix. **Availability:** a stale or wrong `stackId` mints for one namespace and LISTs another, the aggregator 403s, and the surface hides permanently for everyone on the stack — indistinguishable from "this caller lacks list permission", which is the ambiguity the granular reasons exist to kill. **Trust boundary:** `jsonData` is admin-writable (that mutability is exactly what made #1514 necessary), so the one component of the outbound credential's scope that is _not_ server-derived is admin-influenced input on an otherwise fully server-derived path.

**Suggested fix:** pass `PluginConfigFromContext(ctx).Namespace` into `Mint` and drop `stackID` from the `Exchanger`. That removes the untrusted input, makes divergence structurally impossible, and closes the diagnosability gap in one change. Note the completion-records route also still collapses this to the undifferentiated `backend-unavailable` rather than `upstream-403`.

## 6. Medium — `buildManifest` mislabels app-platform manifests as CDN-sourced (latent durable-key hazard)

_Concern: `package-engine`. Empirically confirmed; consequence not yet live._

`app-platform-resolver.ts:80` forces `id` last but not `repository`. `package.schema.ts:138` declares `repository: z.string().default('interactive-tutorials')`, and `repository` is optional in the persisted CR shape. Two skeptics ran `ManifestJsonObjectSchema.loose().safeParse(...)` on a manifest without it: `repository: "interactive-tutorials"`. That value is the sole input to `guideSource` in `resolveCompletionIdentity` (`completion-identity.ts:47`) — no overwrite in between, no fallback — so a private guide's durable completion record would be keyed to the public CDN repository. The inferred-manifest branch sets it correctly, which makes the asymmetry look accidental.

**Latent today:** `onCompletionRecorded` has zero production subscribers on this head, so nothing durable is mislabelled yet. Flagging it here because it becomes a one-way door on a durable key the moment the Track 2 writer lands — so if this is deferred, the follow-up needs to close before #1433/#1434's consumer side.

Second half, same function: a `safeParse` failure silently returns `type: 'guide'`, so a path the catalogue renders as a path card resolves as a plain guide with no milestone chrome and no log.

Adjacent latent bug spotted while verifying: `learning-journey-helpers.ts:485-488` omits `packageManifest` on the milestone guide-kind emission while `:522` passes it 35 lines later in the same function.

## 7. Medium — mint failures enter the namespace-global negative cache and are replayed to other callers

_Concern: `go-backend`. Empirically reproduced; latent today._

A mint failure is wrapped with plain `fmt.Errorf`, so it carries no HTTP status and `isIdentityScopedUpstreamError` can't match it. `completion_records.go:288` therefore writes it into `completionLastFailure[namespace]`. A skeptic reproduced this against the real handler: caller `user:a` mint-fails → 503; caller `user:b` one second later gets 503 with the **upstream call count still 1** — B was served A's cached error. A genuine 403 control produces 2 calls, correctly unshared. That's the invariant `completion_records.go:203-206` explicitly forbids. Per-caller mint failures are reachable because `parseIDToken` checks only structure and `exp`, so auth-api can reject one caller's subject token.

Bounded: cold cache only (`completionCacheEntries` is never evicted, so a warm namespace serves stale), 30 s, availability-only, and nothing on the frontend calls `/completion-records/my` on this head.

**Suggested fix:** give the mint failure its own sentinel and treat it as identity-scoped at the negative-cache decision.

## 8. Medium — the toggle↔API-group pairing is untested, and this PR removed the suite's ability to notice

_Concern: `testing-and-verification`._

Custom guides moved to `aggregation.pathfinderbackend-ext-grafana-app.enabled`; completion-records still reads the legacy `.com` toggle. **That is not a regression today** — CAP isn't retired, both APIServices coexist, so the group/toggle pairing inside each proxy is self-consistent (thanks for the migration note in #1430, that's what settled it). But this PR changed the _shared_ Go fixture from one toggle to both (`completion_records_test.go:75-80`), so the completion suite can no longer detect which toggle governs it — the regression the fixture used to catch, removed in the same commit that repointed the sibling route.

Two cheap tests close it: a table test deriving the expected toggle from the group version for both pairs (the Go mirror of `interactive-guides-api.ts:14`), and a "only `.app` enabled" case for completions. Also fix `app_platform_client.go:25-27`, whose comment now points at a frontend check that reads the _other_ group.

Two smaller gaps in the same concern:

- The completion-records `obo-unavailable` branch is **0%-covered** while its custom-guide twin is tested (`resolveCompletionBackend` 66.7%).
- **The exchange subject is unasserted** — `stubMinter` discards its `idToken` argument, so nothing pins that the caller's token reaches `Mint`. Subject substitution is the one failure mode an OBO design must not have, and this is a one-line fix: record the argument and assert it in `TestAppPlatformListClient_OutboundIdentityAndPagination`.

## 9. Medium — private paths advertise a fabricated public grafana.com URL

_Concern: `docs-retrieval-and-rendering`._

`package-content.ts:241-242` derives `pathSlug` from any truthy `packageManifest.id` and `:283-285` synthesizes `websiteUrl: https://grafana.com/docs/learning-paths/${pathSlug}/` with no provenance check. `LearningJourneyMilestoneToolbar.tsx:171-173` prefers it, `window.open`s it in a new tab, and reports it to analytics as `content_url`. So a customer's private path gets an "Open" button to a public 404, and a fabricated URL lands in RudderStack.

A verifier ran the real `fetchPackageContent` with the manifest shape `CustomGuidesSection.tsx:37` now threads and got `websiteUrl: "https://grafana.com/docs/learning-paths/fe-acme-onboarding/"`; rendering the real toolbar and clicking "Open" called `window.open(thatUrl, '_blank', 'noopener,noreferrer')` and reported the same fabricated URL as analytics `content_url`. `grep websiteUrl` shows no provenance suppression anywhere.

The synthesis is pre-existing and ungated; the (correct) `id`-threading fix is what newly makes it reachable — at the merge base nothing supplied a private path manifest with an `id`, so such a path opened as a plain `interactive` tab with no journey metadata at all. Gate on `manifest.repository === 'app-platform'`. Worth noting the analytics half is emitted data you can't retract, if that affects where you want to draw the defer line.

---

## Low / nits

- **[suggestion] Rejected promises stay cached forever** — `composite-resolver.ts:57-63` evicts only in `.then()`; `.catch(() => {})` doesn't delete. Latent (all four resolvers return Result values today) but the comment asserts the opposite. `.catch(() => this.cache.delete(cacheKey))`.
- **[question] The new cover-validation gate is unreachable.** It needs `loadContent: true`, which no production caller passes, and `repository.json` has no path/journey packages — so no shipping package is stranded (good news for the one-way-door question), but `cover-validation.ts`'s header and `package-content.ts:186-190` both describe protection that doesn't exist in production. Wire a caller, or delete the gate and correct both comments.
- **[suggestion] Private/bundled ID collision is convention-only** (`fe-` prefix, unenforced, `composite-resolver.ts:96-99`). The card opens the private guide via `backend-guide:`, but every resolver-mediated lookup for the same id serves the public package — split-brain with no warning, and hard to change once private guides exist under colliding ids. At minimum document the rule in `package-authoring.md` where authors will see it.
- **[suggestion] `fetchCustomGuideRepository` swallows everything and discards `capability.reason`**, so `usePublishedGuides`'s `catch` and `error` state are dead code (`error` can now only ever be `'No namespace available'`). This is the exact shape that made this week's incident invisible: `obo-unavailable` presented as "no guides", nothing in the console. A `logger.warn` with the reason before returning `[]` is the whole fix.
- **[suggestion] Debug-only diagnosability on the live route.** A wrong CAP token or unreachable auth-api on `/custom-guide-repository` — the only live mint path on this head — emits `logger.Debug` (`custom_guide_repository.go:166`) and 503s indefinitely, never becoming `capability=false`. (`getCompletionIndex` does log this at Info, so the completions route is fine.)
- **[nit] `resetPath` clears `bundled:` keys for App Platform paths** (`learning-paths.hook.ts:445-461`) and still fires `InteractiveProgressCleared`. Currently unreachable — the Restart button only renders when `isCompleted`, which finding 1 makes impossible — so it goes live the moment finding 1 is fixed. Worth fixing in the same pass.
- **[nit] `tokenExchangeURL`** (`app.go:16-18`) is a mutable package global whose comment says "tests override it"; no test does. And `NewApp`'s error path logs both "setup failed" and "not provisioned", pointing an operator at the wrong cause — guard the Info with `else`.
- **[nit] QC8** — several new comments are multi-paragraph where AGENTS.md asks for one short line: `composite-resolver.ts:45-56` (12 lines for a 7-line block, and it has already drifted from the code per finding 4), `cover-validation.ts:1-16` (15-line header on a 17-line function), `package-content.ts:191-206`. The i18n-gap comment also rests on a false premise — it claims to be "the one user-facing string authored in docs-retrieval", and `learning-journey-helpers.ts` already ships seven (`"Ready to begin?"`, `"← Previous"`, `"Next →"`, `"Step N of N"`, …). The decision to skip `t()` is fine; the justification shouldn't be disprovable with one grep.

---

## Dropped by adversarial verification

Recorded so you can check my work rather than take it on trust.

**A privilege escalation via the OBO token cache — refuted 3 of 3.** The claim: because authlib keys its exchange cache on the subject token's `sub` read with `UnsafeClaimsWithoutVerification`, and the plugin validates the ID token only structurally, a forged token bearing a victim's `sub` could hit a warm cache entry and receive a real token minted for that victim. The mechanism is real — one skeptic built a go-jose v4.1.4 harness and confirmed `jwt.ParseSigned` accepts a garbage _or empty_ signature and yields a cache key identical to a genuinely signed token's. It dies on the precondition: Grafana's `ForwardIDMiddleware` stamps `X-Grafana-Id` from the server-side authenticated requester on every `CallResource`, and `SetHTTPHeader` **replaces** any existing value. Another skeptic showed the equivalent exposure predates this PR anyway, since `/completion-records/*` on main already keys per-user data on the unverified `sub` out of a namespace-global cache.

Two doc residuals survive, no code change: `CODA.md:164-166` says the plugin "is not independently reachable with a client-set `X-Grafana-Id`" as though the header were _stripped_, when core only overwrites it (`ForwardIDMiddleware` never calls `DeleteHTTPHeader`, so a client value survives on a request where the requester's ID token is empty). And this PR newly writes "no hop on the outbound path accepts [the ID token] as authentication", which auth-api's `/v1/sign-access-token` does. The PR also shifts this from fail-closed to fail-open-within-585s **if** that boundary is ever broken — which is a genuine argument for the JWKS item `CODA.md` already lists as future hardening, and worth one sentence there.

**"The credential diagnostic can't fire on the failure it exists to diagnose" — refuted on its load-bearing evidence.** The claim said nothing logs above Debug; `getCompletionIndex` logs `logger.Info("completion index refresh failed", …, "error", err)` at `completion_records.go:292` with the full wrapped mint error. The narrow true remainder is kept as a Low above.

**"Nothing asserts the minted token is on-behalf-of" — refuted 2 of 3.** The code facts are accurate (`auth.go:92` returns `resp.Token` unexamined; nothing in `pkg/` reads `Actor`), but the failure mode needs auth-api to violate its own contract by minting a non-actor token for a request that carries `SubjectToken`. The claimed _direction_ of harm is also unproven: with `id == nil` authlib falls back to the CAP's own `Permissions` (`auth_info.go:157-166`), and an OBO-scoped CAP is provisioned with _delegated_ permissions, so that fallback is plausibly empty → 403 at the front door, i.e. fail closed rather than broader. And the proposed `IsOnBehalfOfUser()` guard is imprecise anyway — it returns true for a service-account actor too. Cheap, sensible hardening as a follow-up against the pattern; not a defect this PR demonstrates.

---

## Tech debt

Non-blocking, for awareness. Top five: a byte-identical 29-line card block rendered twice in `CustomGuidesSection.tsx:243-271` and `:280-308`, emitting the same `data-testid` values so tests can't tell the two lists apart; three overlapping views of one dataset in that component's props (`guides ⊇ paths ∪ orphanGuides`), where `paths.length === 0` makes the whole else-branch redundant with the orphan section; near-duplicate `resolveCustomGuideBackend`/`resolveCompletionBackend` with verbatim-identical doc comments, against `BACKEND_PROXY_PATTERN.md`'s own "one definition each" rule; the dead `tokenExchangeURL` seam; and the capability-reason vocabulary now split across two Go files that depend on each other for it, plus a third hand-maintained copy in a TS doc comment. Suggestive: the `path | journey` predicate is now inlined twelve times with no exported tier-0 helper, and the same empty-cover invariant is enforced twice with opposite outcomes (hard failure on bundled/CDN, friendly placeholder on App Platform).

Test hygiene is genuinely good — every new non-test `src/**` file has a colocated test, and the new tests are real rather than existence-only.

## Doc-drift updates recommended

The PR's own doc edits are good; what's stale is the **routing** layer, which matters because `CONCERNS.md` sends every future Go reviewer to `AGENTS.md` first.

1. **`AGENTS.md` §Backend** still says "thin bridge to Coda… three primary request paths" — covers neither the App Platform proxies nor OBO. Add the fourth path, the `pkg/plugin/auth` seam, and a pointer to `BACKEND_PROXY_PATTERN.md`.
2. **`.cursor/rules/systemPatterns.mdc`** has no file-map row for `auth/auth.go`, `custom_guide_repository.go`, or `custom_guide_repository_client.go`; the `package-engine` entry still describes a three-tier chain; the dependency-edges table is missing the catalogue-client and app-platform-resolver edges.
3. **`.cursor/rules/coda.mdc`** should name the wire keys (`jsonData.stackId`, `secureJsonData.accessToken`) — that's what an agent debugging "custom guides don't load on this stack" actually needs — and `resources.go`'s route list omits `/custom-guide-repository`.
4. **`docs/developer/FEATURE_FLAGS.md`** documents neither aggregation toggle, so an agent reading it concludes Pathfinder has no core toggles. Two now gate load-bearing behaviour and they are **not** interchangeable.
5. **`docs/developer/CONTEXT_INDEX.md`**'s `BACKEND_PROXY_PATTERN.md` row says "identity forwarding" (the model this PR deleted), and its `pkg/plugin/*.go` glob doesn't match `pkg/plugin/auth/auth.go`.

Separately (not this PR's job): `CONCERNS.md`'s `package-engine` entry still says "bundled → recommender → CDN", so a reviewer prompted from that text wouldn't know to check the fourth tier's cache semantics — which is exactly where finding 4 lives.

## Instrumentation

Per `TELEMETRY.md`'s decision rule, the card launches are covered (`OpenResourceClick`, `JumpIntoMilestoneClick`, both with `interaction_location`). One real gap, and it's the operationally expensive one: **the capability degradation ladder has no signal at all.** `custom-guide-repository-client.ts` receives `capability.reason` (`obo-unavailable`, `upstream-<status>`, `feature-toggle-disabled`, `namespace-unavailable`…) and discards it, returning `[]` with no log and no facade event. That is rule 1 — a fallback ladder needs a countable, alertable event, and a log doesn't substitute. It is also the direct reason this week's incident presented as "guides silently vanished" rather than "OBO unavailable on this stack". Rule 2 applies too: the paginated drain has a 60 s budget and no `*_ms` measurement. The locked-milestone degradation is likewise invisible. Keep `reason` out of RudderStack — `upstream-<status>` is open-ended — but it belongs in a typed Faro facade op. No new panel, so no `setFaroViewName` / `reportPathfinderSurface` obligation.

---

## Disposition — requesting changes

The auth work is the strongest part of this PR, and the settings blocker is genuinely resolved. What isn't ready is the My Learning ingestion and the locked-placeholder rollout: two independent holes that mean the headline user journey doesn't work yet. Note that **all of this is invisible to the test suite and CI, which are fully green** — these are untested semantics, not broken code paths.

**Must be fixed in this PR:** findings **1** and **2**. Finding 1 because it's the only item here that isn't cleanly reversible — it starts accreting wrongly-keyed rows in `learningProgressStorage` the moment it ships, and a later keying fix needs a migration. Finding 2 because a path whose first member hasn't published has a dead primary CTA and an unreachable completion, and staggered publication is the feature's whole premise. Both fixes are small. I'd fold in the `resetPath` nit at the same time since finding 1's fix is what makes it live.

**Everything else is yours to triage.** Split the mediums and lows into follow-up issues at your discretion — none of the rest changes a persisted shape or a public contract, so deferring them is a scheduling call, not a risk one. Two riders on that:

- **Finding 6** is latent only because `onCompletionRecorded` has no subscribers yet. If you defer it, the follow-up needs to land before the Track 2 completion writer in #1433/#1434, or it becomes a one-way door on a durable key.
- **Finding 9**'s analytics half emits a fabricated URL into RudderStack, which is downstream data you can't retract — your call whether that clears your defer bar.

Happy to file the follow-ups for whichever you want off your plate — say the word and I'll open them with the evidence above. And the feature being gated on both the aggregation toggle and OBO provisioning (ops/prod still tracked by #1503) is doing real work here: none of this is prod-visible today, which is why only two items are blocking rather than half the list.

<sub>Review method: 11 routed reviewers, then 3 independent verifiers per medium-or-higher finding, each prompted to <em>refute</em> it and to default to refuted when uncertain. Findings ≥2 of 3 killed are in "Dropped" above rather than quietly omitted. Where a verifier could execute a claim it did — against the real modules, in a scratchpad-only Jest config and a copied Go module, leaving both checkouts clean — so the `0% / 0/2 guides`, `0 calls / 0 warns`, `recordJourneyCompletion 0 vs 1`, `disabled === false`, cache-size-0 and `window.open` results above are observed, not inferred. Several severities were revised <strong>down</strong> by that pass; noted inline where it happened.</sub>

tomglenn:
Thanks for the review — the reproductions made the triage easy. Pushed three commits (auth namespace refactor; completion-key + locked-milestone fixes; resolver/content ride-alongs) rebased onto current `custom-guide-packages`.

### Fixed in this PR

- **1 (completion key).** `getMilestoneSlug` now strips the `backend-guide:` scheme at the single choke point feeding `markMilestoneDone`, so private members are written and read under the same bare id. Added a `getMilestoneSlug` unit test and a completion round-trip. Folded in the **`resetPath`** nit (clears both `bundled:` and `backend-guide:` keys) since it goes live with this fix. Corrected the `app-platform-paths.ts` comment.
- **2 (locked milestones).** Ready-to-begin CTA picks the first *unlocked* milestone; the completion gate and `isLastMilestone` derive from an unlocked count (`countUnlockedMilestones`), while `totalMilestones` stays the display count; the two context-panel lists render locked rows disabled with a lock icon, `(not yet available)`, and an onClick guard; the bottom Next hides when there's no unlocked successor. Tests across all four surfaces.
- **3 (draft leak).** `guides: (…​milestones ?? []).filter((id) => id in guideMetadata)`.
- **4 (negative caching).** Split resolver failures into untagged cheap declines (negative-cacheable) vs `repository`-tagged post-attempt failures (evicted); composite resolver now evicts rejected in-flight promises (`.catch(() => this.cache.delete(...))`). Added a `createCompositeResolver`-built test.
- **5 (mint namespace).** Namespace now comes from `PluginConfigFromContext(ctx).Namespace`; `stackID` dropped from `Exchanger`/`New` and from the settings struct. Removes the admin-writable input and makes divergence structurally impossible. This also heals stacks whose `stackId` was wiped before #1514.
- **6 (manifest provenance).** `buildManifest` forces `repository` last; a `safeParse` failure is now logged before falling back to `type: 'guide'`.
- **8 (toggle↔group).** Added a Go table test deriving the expected toggle from the group version for both proxies plus a distinct-toggles assertion; recorded the exchange subject in `stubMinter` and assert the caller's id token + namespace reach `Mint`; fixed the stale `app_platform_client.go` comment.
- **9 (fabricated public URL).** Gated `pathSlug` derivation on `repository !== 'app-platform'`, which suppresses **both** the path-cover and the per-milestone `websiteUrl` synthesis (the #1464 My Learning rework had added the per-milestone one, widening this since the original fix).
- **Nits:** capability `reason` now logged front + back (`Debug`→`Info` on the live route); QC8 i18n comment reworded (dropped the disprovable "only user-facing string" premise).

### Filed as follow-ups

- #1553 — **finding 7** (mint failure enters the namespace-global negative cache). Flagged the ordering constraint: lands before #1433/#1434.
- #1554 — cover-validation unreachable gate.
- #1555 — telemetry: capability ladder + drain timing Faro signal.
- #1556 — doc-drift (AGENTS.md / systemPatterns / coda / FEATURE_FLAGS / CONTEXT_INDEX).
- #1557 — tech-debt roundup (card-block dedup, shared `MilestoneRow`, `resolve*Backend` dedup, capability-reason vocab, private/bundled ID-collision doc, dead `tokenExchangeURL` seam).

### One deliberate non-change

The adjacent `packageManifest` omission at `learning-journey-helpers.ts:485-488` is intentional: the milestone-completion fact keys on the slug, while the journey-completion fact (35 lines down) keys on the manifest — the completion-boundary test pins that split. Passing the manifest into the milestone emission would change its key shape. Left as-is.

Gate is green under Node 24: `go build`/`vet`/`test`/`lint:go`, full frontend suite (409 suites), governance, typecheck, eslint, prettier.

src/package-engine/composite-resolver.ts
```diff
@@ -36,6 +43,21 @@ export class CompositePackageResolver implements PackageResolver {
...
+          this.cache.delete(cacheKey);
+        }
```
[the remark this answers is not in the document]
tomglenn: Addressed: the composite resolver now evicts rejected in-flight promises via `.catch(() => this.cache.delete(cacheKey))`, so a rejection is no longer cached. (Same fix as review finding 4's nit.)

src/package-engine/app-platform-resolver.ts
```diff
@@ -0,0 +1,152 @@
...
+ * (`content?.title ?? manifest?.description ?? id`) would otherwise fall back
+ * to the bare package ID instead of a human-readable title (RFC Appendix A3).
+ */
+function buildManifest(packageId: string, spec: InteractiveGuideResource['spec']): ManifestJson {
+  if (spec?.manifest) {
+    const parsed = ManifestJsonObjectSchema.loose().safeParse({ id: packageId, ...spec.manifest });
+    if (parsed.success) {
+      return parsed.data as ManifestJson;
+    }
+  }
```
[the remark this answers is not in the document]
tomglenn: Addressed: `buildManifest` now forces `id` (and `repository`) last, after spreading the persisted manifest, so a stray `spec.manifest.id` can no longer override the resolved package id.

tomglenn:
Follow-up on the triage above — on a second pass, three of the items I'd filed as follow-ups were really this PR's own footprint (they only exist *because* of this branch's work), so I've rolled them in rather than defer:

- **cover-validation gate** (`a5ab294e`) — the branch-added `nonEmptyCoverBlocksError` gate is unreachable (no production caller passes `loadContent: true`, no bundled path/journey exists), and its comments claimed protection that never ran. Deleted the gate + both call sites; the render-time `ensureNonEmptyCoverContent` placeholder is the real, reachable protection. #1554 closed.
- **capability telemetry** (`9ee4e039`) — added a typed Faro facade op `recordCustomGuideCatalogueUnavailable(reason)` (Faro-only, never RudderStack) so the degradation ladder is countable/alertable, not just a console log. #1555 scoped down to the remaining drain `*_ms` + locked-milestone signal.
- **docs** (`c86efec2`) — FEATURE_FLAGS.md now documents both aggregation toggles (not MTFF, not interchangeable); AGENTS.md §Backend now names the fourth request path + the OBO `auth` seam; package-authoring.md now states the cross-repository ID-uniqueness rule. #1556 scoped to the remaining `.mdc`/CONTEXT_INDEX reference polish.

Still genuinely deferred: **#1553** (finding 7 — the vulnerable per-namespace cache lives only in the completion-records proxy, which keeps no live frontend caller on this branch; the live custom-guide proxy deliberately holds no cross-request cache, so it's unreachable there today — must land before #1433/#1434 wire a `/completion-records/my` caller), and the extraction/dedup tech-debt in #1557.

Gate green under Node 24 (full frontend suite 408 suites, typecheck, governance, eslint, prettier).

src/components/LearningPaths/MyLearningTab.tsx
```diff
@@ -165,10 +165,16 @@ export function MyLearningTab({ onOpenGuide }: MyLearningTabProps) {
...
+      // Static or App Platform guide — resolve via the path-scoped metadata
+      // hook exposes (covers App Platform member guides too, RFC §6.11),
+      // falling back to the static bundled catalogue as before.
+      const resolvedGuideUrl = parentPath ? getGuideUrlForPath(guideId, parentPath.id) : undefined;
+      const staticGuideMetadata = getPathsData().guideMetadata[guideId];
+      const title =
+        (parentPath && getPathGuides(parentPath.id).find((g) => g.id === guideId)?.title) ||
+        staticGuideMetadata?.title ||
+        guideId;
+      const guideUrl = resolvedGuideUrl ?? staticGuideMetadata?.url ?? `bundled:${guideId}`;
```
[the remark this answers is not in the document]
tomglenn: Fixed in 16e5bc4b. My Learning now threads the path manifest through `launch → prepareGuideLaunch` as `packageInfo` (mirroring `CustomGuidesSection`), so an App Platform path member renders inside the path's milestone chrome instead of as a standalone guide. The manifest is carried on the ingested `LearningPath` (app-platform-paths.ts); bundled/CDN paths are unaffected (no manifest → packageInfo undefined). Added regression coverage in MyLearningTab.test.tsx for both cases.

tomglenn:
@moxious ready for re-review when you have a moment — findings 1, 2, 5 and the rest are addressed (or filed with the ordering riders), and the Cursor Bugbot high-severity My Learning launch finding is fixed too. Details in the comments above.

moxious (dismissed):
## Review — second round, routed via `docs/design/CONCERNS.md`

**Approving.** This round had two jobs: the standard review pass over the fix commits (`65b8a579`…`16e5bc4b`), and a comment-resolution audit verifying every previously lodged comment against the code rather than the reply threads. The audit result is genuinely good: **every comment from all 14 reviews — Copilot ×3, Cursor Bugbot ×4, @Jayclifford345 ×8 inline, and my findings 1–6/8/9 plus the lows and nits — is verified resolved in code with real, discriminating tests**, and finding 7's deferral to #1553 still holds on this head (no live `/completion-records/my` caller, no cross-request cache on the custom-guide proxy). Gates re-run locally on `16e5bc4b`: `go build`/`vet`/`test`, `lint:go` (0 issues), and the full changed-subsystem frontend suites — all green. The OBO auth path was re-verified by two independent reviewers plus adversarial skeptics and is clean: one `Mint` call site, namespace structurally unable to diverge from the LIST target, no credential ever logged or reflected, mint failure aborts rather than degrading. What follows is two judgment calls I'm deliberately **not** blocking on, consistent with the defer-to-issues posture we agreed for this PR — plus the follow-up ledger.

**Judgment call 1 — the one unresolved review comment (Bugbot's "Missing repository on private packageInfo") is real, and I'm approving anyway → #1558.** Confirmed empirically: both catalogue launch surfaces (`CustomGuidesSection.tsx:37`, `MyLearningTab.tsx:207-209`) force only `id` onto the threaded manifest, the proxy passes `repository` through as `omitempty`, and our own authoring tooling stamps an explicit `interactive-tutorials` that beats the CRD's server-side default — so the `!== 'app-platform'` gate at `package-content.ts:243-244` fails open, resurrecting the fabricated grafana.com `websiteUrl` (404 + RudderStack `content_url`) and mislabeling the durable completion `guideSource` on exactly the two entry points the feature ships. The reason this isn't a request-changes: the feature is dark outside dev/preview (aggregation toggle + OBO provisioning, #1503), the completion-identity half sits under #1423's explicitly accepted keying gap, and no durable completion consumers exist until #1433/#1434. But it defeats a control this PR itself introduced in `3d59caa9`, and the fix is one stamped field at the catalogue-client boundary — **if you'd rather fold it into this PR than take #1558, I'd mildly prefer that**, since it closes the last open review comment and the `MyLearningTab.test.tsx:267` fixture currently pins the broken shape.

**Judgment call 2 — the finding-2 completion-gate fix is incomplete at 2 of 4 call sites, and I'm approving anyway → #1559.** `DocsPanelContentArea.tsx:440` and `link-handler.hook.ts:537` still pass the locked-inclusive `totalMilestones` into `markMilestoneDone`, and skeptic verification proved the two fixed sites are unreachable on the interactive route (all gated on `!hasInteractiveSteps`) — so a partially-published path completed through interactive members emits zero `journey_completed` facts, and reverting the fixed sites passes the entire 6,443-test suite because the fixtures are all-unlocked. The reason this isn't a request-changes: the user-visible consequences today are fully inside #1423's accepted gap ("path-level completion fails closed and is never recorded"), the App Platform badge is inert (`badgeId: ''`), and the durable fact has no subscribers — this is a latent internal defect of an agreed fix, not broken shipping functionality. It does need to land **before** #1433/#1434 wire consumers, same ordering rider as #1553; #1559 carries the two-line fix, the structural option (derive the denominator inside `markMilestoneDone`), and the discriminating-test requirements.

**Follow-ups filed from this round** (all linked back here): #1558 (repository normalization, above), #1559 (completion gate, above), #1560 (`resetPath` leaves path-level `backend-guide:<pathId>` completion keys — reset half-works, journey trigger can re-fire after reload), #1561 (URL-only resolution returns unconditional success — drafts open by id via `recommends`/restored tabs, `not-found` error class regressed), #1562 (hard 503/network catalogue failures bypass the new Faro op — the "wrong CAP token 503s indefinitely" mode is frontend-invisible), #1563 (CODA.md/BPP security-model wording residuals from round one's dropped-finding notes), #1564 (stale doc references incl. the coda.mdc `StackID` row — the field no longer exists), #1565 (small hardening: prototype-safe `guideMetadata`, rejected-promise eviction timing, stale-slug numerator, progress-store/gate disagreement, `manifest.type` optionality), #1566 (additive test gaps incl. the 0%-covered completion-records `obo-unavailable` branch), plus two tech-debt additions on #1557. Already-lodged items re-verified: #1553 premise intact; #1555/#1556 partially satisfied in-PR and due a trim; #1554 confirmed closed by deletion.

**Nit you may want to fold in before merge:** `gofmt -w pkg/plugin/custom_guide_repository.go` (the only PR-introduced unformatted file — the comment continuation at `:167-169`).

The auth work, the settings-blocker elimination (belt: jsonData preservation from #1514; braces: no jsonData dependency left at all), and the fix-pass discipline — 11 of 12 externally-lodged comments closed with tests that would actually fail on regression — are the strongest parts of this PR. Nice work.

moxious (approved):

tomglenn:
Folded in the two judgment-call items from your approval so nothing carries over:

- **#1558 (repository provenance)** — `2a14664a`: the catalogue client now stamps `repository: 'app-platform'` at the boundary feeding both launch surfaces, so the app-platform gate no longer fails open (no fabricated public websiteUrl, no CDN-mislabeled completion). Closes the last open review comment.
- **#1559 (completion gate)** — `2cbb0355`: the remaining two `markMilestoneDone` sites (DocsPanelContentArea, link-handler) now pass `countUnlockedMilestones`, completing finding 2. Discriminating locked-milestone tests added at both — they fail if the denominator regresses.
- gofmt'd `custom_guide_repository.go`.

Full gate green under Node 24 (Go build/test, 408 frontend suites, typecheck, lint, prettier). #1558/#1559 closed; #1560–#1566 remain as follow-ups. Re-requested your review — sorry for the churn.

moxious:
## Follow-up index

Two new issues filed out of the triage pass over this PR's deferred work, so they're linked here alongside the rest:

- https://github.com/grafana/grafana-pathfinder-app/issues/1567 — journey percentage is never recorded as 100 for `backend-guide:` paths (the only 100% write is `bundled:`-gated at `DocsPanelContentArea.tsx:430-432`). Split out of https://github.com/grafana/grafana-pathfinder-app/issues/1565 item 4 with a corrected diagnosis — locked milestones are not the cause.
- https://github.com/grafana/grafana-pathfinder-app/issues/1568 — verify the inbound `X-Grafana-Id` against JWKS before trusting its namespace. This is the code half of https://github.com/grafana/grafana-pathfinder-app/issues/1563; it was previously only an aspirational note in `CODA.md` with nothing tracking it.

Full set of follow-ups deferred out of this PR, each now carrying a triage comment with verified code basis, severity, and the judgment calls an implementer has to settle:

| Issue                                                         | Area                                                                      | Severity              |
| ------------------------------------------------------------- | ------------------------------------------------------------------------- | --------------------- |
| https://github.com/grafana/grafana-pathfinder-app/issues/1553 | Mint failures poison the namespace-global negative cache                  | medium                |
| https://github.com/grafana/grafana-pathfinder-app/issues/1558 | Force `repository: 'app-platform'` on catalogue manifests (Bugbot B6)     | medium                |
| https://github.com/grafana/grafana-pathfinder-app/issues/1559 | Completion gate locked-inclusive at 2 of 4 call sites                     | medium                |
| https://github.com/grafana/grafana-pathfinder-app/issues/1560 | `resetPath` never clears path-level `backend-guide:<pathId>` keys         | medium                |
| https://github.com/grafana/grafana-pathfinder-app/issues/1561 | Resolver URL-only mode vouches for unverified ids                         | medium                |
| https://github.com/grafana/grafana-pathfinder-app/issues/1568 | JWKS verification of inbound `X-Grafana-Id` (**new**)                     | medium                |
| https://github.com/grafana/grafana-pathfinder-app/issues/1563 | Security-model overstatements in `CODA.md` / `BACKEND_PROXY_PATTERN.md`   | medium                |
| https://github.com/grafana/grafana-pathfinder-app/issues/1556 | Doc-drift: `systemPatterns` / `coda` / `CONTEXT_INDEX` reference layer    | medium                |
| https://github.com/grafana/grafana-pathfinder-app/issues/1555 | Drain `*_ms` timing + locked-milestone Faro signal                        | medium                |
| https://github.com/grafana/grafana-pathfinder-app/issues/1562 | Catalogue hard failures (503/network) emit no frontend telemetry          | low-med               |
| https://github.com/grafana/grafana-pathfinder-app/issues/1564 | Stale doc references (`coda.mdc` StackID row, toggle attribution, §8)     | low — **ready as-is** |
| https://github.com/grafana/grafana-pathfinder-app/issues/1565 | Correctness/hardening follow-ups (items 1, 2, 5 after the split)          | low                   |
| https://github.com/grafana/grafana-pathfinder-app/issues/1566 | Test-coverage gaps (additive tests only)                                  | low                   |
| https://github.com/grafana/grafana-pathfinder-app/issues/1567 | Journey percentage never reaches 100 for `backend-guide:` paths (**new**) | low                   |
| https://github.com/grafana/grafana-pathfinder-app/issues/1557 | Tech-debt roundup (7 items after 2 strikes; 6-PR split plan in-issue)     | low                   |

Three things worth knowing before anyone picks one of these up:

1. **All of it is gated on this PR merging.** Every file these issues cite — `app-platform-resolver.ts`, `custom-guide-repository-client.ts`, `app-platform-paths.ts`, `pkg/plugin/auth/`, the `isLocked` field — is absent from `main`. Line numbers in every issue are current at `16e5bc4b` and will drift.
2. **Nothing here is user-visible yet.** `onCompletionRecorded` has zero production subscribers on `main` or on this branch, and the App Platform badge is inert. That's why the two originally-High findings (https://github.com/grafana/grafana-pathfinder-app/issues/1558, https://github.com/grafana/grafana-pathfinder-app/issues/1559) triaged to medium — but it's also why they must land before the completion-records consumers in https://github.com/grafana/grafana-pathfinder-app/issues/1433 / https://github.com/grafana/grafana-pathfinder-app/issues/1434 attach.
3. **Two restructurings were applied during triage.** https://github.com/grafana/grafana-pathfinder-app/issues/1565 item 3 folds into https://github.com/grafana/grafana-pathfinder-app/issues/1559 (it needs the identical `markMilestoneDone` signature change, so doing them separately means doing that refactor twice), and item 4 became https://github.com/grafana/grafana-pathfinder-app/issues/1567. https://github.com/grafana/grafana-pathfinder-app/issues/1557 also lost two items — the ID-collision doc landed in-PR via `c86efec2`, and the completion-marking fragment dedup is subsumed by https://github.com/grafana/grafana-pathfinder-app/issues/1559.

Related pre-existing dependencies: https://github.com/grafana/grafana-pathfinder-app/issues/1503 (ops OBO provisioning) and https://github.com/grafana/grafana-pathfinder-app/issues/1423 (private-guide completion umbrella).

Issue #1411, opened 2026-07-23
Epic: Completion Records — durable, queryable guide/journey completions


moxious:
## What

Durable, queryable completion records for Pathfinder guides and learning journeys. The badge/completion product already ships browser-storage-backed (localStorage + plugin user storage); this epic makes completions **durable, queryable, and auditable**, and feeds them to the recommender as per-request context.

**Design source of truth:** [Completion Records RFC](https://github.com/grafana/pathfinder-rfcs/blob/main/rfc/COMPLETION_RECORDS.md) (grafana/pathfinder-rfcs#6). All architectural decisions live there; this issue is the execution tracker. The RFC was reconciled to the landed implementation in grafana/pathfinder-rfcs#17 (merged 2026-08-19).

Architecture in one line: `grafana-pathfinder-backend` provides the `CompletionRecord` CRD (thin CRD-only App Platform store); all per-user logic (identity stamping, collation, caching, capability signal) lives in this repo's plugin backend as a proxy in front of the generic App Platform CRUD API; the front end records completions through a single recorder boundary and reads/writes only through the proxy.

### Landed

- [x] RFC — grafana/pathfinder-rfcs#6 (merged 2026-07-21)
- [x] `CompletionRecord` CRD schema — grafana/grafana-pathfinder-backend#65 (merged 2026-07-21)
- [x] Completion-recorder boundary (single emission seam, manifest-keyed identity, `journey_completed` trigger) — #1386 (merged)
- [x] Plugin-backend "my completions" read proxy + capability route (reference implementation of the shared App Platform proxy pattern) — #1398 (merged)
- [x] App Platform (GAP) endpoints and on-behalf-of auth reached production (2026-08-18), clearing the deployment gates this epic was blocked on
- [x] Track 2 durable write hook — split from #1415 (closed, replaced by the two focused PRs below):
  - [x] Backend write proxy: `POST /completion-records`, forwarded-caller identity, per-user rate limit, cache invalidation — #1433 (merged 2026-08-19)
  - [x] Frontend write hook + retry queue: capability-gated, browser-persisted, normalized single-fact emission, 404 deployment-skew disarm — #1434 (merged 2026-08-19)
- [x] RFC reconciliation: bring `COMPLETION_RECORDS.md` back in line with the landed implementation — grafana/pathfinder-rfcs#17 (merged 2026-08-19)

### In flight

[...]

Issue #1256, opened 2026-07-07
Validate the proposed Edit/Preview/JSON toggle + Save redesign


Jayclifford345:
## Context

Jess mocked up an alternative guide-editor toolbar (Edit/Preview/JSON button group + Save action) — needs review against the current implementation before building.

## Code pointers

- `src/components/block-editor/BlockEditorHeader.tsx:619-641` — existing Edit/Preview/JSON toggle, a Grafana `<ButtonGroup>` with three `<Button>`s (icons `pen`/`eye`/`brackets-curly`), driven by `viewMode: ViewMode` (`'edit' | 'preview' | 'json'`, defined in `src/components/block-editor/types.ts:13`) and `onSetViewMode`
- Mode-switching logic: `src/components/block-editor/hooks/useBlockEditor.ts`, `src/components/block-editor/hooks/useJsonModeHandlers.ts`
- Save is not currently a single dedicated "Save" button — `renderBackendButton()` (`BlockEditorHeader.tsx:432-498`) renders a status-dependent primary/secondary button (`Save as draft` / `Update draft` / `Publish` / `Update`), shown only when `isBackendAvailable`. When the backend is unavailable, saving is implicit via a local-save indicator (lines 533-546), with no explicit Save action.

Since a toggle already exists in roughly the shape Jess proposed, this issue is primarily a design-review task (`needs-review`): compare her mock against the current `ButtonGroup` + `renderBackendButton` pattern and decide what actually changes (positioning, the Save/local-save split, or visual style) before writing code.

---
Part of the Pathfinder UX review epic — [Figma board](https://www.figma.com/design/MuEIAafuIdrJOljvIy7E0T/Pathfinder-UX-review?node-id=1-4124)

Jayclifford345:
My apologies, I should have moved the images over with the ticket: 

This is the current version:

<img width="450" height="900" alt="Image" src="https://github.com/user-attachments/assets/609cd3d7-5a8c-4b27-a666-e112ef4063a4" />

This is the version that Jess recommends:

<img width="450" height="900" alt="Image" src="https://github.com/user-attachments/assets/4dfcc14d-81c6-4a95-97c0-caa3dc202974" />

moxious:
[...]

PR #1415, opened 2026-07-23
feat(completion-records): durable completion writes via backend proxy and retry queue

moxious:
## Intent

Guide and journey completions are ephemeral today. This PR makes a terminal completion durable by persisting it as a `CompletionRecord` in the stack's App Platform store, written through a plugin-backend proxy so identity is stamped server-side and never trusted from the client. A front-end retry queue keeps the write reliable across transient failures, navigation, and reload.

## What's in the PR

**Backend write proxy** (`POST /completion-records`)

- Persists one completion as a `CompletionRecord` on the aggregated App Platform API.
- **Server-stamped identity:** `userId` (ID-token `sub`), `userLogin`, `userDisplayName`, `orgId`, `stackNamespace`, and `recordedAt` come from the verified request context; the typed request body carries only *what* was completed, so any identity a client smuggles in is dropped on decode. The CRD validates field presence, not truth — only the server can stamp these trustworthily.
- **Validation:** client facts are checked against the CRD value domains (source / category / platform enums, `completionPercent` bounds, per-field byte caps, control-character reject); `completedAt` is bounded to `[now − 30d, now + 5m]`. Violations are terminal `400`s.
- **Per-user rate limit:** a token-bucket guard runs before any upstream work; exhaustion returns `429` with `Retry-After`.
- **Cache invalidation:** a successful create invalidates the namespace read cache (generation fence) and clears the negative-cache cooldown, so a post-write "my completions" reflects the new record.
- **Error taxonomy:** `201` created (durable); `404` = route not deployed here (structural); other non-`429` `4xx` terminal (dropped, no retry); `401` / `429` / `5xx` / network transient (retried).

**Front-end**

- **Retry queue** with capped exponential backoff over the transient / terminal / route-missing outcomes.
- **Persistence & partitioning:** pending writes persist across reload and are partitioned per user, with a cross-tab lease so a single writer drains the queue.
- **Capability gating:** writes arm only where the feature reports available.
- **404 deployment-skew disarm:** a `404` disarms writes for the session (persisted items survive for the next load), so a stack without the served group stays silently unavailable instead of storming retries.

## API surface

- **Group:** `pathfinderbackend.ext.grafana.app/v1alpha1`, kind `CompletionRecord` (plural `completionrecords`). Availability is advertised by the boot toggle `aggregation.pathfinderbackend-ext-grafana-app.enabled`; the completion routes gate on it alone.
- **Plugin routes:** `POST /completion-records` (write), `GET /completion-records/my`, `GET /completion-records/capability`.

## Testing

Focused Go (`go test ./pkg/plugin/ -run Completion`) and front-end (`jest src/completion-records`) suites pass: server-side identity stamping and the fail-closed inbound gate, the validation matrix, rate limiting, cache invalidation, the four-way error taxonomy, and the queue's persistence / partitioning / backoff and `404` disarm.

## Merge gate

- **Outstanding:** a live Viewer-attributed write through the **deployed** plugin proxy against the served API. The residual unverified piece is the proxy's identity forwarding in a real deployment.
- A direct browser write by a real Viewer user to the new group was verified 2026-07-24 (`201 Created`, RBAC enforced), which cleared the RBAC-layer question.
- This PR stays a **draft** until the merge gate clears and a human reviews.

Epic: https://github.com/grafana/grafana-pathfinder-app/issues/1411


moxious:
Superseded by a backend/frontend split for focused, high-confidence review:

- **Backend** (`POST /completion-records` write proxy): #1433
- **Frontend** (write hook + retry queue): #1434

The two branches were cut fresh from current `main` and together reproduce this PR's diff exactly (verified: backend ∪ frontend == this PR's changes over its paths, with unrelated main advancement such as #1256 preserved). `.cursor/rules/systemPatterns.mdc` was hunk-split by layer; `CHANGELOG.md` rides with the frontend PR. No hard merge dependency between them — backend-first is a soft sequencing preference only. Closing this in favor of the two.

Refs #1411

PR #1413, opened 2026-07-23
fix(e2e): use full Chromium for containerized headless runs

leslie-heinzen:
## Summary

A continuous guide-health pilot exposed unreliable containerized E2E executions: more than half of initial task attempts intermittently produced `REPORT_MISSING`, including guides that run consistently outside Docker. The failure occurred before guide execution: Playwright successfully authenticated and loaded Grafana, but timed out waiting for `docs-panel-container` after clicking Help. A repeated execution would either pass or reach the expected guide-specific result, making worker exit behavior appear inconsistent.

An investigation reproduced the failure locally with the published runner image and ruled out Cloud Run scheduling, task parallelism, Node version, container directory setup, and container mem limits. Native headless runs were reliable, as were headed runs inside the container under `Xvfb`. Linux headless runs failed consistently.

Our Playwright version uses a separate `chromium_headless_shell` binary when Chromium runs headless without an explicit channel, while headed runs use full Chromium. The runner now selects channel: 'chromium', making containerized headless execution use the same full Chromium binary that succeeded in headed mode. The image smoke test inspects the actual Linux browser process to prevent regression.

A locally built `linux/amd64` e2e image passed the browser-process smoke test and completed multiple consecutive pool-backed runs across disposable Grafana Cloud stacks without reproducing the docs-panel bootstrap failure.

## What to look at

- [Guide runner configuration](https://github.com/grafana/grafana-pathfinder-app/blob/3d21d649c4281eb19e53518839f1dd03b8b14919/tests/e2e-runner/playwright.config.ts): `channel: 'chromium'` makes headless guide execution use full Chromium instead of `chromium_headless_shell`.
- [Image smoke configuration](https://github.com/grafana/grafana-pathfinder-app/blob/3d21d649c4281eb19e53518839f1dd03b8b14919/tests/e2e-runner/runner-image-smoke.config.ts): the published-image smoke path uses the same browser selection as real guide runs.
- [Image smoke assertion](https://github.com/grafana/grafana-pathfinder-app/blob/3d21d649c4281eb19e53518839f1dd03b8b14919/tests/e2e-runner/runner-image-smoke.spec.ts): inspects Linux process command lines to fail if Playwright launches from `chromium_headless_shell-*`.

## How to verify

1. Build `Dockerfile.e2e-runner` for `linux/amd64`.
[...]

PR #1426, opened 2026-07-24
fix(e2e): harden docs panel bootstrap

leslie-heinzen:
## Summary

#1413 dramatically reduced the amount of errors related to `docs-panel-container` initialization in docker contrainer-driven e2e tests runs.

However, the issue remains intermittent. Containerized guide runs can still time out waiting for `docs-panel-container` when Grafana's Help button opens the generic Help menu before Pathfinder registration completes, or when an unconditional click closes a panel that is already opening. 

This replaces the unconditional Help toggle click with an ensure-open flow that detects when Pathfinder is already open or opening, retries only the setup sequence once, and leaves the 120-second total guide timeout and Playwright's whole-test retry policy unchanged.

## What to look at

- [Panel bootstrap helper](https://github.com/grafana/grafana-pathfinder-app/blob/4606a90a5305790448c8dcf91efaeb96a66f21bf/tests/e2e-runner/utils/guide-runner/bootstrap.ts) — combines panel visibility, docked-sidebar ownership, the sidebar-mounted event, and the expanded direct Help toggle before deciding to click. `aria-expanded` is deliberately not sufficient by itself because Grafana's generic Help dropdown also supplies it while its menu is open.
- [Guide runner integration](https://github.com/grafana/grafana-pathfinder-app/blob/4606a90a5305790448c8dcf91efaeb96a66f21bf/tests/e2e-runner/guide-runner.spec.ts) — reloads and reinjects the guide at most once after a complete bootstrap failure, before content dispatch, discovery, or guide-step execution begins.

## Why

PR 1413 removed the primary browser-level cause of this failure, but occasional runs still fail before guide execution begins. The runner currently assumes clicking Help always means "open Pathfinder," while Grafana can interpret the same click as opening the generic Help menu or closing a panel that is already opening. When that happens, the runner waits for a panel that will never mount instead of recovering from the missed open action.

## How to verify

1. Start the local Grafana development stack.
2. Build the same Linux runner image used in deployed jobs:

   ```shell
   docker buildx build --platform linux/amd64 --load -f Dockerfile.e2e-runner -t pathfinder-e2e-runner:local .
   ```

3. Run the E2E framework guide from inside the image against the host Grafana instance:

   ```shell
   docker run --rm --platform linux/amd64 \
     --add-host=host.docker.internal:host-gateway \
     pathfinder-e2e-runner:local \
     bundled:e2e-framework-test \
     --grafana-url http://host.docker.internal:3000 \
     --verbose
   ```

4. Repeat the container run several times and confirm each run reaches step discovery without timing out waiting for `docs-panel-container`.

## Breaking changes

## Out of scope

- Increasing the 120-second total guide timeout.
- Enabling whole-test Playwright retries.
- Changing bearer-token routing, feature-flag handling, or bootstrap reporting.


leslie-heinzen:
@cursoragent review

leslie-heinzen:
Addressed the blocking review feedback in `3669d803`:

- **Timeout/deadline coverage:** added tests for custom timeout propagation, expired-deadline clamping to 1 ms, and timeout propagation without a setup retry.
- **Sidebar mount detection:** added a stateful jsdom test that executes the real `page.evaluate` callback, arms `pathfinder-sidebar-mounted`, fires the event from the Help action, and verifies the mount signal is observed.
- **Retry/state transitions:** the harness now supports sequential panel and bootstrap states; new tests cover the panel becoming visible during readiness, dock ownership appearing after an unconfirmed click, generic-menu recovery, and the existing one-retry bound.
- **Docked-value parsing:** added null, malformed JSON, and triple-encoded depth-limit cases.
- **Mock fidelity:** replaced static-only mocks with state sequences and optional execution of the browser predicates.

Also addressed two non-blocking maintainability points:

- Removed the `.last()` menu-order assumption; generic Help detection now checks whether any accessible menu exists.
- Named the docked-value parse depth and Help-open attempt limits.

I left repo-wide extraction of `pathfinder-sidebar-mounted` and `__pathfinderPluginConfig` constants for a separate consolidation change, since those contracts are already scattered across production code and changing every owner would broaden this tests-only fix substantially.

Validation: 20 focused bootstrap tests, typecheck, Prettier, governance tests, and the full frontend coverage suite under Node 24 all pass.

tests/e2e-runner/utils/guide-runner/bootstrap.ts
```diff
@@ -0,0 +1,179 @@
...
+    }
+    state = await readBootstrapState(page, false);
+    if (hasOpenSignal(state)) {
+      break;
+    }
+    if (await isExpandedPathfinderToggle(page, helpButton)) {
+      break;
+    }
+
+    await helpButton.click();
```
moxious: [blocking] This click can also outlive the helper’s 10-second deadline when Help is obscured, detached, or not actionable. That bypasses the reload/reinjection recovery this change introduces. Pass remainingTimeout(deadline) as the click timeout.

tests/e2e-runner/utils/guide-runner/bootstrap.ts
```diff
@@ -0,0 +1,179 @@
...
+  return (await page.getByRole('menu').count()) > 0;
+}
+
+async function dismissGenericHelpMenu(page: Page): Promise<void> {
+  if (await isGenericHelpMenuVisible(page)) {
+    await page.keyboard.press('Escape');
+  }
+}
+async function isExpandedPathfinderToggle(page: Page, helpButton: Locator): Promise<boolean> {
+  return (await helpButton.getAttribute('aria-expanded')) === 'true' && !(await isGenericHelpMenuVisible(page));
```
moxious: [blocking] This locator wait is not bounded by the bootstrap deadline. Because the runner does not configure actionTimeout, a missing Help button can keep getAttribute waiting until the whole-test timeout and prevent beforeRetry from running. Please pass the remaining bootstrap timeout into this operation.

tests/e2e-runner/utils/guide-runner/bootstrap.test.ts
```diff
@@ -0,0 +1,278 @@
...
+      openConfirmationResults: [new Error('Mount event delayed')],
+    });
+
+    await ensureDocsPanelOpen(page);
+
+    expect(helpButton.click).toHaveBeenCalledTimes(1);
+    expect(panelWaitFor).toHaveBeenCalledTimes(1);
+  });
+
+  it('dismisses the generic Help menu and retries the open action once', async () => {
```
moxious, line 240: [blocking] Please pin the remaining bootstrap semantics with tests for: (1) aria-expanded="true" while the generic Help menu is visible, (2) both confirmation attempts failing and the second error propagating after exactly two clicks, and (3) panel DOM presence as the sole open signal. Each branch is central to this hardening but can currently regress while the focused suite remains green.

moxious (changes requested):
Final thought on general review, is that you're forming a needed contract between you & docspanel in order to do your e2e work.  Not sure where the most appropriate docs are, but you might consider essentializing what must be true about docs panel in order for you to do your thing, and putting either comments in docs panel to note this, or else something in the concern-based routing system.

This PR is pretty fine as is, but the thing that scares me about it is that docs panel needs a refactor which will almost certainly happen later, and the main thing is that your testing be tight enough that when that refactor happens, your contract is still honored.

tests/e2e-runner/utils/guide-runner/bootstrap.ts
leslie-heinzen: Addressed in a97b1de9. The Help click now receives `{ timeout: remainingTimeout(deadline) }`, so an obscured, detached, or non-actionable button fails within the bootstrap budget and reaches the setup retry. The custom-timeout and expired-deadline tests assert the exact click timeout.

tests/e2e-runner/utils/guide-runner/bootstrap.ts
leslie-heinzen: Addressed in a97b1de9. `getAttribute("aria-expanded")` now receives the remaining bootstrap timeout, preventing locator auto-wait from consuming the whole-test timeout. The timeout propagation and one-millisecond clamp tests pin this operation.

tests/e2e-runner/utils/guide-runner/bootstrap.test.ts
leslie-heinzen: Addressed in a97b1de9. Added focused cases for: (1) `aria-expanded="true"` with a visible generic Help menu, which still performs the open click; (2) two failed confirmations, which propagate the second error after exactly two clicks; and (3) panel DOM presence as the sole open signal with no mount event. The focused bootstrap suite now has 23 passing cases.

leslie-heinzen:
Addressed the overall docs-panel contract concern in `a97b1de9`:

- Added a **Docs panel bootstrap contract** section to `docs/developer/E2E_TESTING_CONTRACT.md` defining the four load-bearing signals: `__pathfinderPluginConfig`, `pathfinder-sidebar-mounted`, `docs-panel-container`, and `pathfinder-auto-open-docs`.
- Added source-level contract tripwires for plugin readiness and the sidebar mount event; existing contract tests already pin the panel container and content-open listener.
- Updated the `docs-retrieval-and-rendering` concern so docs-panel work loads the E2E contract, activates on all four signal names, explicitly checks that refactors preserve or atomically update them, and calls for the bootstrap contract tests during verification.

This should make both agents and human reviewers aware of the runner dependency when the docs panel is refactored, while the tests prevent silent contract removal.

Validation: 23 focused bootstrap tests, docs-panel contract tests, typecheck, Prettier, governance tests, and the full Node 24 coverage suite pass.

tests/e2e-runner/utils/guide-runner/bootstrap.ts
```diff
@@ -0,0 +1,181 @@
...
+    ({ panelTestId }) => {
+      const runtimeWindow = window as Window & { __pathfinderE2ESidebarMounted?: boolean };
+      return (
+        runtimeWindow.__pathfinderE2ESidebarMounted === true ||
+        document.querySelector(`[data-testid="${panelTestId}"]`) !== null
+      );
+    },
+    { panelTestId: testIds.docsPanel.container },
+    { timeout }
+  );
```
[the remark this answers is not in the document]
leslie-heinzen: Valid finding; addressed in baf06b97. The sticky `pathfinder-sidebar-mounted` flag is now only a pre-click guard. Post-click confirmation requires current `docs-panel-container` DOM, and a failed DOM confirmation only trusts current dock ownership before deciding to wait rather than retry. Added a regression where the mount event fires during the second click, the retry closes the opening panel, and the stale flag must not confirm success. Focused/contract tests, typecheck, formatting, governance, and the full Node 24 suite pass.

moxious (approved):

PR #1429, opened 2026-07-27
feat(block-editor): editor header redesign — two-row toolbar, labeled rocker + Save (#1256)

Spidy88:
## What & why

Editor header redesign ([#1256](https://github.com/grafana/grafana-pathfinder-app/issues/1256)) — rows 2–3 of Jess's mockup: a labeled view-mode rocker on its own line, a two-row title/toolbar layout, and a labeled Save, with a measured responsive collapse.

> **Stacked on the kebab consolidation ([#1428](https://github.com/grafana/grafana-pathfinder-app/pull/1428)).** Base retargets to `main` once #1428 merges — review #1428 first. Doing the kebab first shrank the toolbar, which is what makes this bar's responsive behavior work.

## Changes

- **Rocker** — `ButtonGroup` → labeled `RadioButtonGroup<ViewMode>` (Edit/Preview/JSON) with an icon-only container-query fallback (two sibling groups toggled by `display:none`, so it doesn't depend on `@grafana/ui` internal markup). `viewModeToggle` testid stays on the wrapper (tour anchor); `onChange` still routes through `onSetViewMode` (preserves viewMode persistence).
- **Two-row layout** — the single header row splits into `titleRow` (title + guide id + publish status) and `toolbarRow` (rocker left; undo/redo, Save, kebab right). `containerType` moves to `toolbarRow`; single-line, never wraps.
- **Save** — label-only at width, icon-only when collapsed. The two draft-save states ("Save as draft"/"Update draft") both simplify to **"Save"** (mirroring the already-simple Publish/Update); smart routing (save vs publish) unchanged.
- **Collapse breakpoint 420px** — measured: the labeled toolbar needs ~370–386px, so the rocker + Save collapse to icons at ≤420px (one breakpoint, not two).
- **Undo/redo always visible** — the plan hid them below 420px "via keyboard shortcuts", but no such shortcut exists (`useGuideHistory` has none), so hiding stranded the action. They fit even at the 320px floating minimum, so they stay.

## Deferred / out of scope
- **Title-row polish → PR 5**: Draft badge recolor (purple → blue) and the gradient underline. Preview keeps the title row rendered (spacer, no duplicate `<h1>`); fully hiding it in preview is PR-5 work.
- **Tab-bar chrome (wordmark + book/editor tabs) → PR 6**; search / link / resize stay out (per #1256).

## Testing
- `npm run test:ci`: **5688 pass** (+10 new). New `SaveActions.test.tsx` pins the 4-state label + save-vs-publish routing (a private draft must never publish); `BlockEditorHeader.test.tsx` now covers the rocker radios, undo/redo visibility, and preview mode. The one unrelated failure (`schema-e2e-report.test.ts`) is a pre-existing missing `ajv` dep.
- Typecheck / lint / prettier clean.
- Ran a max-effort `/code-review`: it caught preview-mode regressions from the two-row split (local-save indicator dropped for no-backend users, missing top padding, potential narrow-width overflow) — all fixed here by not special-casing preview.

#### Action row with labels
actions-with-labels

#### Action row collapsed to icon only
actions-icon-only

#### Before and after of action row (full width and collapsed)
save-matrix

### DIFF
--- a/src/components/block-editor/BlockEditor.persistence.test.tsx
+++ b/src/components/block-editor/BlockEditor.persistence.test.tsx
@@ -31,7 +31,7 @@ describe('BlockEditor persistence', () => {
   it('writes preview mode immediately when selected', () => {
     render(<BlockEditor />);
 
-    fireEvent.click(screen.getByRole('button', { name: 'Preview' }));
+    fireEvent.click(screen.getByRole('radio', { name: 'Preview' }));
 
     expect(screen.queryByTestId(testIds.blockEditor.palette)).not.toBeInTheDocument();
     expect(JSON.parse(localStorage.getItem(StorageKeys.BLOCK_EDITOR_STATE)!).viewMode).toBe('preview');
@@ -40,7 +40,7 @@ describe('BlockEditor persistence', () => {
   it('restores preview mode after a remount', () => {
     const { unmount } = render(<BlockEditor />);
 
-    fireEvent.click(screen.getByRole('button', { name: 'Preview' }));
+    fireEvent.click(screen.getByRole('radio', { name: 'Preview' }));
     unmount();
     render(<BlockEditor />);
 
--- a/src/components/block-editor/BlockEditorHeader.test.tsx
+++ b/src/components/block-editor/BlockEditorHeader.test.tsx
@@ -215,3 +215,38 @@ describe('BlockEditorHeader: selection mode (in the more-actions kebab)', () => 
     expect(onToggleSelectionMode).toHaveBeenCalledTimes(1);
   });
 });
+
+describe('BlockEditorHeader: toolbar row + undo/redo', () => {
+  it('renders the view-mode rocker as radios', () => {
+    render(<BlockEditorHeader {...baseProps} viewMode="edit" />);
+    expect(screen.getByRole('radio', { name: 'Edit' })).toBeInTheDocument();
+    expect(screen.getByRole('radio', { name: 'Preview' })).toBeInTheDocument();
+    expect(screen.getByRole('radio', { name: 'JSON' })).toBeInTheDocument();
+  });
+
+  it('shows undo/redo in edit mode, disabled when there is no history', () => {
+    render(<BlockEditorHeader {...baseProps} viewMode="edit" canUndo={false} canRedo={false} />);
+    expect(screen.getByRole('button', { name: 'Undo' })).toBeDisabled();
+    expect(screen.getByRole('button', { name: 'Redo' })).toBeDisabled();
+  });
+
+  it('does not render undo/redo outside edit mode', () => {
+    render(<BlockEditorHeader {...baseProps} viewMode="preview" />);
+    expect(screen.queryByRole('button', { name: 'Undo' })).not.toBeInTheDocument();
+    expect(screen.queryByRole('button', { name: 'Redo' })).not.toBeInTheDocument();
+  });
+});
+
+describe('BlockEditorHeader: preview mode', () => {
+  it('renders a spacer instead of the editable title input in preview', () => {
+    render(<BlockEditorHeader {...baseProps} viewMode="preview" />);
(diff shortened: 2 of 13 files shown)


Spidy88:
@cursoragent review

moxious (changes requested):
Thanks for the thoughtful header redesign. The PR otherwise looks merge-ready: mode persistence, save/publish routing, accessibility, architecture, and reversibility all reviewed clean, and CI is passing.

I found one remaining blocking responsive issue:

At the supported 320 px floating-panel minimum, the title row overflows when the status is `Published (modified)`. In `src/components/block-editor/header/header.styles.ts`, the combination of a non-wrapping `titleRow`, `titleArea` with `minWidth: 180`, and a non-shrinking `rightCluster` produces a measured `scrollWidth` of 353 px in a 320 px container.

Please:

1. Add a narrow-width treatment for the title row—for example, allow the title area to shrink below 180 px, responsively collapse the badge, or wrap the row.
2. Add a browser/layout regression test at 320 px using the `Published (modified)` state. It should verify that the row has no horizontal overflow and the status remains visible.

The previously reported Reset-guide toolbar overflow does not reproduce at the supported minimum, so no change is requested there.

Once the title-row overflow is fixed and covered by the focused regression test, I’m comfortable approving this PR.

Spidy88:
Both points addressed in `4a86b9bc`:

1. **Title-row overflow at 320px** — the culprit was `titleArea`'s hard `minWidth: 180` against a non-shrinking `rightCluster`. Changed it to `minWidth: 0` so the title area shrinks (the input truncates via its existing `minWidth: 0`) and the status badge stays fully visible instead of overflowing the row.

2. **Regression test** — added `tests/block-editor-title-row.spec.ts`: it opens the editor, constrains it to 320px, and asserts the title row has **no horizontal overflow** (`scrollWidth <= clientWidth`) and the status **stays visible** (badge right edge within the row). Verified **red before the fix (28px overflow) / green after**.

Note on the test: the `Published (modified)` state can't be reached locally without a live backend (it needs a published-then-modified guide), so the test forces the optional-backend toggle client-side to render the badge and rewrites the rendered badge's label to the widest value — the title-row flex layout is what's exercised. A `block-editor-title-row` testid was added to anchor the measurement.

The earlier Reset-guide toolbar report — agreed, no repro at the supported minimum, no change made there.

Spidy88:
Copilot review triage:

- **Ephemeral PR/issue refs in the title-row spec header comment** — valid, fixed in `3fd755db`. Matches our comment-hygiene rule (QC8 #3: no dead process-artifact refs); dropped `#1256 / PR #1429`, kept the invariant description.

- **`aria-disabled` vs `toBeDisabled()` in SaveActions.test.tsx** — keeping as-is. The button has `disabled` + `tooltip`, so `@grafana/ui` renders `aria-disabled` (a tooltipped button stays focusable) rather than the native `disabled` attribute, and `toBeDisabled()` fails against it (verified). The cited example (`InputBlockForm.test.tsx`) is a plain submit button with no tooltip, so `toBeDisabled()` works there. The test already includes the robust behavioral assertion (click is a no-op while posting) plus an inline comment explaining the `aria-disabled` choice.

Spidy88:
Cursor (icon rocker lacks hover labels) — valid, fixed in `2536fdce`.

The old ButtonGroup rocker had `tooltip="Edit"/"Preview"/"JSON"` per mode; the icon-only `RadioButtonGroup` only set `ariaLabel` (accessible name, but only a weak native `title` tooltip for sighted users). `RadioButton` renders `description` as a proper Grafana `<Tooltip>`, so I added `description` to the three icon-only options to restore the immediate styled hover label when the toolbar collapses. `ariaLabel` still provides the accessible name; the labeled (wide) variant is unchanged. Rocker unit + e2e (`getByRole('radio')`) still green.

moxious (approved):

Issue #1431, opened 2026-07-27
Migrate remaining design docs from legacy CAP group to new GAP group


moxious:
## Context

Follow-up to #1430 (migrate interactive guides API to the new Grafana App Platform group `pathfinderbackend.ext.grafana.app`, hard cutover). That PR migrated the active code, the user-facing developer docs (`EXTERNAL_API.md`, `CUSTOM_GUIDES.md`, `architecture/_index.md`), the `APP-PLATFORM-PUBLISH-HANDOFF.md` handoff spec, and `scripts/upsert-guide.sh` to the new group.

Out of scope for #1430, these design/reference docs still reference the legacy group `pathfinderbackend.ext.grafana.com`:

- `docs/design/AI-AUTHORING-IMPLEMENTATION.md`
- `docs/design/BACKEND_PROXY_PATTERN.md`
- `docs/design/phases/ai-authoring-0-assistant-spike.md`
- `docs/design/phases/ai-authoring-4-assistant-handoff.md`

## Impact

Minor doc drift. These are historical/design-record documents, so this is non-blocking — but they describe a group that active code no longer targets, which will confuse future readers once CAP is retired.

## Task

Decide per-doc whether to (a) update the group reference to `pathfinderbackend.ext.grafana.app`, or (b) add a dated note that the doc describes the pre-migration CAP state (appropriate for phase/spike records that are point-in-time by nature). Sweep after the GAP migration fully lands.

Verify afterward with:

```
git grep -n "pathfinderbackend.ext.grafana.com\|pathfinderbackend-ext-grafana-com" docs/
```

Related: #1430

moxious:
Sibling follow-up from the same #1430 review: #1432 (Grafana Assistant authoring tools group alignment).

moxious:
## 🧭 Implementation readiness

**Stakes:** Nothing changes for the user. This is internal work. It matters because one stale reference is a live, unchecked step that names a group the plugin no longer targets.

**Sanity check:** Three of the four listed docs still carry pre-migration references at this base. Two things moved since filing, because one listed doc was already corrected and a new deliberate legacy mention landed elsewhere.

**Severity:** 🟡 Low — documentation drift in design and phase records. No runtime behaviour and no user-visible surface. The worst case is wasted engineer time following a stale step.

**Code basis:**
[...]

PR #1433, opened 2026-07-27
feat(completion-records): add durable write proxy on the App Platform .app group

moxious:
Adds `POST /completion-records` to the Go backend: a durable write proxy that records a guide completion as a `CompletionRecord` on the stack's aggregated App Platform API, with identity stamped server-side.

Part of the durable completion records epic, https://github.com/grafana/grafana-pathfinder-app/issues/1411. The front-end queue that consumes this contract is #1434.

## What this adds

**The write route** (`pkg/plugin/completion_records_write.go`, `completion_records_write_ratelimit.go`, registered in `resources.go`). It stamps `userId`/`userLogin`/`userDisplayName`, `orgId`, `stackNamespace`, `recordedAt` and `schemaVersion` from the verified request context — never from the request body, so any identity a client smuggles in is dropped on decode. It validates one bounded JSON fact against the CRD's value domains (enums, byte caps, UTF-8 validity, a 30-day backdating horizon), applies a per-user token-bucket rate limit before any upstream work, and stales the namespace read index on success.

**Identity.** The stamped subject comes from the **inbound** `X-Grafana-Id`, via `deriveCompletionUserID` — the single identity contract shared with the read path, so reads and writes cannot drift apart on the key they join on. It is required and fails closed with a 401, which the client retries after re-auth. `userLogin`/`userDisplayName` are best-effort display snapshots read only from the signed token's `username`/`name` claims; there is deliberately no fallback, so an absent claim omits the field (and logs) rather than substituting an unverified value.

**Idempotency.** A non-blank client `idempotencyKey` is required. The record name is a DNS-safe `hash(userId ‖ key)` over the *trusted* userId, so a retried POST targets the same object (an upstream 409 is an idempotent success) and two users' identical keys can never collide in the shared namespace.

**Outbound credential.** The shared App Platform client gains a `create` path alongside `listPage`, both authenticated with an on-behalf-of access token minted per request and sent as `X-Access-Token` only — no `Authorization` header, no outbound `X-Grafana-Id`. `CheckRedirect` returns `http.ErrUseLastResponse` so a minted bearer credential is never handed to a redirect target; an unfollowed 3xx becomes a retryable status instead. A stack with no provisioned credential resolves as `obo-unavailable` and degrades through the structural 404 path rather than retrying forever.

**Store cutover.** Completion records move from the legacy `.com` group to GAP `.app`. Group-version and the aggregation toggle derive from a single `appPlatformGroup` const shared by both proxies, and the routing tests pin that neither route gates on the legacy `.com` toggle. The backend owner confirmed the legacy store empty, so this is a hard cut with no migration.

## Retry taxonomy

The response status *is* the client's retry contract — four outcomes across five status classes:

| status | outcome |
| --- | --- |
| 201 | created (durable) |
| 401 / 408 / 429 / 5xx / 3xx / network / token-exchange failure | transient — retried with capped backoff |
| 403 | disarm for the session, **records retained** for a later drain |
| 404 | structural "route not deployed here" — disarm, records retained |
| other 4xx | terminal (validation / schema) — dropped |

## This review round

- **The stale-serve fallback survives a write.** `invalidateCompletionIndex` deleted the cached entry, which also destroyed the warm index the read path deliberately serves when a refresh fails — so the first upstream blip after any write became a cold 503 instead of a slightly-stale 200. The entry is now marked stale rather than deleted. Staling is what actually skips the TTL fast path and forces the refresh; the generation bump only coalesces in-flight refreshes. The namespace-wide keying question is tracked separately as #1653.
- **Write identity routes through the canonical helper.** `completionWriterIdentity` called `subjectFromIDToken` directly while `deriveCompletionUserID` was documented as the contract. They agreed, but routing both through one helper makes future divergence in the join key impossible rather than merely unlikely.
- **Comments corrected where they described behaviour the code does not have.** The header and handler claimed a 403 was terminal and dropped the completion; it is echoed verbatim and the client disarms-and-keeps, so nothing is lost. `BACKEND_PROXY_PATTERN.md` §11 is corrected to match, as is a test comment describing a `PluginContext.User` profile fallback that does not exist.
- **`durationMs` clamps instead of rejecting.** An out-of-range value was a terminal 400 that discarded an otherwise valid completion over one denormalized convenience field. It now clamps into `[0, 24h]` with a log line. Unreachable today — nothing in `src/` populates the field — so this lands before the first producer can silently lose completions.
- **Contract anchor added** to `docs/design/CONCERNS.md`, per the convention that the anchor lands with the PR establishing the contract. It points at `BACKEND_PROXY_PATTERN.md` §11.
- **Test added for the 2xx/3xx classification** in `isTransientUpstreamStatus`, which also changed the read path: a 3xx LIST now reports a transient 503 rather than a terminal unavailable.
- **Unrelated changes dropped.** A CI triage comment, a `CUSTOM_GUIDES.md` paragraph and two `systemPatterns.mdc` rows described the custom-guide package work, not completion records. They were accurate but widened what a revert of this PR would take with it, so they are reverted to `main` and left for their own PR.

This branch also carries a merge of current `origin/main`, which brought in JWKS signature verification of the inbound ID token (#1604). The write path adopts that shape: `completionWriterIdentity` returns an `identityStatus`, so an unreachable JWKS is served as a transient 503 rather than reported as a bad caller token.

## Reversibility

Reversibility is asymmetric. Once `.app` writes land, a plain revert to the `.com` reader would orphan the new records — if this route is ever backed out, **remove the write and keep the `.app` reader**.

## Testing

`go build`, `go vet`, `golangci-lint run` and `go test -race ./pkg/...` are clean, as are `npm run prettier-test`, `npm run test:review-contract` and the `src/validation` suite.

Behavioural changes this round carry tests: the stale-serve fallback is pinned by `TestCache_InvalidationKeepsStaleServeFallback` (verified to fail against the previous delete-based invalidation), the clamping by `TestCompletionWrite_DurationDomain`, and the read-path classification by `TestErrors_UnexpectedRedirectOnListIsTransient`.

Earlier rounds drove the real handler through the real outbound client against a stub aggregated App Platform server, confirming that the outbound create carries only `X-Access-Token`, that the stamped `spec.userId` comes verbatim from the inbound token, that an unprovisioned stack short-circuits to a structural 404 with no upstream call, that a 307 reaches the redirect target with no request at all, and that idempotency keys are identity-scoped.

No screenshot: the user-facing surface here is an HTTP API.


pkg/plugin/app_platform_identity.go
```diff
@@ -48,6 +48,61 @@ func forwardIdentityHeaders(dst http.Header, idToken string) {
...
+// (an empty string is a valid, schema-permitted value): login prefers the
+// ID-token `login`/`preferred_username` claim, then the X-Grafana-User header;
+// display name prefers the token `name` claim, then the login.
+func completionWriterIdentity(r *http.Request) (userID, userLogin, userDisplayName string, ok bool) {
+	userID, ok = subjectFromIDToken(r)
+	if !ok {
+		return "", "", "", false
+	}
+	login, name := idTokenProfile(r.Header.Get(backend.GrafanaUserSignInTokenHeaderName))
+	userLogin = firstNonEmpty(login, strings.TrimSpace(r.Header.Get("X-Grafana-User")))
```
tomglenn: `[blocking][security]` **This fallback is not a fallback — it's the only path, and it reads a spoofable header.**

`idTokenProfile` below (~L86-94) reads the `login` and `preferred_username` claims. Grafana's ID token emits **neither**. From `authlib/authn/verifier_id_token.go` in the module cache:

```go
// Username of the user (login attribute on the Identity)
Username    string `json:"username,omitempty"`
// Display name of the user (name attribute if it is set, otherwise the login or email)
DisplayName string `json:"name,omitempty"`
```

So `login` is always `""`, `firstNonEmpty` always takes the second leg, and `spec.userLogin` comes from the `X-Grafana-User` request header **100% of the time** in production.

This repo already refuses that header for exactly this reason — `coda_exec.go:82-88`:

> The `X-Grafana-User` header is not an acceptable fallback for this endpoint because it can be set by any client whose request reaches the plugin without going through Grafana's auth proxy [...] `PluginContext.User` is set by the SDK from the authenticated session and cannot be spoofed.

The result is that a caller can durably persist an arbitrary chosen login (up to 256 bytes) on their own records — in the one identity field on the record that has no upstream validation. `userId` is fine (it's the token `sub`, backstopped by upstream RBAC on the forwarded token); this field is the exception. Records survive a revert, so this is worth fixing before any exist.

Fix:
- read `username` (keeping `preferred_username` is harmless), and
- prefer `userLoginFromContext(r.Context())` — already in this codebase at `coda_exec.go:166` — over the raw header for the fallback leg.

`userDisplayName` is mostly fine: the `name` claim is real, so it resolves from the token whenever it's set.

Worth noting *why* this slipped through: no test token in the suite carries a login or name claim — `completion_records_write_test.go:509-510` says so explicitly — so the dead branch is never exercised. A test with a token carrying `username`/`name`, asserting the header is **not** consulted, would have caught it.

pkg/plugin/completion_records_write.go
```diff
@@ -0,0 +1,427 @@
...
+		// numeric org — so this remains the correct field for the value itself.
+		OrgID:          backend.PluginConfigFromContext(r.Context()).OrgID, //nolint:staticcheck // numeric orgId required by CRD; Namespace is the string scope, not a number
+		StackNamespace: namespace,
+		SchemaVersion:  completionWriteSchemaVersion,
+	}, nil
+}
+
+// validateBoundedText rejects oversized or control-character content in a
+// client-supplied free-text field (→ terminal 400).
+func validateBoundedText(field, value string, maxBytes int) error {
```
tomglenn, line 338: `[blocking]` **Invalid UTF-8 gets past this cap and then triples in size.**

The length check counts pre-marshal bytes, and the control-char loop ranges *runes* — an invalid byte decodes to `utf8.RuneError` (U+FFFD), which is neither `< 0x20` nor `== 0x7f`, so it passes. `json.Marshal` then substitutes a 3-byte replacement character per invalid byte.

Verified with a scratch test against this branch:

```
input len=1024 bytes, validation PASSED, marshalled-field len=3072 bytes (expansion 3.0x)
```

That's `guideTitle` at 3× its 1024-byte cap; same for `guideId`/`guideSource`/`pathId`/`idempotencyKey` at 3× their 256. It defeats the invariant stated at L88-91 of this file — that this writer is the only bound between a hostile body and a durable record, because oversized stored strings can push read-path LIST pages past their byte cap.

One line fixes it:

```go
if !utf8.ValidString(value) {
    return fmt.Errorf("%s contains invalid UTF-8", field)
}
```

`boundedIdentityField` (L415) is **not** affected — it accounts with `utf8.RuneLen(r)`, which correctly returns 3 for U+FFFD, matching what `WriteRune` emits. Worth a row in the validation table at `completion_records_write_test.go:268`.

pkg/plugin/completion_records_test.go
```diff
@@ -68,10 +68,13 @@ func rec(userID, guideSource, guideID, title, category, pathID, source, complete
...
 
-// testGrafanaConfig is the healthy config: aggregation toggle on, app URL set.
+// testGrafanaConfig is the healthy config shared by the completion and
+// custom-guide proxy tests: both aggregation toggles on (they gate on different
+// group-derived toggles — completion on the .app group, custom guides on the
+// .com group) and an app URL set.
 func testGrafanaConfig() map[string]string {
 	return map[string]string{
-		featuretoggles.EnabledFeatures: pathfinderBackendAggregationToggle,
+		featuretoggles.EnabledFeatures: completionRecordsAggregationToggle + "," + pathfinderBackendAggregationToggle,
```
tomglenn: `[blocking]` **Enabling both toggles makes the gate untestable, and the new toggle string is pinned nowhere.**

Two tautologies stack here:

1. This fixture enables the `.app` **and** `.com` toggles, so every completion read/write/capability test passes whether the gate reads `completionRecordsAggregationToggle` or `pathfinderBackendAggregationToggle`. Reverting the gate to the legacy toggle would not fail a single test.
2. No test contains the literal `aggregation.pathfinderbackend-ext-grafana-app.enabled`. The fixture references the constant *symbolically*, and `aggregationToggle()` is never called directly with an assertion, so a wrong derived string is invisible. (Grepping `pkg/` for the toggle literals returns only the legacy `.com` const at `app_platform_client.go:44`.)

The failure mode matters: a wrong derived toggle name means `resolveCompletionConfig` returns `available=false`, so **every write 404s and every read reports `capability=false` on every stack** — with a fully green suite. The two toggle-off tests (`completion_records_test.go:723`, `completion_records_write_test.go:249`) set no toggle at all, so they don't discriminate either.

Two cheap additions close it:
- assert the literal: `aggregationToggle("pathfinderbackend.ext.grafana.app") == "aggregation.pathfinderbackend-ext-grafana-app.enabled"`
- a test that a config enabling **only** the legacy `.com` toggle yields `capability=false` / 404 on the completion routes

Then split this fixture so the completion tests enable only the `.app` toggle.

Note the group string itself *is* pinned once, at `app_platform_client_test.go:174` — the read path inherits `.app` correctness solely from sharing that const with the write test. The toggle has no equivalent anchor.

.cursor/rules/coda.mdc
```diff
@@ -86,10 +86,12 @@ Terminal states: `destroyed`, `destroying`, `error`.
...
+| `pkg/plugin/resources.go` | HTTP handlers: `/coda/register`, `/coda/exec`, `/vms`, `/vms/{id}`, `/sample-apps`, `/alloy-scenarios`, `/package-recommendations`, `/completion-records` (POST write), `/completion-records/my`, `/completion-records/capability`, `/custom-guide-repository`, `/health` |
+| `pkg/plugin/completion_records.go` | App Platform read proxy for completion records: per-user collation, 5-min TTL cache with single-flight refresh, invalidation generations, stale-serve/503/capability=false error paths (not a Coda feature) |
+| `pkg/plugin/completion_records_write.go` | App Platform durable write proxy (`POST /completion-records`): server-stamps identity/org/stack, validates the client fact against the CRD, builds the full spec, derives the record name deterministically from a client `idempotencyKey` when present (409 → success), invalidates the read cache on success, maps failures onto the retry-queue contract (201 created / 401 transient after re-auth / structural 404 disarm — persisted data survives / 429+5xx+network transient / other 4xx terminal) |
```
tomglenn: `[blocking]` **"when present" describes the behaviour this PR removed as a blocker.**

The key is **required**: `completion_records_write.go:236-238` rejects a blank/missing/whitespace key with a terminal 400 before any upstream work, and L53-56 of the same file states there is no random-name fallback. "when present" reads as optional-with-an-implied-fallback — which is precisely the keyless/random-name path the PR description classifies as the **High (blocker)** it just fixed.

This row also omits that the derivation is scoped to the trusted `userID`, which is the invariant that makes an upstream 409 provably the caller's own record.

Since the handler doc-comment at L36-37 names this file as co-canonical for the taxonomy, the sync assertion is literally false as written, and an agent trusting this row could reintroduce a keyless branch or write a test asserting a keyless POST succeeds.

Suggest: *"derives the record name deterministically from the server-stamped `userID` plus a **required** non-blank client `idempotencyKey` (blank/missing → terminal 400; 409 → idempotent success)"*.

Same wording needs fixing at `systemPatterns.mdc:185`. `BACKEND_PROXY_PATTERN.md:250-252` already says "required" correctly, so the divergence is internal to this changeset.

.cursor/rules/systemPatterns.mdc
```diff
@@ -180,9 +180,11 @@ The Go backend is a **thin bridge** between the React frontend and the **Coda VM
...
+| `plugin/completion_records_client.go` | Per-kind wrapper over the shared App Platform client for `completionrecords`: supplies group/version + resource, decodes each `items[].spec` for LISTs, and issues the create (POST) for the write path. |
+| `plugin/completion_records_write.go` | Completion Records durable write proxy (`POST /completion-records`). Fails closed with 401 on missing identity (the client retries after re-auth); server-stamps identity/org/stack/`recordedAt`/`schemaVersion`; validates a single bounded JSON fact; builds the CRD spec; derives the record name deterministically from a client `idempotencyKey` when present (409 → success); invalidates the read cache on success; maps outcomes onto the retry contract (201 created / 401 transient after re-auth / structural 404 disarm — persisted data survives / 429+5xx+network transient / other 4xx terminal). |
```
tomglenn: `[blocking]` Same "when present" issue as `coda.mdc:91` — the idempotency key is required, and a blank one is a terminal 400 with no fallback. Please fix both rows together.

docs/design/BACKEND_PROXY_PATTERN.md
```diff
@@ -212,6 +217,70 @@ One definition each, package-wide:
...
+  the free-text fields) and `completedAt` is bounded to a sane window
+  (`[now − 30d, now + 5m]`) to tolerate delayed offline/queued retries while rejecting gross
+  backdating; any violation is a terminal 400.
+- **A per-user token-bucket write rate limit** (`completion_records_write_ratelimit.go`, §9 flood
+  guard) runs before any upstream work; exhaustion returns 429 with `Retry-After`.
+- **A successful create invalidates the namespace read cache** (§4), advances its generation, and
+  clears the negative-cache cooldown (a create is fresh proof the upstream is reachable).
+  Any LIST that began before the write may finish for its caller but cannot repopulate that cache;
+  a post-write GET starts a new refresh.
+- **Outcomes map onto the front-end retry-queue contract (four-way):** 201 created (durable);
```
tomglenn: `[blocking]` **This paragraph contradicts both the code and line 248 of this same file.**

Two problems:

1. **"four-way"** — the taxonomy is five-way. `completion_records_write.go:38-51` lists 201 / 401 / 404 / 429+5xx+network / other-4xx.
2. **"other non-429 4xx terminal (validation / auth / schema — the client drops it)"** — 401 is *not* dropped. `isTransientUpstreamStatus(401)` is false so it takes the terminal branch and is echoed verbatim, but the frontend retries it as transient after re-auth. Line 248 of this file already says so: *"the client retries 401s as transient"*. The code comment at L50 deliberately says "validation / schema / 403" — not "auth" — for this reason.

This is the canonical doc the PR says was "updated to the real contract", so an agent implementing or reviewing #1434 against §11 would drop 401s and silently lose completions on token expiry — the exact four-location drift the PR set out to fix.

Suggest restating as: *201 created · 401 transient (retried after re-auth) · 404 structural, preserved verbatim · 429/5xx/network transient · other 4xx (validation / schema / 403) terminal*.

While you're in §11: the write path's own structural-unavailability gate returns **404** (`completion_records_write.go:161-168`), which is a deliberate exception to §7's "signal structural unavailability in-band as HTTP 200 with `capability.available=false`" rule — a write has no envelope to carry `capability`, and the frontend needs a disarm signal. §11 doesn't record that, so §7 still reads as governing it, and an agent could "correct" the 404 into a soft-200 and break the disarm. Worth one bullet plus a forward pointer from §7.

pkg/plugin/completion_records_write_test.go
```diff
@@ -0,0 +1,723 @@
...
+		err            error
+		wantStatus     int
+		wantRetryAfter string // "" means: header must be present (transient) but value unchecked; "-" means absent
+	}{
+		{"transient 503", &appPlatformUpstreamError{status: 503, msg: "boom"}, http.StatusServiceUnavailable, ""},
+		{"transient 429 echoes retry-after", &appPlatformUpstreamError{status: 429, retryAfter: "12", msg: "slow down"}, http.StatusTooManyRequests, "12"},
+		{"terminal 400 schema", &appPlatformUpstreamError{status: 400, msg: "bad spec"}, http.StatusBadRequest, "-"},
+		{"terminal 422 schema", &appPlatformUpstreamError{status: 422, msg: "unprocessable"}, 422, "-"},
+		{"collection 404 preserved as structural disarm signal", &appPlatformUpstreamError{status: 404, msg: "not found"}, http.StatusNotFound, "-"},
+		{"identity-scoped 403", &appPlatformUpstreamError{status: 403, msg: "forbidden"}, http.StatusForbidden, "-"},
```
tomglenn: `[blocking]` **403 here means the user's completion is destroyed, not retried — and the one status where the frontend contract diverges has no row at all.**

**On the 403 row.** Checked against #1434's actual code: `classifyWriteError` in `completion-write-client.ts` returns `{ kind: 'terminal' }` for any 4xx except 401 and 429, so 403 matches. `completion-write-queue.ts` then does `remove(item)` on a terminal outcome — dropping it from memory **and** from persisted storage, on the first attempt, with only a debug log. No retry, no telemetry, no operator signal.

A 403 on a *collection* create is a stack/role-level condition — an RBAC grant lagging the aggregation rollout is the obvious one — not a defect in the posted record. So "the write can never succeed as posted, so the client drops it" is the wrong bucket for it: the record is fine, the stack isn't. It belongs with the 404 disarm-and-keep path.

The counter-argument I expected to hold — that disarm-and-keep means an unbounded local queue on a permanently-403 stack — doesn't: #1434 caps the queue at 100 items with oldest-first eviction, and `processDue()` returns early once disarmed. So keeping costs a bounded footprint and one wasted POST per session. Data is lost either way in the truly-permanent case, and disarm-and-keep strictly dominates in every recoverable one.

In fairness, a Kubernetes 403 isn't exclusively role-level (admission denial and quota-exceeded also surface as 403) — but quota-exceeded argues for backoff, not drop, and no validating webhook sits in this path. Neither rescues terminal-drop.

Not blocking *this* PR since it ships dark, but this is the right place to fix it — remapping 403 alongside 404 is a backend change.

**On the missing row.** There's no upstream-**401** case in this table. It's the single 4xx where backend and frontend behaviour diverge (echoed verbatim, then retried client-side rather than dropped), and `isIdentityScopedUpstreamStatus` groups 401 with 403 right next door — so a future refactor routing identity-scoped statuses through one branch would silently convert an expired-token 401 into a drop and permanently lose completions that would have succeeded. One row guards it:

```go
{"identity-scoped 401 echoed verbatim", &appPlatformUpstreamError{status: 401, msg: "unauthorized"}, http.StatusUnauthorized, "-"},
```

pkg/plugin/completion_records_write.go
```diff
@@ -0,0 +1,427 @@
...
+	"net/http"
+	"strconv"
+	"strings"
+	"time"
+	"unicode/utf8"
+
+	"github.com/grafana/grafana-plugin-sdk-go/backend"
+)
+
+// Completion Records durable write proxy (docs/design/BACKEND_PROXY_PATTERN.md).
```
tomglenn, line 19: `[suggestion]` **QC8 — this file is carrying ~46 lines of prose before its first declaration, and one invariant is stated eight times.**

`AGENTS.md` §Comments is "default to no comments", with a four-item keep-list: counterintuitive-but-correct code, hidden invariants the type system can't express, external-bug workarounds with an upstream link, and security/correctness warnings. Much of what's here is outside that.

Concretely:

- **The identity-scoped-name invariant** is restated at L52-64, L182-184, L194-196, L371-380, L383-386, L390-392, L398-403, and again at `completion_records_client.go:107-110`. Eight places to update; the first one that drifts becomes an authoritative-looking contradiction on this file's security-critical property.
- **Process artifacts that rot** (shape 3): `#1434` at L54, L63, L130; `#1431/#1432` at `app_platform_client.go:42`; a test name at L386 (`Find References` gives that for free, and it breaks on rename); *"divergence re-creates the split-review bug"* at L37, which is meaningless to anyone outside this review. Also `app_platform_identity.go:43` cites `commit 89d6bd5e on feat/external-import-api` — a branch that no longer exists — as the justification for the `forwardIdentityHeaders` the new create path now depends on; "trim on touch" applies since you're editing that file.
- **Narrating the next line** (shape 1): L202 *"Surface the new record promptly on the next GET"* above `invalidateCompletionIndex(namespace)`, which says that in its name and again in its own doc comment; L152 *"Per-user flood guard … before any upstream work"* above `rateLimiter.allow(userID)`.
- **Defending a non-action** (shape 2): L114-116, L211-212, and six lines at `app_platform_client.go:209-214` defending one discarded error.
- **`RFC §6.9` / `§9` / `§7.1`** (L35, L44, L81, L152, L292) point at an internal design doc but read as IETF RFCs. Either spell them `docs/design/<doc>.md §6.9` or drop them.
- L36-38 self-declares a four-way duplication with the PR body and two `.mdc` files and no single source of truth — and two of the four have in fact drifted (see the `coda.mdc` and §11 comments). That's the failure mode, documented in the comment that was meant to prevent it.

I'd keep exactly two of the eight restatements: the `completionRecordName` doc comment (compressed to the invariant — "the name is scoped to the trusted userID, so a 409 cannot be another caller's record") and the `CORRECTNESS:` warning at L383, minus the test-name reference. A security warning against a specific future edit earns its place. The taxonomy should point at the design doc rather than restating it.

Non-blocking and no single comment here blocks merge — but at this ratio it's worth a pass before it becomes eight things to keep in sync.

pkg/plugin/completion_records_write.go
```diff
@@ -0,0 +1,427 @@
...
+		return
+	}
+
+	spec, err := a.buildCompletionSpec(r, req, userID, userLogin, userDisplayName, namespace)
+	if err != nil {
+		a.writeError(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	// Identity-scoped deterministic name (userID guaranteed present at :140, key
```
tomglenn: `[nit]` This line reference is already wrong on the commit that introduced it — L140 is the `return` inside the method-not-allowed guard; identity is derived at L146.

In-source line numbers can't survive an edit above them. The sentence works without coordinates: *"userID is non-empty before this point; key validated in buildCompletionSpec"* — which is the actual invariant. Worth fixing rather than re-numbering, since this is the one that matters (it's the audit trail for "userID is always present before name derivation").

pkg/plugin/completion_records_write.go
```diff
@@ -0,0 +1,427 @@
...
+		return completionRecordWriteSpec{}, fmt.Errorf("invalid platform")
+	}
+	if req.CompletionPercent < 0 || req.CompletionPercent > 100 {
+		return completionRecordWriteSpec{}, fmt.Errorf("completionPercent out of range")
+	}
+	if err := validateCompletedAt(req.CompletedAt); err != nil {
+		return completionRecordWriteSpec{}, err
+	}
+
+	durationSeconds := int64(0)
```
tomglenn, line 299: `[suggestion]` `durationMs` is the only numeric client field with no bounds check, and it's inconsistent with its neighbour four lines up.

Two things:
- A **negative** value is silently coerced to 0 rather than rejected — so a buggy client's bad data is durably accepted instead of surfacing as the terminal 400 the taxonomy defines for schema violations. `completionPercent` right above is explicitly range-rejected.
- There's **no upper bound**: `durationMs: 9223372036854775807` persists `durationSeconds = 9223372036854775`. If the CRD types that field narrower than int64 it becomes an upstream 400 the retry queue drops; if not, garbage is durable.

Also worth confirming intent: sub-second completions collapse to `durationSeconds: 0`, indistinguishable from "no duration supplied", and the field has no `omitempty`.

Suggest rejecting `< 0` and `> ` some ceiling with a 400, and adding both to the validation table.

pkg/plugin/app_platform_client.go
```diff
@@ -150,20 +168,96 @@ func (c *appPlatformListClient) listPage(ctx context.Context, groupVersion, name
 	return &appPlatformListPage{Specs: specs, Continue: list.Metadata.Continue}, nil
 }
 
+// create POSTs a single object to a namespace collection (the write companion
+// to listPage). It returns the created object body on 200/201, or an
+// appPlatformUpstreamError carrying the upstream status (and Retry-After, when
+// present) so the caller can classify transient/terminal/identity-scoped and
+// echo the upstream backpressure hint. Body is bounded by maxBytes.
+func (c *appPlatformListClient) create(ctx context.Context, groupVersion, namespace, resource string, obj []byte, maxBytes int64) ([]byte, error) {
```
tomglenn: `[suggestion]` Two things on this new method.

**1. The `[]byte` return has no production consumer.** The sole caller, `completionRecordsClient.Create` (`completion_records_client.go:181`), discards it: `if _, err := c.inner.create(...)`. So `maxBytes`, the `body[:maxBytes]` truncation at L216-218, and `completionWriteMaxBytes` exist only to satisfy tests. Worse, the doc comment promises "the created object body", but on the over-cap path it returns a silently truncated fragment and on a read error a partial prefix — with no distinguishing signal. Swallowing those errors is *right* (the write is already durable), but the return value shouldn't be typed as a usable object body.

Suggest `create(...) error` and a single `io.Copy(io.Discard, io.LimitReader(resp.Body, cap))` drain. That also resolves the unbounded drain at L219 — `io.Copy` with no limit after the cap has already been exceeded lets a pathological upstream stream for the full 15s timeout to no benefit.

**2. The first-request credential diagnostic never fires for writes.** `credentialDiagOnce` (L89-93) is a single package-level `sync.Once` fired only inside `listPage` (L127). Because reads already ship, a LIST will always consume it before any write — so the write path gets no diagnostic *ever*, not merely a duplicate one. §9 of the pattern doc justifies that log on the grounds that the likeliest incident for this shape is "the credential model doesn't authenticate on a real stack", and the two verbs can diverge: a POST can 403 where a GET returns 200. That's exactly the failure it exists to make one-line diagnosable. Suggest a small `logFirstUpstreamCall(op, status)` helper called from both, or a separate once-gate in `create`.

Also `[nit]`: the type is now a read+write resource client but is still named `appPlatformListClient` (3 in-package call sites, mechanical rename).

pkg/plugin/completion_records_write.go
```diff
@@ -0,0 +1,427 @@
...
+	// with backoff — an expired forwarded token is time-recoverable after re-auth.
+	userID, userLogin, userDisplayName, ok := completionWriterIdentity(r)
+	if !ok {
+		a.writeError(w, "unauthenticated", http.StatusUnauthorized)
+		return
+	}
+
+	// Per-user flood guard (RFC §9) before any upstream work.
+	if a.completionWriteRateLimiter != nil {
+		if allow, retryAfter := a.completionWriteRateLimiter.allow(userID); !allow {
```
tomglenn, line 181: `[suggestion]` This rate-limit rejection returns with no log line at any level, and the structural-404 return at L164-170 does the same.

Net effect: an operator can't distinguish "no writes arriving" from "every write rejected". The 404 is the sharper one — the response triggers a **sticky, session-wide** client-side disarm, so one unlogged response silences every subsequent write for that session with no server-side evidence.

Both sibling read proxies log their equivalent rungs (`completion_records.go:336`, `custom_guide_repository.go:130`), so the write path is currently less observable than the reads for the same failure class. In-convention fix: one `logger.Debug` on the 429 path, one `logger.Info` on the 404 (matching `custom_guide_repository.go`'s use of Info for "never works here", and it's at most once per session).

Keep attributes to `namespace` and the existing `reason` token — `userID`, `userLogin`, `guideTitle`, and `idempotencyKey` are all in scope at both sites and none belong in a log attribute. (The existing Debug at L204 correctly stays on `guideSource`/`guideId`/`name`.)

No `TELEMETRY.md` obligation applies here, for the record — that doc is frontend-only, and there's no metrics transport in `pkg/`. This is purely about matching what the sibling routes already do.

pkg/plugin/completion_records.go
```diff
@@ -557,26 +603,50 @@ func (a *App) writeMyCompletions(w http.ResponseWriter, resp myCompletionsRespon
...
+	if completionListerOverride != nil {
+		return completionListerOverride, namespace, true, ""
+	}
+	return newCompletionHTTPClient(appURL, idToken, a.ctxLogger(r.Context())), namespace, true, ""
+}
+
+// resolveCompletionWriteBackend is the write-path companion to
+// resolveCompletionBackend: same structural-availability gate, but it returns a
+// creator (POST) and honors completionCreatorOverride for tests.
+func (a *App) resolveCompletionWriteBackend(r *http.Request) (creator completionRecordCreator, namespace string, available bool, reason string) {
```
tomglenn, line 670: `[suggestion]` This and `resolveCompletionBackend` above are identical after normalization — resolve config, early-return, return the package-global override if set, else construct `newCompletionHTTPClient(appURL, idToken, ...)`. Both return the **same concrete type** widened to two different single-method interfaces; the only real difference is which override global is consulted.

Collapsing them would also halve the test-override surface, which is the part that's starting to bite: `completion_records.go` now carries 10 package-level mutable vars, and `withCreator` (`completion_records_write_test.go:36`) swaps only `completionCreatorOverride` without resetting the cache — while a successful write test calls `invalidateCompletionIndex`, mutating shared `completionGenerations` and deleting cache entries that persist into whatever test runs next.

```go
type completionBackend interface {
    completionRecordLister
    completionRecordCreator
}
```

`completionHTTPClient` already satisfies both, so one resolver and one override seam would do. Longer term the six cache maps + mutex would be better as a struct on `App` next to `completionWriteRateLimiter` — which this PR already models correctly one file over.

Related small duplication in the same seam: `completionRetryAfterSeconds` (L44) and `completionWriteRetryAfterSeconds` (`completion_records_write.go:73`) are both `30` for the same hint, and the new `completionWriteRateLimiter` duplicates `execRateLimiter` almost line-for-line (same field pair, same `allow` body) — differing only in two constants and the clock seam.

pkg/plugin/app_platform_client.go
```diff
@@ -150,20 +168,96 @@ func (c *appPlatformListClient) listPage(ctx context.Context, groupVersion, name
...
+	if errors.As(err, &ue) {
+		return ue.retryAfter
+	}
+	return ""
+}
+
+// isTransientUpstreamStatus reports whether an HTTP status should be retried:
+// 429, 5xx, and unexpected 2xx responses that did not satisfy an operation's
+// narrower success contract.
 func isTransientUpstreamStatus(status int) bool {
```
tomglenn, line 345: `[question]` Adding the 2xx clause here also changes the **read** path's taxonomy, since `isTerminalUpstreamError` inverts this function and the read proxy classifies through it (`completion_records.go:656-658`).

`listPage` errors on any `status != 200`, so a 202/204/206 on a LIST now produces a 2xx-status error that flips from terminal (`available:false`, `reason: backend-unavailable`) to transient (503 + `Retry-After`). That may well be an improvement, but it's an unstated behavioural change to a shipped route with no read-path test.

Was that intended? If the 2xx handling is really about the write path's narrower success contract (`create` accepts only 200/201), it might belong locally in `writeCompletionUpstreamError` rather than in the shared classifier.

Separately: the remap at `completion_records_write.go:352-355` covers 2xx only, so a non-followed 3xx (300/304/305) falls through and is echoed verbatim — a bucket the documented taxonomy doesn't define. A `status >= 400` guard before echoing would close it.

tomglenn (changes requested):
## Request changes — 4 must-fix items, all small

`go build`, `go vet`, `go test ./pkg/...` all pass at `0072b02b`. The route ships dark (confirmed: no in-tree caller), so merge risk is low — but three of the four blocking items write bad data or hide a broken feature, and all are cheaper now than after records exist. Total fix size is roughly 30 lines including tests.

### Must fix
1. **The login claim is read from a JWT field Grafana doesn't emit** (`app_platform_identity.go`) — so `spec.userLogin` always falls back to the spoofable `X-Grafana-User` header. Verified against `authlib/authn/verifier_id_token.go`.
2. **Invalid UTF-8 defeats the byte caps by 3×** (`completion_records_write.go`) — verified with a scratch test.
3. **The new `.app` aggregation toggle is pinned by nothing**, and the test fixture enables both toggles, so the gate is tautological.
4. **Three docs contradict the code this PR just wrote** — both `.mdc` rows say the idempotency key is optional; §11 says "four-way" and puts auth in the drop bucket.

### Before the frontend arms writes (not blocking this PR)
An upstream **403 permanently destroys the completion** — #1434 classifies it terminal and drops the item from memory and storage on the first attempt. A 403 here is a role/grant condition, not a defect in the posted record, so it belongs with the 404 disarm-and-keep path.

Also: the partial-revert recipe currently lives only in this PR body. Move it into `BACKEND_PROXY_PATTERN.md` §11 — on-call doesn't read PR descriptions. Cleaner still, land the `.com`→`.app` group migration as its own PR so both halves revert mechanically. And it's worth one line of confirmation from the backend owner that no `CompletionRecord` objects exist on the old group, since the reader stops addressing it after this merge.

### Findings I checked and dropped
Flagging these so you don't chase them — several were argued confidently by earlier passes but don't survive scrutiny:
- The **cache-invalidation "load lever"** is not a meaningful new capability: the same signed-in user can already issue an unthrottled namespace LIST directly against the aggregated API from their browser session, exactly as `fetchBackendGuides.ts` does. The proxy grants nothing extra.
- The **single-flight generation machinery is sound**. Kubernetes continue tokens are bound to the initial LIST's resourceVersion, so a paginated drain that started before a write *cannot* observe that write on any later page — discarding it is necessary, and publishing it would install a stale-but-fresh-looking index for a full 5-minute TTL. Proposals to serialize the flights instead would be worse.
- The **record name is not a frozen persisted contract** — it's never read back (the read path decodes only `items[].spec`; there's no get-by-name/put/patch/delete anywhere), so a golden-vector test would be a brittle change-detector.
- **No rate-limiter data race**: `tokenBucket` carries its own mutex, and `Retry-After` is always exactly `1`.
- Clearing the failure cooldown on create is correct — identity-scoped 401/403 never enter that cache, and without the clear every post-write read would 503 for up to 30s.

### One thing I could not verify
Whether a client can set `X-Grafana-Id` on a plugin resource call, or whether Grafana overwrites it server-side. This is pre-existing on `main` (the read routes already depend on it), but this PR escalates the consequence from an empty read to a durable record. Your open merge gate covers the positive case — consider adding a negative one, since a successful write can't distinguish "upstream verified the signature" from "upstream also trusts the forwarded header".

Non-blocking observations are inline: comment hygiene (the largest one), test gaps, an unbounded `durationMs`, the dead `[]byte` return, missing logs on two failure rungs, and the duplicated resolvers.

pkg/plugin/completion_records_write.go
moxious: Thanks for this review — it was substantive, and three of the four blocking items were real. All three are fixed: the `username`/`name` claim decoding with the trusted `PluginContext` fallback, the `.app` toggle literal now pinned with a `.com`-only-unavailable resolution case, and the docs corrected on the required idempotency key and the five-way taxonomy.

On this item I want to record a disagreement, with evidence, because I could not reproduce the consequence through the handler.

Your measurement reproduces exactly — 1024 invalid bytes in, 3072 out. But `encoding/json` performs the replacement at **decode**, not at marshal, so on the request path the expansion happens *before* `validateBoundedText` runs:

```
wire invalid bytes = 1024
decoded len(string) = 3072
utf8.ValidString(decoded) = true
existing cap len(value) > 1024 rejects it = true
```

The sole caller is `buildCompletionSpec`, operating on `json.NewDecoder(...).Decode` output, so `utf8.ValidString` cannot be false at that call site, and the existing byte cap already returns a terminal 400 with nothing written upstream. The short case behaves the same way: 10 invalid bytes decode to 30 valid bytes, well under the cap. I believe the scratch test called the validator directly with a raw-invalid-byte Go string, which is a state the HTTP path cannot produce — correct measurement, but the inflated bytes don't reach the durable record.

Residual effect is that invalid bytes persist as U+FFFD replacement characters in stored text: cosmetic and bounded, not a cap breach.

Separately, the item you flagged as the one thing you could not verify — whether a client can set `X-Grafana-Id` on a plugin resource call and thereby forge a record identity — is now closed by the on-behalf-of rework in this PR. The inbound token is exchanged at auth-api to mint the outbound credential, so a forged subject token cannot mint one at all and the write never reaches the store. Your concern got answered by the auth change rather than by a negative test.

Your **403** point is deliberately still open. It's a real disagreement, and the credential rework strengthens your side of it — the likeliest real-world 403 is now a missing delegated-permissions grant in ops or prod, which is exactly the recoverable stack-level condition you described rather than a defect in the posted record. I'd rather settle that with you than decide it while you're out. Same for your non-blocking suggestions (comment density, the duplicated resolvers, the unused `[]byte` return, the `appPlatformListClient` rename, the two missing log rungs) — all noted, none dismissed, and better done in one pass once #1408 has landed.

Note on this PR's current state, since the diff will look larger than expected: it temporarily includes your #1408 commits, because the on-behalf-of exchanger it now depends on isn't on `main` yet. That collapses to just our own changes once #1408 merges. No rush from our side — this PR is not being driven to merge-ready yet, and we expect further changes after yours lands.

moxious:
## Stable checkpoint — and the calls deliberately deferred until #1408 merges

This PR is at a **stable, green checkpoint**, not at merge-readiness. That is intentional. Recording the open items here so they live on the PR rather than in someone's notes.

### Where this stands

- Reworked onto the **on-behalf-of auth** from #1408 (`pkg/plugin/auth`): the create path sends only `X-Access-Token`, the write resolver guards a nil exchanger into `obo-unavailable` → structural 404, and a `Mint` failure classifies transient with a Faro-visible warning. Identity stamping still derives from the **inbound** `X-Grafana-Id`, unchanged.
- Added an outbound-credential regression test that fails if the old ID-token-as-credential shape ever returns, plus a test pinning that the record's subject still comes from the inbound token.
- Renamed the credential log field to `outboundCredential` — the old `identityHeaders` name described the superseded model.
- Added a UTF-8 rejection to `validateBoundedText`, labelled in-source as belt-and-braces: see the discussion below for why it is unreachable on today's JSON path.
- Documented the `POST {plugin}/completion-records` route in the `CODA.md` route table and the matching `systemPatterns.mdc` row. Neither this PR nor #1434 owned those rows before; this PR adds the handler, so it owns them.

### ⚠️ This diff temporarily contains #1408's commits

The on-behalf-of exchanger this PR now depends on is not on `main` yet, so #1408 is merged in as a temporary dependency. **That is why the file count looks large.** It collapses to only this PR's own changes once #1408 lands. No rebase and no force-push were used at any point.

### Deferred until #1408 merges — then reassessed depending on how it merges

**1. Four findings in inherited #1408 code — approved unfixed.**
This PR's review scanned the merged-in frontend, so these are in `custom-guide-packages` files verified byte-identical to that branch, not in this PR's own commits. They were **approved without fixing**: patching another author's in-flight PR from inside this one would worsen the merge-order coordination rather than help it.

- **Confirmed** — `AppPlatformPackageResolver` returns `ok: true` immediately when `loadContent` is falsy, with no existence or published check, and `createCompositeResolver` appends it last unconditionally. Resolution therefore succeeds for *any* package id on a stack with a namespace and the toggle on.
- **Confirmed** — locked milestones are guarded in `src/components/docs-panel/CustomGuidesSection.tsx` (early return on click, `disabled`, locked styling) but `src/components/docs-panel/context-panel.tsx` has no `isLocked` handling, so an unpublished milestone renders as an enabled button and clicking it navigates to an empty URL.
- *Reported, not independently verified* — negative-cache eviction keys off `result.repository` while `resolveUncached` returns the last resolver failure (always app-platform-tagged), which would disable negative caching for all packages, including OSS stacks with no namespace.
- *Reported, not independently verified* — `invalidateCustomGuideRepositoryCache` has no production caller despite a docstring describing post-publish use, leaving the catalogue up to 30s stale.

**2. The upstream-403 taxonomy.** Currently terminal-drop. @tomglenn argues it belongs with the 404 disarm-and-keep path, since a 403 on a collection create is a stack or grant condition rather than a defect in the posted record, and the frontend queue is already bounded. The on-behalf-of rework **strengthens that argument**: the likeliest real-world 403 is now a missing `set:pathfinder-app-delegated-permissions` grant in ops or prod (see #1503), which is exactly the recoverable case. Left open deliberately to settle with him rather than decide it in his absence.

**3. Re-sync with `main`** once #1408 lands, at which point this diff should shrink to its own changes.

**4. @tomglenn's non-blocking suggestions**, batched for one pass rather than piecemeal: comment density in the write handler, the duplicated resolvers, the unused `[]byte` return on `create`, the `appPlatformListClient` rename, and the missing log rungs on the local-429 and structural-404 paths.

**5. Re-review after all of the above.** The existing change requests are being left in place, not dismissed.

### Related

- Frontend companion #1434 is at its own green checkpoint, also deliberately unmerged.
- Epic: #1411

tomglenn (approved):

tomglenn (changes requested):
Switching my approval to request-changes. Nothing here is a design problem — the identity handling, the OBO minting, the idempotency scoping and the redirect refusals all hold up, and I re-verified the write path against the current head. These are small, and they belong on this branch rather than in a follow-up, because each one only exists because of the code this PR introduces.

## Please fix before merge

**1. `invalidateCompletionIndex` deletes the cached entry, which also removes the stale-serve fallback** — `pkg/plugin/completion_records.go:218-224`

`getCompletionIndex` deliberately serves a warm-but-stale index when a refresh fails, so a brief upstream problem shows users slightly old data instead of an error. After any successful write there is nothing warm left to serve, so the next upstream blip turns what would have been a stale 200 into a cold 503.

The function already bumps the generation, which is what forces the refresh. Dropping the `delete(completionCacheEntries, namespace)` keeps the fallback and changes nothing else about freshness:

```go
completionGenerations[namespace]++
delete(completionLastFailure, namespace)
```

The wider question — that this invalidation is namespace-wide, so one user's writes drop the shared index for every reader — is a keying decision rather than a bug in this PR, and I've filed it separately as #1653. Only the stale-serve half needs to happen here.

**2. The write path derives its identity through a different helper than the read path** — `pkg/plugin/app_platform_identity.go:59`

`completionWriterIdentity` calls `subjectFromIDToken` directly, but `deriveCompletionUserID` (`completion_records.go:62-68`) is documented as the canonical identity contract and says in terms that the write hook must stamp `spec.userId` with that exact helper.

They return the same value today, so this is a one-line change. It matters because that value is the join key between reads and writes: if the two paths ever diverge — someone adds normalisation, a prefix rule, or a fallback to one and not the other — writes land under a user ID that reads never match, and completions simply stop appearing with nothing logged anywhere. Routing both through the documented helper makes that divergence impossible rather than merely unlikely.

## Comments that describe behaviour the code doesn't have

`completion_records_write.go:56` says "other 4xx terminal — validation / schema / 403", and `:415-417` says "A 403 is terminal and drops the completion".

Neither is true as of #1434's current head. This handler echoes the upstream 403 verbatim (`:400`), and the client maps a 403 to its `forbidden` outcome, which takes the disarm-and-keep path and leaves the record persisted for a later drain (`completion-write-client.ts:127`, `completion-write-queue.ts:213-227`). Nothing is dropped — which is the behaviour we want, and it reads here as though it isn't.

Worth correcting because I initially triaged this as silent data loss on the strength of these comments, and only found otherwise by reading #1434. The next person to touch the taxonomy will do the same.

Two smaller ones in the same class:

- `completion_records.go:67` points at `docs/developer/CODA.md` for the trust boundary. That statement now lives in `docs/design/BACKEND_PROXY_PATTERN.md` — `CODA.md` carries no identity prose at all after the plugin split.
- `completion_records_write_test.go:74-77` describes `PluginContext.User` as "the profile-snapshot fallback source when the ID token carries no username/name claim". There is no such fallback, and the very next test asserts it must never reach the record.

## Contract anchor

This PR establishes the durable-write contract that #1434 conforms to — the retry taxonomy, the identity-scoped deterministic name, the 404 disarm exception — but `docs/design/CONCERNS.md` has no `completion-records` row. The convention is that the anchor lands in the PR that establishes the contract, so that the next change to this subsystem gets reviewed against something. `BACKEND_PROXY_PATTERN.md` section 11 is doing the job in practice; it just needs the pointer.

## Minor, take or leave

- `completion_records_write.go:287-289` — a `durationMs` over the 24-hour ceiling is a terminal 400, which discards an otherwise valid completion. Nothing in `src/` populates that field yet, so it's unreachable today, but clamping would be kinder than dropping the record once a producer lands.
- `app_platform_client.go:345-350` — the 2xx/3xx clause in `isTransientUpstreamStatus` also changes the read path, where a 3xx LIST now reports as a transient 503 rather than a terminal unavailable. The client treats both as unavailable so there's no visible difference today; it's just untested for the classification change as opposed to the credential-leak property.
- The CI triage comment and the custom-guides/systemPatterns rows are follow-on documentation for the package work with no relation to completion records. I checked the claims and they're accurate — they just widen what a revert of this PR would take with it.

Everything else — the fail-closed gating, the toggle pinning, the byte caps, the concurrency tests — looks right, and the local `go build`, `go vet`, `golangci-lint`, `go test -race` and shuffled runs are all clean.

tomglenn (approved):
Approving. Everything from the last round is addressed, and one of my suggestions was wrong — worth recording which.

**I had the cache fix backwards.** I said to drop the `delete` and let the generation bump force the refresh. That would have broken the feature: the bump only coalesces in-flight refreshes, it does not skip the TTL fast path, so the warm entry would have kept serving and a freshly written record would not have surfaced for up to five minutes — exactly what the invalidation exists to prevent. The `stale` flag is the right shape, and the comment explaining why is clearer than my finding was.

Confirmed on the rest:

- `completionWriterIdentity` now routes through `deriveCompletionUserID`, so reads and writes derive the join key from one place.
- The 403 comments now match the behaviour, with 401/403/408/429 carved out of the terminal class explicitly rather than by implication. The log line reading "disarms client, records retained" is the part someone debugging a grant rollout will actually see.
- The trust-boundary pointer moved to `BACKEND_PROXY_PATTERN.md` §3, and the test comment describing a profile-snapshot fallback that never existed is gone.
- `durationMs` clamps and logs instead of failing the write, so a producer bug costs one denormalized field rather than the whole completion.
- The 3xx read-path reclassification has a test, and the CI comment, `CUSTOM_GUIDES.md` paragraph and two `systemPatterns` rows are dropped, so a revert stays scoped.
- The `CONCERNS.md` anchor is substantive rather than a placeholder — identity contract, record-name derivation, the full retry taxonomy and the stale-not-delete rule are all stated where the next reviewer will be sent.

Verified locally on `79238008`: `go build ./...`, `go vet ./pkg/...`, `go test ./pkg/...`, and `go test -race` over the changed paths all pass. CI green with `Build` still pending at time of review.

The namespace-wide invalidation question — one user's writes dropping the shared index for every reader — is unchanged by this and stays in #1653; only the stale-serve half belonged here, and it landed.

If these go in together, this one first: its route is dark until #1434 arms writes, so that order never has a client calling a proxy that isn't deployed.

PR #1434, opened 2026-07-27
feat(completion-records): durable completion write hook and retry queue

moxious:
Part of the completion-records epic, #1411. Frontend half; the backend route this writes to is #1433.

## Summary

Today a finished guide is remembered only in the browser. Clear site data, switch machines, or open Grafana in another browser and the user's learning history is gone. This PR adds the frontend write path that makes those completions durable, so a completion survives the browser it happened in.

The shape is a background queue, not a blocking write. When a guide or learning journey reaches a terminal completion, the fact is normalised and persisted immediately to local storage, then drained to the backend on a timer. Recording never awaits the network and never throws into the completion path — if the write cannot happen now, the record waits rather than being lost.

What it adds, under `src/completion-records/`:

- **A single recorder boundary** (`completion-recorder.ts`) that every terminal completion flows through, plus a surface-neutral emitter (`recordGuideCompletionForSurface`) wired from the sidebar, floating panel and full-screen reader, so a guide finished in any surface records the same way.
- **A bounded, per-`(user, org)` queue** (`completion-write-queue.ts`, `completion-write-storage.ts`) in local storage, capped at 100 items with a 30-day retention horizon. Draining holds a cross-tab lease so several open Grafana tabs cannot double-send.
- **A POST client** (`completion-write-client.ts`) that never throws and classifies every result into a typed outcome: `created`, `transient`, `terminal`, `route-missing`, `forbidden`. Each queued event's stable id is sent as a required idempotency key, so a retry after a lease takeover dedupes to one durable record.
- **Degradation telemetry** (`pathfinder_completion_write_degraded`) reporting queue evictions, disarms and drops, in place of a latency measurement.

Two environment conditions get deliberate handling rather than a drop. A **404** means the write route is not served on this stack — expected during rollout, since the backend lands separately. A **403** means this identity holds no grant for the route yet. Both disarm the session's network draining but keep persisting, so records survive to a later load and drain once the route exists or the grant arrives. Neither destroys a completion.

Completion identity is keyed on `(repository, manifest.id)` read off the resolved manifest, never off a loader URL, with the explicit or resolved repository taking precedence over the manifest's schema-defaulted value — otherwise records key against the `interactive-tutorials` default and land in the wrong place. Whole-journey completion now requires every currently-expected milestone id rather than a bare count, so stale slugs from a renamed or reordered path cannot cross the threshold.

## Review round 2026-08-19

Six items from @tomglenn's review, all on code this PR introduces:

1. **"Reset all learning progress" now discards queued writes.** It cleared five stores but not the write queue, so anything already queued survived the reset and drained afterwards — minting durable records for exactly the guides the user just asked us to forget, with no second lever to stop it. Queued-but-unsent records are now discarded rather than merely stopped: the intent is unambiguous, and a record that has not left the browser is still the user's to withdraw. Closing this surfaced a related gap — a reset landing while a POST was in flight was undone by the transient re-persist, which now re-checks queue membership before writing back.

2. **The write stack is off the entry bundle.** A static import in `module.tsx` pulled the whole stack into `module.js`, paid on every page load of every Grafana with this plugin installed, including by users who never open Pathfinder. Arming now happens behind a dynamic `import()` in `plugin.init`. **`module.js`: 145,062 → 116,740 bytes (−28,322, −19.5%)**, measured by building with and without the static import; the stack contributes zero bytes to the entry chunk.

3. **`createWriteQueue` no longer calls `refresh()` from its factory body.** With arming on the bootstrap path that ran an origin-wide key scan, a `JSON.parse` per match, and un-leased `storage.remove()` calls on every page load. The first scheduled drain reconciles instead, under the lease — so no queue mutation happens outside it. This needed a companion fix: `start()` gated its initial drain on `queue.size() > 0`, which is now always 0 at start, so it schedules unconditionally or persisted records would never drain.

4. **The six contract anchors cited in comments and test names now resolve**, as a named-invariants table in `docs/design/CONCERNS.md` covering queue ownership, the lease protocol, the required idempotency key, the outcome taxonomy and the repository-identity precedence rule.

5. **An unresolvable identity is now visible.** Arming left the module-level controller assigned but inert, indistinguishable from a broken one at the earliest point in bootstrap. It now leaves the controller null and warns — which also fixes a latch, since an inert controller made every later arm a no-op even once an identity resolved.

6. **`CONTEXT_INDEX.md`** row path corrected to `.cursor/rules/systemPatterns.mdc`.

### Two behaviour changes worth a second look

Both came out of review on the round above rather than from Tom's list, and both change runtime behaviour, so calling them out explicitly:

- **A vanished lease now halts a drain** (`completion-write-storage.ts`). Fixing item 1 only bound the resetting tab: a second tab already inside `processDueWithLease()` kept POSTing from its own in-memory `items`, and because `renewLease()` treated an absent lease key as available, `clear()` sweeping the owner prefix neither stopped it nor preserved mutual exclusion for a third tab. `renewLease()` now fails when a lease this tab persisted has since vanished — takeover-or-clear either way — which bounds the cross-tab leak from `DRAIN_BUDGET_PER_PASS` (10) records to at most the one already in flight. This changes the lease contract, so the `queue-eviction-concurrency` invariant is reworded in the same commit.

- **`fetchPackageContent` falls back to its own resolved repository** (`package-content.ts`). It stamped `metadata.repository` only when a caller supplied one, so the same package recorded under two different durable `guideSource` values depending on entry path: a journey nav link keyed `('online-cdn', id)`, while My Learning → Discover More supplied a manifest with no repository and fell through to the schema default, keying `('interactive-tutorials', id)` — records for one guide split across two keys. This surfaced from the merge in Part 1: the Discover More launch path came in from main's #1638. The fallback is guarded on `.ok`, because `PackageResolutionFailure.repository` exists for negative-caching policy and is not an identity claim — app-platform is the composite's last tier and tags every probed miss with `repository: 'app-platform'`, which would have mis-attributed a public CDN path to the private App Platform repository.

## What to look at

- `src/completion-records/completion-write-queue.ts` — the state machine: lease, backoff, eviction, retention, disarm.
- `src/completion-records/completion-write-storage.ts` — `renewLease()`, and the `leasePersisted` flag that lets it tell a vanished lease from one never written.
- `src/completion-records/completion-write-hook.ts` — arming, the drain scheduler, and the discard entry point.
- `src/module.tsx` — the deferred arming, pinned by `module.bootstrap.test.ts` (armed once, from `init`, before any early return) and by a static-import ban in `entry-bundle-boundary.test.ts`.
- `docs/design/CONCERNS.md` — the named invariants. #1433 adds the top-level `completion-records` anchor row separately; both are meant to land.

## How to verify

- `npm run typecheck` — clean.
- `npm run lint` — 0 errors.
- `npm run test:ci` — 457 suites, 7316 tests, green.
- Bundle delta measured by building `module.js` with the static import and with the dynamic one: 145,062 → 116,740 bytes, unchanged by the follow-up fixes. The write stack lands in its own chunk.

Behavioural coverage for this round: reset leaves nothing to drain (including a POST already in flight, and now a second tab mid-drain); the queue neither scans nor mutates storage when constructed, and reconciles on the first drain; arming warns and stays re-armable when no identity resolves; a failed package resolution is not stamped as a repository.

## Breaking changes

None. No API, schema, or storage-format change to anything that already shipped; the queue's storage keys are new to this PR.

## Reversibility

Reversible. The feature is inert without the backend route — a stack without it disarms and retains rather than erroring, so this can ship ahead of #1433 without user-visible effect. Nothing here is a one-way door: no durable server-side state is created by the frontend alone, and the local queue is bounded and self-expiring.


tomglenn (changes requested):
## Review summary

Requesting changes on two defects. The rest of this is non-blocking.

The encouraging part first: the design here is sound and the hard parts are done well. The lease-plus-idempotency-key handling of the two-tab race is genuinely correct, and its test models a real lease rather than asserting on a mock. The 404-disarm-then-persist lifecycle test verifies durability against real `localStorage` rather than a stub. The tier model is respected — the new dependency runs downward and nothing in the Tier 1 module reaches upward. The shared `currentPlatform()` and fetch-error helpers were checked line by line and are behaviour-preserving (the context-service case is slightly more conservative than before, which is the right direction). CI is green across the board, and the reversibility section in the description is more honest than most.

The two blockers are both about wiring rather than about the queue machinery itself.

---

## Blocker 1: the write hook is never armed on any surface except the plugin's own app page

`armCompletionWriteHook()` is reachable only through `onPluginStart()`, and that has exactly one caller — the mount effect at `src/components/App/App.tsx:148`. `App` is registered with `plugin.setRootPage(App)` (`src/module.tsx:111`), so it mounts only when someone navigates to `/a/grafana-pathfinder-app`.

The extension sidebar is a separate registration (`src/module.tsx:299`) that renders `LazyContextPanel` and never mounts `App`. The same is true of the floating panel, the fullscreen panel, and the guide-reader overlay.

So in the normal flow — open Pathfinder in the sidebar from a dashboard, complete a guide — `recordGuideCompletion` fires into an empty listener set. Nothing is enqueued, nothing is persisted, no POST is attempted, and there is no log line at any level. The completion is lost rather than deferred.

It also cannot be recovered later in the same page load: `record()` inserts the dedupe key into `emitted` before notifying listeners, so if the user subsequently visits the app page and the hook arms, a re-record of the same fact is suppressed.

Worth noting the codebase already compensates for this exact gap elsewhere — `src/context-engine/context.service.ts:104` carries the comment "Ensure EchoSrv is initialized (fallback if onPluginStart wasn't called)". The new hook has no equivalent.

Suggested fix: arm from a surface-independent location that every entry point shares (module scope / `plugin.init`, or the panel activation handler), and add a test asserting the sidebar mount path arms it. The existing singleton guard in `armCompletionWriteHook` already makes multiple arming sites safe.

This also means the "How to verify" steps in the description pass if exercised from the app page and silently fail from the sidebar. Worth confirming which surface the manual verification actually used.

---

## Blocker 2: the durable key resolves to the wrong source on several entry paths

This is the bug that fix 1 in the description set out to eliminate, still reachable on a subset of paths.

`resolveCompletionIdentity` (`src/completion-records/completion-identity.ts:50-54`) checks `packageManifest.repository` **first** and only then falls back to the threaded `repository`. But `ManifestJsonObjectSchema` declares `repository: z.string().default('interactive-tutorials')` (`src/types/package.schema.ts:138`). Any manifest that has passed through that schema therefore arrives carrying a synthesised `'interactive-tutorials'` that is indistinguishable from a real value — and it outranks the correct threaded one.

The recommender path avoids this because `sanitizeV1PackageManifest` (`src/context-engine/context.service.ts:687`) strips `repository` out of the manifest and carries it as a sibling. That is precisely why the other paths are exposed.

Two confirmed leaks:

**1. `fetchPackageInfoFromUrl`** (`src/docs-retrieval/package-info-from-url.ts:62-70`) parses the fetched manifest through the schema and returns it as `packageManifest`, without setting `repository` on the `PackageOpenInfo`. This is the deep-link (`?doc=`), fullscreen-handoff and floating-to-fullscreen path.

**2. `resolvePackageNavLinks`** (`src/docs-retrieval/content-fetcher/package-content.ts:143-153`) receives a resolution that *does* know the true source — `src/package-engine/online-cdn-resolver.ts:69` literally sets `repository: 'online-cdn'`, and `recommender-resolver.ts:79` forwards the resolved value — but `ResolvedNavLink` has no field for it, so it is dropped. The four `openDocsPage(..., { packageId, packageManifest: link.manifest })` call sites in `src/components/docs-panel/context-panel.tsx` (around lines 619, 653, 905, 939) were not updated, so completions reached through the recommends, suggests and milestone links key on the manifest default.

Concrete result: guide `foo` served from `interactive-learning.grafana.net` completed after opening from a recommendation card writes `('online-cdn', 'foo')`. The same guide opened via a `?doc=` deep link writes `('interactive-tutorials', 'foo')`. Two `CompletionRecord` identities for one guide, and a revert stops new writes but leaves both behind — anything counting distinct completions double-counts.

There is also a precedence inconsistency inside a single function. In `markMilestoneDone` (`src/docs-retrieval/learning-journey-helpers.ts:531`) the milestone fact passes `repository: context?.repository ?? manifestGuideSource(...)` with no manifest, so the threaded value wins. The journey fact emitted from the same call (`:570`) passes the manifest, so `manifest.repository` wins. One user action, two facts, potentially two different sources. Latent today, but leak 1 above is already a producer that supplies both values with different contents.

Suggested fix: make the explicit `repository` argument take precedence over `manifest.repository`, or drop the `.default()` on that field and apply it only at routing/render sites rather than on the identity path. Then thread the value on the two paths above, and unify the precedence in `markMilestoneDone` (an explicit `overrideId` would express "milestone slug wins" better than withholding the manifest does).

---

## Why these were not caught: the threading is untested in the middle

The `repository` field crosses roughly six hops and only the first and last are covered.

No test sets `repository` on content metadata, and no test passes a fourth argument to `fetchPackageContent`, so the `...(repository !== undefined && { repository })` spread at `src/docs-retrieval/content-fetcher/package-content.ts:258` is unexercised.

The load-bearing detail is in `src/components/docs-panel/components/DocsPanelContentArea.test.tsx`: the new assertion uses `toHaveBeenCalledWith({ packageManifest, guideTitle })`, and Jest's object equality ignores keys whose value is `undefined`. So it passes against an actual containing `repository: undefined`. You can delete the threading from the component outright and the suite stays green.

Two tests would close this: one asserting `fetchPackageContent(url, manifest, undefined, 'app-platform')` surfaces `metadata.repository === 'app-platform'`, and one in the component test that sets `metadata.repository` and asserts the full context object strictly, so a dropped field fails.

---

## Non-blocking, but worth addressing before this feature carries real traffic

**No production observability on any failure path.** All five degradation paths use `logger.debug`, and the free Faro log bridge described in `docs/developer/TELEMETRY.md` covers `info`/`warn`/`error` only — so none of them surface. The startup eviction in `refresh()` has no log at all. If the backend began returning 403, `classifyWriteError` would map it to `terminal`, the queue would drop every record permanently, and the dashboard would show zero records — indistinguishable from zero usage. The telemetry doc's decision rule about degradation ladders needing a countable, alertable event applies directly here. One low-cardinality typed event with a `reason` field would cover it; the random ids and raw error strings currently being logged must not be lifted into Faro attributes.

**A capability endpoint already exists and is unused.** The backend on `main` already serves `GET /completion-records/capability` returning `{available, reason}` (`pkg/plugin/resources.go:22`). This PR instead probes by POSTing and treating a 404 as structural. On stacks where the route can never exist, that is a wasted failed request per load, plus a queue that fills to its 100-item cap and is never drained or cleared. Gating the arm on the existing probe would also give `CompletionWriteStorage.clear()` — currently with no production caller at all — somewhere to be called from.

**A drain pass has no time budget.** `processDueWithLease` walks every due item sequentially, renewing the lease before each send, with no elapsed-time check and no per-pass item cap. At the 100-item cap against a backend that hangs rather than refusing, one pass is 100 × 20s ≈ 33 minutes of continuously renewed lease, during which every other tab is locked out of `processDue` entirely. When it ends, all 100 items have fresh backoffs capped at five minutes, so it repeats — across every future page load. Only a 404 breaks the cycle. A per-pass wallclock budget that bails and returns the next delay would bound it.

**No attempt cap and no age limit.** `createdAt` is persisted but only ever read for sort ordering. A revert-then-reland would replay weeks-old completions with an original `completedAt` and a much later server-stamped `recordedAt`. One of the new tests currently locks the unbounded behaviour in as intentional.

**"Reset all learning progress" does not clear the queue.** `handleResetProgress` (`src/components/LearningPaths/MyLearningTab.tsx:355-381`) clears five stores but not `COMPLETION_WRITE_QUEUE_PREFIX`. A user told their completions were cleared can later have queued facts drain and mint records for guides they explicitly reset. Either call the queue's `clear()` there, or state in the dialog that durable records are intentionally immutable.

**Unresolvable identity arms into a state that can never recover.** When `ownerKey()` returns null (anonymous access, or `bootData.user` not yet populated), the queue is null and `start()` returns before subscribing — but the module-level `controller` is still assigned, and `armCompletionWriteHook` early-returns whenever it is set. No later call can fix it once the session has armed degraded. Leaving `controller` null in that branch would allow a retry, and a `warn` would make the state visible.

**Org switching orphans whole key prefixes.** The owner key is captured once at construction and `clear()` is scoped to that one prefix; nothing ever enumerates the base prefix. Each prefix is capped at 100 items, but the number of prefixes is not.

---

## Smaller items

- A queued entry whose stored `id` diverges from the `localStorage` key it was read from can never be removed: `list()` maps by `item.id` while `remove(id)` deletes `${itemPrefix}${id}`. It is re-parsed and re-sent on every drain and every page load indefinitely. Having `parseQueuedWrite` take the key and reject a mismatch would close it.
- `guideTitle`, `guideId`, `guideSource` and `pathId` reach the durable payload with no length bound. Custom-guide titles also skip the sanitiser that recommender titles get at `src/components/docs-panel/CustomGuidesSection.tsx:46`. Not an XSS path today — nothing renders these — but it is durable storage of arbitrary author-supplied markup, and 100 large titles is a real `localStorage` quota risk on an origin Grafana shares.
- The cross-repo contract exists only as TypeScript comments: `grep -i idempotency pkg/` returns nothing on this branch. If the backend ships the field as `idempotency_key`, the client keeps sending `idempotencyKey`, the backend falls back to a server-random name, and every retry after a lost response mints a duplicate — invisible from both ends. The frontend test asserts via the constant, so a rename stays green. A Go test pinning the JSON tag plus one shared derivation vector would make it real.
- The new dependency-table row at `.cursor/rules/systemPatterns.mdc:164` has producer and consumer swapped relative to the other twelve rows, so its "(Tier 2 → Tier 1, downward)" annotation reads as an upward violation. It is the only worked example of that annotation in a file agents read to learn the tier rules.
- `docs/developer/engines/context-engine.md:271` says the arm "never blocks startup", but `createWriteQueue` calls `refresh()` in its factory body, which walks every key in the origin and issues synchronous `removeItem` calls — outside the lease that `processDue` otherwise requires. Dropping the constructor `refresh()` and letting the first scheduled drain reconcile under the lease would fix the claim and the un-leased mutation together.
- The comment at `src/completion-records/completion-write-queue.ts:124-126` describes a mid-drain `enqueue()` → `refresh()` that no longer exists. The guard is still load-bearing, but for a different reason: `enqueue()`'s `items.shift()` eviction can drop an item that is in the current `due` list. Related: when that happens to an item already in flight and the send returns `transient`, line 164 re-`put`s it and resurrects an evicted record.
- `enqueue()` returns `boolean` but only ever returns the literal `true`, and the sole caller discards it — while a real failure mode does exist and is silent (`put()` falling back to the volatile map means the fact was not persisted, which is exactly what a boolean should report). Three tests assert `toBe(true)` against the constant.
- The `QueuedWrite` shape guard is duplicated between `isQueuedWrite` and `parseQueuedWrite` in two files, neither importing the other, and `refresh()` re-validates data that `storage.list()` has already validated and typed.
- The `-v2:` suffix on the storage key versions against a v1 that never reached `main`, so no migration is owed — worth either dropping or annotating.
- The 404 invariant is restated in seven places across five files. One statement on the `route-missing` member of `WriteOutcome` would carry it; the rest will drift.

---

## Test coverage notes

The branch behaviour in the `onGuideComplete` rewrite was enumerated against `main` and is **correct** — no case loses a progress write, the two branches are mutually exclusive so there is no double-emit, and the one emission delta (bundled journey with no manifest) is unreachable because that shape requires a path/journey manifest, which the previous code already early-returned on. The gap is coverage, not behaviour: four of six branch combinations are untested, including the two most common plain-bundled-guide paths, and the milestone test does not assert that `setJourneyCompletionPercentage` was *not* called — which is the double-emit the split exists to prevent.

Also uncovered: `refresh()`'s load-time eviction (no test seeds storage above `maxSize`, and this path permanently deletes records); the lease fail-open behaviour when `setItem` throws but `getItem` still works, which is the realistic full-storage case and where mutual exclusion silently disappears; the `timeout()` operator actually being applied to the request, as opposed to the constant relationship between the two timing values, which *is* asserted; and `subscribe()`'s key filter.

One deleted assertion is worth restoring. The case "does not emit for a non-bundled key at 100%" was replaced by one that calls a different function with a bundled key, so no test now passes a non-`bundled:` URL to `setJourneyCompletionPercentage` at 100%. That guard moved into `persistJourneyCompletionPercentage` during this refactor and is now unprotected — and `src/components/docs-panel/docs-panel.tsx:452` calls that function on every learning-journey tab load with a `grafana.com` URL as the key.

---

## How this review was produced

Six concern-scoped passes over the diff in an isolated worktree, plus local verification. Typecheck and lint are clean on all changed files; the targeted suites pass 709/710, where the single failure is a pre-existing environmental one unrelated to this branch (a test asserting `node:test` is a builtin, on Node 22 rather than the required 24). The architecture test passes 26/26.

Both blockers were confirmed by direct code reading and were each found independently by more than one pass. The lower-severity items above did not go through the same adversarial verification, so treat them as single-source and worth a second look before acting.

moxious:
## Open telemetry-policy question: the degradation signal is dropped on the drain-on-next-load path

Surfaced by review during the merge-with-main pass. **Deliberately not resolved in this PR** — the fix is a change to a shared telemetry boundary, which is a policy call (see `docs/developer/TELEMETRY.md`) rather than something this merge-only PR should decide. Filing it here so it is not lost.

### What happens

`passesActivityGate` (`src/lib/telemetry/filtering.ts:216`) unconditionally passes **only** exceptions and `error`-level logs; everything else is discarded until a Pathfinder surface has been open in the page load. Both halves of the completion-write degradation signal sit below that bar:

- the typed event `pathfinder_completion_write_degraded` is an **event item**, and
- all six call sites use `logger.warn` (queue lines 83, 92, 135, 184, 217, 223; hook lines 116, 128, 184).

Since `armCompletionWriteHook()` now runs as the first statement of `plugin.init`, `controller.start()` (`completion-write-hook.ts:199`, `:65`) fires `scheduleDrain(0)` when `this.queue.size() > 0` (`:72-73`). That means the drain of facts persisted by a **previous** session executes before any surface mounts or calls `reportPathfinderSurface`. At that moment `pathfinderWasOpen` is `false`, so both the event and the warn log are dropped. The event fires once, so a later surface open does not recover it.

**Net effect:** route-missing disarm, terminal-drop and expired-drop originating from the drains-on-next-load path never reach Faro — and that is the rollout-health signal the write client's docstring relies on in place of a latency measurement. This matters more than it looks right now, because the on-behalf-of credential is provisioned in dev only today, so the route-missing (404) disarm path is the common case in production.

Worth noting the existing coverage is not wrong, just narrower than the code now is: the gate test at `src/lib/faro.test.ts:1178` asserts the event clears the gate **from an open surface**, and its comment encodes the assumption that no longer holds universally — "a guide can only be completed from an open surface". True for the completion path; not true for the load-time drain path.

### The two options for a reviewer to choose between

1. **Exempt the event in the gate** — let `TELEMETRY_EVENTS.completionWriteDegraded` through `passesActivityGate` alongside exceptions and `error` logs, on the grounds that a degradation is a *health* signal rather than a *usage* signal. This is the earliest shared boundary that fixes both halves at once, but it widens what a collector session can mean.
2. **Promote the log level** — raise the disarm/terminal-drop/expired-drop classes from `logger.warn` to `logger.error`, which clears the existing gate with no change to shared filtering. Narrower blast radius, but it puts operationally-expected degradation at `error` severity.

Either way the gate's stated contract — collector sessions mean "used Pathfinder or Pathfinder errored", not "loaded a Grafana page" — is the thing being traded against, so it should be an explicit decision.

### Not addressed here either (follow-up grade, left unapplied on purpose)

Both touch code this PR deliberately froze:

- `sync-scan-on-bootstrap` — `createWriteQueue` calls `refresh()` synchronously in its constructor, so arming on `plugin.init` does an origin-wide `localStorage` enumeration plus a `JSON.parse` per stored item (up to 100) before init continues. The queue's own `enqueue` comment deliberately keeps that scan off the completion path for exactly this reason.
- `unused-storage-clear` — `CompletionWriteStorage.clear()` is declared and implemented but has no production caller.

Tracked under epic #1411.

tomglenn (approved):

tomglenn (changes requested):
Switching my approval to request-changes. The queue, lease and eviction logic all hold up — I read them against the tests and the earlier blockers are properly closed, including the AST-level pinning of the bootstrap wiring, which is a nice way to make that regression un-repeatable. Two things need to happen on this branch though, and both only exist because of the code this PR adds, so they can't reasonably be a follow-up.

## Please fix before merge

**1. "Reset all learning progress" doesn't clear the write queue** — `src/components/LearningPaths/MyLearningTab.tsx:238-268`

`handleResetProgress` clears five stores. The completion-write queue isn't one of them, so anything already queued survives the reset and drains afterwards — minting durable records for exactly the guides the user just asked us to forget.

The dialog tells them this clears their completed guides, badges and streaks. A user who resets and then watches those completions reappear has no way to understand what happened, and no second lever to pull. `CompletionWriteStorage.clear()` already exists with no production caller, so I think this is wiring rather than design.

Worth deciding explicitly whether reset should also discard queued-but-unsent facts or only stop future ones — I'd argue discard, since the user's intent is unambiguous and a record that hasn't left the browser yet is still theirs to withdraw.

**2. The entry bundle grows 27.8 KB (+24.7%)** — `src/module.tsx:27`

The static `import { armCompletionWriteHook } from './completion-records'` pulls the whole write stack into `module.js`: the queue at 13.91 KB, the hook at 9.25 KB, storage at 8.45 KB, the client at 6.52 KB, plus normalise/timing/telemetry. The bundle bot's per-module table has the breakdown.

That's paid on every page load of every Grafana with this plugin installed, including by users who never open Pathfinder, for a feature that runs entirely in the background. This file already carries a comment about keeping the telemetry barrel out of `module.js` for exactly this reason, so the precedent is set.

A dynamic `import()` inside `plugin.init` keeps it off the entry chunk. `src/module.bootstrap.test.ts` asserts a top-level call, so it would need to accept the deferred shape — but the property it's really protecting (armed once, from `init`, before any early return) survives that change.

**3. Related, and cheap while you're in there** — `src/completion-records/completion-write-queue.ts:99`

`createWriteQueue` calls `refresh()` from its factory body, which scans every key in the origin, does a `JSON.parse` per match, and issues un-leased `storage.remove()` calls for expired or over-cap items. Moving arming to `plugin.init` means that now runs on every page load rather than only on visits to the app page.

It's bounded, so this isn't urgent — but the first scheduled drain already reconciles under the lease, and it's the safer place to do it. Dropping the constructor call removes both the startup cost and the only mutation that happens outside the lease.

## Contract anchors that don't resolve

Six anchor names appear in comments and test names, and none of them exist anywhere in `docs/` or `.cursor/`:

- `payload-boundary-normalization` — `completion-write-normalize.ts:7`
- `disarmed-shared-queue-bound` — `completion-write-queue.ts:43`
- `queue-eviction-concurrency` — `completion-write-queue.ts:65`
- `completion-write-latency-signal` — `completion-write-client.ts:80`
- `repository-identity-authority` and `journey-threshold-membership` — used in test names

A "see `<name>`" that points nowhere costs the next reader real time before they conclude it isn't a typo on their end. Either add the rows or drop the parentheticals.

Relatedly, `docs/design/CONCERNS.md` isn't touched, though between this PR and #1433 you're establishing queue ownership, the lease protocol, the required idempotency key, the outcome taxonomy and the repository-identity precedence rule. The convention is that the anchor lands with the PR that establishes the contract, so the next change here gets reviewed against something.

## Minor

- When `currentCompletionQueueOwnerKey()` returns null, the queue is left null and `start()` returns, but `armCompletionWriteHook` still assigns the module-level controller, so every later call quietly early-returns. That's the right outcome for anonymous users, but arming now happens at the earliest point in bootstrap, and there's no log at any level to distinguish "correctly inert" from "silently broken". Leaving the controller null in that branch plus a warn would make it both visible and recoverable.
- The new `CONTEXT_INDEX.md` row gives its file as `systemPatterns.mdc (completion-records section)`. Every other row in that table is relative to `docs/developer`; this one actually lives at `.cursor/rules/systemPatterns.mdc`.

Everything else looks good — the identity precedence fix is threaded through cleanly, the milestone-membership change is equivalent to the old count for every current producer, and the degradation telemetry is a real improvement over the debug logging it replaces.

tomglenn (approved):
Approving. Both blockers and all four minors are addressed, and two of the fixes went further than what I raised.

**The reset fix caught a race I missed.** I flagged only that the queue was not cleared. Calling `discardQueuedCompletionWrites()` synchronously before the first `await` closes the window where a drain scheduled earlier fires mid-reset and mints records for exactly the guides being cleared — which would have been a confusing partial failure to debug, since the reset would look like it worked and then quietly undo itself.

**The bundle regression is genuinely gone**, not just moved: the bot reports +0.41%, down from +24.7%. `entry-bundle-boundary.test.ts` is the part that makes it stay fixed — a static import back into `completion-records` from `module.tsx` now fails a test rather than showing up as a number on a future PR that nobody reads.

The rest:

- `refresh()` moved from the queue constructor into `processDue()` under the acquired lease. That removes the bootstrap scan and the only mutation happening outside the lease, and the comment on the now-unconditional first drain explains why in-memory size says nothing about what a previous load persisted.
- The unresolvable-identity path leaves the controller null with a warn instead of assigning an inert one. The latching problem is the reason this matters — an assigned controller would have made every later arm early-return even once an identity resolved.
- `discardQueuedCompletionWrites` handling the not-yet-armed case is a good catch; the dynamic import introduced exactly that window.

On the six unresolved anchor names: introducing a documented "named invariants" concept in `CONCERNS.md` is a better answer than the one I suggested. I framed it as "add the rows or drop the parentheticals", which assumed the only two options were a full contract anchor or nothing. A named rule inside an anchored contract is the thing those comments were actually reaching for, and now `see <name>` resolves to a definition of what that citation means.

Verified locally on `51fa2f1b`: 216 tests across the completion-records, bootstrap, entry-bundle-boundary, My Learning and package-content suites pass. All CI green including the seven-image Playwright matrix. I also second-passed the one piece of new code beyond the review items — guarding the repository fallback on `baseUrlResolution?.ok` so a failed package resolution cannot supply a repository — and it is correct and covered.

Suggest landing #1433 first, since this arms writes against the proxy that ships there.

PR #1441, opened 2026-07-28
fix(e2e): align step execution with rendered controls

leslie-heinzen:
## Summary

The guide runner treated Do it as the only executable control, converted some mandatory requirement failures into successful skips, and trusted stale discovery metadata after navigation. This change executes the live rendered control, supports **Show me** for `doIt: false` steps, fails unmet mandatory requirements, and hardens guided target resolution against deployed `completeEarly`, hover-only controls, and Pathfinder’s `:nth-match(n)` selectors.

## What to look at

- [Step discovery](https://github.com/grafana/grafana-pathfinder-app/blob/7b9a226d43d9b930b9b269b9823fde5a33839268/tests/e2e-runner/utils/guide-runner/discovery.ts) — records **Show me** controls and preserves explicit skippability.
- [Step execution](https://github.com/grafana/grafana-pathfinder-app/blob/7b9a226d43d9b930b9b269b9823fde5a33839268/tests/e2e-runner/utils/guide-runner/execution.ts) — re-reads live controls, distinguishes optional and mandatory requirement outcomes, drives deployed guided substeps, translates `:nth-match(n)`, and reveals hover-only panel menus without force-clicking.
- [Interactive step behavior](https://github.com/grafana/grafana-pathfinder-app/blob/7b9a226d43d9b930b9b269b9823fde5a33839268/src/components/interactive-tutorial/interactive-step.tsx) — **Show me** completion reflects the underlying action result, and action errors remain observable.

## How to verify

1. Run a guide containing `doIt: false` steps and confirm each **Show me** action completes without a 15-second Do it wait.
2. Trigger a mandatory unmet requirement with no automatic fix and confirm the guide fails rather than reporting a successful skip.
3. Navigate away during an earlier step and confirm the next step re-reads its current controls instead of waiting on a stale locator.
4. Run a guided panel-edit sequence using a hover-only menu and confirm both guided substeps execute.
5. Run a guide using Pathfinder’s appended `:nth-match(n)` selector and confirm the intended occurrence is selected.

## Breaking changes

None. The change makes existing rendered controls authoritative and removes false-positive outcomes.

## Out of scope

- React guided-state precedence and the wide viewport are handled in the dependent PR #1442.
- Trace persistence and report-level safety are handled independently in PR #1443.

🤖 Generated with Warp


leslie-heinzen:
@cursoragent review

leslie-heinzen:
Addressed the valid review gaps in `7b9a226d`:

- Added a component regression proving a live-tab `doIt: false` step remains incomplete and exposes error state when Show me fails.
- Documented that live-tab Show me completion is outcome-gated, while controller-mode simple steps intentionally remain optimistic because the cross-tab channel has no live acknowledgement.

The review’s claim that controller completion became outcome-gated does not apply to this diff; that behavior remains unchanged. The additive error/Show me/mandatory-requirement contracts were already covered by the existing runner and contract suites.

Validation: 121 focused tests passed with 12 skipped, typecheck passed, and touched-file Prettier passed.

moxious (dismissed):
**Verdict: Approve.** No blocking issues. Typecheck clean (`tsc --noEmit`), and the three changed unit suites pass (121 tests). Two low-severity observations below, neither blocking.

I verified the one production file (`interactive-step.tsx`) end-to-end against the surrounding code and confirmed ground truth on the branch; the rest is e2e-runner infra with solid unit coverage of the new pure helpers.

### Low-severity observations

**1. `deriveInteractiveStepState` orders `hasError` before `isCompleted` (`interactive-step.tsx:76`)**
Both error DOM elements are suppressed when `isCompletedWithObjectives` (lines ~1137, ~1147), but the state derivation checks `hasError` before `isCompleted`. A step completed via objectives that still carries a stale error would report `data-test-step-state="error"` with no error DOM present, and the runner's `waitForCompletionWithObjectivePolling` would throw "entered error state". In practice this is effectively unreachable — the runner sees `error` on the first poll before objectives can win the race — so it's cosmetic. Optional one-line fix: check `isCompleted` before `hasError` to keep the attribute consistent with the render's completion-suppression.

**2. Orphaned helpers (`execution.ts:99,120`)**
`waitForDoItButtonToAppear` / `waitForDoItButtonEnabled` lose all internal callers after this refactor (still re-exported from `index.ts`, so no lint error). Dead code in test infra — safe to delete or leave.

### What I checked that could have been bugs, but aren't
- **`handleShowAction` outcome guard:** switching from `!result.elementFound` to `result.outcome === 'error'` still catches element-not-found, since `executeWithLazyScroll` returns `outcome: 'error'` in that case. No regression; the new "show-only failure" test exercises exactly this path.
- **`executeInteractiveAction` outcome:** for non-sequence actions it always returns `'ok'` (errors are caught/swallowed), so the outcome only fully reflects failures for `sequence`/element-not-found — but this is strictly better than the old always-`true`, not a defect introduced here.
- **Error clearing:** both handlers clear `postVerifyError`/`lazyScrollError` at entry, so a successful retry lands with errors null and `completed` reports correctly.
- **Mandatory-fail path:** `determineUnmetRequirementOutcome` only fires on `status === 'unmet'`; the sequential-dependency "disabled button, no explanation" case is `status === 'unknown'` and correctly falls through to the action-wait logic — not over-aggressive.
- **`completeEarly`/guided semantics:** `completed` with a non-null substep index is consistently treated as still-active across all three call sites; terminal only when `substepIndex === null`.

### Boundaries / assumptions
- The Playwright-only integration paths (`resolveGuidedTarget`, `revealGuidedTarget` hover→panel-menu escalation, `guidedSelectorLocator`) are domain heuristics that can't be unit-driven; reviewed for logic soundness only.
- The runner is now materially stricter (guided steps no longer auto-skippable; mandatory unmet → fail; no-action mandatory → throw). This is the stated intent and may surface real gaps in existing guides — expected, not a defect.

🤖 Reviewed with Claude Code

moxious (approved):

PR #1442, opened 2026-07-28
fix(e2e): stabilize guided UI state and viewport

leslie-heinzen:
## Summary

`completeEarly` guided steps persisted completion before their subactions finished, causing the DOM contract to report completed while execution was still active. This dependent change keeps guided execution observable until subactions settle and uses a 1920×1080 desktop viewport so Grafana editor controls remain available beside Pathfinder.

## What to look at

- [Guided state derivation](https://github.com/grafana/grafana-pathfinder-app/blob/f34aa112bcc527e14320cc6a0261017c680d71f4/src/components/interactive-tutorial/interactive-guided.tsx) — active execution takes precedence over persisted completion, while final completion remains navigation-safe.
- [Playwright viewport](https://github.com/grafana/grafana-pathfinder-app/blob/f34aa112bcc527e14320cc6a0261017c680d71f4/tests/e2e-runner/playwright.config.ts) — the supported desktop viewport prevents responsive Grafana editor controls from being hidden by the open Pathfinder panel.

## How to verify

1. Run a `completeEarly` guided panel-edit sequence and confirm the step exposes executing state while each substep is active.
2. Confirm the step transitions to completed only after the guided actions settle.
3. Open a panel editor with Pathfinder visible and confirm Back to dashboard remains visible.
4. Complete a final Back/exit-edit guided sequence and confirm the dashboard URL no longer contains `editPanel`.

## Breaking changes

None. The persisted completion model is unchanged; only the observable in-progress state and E2E viewport are corrected.

## Out of scope

- Runner-side guided selector and control execution are provided by the base PR #1441.
- Artifact and report safety are handled independently in PR #1443.

🤖 Generated with Warp


leslie-heinzen:
@cursoragent review

src/components/interactive-tutorial/interactive-guided.tsx
```diff
@@ -118,6 +118,34 @@ interface InteractiveGuidedProps {
...
+interface GuidedUiStateInput {
+  isCompleted: boolean;
+  isExecuting: boolean;
+  hasError: boolean;
+  wasCancelled: boolean;
+  isChecking: boolean;
+  isEnabled: boolean;
+}
+
+export function deriveGuidedUiState(input: GuidedUiStateInput): StepStateValue {
```
Jayclifford345, line 131: `[suggestion]` The derivation is right, but the invariant the PR describes — "active execution takes precedence over persisted completion" — isn't actually established for the `completeEarly` path. It's established ~50ms late.

`executeStep` persists *and then yields* before flipping `isExecuting`:

```ts
if (completeEarly) {
  persistCompletion();
  onStepComplete?.(stepId);
  onComplete?.();
  await new Promise((resolve) => setTimeout(resolve, 50));  // <- window
}
setIsExecuting(true);
```

Inside that window the store write has landed, so the step renders `data-test-step-state="completed"` with `data-test-substep-index` absent — exactly the shape `waitForGuidedExecutionStart` (`execution.ts:289`) and `runGuidedSubstepLoop` (`state === 'completed' && indexStr === null`) treat as "already finished".

The runner doesn't hit it today because `POST_CLICK_SETTLE_DELAY_MS` (500ms) elapses before the first poll, leaving a ~450ms margin. That margin is a timing coincidence, not a contract: `markStepCompleted` fans a store notify out to every sibling step in the section plus a `triggerReactiveCheck`, which is precisely the work that gets slow on a loaded CI container.

Hoisting `setIsExecuting(true)` above the `if (completeEarly)` block closes it deterministically, and tightens the re-entry guard at the top of `executeStep` as a bonus. `interactive-multi-step.tsx` has the same shape.

The new `completeEarly lifecycle` test can't catch this — it asserts state only after `mockExecuteGuidedStep` has been called, i.e. after `setIsExecuting(true)`.

src/components/interactive-tutorial/interactive-guided.tsx
```diff
@@ -118,6 +118,34 @@ interface InteractiveGuidedProps {
...
+  if (input.isExecuting) {
+    return STEP_STATES.EXECUTING;
+  }
+  if (input.hasError) {
+    return STEP_STATES.ERROR;
+  }
+  if (input.wasCancelled) {
+    return STEP_STATES.CANCELLED;
```
Jayclifford345, line 142: `[question]` Is this branch reachable? `setWasCancelled` is called at five sites in this file and every one passes `false`. The cancelled outcome in `executeStep` returns without touching it:

```ts
} else if (result === 'cancelled') {
  return false;
}
```

So `uiState === 'cancelled'` looks dead, and with it:

- the `STATE: CANCELLED` / "Tour cancelled" block (~line 1003) — `handleCancel` drops the step back to `idle` instead
- the two new `deriveGuidedUiState` rows asserting cancellation precedence: they pass, but they pin behavior no code path can produce
- the runner's cancelled diagnostics from #1441 (`waitForSubstepAdvance`, `runGuidedSubstepLoop`, `waitForGuidedExecutionStart`) — a cancel surfaces as `Unexpected guided step state: idle` or a substep timeout, never "Guided step was cancelled"

Either add `setWasCancelled(true)` to the `result === 'cancelled'` branch (which makes the runner's existing throws work, and is probably what was intended), or drop the branch, its tests, and the render block. Pre-existing, but this PR is the one formalizing the ordering, so it's the natural place to settle it.

src/components/interactive-tutorial/interactive-guided.tsx
```diff
@@ -667,8 +695,11 @@ export const InteractiveGuided = forwardRef<{ executeStep: () => Promise<boolean
       }
     }, [disabled, isExecuting, stepId, onStepReset, persistReset]);
 
-    // Handle skip current step on timeout
     const handleSkipStep = useCallback(async () => {
```
Jayclifford345, line 698: `[question]` This handler and the multi-step error-skip button now do different things for the same user gesture:

- multi-step (`interactive-multi-step.tsx:981`): `await checker.markSkipped()` → `setExecutionError(null)` → `persistCompletion('skipped')`
- guided (here): no `checker.markSkipped()`

`persistCompletion` is a no-op unless `isStandalone` (`!onStepComplete`), so for a **section-managed** guided step — how real guides run — the `'skipped'` reason is dropped and the section's `handleStepComplete` writes `markStepCompleted(stepId, sectionId, 'manual')` instead. Same click, different `pathfinder:progress` reason depending on component and standalone-ness, and the checker FSM never learns the step was skipped.

The new `skip recovery` test uses a harness with no `onStepComplete`, so it only exercises the standalone branch. Worth a section-managed case too, since that's the path where clearing `executionError` here hands responsibility for reaching `completed` entirely to the parent.

src/components/interactive-tutorial/interactive-guided.tsx
```diff
@@ -667,8 +695,11 @@ export const InteractiveGuided = forwardRef<{ executeStep: () => Promise<boolean
...
+      setExecutionError(null);
+      setFailedStepIndex(-1);
+      setCurrentStepStatus('completed');
```
Jayclifford345: `[nit]` These two look inert. `currentStepStatus` is only read inside the `executing` block and as the error-detail fallback, both of which are gone once `executionError` is cleared on the line above; `wasCancelled` is never `true` (see the other comment). The two lines doing real work are `setExecutionError(null)` and `setFailedStepIndex(-1)`.

src/components/interactive-tutorial/interactive-multi-step.tsx
```diff
@@ -65,6 +65,30 @@ interface InteractiveMultiStepProps {
...
 
+interface MultiStepUiStateInput {
+  isCompleted: boolean;
+  isExecuting: boolean;
+  hasError: boolean;
+  isChecking: boolean;
+  isEnabled: boolean;
+}
+
+export function deriveMultiStepUiState(input: MultiStepUiStateInput): StepStateValue {
```
Jayclifford345, line 78: `[suggestion]` `deriveMultiStepUiState` and `deriveInteractiveStepState` (`interactive-step.tsx:51`) are now identical, and `deriveGuidedUiState` is the same function plus one `cancelled` branch. Three copies of a state-precedence rule in three files is the drift this refactor is trying to eliminate. One `deriveStepUiState(input)` in `step-states.ts` with an optional `wasCancelled` would make the ordering impossible to diverge per component later.

src/components/interactive-tutorial/interactive-guided.test.tsx
```diff
@@ -227,6 +234,112 @@ describe('InteractiveGuided — double skip button (issue #786)', () => {
...
+    ['reports settled completion', { isCompleted: true }, 'completed'],
+    ['reports requirement checks', { isChecking: true }, 'checking'],
+    ['reports idle when enabled', {}, 'idle'],
+    ['reports unmet requirements when disabled', { isEnabled: false }, 'requirements-unmet'],
+  ])('%s', (_name, overrides, expected) => {
+    expect(deriveGuidedUiState({ ...baseState, ...overrides })).toBe(expected);
+  });
+});
+
+describe('InteractiveGuided — completeEarly lifecycle', () => {
```
Jayclifford345, line 285: `[suggestion]` Neither new describe resets `mockExecuteGuidedStep`, and `jest.config.js` sets no `clearMocks`/`resetMocks`, so implementations leak forward: the never-resolving promise set here survives into `skip recovery` (which happens to override it), and that describe's `mockResolvedValue('timeout')` then survives into `AI "Fix this" gating` and everything after it.

Harmless today because the later describes don't execute guided steps, but it makes the file order-dependent. A `mockExecuteGuidedStep.mockReset()` alongside the existing `mockStoredCompleted` reset at line 149 would cover it.

tests/e2e-runner/playwright.config.ts
```diff
@@ -47,6 +47,7 @@ export default defineConfig<PluginOptions>({
       name: 'chromium',
       use: {
         ...devices['Desktop Chrome'],
+        viewport: { width: 1920, height: 1080 },
```
Jayclifford345, line 50: `[question]` Two things:

1. This is the runner config only — the root `playwright.config.ts` still uses bare `devices['Desktop Chrome']` (1280×720) for the main suite, which also drives Pathfinder with panel editors open. If the responsive-overflow failure ("Back to dashboard" hidden behind the panel) is real, the main suite still has it, and the two suites now resolve selectors at different widths. Intentional, or should the root config match?
2. 1920×1080 is ~2.25x the pixels per screenshot. Worth confirming against the artifact budget in #1443, since guided runs capture pre/post plus loop diagnostics.

docs/developer/E2E_TESTING_CONTRACT.md
```diff
@@ -353,23 +353,27 @@ This ensures consistency between:
...
+            ? 'checking'
+            : !isEnabled
+              ? 'requirements-unmet'
+              : 'idle'
   }
   data-test-substep-index={isExecuting ? currentIndex : undefined}
 />
 ```
 
+For guided and multi-step components, `executing` takes precedence over `completed`. This keeps the active substep observable when `completeEarly` persists completion before the remaining actions settle.
```
Jayclifford345, line 375: `[nit]` Two small inaccuracies now that #1441 has landed:

- The snippet was retitled to `interactive-guided.tsx / interactive-multi-step.tsx` and this sentence scopes the rule to those two — but `deriveInteractiveStepState` in `interactive-step.tsx` uses the same `executing → error → completed` order. As written, the doc implies a divergence that doesn't exist. Naming all three (and noting `cancelled` is guided-only) matches the code.
- With `executing` now outranking `completed`, `state === 'completed' && substepIndex !== null` can no longer occur, so the `isGuidedExecutionActive` branch handling it (`execution.ts:322`) is dead — it was the workaround for the old precedence. Better retired here than left as a second, competing model of "is this guided step running".

Jayclifford345 (commented):
## Review: `fix(e2e): stabilize guided UI state and viewport`

**Disposition: approve with minor.** The core move is right — hoisting three inline state ternaries into tested pure functions, and making `interactive-multi-step` render its blocks *from the same `uiState` it publishes*, removes a real class of bug (the attribute and the rendered block could previously disagree). Nothing here is a one-way door; every finding below is a tweak or a straight revert.

Verified locally on `58944ee`:

- `npx jest src/components/interactive-tutorial/{interactive-guided,data-attributes.contract}.test.tsx` → 72 passed, 12 skipped
- `npm run typecheck` → one failure in `src/cli/__tests__/schema-e2e-report.test.ts` (`Cannot find module 'ajv/dist/2020'`), pre-existing from #1372 and unrelated to this diff

### Findings

| # | Prefix | Finding | Where |
| - | ------ | ------- | ----- |
| 1 | `[suggestion]` | The `executing`-over-`completed` invariant is established ~50ms *late* for `completeEarly`; only `POST_CLICK_SETTLE_DELAY_MS` hides the window | `interactive-guided.tsx:130` |
| 2 | `[question]` | `wasCancelled` is never set to `true` — the `cancelled` branch, its two new test rows, and the runner's cancelled diagnostics are all unreachable | `interactive-guided.tsx:138` |
| 3 | `[question]` | Guided error-skip and multi-step error-skip now diverge on `checker.markSkipped()` and on the persisted progress reason; the new test only covers the standalone path | `interactive-guided.tsx:698` |
| 4 | `[nit]` | Two of the four new lines in `handleSkipStep` are inert | `interactive-guided.tsx:701` |
| 5 | `[suggestion]` | Three near-identical `derive*State` functions across three components | `interactive-multi-step.tsx:76` |
| 6 | `[suggestion]` | New mock implementations leak forward across describes | `interactive-guided.test.tsx:269` |
| 7 | `[question]` | Viewport bumped only in the runner config; the root `playwright.config.ts` still runs the main suite at 1280×720 | `playwright.config.ts:50` |
| 8 | `[nit]` | Contract doc now implies `interactive-step.tsx` uses a different order than it does; the runner's `completed && substepIndex !== null` workaround is now dead | `E2E_TESTING_CONTRACT.md:375` |

### Things I checked and cleared

- **`uiState === IDLE` tightening in multi-step.** The old idle gate was `... && checker.isEnabled && !executionError` and the new one additionally requires `!isChecking`. Not a regression: `createCheckingState` sets `isEnabled: false`, so `isEnabled && !isChecking ≡ isEnabled`. The "Do it" button already vanished during rechecks.
- **`REQUIREMENTS_UNMET` losing the explicit `completionReason !== 'objectives'` guard.** Preserved transitively — `isCompletedWithObjectives` already folds in `completionReason === 'objectives'`, and the `completed` branch short-circuits first.
- **`executing` outranking `error`.** Not reachable as a rendered state in either component today: every `setExecutionError` is followed by `return false` into a `finally { setIsExecuting(false) }` in the same batch. Fine as a latent ordering choice; the docs and tests pin it.
- **Error block no longer suppressed while `checker.isChecking`.** Behavior change (the old gate was `executionError && !checker.isChecking`), and an improvement — a background recheck no longer blanks the error box out from under the user.

**Tech debt:** items 4, 5, and 8 above. No other high-confidence patterns in the changed files.

**Instrumentation:** no new user-facing action, async budget, or fallback ladder — existing `DoItButtonClick` / `StepAutoCompleted` coverage is unchanged. No gaps.

**Doc drift:** `E2E_TESTING.md` and `E2E_TESTING_CONTRACT.md` are both updated in-PR (good); see item 8 for the one inaccuracy. No `AGENTS.md` / `.cursor/rules/` updates needed.

Jayclifford345:
### Adjudicating the three Bugbot findings

Two are stale (fixed by later commits in this same PR), one is real and I missed it. Details so nobody re-litigates them:

**❌ Stale — "Multi-step UI ignores derived state" (High→n/a).** Reviewed `90cd9224`. `a2ebe65c` ("render multi-step state consistently") is precisely the fix: the COMPLETED / EXECUTING / REQUIREMENTS_UNMET / ERROR blocks are all gated on `uiState === …` now, so they're mutually exclusive. The only remaining independent gate is the `' executing'` CSS class, which keys off the `isCurrentlyExecuting` **prop** (section-driven), not local `isExecuting` — cosmetic, and unchanged by this PR.

**❌ Stale — "Skip stuck after error precedence" (High→n/a).** Reviewed `a2ebe65c`. `58944ee7` ("clear errors when skipping failed steps") adds exactly the `setExecutionError(null)` / `setFailedStepIndex(-1)` the bot asked for, to both skip handlers.

**✅ Real — "Retry fails after early completion" (medium).** Reviewed head; confirmed, and it's a good catch I missed. Concretely:

1. A `completeEarly` guided step persists completion, then substep *i* fails → `executionError` set, `isExecuting` false, `storedCompleted` already true.
2. Pre-PR, `completed` outranked `error`, so the error UI was unreachable here — the step just showed "Completed".
3. Post-PR it correctly shows the error box. But `handleRetry` ([interactive-guided.tsx:715](https://github.com/grafana/grafana-pathfinder-app/blob/58944ee78c0847782f63b99cfdbc06d8538d8906/src/components/interactive-tutorial/interactive-guided.tsx#L715)) only clears `executionError` and calls `executeStep`, which bails immediately on its entry guard at [line 318](https://github.com/grafana/grafana-pathfinder-app/blob/58944ee78c0847782f63b99cfdbc06d8538d8906/src/components/interactive-tutorial/interactive-guided.tsx#L318) (`isCompletedWithObjectives` is still true — nothing reset the store).
4. Net: "Try again" clears the error, reruns nothing, and the step settles back to `completed`. The failed action never happens and the user gets no signal.

So the PR makes a state reachable whose primary recovery action is a silent no-op. Same shape in multi-step: the error block's retry ([interactive-multi-step.tsx:949](https://github.com/grafana/grafana-pathfinder-app/blob/58944ee78c0847782f63b99cfdbc06d8538d8906/src/components/interactive-tutorial/interactive-multi-step.tsx#L949)) hits the identical guard at [line 310](https://github.com/grafana/grafana-pathfinder-app/blob/58944ee78c0847782f63b99cfdbc06d8538d8906/src/components/interactive-tutorial/interactive-multi-step.tsx#L310).

`handleStepRedo` already models the right thing — it calls `persistReset()` before clearing local state. Adding `persistReset()` to both retry paths would fix it, and it pairs naturally with hoisting `setIsExecuting(true)` above the `completeEarly` block (finding 1 in my review): both are "the `completeEarly` early-persist leaks into paths that assume completion means done."

Worth a unit test alongside the new `completeEarly lifecycle` one: fail a substep on a `completeEarly` step, click Try again, assert `executeGuidedStep` is called a second time.

This bumps my disposition from *approve with minor* to **request changes** — not for the derivation itself, which is right, but because the newly-reachable error state has a dead recovery button.

leslie-heinzen:
## Review feedback addressed

Thanks for the thorough reviews. I addressed the actionable findings from Jay's review and the subsequent Cursor Bugbot rounds.

### Jay review

- Established `executing` before `completeEarly` persistence in guided and multi-step flows, closing the transient completed-state window.
- Made guided cancellation reachable and observable through the DOM contract.
- Routed guided error-state Skip through `checker.markSkipped()` so section-managed steps retain the `skipped` progress reason.
- Removed inert skip-state updates and reset leaked mock implementations between test cases.
- Removed the obsolete runner compatibility path for `completed` with a live substep index.
- Clarified the intentional state-ordering difference: `InteractiveStep` lets settled completion beat stale local errors, while composite steps surface live errors before completion.
- Aligned both Playwright configs to the same 1920×1080 viewport.

I did not consolidate the three state helpers because their precedence is intentionally different for single-step versus composite rendering.

### Cursor Bugbot

- Made every multi-step status region render exclusively from the derived `uiState`, preventing completed/Redo controls from appearing during active execution.
- Cleared transient guided and multi-step errors before skipped completion so recovery reaches the completed UI.
- Added standalone and section-managed skip recovery coverage.
- Allowed Retry to rerun failed guided and multi-step actions when `completeEarly` already persisted completion.
- Added regression coverage proving a failed `completeEarly` guided action executes again before returning to completed.

### Validation

- Focused suites: 4 passed, 149 tests passed, 12 skipped.
- Full frontend coverage: 368 suites passed, 5,777 tests passed.
- Typecheck and Prettier passed.
- ESLint: 0 errors; 24 pre-existing deprecation warnings.

moxious (dismissed):
## Review summary

The core change is sound and well-verified. Reordering `executeStep` so `setIsExecuting(true)` + `waitForReactUpdates()` fire **before** `persistCompletion()`, and inverting the derivation so `executing` beats `completed`, makes `data-test-step-state` monotonic-safe — any committed render where `isCompleted` is true still has `isExecuting` true, so it renders `executing`, never a transient `completed`. That lets the runner treat any `completed` as terminal and drop the `isGuidedExecutionActive` hack (all references cleaned up). The issue #786 double-skip fix is preserved, and the change is fully reversible (producer, consumer, and contract doc move atomically; `persistCompletion` args and storage keys unchanged). No security surface touched.

Requesting changes on two items — neither a data-safety one-way-door, but both worth fixing before merge given the PR's stated purpose is E2E/UI state consistency. A single fix resolves the correctness finding; adding multi-step component tests closes the coverage gap that would have caught it.

---

### 🟠 Medium — completed-via-objectives step reports `data-test-step-state='error'` (or `cancelled`)

The precedence inversion puts `error`/`cancelled` above `completed` in `deriveGuidedUiState`/`deriveMultiStepUiState`, but nothing clears the component-local `executionError`/`wasCancelled` when a step completes via the **objectives auto-completion** path (`step-checker.hook.ts` ~L884-890), which runs independently of `executeStep`.

**Scenario:** a non-`completeEarly` guided/multi-step step carries an `objectives` requirement _and_ internal actions. An action returns `'error'` → `executionError` set, `isExecuting=false`, `persistCompletion` not called, so the step stays eligible for objectives polling. The user then satisfies the objective out-of-band → `completionReason='objectives'` → `isCompletedWithObjectives` true and completion persisted — but `executionError` is still set, so the derive returns `error`. Pre-PR (completed-first) this showed `completed`.

**In-file confirmation:** multi-step's `getButtonTitle` still uses the _old_ completion-first ordering ("Already done!"), so in this state the button title says completed while `data-test-step-state='error'` — the two disagree.

Bounding factors: persisted state is **not** corrupted (`writeStoreCompletion('objectives')` is canonical), it self-heals on remount (`executionError` is `useState(null)`), and the E2E runner never satisfies objectives out-of-band. But a genuinely-completed step reporting `error` is a real regression for this PR's purpose.

**Suggested fix:** clear `executionError`/`wasCancelled` when `completionReason === 'objectives'` (an effect, mirroring the existing objectives-completion effect), or give objectives-completion precedence in the derive — so `completeEarly`-then-error still shows `error` (intended, retry stays reachable) but genuine objectives completion resolves to `completed`. This also realigns `getButtonTitle`.

### 🟠 Medium — `InteractiveMultiStep`'s changed behaviors have no component test

Multi-step received the same three runtime changes as guided — the `allowCompletedRetryRef` retry-bypass guard, the `completeEarly` reorder, and the skip handler clearing `executionError`/`failedStepIndex` — but its only new coverage is the pure `deriveMultiStepUiState` table. Guided got full component coverage (executing-during-completeEarly, completeEarly-retry, skip-recovery); multi-step's identical logic is untested at the component level, so a future regression there ships green. Notably, the multi-step manifestation of the finding above is exactly what a component test would surface. **Please port the guided completeEarly/retry/skip component tests to multi-step.**

---

### 🟡 Low — guided `handleSkipStep` clears error UI before awaiting `markSkipped`

`handleSkipStep` clears `executionError`/`wasCancelled` and _then_ `await markSkipped?.()` before `persistCompletion('skipped')`. `markSkipped` (`step-checker.hook.ts` L782) is a synchronous void callback, so the practical risk is negligible — but it is asymmetric with multi-step's skip handler, which clears error _after_ the await. Optional: match multi-step's ordering.

### 🟡 Low — `cancelled`-beats-`completed` only asserted at the pure-function level

The component cancellation test uses a non-`completeEarly` step (`isCompleted=false`), so it proves cancelled-beats-idle, not cancelled-beats-completed. Low, since derivation is centralized in the tested pure function.

---

### Non-blocking notes

- **Nit:** the `waitForReactUpdates()` relocation dropped the load-bearing comment explaining the issue #786 ordering. The fix still holds, but the comment documented counterintuitive-but-correct ordering — consider restoring a one-liner.
- **Tech debt (non-blocking):** three near-duplicate derivation functions now exist with intentionally divergent precedence (InteractiveStep: `completed` before `error`; guided/multi: `error` before `completed`). Leave as-is, but it's a subtlety future editors must respect — fixing the medium above would make the "except genuine objectives" rule explicit.
- **Docs/instrumentation:** contract-doc updates are accurate; no telemetry gaps.

_Automated Principal-level review (concern-routed + 3-skeptic adversarial verification on medium findings). The core fix verified clean; findings above are the residue._

leslie-heinzen:
## Latest review feedback addressed

Addressed the findings from review `4809512891` in commit `7e2d4098`.

- Made genuine objectives completion authoritative over stale guided or multi-step error/cancellation state after execution settles.
- Kept `completeEarly` failures recoverable: a persisted early-completion write alone does not suppress Retry or Skip.
- Matched guided Skip ordering to multi-step and restored the execution-order invariant comment.
- Strengthened guided cancellation coverage with `completeEarly` and added objectives-after-error component coverage.
- Added a dedicated `InteractiveMultiStep` component suite covering early execution visibility, completeEarly Retry, Skip recovery, and objectives completion after a stale error.

Validation:

- 5 focused suites passed: 156 tests.
- Full coverage passed: 369 suites, 5,784 tests.
- Typecheck, Prettier, and docs sync passed.
- ESLint reported 0 errors and 24 pre-existing warnings.

Ready for re-review.

Jayclifford345:
## Re-review: 🟢 greenlight

Re-reviewed at `7e2d409`. Every finding from both prior rounds is addressed, and the two mediums are fixed in the shape that was asked for rather than papered over. Good to merge from my side.

### Verified locally on `7e2d409`

- `npx jest src/components/interactive-tutorial/ --coverage=false` → **24 suites, 368 passed**, 12 skipped, 5 todo
- `npm run typecheck` → **clean** (the `ajv/dist/2020` failure I hit on `58944ee` is gone after the `main` merge — it was pre-existing and unrelated)
- All 37 CI checks green, including all six `e2e test` matrix legs and Cursor Bugbot

### Findings closed

| Round | Finding | Resolution |
| ----- | ------- | ---------- |
| @moxious 🟠 | objectives-completed step reported `error`/`cancelled` | `isCompletedByObjectives` now sits directly under `isExecuting` in both derives |
| @moxious 🟠 | no multi-step component coverage | new `interactive-multi-step.test.tsx` — completeEarly executing window, retry-after-early-completion, skip-clears-error, objectives-over-stale-error |
| @moxious 🟡 | guided skip cleared error before `await markSkipped` | now awaits first, matching multi-step |
| @moxious 🟡 | cancelled-beats-completed only asserted on the pure function | component test now uses `completeEarly={true}`, so `isCompleted` is genuinely true |
| #1 | `executing` invariant established ~50ms late | `setIsExecuting(true)` + `waitForReactUpdates()` now precede `persistCompletion()` |
| #2 | `wasCancelled` never set `true` | set on both the `'cancelled'` action result and `handleCancel`; the `uiState === 'cancelled'` block at `interactive-guided.tsx:1009` is now reachable and tested |
| #3 | guided/multi-step skip divergence | guided now awaits `checker.markSkipped` before `persistCompletion('skipped')` — same pair as multi-step |
| #4 | inert lines in `handleSkipStep` | load-bearing now that `error` outranks `completed`, and pinned by the skip-recovery test |
| #6 | mock implementations leaking across describes | top-level `beforeEach` resets `mockExecuteGuidedStep`, `mockMarkSkipped`, `mockStoredCompleted`, `mockCompletionReason` |
| #7 | viewport bumped only in the runner config | root `playwright.config.ts:52` now sets 1920×1080 too, and `E2E_TESTING.md` documents both |
| #8 | contract doc inaccuracy + dead runner workaround | doc now names guided/multi-step and describes `InteractiveStep` separately (matches `deriveInteractiveStepState`); `isGuidedExecutionActive` deleted along with its barrel export and test |

### What I checked on the objectives fix specifically

The load-bearing claim in the contract doc — *"a `completeEarly` write alone does not receive this exception"* — holds, and I traced why rather than taking it on faith:

- `persistCompletion()` defaults to `reason: 'manual'` (`interactive-guided.tsx:231`), so a `completeEarly` write is `'manual'`.
- `checker.completionReason` becomes `'objectives'` only via `createObjectivesCompletedState`, reached from exactly one call site (`step-checker.hook.ts:503`) after an authored `objectives` string passes a real check. The store record's reason never feeds back into `completionReason`.

So `completeEarly` → error still derives `error`, and Retry/Skip stay reachable — which is what the new `it('reruns failed actions even though completion was persisted early')` tests in both components assert. `getButtonTitle` in multi-step is realigned as a side effect (it already led with `completionReason === 'objectives'`), and it's only rendered on the idle button anyway, so the two can no longer disagree.

### Non-blocking, for later

- **Residual on the objectives fix (low).** Precedence in the derive fixes the display but leaves `executionError` set in component state. If a step completes via objectives (persisting `storedCompleted`) and the objective later stops passing without a remount, `completionReason` falls back to `'none'` while `storedCompleted` stays `true` — so the stale error resurfaces as `error` on a persisted-complete step. Much narrower than the original finding, self-heals on remount, and unreachable from the runner. The effect-based variant (clear `executionError` when `completionReason === 'objectives'`) would have closed it fully; not worth another round.
- **Tech debt (#5), unchanged.** Three near-duplicate derive functions with intentionally divergent precedence. Now explicitly documented in `E2E_TESTING_CONTRACT.md`, which is the important half.
- **Nit.** The relocated `waitForReactUpdates()` comment explains the observability reason but no longer mentions issue #786. The `describe('InteractiveGuided — double skip button (issue #786)')` suite still pins the behavior, so nothing is at risk.

Jayclifford345 (approved):

PR #1443, opened 2026-07-28
fix(e2e): secure runner artifacts and reporting

leslie-heinzen:
## Summary

Concurrent runs shared Playwright output paths, bearer-authenticated traces persisted credentials, broad timeout matching misclassified guide failures as infrastructure, and skip-only or zero-pass runs could report misleading summaries. This change isolates output, disables traces when bearer authentication is active, narrows automatic infrastructure classification, and makes report outcomes reflect verified execution.

## What to look at

- `src/cli/e2e/playwright-runner.ts` — each invocation gets a unique output directory, bearer-token runs explicitly disable tracing, and the security warning is always visible.
- `src/cli/e2e/e2e-console-reporter.ts` — pre-run skips contribute to total, failed, and skipped guide counts.
- `tests/e2e-runner/utils/guide-runner/constants.ts` — only high-confidence network, authentication, browser-crash, and closed-target failures are automatically infrastructure.
- `tests/e2e-runner/utils/guide-runner/execution.ts` — a run with failures and no verified pass cannot report success, while an all-skipped clean run still succeeds.

## How to verify

1. Run two local guide tests with trace enabled and confirm their Playwright outputs use different `playwright-{invocation-id}` directories.
2. Run a bearer-authenticated cloud guide with `--trace` and confirm the runner warns that tracing is disabled and creates no trace archive.
3. Run a credential-safe local or form-login guide with tracing and confirm the trace path is still reported.
4. Produce all-skipped and mixed failed/skipped pre-run reports and confirm their guide counts and outcomes.
5. Trigger messages containing `timeout`, `timed out`, `waiting for`, and `exceeded`; confirm they remain `unknown` unless a high-confidence infrastructure pattern also matches.
6. Produce a run with an optional failure and zero verified passes and confirm the overall outcome is failed.

## Breaking changes

- Bearer-token-authenticated runs no longer emit Playwright traces, even when `--trace` is requested. Screenshots, DOM snapshots, console artifacts, and credential-safe local traces remain available.
- Success now requires zero mandatory failures and either at least one verified passed step or zero failed steps. Previously, a guide with only failed skippable steps could report success.
  - **Impact:** optional-only guides and CI pipelines may see new failures.
  - **Migration:** ensure a guide has at least one step that verifies core functionality, or that all skipped steps are clean skips rather than failed optional executions.
- Generic timeout failures are no longer automatically classified as `infrastructure`; they default to `unknown` unless they also match a high-confidence network, authentication, browser-crash, or closed-target pattern.
  - **Impact:** automated triage and dashboards may route timeout failures differently.
  - **Migration:** update filters and alert routing to account for timeout failures under `unknown`.
- Playwright artifacts now use a per-invocation `playwright-{invocation-id}` subdirectory instead of sharing a flat output path.

## Historical discontinuity

Pass-rate and failure-classification metrics will show a step change at this commit. Reports generated after this change should not be compared directly with earlier reports without accounting for the stricter verified-pass rule and timeout reclassification. Artifact consumers must also follow the new per-invocation output directory.

## Reversibility

The code changes are reversible and do not migrate application state. A revert restores future trace, classification, success, and output-directory behavior, but it cannot remove credential-bearing traces already captured after a revert or erase the historical metrics discontinuity created while this version was active.

## Out of scope

- Semantic-layer shared-stack safety classification remains unchanged.
- Guide execution context and rendered control behavior are handled in PRs #1440–#1442.
- Splitting the security, reporting, and classification changes into separate deployments is not planned for this PR; the added tests and explicit migration guidance provide the rollout boundary.

🤖 Generated with Warp

### DIFF
--- a/docs/developer/E2E_TESTING.md
+++ b/docs/developer/E2E_TESTING.md
@@ -257,7 +257,7 @@ When a step fails, the runner captures:
 - **DOM snapshot**: `{stepId}-dom.html` for selector debugging
 - **Console errors**: `{stepId}-console.json` when the step records console errors
 
-Artifacts are saved to the `--artifacts` directory (or a temp directory by default). With `--always-screenshot`, the runner also captures pre-step screenshots, success screenshots, and a final screenshot. `--trace` records a Playwright trace in a per-invocation output directory and surfaces the trace path in CLI output. Trace capture is disabled for bearer-token-authenticated cloud runs because Playwright traces can contain authorization headers, cookies, and temporary credentials.
+Artifacts are saved to the `--artifacts` directory (or a temp directory by default). With `--always-screenshot`, the runner also captures pre-step screenshots, success screenshots, and a final screenshot. `--trace` records a Playwright trace in a retained per-invocation output directory and surfaces the trace path in CLI output. Non-trace Playwright output directories are removed after each invocation. Trace capture is disabled for bearer-token-authenticated cloud runs because Playwright traces can contain authorization headers, cookies, and temporary credentials.
 
 ## Guided-block test guide
 
--- a/docs/developer/E2E_TESTING.md
+++ b/docs/developer/E2E_TESTING.md
@@ -255,7 +255,7 @@ When a step fails, the runner captures:
 - **DOM snapshot**: `{stepId}-dom.html` for selector debugging
 - **Console errors**: `{stepId}-console.json` when the step records console errors
 
-Artifacts are saved to the `--artifacts` directory (or a temp directory by default). With `--always-screenshot`, the runner also captures pre-step screenshots, success screenshots, and a final screenshot. `--trace` records a Playwright trace and surfaces the trace path in CLI output.
+Artifacts are saved to the `--artifacts` directory (or a temp directory by default). With `--always-screenshot`, the runner also captures pre-step screenshots, success screenshots, and a final screenshot. `--trace` records a Playwright trace in a per-invocation output directory and surfaces the trace path in CLI output. Trace capture is disabled for bearer-token-authenticated cloud runs because Playwright traces can contain authorization headers, cookies, and temporary credentials.
 
 ## Guided-block test guide
 
@@ -455,11 +455,11 @@ When a step fails, the runner assigns an error classification to help with triag
 | `SELECTOR_NOT_FOUND` | `unknown`        | Could be content-drift OR product-regression |
 | `ACTION_FAILED`      | `unknown`        | Needs human triage                           |
 | `REQUIREMENT_FAILED` | `unknown`        | Could be content-drift OR missing setup      |
-| `TIMEOUT`            | `infrastructure` | Likely environmental                         |
+| `TIMEOUT`            | `unknown`        | Could be content, product, or performance    |
 | `NETWORK_ERROR`      | `infrastructure` | Definitely environmental                     |
 | `AUTH_EXPIRED`       | `infrastructure` | Definitely environmental                     |
 
-Only `infrastructure` failures are auto-classified. `SELECTOR_NOT_FOUND`, `ACTION_FAILED`, and `REQUIREMENT_FAILED` default to `unknown` and require human triage — they could indicate content drift, a product regression, or a missing test environment setup.
+Only high-confidence network, authentication, browser-crash, and closed-target failures are auto-classified as `infrastructure`. Selector, action, requirement, and step timeout failures default to `unknown` and require human triage.
 
 The implemented classifier lives in `tests/e2e-runner/utils/guide-runner/classification.ts`.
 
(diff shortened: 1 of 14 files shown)


leslie-heinzen:
@cursoragent review

moxious (changes requested):
1. **`medium` — Report `summary.*` counts have path-dependent meaning.** This PR makes the skip-only path fold pre-run skips into `summary.{totalGuides,skippedGuides,failedGuides}` (`src/cli/e2e/e2e-console-reporter.ts:56-61`), but the mixed multi-guide path (`:250-259`) attaches `preRunSkipped` only as a sibling array — `summary` still comes from `generateMultiGuideReport(resultsWithData)`, which counts only executed guides. So the same pre-run-skipped guide is counted in `summary.totalGuides` in an all-skipped run but not in a mixed run. A consumer reading `summary.totalGuides`/`skippedGuides` undercounts in mixed runs; one that adds `preRunSkipped.length` double-counts in skip-only runs. The new `counts mixed failed and skipped pre-run guides` test only exercises the skip-only branch, and the JSON shape is unchanged so no schema test catches it. It also exposes a related quirk: a failed pre-run guide in the mixed path flips `outcome` to `failed` but leaves `summary.failedGuides === 0`, so `isMultiGuideReportSuccess` disagrees with `outcome`. Report-fidelity only — process exit code reads `GuideRunResult` statuses directly (`e2e-results.ts:307`), so this is not a pass/fail regression. **Fix:** fold `preRunSkipped` counts into `summary` in the mixed path too (where `report.preRunSkipped` is attached at `:254`), which also closes the `outcome`-vs-`failedGuides` disagreement; add a mixed run+skip multi-guide test asserting summary counts.

2. **`low` — Stale JSDoc in `classification.ts` now contradicts a committed test.** The PR removed the timeout regexes from `INFRASTRUCTURE_ERROR_PATTERNS` and updated `types.ts` + `E2E_TESTING.md`, but `tests/e2e-runner/utils/guide-runner/classification.ts` (not in the diff) still says `TIMEOUT … → infrastructure` (`:16-17`) and carries `@example classifyError('Timeout waiting for step completion') // → 'infrastructure'` (`:31`) — the exact opposite of what the new `classification.test.ts` asserts. **Fix in this PR:** drop TIMEOUT from the infrastructure bullet and change the example to `// → 'unknown'`.

3. **`low` — Per-invocation `playwright-<id>` output dir is never cleaned up.** `playwrightOutputDir = join(artifactsDir, 'playwright-<tempbasename>')` (`src/cli/e2e/playwright-runner.ts:186`) lives under the persistent `artifactsDir`; only `tempDir` is removed in `finally`. Bounded and fine with the default ephemeral `--artifacts` dir, but a fixed `--artifacts` dir across repeated runs accumulates unbounded subdirs. (The layout change itself is a legitimate fix — the old flat path let Playwright wipe the prior guide's trace in multi-guide runs.) **Fix:** rm the dir in `finally` after the trace file is read, or document the accumulation.

4. **`low` — Trace gate has no positive-direction regression guard.** The new test covers only the disable direction (trace + token → no `--trace`, `E2E_TRACE=false`, warning). There is no test that `trace: true` with no token still pushes `--trace on` and sets `E2E_TRACE=true`. If the gate is later inverted or coerced, the disable test still passes while traces silently vanish for legitimate local/debug runs. **Fix:** add the mirror-image test.

5. **`nit` — "No verified pass" impact wording re-derived in two reporters.** `printDetailedSummary` (`tests/e2e-runner/utils/console-reporter.ts:~276`) and `logExecutionSummary` (`tests/e2e-runner/utils/guide-runner/execution.ts:~1276`) both inline `mandatoryFailed === 0 && passed === 0`, a third re-derivation of the run-success rule alongside `summarizeResults`. Derive the wording from the already-computed `summary.success` (small shared helper) so display can't drift from the rule.

leslie-heinzen:
## Latest review feedback addressed

Addressed the findings from moxious' review in commit `e862f722`.

- Folded pre-run failed/skipped guides into mixed-report `summary` counts, so mixed and skip-only reports use consistent meanings and `outcome` agrees with `failedGuides`.
- Corrected stale timeout-classification documentation.
- Added cleanup for non-trace per-invocation Playwright output directories; trace-enabled directories remain retained because the surfaced trace path points inside them.
- Added positive trace-enable coverage alongside bearer-token trace suppression.
- Centralized the skippable-failure impact predicate from `summary.success` and reused it in both console summary paths.
- Added regression coverage for mixed reports, output cleanup/retention, and both trace directions.

Validation:

- 4 focused suites passed: 90 tests.
- Full coverage passed: 376 suites, 5,983 tests.
- Typecheck, Prettier, docs sync, and CLI build passed.
- ESLint reported 0 errors and 24 pre-existing warnings.

Ready for re-review.

moxious (approved):

PR #1445, opened 2026-07-28
feat(docs-panel): move kiosk, guide editor, and dev tools into overflow menu

### LINKED ISSUE #1249: Consolidate the overflow "..." menu into settings
## Context

The overflow menu currently mixes "Create guide," internal-only Kiosk mode, extra settings, and feedback. Combine these into a single settings dropdown.

> "this contains 'Create guide', Kiosk mode for us internally, extra settings and feedback. just combine those into the settings dropdown menu"

Image

## Code pointers

- `src/components/docs-panel/components/TabBarActions.tsx:83-143` — the ellipsis (`ellipsis-v`) dropdown in the tab bar's trailing slot; currently contains `Menu.Item`s for "Give feedback" (line 108) and "Settings" (line 113, permission-gated)
- Kiosk mode toggle (`presentation-play`, lines 71-92) is a sibling `IconButton` next to this dropdown, not inside the menu — moving it in is part of this change
- **No "Create guide" menu item was found** anywhere in this component or elsewhere. The closest analog is the permanent editor icon-tab in `src/components/docs-panel/components/DocsPanelTabBar.tsx:76-85` (pencil icon, gated by `isEditorUser`), which is a separate tab, not a menu entry — confirm with Jess whether that's what she means before folding it into the menu

---

dgehring-grafana:
## Summary

Per UX guidance, Kiosk, the guide editor, and Dev tools now live in the docs-panel overflow menu instead of the left rail. Recommendations remains the only home/rail tab. Permanent-tab auto-injection is removed in favor of explicit tab-kind helpers so each chrome surface (home, strip-excluded Dev tools, navigable guide editor) can behave differently — including upcoming multi-editor work.

## What to look at

- **Overflow menu** ([`TabBarActions.tsx`](https://github.com/grafana/grafana-pathfinder-app/blob/85c3ff3d44098c4ca056a688c49b0a70f78b85b5/src/components/docs-panel/components/TabBarActions.tsx)) — Kiosk (still gated by `__pathfinderKioskConfig`), guide editor entry (`isEditorUser`; menu label “Create guide”), and Dev tools (`isDevMode`) as menu items; gates are relocated, not loosened.
- **Tab taxonomy** ([`tab-kinds.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/85c3ff3d44098c4ca056a688c49b0a70f78b85b5/src/components/docs-panel/utils/tab-kinds.ts)) — Replaces `usePermanentTabs` / `PERMANENT_TAB_IDS`. Strip exclusion (recs + Dev tools) is separate from non-content (those + guide editor). `hasNoGuideStripTabs` drives empty-strip/close UX; `hasOnlyNonContentTabs` drives restore so a lone guide editor tab does not block handoff restore from fullscreen/floating.
- **Panel lifecycle** ([`docs-panel.tsx`](https://github.com/grafana/grafana-pathfinder-app/blob/85c3ff3d44098c4ca056a688c49b0a70f78b85b5/src/components/docs-panel/docs-panel.tsx)) — `pruneGatedTabs` / `withoutUnauthorizedGatedTabs` strip unauthorized editor/devtools on restore and when role/dev-mode flips. Guide editor opens as a navigable strip tab (static `"New Guide"` title for now).
- **Tab bar** ([`DocsPanelTabBar.tsx`](https://github.com/grafana/grafana-pathfinder-app/blob/85c3ff3d44098c4ca056a688c49b0a70f78b85b5/src/components/docs-panel/components/DocsPanelTabBar.tsx)) — Filters `GUIDE_STRIP_EXCLUDED_TAB_IDS` so Dev tools never appears as strip chrome even on visibility early returns. Overflow width logic is unchanged in spirit; confirm it still holds with an open guide editor + several guides.

## Why

UX asked for the guide editor and Kiosk under the overflow menu; follow-up discussion moved Dev tools there too so Recommendations is the only home tab. Keeping the old permanent-tabs / left-rail render path was considered, but rejected: Dev tools is no longer a strip tab, and upcoming multi non-content / multi–guide-editor work needs clearer kinds than a single “permanent” set. Some ID→constant renames are sanitization/future-proofing beyond the strict UX ask.

## How to verify

1. Enable **kiosk** (plugin config / existing kiosk enablement path) and **pathfinder dev mode**; open the docs-panel `⋯` menu — confirm Kiosk and Dev tools appear only when their gates are on, and disappear when off.
2. As an editor/admin, open **Create guide** from the menu — confirm a navigable guide editor strip tab opens and content renders; close it and confirm strip/overflow behavior with other guides still looks right.
3. Open several content tabs until overflow kicks in — confirm chevron/overflow dropdown still lists guide tabs correctly and Dev tools never appears as a strip or overflow tab.
4. Open Dev tools from the menu — confirm content switches without a strip tab, and recommendations is not falsely “active.”
5. With the guide editor open (no other content tabs), switch panel mode sidebar ↔ fullscreen/floating and back — confirm guides written on the other surface still restore into the sidebar (guide editor alone must not block restore).
6. With a guide editor or Dev tools tab persisted, drop editor role / leave dev mode — confirm those tabs are pruned from state/storage.

## Breaking changes

None for end users beyond UI placement. No kiosk/dev-mode enablement shims; defaults and gates are unchanged. Tab storage shape is the same; unauthorized gated tabs are now pruned more consistently on restore.

## Out of scope

- Multiple concurrent guide editor tabs and dynamic tab title sync from the working guide — planned follow-up PR.
- Header affordance when Dev tools is active (no strip active state by design for now).
- Model-level auth checks inside `openEditorTab` / `openDevToolsTab` (UI + prune remain the gates).

Fixes #1249


moxious (changes requested):
I rechecked the updated head (`98987467`). The editor E2E entry path has been updated to use **More options → Create guide**, and the current checks are green. That finding is resolved.

Three verified runtime/security findings remain:

1. **[medium] Do not let closing a visible tab activate strip-hidden Dev Tools.**

   `closeTab` selects the replacement from raw `currentTabs` adjacency, although Dev Tools is excluded from the rendered strip. With an editor/guide remaining, closing the active guide can switch to Dev Tools while no visible tab has an active marker.

   Choose the replacement from strip-visible tabs or fall back to Recommendations. Add close-adjacency tests with Dev Tools next to an editor or guide.

   Reversibility: reversible.

2. **[security][medium] Validate reserved tab IDs and types together.**

   A persisted `{ id: "devtools", type: "docs" }` record takes the generic restore path, survives type-based pruning, and then reaches the ID-based Dev Tools renderer with dev mode off. The new pruning contract therefore does not consistently enforce the stated gate at the persisted-state boundary.

   Canonicalize or reject reserved ID/type mismatches, prune using the same canonical predicate used by rendering, and retain a render-level dev-mode guard. Add a storage-restore regression test for the mismatch.

   Reversibility: reversible.

3. **[medium] Use one Dev Tools configuration source on `/context`.**

   `docsPage.ts` constructs `CombinedLearningJourneyPanel({})`, while the renderer reads live plugin configuration and can show the Dev Tools menu item. After Dev Tools opens, the new tabs-dependent pruning effect evaluates the model's empty configuration and immediately removes the authorized tab.

   Construct the scene model with the same resolved configuration used by the renderer, synchronize model configuration, or pass the resolved gate into pruning. Add a `/context` integration test that opens Dev Tools in dev mode and verifies it remains active.

   Reversibility: reversible.

The last two findings share the same architectural cause: Dev Tools authorization is decided from different representations in different layers—renderer configuration versus model configuration, and tab ID versus tab type. Consolidating gate evaluation and tab-kind validation should resolve both cleanly.

### Follow-up scrutiny

I tested the current singleton editor design as a blocker for the planned multi-editor follow-up. Three independent verification passes agreed it is not a defect in this PR because no current path creates alternate editor IDs. The follow-up should still change type/ID dispatch, persistence, sidebar rendering, selection, and surface handoff together.

### Tech debt

Non-blocking changed-files scan:

- `FullScreenPanelRenderer` mixes multiple event lifecycles with restoration, routing, telemetry, and surface transitions; consider extracting focused surface-event hooks.
- `tab-kinds.ts` is mostly commentary around one-line predicates. Keep the non-obvious strip-versus-content invariant and let names/types describe the rest.
- Suggestive: `docs-panel.tsx` remains a high-churn drift hub; raw Pathfinder event names remain scattered across floating/fullscreen surfaces; `useTabRestoration.ts` retains substantial process-history commentary.

Instrumentation: new behavior is covered by the free telemetry channels (or existing facade operations); no gaps found.

Doc drift: none detected. Coverage confidence: high.

dgehring-grafana:
@moxious 

Addressed 1, 2, and 3 - 

1. closeTab activating strip-hidden Dev Tools
closeTab now picks the next focus from getGuideStripTabs(currentTabs) (excludes recommendations + Dev Tools), not raw currentTabs. Empty strip falls back to recommendations. Covered in docs-panel.close-tab.test.ts (guide next to Dev Tools + editor, last strip tab with Dev Tools still open, background close leaves Dev Tools alone).

2. Reserved ID/type mismatch
getCanonicalPersistedTabType in tab-storage-restore.ts requires reserved IDs and types to match both ways — e.g. { id: "devtools", type: "docs" } is rejected before it enters state. Pruning/rendering key off type. DocsPanelContentArea still gates Dev Tools with activeTab?.type === 'devtools' && isDevMode. Regression: rejects noncanonical reserved IDs… in tab-storage-restore.test.ts, plus content-area tests for type/dev-mode mismatch.

3. Split Dev Tools config on /context
docsPage.ts builds the panel with getConfigWithDefaults(window.__pathfinderPluginConfig || {}) instead of {}. The renderer syncs via model.syncPluginConfig(...) (null while unresolved so it doesn’t look like “dev mode off”), and pruning uses this.state.pluginConfig only. Pinned by docsPage.test.ts.

Investigating these issues brought to my attention a systemic problem where we seem to derive config and state from multiple sources of truth depending on where it's convenient - I'd like to either understand why or propose some work to create a centralized place to service this information.

Regarding Cursor Bugbot's Findings -

Addressed point 1 by filtering out dev tools from the shortcut. Did not address point 2 because state restoration was not initially present for these tabs on toggle of either mode and the sprawl is already huge. Perhaps it can be addressed in a future item?

moxious (approved):

PR #1446, opened 2026-07-28
feat(docs-panel): choose launch surface from guide content for My Learning

moxious:
## Intent

A large amount of **non-interactive** content is coming to Pathfinder, and it will be launched from the **My Learning** page. Non-interactive content — prose, images, video — is a poor fit for the sidebar: it doesn't use the available screen real estate, and the sidebar's whole reason for being is to let you _use Grafana while a guide runs beside it_. That benefit means nothing for content you only read or watch.

So this began with one question: **how do we make sure non-interactive content launches in full-screen mode, since it needs no app interaction?**

Answering it honestly pulled in the other half of the problem: sometimes you launch content that **is** interactive, and that content genuinely needs the sidebar so its "show me / do it" steps can drive the real Grafana UI. A launch therefore has to decide, per guide, which surface fits — without a jarring switch, and without trampling a preference the user set themselves.

That gives three requirements:

1. **Determine whether a guide needs the Grafana UI.** We deliberately did **not** add another manifest metadata field for this. Whether a guide drives the Grafana UI is _computable from the guide's own content_ — so we compute it, and a value derived from the content it describes can never drift out of sync with that content. A hand-maintained flag inevitably would (an author adds an interactive step, forgets the flag → wrong surface). The classifier is `src/components/docs-panel/utils/requires-grafana-ui.ts`; it detects the Grafana-driving actions `highlight`, `button`, `formfill`, `navigate`, `hover`, plus `code-block` (whose required `reftarget` targets a live Monaco editor) (`src/types/json-guide.schema.ts`), recursing through sections, multistep/guided steps, conditionals, and resolved snippet references.
2. **Decide the launch surface as the guide loads.** `src/components/docs-panel/utils/prepare-guide-launch.ts` fetches the guide, resolves its snippets, classifies it, and carries that _same_ fetched content into the chosen surface — one fetch, surface committed before render, so there's no second load and no visible bounce. Wired into the My Learning launch path (`src/components/LearningPaths/MyLearningTab.tsx`, `src/components/Home/HomePanel.tsx`, `src/components/docs-panel/docs-panel.tsx`).
3. **Respect the user's own preference.** An automatic launch choice is _transient_ — it never overwrites the panel-mode preference the user set deliberately; only genuine user surface choices (pop-out, switch to full-screen/floating, `panelMode=` deep link) persist. Enforced centrally in `src/global-state/panel-mode.ts`.

## What changes

- A **non-interactive** guide launched from My Learning now opens **full-screen** automatically, using the whole content area for reading (previously: sidebar).
- An **interactive** guide launched from My Learning opens in the **sidebar** (or a floating panel if another plugin owns the sidebar), chosen from the guide's real content rather than a fixed default.
- The surface is decided **as the guide loads**, from a single fetch — no switch after render.
- **Code blocks count as interactive**: a guide whose only interactivity is a `code-block` still opens beside Grafana, because its "Show me" highlights and "Insert" mutates a live Monaco editor that doesn't exist in full-screen mode.

## How the prepared content travels (hardened after review)

The launch handoff crosses two kinds of channel, and after the review pass they are treated differently:

- **Trusted, module-owned memory** — `panelModeManager.setPendingGuide` (full-screen and occupied-sidebar→floating handoffs). Carries the prepared payload directly; consumed once on the receiving surface's mount via the shared `consumePendingGuideOnMount` router, so all surfaces route identically. The pending guide also carries the original launch `source`, so the starting-location alignment check behaves the same regardless of which surface the guide lands in.
- **Public channels** — the `pathfinder-auto-open-docs` CustomEvent and the cold-sidebar link queue, both writable by any same-page script. These never carry the payload. The producer stages it in a module-owned, consume-once store (`src/global-state/guide-launch.ts`) and sends only an opaque `launchKey`; the listener redeems the key at the trusted boundary. Redemption requires the key to be genuine, unused, less than 60s old, and presented with the exact URL it was staged for. **Any miss falls back to the loader's normal validated fetch** — the optimization's failure mode is the pre-existing two-fetch behavior, never trust in unvalidated content.

Snippet resolution during launch preparation is deadline-bounded (`AbortSignal.timeout` on the CDN fetches); a timeout flows into the existing unresolved-snippet fail-safe, which forces the sidebar rather than hiding a possibly-interactive guide in full-screen mode.

## What does NOT change

- **Your saved panel-mode preference is never overwritten by an auto-launch.** Reading a prose guide (transient full-screen) and exiting leaves your saved preference untouched. Deliberately choosing a surface still persists as before.
- **Manual full-screen entry/exit** behaves exactly as it always did.
- **Guides launched from anywhere other than My Learning** are unaffected — this is scoped to the My Learning launch path.
- **Cold `/fullscreen?doc=` deep links** are honored exactly as before.
- **The interactive engine is untouched** — "show me / do it / do section", section completion, starting-location alignment, journey navigation, and completion tracking all behave as before. This PR decides _which surface a guide opens in_, not how guides run.
- **No new manifest field, no authoring burden** — authors tag nothing; the behavior falls out of the content.
- **The `pathfinder-auto-open-docs` event keeps its pinned pre-PR detail shape** (`{url, title, source}`), extended only by the opaque `launchKey`.

## Testing

- **Classifier** (`requires-grafana-ui.test.ts`): the five Grafana-driving actions detected through sections, multistep/guided steps, conditionals, and snippet-refs; regression tests prove a `code-block`-only guide (top-level and nested) forces the sidebar; negative cases confirm `noop`, `popout`, quiz/input/terminal/etc. do **not**.
- **Prepared launch** (`prepare-guide-launch.test.ts`): fetch → resolve snippets → classify → carry, verifying a single content fetch.
- **Trust boundary** (`guide-launch.test.ts`, `useAutoOpenListener.test.ts`): consume-once redemption; forged, replayed, stale, and URL-mismatched keys all fall back to the normal fetch; a forged event smuggling `preparedContent`/`packageInfo` inline is ignored.
- **Cold-sidebar relay** (`useAutoOpenListener.cold-sidebar-relay.test.ts`): the real queue, real store, and real listener relay a staged payload end to end (queue → drain → event → listener → model) with zero network fetches; the loader's no-fetch consumption of `preparedContent` is pinned in `docs-panel.load-tab-content.test.ts`.
- **Occupied-sidebar → floating handoff** (`pendingGuideRouter.test.ts`, `floating-panel.pending-guide.test.ts`): the floating surface consumes the pending guide exactly once on mount, marks the open in-flight before the restoration/empty-state gates can run, and preserves the original launch source.
- **Launch deadline** (`online-snippet-resolver.test.ts`): a never-settling snippet CDN fetch resolves into the existing failure shapes instead of hanging the launch.
- **Acceptance criteria:**
  - **A** — a non-interactive guide launched from My Learning opens full-screen.
  - **B** — an interactive guide launched from My Learning opens in the sidebar.
  - **C** — launch independence: B opens in the sidebar regardless of whether A launched first, and A opens full-screen regardless of whether B launched first (both orderings).
- **Preference round-trip invariant** (`panel-mode.test.ts`): a stored non-default preference (e.g. `floating`) survives an auto-launch enter + teardown across sidebar, full-screen exit, and floating teardown (including the auto-dock-to-floating path); a deliberate mode change after a round-trip still persists.
- **CI:** all checks green on the head commit, including the six-image Playwright e2e matrix.

## Review response

All six findings from the [review](https://github.com/grafana/grafana-pathfinder-app/pull/1446#issuecomment-5110518588) are dispositioned ([full mapping](https://github.com/grafana/grafana-pathfinder-app/pull/1446#issuecomment-5111106120)): findings 1–5 fixed in dedicated commits (`8b9c47b6`, `5c92ccdf`, `74cfec0a`, `5849d59b`, `c41bdbbe` — one per finding, rationale in each message); finding 6 explicitly deferred to #1448 (needs a new panel-mode primitive and touches the race-sensitive auto-dock — split so this PR's surface stays reviewable). Adjacent debt discovered while verifying the review is tracked in #1449 and #1450, both out of scope here.

### DIFF
--- a/src/components/Home/HomePanel.tsx
+++ b/src/components/Home/HomePanel.tsx
@@ -58,6 +58,7 @@ function pendingGuideFrom(launch: PreparedGuideLaunch): PendingGuide {
     type: launch.type,
     packageInfo: launch.packageInfo,
     preparedContent: launch.preparedContent,
+    source: launch.source,
   };
 }
 
--- a/src/components/Home/HomePanel.tsx
+++ b/src/components/Home/HomePanel.tsx
@@ -30,7 +30,7 @@ import { guideLaunchStore } from '../../global-state/guide-launch';
 import { linkInterceptionState } from '../../global-state/link-interception';
 import { panelModeManager, type PendingGuide } from '../../global-state/panel-mode';
 import { isExtensionSidebarOwnedByOther } from '../../lib/storage/extension-sidebar';
-import { REQUEST_FLOATING_GUIDE_EVENT } from '../../lib/event-names';
+import { AUTO_OPEN_DOCS_EVENT, REQUEST_FLOATING_GUIDE_EVENT } from '../../lib/event-names';
 import { PLUGIN_BASE_URL, ROUTES } from '../../constants';
 import { reportAppInteraction, UserInteraction } from '../../lib/analytics';
 import { buildFullScreenRouteUrl } from '../../utils/pathfinder-search-params';
@@ -113,7 +113,7 @@ export function HomePanelRenderer() {
     });
     if (sidebarState.getIsSidebarMounted()) {
       document.dispatchEvent(
-        new CustomEvent('pathfinder-auto-open-docs', {
+        new CustomEvent(AUTO_OPEN_DOCS_EVENT, {
           detail: { url: launch.url, title: launch.title, source: launch.source, launchKey },
         })
       );
--- a/src/components/Home/HomePanel.tsx
+++ b/src/components/Home/HomePanel.tsx
@@ -29,6 +29,7 @@ import { guideLaunchStore } from '../../global-state/guide-launch';
 import { linkInterceptionState } from '../../global-state/link-interception';
 import { panelModeManager, type PendingGuide } from '../../global-state/panel-mode';
 import { isExtensionSidebarOwnedByOther } from '../../lib/storage/extension-sidebar';
+import { REQUEST_FLOATING_GUIDE_EVENT } from '../../lib/event-names';
 import { PLUGIN_BASE_URL, ROUTES } from '../../constants';
 import { buildFullScreenRouteUrl } from '../../utils/pathfinder-search-params';
 import pluginJson from '../../plugin.json';
@@ -73,6 +74,10 @@ export function HomePanelRenderer() {
     if (isExtensionSidebarOwnedByOther(pluginJson.id)) {
       panelModeManager.setPendingGuide(pendingGuideFrom(launch));
       panelModeManager.setModeTransient('floating');
+      // A same-mode transient launch dispatches no mode-change event, so an
+      // already-mounted floating panel never remounts to consume. Signal it
+      // directly; unheard when the panel isn't up yet (mount consumes then).
+      document.dispatchEvent(new CustomEvent(REQUEST_FLOATING_GUIDE_EVENT));
       return;
     }
 
--- a/src/components/Home/HomePanel.tsx
+++ b/src/components/Home/HomePanel.tsx
@@ -3,16 +3,36 @@
  *
  * SceneObjectBase wrapper + React composition root for the home page.
  * Renders MyLearningTab as the full-page learning hub at /a/grafana-pathfinder-app.
- * Guide opens are dispatched to the sidebar panel.
+ *
+ * When a guide is launched from My Learning it arrives here already fetched,
+ * snippet-expanded, and classified (see `prepareGuideLaunch`). This handler
+ * picks the display surface from that classification:
+ *
+ * - reading-only content (no Grafana-driving action) → full screen, so the
+ *   whole viewport is used for reading;
+ * - content that drives the Grafana UI → the sidebar, so its "show me / do it"
+ *   actions have the Grafana main area to work on — or a floating overlay when
+ *   another plugin owns the extension sidebar.
+ *
+ * The surface choice is transient: it never overwrites the user's persisted
+ * panel-mode preference. The prepared content is carried through to the
+ * destination so no second fetch happens.
  */
 
 import React, { useCallback } from 'react';
 import { SceneObjectBase, type SceneObjectState } from '@grafana/scenes';
+import { locationService } from '@grafana/runtime';
 import { useStyles2 } from '@grafana/ui';
 
 import { sidebarState } from '../../global-state/sidebar';
 import { linkInterceptionState } from '../../global-state/link-interception';
+import { panelModeManager, type PendingGuide } from '../../global-state/panel-mode';
+import { isExtensionSidebarOwnedByOther } from '../../lib/storage/extension-sidebar';
+import { PLUGIN_BASE_URL, ROUTES } from '../../constants';
+import { buildFullScreenRouteUrl } from '../../utils/pathfinder-search-params';
+import pluginJson from '../../plugin.json';
 import { MyLearningTab } from '../LearningPaths';
+import type { PreparedGuideLaunch } from '../docs-panel/utils/prepare-guide-launch';
 import { MyLearningErrorBoundary } from '../docs-panel/components';
 import { getHomePageStyles } from './home.styles';
 import { testIds } from '../../constants/testIds';
@@ -31,12 +51,35 @@ export class HomePanel extends SceneObjectBase<HomePanelState> {
 // RENDERER
 // ============================================================================
 
+function pendingGuideFrom(launch: PreparedGuideLaunch): PendingGuide {
+  return {
+    url: launch.url,
+    title: launch.title,
+    type: launch.type,
+    packageInfo: launch.packageInfo,
+    preparedContent: launch.preparedContent,
+  };
+}
+
 export function HomePanelRenderer() {
   const styles = useStyles2(getHomePageStyles);
 
-  const handleOpenGuide = useCallback((url: string, title: string) => {
-    const detail = { url, title, source: 'home_page' };
+  // Open beside Grafana: sidebar, or a floating overlay when another plugin
+  // owns the extension sidebar. Carries the prepared content so the tab opens
+  // without a second fetch.
+  const openBesideGrafana = useCallback((launch: PreparedGuideLaunch) => {
+    if (isExtensionSidebarOwnedByOther(pluginJson.id)) {
+      panelModeManager.setPendingGuide(pendingGuideFrom(launch));
+      panelModeManager.setModeTransient('floating');
+      return;
+    }
 
+    const detail = {
+      url: launch.url,
+      title: launch.title,
+      source: launch.source,
+      preparedContent: launch.preparedContent,
+    };
     if (sidebarState.getIsSidebarMounted()) {
       document.dispatchEvent(new CustomEvent('pathfinder-auto-open-docs', { detail }));
     } else {
@@ -46,9 +89,42 @@ export function HomePanelRenderer() {
         title: detail.title,
         timestamp: Date.now(),
       });
-      linkInterceptionState.addToQueue({ ...detail, timestamp: Date.now() });
+      linkInterceptionState.addToQueue({
+        url: launch.url,
+        title: launch.title,
+        timestamp: Date.now(),
+        preparedContent: launch.preparedContent,
+      });
     }
   }, []);
+
+  // Open full screen for reading-only content. Mirrors the sidebar→full-screen
+  // handoff order (pending guide → prior path → mode → route), but selects the
+  // mode transiently so the user's stored preference is untouched.
+  const openFullScreen = useCallback((launch: PreparedGuideLaunch) => {
+    panelModeManager.setPendingGuide(pendingGuideFrom(launch));
+    panelModeManager.capturePriorPath(window.location.pathname + window.location.search);
+    panelModeManager.setModeTransient('fullscreen');
+    locationService.push(
+      buildFullScreenRouteUrl({
+        pluginBaseUrl: PLUGIN_BASE_URL,
+        fullScreenRoute: ROUTES.FullScreen,
+        doc: launch.url,
+        guideType: launch.type,
+      })
+    );
+  }, []);
+
+  const handleOpenGuide = useCallback(
+    (launch: PreparedGuideLaunch) => {
+      if (launch.requiresGrafanaUi) {
+        openBesideGrafana(launch);
+      } else {
+        openFullScreen(launch);
+      }
+    },
+    [openBesideGrafana, openFullScreen]
+  );
 
   return (
     <div className={styles.container} data-testid={testIds.homePage.container}>
--- a/src/components/Home/HomePanel.tsx
+++ b/src/components/Home/HomePanel.tsx
@@ -25,6 +25,7 @@ import { locationService } from '@grafana/runtime';
 import { useStyles2 } from '@grafana/ui';
 
 import { sidebarState } from '../../global-state/sidebar';
+import { guideLaunchStore } from '../../global-state/guide-launch';
 import { linkInterceptionState } from '../../global-state/link-interception';
 import { panelModeManager, type PendingGuide } from '../../global-state/panel-mode';
 import { isExtensionSidebarOwnedByOther } from '../../lib/storage/extension-sidebar';
@@ -77,28 +78,31 @@ export function HomePanelRenderer() {
 
     panelModeManager.setModeTransient('sidebar');
 
-    const detail = {
+    // The event and the cold-sidebar queue are public channels — stage the
+    // prepared payload in module-owned memory and send only the opaque key.
+    const launchKey = guideLaunchStore.stage({
       url: launch.url,
-      title: launch.title,
-      source: launch.source,
       preparedContent: launch.preparedContent,
       packageInfo: launch.packageInfo,
-    };
+    });
     if (sidebarState.getIsSidebarMounted()) {
-      document.dispatchEvent(new CustomEvent('pathfinder-auto-open-docs', { detail }));
+      document.dispatchEvent(
+        new CustomEvent('pathfinder-auto-open-docs', {
+          detail: { url: launch.url, title: launch.title, source: launch.source, launchKey },
+        })
+      );
     } else {
       sidebarState.setPendingOpenSource('home_page');
       sidebarState.openSidebar('Interactive learning', {
-        url: detail.url,
-        title: detail.title,
+        url: launch.url,
+        title: launch.title,
         timestamp: Date.now(),
       });
       linkInterceptionState.addToQueue({
         url: launch.url,
         title: launch.title,
         timestamp: Date.now(),
-        preparedContent: launch.preparedContent,
-        packageInfo: launch.packageInfo,
+        launchKey,
       });
     }
   }, []);
--- a/src/components/Home/HomePanel.tsx
+++ b/src/components/Home/HomePanel.tsx
@@ -74,11 +74,14 @@ export function HomePanelRenderer() {
       return;
     }
 
+    panelModeManager.setModeTransient('sidebar');
+
     const detail = {
       url: launch.url,
       title: launch.title,
       source: launch.source,
       preparedContent: launch.preparedContent,
+      packageInfo: launch.packageInfo,
     };
     if (sidebarState.getIsSidebarMounted()) {
       document.dispatchEvent(new CustomEvent('pathfinder-auto-open-docs', { detail }));
@@ -94,6 +97,7 @@ export function HomePanelRenderer() {
         title: launch.title,
         timestamp: Date.now(),
         preparedContent: launch.preparedContent,
+        packageInfo: launch.packageInfo,
       });
     }
   }, []);
--- a/src/components/Home/HomePanel.tsx
+++ b/src/components/Home/HomePanel.tsx
@@ -12,7 +12,8 @@
  *   whole viewport is used for reading;
  * - content that drives the Grafana UI → the sidebar, so its "show me / do it"
  *   actions have the Grafana main area to work on — or a floating overlay when
- *   another plugin owns the extension sidebar.
+ *   another plugin owns the extension sidebar or the floating panel is already
+ *   the current surface.
  *
  * The surface choice is transient: it never overwrites the user's persisted
  * panel-mode preference. The prepared content is carried through to the
@@ -71,7 +72,13 @@ export function HomePanelRenderer() {
   // owns the extension sidebar. Carries the prepared content so the tab opens
   // without a second fetch.
   const openBesideGrafana = useCallback((launch: PreparedGuideLaunch) => {
-    if (isExtensionSidebarOwnedByOther(pluginJson.id)) {
+    // A floating panel already IS beside Grafana — deliver the guide there.
+    // Forcing 'sidebar' would unmount the floating panel without anything
+    // opening the extension sidebar: `isSidebarMounted` stays true (the
+    // floating cleanup leaves it for the expected sidebar mount), so the
+    // auto-open event below would fire with no listener and the guide would
+    // be lost. See #1450 for the listener-ownership gap.
+    if (isExtensionSidebarOwnedByOther(pluginJson.id) || panelModeManager.getMode() === 'floating') {
       panelModeManager.setPendingGuide(pendingGuideFrom(launch));
       panelModeManager.setModeTransient('floating');
       // A same-mode transient launch dispatches no mode-change event, so an
--- a/src/components/Home/HomePanel.tsx
+++ b/src/components/Home/HomePanel.tsx
@@ -32,6 +32,7 @@ import { panelModeManager, type PendingGuide } from '../../global-state/panel-mo
 import { isExtensionSidebarOwnedByOther } from '../../lib/storage/extension-sidebar';
 import { REQUEST_FLOATING_GUIDE_EVENT } from '../../lib/event-names';
 import { PLUGIN_BASE_URL, ROUTES } from '../../constants';
+import { reportAppInteraction, UserInteraction } from '../../lib/analytics';
 import { buildFullScreenRouteUrl } from '../../utils/pathfinder-search-params';
 import pluginJson from '../../plugin.json';
 import { MyLearningTab } from '../LearningPaths';
@@ -53,6 +54,17 @@ export class HomePanel extends SceneObjectBase<HomePanelState> {
 // ============================================================================
 // RENDERER
 // ============================================================================
+
+// Answers "is the classifier picking the right surface in the wild?" — the
+// surface itself is also reported by mounts/mode changes, but only this event
+// ties it to the content and the classification that chose it.
+function reportSurfaceChoice(surface: 'sidebar' | 'floating' | 'fullscreen', launch: PreparedGuideLaunch): void {
+  reportAppInteraction(UserInteraction.GuideLaunchSurfaceChosen, {
+    surface,
+    requires_grafana_ui: launch.requiresGrafanaUi,
+    content_url: launch.url,
+  });
+}
 
 function pendingGuideFrom(launch: PreparedGuideLaunch): PendingGuide {
   return {
@@ -79,6 +91,7 @@ export function HomePanelRenderer() {
     // auto-open event below would fire with no listener and the guide would
     // be lost. See #1450 for the listener-ownership gap.
     if (isExtensionSidebarOwnedByOther(pluginJson.id) || panelModeManager.getMode() === 'floating') {
+      reportSurfaceChoice('floating', launch);
       panelModeManager.setPendingGuide(pendingGuideFrom(launch));
       panelModeManager.setModeTransient('floating');
       // A same-mode transient launch dispatches no mode-change event, so an
@@ -88,6 +101,7 @@ export function HomePanelRenderer() {
       return;
     }
 
+    reportSurfaceChoice('sidebar', launch);
     panelModeManager.setModeTransient('sidebar');
 
     // The event and the cold-sidebar queue are public channels — stage the
@@ -123,6 +137,7 @@ export function HomePanelRenderer() {
   // handoff order (pending guide → prior path → mode → route), but selects the
   // mode transiently so the user's stored preference is untouched.
   const openFullScreen = useCallback((launch: PreparedGuideLaunch) => {
+    reportSurfaceChoice('fullscreen', launch);
     panelModeManager.setPendingGuide(pendingGuideFrom(launch));
     panelModeManager.capturePriorPath(window.location.pathname + window.location.search);
     panelModeManager.setModeTransient('fullscreen');
--- a/src/components/LearningPaths/MyLearningTab.tsx
+++ b/src/components/LearningPaths/MyLearningTab.tsx
@@ -18,6 +18,7 @@ import { BadgeIcon } from './BadgeIcon';
 import { SkeletonLoader } from '../SkeletonLoader';
 import { FeedbackButton } from '../FeedbackButton/FeedbackButton';
 import { reportAppInteraction, UserInteraction, AnalyticsContentType } from '../../lib/analytics';
+import { logger } from '../../lib/logging';
 import { StorageEvents } from '../../lib/event-names';
 import {
   learningProgressStorage,
@@ -336,9 +337,16 @@ export function MyLearningTab({ onOpenGuide }: MyLearningTabProps) {
         if (result.ok) {
           onOpenGuide(result.launch);
         } else {
+          // The raw error is internal-shaped — keep it for the logs (Faro
+          // bridge makes launch failures countable) and show a translated
+          // generic message.
+          logger.error('[MyLearning] Guide launch preparation failed', { url, error: result.error });
           getAppEvents().publish({
             type: 'alert-error',
-            payload: ['Could not open the guide', result.error],
+            payload: [
+              t('myLearning.launchErrorTitle', 'Could not open the guide'),
+              t('myLearning.launchErrorMessage', 'Something went wrong while loading the guide. Please try again.'),
+            ],
           });
         }
       } finally {
--- a/src/components/LearningPaths/MyLearningTab.tsx
+++ b/src/components/LearningPaths/MyLearningTab.tsx
@@ -5,7 +5,7 @@
  * Provides a unified experience for users to explore and track their learning journey.
  */
 
-import React, { useState, useCallback, useMemo, useRef } from 'react';
+import React, { useState, useCallback, useEffect, useMemo, useRef } from 'react';
 import { useStyles2, Icon } from '@grafana/ui';
 import { getAppEvents } from '@grafana/runtime';
 import { t } from '@grafana/i18n';
@@ -222,6 +222,16 @@ export function MyLearningTab({ onOpenGuide }: MyLearningTabProps) {
   const styles = useStyles2(getMyLearningStyles);
   // Guards against a second launch while the first is still fetching/classifying.
   const launchInFlightRef = useRef(false);
+  // Drives the pending affordance on the launching card while the ref above
+  // stays the correctness guard.
+  const [launchingPathId, setLaunchingPathId] = useState<string | null>(null);
+  const mountedRef = useRef(true);
+  useEffect(() => {
+    mountedRef.current = true;
+    return () => {
+      mountedRef.current = false;
+    };
+  }, []);
   const [showAllBadges, setShowAllBadges] = useState(false);
   const [showAllPaths, setShowAllPaths] = useState(false);
   const [selectedBadge, setSelectedBadge] = useState<EarnedBadge | null>(null);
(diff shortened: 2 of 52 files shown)


moxious:
## Review findings

I recommend addressing the following before merge.

### 1. High — Validate prepared launch payloads at the public event boundary

`useAutoOpenListener` now accepts `preparedContent` and `packageInfo` directly from the document-level `pathfinder-auto-open-docs` `CustomEvent`. Any same-page script or co-plugin can dispatch this event. `parseUrlSafely` only helps choose the open method; it does not validate the new payloads or establish that they came from `prepareGuideLaunch`.

When `preparedContent` is present, `loadTabContent`/`loadDocsTabContent` skips the normal fetch pipeline. `packageInfo` also controls render type and starting-location behavior and is persisted with the tab. This lets a forged event bypass the normal URL/HTTPS/package-manifest validation boundary.

Keep the enriched payload in trusted module-owned memory and put only a consume-once key in the public event, or fully validate `preparedContent` and `packageInfo` at the listener/model boundary before using or persisting them. Add a negative test that dispatches malformed/forged prepared content and package metadata.

Relevant code: `src/components/docs-panel/hooks/useAutoOpenListener.ts:35-63`, `src/components/docs-panel/docs-panel.tsx:404-450,675-746`.

### 2. High — Treat `code-block` as requiring the Grafana UI

The classifier deliberately returns `false` for `code-block`, but `CodeBlockStep` uses `refTarget` to highlight a live Monaco editor and `clearAndInsertCode` to mutate it. A guide containing only a code block is therefore launched fullscreen, where the target Grafana editor is not available and both “Show me” and “Insert” cannot perform their intended actions.

Classify `code-block` as Grafana-driving and replace the current negative assertion with a regression test proving it opens beside Grafana.

Relevant code: `src/components/docs-panel/utils/requires-grafana-ui.ts:46-63`, `src/components/docs-panel/utils/requires-grafana-ui.test.ts:41-58`, `src/components/interactive-tutorial/code-block-step.tsx:195-226`.

### 3. High — Consume the pending guide in the occupied-sidebar floating path

When another plugin owns the extension sidebar, `HomePanel` stores the prepared guide with `setPendingGuide` and switches to transient floating mode. `FloatingPanelManager` never consumes that pending guide; it only restores tab storage and listens for the separate auto-launch event. On a cold launch, the requested guide is not opened and the empty-state fallback can switch the panel back to sidebar mode.

Consume the pending guide when `FloatingPanelInner` mounts, using the same `openPendingGuide` routing used by fullscreen, and mark the guide open as in-flight before restoration/fallback can run. Add an integration test for My Learning → occupied sidebar → floating panel with empty tab storage.

Relevant code: `src/components/Home/HomePanel.tsx:70-74`, `src/components/floating-panel/FloatingPanelManager.tsx:90-174`.

### 4. High — Bound snippet resolution so a launch cannot hang indefinitely

`prepareGuideLaunch` waits for all snippet references before choosing a surface. The online resolver uses unbounded `fetch` calls, and `resolveRefs` waits on `Promise.all`. If any request never settles, `MyLearningTab.launch` never reaches its `finally`, leaving `launchInFlightRef` set forever and preventing every later guide launch from the page.

Add an abort signal/deadline to snippet catalog and content requests, convert timeout/abort to the existing unresolved-snippet fail-safe result, and test with a never-resolving fetch.

Relevant code: `src/components/docs-panel/utils/prepare-guide-launch.ts:65-108`, `src/snippet-engine/inline-refs.ts:67-77`, `src/snippet-engine/online-snippet-resolver.ts:35-90`, `src/components/LearningPaths/MyLearningTab.tsx:310-330`.

### 5. Medium — Add a cold-sidebar payload-relay boundary test

The new one-fetch path crosses several independently mocked boundaries when the sidebar starts closed:

`HomePanel` queue → `linkInterceptionState.processQueuedLinks` → `pathfinder-auto-open-docs` → `useAutoOpenListener` → model open method → loader.

There is no test that proves `preparedContent` and `packageInfo` survive that complete route and are consumed without a second fetch. The current unit tests can all remain green if one handoff silently drops either field.

Add a boundary-level test that starts with an unmounted sidebar, queues a prepared launch, mounts the listener/model, drains the queue, and asserts that the loader receives the exact prepared content/package metadata and performs no network fetch.

Relevant code: `src/components/Home/HomePanel.tsx:86-101`, `src/global-state/link-interception.ts:83-100`, `src/components/docs-panel/hooks/useAutoOpenListener.ts:35-63`.

### 6. Medium — Browser Back defeats the prose-only fullscreen decision

The persisted `panelMode` preference is preserved, but the sidebar's actual open/closed state is not. A prose-only launch from My Learning enters transient fullscreen. Browser Back then reaches the pre-existing generic fullscreen auto-dock handler, which switches to sidebar mode and calls `openSidebar` unconditionally. Shared tab storage restores the same prose guide there, even when Pathfinder's sidebar was closed before launch.

The generic auto-dock code predates this PR and is appropriate for an interactive step navigating into Grafana. This PR introduces the prose-only My Learning entry path that exposes the mismatch: content classified as a poor sidebar fit is reopened in the sidebar on the normal return path.

Carry the fullscreen entry context, prior sidebar-open state, and history action into the auto-dock decision. A browser `POP` back from a transient My Learning prose launch should end the transient session without opening a previously closed sidebar; interactive navigation should retain the existing deferred sidebar/floating auto-dock. Add regression tests for both transitions.

Relevant code: `src/components/Home/HomePanel.tsx:108-119`, `src/components/full-screen/FullScreenPanel.tsx:197-224`, `src/components/full-screen/full-screen-autodock.ts:46-100`, `src/global-state/panel-mode.ts:108-149`.

CI is green and the targeted tests pass, but these are semantic gaps at subsystem boundaries rather than failures exercised by the current suite.

moxious:
## Review findings — disposition

All six findings from the review are dispositioned. Findings 1–5 are fixed on this branch, one commit per finding with full rationale in the commit message:

| Finding                                                             | Disposition                                                                                                                                                                                  |
| ------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| 1 — validate prepared payloads at the event boundary                | Addressed in commit 5849d59b — payloads never cross the public event/queue; an opaque consume-once key redeems them from module-owned memory, misses fall back to the normal validated fetch |
| 2 — treat `code-block` as requiring the Grafana UI                  | Addressed in commit 8b9c47b6                                                                                                                                                                 |
| 3 — consume the pending guide in the occupied-sidebar floating path | Addressed in commit 74cfec0a                                                                                                                                                                 |
| 4 — bound snippet resolution so a launch cannot hang                | Addressed in commit 5c92ccdf                                                                                                                                                                 |
| 5 — cold-sidebar payload-relay boundary test                        | Addressed in commit c41bdbbe                                                                                                                                                                 |
| 6 — browser Back defeats the prose-only fullscreen decision         | **Explicitly deferred from PR scope** → #1448                                                                                                                                                |

Two further issues discovered while verifying the review are also **explicitly deferred from PR scope**: #1449, #1450.

leslie-heinzen (commented):
I don't have time to do a review with my fleshy eyeballs right now, but wanted to leave you an additional aggregated agent review before I peace out for the night. I will take a look myself tomorrow.

Findings by claude 4.7 and gpt-5.6:

## PR #1446 Aggregated Review

### Overall verdict
GPT-5.6 Sol: **Request changes** | Claude 4.7: **Approve with fixes**

Both reviewers agree the PR is well-constructed — the trust boundary design (`guideLaunchStore`, consume-once key, URL-binding) is correct and well-tested, and the `FloatingPanelInner` restoration-gate fix is legitimately good. The disagreement on disposition is driven by one finding GPT-5.6 caught that Claude 4.7 missed.

---

### Findings in BOTH reviews

**1. Transient-mode reload leak — the preference invariant breaks on page refresh** *(both rated Medium)*

Both reviewers independently identified that the PR's core guarantee — "an auto-launch never overwrites your panel-mode preference" — holds within a session but not across a reload. If a user is auto-launched to `/fullscreen`, reloads the page, and `_transientActive` is lost, the cold-load sync path persists `fullscreen` to `localStorage`, silently overwriting whatever the user had set.

- Claude 4.7 suggested: a URL query flag (`?panelMode=transient`) or a `sessionStorage` marker consumed by the cold-load sync.
- GPT-5.6 Sol noted this is also linked to the already-tracked issues #1448/#1449, suggesting explicit maintainer sign-off if deferring.

**2. Non-blocking telemetry gaps** *(both Low/non-blocking)*

Both flagged that failure/fallback paths in the new launch pipeline emit no countable, alertable event — only toasts. The specific gaps differ slightly but converge on the same area: `prepareGuideLaunch` failure paths and the unresolved-snippet silent-fallback to sidebar.

---

### High-severity finding: GPT-5.6 only (`ARCH-1446-1`)

**Interactive guides silently dropped when Pathfinder is already floating** *(High/Blocking — GPT-5.6 only)*

When the user already has Pathfinder open as a floating panel, clicking a My Learning guide triggers `setModeTransient('sidebar')`, which dispatches `pathfinder-auto-open-docs` — but the auto-open listener is only mounted by the **sidebar** renderer, not the floating root. The "is sidebar mounted" flag is `true` because the floating panel is mounted, so HomePanel treats it as ready and never calls `openSidebar`. The guide silently disappears, staged payload unconsumed.

Claude 4.7 did not catch this. GPT-5.6 verified it against the source with three skeptics. This is the main reason for the "Request changes" disposition.

---

### Medium finding: Claude 4.7 only

**`prepareGuideLaunch` has no cancellation on unmount** *(Medium — Claude 4.7 only)*

If a user clicks a guide and then navigates away before the 15s fetch deadline, the launch callback still fires and pushes the user to `/fullscreen?doc=…` from whatever page they landed on. Suggested fix: thread an `AbortSignal` through `prepareGuideLaunch` and abort on `MyLearningTab` unmount.

---

### Minor findings (one or both)

- **`guideLaunchStore` has no sweep for abandoned entries** (Claude 4.7 Low) — easy one-liner fix.
- **`pathfinder-auto-open-docs` event name scattered as raw string** (GPT-5.6 tech debt) — should be centralized in `event-names.ts` since the PR is extending its contract.
- **Contract test (`panel-mode-surface-toggles`) is a substring tripwire** (Claude 4.7) — brittle against renames, documented as intentional, but worth noting.

---

### Bottom line

The blocking issue to resolve before merge is `ARCH-1446-1` (floating→sidebar launch drops the guide). The reload preference leak is medium severity and the two reviewers converge on it independently — worth fixing or explicitly deferring with a tracked issue. Everything else is non-blocking.

src/components/Home/HomePanel.tsx
```diff
@@ -31,25 +53,93 @@ export class HomePanel extends SceneObjectBase<HomePanelState> {
...
+      panelModeManager.setPendingGuide(pendingGuideFrom(launch));
+      panelModeManager.setModeTransient('floating');
+      // A same-mode transient launch dispatches no mode-change event, so an
+      // already-mounted floating panel never remounts to consume. Signal it
+      // directly; unheard when the panel isn't up yet (mount consumes then).
+      document.dispatchEvent(new CustomEvent(REQUEST_FLOATING_GUIDE_EVENT));
+      return;
+    }
+
+    panelModeManager.setModeTransient('sidebar');
```
Jayclifford345, line 105: **Blocking.** This branch is reached when the user's own preference is `floating`, and it breaks the launch.

`isExtensionSidebarOwnedByOther` is `false` in floating mode — `module.tsx` calls `clearExtensionSidebarDocked()` when the persisted mode is `floating`/`fullscreen`, so nothing is docked. So a floating-preference user falls through to here and gets `setModeTransient('sidebar')`.

What happens next:

1. `FloatingPanelManager` returns `null` as soon as mode !== `'floating'`, so the floating panel unmounts.
2. Nothing opens the extension sidebar. `applyModeChange` for `'sidebar'` only reports surface-closed and dispatches the mode-change event — compare `handleSwitchToSidebar` in `FloatingPanelManager.tsx`, which deliberately pairs `setMode('sidebar')` with `sidebarState.openSidebar('Interactive learning')`.
3. The cold path below is skipped, because `sidebarState.getIsSidebarMounted()` is still `true` — the floating panel set it on mount, and `FloatingPanelInner`'s cleanup intentionally does *not* clear it when `getMode() === 'sidebar'`.
4. So `pathfinder-auto-open-docs` is dispatched into the docs panel that is one render away from unmounting.

The user ends up with no visible Pathfinder surface (the tab exists only in tab storage), and `isSidebarMounted` is now permanently `true` with nothing mounted, so subsequent auto-opens are dropped as well. On `main` this case worked: floating mode took the warm path and the guide opened in the floating panel.

It also contradicts the summary's third requirement — switching a floating user to the sidebar is exactly "trampling a preference the user set themselves".

Two ways out:

- Treat `panelModeManager.getMode() === 'floating'` as already-beside-Grafana and reuse the branch above (`setPendingGuide` + `REQUEST_FLOATING_GUIDE_EVENT`), which is already proven for the occupied-sidebar case; or
- if the sidebar is genuinely the intended target, mirror `handleSwitchToSidebar` and call `sidebarState.openSidebar('Interactive learning')` after the mode flip rather than gating on `getIsSidebarMounted()`.

The first seems right: a floating panel *is* beside Grafana, and it's what the user asked for.

src/components/Home/HomePanel.tsx
```diff
@@ -31,25 +53,93 @@ export class HomePanel extends SceneObjectBase<HomePanelState> {
...
     }
   }, []);
 
+  // Open full screen for reading-only content. Mirrors the sidebar→full-screen
+  // handoff order (pending guide → prior path → mode → route), but selects the
+  // mode transiently so the user's stored preference is untouched.
+  const openFullScreen = useCallback((launch: PreparedGuideLaunch) => {
+    panelModeManager.setPendingGuide(pendingGuideFrom(launch));
+    panelModeManager.capturePriorPath(window.location.pathname + window.location.search);
+    panelModeManager.setModeTransient('fullscreen');
```
Jayclifford345, line 143: This pushes a real, bookmarkable, reloadable URL, but transiency lives only in `_transientMode` / `_transientActive`, which don't survive a reload.

Reload (or back-forward, or opening the URL in a new tab) on `/fullscreen?doc=…`:

- `getMode()` falls back to localStorage → the stored preference, e.g. `sidebar`.
- `FullScreenPanel`'s reconcile effect sees `getMode() !== 'fullscreen'` and calls plain `setMode('fullscreen')`.
- `_transientActive` is `false` after the reload, so that write **persists** `fullscreen`.

The user's stored preference is silently replaced by the surface an auto-launch chose — the one thing "What does NOT change" promises won't happen. And reloading is especially likely here, since full screen is reserved for long-form reading content.

The route is the only state that survives the reload, so it probably has to carry the transiency: e.g. stamp the launch URL and have the reconcile effect use a non-persisting write when it is only aligning mode to the route it already landed on. Worth pinning with a test that reloads (fresh manager + preserved localStorage + fullscreen route) and asserts the stored preference is intact.

src/components/LearningPaths/MyLearningTab.tsx
```diff
@@ -293,6 +303,33 @@ export function MyLearningTab({ onOpenGuide }: MyLearningTabProps) {
...
+  // Fetch + snippet-expand + classify the target, then hand the prepared
+  // launch to the host so it can pick the surface without re-fetching. The
+  // fetch happens while My Learning stays mounted; on failure My Learning stays
+  // visible and the error is surfaced rather than committing a surface.
+  const launch = useCallback(
+    async (url: string, title: string) => {
+      if (launchInFlightRef.current) {
+        return;
+      }
+      launchInFlightRef.current = true;
```
Jayclifford345, line 326: The click is now silent for the whole prepare step. `fetchContent` and the snippet CDN fetches are each bounded by `DEFAULT_CONTENT_FETCH_TIMEOUT` (10s), so on a slow or unreachable CDN the card can be dead for ~10s and then show an error — and this ref swallows repeat clicks with no disabled or pending affordance, so the user's natural response (click again) also does nothing.

Before this PR the click opened the sidebar immediately and the panel showed its own loading state; the surface decision has moved that feedback off the critical path. Keep the ref as the correctness guard, but add a pending state on the launching card so the UI acknowledges the click.

src/components/LearningPaths/MyLearningTab.tsx
```diff
@@ -293,6 +303,33 @@ export function MyLearningTab({ onOpenGuide }: MyLearningTabProps) {
...
+      try {
+        const result = await prepareGuideLaunch(url, { title, source: 'home_page' });
+        if (result.ok) {
+          onOpenGuide(result.launch);
+        } else {
+          getAppEvents().publish({
+            type: 'alert-error',
+            payload: ['Could not open the guide', result.error],
```
Jayclifford345: User-facing strings, not wrapped in `t()` — this file already imports `t` from `@grafana/i18n`. Compare `useContentReset.ts:90`, which translates both the title and the body of its `alert-error`.

`result.error` also reaches the user verbatim; some of those strings are internal-shaped (`'Guide content could not be parsed'`, `'Failed to load content'`). Worth a translated generic message with the raw error kept to the logs.

src/global-state/guide-launch.ts
```diff
@@ -0,0 +1,83 @@
...
+class GuideLaunchStore {
+  private _staged = new Map<string, { payload: StagedLaunchPayload; stagedAt: number }>();
+
+  /**
+   * Stage a prepared payload and get back the opaque key to send over the
+   * public channel. Keys are single-use.
+   */
+  public stage(payload: StagedLaunchPayload): string {
+    const key = crypto.randomUUID();
+    this._staged.set(key, { payload, stagedAt: Date.now() });
```
Jayclifford345: `consume()` is the only eviction path, so an entry that is never redeemed lives for the life of the page — holding a full snippet-expanded guide. `STALE_AFTER_MS` is checked at redemption, which is exactly the moment that never comes for a leaked entry.

Unredeemed staging is reachable: a cold sidebar that the user never lets mount, navigating away before the queue drains, and (today) the floating-preference path in `HomePanel` where the event has no live listener.

`MyLearningTab`'s in-flight guard means at most one key is ever useful at a time, so the fix is cheap — drop expired entries (or just clear the map) at the top of `stage()`. That also keeps the store's memory bounded by construction rather than by the redeemer's good behaviour.

src/components/docs-panel/utils/requires-grafana-ui.ts
```diff
@@ -0,0 +1,79 @@
...
+      return blocksRequireGrafanaUi(block.blocks);
+    case 'conditional':
+      return blocksRequireGrafanaUi(block.whenTrue) || blocksRequireGrafanaUi(block.whenFalse);
+    case 'code-block':
+      // `reftarget` is schema-required and targets a live Grafana Monaco
+      // editor; without the Grafana UI both "Show me" and "Insert" are dead.
+      return true;
+    case 'snippet-ref':
+      return true;
+    default:
```
Jayclifford345: The switch covers every current member of the `JsonBlock` union — I checked — but `default: return false` means the next container block type added to that union classifies as non-interactive silently, and any Grafana-driving action nested inside it gets hidden in full screen with dead "show me / do it" buttons.

The failure is invisible: no type error, no test failure, just a wrong surface. Since the module's whole premise is "computed from content can never drift", the drift risk deserves a compile-time guard — an `assertNever(block)` / `const _exhaustive: never = block` in the default arm turns a future `JsonBlock` addition into a build failure that points straight here.

src/components/docs-panel/utils/prepare-guide-launch.ts
```diff
@@ -0,0 +1,109 @@
...
+
+interface PrepareGuideLaunchContext {
+  title: string;
+  source: LaunchSource;
+  /** Pre-resolved package context (recommender path); otherwise derived from the URL. */
+  packageInfo?: PackageOpenInfo;
+}
+
+/** Mirrors the routing predicate in `useAutoOpenListener` so the surface handoff picks the same open method. */
+function isLearningJourneyUrl(url: string): boolean {
```
Jayclifford345: The doc comment says this mirrors `useAutoOpenListener`, but the two predicates differ: the listener runs the URL through `parseUrlSafely` and tests `urlObj?.pathname`, this one does a raw substring test on the whole URL.

So `https://grafana.com/docs/foo/?ref=/learning-paths/x` is a journey here and a docs page there. `type` is what `pendingGuideRouter` uses to pick `openLearningJourney` vs `openDocsPage`, so the divergence routes through the wrong loader on the full-screen and floating handoffs while the sidebar path gets it right.

Narrow in practice, but it's free to fix and the "mirrors" claim should be true: reuse `parseUrlSafely` (or export the listener's predicate and call it from both).

src/components/docs-panel/docs-panel.tsx
```diff
@@ -723,11 +733,17 @@ class CombinedLearningJourneyPanel extends SceneObjectBase<CombinedPanelState> i
...
+      // appears. See package-info-from-url.ts for the URL pattern. Skipped for
+      // prefetched launches — `prepareGuideLaunch` already derived it.
+      if (!prefetched && !packageInfo && isPackageContentUrl(url)) {
         packageInfo = await fetchPackageInfoFromUrl(url);
       }
-      const result = await loadDocsTabContentResult(url, { skipReadyToBegin, packageInfo });
+      // Prefetched content skips the network fetch (one-fetch launch) but runs
+      // the identical finalization below so alignment / journey / package
+      // parity holds.
+      const result = prefetched
```
Jayclifford345, line 750: `skipReadyToBegin` is silently ignored when `prefetched` is set — the argument is still accepted, but the only consumer is the `loadDocsTabContentResult` call it now skips. `prepareGuideLaunch` never passes it either, so prepared journey content always carries the ready-to-begin extras that `generateJourneyContentWithExtras` injects.

No caller hits the combination today (the pending-guide router and the auto-open listener both leave it unset), so this is a latent trap rather than a live bug: the next caller that passes both gets a silently wrong render with no signal. Either thread it into `prepareGuideLaunch` so the prepared content is built the same way, or make the incompatibility explicit rather than dropping the option on the floor.

Jayclifford345 (commented):
## Review round 2

Re-reviewed against the current head (`dabbc010`). The trust-boundary hardening from round 1 holds up — the opaque-key staging, the consume-once redemption with URL binding, and the deadline on the snippet CDN fetches all do what the summary claims, and the classifier's block coverage is complete over today's `JsonBlock` union.

One blocking finding, one that breaks a stated invariant, and a handful of smaller ones. Details inline; summary here.

### Blocking

**1. A user whose preference is `floating` loses the guide entirely.** `openBesideGrafana` only asks `isExtensionSidebarOwnedByOther`. When the persisted preference is `floating`, nothing is docked (module.tsx clears the docked key in floating mode), so that check is `false` and the launch forces `setModeTransient('sidebar')`. That unmounts `FloatingPanelManager`, and nothing opens the extension sidebar — `applyModeChange` for `sidebar` only dispatches the mode-change event, and the cold-path `openSidebar(...)` is skipped because `getIsSidebarMounted()` is still `true`. The auto-open event lands in the panel that is about to unmount. Net result: no visible Pathfinder surface, and `isSidebarMounted` is stuck `true` so later auto-opens are dropped too. On `main` this case worked (warm path → guide opens in the floating panel), so it's a regression, and it tramples the same preference the summary says is never trampled.

### Invariant

**2. Reloading a transient full-screen guide persists `fullscreen`.** `openFullScreen` pushes a real, reloadable `/fullscreen?doc=…` URL but records transiency in memory only. After a reload `FullScreenPanel`'s reconcile effect ([FullScreenPanel.tsx:64](https://github.com/grafana/grafana-pathfinder-app/blob/dabbc01015f08277477a092c6b5378b9c1a8cef1/src/components/full-screen/FullScreenPanel.tsx#L64)) sees a mode mismatch and calls plain `setMode('fullscreen')` outside any round-trip — which writes localStorage. "Your saved panel-mode preference is never overwritten by an auto-launch" stops holding the moment the user reloads, and reloading is especially likely on the long prose content this feature exists to serve.

### Smaller

3. No pending UI while the launch is prepared (inline).
4. `guideLaunchStore` has no sweep — unredeemed keys retain a full expanded guide for the page's lifetime (inline).
5. Classifier `default` arm has no exhaustiveness guard (inline).
6. `isLearningJourneyUrl` claims to mirror the listener's predicate but doesn't (inline).
7. `skipReadyToBegin` is silently dropped on the prefetched path (inline).
8. The new error alert is hardcoded English; compare `useContentReset`, which wraps both strings in `t()` (inline).

### Testing

Coverage of the new surfaces is genuinely good, but two gaps line up with the findings above: `HomePanel.test.tsx` mocks `panelModeManager` and `sidebarState` wholesale, so neither the floating-preference launch nor "who actually opens the sidebar" is exercised; and `panel-mode.test.ts`'s no-op case (`it('is a no-op when target mode equals current mode')`) asserts only that no event publishes — it doesn't pin that localStorage is untouched, which is now the load-bearing half of `applyModeChange` since `commit()` moved above the early return.

### Instrumentation

The prepare-failure path surfaces a user alert but has no typed facade event or stable `src/lib/logging.ts` signal, so a CDN degradation that makes launches fail is invisible in aggregate — that's the countable-fallback rule in `docs/developer/TELEMETRY.md`. The surface decision itself is also uncountable: `reportPathfinderSurface` fires for the full-screen and floating branches (via `applyModeChange`) but not for the sidebar branch, and nothing records *why* a surface was chosen, so "is the classifier picking the right surface in the wild?" can't be answered from data.

### Comment hygiene (non-blocking)

`AGENTS.md` says default to no comments, with a narrow keep-list. Several new modules here are majority prose, and some of it is the shape the guide explicitly lists as bad — narration of process artifacts: `pendingGuideRouter.ts`'s "the floating panel shipped exactly that bug once … (PR #1446 review finding 3)" and the `describe('… (PR #1446 review finding 4)')` block in `online-snippet-resolver.test.ts`. Those read as review-response bookkeeping that a future reader can't act on; the git history already holds them. The invariant comments (why `liveTabs` instead of the render snapshot, why the payload never rides the public channel) are on the keep-list and should stay.

moxious:
Fix pass for both reviews pushed (`dabbc010..ffa38664`). Every finding was independently re-verified against the code before fixing; all confirmed except two that were partially overstated (notes at the end). Mapping:

**Blocking — floating-preference user loses the guide** (@Jayclifford345 #1, = the aggregated review's ARCH-1446-1) → `ddea4a65`. Took the first suggested route: a current floating surface is treated as already-beside-Grafana and reuses the proven `setPendingGuide` + `REQUEST_FLOATING_GUIDE_EVENT` branch. New HomePanel test pins it. One framing correction: this wasn't a clean regression — on `main` the same click also dispatched `pathfinder-auto-open-docs` into the void (`useAutoOpenListener` only ever mounts in the sidebar renderer; the floating panel renders `ContentRenderer` directly). That listener-ownership gap predates the PR and is tracked in #1450. What this PR *added* was the damage: tearing down the floating panel and leaving `isSidebarMounted` stuck true. Both halves of the added damage are gone; the pre-existing link-interception variant stays with #1450.

**Invariant — reloading a transient full-screen guide persists `fullscreen`** (Jay #2, also in the aggregated review) → `596159e0`. The `FullScreenPanel` reconcile effect now writes via `setModeTransient`: it only fires on a cold load/reload of `/fullscreen` where route and mode disagree, which is never a preference expression. No URL stamping needed — this also correctly reopens the round-trip so the eventual exit doesn't persist either. Added the suggested reload test (fresh manager + preserved localStorage + fullscreen route → stored preference intact) plus a contract-test entry pinning the call site. The structural surface/preference split stays deferred to #1449.

**Smaller findings**
- No pending UI (Jay #3) → `9f8814b5`. Launching card shows a spinner + "Opening…", continue buttons disable while a launch is in flight. Also fixed the adjacent issue from the aggregated review (no cancellation on unmount): a launch resolving after My Learning unmounts is dropped instead of yanking the user to `/fullscreen` from wherever they navigated. Both pinned by new MyLearningTab tests.
- Store sweep (Jay #4) → `89f3217f`. `stage()` evicts expired entries; memory bounded by construction.
- Classifier exhaustiveness (Jay #5) → `9feb7cfd`. All 19 `JsonBlock` members enumerated; the default arm is a compile-time `never` guard with a runtime fail-safe toward the sidebar (unknown CDN block types fail in the never-hide-an-action direction, matching the module's own rule).
- Predicate divergence (Jay #6) → `74ac8eaa`. One shared pathname-based `isLearningJourneyUrl` in `url-validation.ts`, used by both the listener and `prepareGuideLaunch`; the `?ref=/learning-paths/x` case is a test now.
- `skipReadyToBegin` dropped (Jay #7) → `ffa38664`. Latent (no caller passes both), so the conflict now logs a warning instead of threading the flag through — the prepared-content path can't honor it by construction.
- i18n (Jay #8) → `935ec152`. Title + generic body wrapped in `t()`; the internal-shaped raw error goes to the logger instead of the toast.
- Comment hygiene → `ffa38664`. The PR-review bookkeeping refs are trimmed from comments and test titles (all of them, not just the two named); the invariant comments stay.
- Event-name centralization (aggregated review) → `ffa38664`. `AUTO_OPEN_DOCS_EVENT` in `event-names.ts`; the tripwire now pins the constant's *value* as the wire contract.

**Instrumentation** (Jay) → `935ec152`, scoped after verification. Two of the three claims were narrower than stated: the fetch ladder already emits typed facade events with ok/error outcomes (`recordContentFetch` + fallback events + Faro-bridged logger errors), and the sidebar surface *is* counted — at its mount (`module.tsx`), per the documented convention in `panel-mode.ts`. What was genuinely silent and is now covered: the JSON-parse failure branch in `prepareGuideLaunch` (logger signal), launch-level failure logging, and a new `guide_launch_surface_chosen` analytics event carrying `surface` + `requires_grafana_ui` + `content_url` — that's the "is the classifier right in the wild?" question made answerable.

**Testing** (Jay) — the no-op `setMode` test now also asserts the stored *value* is unchanged (a literal "untouched" assertion would fail by design: `commit()` legitimately rewrites the same value outside a round-trip). Note the dangerous variant you were worried about — no-op `setMode` during a transient round-trip clobbering the preference — was already pinned at `panel-mode.test.ts` (`it.each` sidebar case). The HomePanel mock-boundary observation: the floating *fallback* branch was already exercised; what didn't exist was the floating-*preference* branch, because the code branch itself was missing (finding 1). It exists and is tested now.

moxious:
**Scoping note: unresolved snippets and surface classification — deliberately out of scope**

Recording a decision from the review discussion about snippet resolution failing at guide-open time.

First, the mechanics: the classifier already fails safe. `prepareGuideLaunch` computes `requiresGrafanaUi(expandedGuide) || unresolvedSnippetIds.length > 0`, so a snippet that fails to resolve can only push a guide *toward* the sidebar. The scary direction — interactive steps hidden behind a failed snippet, guide sent full screen with dead buttons — cannot occur. The entire residual is the harmless direction: a reading-only guide with a failed snippet renders in the sidebar instead of full screen.

We are not engineering beyond that fail-safe, because:

(a) Snippets are a reusability mechanism, not a primary content channel — most guides don't use them, and interactive guides that do typically have interactive elements outside the snippet too, so classification usually lands right even when resolution fails.

(b) When resolution fails, the guide renders with a broken placeholder — the user-visible problem is the content, not which surface it's on. Failed-resolution UX is its own thread of work, separate from this PR.

(c) The failure window is dominated by runtime resolution (transient network, CDN drift): snippet bodies are schema-validated at PR time in interactive-tutorials (`build-snippets` exits non-zero on any invalid body). One known gap: guide→snippet *references* aren't cross-checked against the catalog at publish time, so a dangling `snippetId` can ship — but under the fail-safe its consequence is identical to a transient failure (sidebar + placeholder). Adding a catalog cross-check to `pathfinder-cli validate` is cheap and tracked separately.

Net: the worst case is prose rendering in the sidebar next to a visibly broken snippet placeholder. Esoteric enough that we're consciously not spending more of this PR on it.

Jayclifford345:
## Review round 3 — looks good to me

Re-checked all eight of my round-2 findings against `b4f5a765`, plus the two the aggregated review raised. All resolved, and the two you pushed back on were fair pushback — the no-op `setMode` test genuinely can't assert "untouched", and `content_url` is normalized at the Faro boundary by the `*_url` convention, so the privacy invariant holds.

The two I cared most about I traced properly rather than taking the commit message for it:

- **Floating-preference launch** — `module.tsx` mounts `FloatingPanelManager` eagerly when the persisted mode is `floating`, and `FloatingPanelInner` listens for `REQUEST_FLOATING_GUIDE_EVENT`, so the same-mode transient launch (which fires no mode-change event) still gets consumed. Test asserts `setModeTransient('sidebar')` is never reached.
- **Reload invariant** — walked all four route/mode combinations. Transient launch + reload leaves localStorage alone; a deliberate `setModePersisted('fullscreen')` falls into the `else` branch, so nothing regresses.

CI is green on the head commit, e2e matrix included.

Two things to note, neither blocking:

- **The new `t()` keys aren't registered.** `myLearning.launchErrorTitle` / `myLearning.launchErrorMessage` aren't in `src/locales/en-US/grafana-pathfinder-app.json` (no locale file is touched in the PR). Nothing breaks — the inline default renders and there's no CI gate — but the key it was modelled on, `docsPanel.resetGuideErrorTitle`, is registered, so this is drift we otherwise keep clean. One `i18n-extract` away, either here or as a follow-up.
- **A behaviour change worth recording.** The reload fix means a cold bookmarked `/fullscreen?doc=…` load no longer persists `fullscreen` as the preference — on `main` it did. I think that's the right reading (a route isn't a preference expression), and `?panelMode=fullscreen` still persists via the deep-link handler. Just flagging it because "What does NOT change" says cold deep links behave exactly as before, and that's now true of the surface but not of the persistence side effect.

Also: `'Opening…'` on the continue button is hardcoded English, but so are `Start` / `Continue` / `Restart` in that file, so it matches its surroundings — not worth a commit on its own.

Nice work on the dispositions, the one-commit-per-finding split made this much faster to verify.

Jayclifford345 (approved):

PR #1447, opened 2026-07-28
feat(docs-panel): tab-bar chrome — Pathfinder wordmark + My learning tab (#1256)

Spidy88:
PR 6 (final) of the block-editor header redesign stack (issue #1256, change 7). **Independent of the editor header** — branches off `main`.

## What this does

Adds the **"Interactive Learning" wordmark** (the plugin's external name) + a vertical divider to the docs-panel tab bar. The wordmark hides below a 360px container width so branding yields first when the bar is narrow.

`container-type: inline-size` (the wordmark's query container) makes the tab bar a stacking context, so the bar takes `theme.zIndex.navbarFixed` to keep the tab-overflow dropdown painting above page content (below modals).

## Scope note — reduced from the original change-7 plan

Parallel work is removing the editor/devtools permanent tabs (leaving a single, non-dismissable Recommendations tab, so no in-panel navigation is needed) and separately owns the "My learning" affordance. To avoid conflicts and wasted effort, this PR was reduced to **just the wordmark + divider** and leaves the following to that parallel work:

- The recommendations tab's icon and title — a new "recommendations" icon is wanted (TBD; a book icon would conflate it with My learning), so it's left as-is here.
- The "My learning" book button and its placement (stays right-justified per design; not moved into a kebab).
- The tab-overflow reserve calc.

## Verification

- Unit (`DocsPanelTabBar` + contract), lint, prettier, `tsc` — green.
- The wordmark's responsive hide is a container query (not unit-testable in jsdom); the unit test asserts presence.

## Screenshots
pathfinder-header-wordmark

### DIFF
--- a/src/components/docs-panel/components/DocsPanelTabBar.test.tsx
+++ b/src/components/docs-panel/components/DocsPanelTabBar.test.tsx
@@ -80,6 +80,13 @@ describe('DocsPanelTabBar', () => {
     expect(props.onSetActiveTab).not.toHaveBeenCalled();
   });
 
+  // jsdom can't evaluate the @container query that hides the wordmark at narrow
+  // widths, so this asserts presence only — the responsive hide isn't unit-testable.
+  it('renders the "Interactive Learning" wordmark', () => {
+    render(<DocsPanelTabBar {...makeProps()} />);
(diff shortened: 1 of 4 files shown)


Spidy88:
@cursoragent review

moxious (approved):

PR #1455, opened 2026-07-29
Support private CDN guide URLs

jdbaldry:
Add support for private interactive-learning CDN content.

- `pkg/plugin/resources.go`: registers `POST /cdn/sign` on the app plugin resource router.
- `pkg/plugin/cdn_sign.go`: adds HMAC URL signing with path and TTL validation.
- `pkg/plugin/settings.go`: reads `cdn_private_base_url`, `cdn_signing_key_id`, and `cdn_signing_secret` from secure settings.
- `src/constants.ts`: allowlists private interactive-learning hostnames.
- `src/utils/find-doc-page.ts`: accepts private hosts through `isInteractiveLearningUrl` validation instead of a hardcoded substring gate.
- `docs/developer/LOCAL_DEV.md`: documents the new signer flow with session-cookie auth and real key retrieval from `deployment_tools` Terraform state.

I verified the signer against the real dev Fastly verifier using the current `key1` secret from `deployment_tools` Terraform state. To test yourself:

1. Start local Grafana with `npm run server`.
2. Upload fixture files to `gs://interactive-learning-dev-private/internal/e2e/<ID>/`.
3. Log into local Grafana and saved signer settings on `grafana-pathfinder-app`.
4. Mint a signed URL from `POST /api/plugins/grafana-pathfinder-app/resources/cdn/sign`.
5. Confirm status codes:
   - Signed URL returns `200`.
   - Same unsigned URL returns `403`.
   - Tampered signature returns `403`.
   - Expired token returns `403`.
6. Open `http://localhost:3000/a/grafana-pathfinder-app?doc=<URL ENCODED SIGNED URL>` and confirm the guide loads.

More detail is in the `LOCAL_DEV.md` documentation

### DIFF
--- a/.config/Dockerfile
+++ b/.config/Dockerfile
@@ -14,7 +14,7 @@ ENV DEV "${development}"
 # Do NOT enable these settings in a public facing / production grafana instance
 ENV GF_AUTH_ANONYMOUS_ORG_ROLE "Admin"
 ENV GF_AUTH_ANONYMOUS_ENABLED "${anonymous_auth_enabled}"
-ENV GF_AUTH_BASIC_ENABLED "false"
+ENV GF_AUTH_BASIC_ENABLED "true"
 # Set development mode so plugins can be loaded without the need to sign
 ENV GF_DEFAULT_APP_MODE "development"
 
--- a/.config/docker-compose-base.yaml
+++ b/.config/docker-compose-base.yaml
@@ -19,7 +19,7 @@ services:
 
     environment:
       NODE_ENV: development
-      GF_LOG_FILTERS: plugin.grafana-pathfinder-app:debug
-      GF_LOG_LEVEL: debug
+      GF_LOG_FILTERS: plugin.grafana-pathfinder-app:info
+      GF_LOG_LEVEL: info
       GF_DATAPROXY_LOGGING: 1
       GF_PLUGINS_ALLOW_LOADING_UNSIGNED_PLUGINS: grafana-pathfinder-app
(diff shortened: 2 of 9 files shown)


.config/Dockerfile
```diff
@@ -14,7 +14,7 @@ ENV DEV "${development}"
 # Do NOT enable these settings in a public facing / production grafana instance
 ENV GF_AUTH_ANONYMOUS_ORG_ROLE "Admin"
 ENV GF_AUTH_ANONYMOUS_ENABLED "${anonymous_auth_enabled}"
-ENV GF_AUTH_BASIC_ENABLED "false"
+ENV GF_AUTH_BASIC_ENABLED "true"
```
jdbaldry, line 17: Can drop this, this was just for my convenience in testing

.config/docker-compose-base.yaml
```diff
@@ -19,7 +19,7 @@ services:
 
     environment:
       NODE_ENV: development
-      GF_LOG_FILTERS: plugin.grafana-pathfinder-app:debug
-      GF_LOG_LEVEL: debug
+      GF_LOG_FILTERS: plugin.grafana-pathfinder-app:info
+      GF_LOG_LEVEL: info
```
jdbaldry, line 23: Can also drop this because it was just a bit more logging than I needed

Jayclifford345:
Hey @jdbaldry, just checking before review again are you still going ahead with the private cdn?

jdbaldry:
No need to review this in its current state. As we discussed before in our meetings, the GAP migration means that we no longer need this machinery to enable private guides and I've not been given any concrete use case for the asset server CDN yet so this has been broadly deprioritized. 

If you're keen to keep the repo tidy, I can close this for now and track elsewhere :)

Issue #1460, opened 2026-07-29
Unify Pathfinder's three parallel guide-launch channels into one surface-agnostic entry point


moxious:
**Summary**

Pathfinder launches a guide onto a surface through **three parallel channels** with different delivery semantics, and that divergence is a recurring bug source:

1. `AUTO_OPEN_DOCS_EVENT` — a **fire-and-forget** `document` CustomEvent. Dispatched by global link interception (`link-interception.ts`), `HomePanel`'s mounted-surface launch branch, and the cold-sidebar queue drain. If no live listener is registered when it fires, the launch is silently lost.
2. `REQUEST_FLOATING_GUIDE_EVENT` — a floating-only signal consumed via `consumePendingGuideOnMount`.
3. `autoLaunchChannel` — a **latched** channel (`emit` stores the value and replays it on the next surface mount). Consumed on every surface via `useAutoLaunchTutorial`, so it is surface-agnostic and cannot fire into the void.

**Why this is a problem**

Because channel (1) is fire-and-forget and (until #1450) was only listened to on the sidebar, dispatching it while another surface owned the display dropped the launch — the class of bug fixed in #1450 (a docs-link click doing nothing when the floating/fullscreen panel owns the surface). #1450 closed the immediate instance by mounting the auto-open listener on all three surfaces (mode-gated), but that leaves three channels with three delivery models, each of which a future dispatcher must reason about correctly. New launch entry points have to pick the right channel and the right mount-timing or silently regress.

**Proposed direction**

Collapse the three into **one surface-agnostic, latched launch entry point** — a single `global-state` API (e.g. `surfaceLaunch.openDocs(url, title, { source, launchKey })`) that:

- routes to whatever surface currently owns the display, and
- latches when no surface is mounted yet, replaying to the first surface that mounts (the `autoLaunchChannel` model),

so no dispatcher can ever fire a launch into the void, and no dispatcher needs mode-awareness. Migrate `link-interception`, `HomePanel`, the deep-link handler, the highlighted-guide orchestrator, and the cross-tab live-tab executor onto it, then retire `AUTO_OPEN_DOCS_EVENT` and `REQUEST_FLOATING_GUIDE_EVENT`.

**Relationship**

- Spun out of the #1450 review/scoping as the structural endgame; #1450 is the tactical fix.
- Overlaps the `panel-mode` / surface-persistence area, so it should be sequenced against that work rather than run concurrently.

moxious:
[...]

PR #1461, opened 2026-07-29
fix(docs-panel): mount auto-open docs listener on all launch surfaces

### LINKED ISSUE #1450: Auto-open events have no listener when the floating panel owns the surface
**Summary**

`useAutoOpenListener` — the handler for the `pathfinder-auto-open-docs` event — is mounted only by the sidebar renderer (`src/components/docs-panel/docs-panel.tsx:955`). But `sidebarState.isSidebarMounted` is set to `true` by _all three_ surfaces: the sidebar, the floating panel (`src/components/floating-panel/FloatingPanelManager.tsx:118`), and the full-screen panel (`src/components/full-screen/FullScreenPanel.tsx:88`) — the flag means "some Pathfinder surface is mounted," not "the auto-open listener is registered." Dispatch paths that gate on that flag therefore fire events into the void whenever the floating panel owns the surface: global link interception (`src/global-state/link-interception.ts:33-38`) and HomePanel's mounted-sidebar launch branch (`src/components/Home/HomePanel.tsx:88`).

There is also a race variant introduced by the #1446 launch path: a user whose current surface is floating clicks an interactive guide on My Learning. `openBesideGrafana` calls `setModeTransient('sidebar')` (`HomePanel.tsx:79`), which starts unmounting the floating panel — but `getIsSidebarMounted()` still reads `true` at `HomePanel.tsx:88` (the floating panel's cleanup deliberately leaves the flag set during dock-back, `FloatingPanelManager.tsx:132-136`), so the event is dispatched toward a model that is being torn down, and `sidebarState.openSidebar` is never called.

Predates #1446 in its link-interception form (the flag overload exists on `main`); surfaced during the #1446 review verification pass. Fixing it likely means either mounting the auto-open listener on every surface's model, or splitting the overloaded flag into "surface mounted" vs "listener ready" — either way it is its own scoped change.

**Use cases / user flows impacted**

1. _Swallowed docs link._ A user is working with the floating panel open (their preferred surface, or because another plugin owns the extension sidebar). They click a grafana.com docs link anywhere in the Grafana UI. Link interception `preventDefault()`s the click (`link-interception.ts:22`), sees "mounted," and dispatches the auto-open event — which nothing is listening to. Net effect: the click does _nothing at all_. The user doesn't even get the browser navigation they would have had with Pathfinder uninstalled.
2. _My Learning launch into a dying panel._ The race above: a floating-surface user clicks an interactive guide on My Learning and the guide never appears — the tab may be written to storage by the unmounting model, so it can ghost into existence on the _next_ panel open, which reads as "Pathfinder randomly restored a guide I never opened."

Found while verifying the review of #1446 (adjacent to finding 3); out of scope there because it predates the PR and needs its own listener-ownership decision.

moxious:
## Intent

Implement follow-up issue #1450 from the merged #1446 launch-surface work: the auto-open docs listener (AUTO_OPEN_DOCS_EVENT, a fire-and-forget document CustomEvent) was mounted ONLY on the sidebar renderer, so a global docs-link click intercepted while the floating panel or the full-screen page owned the surface was dispatched into the void — the click did nothing at all (worse than an uninstalled Pathfinder). This was a live user-facing bug that predates #1446.

DECISIONS LOCKED WITH THE USER (each deliberate, do not flag as a mistake):
- Fix strategy = Option A: mount useAutoOpenListener on the floating AND full-screen surfaces too, symmetric with the existing useAutoLaunchTutorial which is already mounted on all three. NOT the 'split the flag' alternative, which is insufficient on its own (the fallback openSidebar() is a no-op in floating/fullscreen mode).
- Mode-gate (decision 3a): useAutoOpenListener now takes a (default 'sidebar' to keep the ~15 existing direct-hook test call sites green) and the handler acts only when panelModeManager.getMode() === surface. This prevents double-open during a dock-back when the old and new surfaces are briefly mounted at once. This mode-gate is INTENTIONAL, not dead code.
- Scope (decision 4) = BOTH floating and full-screen (full screen has the identical gap: own model, no listener, sets isSidebarMounted).
- Decision 2 = do NOT split/rename the overloaded isSidebarMounted flag in this PR. Option A makes its existing meaning ('a surface is mounted') also imply 'a listener is ready', so its 8 consumers stay correct untouched. I added only a clarifying comment. Renaming would touch many files and is deliberately out of scope (keeps this PR disjoint from an in-flight panel-mode PR).
- Decision 5 = confirm-and-lock: the related 'launch into a dying panel' race (issue point 2) was already resolved by #1446's HomePanel floating-branch fix; it stays covered by the existing HomePanel test 'delivers to the floating panel when it is the current surface'. I did not reopen it or add a duplicate test.
- Decision 6 = defer the structural unification of the three parallel launch channels; filed as follow-up issue #1460 and referenced in the commit.

TESTS: new useAutoOpenListener.surface-ownership.test.ts proves the floating and full-screen listeners open an intercepted link, the mode-gate skips a non-current surface, and a dock-back overlap opens exactly once (only the current-mode surface's model). Existing 30 auto-open/HomePanel tests still pass (they use the default 'sidebar' surface, whose gate passes under getMode()'s default).

ARCHITECTURE NOTES for the reviewer: useAutoOpenListener lives in docs-panel/hooks and is now imported by floating-panel and full-screen renderers — both are Tier-4 components importing a same-tier hook, already an established pattern in those files (they import CombinedLearningJourneyPanel, pendingGuideRouter, useContentReset from docs-panel). No tier violation. The diff is intentionally minimal (5 source files, 1 new test) and disjoint from the panel-mode persistence PR.

## What Changed

- `useAutoOpenListener` now takes a `surface` parameter (default `'sidebar'`) and only acts when `panelModeManager.getMode()` matches, so an intercepted `AUTO_OPEN_DOCS_EVENT` is handled by exactly one surface even during a dock-back overlap when two surfaces are briefly mounted.
- Mounted the listener on the floating-panel and full-screen renderers in addition to the sidebar, closing the gap where a global docs-link click was dispatched into the void when the floating panel or full-screen page owned the surface; added a clarifying comment on the still-overloaded `isSidebarMounted` flag in `global-state/sidebar.ts`.
- Added `useAutoOpenListener.surface-ownership.test.ts` proving each surface opens an intercepted link, the mode-gate skips a non-current surface, and a dock-back overlap opens exactly once.

## Risk Assessment

✅ Low: A minimal, symmetric five-file fix with a new test that mounts an existing listener on two more surfaces behind an intentional, verified mode-gate; every cold-relay dispatcher aligns mode before queuing so no legitimate launch is dropped, and it conforms to all locked decisions in the intent.

## Testing

Installed dependencies (node_modules was missing in this worktree — a setup fix, not a product issue), then ran the new surface-ownership suite plus the existing auto-open, cold-sidebar-relay, event-contract, and HomePanel suites: 34 tests pass. The surface-ownership scenarios exercise the intent end-to-end by dispatching the actual `pathfinder-auto-open-docs` DOM CustomEvent the live link interceptor fires and asserting the real model opens the doc — proving the floating and full-screen listeners now open an intercepted link (the pre-fix click did nothing at all), the mode-gate skips a non-current surface, and a dock-back overlap opens exactly once across two mounted models. No visual GIF/screenshot was captured because this change has no rendered surface of its own: it is pure event-listener wiring whose visible outcome (a doc opening in a panel) is identical to the already-working sidebar path, and the key dock-back double-open race cannot be reliably triggered by hand — capturing it would require a full Grafana Docker build plus manual surface-ownership orchestration that the deterministic tests already reproduce faithfully at the DOM-event boundary. Worktree left clean (node_modules is gitignored).

<details>
<summary>Evidence: Auto-open surface-ownership scenarios — per-test pass transcript</summary>

<code>useAutoOpenListener surface ownership (#1450):&#10;[PASSED] the floating listener opens an intercepted link when floating owns the surface&#10;[PASSED] the full-screen listener opens an intercepted link when full screen owns the surface&#10;[PASSED] a listener ignores events when it is not the current surface (mode-gate)&#10;[PASSED] during a dock-back overlap, exactly one surface handles the event&#10;HomePanel (locked #1446 floating-branch coverage):&#10;[PASSED] delivers to the floating panel when it is the current surface, instead of tearing it down&#10;TOTAL: 24 passed / 24 across 4 suites (plus useAutoOpenListener.test.ts default-surface suite = 34 total)</code>

```text
Auto-open listener surface-ownership — end-to-end behavioral scenarios (#1450)

• components/docs-panel/hooks/useAutoOpenListener.surface-ownership.test.ts
   [PASSED] useAutoOpenListener surface ownership (#1450) the floating listener opens an intercepted link when floating owns the surface
   [PASSED] useAutoOpenListener surface ownership (#1450) the full-screen listener opens an intercepted link when full screen owns the surface
   [PASSED] useAutoOpenListener surface ownership (#1450) a listener ignores events when it is not the current surface (mode-gate)
   [PASSED] useAutoOpenListener surface ownership (#1450) during a dock-back overlap, exactly one surface handles the event

• components/docs-panel/hooks/useAutoOpenListener.cold-sidebar-relay.test.ts
   [PASSED] cold-sidebar relay: queued prepared launch reaches the model intact delivers preparedContent and packageInfo through queue → drain → event → listener without a network fetch
   [PASSED] cold-sidebar relay: queued prepared launch reaches the model intact routes a queued journey URL through openLearningJourney with the payload intact
   [PASSED] cold-sidebar relay: queued prepared launch reaches the model intact a queued link without a key still opens — via the normal fetch path

• components/Home/HomePanel.test.tsx
   [PASSED] HomePanelRenderer composition renders MyLearningTab
   [PASSED] HomePanelRenderer composition renders home-page container
   [PASSED] HomePanelRenderer reading-only content opens full screen carries the prepared content into a pending guide and enters full screen transiently
   [PASSED] HomePanelRenderer content that drives the Grafana UI opens beside Grafana dispatches pathfinder-auto-open-docs with a redeemable launch key (payload stays off the event) when the sidebar is mounted
   [PASSED] HomePanelRenderer content that drives the Grafana UI opens beside Grafana opens the sidebar and queues a redeemable launch key when the sidebar is NOT mounted
   [PASSED] HomePanelRenderer content that drives the Grafana UI opens beside Grafana falls back to a floating overlay when another plugin owns the sidebar
   [PASSED] HomePanelRenderer content that drives the Grafana UI opens beside Grafana delivers to the floating panel when it is the current surface, instead of tearing it down
   [PASSED] HomePanelRenderer content that drives the Grafana UI opens beside Grafana signals an already-mounted floating panel to consume, once per launch, after staging
   [PASSED] HomePanelRenderer content that drives the Grafana UI opens beside Grafana carries packageInfo through the staged payload on both the event and the queued-link paths
   [PASSED] HomePanelRenderer launches are independent — no cross-launch surface leakage reading-only then interactive: the interactive launch resets the surface to sidebar (A→B)
   [PASSED] HomePanelRenderer launches are independent — no cross-launch surface leakage interactive then reading-only: the reading-only launch still enters full screen (B→A)

• components/docs-panel/docs-panel.auto-open-event.test.tsx
   [PASSED] Phase 0 tripwire: pathfinder-auto-open-docs CustomEvent contract the event name constant still carries the wire-contract string
   [PASSED] Phase 0 tripwire: pathfinder-auto-open-docs CustomEvent contract the shared routing predicate still tests journey pathnames
   [PASSED] Phase 0 tripwire: pathfinder-auto-open-docs CustomEvent contract listener registration exists in exactly one tracked file
   [PASSED] Phase 0 tripwire: pathfinder-auto-open-docs CustomEvent contract listener cleanup pairs with registration in the same tracked file
   [PASSED] Phase 0 tripwire: pathfinder-auto-open-docs CustomEvent contract handler routes journey URLs through openLearningJourney and others through openDocsPage
   [PASSED] Phase 0 tripwire: pathfinder-auto-open-docs CustomEvent contract untrusted event.detail.source is coerced via coerceLaunchSource at the boundary

TOTAL: 24 passed / 24 tests across 4 suites
```
</details>

## Pipeline

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.
- <code>`npm ci` — installed missing dependencies (node_modules absent in worktree; setup fix)</code>
- <code>`npx jest --coverage=false src/components/docs-panel/hooks/useAutoOpenListener.surface-ownership.test.ts src/components/docs-panel/hooks/useAutoOpenListener.test.ts src/components/docs-panel/hooks/useAutoOpenListener.cold-sidebar-relay.test.ts src/components/docs-panel/docs-panel.auto-open-event.test.tsx src/components/Home/HomePanel.test.tsx` — 34 passed</code>
- <code>Verified `panelModeManager.getMode()` defaults to &#39;sidebar&#39;, confirming existing default-surface call sites stay green under the new mode-gate</code>
- <code>Generated per-scenario pass transcript via jest `--json` reporter into the evidence dir</code>

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

---
Closes #1450
Follow-up: #1460


src/global-state/sidebar.ts
```diff
@@ -33,6 +33,11 @@ interface PendingOpenInfo {
  * Manages sidebar mounting and unmounting.
  */
 class GlobalSidebarState {
+  // Despite the name, this means "some Pathfinder surface is mounted" — set by
+  // the sidebar, the floating panel, AND the full-screen page. As of #1450 every
+  // one of those surfaces registers `useAutoOpenListener`, so a `true` value now
+  // also implies "the auto-open listener is ready," which is what the dispatch
```
leslie-heinzen, line 39: [blocking] This new invariant is not maintained by `ContextSidebar`: its cleanup in `src/module.tsx` still unconditionally calls `setIsSidebarMounted(false)`. Because the incoming floating/full-screen root can set this shared flag to `true` before the outgoing sidebar unmounts, that later cleanup can clobber readiness. Subsequent intercepted links are queued, `openSidebar()` no-ops in floating/full-screen mode, and the active listener's one-shot queue drain has already passed. Please make cleanup owner/mode-aware (or track explicit/ref-counted ownership) and add a transition-order test.

src/components/docs-panel/hooks/useAutoOpenListener.surface-ownership.test.ts
```diff
@@ -0,0 +1,98 @@
...
+  afterEach(() => {
+    jest.useRealTimers();
+    processQueuedLinksSpy.mockRestore();
+    localStorage.clear();
+  });
+
+  it('the floating listener opens an intercepted link when floating owns the surface', () => {
+    setMode('floating');
+    const model = makeModel();
+    renderHook(() => useAutoOpenListener(model, 'floating'));
```
leslie-heinzen, line 56: [blocking] This proves the hook behaves correctly when a caller supplies `'floating'`, but it does not verify that `FloatingPanelManager` or `FullScreenPanel` actually mounts the hook. The existing source-contract test tracks only `docs-panel.tsx` and the hook, so deleting either production call would leave these focused tests green and recreate the regression. Please add a component-boundary test or source-contract tripwire for all three exact owners, ideally with a real queued-link relay case for a non-sidebar surface.

leslie-heinzen (changes requested):
Requesting changes for two blocking issues found in the routed review and sustained by adversarial verification.

Non-blocking follow-ups:
- Centralize the repeated `pathfinder-panel-mounted` event contract in `src/lib/event-names.ts`.
- Consider extracting full-screen lifecycle/event wiring into a focused hook.
- Update `.cursor/rules/systemPatterns.mdc` so its load-bearing edge documents any-surface mounted/listener readiness across sidebar, floating, and full-screen surfaces.

Security review was clean, rollback is fully reversible, and focused Jest, TypeScript, and changed-file ESLint validation passed.

moxious:
Thanks @leslie-heinzen — both blocking findings were grounded and are now fixed in eee227e8.

**1. Unconditional `isSidebarMounted` clear in `ContextSidebar` (module.tsx).** Confirmed the asymmetry: `FloatingPanelManager` and `FullScreenPanel` already mode-gate their `setIsSidebarMounted(false)` cleanup, but the sidebar cleared it unconditionally. Since `setMode('floating'|'fullscreen')` commits the new mode and publishes `close-extension-sidebar` *before* the surfaces swap, the incoming root can set the flag `true` before the outgoing sidebar's cleanup clobbers it back to `false` — after which link-interception queues into a no-op `openSidebar()` and the one-shot drain has already passed (exactly the #1450 regression). Fixed by gating the clear on `panelModeManager.getMode() === 'sidebar'`, mirroring the other two surfaces.

**2. No test pins the three production mount sites.** Confirmed — the source-contract test tracked only `docs-panel.tsx` + the hook, and the surface-ownership test drives the hook directly, so deleting a production call left everything green. Added:
- a source-contract tripwire asserting all three surfaces mount `useAutoOpenListener` with their own surface gate (`'sidebar'` / `'floating'` / `'fullscreen'`);
- a transition-order tripwire pinning module.tsx's mode-gated clear (ContextSidebar isn't runtime-mountable in Jest — needs a scenes/ui theme provider — so this follows the file's established source-assertion style);
- a real queued-link relay test for the floating surface that drives the actual `linkInterceptionState.processQueuedLinks()` drain rather than a hand-dispatched event.

I verified both tripwires *bite*: reverting the module.tsx guard to unconditional, and removing the floating mount call, each fails its corresponding test.

The three non-blocking follow-ups (centralize `pathfinder-panel-mounted` in `event-names.ts`, extract full-screen lifecycle wiring, and the `systemPatterns.mdc` load-bearing-edge doc) I've left out of this PR to keep it minimal and disjoint from the in-flight panel-mode work — happy to file them as follow-ups if you'd prefer.

leslie-heinzen (approved):

PR #591, opened 2026-02-19
feat(coda): v1 plugin of the coda integration

Jayclifford345:
This pull request introduces the initial Go backend implementation for the Grafana Pathfinder app, integrating Coda VM provisioning and management capabilities. It adds the foundational Go code, configures the build system and dependencies, and updates the development workflow to support building and testing the Go backend alongside the frontend.

**Backend implementation and VM provisioning:**

* Introduced a new Go backend application with entrypoint `pkg/main.go` that starts the plugin and integrates with Grafana via the plugin SDK.
* Added the core plugin logic in `pkg/plugin/app.go`, including settings parsing, health checks, and management of VM terminal sessions.
* Implemented a Coda client in `pkg/plugin/coda.go` for registering with the Coda API, creating, listing, retrieving, deleting, and polling VMs, as well as managing authentication with JWT tokens.

**Build and development tooling:**

* Added `Magefile.go` for Mage-based Go build automation, and updated `package.json` scripts to support building, testing, and linting the Go backend for multiple platforms. [[1]](diffhunk://#diff-6680eec3986e59a1c51dad609391d10e23e28d43dd7099a4ecb96c43e29182a8R1-R12) [[2]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R12-R31)
* Introduced a new `go.mod` file to define module dependencies, including the Grafana plugin SDK, Mage, Gorilla WebSocket, and several indirect dependencies required for plugin development.

**Development environment enhancements:**

* Updated `docker-compose.yaml` to enable the Grafana Live feature and allow all origins, supporting terminal streaming and real-time features for the app.

---

> [!NOTE]
> **Medium Risk**
> Introduces a new Go backend build/dependency toolchain and tightens local dev auth defaults; failures are most likely in CI/build compatibility and local environment expectations rather than runtime behavior.
> 
> **Overview**
> Adds initial Go backend scaffolding for the plugin: introduces `go.mod`/`go.sum` (Go 1.25.7) and a `Magefile.go` to build via `grafana-plugin-sdk-go`’s Mage targets.
> 
> Updates workflow/docs to treat the repo as *frontend + Go backend*: CI now conditionally sets up Go and runs Mage `coverage`/`buildAll`, Cursor review/tidy rules add Go-specific checks, and `AGENTS.md`/`techContext.mdc` are updated with backend commands and structure.
> 
[...]

Issue #76, opened 2025-07-24
Separate cloud & OSS recommendations


moxious:
We now send whether its OSS or cloud to the recommender but we still send both OSS and Cloud results to any instance this is fine for now as we are demoing but in the end we will need to separate these

This can be done by editing the static recommendations file and simply adding a flag, something akin to "OSS only", "cloud only", or "both", and have the recommender filter based on which environment is asking the question.

PR #1468, opened 2026-07-30
feat(coda): consume the grafana-coda-app plugin instead of embedding its backend

Jayclifford345:
## What

Deletes Pathfinder's Coda backend (~5,300 lines of Go) and rewires the terminal onto the versioned `v1` API of the new [`grafana/grafana-coda-app`](https://github.com/grafana/grafana-coda-app) plugin, consumed through its published SDK, [`@grafana/coda-client`](https://github.com/grafana/grafana-coda-app/tree/main/packages/coda-client).

This is the Pathfinder half of an extraction whose other half has already landed: the plugin repo now owns the Coda REST client, JWT refresh, the WebSocket-relay SSH transport, the PTY session, the session registry, VM resolution, quota and rate limiting.

## Why

Pathfinder owning that stack meant it was the only plugin that could offer a sandbox, and put its release cadence in front of every sandbox change. The capability is now available to any Grafana plugin over a documented contract.

**Pathfinder keeps the terminal UI.** The xterm panel, scrollback persistence, search, resize handle, step-registry participation and challenge-block lifecycle are tightly coupled to guide execution and were not worth generalising away. It gives up the backend only.

## Behaviour

Coda stays **optional** — availability is probed at runtime with `isAppPluginEnabled('grafana-coda-app')` rather than declared as a plugin dependency. An absent plugin means a hidden terminal, not an error, and guides that need no sandbox are unaffected.

`CodaBackendStatus` on the config page names whichever of the three gates is unmet and links to the plugin's own page.

## ⚠️ Operator action required

**The Coda refresh token cannot be migrated.** Grafana plugin settings are per-plugin and `secureJsonData` is write-only — `secureJsonFields` exposes booleans, not values — so nothing in the frontend can copy it across. Every operator with Coda enabled must re-enter their enrollment key at `/plugins/grafana-coda-app`.

`docs/sources/upgrade-notes` covers this. The Pathfinder-side settings that moved (`codaApiUrl`, `codaRelayUrl`, `codaRegistered`) are dropped from `DocsPluginConfig`; only `enableCodaTerminal` remains.

## 📦 Lockfile handling (this is not a merge blocker)

An earlier revision of this description claimed CI could not pass until `2026-08-15T15:26:59Z`. **That was wrong** — CI is green on this head, and `npm ci` runs fine today.

Our `.npmrc` sets `min-release-age=3`, which is a **resolution-time** filter. `npm ci` installs the tree already pinned in `package-lock.json` and never re-resolves, so the age of `@grafana-coda/sandbox-client@1.1.1` (published `2026-08-12T15:26:59Z`) does not affect it.

What *is* true, and worth knowing when resolving conflicts: **regenerating** the lockfile fails until the window passes, with an opaque error that does not mention the policy:

```
npm error code ENOVERSIONS
npm error No versions available for @grafana-coda/sandbox-client
```

So resolve any `package-lock.json` conflict by keeping this branch's lockfile and running `npm ci` — never `npm install`. Do not weaken `.npmrc` to "unblock" anything; there is nothing blocked.

## Consuming the SDK rather than a hand copy

The client at `src/integrations/coda/coda-api.ts` is now a thin adapter over the published package (folded in from #1607, which is closed):

```json
"@grafana/coda-client": "npm:@grafana-coda/sandbox-client@^1.1.1"
```

The alias is interim, until the `@grafana` scope is claimed. Every consumer keeps its existing import path and call signature (`getCapabilities`, `createSession`, `execInSession`, `deleteSession`, `isRoleForbidden`, `sampleAppsUrl`, `alloyScenariosUrl`, `PATHFINDER_READY_FILE`), so the adapter and `useTerminalLive.hook.ts` are the only files that really change shape.

Two behaviour differences worth a reviewer's eye, called out rather than papered over:

- **`isUnavailable` now also treats `coda_auth_failed` as unavailable.** That is the SDK's classification to own — this PR consumes it rather than patching it.
- **Session cleanup now proactively frees the slot.** `CodaSession.close()` issues an idempotent `DELETE /sessions/{id}` from every cleanup path; the hand copy just let the backend's own sweep collect abandoned sessions. This reads as a strict improvement — flagging in case it isn't.

Two cosmetic signals from the raw Live channel are gone, since `CodaSession` does not surface them: the transient "Waiting for SSH handshake..." line, and instant client-side detection of the underlying WebSocket dropping. Both are still covered, with a delay, by the connecting banner and the idle timer's `terminal_disconnected`.

## Follow-up, not a blocker

`isCodaPluginAvailable()` in `useCodaAvailability.hook.ts` duplicates the SDK's own `CodaClient.isAvailable()`, and is deliberately kept for now. The SDK's version calls `isAppPluginEnabled(...)` unguarded, but that function is absent from core's `@grafana/runtime` below ~13.1, where it throws **synchronously** — `.catch()` cannot see it, and it previously crashed the whole docs panel. Our `grafanaDependency` floor is `>=12.3.0-0` and three of the five images in the e2e matrix (12.3.10, 12.4.8, 13.0.6) sit in that range, so the local `typeof` guard stays load-bearing; `useCodaAvailability.hook.test.ts` pins it.

The guard is being fixed upstream in `grafana-coda-app`. Once a release carries it, `isCodaPluginAvailable` collapses to a one-line delegate and the local cache goes with it — a separate PR, tracked separately.

The rest of that hook is not SDK-shaped and stays regardless: the React hooks have no counterpart in a framework-agnostic package, `useCodaTerminalGate` folds in Pathfinder's own `enableCodaTerminal` setting, and the `checking` tri-state exists so the config page does not flash a failure on mount. `isCodaUsable` and `codaSessionEligibility` are already consumed from the SDK.

## Notes for review

Three details that used to be load-bearing in Pathfinder's hand-rolled client are now the SDK's responsibility, and each cost a debugging session before it was understood:

- publishes must pass `{ useSocket: true }`, or they silently no-op on multi-node Grafana while working fine on a laptop;
- the `channel` string from the session response is opaque and must be split, never constructed — an Alloy scenario id can contain `/`;
- every Live frame carries a single JSON-encoded string field.

`useTerminalLive.hook.ts` now keeps only what is Pathfinder-specific — progress bar, banner text and xterm wiring — hung off `CodaSession`'s handlers.

The dev compose mount takes the plugin's `dist/` from a sibling checkout, with `CODA_PLUGIN_DIST` to override rather than editing the compose file. `docs/developer/CODA.md` documents the two-plugin setup; `provisioning/plugins/app.yaml` provisions both.

Rebased onto `main`; the one conflict was in `docs-panel.tsx`, where `main` had extracted the editor check into `isCurrentUserEditor()` — kept that helper (logic is identical) alongside the new `useCodaPluginAvailable()` hook.

`jest.config.js` gained `@grafana/coda-client` in the ESM `transformIgnorePatterns` list (the package ships ESM-only, same treatment already given to `@grafana/schema`, `rxjs`, etc.), via the documented extension point in `.config/README.md#esm-errors-with-jest` rather than by editing `.config/`.

## Verification

- `npm run typecheck` — clean
- `npm run lint` — clean
- `npm run check` — clean end to end (typecheck, lint, prettier, lint:go, test:go, coverage)
- `npx jest` — 429 suites, 6793 passed (18 skipped, 5 todo, pre-existing)
- `go build ./...`, `go vet ./...`, `go test ./pkg/...` — clean; `pkg/` is now an App Platform read proxy only
- The architecture test gained one allowlist entry for `requirements-manager/checks/coda.ts -> integrations`, matching the existing dynamic-import treatment of `terminal.ts`

**Not covered by CI, stated plainly because an earlier revision of this section got it wrong.** CI does *not* exercise Pathfinder against a real `grafana-coda-app`. The `docker-compose.coda.yaml` overlay is opt-in and its own header says CI must not mount it, so no check crosses the plugin boundary: the consumer tests here are mocked, and the e2e matrix runs the base stack without the provider. What CI does cover is that this branch builds, typechecks, lints, and passes 444 suites with the locked SDK in the tree.

A real consumer/provider contract job needs a pinned, published `grafana-coda-app` build for CI to mount, which does not exist yet. That is a follow-up against both repos, in that order — land the boundary, get versioned provider artifacts published, then pin one in CI. Also not verified locally: the live end-to-end path against a real Coda dev stack.


Jayclifford345:
## Cross-repo triage — 8 items

**Coda plugin issue.** `CodaBackendStatus` is a correct reading of a contract that cannot answer the
question it is asked. `/v1/capabilities` sets `registered` from "a Coda client could be built" — which
the plugin's own 1.1.0 release established does *not* mean the credential works; an expired refresh
token reports `registered: true` while every call `401`s. The `credential` object that fixed it went
onto `/v1/health` only, while `docs/API.md` still says "start here" about capabilities. So
`capabilities.registered ? 'ready' : 'not-registered'` renders a green "Coda is ready" on a stack where
every learner will hit an auth failure, and there is no field you could branch on instead.

Tracked in `grafana/grafana-coda-app`, owned by @Jayclifford345 and the Coda plugin maintainers:
https://github.com/grafana/grafana-coda-app/issues/51

Once `/v1/capabilities` carries `credential`, `CodaBackendStatus` wants a fourth state for it.

Triaged as part of the Coda extraction (grafana/grafana-pathfinder-app#1468).

---

**Coda plugin issue**, and it is the reason this PR carries a hand-rolled client.

Confirmed `@grafana/coda-client` is genuinely unpublished — `registry.npmjs.org/@grafana/coda-client`
returns 404, while `packages/coda-client/package.json` declares `1.1.0`. Per that repo's
`packages/coda-client/RELEASING.md`, publishing is blocked on **one human action**: an `@grafana` npm
organisation owner configuring a GitHub Actions trusted publisher (no token to provision; it is OIDC).
So `src/integrations/coda/coda-api.ts` is a forced workaround, not a boundary violation — and your file
header's account of why the git-ref fallback does not work matches that repo's own findings exactly.
Nothing to change here now.

Tracked in `grafana/grafana-coda-app`, owned by @Jayclifford345 and the Coda plugin maintainers:
https://github.com/grafana/grafana-coda-app/issues/52

Once it is published, the swap your header describes unblocks — and it removes the stale-`sessionId`
class too, since the package's `CodaSession` cannot be used after it closes.

Triaged as part of the Coda extraction (grafana/grafana-pathfinder-app#1468).

---

**Coda plugin issue** — documentation drift found in `grafana/grafana-coda-app` while triaging this PR, not a defect in the PR itself.

`docs/API.md` in the plugin repo still closes with "A typed TypeScript client wrapping these calls and the Live protocol **is planned**", although `packages/coda-client/` exists at version `1.1.0` and `docs/ROADMAP.md` §2 marks it DONE. A consumer reading the contract document is therefore told to reimplement the Live handshake by hand — which is a contributing signal towards this PR's hand-rolled `src/integrations/coda/coda-api.ts`, even though the dominant cause is that the package is not yet on npm (`npm view @grafana/coda-client` → 404, blocked on one npm trusted-publisher configuration per `packages/coda-client/RELEASING.md` §1).

No action required in Pathfinder for this item. Tracked in `grafana/grafana-coda-app`, owned by @Jayclifford345 and the Coda plugin maintainers: https://github.com/grafana/grafana-coda-app/issues/53

Triaged as part of the Coda extraction (grafana/grafana-pathfinder-app#1468).

---

**Coda plugin issue**, found while triaging this PR rather than in it.

Your #1514 (`fix(settings): preserve provisioned jsonData fields on plugin settings save`) landed on
`main` after this PR was opened, and the identical defect exists in the extracted plugin's own
configuration page: it sends `jsonData: { apiUrl, relayUrl }` wholesale, which deletes
`jsonData.minimumSessionRole` on every save. That field is the only way to let Viewer learners start a
sandbox — RBAC cannot grant past the backend's basic-role check — so an operator editing a URL
silently re-locks sandboxes to Editor and every Viewer gets `403 role_forbidden`.

Tracked in `grafana/grafana-coda-app`, owned by @Jayclifford345 and the Coda plugin maintainers:
https://github.com/grafana/grafana-coda-app/issues/54

Nothing to do on the Pathfinder side; noting it here because #1514 is the precedent and the fix is the
same one-line spread.

Triaged as part of the Coda extraction (grafana/grafana-pathfinder-app#1468).

---

**Split issue** — the migration path in `docs/sources/upgrade-notes/_index.md` does not match what the
Coda plugin does. Three problems:

1. **"select Register with Coda" is harmful.** Registration is automatic on save; the button that exists
   is `Register now` and posts `{"force": true}`, which the plugin's `docs/API.md` calls destructive —
   Coda revokes every token for an instance when one is redeemed. Save-then-press force-redeems a
   one-time key the automatic path already spent, producing a "your enrollment key was rejected" banner
   moments after registration in fact succeeded.
2. **Stale names** — the plugin's display name is "Coda", not "Coda-App".
3. **`minimumSessionRole` is missing.** The deleted backend had no role check; the extracted plugin needs
   Editor or above to spend VM quota and RBAC cannot grant past it, so any stack whose learners are
   Viewers finishes this migration with every sandbox returning `403 role_forbidden`.

Coda half (guard the force-register, document the operator cutover) is tracked in
`grafana/grafana-coda-app`, owned by @Jayclifford345 and the Coda plugin maintainers:
https://github.com/grafana/grafana-coda-app/issues/55 — that half goes first, since the button's semantics decide the wording. The
upgrade-note rewrite stays with the Pathfinder maintainers.

Triaged as part of the Coda extraction (grafana/grafana-pathfinder-app#1468).

---

**Pathfinder issue** — two Bugbot threads here are marked resolved but the branch shows no fix, and
resolved-without-a-fix feedback is what disappears at merge.

1. **Stale `sessionId`** (`useTerminalLive.hook.ts`). `setSessionId(null)` appears only at line 554 (top
   of `connect`) and 597 (`disconnect`). The `'disconnected'`/`'error'` branches,
   `LiveChannelConnectionState.Disconnected` and the subscription-error path all leave it set, so
   `getTerminalSessionId()` keeps handing a dead id to `coda-exit-zero` and `challenge-block.tsx:238-248`.
   Exec is session-scoped now — the deleted backend found any live client for the user — so tracking
   session death became the consumer's job.
2. **Challenge hang when the plugin is absent** (the other two threads). Same mechanism as #1541; the
   extraction just adds a second trigger. Please fold it into #1541 rather than tracking it twice.

Stays with the Pathfinder maintainers; mirrored in `grafana/grafana-coda-app` for visibility only:
https://github.com/grafana/grafana-coda-app/issues/56. The durable fix for (1) is the `@grafana/coda-client` swap — its `CodaSession` closes
and refuses `exec` locally — which is blocked on publishing the package.

Triaged as part of the Coda extraction (grafana/grafana-pathfinder-app#1468).

---

**Split issue** — one factual correction, plus follow-up in both repos.

> Not verified locally: the e2e suite, which needs both plugins built and mounted. CI covers it.

CI does not. This PR's own `docker-compose.coda.yaml` says the opposite — "the base stack deliberately
does not include it — **and CI must not**, or the bind-mount below would point at a path that does not
exist on the runner" — and the two follow-up commits are that decision being made. The overlay needs
`COMPOSE_FILE` and `CODA_PLUGIN_DIST` set by hand. No Playwright spec on this branch touches a terminal,
challenge VM or Coda, and `grafana-coda-app` removed its own suite (`run-playwright: false` in its CD).
So after this merges the interactive terminal has no automated coverage on either side of the boundary.

Coda half — stale two-plugin setup instructions in that repo's `AGENTS.md` and `.claude/PATHFINDER.md`,
still describing the old root-compose mount, plus the fake Coda service its `ROADMAP` §5 designs as the
prerequisite for any real e2e — is tracked in `grafana/grafana-coda-app`, owned by @Jayclifford345 and
the Coda plugin maintainers: https://github.com/grafana/grafana-coda-app/issues/57. Correcting the claim above stays with the Pathfinder
maintainers; everything else in the verification list checks out.

Triaged as part of the Coda extraction (grafana/grafana-pathfinder-app#1468).

---

**Coda plugin issue** — the extraction adds a second unverified contract, and its version-negotiation
half is design rather than working code. `/v1/capabilities` advertises `apiVersions` and
`pluginVersion`; both this PR's `CodaCapabilities` and the plugin's `@grafana/coda-client` declare them
and **neither reads them**. What is implemented is the `V1_DEFAULTS` fallback for the three `timings`
numbers, which works — but only covers tunables, not capabilities. `readyFile` is the case that matters:
a backend predating it ignores the unknown field and runs the command **ungated** with a `200`, so
`codaExitZeroCheck` would pass against a half-built environment rather than erroring. Every released
version has `readyFile`, so this is not a live incident — but nothing marks which `v1` additions change
behaviour when ignored, nothing states a minimum backend version, and no test drives a client against an
older backend.

Distinct from the golden-fixture gap (#1526's excluded Coda routes, carried separately): that is
types-within-one-build, this is two independently released artefacts meeting at runtime.

Tracked in `grafana/grafana-coda-app`, owned by @Jayclifford345 and the Coda plugin maintainers:
https://github.com/grafana/grafana-coda-app/issues/58

Triaged as part of the Coda extraction (grafana/grafana-pathfinder-app#1468).

Jayclifford345:
## Brought current with `main` (57 commits) and fixed four issues raised against this PR

Two merge commits (not a rebase — Pathfinder squash-merges, so the branch's internal history is
discarded at merge anyway, and a merge commit lets you audit each conflict resolution instead of
trusting a silent re-resolve). No force-push. `MERGEABLE` again.

### How each conflict was resolved

The branch predated **#1408** (custom guide packages, App Platform + on-behalf-of proxy auth) and its
follow-ups **#1569**, **#1570**, **#1594**, **#1596**, so it carried pre-#1408 copies of six docs. The
specific hazard was that a careless resolution would have reinstated the `Authorization: Bearer
<id-token>` + `X-Grafana-Id` outbound arrangement that `docs/design/BACKEND_PROXY_PATTERN.md` §3 now
explicitly forbids ("Do not reintroduce it"). Every conflicted hunk therefore takes `main` as the base
and re-applies only the Coda-specific edit.

| File | Resolution |
| --- | --- |
| `pkg/plugin/settings.go` | Deleted here, modified on `main`. **Kept the file**, with `OBOToken` only — the Coda fields have no readers left. `settings_test.go`'s fixture lost its now-meaningless `codaApiUrl`. |
| `pkg/plugin/app.go` | Kept `main`'s `ParseSettings` + `oboExchanger` wiring; dropped the `CodaClient`, stream-session map and per-user VM cache. |
| `go.mod` / `go.sum` | Took `main` (authlib, SDK v0.296.1), then `go mod tidy` dropped `gorilla/websocket` and `golang.org/x/crypto` along with the Coda code. |
| `docs/design/BACKEND_PROXY_PATTERN.md` | The canonical trust-boundary statement still moves here out of `CODA.md`, but now states `main`'s OBO model. Also dropped this branch's dev-stack parenthetical on the `Authorization` bullet — that is the falsified evidence `main`'s §3 repudiates. |
| `docs/developer/CODA.md` | Took this side wholesale. Everything `main` changed in the conflict region was Coda-backend prose being deleted; the identity trust boundary it also held is preserved, relocated, in the file above. |
| `.cursor/rules/systemPatterns.mdc` | Kept the Coda-free file map; took `main`'s wording for `app_platform_client.go` (two-axis classification, #1570) and `app_platform_identity.go` — **notably `forwardIdentityHeaders`, which #1408 deleted, is gone**. Added rows for `plugin/auth/` and `settings.go`, and corrected two now-false claims this branch had introduced (`grafana-plugin-sdk-go` as the only direct dependency; `app.go` holding only a logger). |
| `.cursor/rules/coda.mdc` | Dropped the Coda Go rows. #1569's `settings.go` correction moves to `systemPatterns.mdc`, which is where that file is described now. |
| `docs/developer/CONTEXT_INDEX.md` | Took `main`'s row for `BACKEND_PROXY_PATTERN.md` (#1580's contract goldens) and this branch's globs/descriptions for `coda.mdc` / `CODA.md`. |
| `AGENTS.md` | Kept the "read proxy, and nothing else" framing; restated identity as `main` does post-#1408. |
| `src/components/AppConfig/ConfigurationForm.tsx` | Dropped the Coda registration handlers; kept `main`'s #1514 comment about preserving provisioned `jsonData`. |
| `docs/design/CONCERNS.md` (2nd merge) | Took `main`'s table including #1578's `guide-identity.ts` row and review question; re-applied only the `coda-terminal` row. |
| `challenge-block.tsx` + test (2nd merge) | Took #1578's `useGuideRequirements()` hook and kept `useCodaTerminalGate()` alongside it. Dropped the test's stale single-key `requirements-manager` mock in favour of `main`'s factory. |

`pkg/` post-merge: `go build`, `go vet`, `mage lint`, `mage test` all clean, and nothing `main` added to
`pkg/` was lost (`plugin/auth`, `custom_guide_repository*`, `contract_fixtures_test.go`,
`settings_test.go` all present).

### The four fixes

**1. Stale `sessionId` after the stream ends** — grafana/grafana-coda-app#56, the Bugbot thread that was
resolved without a fix. Exec became session-scoped in the extraction, so "know when the session died" is
the consumer's job now, and this consumer only did it on explicit disconnect and at the start of a new
connect. `cleanup()` now clears the id — which makes the invariant structural and covers the backend
`error` / backend `disconnected` / handshake-timeout paths — and the three that do not call it (Live
channel drop, subscription `error`, subscription `complete`) clear it themselves. `getTerminalSessionId()`
additionally returns `null` unless the status is `connected`, so the requirements checker cannot read a
live-looking id out of a dead terminal. New `useTerminalLive.hook.test.ts`, 8 tests; reverting the
production change fails 6 of the 8 (the other two are the positive control and the pre-existing
`disconnect` path).

**2. Coda-mode challenge hangs when the integration is off** — #1541, widened by this PR. `TerminalProvider`
mounts unconditionally while `TerminalPanel` — the only caller of `_register`, and so the only source of a
real `connect` — is gated, so `openTerminal` silently no-ops. The load-bearing fix is `isTerminalRegistered`
on the context, checked on the Start click **and** in the status watcher, so an availability probe still in
flight at click time resolves into an error rather than a hang. `useCodaTerminalGate()` folds the two
operator-owned gates (`enableCodaTerminal`, then the plugin being installed) into one tri-state verdict —
tri-state precisely so "not asked yet" never renders as a failure — and the block renders that reason
instead of a Start button that cannot work. The context also carries the terminal's own `error` now, so a
challenge reports "Coda is not registered" or the role-floor message rather than a generic retry hint;
that covers the third gate the config page reports without a second `/capabilities` call per block.

`ChallengeBlockForm`'s Coda VM radio is **annotated, not disabled**. A guide is often authored on a stack
with no Coda for learners on a stack that has it, so disabling would block legitimate work; the author only
needs to know their own preview will not run. Say if you'd rather it were disabled. 6 new tests on the
block, 3 on the form, 6 on the context; reverting the production change fails 5 of the 6 block tests (the
sixth is the standard-mode negative control, which correctly passes either way).

**3. Coda 401 trips Grafana's global auth handling** — #1539. `retry: 1` is the right knob and `retry`
does exist on `BackendSrvRequest`. Confirmed against the plugin that `coda_auth_failed` really is a 401
(`pkg/plugin/resources.go:186`). Applied to the hook the issue named **and** to `coda-api.ts`'s shared
`request()`, which every session, exec and capabilities call goes through and has the identical defect —
that one is this PR's own code, so it seemed wrong to leave it. Not adding `showErrorAlert: false`, per
the issue.

**4. Upgrade notes prescribe a sequence that does not work** — `docs/sources/upgrade-notes/_index.md`.
Saving *is* the registration, so the "then select Register with Coda" step was wrong; the button that
exists posts `{force: true}` and re-submits a key that has just been spent. The note now says plainly
that this does **not** break a healthy stack — the plugin only overwrites its stored credential on
success, and Coda revokes only on a successful redemption — and that the harm is a false "key rejected"
diagnosis, so "do not press this" isn't read as "you are broken". Fixed the pre-1.0.0 UI labels (display
name is **Coda**; the button is **Register now**). Added `minimumSessionRole`: floor is Editor by default,
RBAC cannot grant past it, and Pathfinder's embedded backend had no role check at all — so a Viewer-learner
stack can finish the whole migration and then answer `403 role_forbidden` everywhere. Detail is linked to
the plugin's `docs/OPERATORS.md` rather than restated.

### Blocked on unmerged plugin PRs — deliberately not done here

- **`@grafana/coda-client`** would make issue #56's whole class structurally impossible (`session.ts:107,126,247`),
  but the package is unpublished (`npm view` → 404) and the git-ref fallback does not work, so there is
  currently no way to depend on it. `coda-api.ts`'s header documents the swap for when there is.
- **grafana/grafana-coda-app#79** adds a `credential` field to `/v1/capabilities`; **#80** adds `code` to
  Live frames. Both unmerged — consuming either now would make this PR unmergeable. The frame `code` in
  particular would let `useTerminalLive` classify backend errors the way the REST paths already do.
- **grafana/grafana-coda-app#76** (`docs/OPERATORS.md`) is still open. The upgrade note links to it at
  `main`; that link 404s until #76 lands, so please sequence #76 first.

### ⚠️ Collision with #1579

**#1579 (`fix/issue-1527-live-terminal-protocol`, open) creates `src/integrations/coda/useTerminalLive.hook.test.ts`
— the same path this PR now adds.** Whichever merges second gets an add/add conflict on that file. It also
rewrites `useTerminalLive.hook.ts` substantially (replacing `parseTerminalOutput` with `parseTerminalFrame`
from a new `terminal-protocol.schema.ts`), touching the same message-handling region as this PR's
`setSessionId(null)` clears — and it is based on `main`'s pre-extraction copy of that file, so the two
versions have diverged well beyond the overlap. I have not touched #1579's files beyond what the merge
forced. Worth agreeing an order: if #1579 goes first, this PR's session-lifetime commit needs replaying
onto its rewrite and the two test files merging; if this one goes first, #1579 rebases onto a file that
no longer has the code it is changing.

### Verification — actually run, on the merged tree

```
npm run check                → exit 0
  typecheck (tsc --noEmit)   → clean
  lint (eslint)              → 0 errors, 24 warnings (all pre-existing @typescript-eslint/no-deprecated)
  prettier-test              → all files formatted
  docs:sync-terms:check      → ok
  lint:go (golangci-lint)    → 0 issues
  test:go                    → ok (pkg/plugin, pkg/plugin/auth)
  test:coverage (jest)       → 427 suites, 6771 passed / 18 skipped / 5 todo — thresholds met
  test:review-contract       → pass
go build ./... && go vet ./...  → clean
npm run build                   → webpack ok
npm run build:cli && validate:packages && repository:check → ok, 0 invalid, repository.json fresh
```

Every new test was discrimination-checked by reverting the production change and confirming failure;
the counts are above. One caveat worth recording: a full-suite run before the second merge lost one
suite to a `SIGSEGV` in a jest worker (`src/cli/mcp/__tests__/failed-mutation-invariant.test.ts`); it
passed alone and on every subsequent full run, so it was an environmental worker crash, not this change.

**Not verified locally:** the e2e suite, which needs both plugins built and mounted, and anything
requiring a live Coda service — so the end-to-end behaviour of fixes 1 and 3 against a real 401 or a real
stream teardown is covered by unit tests only. CI covers e2e.

Jayclifford345:
### CI result, and one pre-existing failure I could not fix

Everything except e2e is green on `70e8e65b`, including Cursor Bugbot, `signed-commits`, `compatibilitycheck`, both CodeQL analyses, zizmor, semgrep and Socket:

`Lint` · `Typecheck` · `Prettier` · `Test` · `Test backend` · `Build` · `Validate packages` · `Validate bundled guides` · `Build CLI, build image, smoke test` · `Build and smoke-test E2E runner` — **pass**.

**Four of the six e2e matrices fail, and they were already failing on this branch before I touched it.**

| matrix | result |
| --- | --- |
| `grafana-enterprise@13.1.3` | pass |
| `grafana-dev@13.1.0-25957875840` | pass |
| `grafana-enterprise@12.3.10` | **fail** |
| `grafana-enterprise@12.4.8` | **fail** |
| `grafana-enterprise@13.0.6` | **fail** |
| `grafana@dev-preview-react19` | **fail** |

Evidence that this is not from the merge or the four fixes:

- The previous run on this branch — `45979285`, pushed 2026-08-03, [run 30822895780](https://github.com/grafana/grafana-pathfinder-app/actions/runs/30822895780) — failed **the same four matrices** (`12.3.9`, `12.4.6`, `13.0.4`, `react19-preview`) with **the same seven tests** and the same signature: `getByTestId('docs-panel-container')` → `element(s) not found`. `block-editor-title-row` ×2, `block-editor`, `online-package-recommendations`, `open-close` ×2, `welcome-journey`.
- `main` at `dddbd33c` — the exact commit merged in — passes all six ([run 31593469515](https://github.com/grafana/grafana-pathfinder-app/actions/runs/31593469515)).
- None of the seven failing specs touch Coda, the challenge block or the block-editor challenge form. They all fail on the same first assertion: the docs panel never opens.

Reduced to the branch's own e2e-relevant delta, it is two lines — `.config/Dockerfile` and everything under `tests/e2e-runner/` in this diff came from `main`, not from here:

1. `src/plugin.json` — `"streaming": true` removed (`2e2b7f88`).
2. `provisioning/plugins/app.yaml` — the `jsonData: { codaRegistered: false }` block removed, leaving an `apps:` entry with no `jsonData` key at all (`38a11d64`). This file **is** mounted in CI (`../provisioning:/etc/grafana/provisioning` in `.config/docker-compose-base.yaml`), so it is live. `codaRegistered` now has zero readers in `src/`, so the value is genuinely dead — but "no `jsonData` key" is a different provisioning input from "`jsonData` with one field", and that is my leading suspect for a version-dependent difference.

I have **not** attempted a fix. I cannot reproduce locally — the Docker daemon is unavailable in this environment — and guessing at a version-dependent provisioning failure I cannot verify seemed worse than leaving it clearly reported.

One thing would make this diagnosable in minutes, and it is a bug on `main`: the workflow's `Docker logs` step runs `docker logs grafanalabs-grafanadocsplugin-app`, but `.config/docker-compose-base.yaml` sets `container_name: 'grafana-pathfinder-app'`. The step exits 1 and uploads nothing, which is why both failing runs have **no Grafana server log** — the one artifact that would say whether the plugin loaded, whether provisioning errored, or whether a chunk 404'd. `main` never notices because its e2e never fails. That step only runs on failure, so fixing it is zero-risk; it just needs `.config/` to stay untouched, so the workflow side is the thing to change.

Suggested next step for whoever picks this up: fix the container name so the server log is captured, re-run, and read it before changing either of the two suspect lines.

Jayclifford345:
## One stale caveat in the upgrade note, introduced by a plugin-side fix

The rewritten `docs/sources/upgrade-notes/_index.md` advises setting the session role floor "before your last configuration-page save". That was correct when this branch was written and is no longer:

grafana-coda-app#73 (merged) makes the plugin's configuration page spread the stored `jsonData` and override only the two fields the form owns, so **Save no longer erases `minimumSessionRole`**. The plugin's own `docs/OPERATORS.md` was corrected for this in grafana-coda-app#83, which keeps the old behaviour only as an explicit caveat for versions ≤ 1.1.1.

Suggested change: drop the ordering advice, or scope it — "on Coda plugin versions 1.1.1 and earlier, set the role after your last configuration-page save". Everything else in that section (saving *is* the registration, no button afterwards, the `403 role_forbidden` symptom, linking `OPERATORS.md` as the authority) matches the current plugin behaviour and reads correctly.

For reference, the plugin side is now fully merged — 15 PRs, `main` green at 293 tests — so the routes, codes and `/v1/capabilities` fields this PR consumes are all live. Two additions worth knowing about for follow-ups rather than this PR: `/v1/capabilities` now carries `credential` (so `CodaBackendStatus` can stop reading `registered` alone, which reports `true` for a dead credential) and `caller.canCreateSessions` (so a sandbox feature can be gated before it spends a request). Both are additive and absent-tolerant, so nothing here breaks by not using them yet.

moxious (changes requested):
> **Amended.** An earlier version of this review was submitted prematurely, before its findings had been through adversarial verification. Two of them did not survive and have been corrected below: the cancel-race no longer claims a quota-charged VM, and the `readyFile` / `codaSupports` finding is withdrawn. Apologies for the churn if you read the first version.

Reviewed with the routed `/review` orchestrator: 14 concern reviewers, 2 contract-evolution scans, a tech-debt pass, a doc-drift pass, and 9 adversarial-verification skeptics over the medium-and-higher findings.

**Disposition: approve once the "fix in this PR" list is addressed.** The extraction itself is clean, and the security posture improves — the deleted `PublishStream` looked sessions up by channel path with **no caller-identity check**, so any authenticated user who knew `terminal/{vmId}` could inject keystrokes into someone else's SSH session. That, `ssh.InsecureIgnoreHostKey()`, and raw private-key handling all go with it. `pkg/` at head is genuinely tidy: `go build`, `go vet`, `go test ./pkg/...`, `npm run lint:go` (0 issues) and `go mod tidy -diff` all clean, with `gorilla/websocket`, `x/crypto` and `x/term` really dropped.

**Blast-radius calibration, which I applied to every runtime finding below:** the terminal needs an admin to set `devMode: true`, add that specific user's ID to `devModeUserIds`, enable `enableCodaTerminal` (`DEFAULT_ENABLE_CODA_TERMINAL = false`), _and_ install and enrol `grafana-coda-app`. Nothing here reaches a production user unopted, so none of these is a shipping regression.

## Fix in this PR

**1. No `coda-terminal` row in the contract anchors table** — `docs/design/CONCERNS.md`

This PR replaces the capability's lifecycle owner outright, and `.cursor/rules/coda.mdc` now says so under a new "Ownership boundary" heading. But the anchors table still lists only `floating-panel` and `block-editor-authoring`. `.cursor/skills/review/SKILL.md` §5 does not allow deferring the anchor row to a follow-up — without it, the next Coda PR's evolution scan reconstructs the contract from history that is now 5,300 deleted lines describing the _old_ owner. Suggested row: `grafana-coda-app` owns session lifecycle, VM provisioning, SSH, relay, quota and credentials; Pathfinder owns the terminal UI and is a v1 consumer through the single adapter `coda-api.ts`; the Live channel is opaque; availability is resolved at runtime, never declared as a dependency.

**2. `plugin.json` declares `>=12.3.0-0`, but the terminal needs Grafana ~13.1 — and the config page blames the wrong component**

`isAppPluginEnabled` is absent from `@grafana/runtime` 12.3.0, 12.4.0 and 13.0.0, and present in 13.1.0 (checked by unpacking all four published tarballs). `useCodaAvailability.hook.ts:29-33` resolves `false` there, which hides the panel, strips terminal blocks from the palette, and makes `CodaBackendStatus.tsx:72-74` render **"Coda app plugin not found"** — the `unreachable` branch can't catch it, because the guard resolves rather than rejects.

The load-bearing part: `grafana-coda-app`'s own manifest also declares `>=12.3.0-0`, and the SDK's data path uses only `getBackendSrv().fetch` and `getGrafanaLiveSrv`, both ancient. Only `CodaClient.isAvailable()` touches `isAppPluginEnabled`. So the sandbox would work on 12.3–13.0 but for this probe, and no doc states a floor — it exists only in a code comment and a test.

Please **don't** hand-roll a settings probe as the fix. I checked 13.1's implementation: `isAppPluginEnabled` delegates to `getPluginSettings`, which branches on an OpenFeature flag and is mid-migration to an App Platform route, so a hand-rolled version pins the legacy half; and `config.apps` is marked deprecated in `@grafana/data`. This is a manifest-and-docs decision: state the floor, and give `CodaBackendStatus` a distinct state so it stops asserting the plugin is missing.

**3. `forwardIdentityHeaders` does not exist** — `.cursor/skills/secure/SKILL.md:100`, `docs/design/CONCERNS.md:59`

The symbol appears only in those two lines at head, and never existed at base — this PR introduced it. It is half of security rule B2 (High) and one of `go-backend`'s trigger keywords, so an agent running `/secure` either silently drops the outbound-header check or invents a citation to satisfy the skill's quote-the-canonical-pattern constraint. The real code is `req.Header.Set(auth.AccessTokenHeader, …)` at `app_platform_client.go:124`.

**4. The `min-release-age` merge blocker in the PR description isn't real**

CI is green across all 30 check runs on head. `min-release-age` is a resolution-time filter inside arborist; `npm ci` installs the pinned tree and never re-resolves, and the Build job's node-modules cache missed on the changed lockfile hash, so `npm ci` genuinely ran ~20 h after `1.1.1` was published and passed. There is no 2026-08-15 block.

Worth correcting before merge because the tempting reaction is to "unblock" it by weakening `.npmrc`. What _is_ true and worth keeping: lockfile **regeneration** on this branch fails with an opaque `npm error code ENOVERSIONS` until the window passes, so conflicts should be resolved by keeping the branch's `package-lock.json` and running `npm ci`, never `npm install`.

**5. Two state bugs in `useTerminalLive.hook.ts`**

- `connect()` sets `'connecting'`, calls `cleanup()`, then `await createSession()` and installs the session **unconditionally** (`:413-427`). `TerminalPanel.tsx:409` renders Cancel for exactly that window. A cancel closes nothing — `sessionRef.current` is null — then the promise resolves, `subscribe()` runs, and provisioning starts; the user sees "Disconnected by user" and then watches status flip back to `connected`. Base's `connect()` was synchronous, so Cancel always had a subscription to tear down; this is a regression. Fix with a generation token checked after the await, closing the losing session.
  _Corrected from the earlier version of this review:_ this does **not** spend a VM against quota. The SDK's `close()` explicitly does not destroy the VM, and the backend reuses a matching one, so the real harm is the leaked subscription and the stale status, not quota.
- `disconnect()` sets `suppressClosedBannerRef = true` unconditionally (`:437`), but only `onClosed` consumes it — and with no open session `onClosed` never fires. The latch survives into the next session, where the first genuine close takes the early return at `:339-347`, skipping both `cleanup()` and `setStatus('disconnected')`. Status sticks at `connected`, so `getTerminalSessionId()` keeps handing a spent id to `coda-exit-zero`. One-line fix: only latch when there is a session to suppress.

**6. Two agent-guidance contradictions the doc sweep missed**

- `.cursor/rules/techContext.mdc:50-51` still lists `gorilla/websocket` and `golang.org/x/crypto` as backend dependencies, which _this PR_ removed from `go.mod`, while `systemPatterns.mdc:209` now asserts the opposite. Two rule files, opposite facts, no precedence rule to break the tie.
- The new "**every route** drains a paginated upstream LIST … and rides the caller's own identity end to end" in `AGENTS.md:75` and `systemPatterns.mdc:172` is false for `/package-recommendations` (anonymous CDN fetch, no namespace, global 6 h cache) and `/health`. This one has a security tail: it invites a future change to put per-user data into a shared cache. `resources.go`'s own comment gets it right.

## Worth deciding, not blocking

- **Supply-chain posture on the new dependency.** The tarball is clean — no lifecycle scripts, zero runtime deps, one `getBackendSrv().fetch` call site, integrity matching the lockfile. But the package has a single individual maintainer, both versions were published 2026-08-12 nine minutes apart, 1.1.0 is deprecated as broken, and `dist.attestations` is absent (no npm provenance). Separately, `renovate.json`'s `matchPackageNames: ["/^@grafana/"]` is **unanchored**, so it also matches `@grafana-coda/…` and waives the repo's own 14-day `minimumReleaseAge` for the package that owns the session protocol. Anchoring to `/^@grafana\//` is a one-character fix; consider pinning `1.1.1` exactly until provenance exists.
- **Orphaned Coda secrets.** The enrollment key and refresh token stay in Pathfinder's `secureJsonData` after upgrade with no reader, no UI to clear them, and no revocation step in the upgrade note. That retention is also what makes a revert work, so it's worth saying which it is. Whether redeeming a new key invalidates the old registration Coda-side is unanswerable from this repo and is the one thing a revert can't undo — one sentence from the Coda team in the upgrade note closes it. Also: `docs/sources/upgrade-notes/_index.md` says the token "cannot be migrated — by Interactive learning or by anyone else", which is broader than provable; the narrow true claim is that Pathfinder has no supported path to write another plugin's secure settings and `grafana-coda-app` has no import path for an already-issued token.
- **Test coverage**, established by mutation rather than inference: deleting the banner-suppression guards, the unmount `cleanup` return, or the progress-interval `clearInterval`, or inverting `useCodaTerminalGate`, all leave the suite green. To be clear about the baseline — the Coda test files here are **new**; base had zero tests for `useTerminalLive` or `TerminalContext`, so frontend coverage rises. These are residual gaps, not regressions. Best value per line: return the fake terminal from `connectedHook()` and add one `unmount()` assertion.
- **Catalogue fetch.** `sampleAppsUrl()` / `alloyScenariosUrl()` build two routes the SDK doesn't model, and `useCodaOptions` reads response keys `apps` / `scenarios` carried over from the deleted Go handler. On mismatch the dropdown silently empties, indistinguishable from "plugin absent" — and it omits `showErrorAlert: false` (which `CodaClient.request` sets on the same resource base), so plugin-absent authoring pops a global toast with no in-place message. Sourcing both from the already-cached `getCapabilities()` (`capabilities.sampleApps` / `alloyScenarios`) removes two unmodelled URLs and the guessed keys in one move, and picks up `status` so the editor can mark experimental entries.
- **Three `CodaErrorCode` → message tables** (`useTerminalLive`, `challenge-block`, `checks/coda.ts`) with four different `role_forbidden` wordings, only two of which name `minimumSessionRole` — the one actionable fact. Worth one shared function.
- **`vm_quota_exceeded` advises "close another terminal"**, which provably doesn't free a slot — `close()` doesn't destroy the VM. Base auto-recovered via `cleanupUserVMsForQuota`; worth confirming whether that carried over, and dropping the clause either way.
- **QC8.** The rewritten hook's docblocks still describe the deleted backend, including "Retries SSH with fresh VM on auth failures", on lines this PR rewrote.
- **`ParseSettings` unmarshals `jsonData` into a struct with no JSON-mapped fields**, and returns early on error before reading `accessToken` — so a malformed blob now silently disables the App Platform proxies for a field nothing reads.

## Withdrawn after verification

The **`readyFile` / `codaSupports` feature-detection finding is withdrawn.** Five reviewers raised it independently; skeptics refuted it 2-of-3. The SDK's own docblock sanctions "accept an ungated run" as a valid degradation; `docs/developer/CODA.md` already calls `readyFile` "a **UI-race guard, not a security boundary**"; the only production caller sits behind the challenge block's state machine, which reaches "Check my work" only after the sentinel write returns exit 0; and step completions go to resettable per-user storage, not the durable-record path (which has no production subscribers and no backend write route). The proposed client-side gate would also reintroduce the shell-quoting hazard the deleted Go `shellSingleQuote` existed to handle.

One residue all three skeptics agreed on, offered as a nit: `checks/coda.ts:5` says the check is "**Always** gated on `PATHFINDER_READY_FILE`" as an unconditional guarantee, which isn't true against a backend predating `exec.readyFile`. The code is defensible; the comment overstates it.

Also withdrawn: a claim that `terminal-connect-step.tsx` regressed — it's zero-diff and `isTerminalRegistered` didn't exist at base, so this PR changed nothing about its reachability. And the 35-second stale-liveness concern is downgraded to informational: `coda-exit-zero` branches on backend error codes rather than status, so that window yields a correct failure, and base's "immediate" detector permanently disposed input on a _reconnectable_ `Disconnected`, which was worse.

## Contract evolution

Both scans returned **`coherent_extension`** — no finding. `go-backend`: the deletion removes the backend's second role, leaving only the App Platform read proxy that `BACKEND_PROXY_PATTERN.md` (#1401) describes, and this PR updated that document and `AGENTS.md` to match. `coda-terminal` ran on a discretionary trigger (the deterministic gate found zero semantic PRs in 30 days); reconstruction from #591 → #667 → #672 → #688 showed a positional Live-channel string co-owned by TypeScript and Go that grew a segment per feature, and this PR consolidates it behind one owner. The only gate it flags as unmet is the missing anchor row (item 1).

## Non-blocking passes

**Tech debt** — net direction strongly reducing. Debt added is concentrated in the adapter seam: it's declared in a header comment but unenforced and already bypassed inside its own directory (`useTerminalLive` imports `CodaSession` from the package directly), five new barrel re-exports in `index.ts` have zero importers, and `challenge-block.tsx` is now 634 lines with 18 hooks. `docs/design/AGENT_HARDENING.md` D4 still scopes itself to the deleted `stream.go` + `terminal.go`.

**Doc drift** — the doc work is above average where it was done; `coda.mdc`, `CODA.md` and the `BACKEND_PROXY_PATTERN.md` §3 move are genuine improvements, and the `coda-terminal` routing row was correctly rewritten so no dead `pkg/` trigger paths remain. Gaps are uniformly in files that weren't opened: `secure/SKILL.md` lines 22/94/250/324 still cite deleted `isAllowedCodaURL` / `IsAllowedRelayURL` / `setAuthHeader` / `coda.go`; `prevent-doc-drift/SKILL.md` has an orphaned rule pointing at a bullet this same PR deleted plus a new bucket with no rule row; three Go comments route the identity trust boundary to `CODA.md`, which now explicitly disclaims it; `json-guide-format.md:1092` still documents the two-gate indefinite hang this PR fixes; `LOCAL_DEV.md` never mentions the new opt-in compose overlay.

**Instrumentation** — the new degradation ladder (plugin-missing → not-registered → credential-expired → `role_forbidden` → unavailable) has no typed facade event, so none of it is countable or alertable per `TELEMETRY.md`. A judgment call while the feature is dark; worth one event before it's un-darked.

**Routing follow-up** — at `min_signals: 2`, a PR changing the availability gate in `docs-panel.tsx` or `BlockPalette.tsx` scores 1 signal and won't activate `coda-terminal`. Dropping it to 1 fixes that, and the signal-counting rule already prevents path-only activation.

## Verified clean

Zero F1–F6 hits, and no new DOM sink — every value crossing from the other plugin renders as a JSX text child. The `docs-panel.tsx` rebase resolution is correct against both merge parents, with `isCurrentUserEditor()` byte-identical to `main` and the editor and Coda gates on disjoint subtrees; all four docs-panel bootstrap signals untouched. `terminal-storage.ts` is zero-diff, so stored values still parse and mean the same thing. The guide schema has no diff at all — no field, enum value or validation rule changed — so existing challenge and terminal content still validates and runs. All three removed `data-test-*` ids are fully unreferenced. `streaming: true` removal is correctly paired with the implementation: a stale Live subscribe gets `codes.Unimplemented`, never a panic. `jest.config.js` extends the transform list through the documented extension point as a strict superset. And the earlier Bugbot findings about the challenge block hanging with no panel, `sessionId` not being cleared, and timers surviving teardown all reproduce as **fixed** at head.

Jayclifford345:
## Review fixes pushed — `782706e9`

All six **"fix in this PR"** items, plus the four follow-ups from **"worth deciding"** and two more the recheck surfaced. `npm run check` passes end to end: 429 suites, 6793 tests, 0 lint errors, prettier clean, `test:go` ok.

### The six blocking items

**1. `coda-terminal` contract anchor** — added as `#688 → #1468`: `grafana-coda-app` owns session lifecycle, VM provisioning, SSH, relay, quota and credentials; Pathfinder owns the terminal UI and is a v1 consumer through the single adapter `coda-api.ts`; the Live channel address is opaque; availability is runtime-resolved, never a declared dependency.

**2. Grafana 13.1 floor** — `grafanaDependency` deliberately **stays** `>=12.3.0-0`. Raising it would block the whole plugin from installing on 12.3–13.0 for one optional, dev-mode-gated feature, and the review's own evidence is that the sandbox works there but for the probe. Instead:

- `isCodaProbeSupported()` separates "this Grafana cannot answer" from "the plugin is absent";
- `CodaBackendStatus` renders a distinct `grafana-too-old` notice instead of asserting the plugin is missing;
- the floor is documented in `coda.mdc` and `CODA.md`, with the reasoning for not raising the manifest.

No hand-rolled settings probe — confirmed zero uses of `getPluginSettings` or `config.apps` in `src/`.

**3. `forwardIdentityHeaders`** — zero references remain. Both the `secure` skill's rule B2 and the `go-backend` trigger keyword now name `auth.AccessTokenHeader`, which is what `app_platform_client.go:124` actually sets.

**4. PR description** — corrected. I verified both halves rather than taking it on trust: `npm ci` installs cleanly today, and a from-scratch regeneration in a throwaway copy reproduces the failure exactly:

```
npm error code ENOVERSIONS
npm error No versions available for @grafana-coda/sandbox-client
```

One correction worth recording for whoever hits this next: `min-release-age` **is** a real npm 12 config (`npm config get min-release-age` → `3`; `config ls -l` lists it alongside `min-release-age-exclude`). The `Unknown project config "min-release-age"` warning npm prints on every command is spurious, and is the most likely route to the original wrong conclusion. The policy is active — it just filters at resolution time, which `npm ci` never reaches.

**5. Two `useTerminalLive` state bugs** — fixed, and verified by mutation rather than assertion. Reverting both fixes makes exactly 3 of the new tests fail; restoring them passes all 11.

- Cancel race: a generation token bumped by every teardown, compared after the `await`; the losing session is closed rather than installed.
- Banner latch: only latched when there is a session whose `onClosed` will actually consume it.

**6. Guidance contradictions** — `techContext.mdc` no longer lists `gorilla/websocket` or `golang.org/x/crypto` (and now names `grafana/authlib`, the real second direct require). The "every route drains a paginated upstream LIST" claim is gone from both `AGENTS.md` and `systemPatterns.mdc`, replaced with the App Platform routes named explicitly and `/package-recommendations` called out as the exception — including the security tail, that its cache is shared across users so per-user data must never enter it.

### Follow-ups

- **`renovate.json`** — anchoring alone was not sufficient. The dependency is consumed through an npm alias (`@grafana/coda-client` → `@grafana-coda/sandbox-client`) and Renovate matches on both names, so the alias would still have inherited the `@grafana` release-age waiver. Anchored to `/^@grafana\//` **and** added an explicit rule restoring the 14-day `minimumReleaseAge` for both names, with a note to drop it once the package moves into the `@grafana` scope.
- **`secure/SKILL.md`** — the other four dead references (`isAllowedCodaURL`, `IsAllowedRelayURL`, `setAuthHeader`, `coda.go`) repointed at `isAllowedInteractiveLearningHost` and the App Platform client.
- **Catalogue fetch** — now reads `capabilities.sampleApps` / `alloyScenarios`; `sampleAppsUrl`/`alloyScenariosUrl` and the now-orphaned `CODA_RESOURCE_BASE` re-export are deleted. I checked the SDK before moving: `CodaClient.request` sets `retry: 1`, so **the #1539 fix is preserved**, and it adds the `showErrorAlert: false` the hand-rolled fetch lacked. Picks up `status`, so experimental entries are labelled, and an unreachable Coda now says so in the placeholder instead of silently emptying.
- **Shared error messages** — the `CodaErrorCode` table and the role refusal move to `coda-api.ts`. The recheck found a **fourth** role wording I had missed at `challenge-block.tsx:390`; there is now exactly one, and it names `minimumSessionRole`.

### Two more the recheck surfaced

- **Routing row** — `review/SKILL.md` §5 forbids deferring *"the anchor row or the concern's routing paths"*, and the row had two gaps: `min_signals: 2` meant a change to the availability gate in `docs-panel.tsx` or `BlockPalette.tsx` scored one semantic signal and did not activate the concern at all, and `ChallengeBlockForm.tsx` / `useCodaOptions.ts` were absent from `trigger_paths` despite now reading the v1 capabilities contract directly. Dropped to 1 and added both paths plus the new symbols. The "never activate on paths alone without semantic evidence" rule still prevents false positives.
- **`checks/coda.ts` header** — the one residue from the withdrawn `readyFile` finding. It claimed the check is *"Always gated on `PATHFINDER_READY_FILE`"* as an unconditional guarantee, which is untrue against a backend predating `exec.readyFile`. Now states the guard's real scope and why the degradation is accepted.

### Deliberately not done

- **`vm_quota_exceeded`** wording is fixed ("wait for one to expire"), and I confirmed from the SDK why: `CodaSession.close()` documents that the VM is **not** destroyed, and only `deleteVM(id, force)` frees a slot — which Pathfinder does not expose. The auto-recovery `cleanupUserVMsForQuota` did **not** carry over. Worth a separate decision on whether it should.
- The **orphaned Coda secrets** question for the upgrade note (needs an answer from the Coda team on whether redeeming a new enrollment key invalidates the old registration).
- A **typed telemetry event** for the degradation ladder.

moxious (changes requested):
Independent review run against this repo's own review orchestrator. Thanks for doing the hard half of this extraction — the verification came back clean on the parts that matter most. Every deleted runtime capability maps to a real implementation: Coda registration and JWT refresh, VM CRUD and reuse, session-bound exec with its bounds and rate limiting, the Live session registry with quota recovery and heartbeat/expiry, and the relay/SSH/PTY transport all have provider-side or SDK equivalents, and quota cleanup is still called before session creation and after a provider quota response. The locked `@grafana-coda/sandbox-client@1.1.1` tarball contains every method this diff calls. The trust boundary does not weaken and is arguably stronger: the companion derives caller and org from trusted plugin context and rechecks role and ownership on REST and on Live Subscribe/Publish/Run, where the deleted `PublishStream` path did not re-establish caller ownership. The aggregate tech-debt scan is net negative. Requesting changes on the cutover, not the extraction.

**[blocking] Bind challenge setup to the newly requested session.** `challenge-block.tsx:295-305,468-473` reads the rendered `sessionId`, asks `openTerminal` for a different VM, then immediately calls `runSetup` whenever the pre-call status was `connected` — while `TerminalContext.tsx:184-209` disconnects the old session and schedules the replacement 100 ms later, leaving the setup callback closed over the old id. Switching from a connected challenge A to a different challenge B can run B's setup on A's VM, return `session_not_found`, or stall before B is prepared. The base had a comparable timing hazard through the user-global exec endpoint, but this is materially worse: exec is now explicitly pinned to the captured old session while SDK cleanup proactively deletes it. Make opening awaitable so it resolves with the new session id, pass that id into setup, and add the connected-A-then-different-B regression test. Note the earlier inline thread on this is marked resolved, but there is no reply and the control flow is unchanged at this head.

```ts
const nextSessionId = await terminalCtx.openTerminal(vmOpts);
await runSetup(nextSessionId);
```

**[blocking] Preserve the terminal on supported Grafana versions.** `src/plugin.json:74` declares `>=12.3.0-0` and the SDK data path supports `>=11.1.0`, but `useCodaAvailability.hook.ts:23-40` returns `false` below 13.1 solely because `isAppPluginEnabled` is absent before core 13.1 — the rule's own comment says nothing else in the data path needs it. An operator on supported 12.3, 12.4 or 13.0 who has a working terminal today loses every terminal-backed guide on upgrade. That is graceful against a crash but not behaviour-preserving. A direct probe of the stable v1 capabilities resource distinguishes an enabled provider from an absent one without raising the plugin's compatibility floor:

```ts
if (isCodaProbeSupported()) {
  return isAppPluginEnabled(CODA_PLUGIN_ID).catch(() => false);
}
return getCapabilities().then(() => true).catch(() => false);
```

**[blocking] Give every sandbox block the same graceful-degradation contract.** `challenge` got the `isTerminalRegistered` guard; `terminal` and `terminal-connect` did not. `TerminalProvider` stays non-null when `TerminalPanel` is gated away, so `terminal-step.tsx:189` and `terminal-connect-step.tsx:138` render apparently-enabled controls wired to a no-op (`docs-panel.tsx:1480`). A learner clicks and nothing happens, with no stated reason — which is what the upgrade note promises will not occur. Apply the same typed gate, surface the reason, and cover missing-plugin and old-Grafana states in tests.

**[blocking] Add a real consumer/provider contract job, and correct the CI claim in the description.** CI never mounts `docker-compose.coda.yaml` — the overlay says it must not — so no check exercises Pathfinder against the locked SDK and an actual provider (`.github/workflows/ci.yml:465`). The 34 deleted Go tests are replaced only by mocked consumer tests. Pin a companion version in CI and test availability, unregistered degradation, session create/subscribe/input/resize/close, exec, error paths and cancellation across the boundary. The body's statement that CI covers the two-plugin e2e suite is not accurate as written.

**[blocking] Make the documented local stack actually work.** `docker-compose.coda.yaml:19` and `docker-compose.yaml:7` omit both `externalServiceAccounts` and `GF_AUTH_MANAGED_SERVICE_ACCOUNTS_ENABLED`, so following the new two-plugin instructions returns `409 no_service_account_token` at registration. Anyone trying to reproduce or verify this PR locally cannot get past setup.

**[blocking] State the credential lifecycle accurately.** `docs/sources/upgrade-notes/_index.md:40` says the old Coda secrets are removed. They are no longer read, but nothing clears them and the provider guide says they remain stored, so they may still be live (`pkg/plugin/settings.go:20`). Correct the note, define the rollback window, confirm revocation semantics, and provide a post-window clear or revoke step rather than describing removal that does not happen.

**[blocking] Do not grow the tier-violation ratchet silently.** `src/validation/architecture.test.ts:103-113` states that additions to this allowlist are architecture degradation, then adds `requirements-manager/checks/coda.ts -> integrations`. Prefer a lower-tier port or registration seam. If the exception stands, document it in both canonical tier descriptions as the changed doc-drift skill requires (`AGENTS.md:59`).

**[blocking] Use one provider-aware authoring contract.** The block palette hides Coda types when unavailable (`BlockPalette.tsx:276`) while the type-switch dropdown will still convert a block into one (`forms/TypeSwitchDropdown.tsx:41`), so an author can create a block the palette says cannot exist. Separately, `ChallengeBlockForm.tsx:34` and `TerminalConnectBlockForm.tsx:18` hardcode the three legacy VM templates although `capabilities.templates` is now authoritative (`docs/developer/CODA.md:305`) — new provider templates will not appear and removed ones will linger. Share the availability predicate across creation and conversion, and populate templates from capabilities with an unknown/custom fallback.

**[blocking] Instrument the new degradation ladder.** `docs/developer/TELEMETRY.md:37-46` requires a typed signal for fallback and degradation, and there are now five distinct ways the terminal can be unavailable — unsupported core, missing or unregistered provider, credential/role refusal, provider outage, session failure — with no signal for any of them (`CodaBackendStatus.tsx:6`). Without it, "nobody uses the terminal" and "it is broken for everyone below 13.1" look identical. A low-cardinality outcome event covering those five states, recording no commands, ids, URLs or guide content, would close it.

**[blocking] Align canonical guidance with the SDK boundary this PR establishes.** `.cursor/rules/coda.mdc:34,43,132` and `docs/developer/CODA.md:44` still teach raw Live address splitting, subscribe/publish and frame parsing — the exact surface this PR deletes — mislocate local symbols, and over-route any `CodaClient` work to the provider repo. The changed drift skill also adds a bucket that detects a new route consumer with no Phase 2 action (`.cursor/skills/prevent-doc-drift/SKILL.md:66`). Rewrite these around `CodaSession`, distinguish provider implementation from local adapter work, and complete the drift rule.

**[blocking] Keep touched quality debt from growing.** `src/integrations/coda/coda-api.test.ts:1` adds a header claiming coverage for two URL builders this PR deletes — trim it. More substantively, extract the new Coda challenge lifecycle from `ChallengeBlock`, now 414 lines and 23 hooks (`challenge-block.tsx:215`): coordinating transport state and rendered state in that component is what produced the wrong-session race above, so this is not abstract debt.

**[question] Two items look outside this PR's scope to me. Your call whether to file them, defer them, or drop them — I have not filed anything.**

- `docs/design/AGENT_HARDENING.md:141,145,148` still assigns terminal and SSH work to Pathfinder files this PR deletes. That reads like design-review work rather than something to fold in here.
- Out-of-diff comments in `pkg/plugin/app_platform_identity.go:17`, `completion_records.go:62` and `contract_fixtures_test.go:68` retain the old identity-doc pointer and state that this PR deletes `resources.go`, which survives.

**[question] On scope:** this is one architectural goal but not one reviewable concern — provider rollout, SDK adoption, transport replacement, challenge lifecycle, auth and gating changes, operator migration, authoring policy, backend deletion, test deletion, CI, and canonical agent guidance across 65 files and 19 commits, with cutover and rollback evidence interleaved with the deletion. I am not asking you to split it after the fact, but flagging that the same merge bar applies either way: the consumer regressions above fixed, and a real boundary job, before the old backend deletion is accepted.

Jayclifford345:
Pushed as 02d824cc, rebased onto current `main` (the `CONCERNS.md` conflict was main's table reformat plus its new `data-check` row against this branch's `coda-terminal` rows — resolved by taking main's table and re-applying the four coda/go-backend rows and the anchor row; the diff against main is exactly those five lines).

Six of the ten blocking items are fixed, one is fixed differently from the suggestion, and three I'm pushing back on. `npm run check` is clean end to end: typecheck, lint (0 errors), prettier, `lint:go` (0 issues), `go build`/`go vet`/`go test ./pkg/...`, 444 suites / 7097 tests, and the governance tests.

## Fixed

**Bind challenge setup to the newly requested session.** This was the real bug, and it was worse than the finding says — there were two.

`openTerminal` now returns `Promise<string | null>` resolving with the session the caller may use, and `runSetup`/`runExec` take that id as an argument rather than reading `terminalCtx.sessionId` off a render. Close to your suggested shape:

```ts
const nextSessionId = await terminalCtx.openTerminal(vmOpts);
if (nextSessionId) { runSetup(nextSessionId); }
```

The second bug: `statusAtStartRef.current` was `'connected'` for challenge A, so when the reconnect *completed* the effect's `status === statusAtStartRef` guard read the new `connected` as "nothing changed" and returned early. Setup never ran at all — the block sat on Provisioning until Cancel. So the old code either exec'd against a deleted session or stalled, depending on timing.

The context resolves a waiter only after it has seen the session leave `connected`, so a reconnect cannot resolve with the id it is tearing down; waiters are also resolved `null` on `error`, on an explicit `disconnect`, and on unmount, so nothing awaits a connect that was cancelled. The effect keeps a `staleSessionIdRef` for the same invariant on the late-arrival path.

Two tests, and the regression one is mutation-checked — restoring the rendered-id read makes it fail:

```
ok  runs setup against the session openTerminal resolved, not the one that was live at click time
ok  does not start setup when the requested session never arrives
```

**Preserve the terminal on supported Grafana versions.** Fixed, essentially as suggested. Below 13.1 the availability probe now asks the provider instead of treating core's missing helper as a missing plugin — `/capabilities` is served by `grafana-coda-app` itself, so a response means installed and enabled and a failure means not reachable, and `CodaClient` suppresses the toast so a miss is silent. `loadCodaCapabilities` reuses that same read, so it is one request rather than two, pinned by a test. This also closes the previous round's "the config page blames the wrong component": `plugin_not_installed` is no longer reported on a 12.3–13.0 stack where the plugin is fine, and `plugin.json`'s `>=12.3.0-0` is now honest rather than aspirational.

**Give every sandbox block the same graceful-degradation contract.** `codaConfigGateMessage` / `codaUnavailableMessage` moved out of `challenge-block.tsx` into `useCodaAvailability.hook.ts`, parameterised by a subject phrase, and `terminal` and `terminal-connect` now use them. A gated-away panel means the reason is stated instead of a Connect button wired to nothing. `terminal` still offers Copy — it never needed a sandbox. Two tests on `terminal-step`, including that Copy survives.

**Use one provider-aware authoring contract.** Both halves. `CODA_BLOCK_TYPES` and `useCodaBlockTypesAvailable()` are now shared by `BlockPalette` and `TypeSwitchDropdown`, so a conversion cannot produce a block the palette says cannot exist. VM templates come from `capabilities.templates` via `useCodaTemplateOptions`, replacing the three hardcoded literals in both forms, with the legacy trio kept only as a fallback for a backend too old to advertise the list. One thing worth flagging that you did not ask for: the authored value is always kept in the options, because otherwise a template the provider stopped offering silently disappears from the form and an author editing an unrelated field saves the block with the VM changed under them. Test for that.

**Instrument the new degradation ladder.** `pathfinder_sandbox_unavailable` with a closed four-value `reason` (`terminal-disabled`, `plugin-missing`, `role-forbidden`, `panel-not-registered`) plus `blockType`. No ids, commands, URLs or guide content. Emitted once per block that degraded, and only when the rung changes, so the probe resolving mid-mount is not a second event. Derived from the same ladder as the message so a rung cannot produce one without the other. You were right that this shouldn't wait for the un-darkening — the whole point is that an unavailable sandbox and an unused one are otherwise the same absence of signal.

**State the credential lifecycle accurately.** Corrected. The note now says Interactive learning no longer *reads* those five settings, that values already stored stay there because Grafana never exposes an encrypted setting's value back so there is no supported way to blank one, that the retention is what makes a downgrade work, that whether the old refresh token is still live Coda-side is a question for a Coda administrator, and that an admin can delete the plugin settings through the API once the downgrade window has passed.

**Keep touched quality debt from growing** (first half). The `coda-api.test.ts` header no longer claims coverage for the two deleted URL builders.

**Do not grow the tier-violation ratchet silently** (the documentation half). The exception is now written into both canonical tier descriptions — `AGENTS.md` and `systemPatterns.mdc` — naming both entries and why they exist (a dynamic import that keeps the integration out of the requirements chunk when the feature is off).

**Align canonical guidance with the SDK boundary**, and the round-1 doc items. Checked each at this head: `coda.mdc` already teaches the opaque `channel` and `sessionChannelAddress` (lines 44–51, 179), `forwardIdentityHeaders` is gone from `secure/SKILL.md` and `CONCERNS.md`, the stale `isAllowedCodaURL` / `IsAllowedRelayURL` / `setAuthHeader` citations are gone, `techContext.mdc` no longer lists `gorilla/websocket`, the `AGENTS.md` "every route" overstatement is gone, `codaErrorCodeMessage` is one shared table, `vm_quota_exceeded` no longer advises closing a terminal, `useCodaOptions` reads `getCapabilities()` rather than two unmodelled URLs, and the `useTerminalLive` docblock no longer describes the deleted backend. What was left and is now fixed: `AGENT_HARDENING.md` D4 was still assigning terminal/SSH audit logging to `pkg/plugin/stream.go` and `terminal.go`, which this PR deletes — the row now routes it to the Coda repo and says there is nothing to build here. Also `ParseSettings` no longer unmarshals a `jsonData` it maps no field from, so a malformed blob can't disable the App Platform proxies over a value nothing reads.

One unrelated fix rode along because my telemetry import made it fail: `createConstantsMock` in the section harness replaced the whole `src/constants` barrel, so six interactive-section suites broke the moment anything in their tree read a different export from it. It spreads `requireActual` now.

## Fixed, but not the way you asked

**"Do not grow the tier-violation ratchet"** — I kept the allowlist entry rather than porting `checks/coda.ts` to a lower tier or adding a registration seam. `checks/terminal.ts` has the identical shape and the identical justification, and it is on `main`. Adding a registration seam for the second instance of a two-instance pattern buys indirection, not architecture; the honest options are to leave both as documented exceptions or to do the seam for both in a PR that is about that. I've documented it in both canonical descriptions as the drift skill requires, which is the part that was actually missing.

## Pushing back

**"Make the documented local stack actually work."** I don't think this is a defect in this PR, and I can't verify the mechanism from here. Registration is a manual step by design and always has been: an administrator enters an enrollment key on the Coda plugin's own configuration page and saves, which redeems it. `CODA.md` says so in three places, and the upgrade note is built around it. The overlay's job is to mount and provision the plugin, which it does. `409 no_service_account_token` is a claim about `grafana-coda-app`'s internals — whether it needs `externalServiceAccounts` and `GF_AUTH_MANAGED_SERVICE_ACCOUNTS_ENABLED` is a question for that repo, and adding two Grafana feature toggles to our compose on the strength of an error string I can't reproduce is exactly the kind of unverified change I'd expect a review to catch.

What I have done is remove the ambiguity that let the finding land: the overlay header and a new `LOCAL_DEV.md` section now say explicitly that mounting the plugin is not registering it, and that until registration the terminal is hidden — the documented degraded state, not a broken stack. If the Coda team confirms the toggles are required, that is a two-line follow-up and I'll take it.

**"Add a real consumer/provider contract job in CI."** Not in this PR. It needs a pinned, published `grafana-coda-app` build for CI to mount, which does not exist yet — that's a cross-repo dependency with its own release cadence, and standing up a contract harness against a moving provider inside the PR that establishes the boundary gets the sequencing backwards. The correct order is: land the boundary, get the provider publishing versioned artifacts, then pin one in CI. Worth an issue against both repos, and I'd rather file that than fake it here.

The half of this that *was* fair, I've taken: the body's claim that CI covers the two-plugin e2e suite is wrong, and I've corrected it rather than leaving it.

**"Extract the Coda challenge lifecycle from `ChallengeBlock`."** Declining for now, and I want to be precise about why, because your framing — "coordinating transport state and rendered state in that component is what produced the wrong-session race" — is the strongest argument in the review. It's also, I think, not quite the diagnosis. The race came from one specific thing: a fire-and-forget `openTerminal` whose completion the caller could not observe, so the caller read the rendered id instead. Making the operation awaitable fixes that at the seam where it was wrong, and the fix would be identical inside an extracted hook. A 400-line component with the right seam is safer than a 250-line component and a hook with the wrong one. The extraction is still worth doing; it just isn't what prevents this bug from recurring, and doing it in the same PR would bury a verified two-line-invariant fix inside a refactor of the file it lives in.

## Answered

**Renovate anchoring.** Not actionable in this repo — `renovate.json` here only extends `grafana/grafana-renovate-config` presets and carries no `matchPackageNames` of its own, so the unanchored `/^@grafana/` you found lives in the shared preset. Worth raising there; a change here can't fix it. Pinning `1.1.1` exactly is a judgement call I'd rather leave to a supply-chain pass than make silently in this diff.

**Scope.** I take the point and I'm not going to argue the PR is one reviewable concern. What I'd note is that the split you'd want is mostly not available: the backend deletion and the SDK adoption are the same commit in any ordering that keeps `main` working, since the frontend cannot consume both transports at once. The pieces that *were* separable — the terminal UI itself, the config page, the docs — are separable because they didn't change shape, which is also why the diff is 65 files and mostly deletion.

**`checks/coda.ts:5`.** Already softened at this head — it says the backend is *asked* to gate on the ready file, not that it always does.

moxious (dismissed):
Thanks for the thorough response. I verified the old-session regression at `02d824cc`, including the claimed mutation check: restoring the rendered session-id read makes the new test fail with both execs going to the old session. I also verified the pre-13.1 capabilities fallback and single request, all three block degradation paths, provider-backed template handling, credential wording, tier documentation, telemetry, and the current 35/0/5 CI result.

I accept the pushback on the contract job, local-registration claim, and challenge-hook extraction. The first two depend on evidence or artifacts this PR does not have, and the lifecycle diagnosis correctly identifies the awaitable seam as the cause of the race.

One non-blocking cleanup remains: `.cursor/rules/coda.mdc` and `docs/developer/CODA.md` still show the deleted raw Live flow, reference nonexistent `sessionChannelAddress`, and state the old Grafana 13.1 terminal floor even though this round added the capabilities fallback. Please clean that up here if convenient, or track it immediately as a small follow-up. With that minor noted, this is good enough to merge.

tomglenn (approved):

Issue #1469, opened 2026-07-30
Learning paths: migrate static/bundled paths to bundled `path` packages (unify with journey rendering)


moxious:
## Summary

Follow-up to #1467. That ticket delivers landing pages + table of contents + boundary-aware navigation for **URL-based (cloud) learning paths**, and deliberately excludes **static/bundled paths** (`src/learning-paths/paths.json` — e.g. `getting-started`, `observability-basics`) because they have no journey root and no cover `content.json`.

This ticket closes that gap by **migrating static paths to bundled `path` packages**, so they render through the exact same learning-journey pipeline as cloud paths — inheriting the cover/landing page, the milestone TOC, and back/forward boundary disabling for free, with no bespoke "synthesized landing page" code.

**Depends on #1467** (the shared React TOC component and launch-routing land there first).

---

## Why this is the right unification (not a wide refactor)

The rendering pipeline is **manifest-driven and already resolves bundled milestone IDs** — most of the work is data authoring, not new rendering code:

- Package manifests support `type: 'guide' | 'path' | 'journey'` (`src/types/package.types.ts:93`); `getPackageRenderType` maps `path`/`journey` → `learning-journey` rendering (`src/types/package.types.ts:108-115`).
- A `path` manifest carries `milestones: string[]` (`src/types/package.types.ts:153`). `fetchPackageContent` resolves those into a full `LearningJourneyMetadata` with a milestone-0 cover (`src/docs-retrieval/content-fetcher/package-content.ts:188-244`).
- `resolvePackageMilestones` resolves each milestone ID through the **package resolver, which already handles bundled content** (`src/docs-retrieval/content-fetcher/package-content.ts:68-108`).
- Bundled dirs already ship `content.json` + `manifest.json` (e.g. `src/bundled-interactives/first-dashboard/`), and `bundled:{dir}/content.json` is a first-class URL shape (`src/docs-retrieval/content-fetcher/bundled.ts:31-34`).
- **Gap today:** every bundled package is `type: "guide"`; no bundled `path` package exists, and static paths live in `paths.json` as a separate `LearningPath` construct that just lists guide IDs.

So a bundled `type: path` package with `milestones: ["welcome-to-grafana", ...]` + a cover `content.json` renders identically to a remote cloud path.

---

## The load-bearing risk: completion + badge reconciliation

This is the part that needs care and is somewhat one-way-door on the data model:

[...]

Issue #12, opened 2025-06-12
feat: "More to Explore" Docs Section


Jayclifford345:
### 📚 "More to Explore" Docs Section

---

#### 🧠 Overview

Some milestones include a **“More to explore”** section that links to optional docs for users who want to dig deeper into a particular topic.

---

#### ✅ Current Behavior

- These links appear under a **“More to explore (optional)”** heading.
- Controlled via **front matter** in the doc.
- Example: links to sending data from multiple Prometheus instances.

---

#### 🤔 Proposed Improvements

- Consider placing these optional docs under a **visually distinct, consistently styled heading**.
- Make it clearer that this section is **optional** and **exploratory**, not part of the core journey flow.

---

#### 📌 Design Thoughts

- Could use collapsible UI or a subtle visual divider.
- Ensure this section does not distract from milestone completion.

Let’s decide if we want to keep this pattern as-is or refine its visual treatment.

PR #1473, opened 2026-07-30
feat(learning-paths): land cloud paths on cover page with a table of contents

### LINKED ISSUE #1467: Learning paths (cloud): land on the cover page with a table of contents, and disable back/forward at boundaries
## Summary

When a user launches a learning path from the **My Learning** page, Pathfinder skips the path's cover page (the `content.json` at the path root) and jumps straight to "module 1" (the first milestone). This ticket makes launching a path land on that cover/landing page, renders a milestone **table of contents** on it, and makes the back/forward buttons **boundary-aware**.

**Scope is URL-based (cloud) learning paths only** (`src/learning-paths/paths-cloud.json`). Static/bundled OSS paths are intentionally out of scope here and tracked in **#1469** (migrate them to bundled `path` packages so they render through the same pipeline). See "Scope decision" below for why.

Design questions are resolved (see "Resolved decisions"). The investigation is grounded with file:line anchors so this ticket is self-contained for an implementing agent.

### Deliverables

1. **Land on the cover page** — launching a not-started URL-based path opens the root `content.json` (milestone 0) instead of module 1.
2. **Table of contents** — a milestone list rendered on the cover page (mirrors the My Learning card's "Learning paths expander").
3. **Boundary-aware nav** — back disabled on the cover page, forward disabled on the last milestone.

---

## Background: two parallel models

| Model                                      | Where                                                  | What it knows                                                                                                                                             |
| ------------------------------------------ | ------------------------------------------------------ | --------------------------------------------------------------------------------------------------------------------------------------------------------- |
| **learning-paths** (My Learning page)      | `src/learning-paths/`, `src/components/LearningPaths/` | `LearningPath.guides[]` — a flat ordered list of **milestones only**. The cover page is **not** in it.                                                    |
| **learning-journey** (docs-panel renderer) | `src/docs-retrieval/`, `src/components/docs-panel/`    | `LearningJourneyMetadata` with `currentMilestone` where **0 = cover page**, `1..N` = milestones. Cover-page rendering + boundary-aware nav already exist. |

The seam between these two models is the whole story: the launch flow lives in learning-paths; the cover/milestone rendering lives in learning-journey.

---

## Resolved decisions

1. **Launch target — Start lands on cover; Continue resumes.** Fresh path (`progress === 0`) → cover/landing page. In-progress "Continue" → current module (existing behavior). The working back button always returns to the cover, so returning users lose nothing.
2. **Scope — URL-based (cloud) paths only.** Static/bundled OSS paths have no journey root / no `content.json`; unifying them is tracked in **#1469** and deliberately deferred to avoid touching the completion/badge boundary in this PR. (This narrows an earlier "also synthesize static paths" note — synthesizing a bespoke static landing page would be throwaway work once #1469 makes static paths render like cloud paths.)
3. **TOC rendering — React component, not HTML injection.** Per-milestone completion checkmarks come from async storage, which is awkward in the HTML-string path. React is clean and testable, and the component is an extraction of the My Learning card's guide list so it is reused in three places.
4. **Nav-consistency nits — deferred to a follow-up** (not #1469; a separate small issue). The injected HTML bottom-nav _hides_ Prev/Next at boundaries rather than disabling, and Alt+←/→ shortcuts don't gate on `canNavigate*` (they no-op safely). Neither blocks this work.

---

## Feature 1 — Land on the cover page

**The cover page is deliberately filtered out before My Learning sees it:**

- `src/learning-paths/fetch-path-guides.ts:71-72` drops any `index.json` item carrying `params.grafana.skip` (the cover). So `guides[0]` is milestone 1, never the cover. (Mirrored at `src/docs-retrieval/content-fetcher/metadata-extract.ts:148-152`.)
- On "Start", `handleContinue` opens `guides[0]`: `src/components/LearningPaths/LearningPathCard.tsx:44-50`.
- `handleOpenGuide` resolves `getGuideUrlForPath(...) ?? parentPath.url`: `src/components/LearningPaths/MyLearningTab.tsx:363-386`. **`parentPath.url` is exactly the cover-page base URL** — used only as a fallback today.

**The cover page already renders fully.** Loading content at the base URL:

- `findCurrentMilestoneFromUrl` returns `0` for the base URL: `src/docs-retrieval/content-fetcher/metadata-extract.ts:202-227`.
- `currentMilestone === 0` triggers cover decoration: `src/docs-retrieval/content-fetcher/package-content.ts:243-245` → `injectJourneyExtrasIntoJsonGuide` (`src/docs-retrieval/content-fetcher/cover-page.ts:120-144`) + `generateJourneyContentWithExtras` (`src/docs-retrieval/learning-journey-helpers.ts:156-193`).

**Implementation:** for a not-started URL-based path, route the launch to `parentPath.url` (milestone 0) rather than `guides[0]`'s milestone URL — a deliberate special-case at the launch decision point (`LearningPathCard.tsx:44-50` and/or `MyLearningTab.tsx:363-386`), gated on `progress === 0`. This is **not** an un-skip of the `grafana.skip` filter.

**Verify during implementation:** launching the base URL through `prepareGuideLaunch` (`src/components/docs-panel/utils/prepare-guide-launch.ts:62-109`) actually produces `learningJourney.currentMilestone === 0` end-to-end.

---

## Feature 2 — Table of contents on the landing page (the main net-new work)

- The My Learning expander TOC is **inline React**, not extracted: `src/components/LearningPaths/LearningPathCard.tsx:183-206` — a `guides.map()` over `PathGuide[]` (`{ title, completed, isCurrent }`) with per-item check/circle icons at `:191-203`.
- The journey TOC data is available: `LearningJourneyMetadata.milestones[]`, each `{ number, title, url, ... }` (`src/types/content.types.ts:47-94`).
- The cover page today renders "what to expect" + "Ready to begin" but **no milestone list**.

**Implementation:**

- Extract the guide-list block from `LearningPathCard.tsx:183-206` into a small reusable component taking `guides` (styles in `src/components/LearningPaths/learning-paths.styles.ts` — `guideList`, `guideItem`, `guideIcon*`, `guideTitle`). Reused by: the My Learning expander, the cover-page TOC here, and (later) #1469's static paths.
- Render it on the cover page from `learningJourney.milestones`. Completion status via `milestoneCompletionStorage` / `useLearningPaths().getPathGuides`.
- **Confirm the insertion point:** check how the cover-page content mounts in the panel (block renderer vs `dangerouslySetInnerHTML`) so the React TOC has a clean home — this is the one open implementation-level design detail.

---

## Feature 3 — Disable back on cover, forward on last milestone (rides on Feature 1)

**The disabling logic already exists and is already correct** in the shared toolbar:

- `src/components/docs-panel/components/LearningJourneyMilestoneToolbar.tsx:197` — back is `disabled={!panel.canNavigatePrevious() || activeTab.isLoading}`; `:217` — forward is `disabled={!panel.canNavigateNext() || activeTab.isLoading}`.
- `canNavigatePrevious()` returns **false on milestone 0**; `canNavigateNext()` returns **false on the last milestone**: `src/components/docs-panel/docs-panel.tsx:701-709`, backed by `getPreviousMilestoneUrl` / `getNextMilestoneUrl` (`src/docs-retrieval/learning-journey-helpers.ts:57-86`). Disabled styling exists (`src/styles/docs-panel.styles.ts:808-811`). One shared toolbar serves sidebar / fullscreen / floating.

**Why back appears "always enabled" today:** you never _land_ on the cover page (Feature 1), so you always start at milestone 1 where back→cover is legitimately enabled. **Once Feature 1 routes entry to milestone 0, back is automatically disabled there — no new nav code needed.** Forward-at-last already works.

---

## Key files (index)

- Launch decision: `src/components/LearningPaths/LearningPathCard.tsx:44-50`, `src/components/LearningPaths/MyLearningTab.tsx:363-386`
- Cover-page skip filter: `src/learning-paths/fetch-path-guides.ts:71-72`, `src/docs-retrieval/content-fetcher/metadata-extract.ts:148-152`
- Cover-page rendering: `src/docs-retrieval/content-fetcher/cover-page.ts`, `src/docs-retrieval/content-fetcher/package-content.ts:243-245`, `src/docs-retrieval/learning-journey-helpers.ts:156-193`
- Boundary nav (already correct): `src/components/docs-panel/components/LearningJourneyMilestoneToolbar.tsx:197,217`, `src/components/docs-panel/docs-panel.tsx:701-709`, `src/docs-retrieval/learning-journey-helpers.ts:57-86`
- Milestone metadata model: `src/types/content.types.ts:47-94`
- TOC to extract/reuse: `src/components/LearningPaths/LearningPathCard.tsx:183-206`, `src/learning-paths/learning-paths.hook.ts` (`getPathGuides`)

---

## Out of scope (tracked elsewhere)

- **Static/bundled OSS paths** → **#1469** (migrate to bundled `path` packages).
- **Nav-consistency nits** (bottom-nav hide-vs-disable; keyboard-shortcut gating) → separate small follow-up.

_Investigation-only ticket; no code changed yet. Related: #1460 (guide-launch channel unification), #1463 / #1462 (My Learning layout)._

moxious:
## Summary

Launching a not-started URL-based (cloud) learning path from **My Learning** now opens the path's cover page (milestone 0) with an "In this path" table of contents showing per-milestone completion, instead of jumping straight to module 1. Because you now land on the cover, the shared toolbar's **Back** button is correctly disabled there with no new navigation code. A companion fix aligns milestone-completion storage keys so those table-of-contents checkmarks are accurate on the flow this feature targets.

## What to look at

- **`src/lib/user-storage.ts` + `src/lib/learning-journey-url.ts`** — the completion-key alignment, the load-bearing fix. `milestoneCompletionStorage` now canonicalizes keys through `getLearningJourneyBaseUrl` on both read and write: `getStoredMilestoneSlugs` matches any stored key whose canonical base (or trailing-slash-stripped launch URL) equals the lookup key, and `markCompleted` writes the canonical key, the raw launch URL, and any pre-existing matching rows. The URL helpers moved to a new tier-1 `lib/` module because tier-1 `user-storage` cannot import from tier-2 `docs-retrieval`; `url-utils.ts` and `learning-journey-helpers.ts` now re-export them.
- **The four completion writers** (`LearningJourneyMilestoneToolbar.tsx`, `DocsPanelContentArea.tsx`, `useLastMilestoneAutoComplete.ts`, `link-handler.hook.ts`) — all switched from the raw `activeTab.baseUrl` to the normalized `learningJourney.baseUrl`, so writes and reads agree on the key.
- **`src/components/content-renderer/content-renderer.tsx`** — the table-of-contents mount point. It renders `LearningPathTableOfContents` through a new generic `beforeContent` slot, gated on the existing `isJourneyCoverPage(content)` predicate plus `milestones.length > 0`. The gate is shape-based: it fires for any learning-journey cover with milestones — today that is cloud paths, and package/bundled `path` covers once they exist (see Out of scope).
- **`src/components/LearningPaths/`** — `GuideList` extracted from `LearningPathCard`'s inline list and reused by both the card and the new `LearningPathTableOfContents`; styles split into `getGuideListStyles` / `getTableOfContentsStyles` so the card owns its own gutter rather than the cover page inheriting card padding.
- **`src/components/LearningPaths/MyLearningTab.tsx`** — `handleOpenGuide` routes a fresh launch (`progress === 0`) to `parentPath.url` (the cover); "Continue" still resumes the current module. The `OpenResourceClick` event now carries a `launch_target: 'cover_page' | 'milestone'` discriminator.

## Why

Issue #1467: launching a learning path skipped its cover/landing page and dropped users on module 1, so they never saw the path overview or a milestone map, and Back appeared permanently enabled because there was nothing before module 1 to land on. Rendering the table of contents as a React sibling rather than injected HTML was deliberate — per-milestone completion comes from async storage, which is awkward and untestable in the HTML-string path, and a React component reaches every panel surface. The completion-key alignment was added after review found the table of contents read completions under a normalized key while every writer stored them under the raw launch URL (with a trailing slash), so freshly-completed milestones rendered unchecked on this feature's own surface.

## How to verify

1. In **My Learning**, open a not-started cloud path (e.g. "Monitor a Linux server"). Confirm it lands on the cover page showing an "In this path" milestone list, and the toolbar **Back** arrow is disabled.
2. Complete milestone 1, then click **Back** to return to the cover. Confirm the completed milestone now shows a green check — before the key-alignment fix it stayed an open circle.
3. Reload or reopen the tab and confirm completion state persists and still renders correctly.
4. Launch an **in-progress** path via "Continue" and confirm it resumes the current module, not the cover.
5. Confirm the table of contents does **not** appear on individual milestone pages, on an empty journey, or on single-doc content.

## Breaking changes

- **Milestone-completion storage keys**: completions are now canonicalized to the trailing-slash-free journey base URL. Reads reconcile legacy raw-URL rows and `markCompleted` continues to write the raw key too, so existing progress is preserved and no migration is required.
- **Analytics**: `OpenResourceClick` from `my_learning_tab` now emits `content_url` as the cover-page URL for fresh launches (previously module 1's URL) and adds a `launch_target` dimension. Downstream views that key on `content_url` to attribute a starting module will see a discontinuity from the cutover date; the new `launch_target` field is the intended discriminator.

## Reversibility

Reverting restores the previous launch and render behavior cleanly; the storage layer stays backward-compatible in both directions because `markCompleted` writes both canonical and raw keys. Emitted analytics rows are immutable, so the `content_url` / `launch_target` change on already-recorded events does not revert — confirm the cutover with the analytics owner.

## Out of scope

- **Static/bundled OSS paths** → #1469 (migrate to bundled `path` packages so they render through this pipeline). The shape-based gate will light up automatically for package `path` covers once they exist.
- **Nav-consistency nits** (bottom-nav hide-vs-disable; Alt+←/→ boundary gating) → separate follow-up noted in #1467.

Fixes #1467

### DIFF
--- a/src/components/LearningPaths/GuideList.tsx
+++ b/src/components/LearningPaths/GuideList.tsx
@@ -1,0 +1,52 @@
+/**
+ * Guide list
+ *
+ * Renders an ordered list of a learning path's guides/milestones with a
+ * per-item completion icon. Shared by the My Learning card expander and the
+ * cover-page table of contents so both read identically.
+ */
+
+import React from 'react';
+import { useStyles2, Icon } from '@grafana/ui';
+import { cx } from '@emotion/css';
+
+import type { PathGuide } from '../../types/learning-paths.types';
+import { getLearningPathCardStyles } from './learning-paths.styles';
+
+export interface GuideListProps {
+  guides: PathGuide[];
+  /** Shows a spinner row instead of the list while guides resolve. */
+  isLoading?: boolean;
+  loadingLabel?: string;
+}
+
+export function GuideList({ guides, isLoading = false, loadingLabel = 'Loading guides...' }: GuideListProps) {
+  const styles = useStyles2(getLearningPathCardStyles);
+
+  return (
+    <div className={styles.guideList}>
+      {isLoading ? (
+        <div className={styles.guideItem}>
+          <Icon name="fa fa-spinner" size="sm" />
+          <span className={styles.guideTitle}>{loadingLabel}</span>
+        </div>
+      ) : (
+        guides.map((guide) => (
+          <div key={guide.id} className={cx(styles.guideItem, guide.isCurrent && styles.guideItemCurrent)}>
+            <span
+              className={cx(
+                styles.guideIcon,
+                guide.completed && styles.guideIconCompleted,
+                guide.isCurrent && styles.guideIconCurrent,
+                !guide.completed && !guide.isCurrent && styles.guideIconPending
+              )}
+            >
+              {guide.completed ? <Icon name="check" size="sm" /> : <Icon name="circle" size="sm" />}
+            </span>
+            <span className={styles.guideTitle}>{guide.title}</span>
+          </div>
+        ))
+      )}
+    </div>
+  );
+}
--- a/src/components/LearningPaths/GuideList.tsx
+++ b/src/components/LearningPaths/GuideList.tsx
@@ -1,34 +1,26 @@
-/**
- * Guide list
- *
- * Renders an ordered list of a learning path's guides/milestones with a
- * per-item completion icon. Shared by the My Learning card expander and the
- * cover-page table of contents so both read identically.
- */
-
 import React from 'react';
 import { useStyles2, Icon } from '@grafana/ui';
+import { t } from '@grafana/i18n';
 import { cx } from '@emotion/css';
 
 import type { PathGuide } from '../../types/learning-paths.types';
-import { getLearningPathCardStyles } from './learning-paths.styles';
+import { getGuideListStyles } from './learning-paths.styles';
 
 export interface GuideListProps {
   guides: PathGuide[];
-  /** Shows a spinner row instead of the list while guides resolve. */
   isLoading?: boolean;
-  loadingLabel?: string;
+  className?: string;
 }
 
-export function GuideList({ guides, isLoading = false, loadingLabel = 'Loading guides...' }: GuideListProps) {
-  const styles = useStyles2(getLearningPathCardStyles);
+export function GuideList({ guides, isLoading = false, className }: GuideListProps) {
+  const styles = useStyles2(getGuideListStyles);
 
   return (
-    <div className={styles.guideList}>
+    <div className={cx(styles.list, className)}>
       {isLoading ? (
         <div className={styles.guideItem}>
           <Icon name="fa fa-spinner" size="sm" />
-          <span className={styles.guideTitle}>{loadingLabel}</span>
+          <span className={styles.guideTitle}>{t('myLearning.loadingGuides', 'Loading guides...')}</span>
         </div>
       ) : (
         guides.map((guide) => (
--- a/src/components/LearningPaths/LearningPathTableOfContents.test.tsx
+++ b/src/components/LearningPaths/LearningPathTableOfContents.test.tsx
@@ -1,8 +1,3 @@
-/**
- * Tests for the cover-page table of contents: it lists journey milestones and
- * marks per-milestone completion resolved from storage (issue #1467).
- */
-
 import React from 'react';
 import { render, screen, waitFor } from '@testing-library/react';
 import { LearningPathTableOfContents } from './LearningPathTableOfContents';
@@ -20,17 +15,6 @@ jest.mock('@grafana/i18n', () => ({
 
 jest.mock('../../lib/user-storage', () => ({
   milestoneCompletionStorage: { getCompleted: jest.fn() },
-}));
-
-// The real docs-retrieval barrel drags content-fetcher (→ @grafana/runtime) at
-// import time; the slug rule under test is a stable one-liner.
-jest.mock('../../docs-retrieval', () => ({
-  getMilestoneSlug: (url: string) =>
-    url
-      .replace(/\/(content\.json|unstyled\.html)$/, '')
-      .replace(/\/+$/, '')
-      .split('/')
-      .pop() || '',
 }));
 
 const getCompletedMock = milestoneCompletionStorage.getCompleted as jest.MockedFunction<
@@ -53,7 +37,12 @@ describe('LearningPathTableOfContents', () => {
     expect(screen.getByText('In this path')).toBeInTheDocument();
     expect(screen.getByText('Set up')).toBeInTheDocument();
     expect(screen.getByText('Explore')).toBeInTheDocument();
-    await waitFor(() => expect(getCompletedMock).toHaveBeenCalledWith(baseUrl));
+    await waitFor(() =>
+      expect(getCompletedMock).toHaveBeenCalledWith(
+        baseUrl,
+        milestones.map((milestone) => milestone.url)
+      )
+    );
   });
 
   it('shows a check for milestones whose slug is in the completed set', async () => {
--- a/src/components/LearningPaths/LearningPathTableOfContents.test.tsx
+++ b/src/components/LearningPaths/LearningPathTableOfContents.test.tsx
@@ -1,0 +1,66 @@
+/**
+ * Tests for the cover-page table of contents: it lists journey milestones and
+ * marks per-milestone completion resolved from storage (issue #1467).
+ */
+
+import React from 'react';
+import { render, screen, waitFor } from '@testing-library/react';
+import { LearningPathTableOfContents } from './LearningPathTableOfContents';
+import { milestoneCompletionStorage } from '../../lib/user-storage';
+import type { Milestone } from '../../types/content.types';
+
+jest.mock('@grafana/ui', () => ({
+  useStyles2: () => new Proxy({}, { get: (_t, p) => String(p) }),
+  Icon: ({ name }: { name: string }) => <span data-icon={name} />,
+}));
+
+jest.mock('@grafana/i18n', () => ({
+  t: (_key: string, fallback: string) => fallback,
+}));
+
+jest.mock('../../lib/user-storage', () => ({
+  milestoneCompletionStorage: { getCompleted: jest.fn() },
+}));
+
+// The real docs-retrieval barrel drags content-fetcher (→ @grafana/runtime) at
+// import time; the slug rule under test is a stable one-liner.
+jest.mock('../../docs-retrieval', () => ({
+  getMilestoneSlug: (url: string) =>
+    url
+      .replace(/\/(content\.json|unstyled\.html)$/, '')
+      .replace(/\/+$/, '')
+      .split('/')
+      .pop() || '',
+}));
+
+const getCompletedMock = milestoneCompletionStorage.getCompleted as jest.MockedFunction<
+  typeof milestoneCompletionStorage.getCompleted
+>;
+
+const baseUrl = 'https://grafana.com/docs/learning-paths/demo/';
+const milestones: Milestone[] = [
+  { number: 1, title: 'Set up', duration: '', url: `${baseUrl}set-up/content.json`, isActive: false },
+  { number: 2, title: 'Explore', duration: '', url: `${baseUrl}explore/content.json`, isActive: false },
+];
+
+describe('LearningPathTableOfContents', () => {
+  beforeEach(() => jest.clearAllMocks());
+
+  it('renders every milestone title with a heading', async () => {
+    getCompletedMock.mockResolvedValue(new Set());
+    render(<LearningPathTableOfContents milestones={milestones} baseUrl={baseUrl} />);
+
+    expect(screen.getByText('In this path')).toBeInTheDocument();
+    expect(screen.getByText('Set up')).toBeInTheDocument();
+    expect(screen.getByText('Explore')).toBeInTheDocument();
+    await waitFor(() => expect(getCompletedMock).toHaveBeenCalledWith(baseUrl));
+  });
+
+  it('shows a check for milestones whose slug is in the completed set', async () => {
+    getCompletedMock.mockResolvedValue(new Set(['set-up']));
+    render(<LearningPathTableOfContents milestones={milestones} baseUrl={baseUrl} />);
+
+    await waitFor(() => expect(document.querySelectorAll('[data-icon="check"]')).toHaveLength(1));
+    expect(document.querySelectorAll('[data-icon="circle"]')).toHaveLength(1);
+  });
+});
--- a/src/components/LearningPaths/LearningPathTableOfContents.tsx
+++ b/src/components/LearningPaths/LearningPathTableOfContents.tsx
@@ -1,0 +1,67 @@
+/**
+ * Learning path table of contents
+ *
+ * Milestone list rendered on a learning path's cover page (milestone 0). Mirrors
+ * the My Learning card's guide expander via the shared GuideList, but sources its
+ * items from journey metadata and resolves per-milestone completion from storage
+ * (issue #1467). Display-only — the cover page's "Ready to begin" button and the
+ * milestone toolbar own navigation.
+ */
+
+import React, { useEffect, useState } from 'react';
+import { useStyles2, Icon } from '@grafana/ui';
+import { t } from '@grafana/i18n';
+
+import type { Milestone } from '../../types/content.types';
+import type { PathGuide } from '../../types/learning-paths.types';
+import { milestoneCompletionStorage } from '../../lib/user-storage';
+import { getMilestoneSlug } from '../../docs-retrieval';
+import { logger } from '../../lib/logging';
+import { testIds } from '../../constants/testIds';
+import { GuideList } from './GuideList';
+import { getTableOfContentsStyles } from './learning-paths.styles';
+
+export interface LearningPathTableOfContentsProps {
+  milestones: Milestone[];
+  /** Journey base URL — the completion-storage key for this path's milestones. */
+  baseUrl: string;
+}
+
+export function LearningPathTableOfContents({ milestones, baseUrl }: LearningPathTableOfContentsProps) {
+  const styles = useStyles2(getTableOfContentsStyles);
+  const [completedSlugs, setCompletedSlugs] = useState<Set<string>>(new Set());
+
+  useEffect(() => {
+    let cancelled = false;
+    milestoneCompletionStorage
+      .getCompleted(baseUrl)
+      .then((slugs) => {
+        if (!cancelled) {
+          setCompletedSlugs(slugs);
+        }
+      })
+      .catch((error) => {
+        logger.warn('[LearningPathTableOfContents] Failed to load milestone completion', { error });
+      });
+    return () => {
+      cancelled = true;
+    };
+  }, [baseUrl]);
+
+  const guides: PathGuide[] = milestones.map((milestone) => ({
+    id: String(milestone.number),
+    title: milestone.title,
+    completed: completedSlugs.has(getMilestoneSlug(milestone.url)),
+    isCurrent: false,
+  }));
+
+  return (
+    <div className={styles.container} data-testid={testIds.learningPaths.tableOfContents}>
+      <h2 className={styles.heading}>
+        <Icon name="list-ul" size="md" className={styles.headingIcon} />
+        {t('coverPage.tableOfContents', 'In this path')}
+      </h2>
+      <GuideList guides={guides} />
+    </div>
+  );
+}
--- a/src/components/LearningPaths/LearningPathTableOfContents.tsx
+++ b/src/components/LearningPaths/LearningPathTableOfContents.tsx
@@ -1,13 +1,3 @@
-/**
- * Learning path table of contents
- *
- * Milestone list rendered on a learning path's cover page (milestone 0). Mirrors
- * the My Learning card's guide expander via the shared GuideList, but sources its
- * items from journey metadata and resolves per-milestone completion from storage
- * (issue #1467). Display-only — the cover page's "Ready to begin" button and the
- * milestone toolbar own navigation.
- */
-
 import React, { useEffect, useState } from 'react';
 import { useStyles2, Icon } from '@grafana/ui';
 import { t } from '@grafana/i18n';
@@ -15,15 +5,13 @@ import { t } from '@grafana/i18n';
 import type { Milestone } from '../../types/content.types';
 import type { PathGuide } from '../../types/learning-paths.types';
 import { milestoneCompletionStorage } from '../../lib/user-storage';
-import { getMilestoneSlug } from '../../docs-retrieval';
-import { logger } from '../../lib/logging';
+import { getMilestoneSlug } from '../../lib/learning-journey-url';
 import { testIds } from '../../constants/testIds';
 import { GuideList } from './GuideList';
 import { getTableOfContentsStyles } from './learning-paths.styles';
 
 export interface LearningPathTableOfContentsProps {
   milestones: Milestone[];
-  /** Journey base URL — the completion-storage key for this path's milestones. */
   baseUrl: string;
 }
 
@@ -33,20 +21,20 @@ export function LearningPathTableOfContents({ milestones, baseUrl }: LearningPat
 
   useEffect(() => {
     let cancelled = false;
-    milestoneCompletionStorage
-      .getCompleted(baseUrl)
+    void milestoneCompletionStorage
+      .getCompleted(
+        baseUrl,
+        milestones.map((milestone) => milestone.url)
+      )
       .then((slugs) => {
         if (!cancelled) {
           setCompletedSlugs(slugs);
         }
-      })
-      .catch((error) => {
-        logger.warn('[LearningPathTableOfContents] Failed to load milestone completion', { error });
       });
     return () => {
       cancelled = true;
     };
-  }, [baseUrl]);
+  }, [baseUrl, milestones]);
 
   const guides: PathGuide[] = milestones.map((milestone) => ({
     id: String(milestone.number),
--- a/src/components/LearningPaths/MyLearningTab.tsx
+++ b/src/components/LearningPaths/MyLearningTab.tsx
@@ -369,9 +369,20 @@ export function MyLearningTab({ onOpenGuide }: MyLearningTabProps) {
       // on the actual next module instead of the path base / first module
       // (issue #744). When dynamic data has not loaded yet, fall back to the
       // path's base URL.
+      //
+      // Fresh launch (no progress yet) lands on the path's cover page
+      // (milestone 0) instead of module 1 (issue #1467): the base URL is
+      // exactly the cover page, and its working Back button returns here so
+      // resuming users lose nothing. "Continue" (progress > 0) still resumes
+      // the current module below.
       if (parentPath?.url) {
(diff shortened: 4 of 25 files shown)


moxious:
@cursoragent review

src/components/LearningPaths/LearningPathTableOfContents.tsx
```diff
@@ -0,0 +1,67 @@
...
+}
+
+export function LearningPathTableOfContents({ milestones, baseUrl }: LearningPathTableOfContentsProps) {
+  const styles = useStyles2(getTableOfContentsStyles);
+  const [completedSlugs, setCompletedSlugs] = useState<Set<string>>(new Set());
+
+  useEffect(() => {
+    let cancelled = false;
+    milestoneCompletionStorage
+      .getCompleted(baseUrl)
```
Jayclifford345: **[blocking] The TOC reads milestone completion under a key that none of the writers use, so the ticks stay empty on the flow this PR targets.**

**Read key** — `baseUrl` here comes from `content.metadata.learningJourney.baseUrl` (`content-renderer.tsx:410`), which is `getLearningJourneyBaseUrl(finalUrl)` (`metadata-extract.ts:70`). That regex capture (`url-utils.ts:88-91`) can never emit a trailing slash:

```
https://grafana.com/docs/learning-paths/linux-server-integration
```

**Write key** — all four writers pass `activeTab.baseUrl` (`LearningJourneyMilestoneToolbar.tsx:160`, `DocsPanelContentArea.tsx:436`, `link-handler.hook.ts:534`, `useLastMilestoneAutoComplete.ts:55`), which is the launch URL verbatim (`docs-panel.tsx:426` `baseUrl: url`, never re-derived — `:881` is `t.baseUrl || fetchedContent.url`). For a My Learning launch that URL comes from `paths-cloud.json:16` and **does** carry the slash:

```
https://grafana.com/docs/learning-paths/linux-server-integration/
```

Lookups are exact map-key hits with no normalisation (`user-storage.ts:771` `data?.[journeyBaseUrl]`), so the read misses and returns an empty `Set`.

**Repro (single session):** open "Monitor a Linux server" from My Learning → "Ready to begin" → complete milestone 1 → click the toolbar Back arrow. The cover reloads in the same tab and shows the milestone you just finished as unchecked. Tab state persists and restores, so it's still wrong next session.

**How this happened** — there are two key conventions in `main` for two different stores, and this reader picked the wrong one. `docs-panel.tsx:551` carries an explicit comment saying the *journey-percentage* store keys on `learningJourney.baseUrl` so readers and writers match; the *milestone* store's writes stayed on raw `activeTab.baseUrl`. Corroboration that the slashed form is the live milestone key: `learning-paths.hook.ts:373` clears it with `path.url`.

**Suggested fix** — either:
1. Normalise inside `milestoneCompletionStorage` on **read and write** (strip trailing `/`). Normalising writes only would orphan every existing row; normalising reads too keeps them readable. This also fixes the resume case, where `tab.baseUrl` is a *module* URL and completions scatter across a third spelling.
2. Or source completion the way the My Learning card already does — slug membership in `progress.completedGuides`. `markMilestoneDone` already calls `markGuideCompleted(milestoneSlug)` (`learning-journey-helpers.ts:486`), so that source is populated and immune to base-URL drift.

**Severity note (recorded dissent):** one adversarial check argued this should be `low` — it's `check` vs `circle` icon only, with no effect on stored data, badges, percentages or navigation, and the competing key conventions plus the unnormalised slash already exist in `main` (`resetPath` vs `journeyCompletionStorage` has the same mismatch today). That's fair on root cause and blast radius. I'm still flagging it as blocking because per-milestone completion **is** the feature, and it's wrong on the PR's own target surface. Revert is clean either way — this PR never writes to the store.

src/components/content-renderer/content-renderer.tsx
```diff
@@ -396,6 +397,19 @@ export const ContentRenderer = React.memo(function ContentRenderer({
...
+  // storage, so it's a React sibling rather than injected HTML — the one shared
+  // mount point that reaches every surface (sidebar, floating, full-screen).
+  const journey = content.metadata.learningJourney;
+  const coverPageToc =
+    content.type === 'learning-journey' &&
+    journey &&
+    journey.currentMilestone === 0 &&
+    journey.milestones.length > 0 ? (
+      <LearningPathTableOfContents milestones={journey.milestones} baseUrl={journey.baseUrl} />
+    ) : null;
```
Jayclifford345: **[question] The gate is broader than the PR's stated scope, duplicates an existing predicate, and nothing tests it.**

Three separate points on these lines:

1. **Scope.** The PR says "Scope is URL-based cloud paths only; static/bundled OSS paths are out of scope (tracked in #1469)". This gate is shape-only, and `fetchPackageContent` produces exactly that shape for any `path`/`journey` manifest, CDN or `bundled:` — `package-content.ts:224-225` sets `currentMilestone = milestoneIndex >= 0 ? milestoneIndex + 1 : 0` and `:254` sets `type: renderType`. So package-backed path covers get the TOC too. No bundled path manifest with `milestones` exists in-tree today, so nothing breaks right now — but the containment the description claims isn't in the code, and #1469 will assume it is. Either narrow the gate or adjust the description.

2. **Duplicated predicate.** `isJourneyCoverPage(content)` (`learning-journey-helpers.ts:122`) already encodes three of these four conditions — and it currently has **zero callers** anywhere in `src/`. Reusing it here would remove the drift risk of two copies of "is this a cover page".

3. **Untested.** This is the load-bearing new integration point, and no test renders `ContentRenderer` — every existing suite stubs it (`DocsPanelContentArea.test.tsx:39`, `BlockPreview.test.tsx:9`, `GuideReaderOverlay.test.tsx:20`). A regression that flips the gate (milestone renumbering, the TOC leaking onto every milestone, or it appearing in the block-editor preview) passes CI today. A focused render test asserting `queryByTestId(testIds.learningPaths.tableOfContents)` present at `currentMilestone: 0` and absent for `currentMilestone: 1`, `milestones: []` and `type: 'docs'` would cover all three.

src/components/LearningPaths/LearningPathTableOfContents.tsx
```diff
@@ -0,0 +1,67 @@
...
+      .then((slugs) => {
+        if (!cancelled) {
+          setCompletedSlugs(slugs);
+        }
+      })
+      .catch((error) => {
+        logger.warn('[LearningPathTableOfContents] Failed to load milestone completion', { error });
+      });
```
Jayclifford345: **[suggestion] This `.catch` is unreachable, so the one degradation path has no signal at all.**

`milestoneCompletionStorage.getCompleted` already swallows everything internally and resolves with an empty set (`user-storage.ts:769-776`: `} catch { return new Set(); }`). It cannot reject, so this handler and its `logger.warn` never fire.

That matters beyond dead code: a genuine storage fault renders as an all-unchecked TOC that's indistinguishable from a genuinely fresh path — the same end state as the key mismatch above, which is why that bug leaves no telemetry trail. Per `docs/developer/TELEMETRY.md` a stable logger signal is sufficient for an ordinary failure, but there currently isn't one.

Either drop the `.catch` (and the now-unused `logger` import) since the storage layer owns the fallback, or move the signal into `getCompleted`'s existing `catch` where it can actually fire.

src/components/LearningPaths/LearningPathTableOfContents.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { milestoneCompletionStorage } from '../../lib/user-storage';
+import { getMilestoneSlug } from '../../docs-retrieval';
+import { logger } from '../../lib/logging';
+import { testIds } from '../../constants/testIds';
+import { GuideList } from './GuideList';
+import { getTableOfContentsStyles } from './learning-paths.styles';
+
+export interface LearningPathTableOfContentsProps {
+  milestones: Milestone[];
+  /** Journey base URL — the completion-storage key for this path's milestones. */
```
Jayclifford345: **[nit] QC8 — this comment asserts a contract the code doesn't currently honour.**

"the completion-storage key for this path's milestones" is precisely the claim the blocking comment above disproves: the writers key on `activeTab.baseUrl`, not this value. A hidden-invariant comment normally earns its place under AGENTS.md §Comments — but only while it's true. Fix it alongside the key mismatch, or drop it.

While you're here, the rest of the QC8 surface on the changed hunks (all non-blocking, `AGENTS.md` defaults to no comments):
- `GuideList.tsx:18` — `/** Shows a spinner row instead of the list while guides resolve. */` restates `isLoading?: boolean`.
- `content-renderer.tsx:444` — `/** Cover-page milestone TOC, rendered above the content… */` restates `coverPageToc?: React.ReactNode`.
- Whole-file docstrings on two small modules (6 lines on 52-line `GuideList.tsx`, 8 lines on 67-line `LearningPathTableOfContents.tsx`).
- `MyLearningTab.tsx:372-377` — the new block ends by restating the ternary two lines below it; the one non-obvious fact worth keeping is that the base URL *is* the cover page.

src/components/LearningPaths/LearningPathTableOfContents.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+  beforeEach(() => jest.clearAllMocks());
+
+  it('renders every milestone title with a heading', async () => {
+    getCompletedMock.mockResolvedValue(new Set());
+    render(<LearningPathTableOfContents milestones={milestones} baseUrl={baseUrl} />);
+
+    expect(screen.getByText('In this path')).toBeInTheDocument();
+    expect(screen.getByText('Set up')).toBeInTheDocument();
+    expect(screen.getByText('Explore')).toBeInTheDocument();
+    await waitFor(() => expect(getCompletedMock).toHaveBeenCalledWith(baseUrl));
```
Jayclifford345: **[suggestion] This assertion can't fail on the bug in the component it covers.**

The test supplies `baseUrl` as a prop and asserts `getCompleted` was called with that same string — self-consistency, not read/write alignment. Since `milestoneCompletionStorage` is mocked, the trailing-slash mismatch is invisible here, which is why CI is green on a broken tick display.

One integration-level test would close it: mock only the underlying user-storage backend, call `markMilestoneDone(<trailing-slash launch url>, slug)`, then read via `getCompleted(getLearningJourneyBaseUrl(launchUrl))` and assert the slug is found. It should fail today and become the regression guard after the fix.

Separately, `jest.mock('../../docs-retrieval')` at line 27 hand-rolls `getMilestoneSlug`, so the completion-matching assertion is validated against a reimplementation. The real helper (`learning-journey-helpers.ts:461`) has **no unit test anywhere**, and a second, *divergent* hand-rolled copy already exists at `LearningJourneyMilestoneToolbar.test.tsx:39` — that one doesn't strip `content.json`, so it returns `'content.json'` for the URL shapes this feature uses. Two mocks, two behaviours, real function unpinned. A small test for the real helper (trailing slash, `content.json`, `unstyled.html`, empty) would make both mocks legitimate; importing from `../../docs-retrieval/learning-journey-helpers` instead of the barrel would also avoid the `@grafana/runtime` pull that forced the mock.

src/components/LearningPaths/MyLearningTab.tsx
```diff
@@ -369,9 +369,20 @@ export function MyLearningTab({ onOpenGuide }: MyLearningTabProps) {
...
       if (parentPath?.url) {
-        const resolvedGuideUrl = getGuideUrlForPath(guideId, parentPath.id) ?? parentPath.url;
-        const guideTitle = getPathGuides(parentPath.id).find((g) => g.id === guideId)?.title;
+        const isFreshLaunch = getPathProgress(parentPath.id) === 0;
+        const resolvedGuideUrl = isFreshLaunch
+          ? parentPath.url
+          : (getGuideUrlForPath(guideId, parentPath.id) ?? parentPath.url);
+        const guideTitle = isFreshLaunch
+          ? parentPath.title
+          : getPathGuides(parentPath.id).find((g) => g.id === guideId)?.title;
```
Jayclifford345, line 375: **[question] `OpenResourceClick` now means something different for fresh launches — worth a discriminator.**

On a fresh launch this emits the **path** title and the **path base URL**; previously it was the first module's title/URL. The event shape is unchanged, so nothing in CI or the schema flags it, but the meaning of the dimension shifts: every first launch of a cloud path now reports the cover URL — the same value the recommendations surface emits — and the event carries no `path_id` or milestone number, so module-vs-path attribution rests entirely on `content_url`.

Anything in the warehouse that segments "which module do users start with" for `interaction_location: 'my_learning_tab'`, or joins `content_url` to a milestone dimension, will bucket new rows differently from historical ones in the same table — a silent discontinuity in a time series rather than a query error.

Either add a discriminator (`milestone: 0`, or `launch_target: 'cover_page' | 'module'`) so the two cases stay separable downstream, or confirm with the analytics owner that no view keys on `content_url` here and note the cutover date in the PR description. Emitted rows are immutable, so this half doesn't revert with the code.

src/components/LearningPaths/GuideList.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { cx } from '@emotion/css';
+
+import type { PathGuide } from '../../types/learning-paths.types';
+import { getLearningPathCardStyles } from './learning-paths.styles';
+
+export interface GuideListProps {
+  guides: PathGuide[];
+  /** Shows a spinner row instead of the list while guides resolve. */
+  isLoading?: boolean;
+  loadingLabel?: string;
```
Jayclifford345: **[nit] `loadingLabel` has no production caller, and the default string isn't translated.**

Both call sites rely on the default (`LearningPathCard.tsx:184`, `LearningPathTableOfContents.tsx:64`); the only place that passes it is `GuideList.test.tsx:35`. A prop exercised solely by its own test is worth deleting — inline the label and assert the real copy.

If you keep it: `'Loading guides...'` is user-visible and bypasses `t()`, while the sibling new component does translate its heading. This is inherited (`LearningPathCard.tsx` is untranslated throughout), so not a regression — but `t('myLearning.loadingGuides', 'Loading guides...')` would match this directory's existing `myLearning.*` namespace, and the card-wide i18n pass can stay a separate task.

src/locales/en-US/grafana-pathfinder-app.json
```diff
@@ -1,4 +1,7 @@
 {
+  "coverPage": {
+    "tableOfContents": "In this path"
+  },
```
Jayclifford345: **[nit] Out of the generator's sort order.**

`src/locales/i18next-parser.config.js:27` sets `sort: true`, so `npm run i18n-extract` will order `contextPanel` before `coverPage` and produce an unrelated diff for whoever next runs it. Moving this block after `contextPanel` (or committing the extractor's output) avoids that.

en-US-only is correct and matches existing practice (`myLearning.launchErrorTitle` is the same) — the other 20 locales fall back to the inline default. `i18n-sync` can stub them later.

src/constants/testIds.ts
```diff
@@ -295,6 +295,7 @@ export const testIds = {
     resetProgressButton: 'learning-paths-reset-progress',
     badgeToast: 'learning-paths-badge-toast',
     badgeToastDismiss: 'learning-paths-badge-toast-dismiss',
+    tableOfContents: 'learning-path-toc',
```
Jayclifford345: **[nit] Prefix diverges from its siblings — cheapest to align now.**

Every neighbour uses the `learning-paths-` prefix (`learning-paths-reset-progress`, `learning-paths-badge-toast`); this one is `learning-path-toc`. It currently has no consumers at all — no unit assertion, no Playwright spec (the new test selects by the copy `'In this path'` instead, which couples it to wording). Renaming a testid once E2E depends on it is a coordinated change, so `learning-paths-toc` now is nearly free.

Using it in the `ContentRenderer` mount-gate test suggested above would give it a consumer and let that test stop asserting on user-visible copy.

Jayclifford345 (commented):
## `/review` — routed PR review

**Classification:** `mixed` (product-runtime + a public DOM/i18n contract touch)
**Activated concerns:** `security`, `correctness-and-reliability`, `testing-and-verification`, `reversibility-and-one-way-door`, `cross-cutting-architecture`, `docs-retrieval-and-rendering` (2 signals) + tech-debt and doc-drift passes
**Coverage confidence:** high — the change's centre of gravity is well modelled by `docs/design/CONCERNS.md`.

All 29 CI checks are green (Test, Lint, Prettier, Typecheck, e2e across 5 Grafana versions), so this review is about semantics, not mechanics.

### Disposition: one issue to resolve before merge

The feature is well shaped — the `GuideList` extraction is clean, the cover-launch test renders the real component tree rather than a stub, and Feature 3's "no new code" claim genuinely holds (`docs-panel.tsx:706` implements `canNavigatePrevious()` as `getPreviousMilestoneUrlFromContent(content) !== null`, which is exactly what the new nav-boundary test pins). The blocker is one storage-key mismatch that makes the TOC's completion ticks wrong on this PR's own target flow — see the inline comment on `LearningPathTableOfContents.tsx`.

### Verified while reviewing (positive evidence)

- **No progress-load race.** `MyLearningTab.tsx:499` returns a skeleton while `isLoading`, so no card renders before `progress.completedGuides` resolves — `getPathProgress` can't transiently read 0 for an in-progress path. And `LearningPathCard.handleContinue` no-ops while `guides` is empty.
- **Button label stays consistent.** `progress === 0` ⇒ label is "Start", so landing on the cover matches the affordance; "Continue" still resumes the module.
- **No duplicate cover-page content.** I checked the live cover page for `linux-server-integration`: the authored "Here's what to expect" block lists *learning objectives* ("Describe the benefits of…", "Set up a Linux server integration…"), not module titles. The new "In this path" milestone list therefore adds distinct information rather than restating that block. Worth re-checking if any path authors the block as a module list.
- **Security: clean.** Remote `milestone.title` reaches the DOM only through JSX text interpolation; `milestone.url` never reaches the DOM at all. No new value class enters `prepareGuideLaunch` — `parentPath.url` was already the pre-PR fallback on that same line.
- **Instrumentation: adequate.** No new user action, latency budget, or panel surface; the ordinary failure path is covered by the free logger→Faro bridge per `docs/developer/TELEMETRY.md`. (Caveat in the inline comment on the dead `.catch`.)
- **Doc drift: none.**

### Tech debt (non-blocking, for awareness)

- `ContentWithVariablesProps` crosses 11 props in a 1519-line file, and the new one is feature-named (`coverPageToc`) inside an otherwise feature-agnostic renderer. A generic slot (`beforeContent?: React.ReactNode`) would let the next surface reuse it instead of adding prop #12.
- `GuideList` reads `getLearningPathCardStyles`, whose `guideList` padding (`0 12px 12px`) is a card-gutter value. Inside the TOC's own `padding: spacing(2)` container that yields a ~28px indent, and any future card-layout tweak silently changes the cover page. Extracting `getGuideListStyles` and letting the card own its gutter would decouple them.
- Pre-existing, not from this PR: `content-renderer.tsx:603` contains a raw NUL byte inside a template literal (also on `main`), which makes `grep`/`rg` silently skip the whole file without `-a`. `block-editor/LintBadge.tsx` has the same. Using `'\u0000'` would fix it — probably its own small PR.

---

*Reviewed against `docs/design/CONCERNS.md` routing. The top finding was independently reached by four concern reviewers and survived a three-skeptic adversarial pass (2 of 3 could not refute it; the third disputed severity, not mechanism — that dissent is recorded inline).*

Jayclifford345 (approved):

PR #1474, opened 2026-07-30
feat(e2e): support path (journey) packages

leslie-heinzen:
## Summary

Allows `pathfinder-cli e2e --package` to execute `path` packages by recursively expanding them into ordered leaf-guide chains. Hard `depends` relationships remain the only source of failure propagation, while the requested metapackage is recorded separately through additive multi-guide `selection` metadata.

Repository-wide runs also keep independent guides runnable when another root has an unavailable or structurally invalid prerequisite.

## What to look at

- [Package planning and hydration](https://github.com/grafana/grafana-pathfinder-app/blob/351da70bed0f1ca94c6ef0739cf51f927487937f/src/cli/e2e/guide-chains.ts) — trace how CNF dependencies, virtual providers, recursive milestones, mixed-edge cycles, and shared leaves become an ordered guide plan. Milestone order intentionally does not create hard dependency edges.
- [Local and remote package resolution](https://github.com/grafana/grafana-pathfinder-app/blob/351da70bed0f1ca94c6ef0739cf51f927487937f/src/cli/commands/e2e.ts) and [remote hydration](https://github.com/grafana/grafana-pathfinder-app/blob/351da70bed0f1ca94c6ef0739cf51f927487937f/src/cli/e2e/e2e-package.ts) — local metapackages are manifest-first so cover content stays optional, remote content identity is checked against the repository, and pre-run failures retain the selected root.
- [Multi-guide report contract](https://github.com/grafana/grafana-pathfinder-app/blob/351da70bed0f1ca94c6ef0739cf51f927487937f/src/cli/e2e/schemas/e2e-report.schema.ts) — `selection` is optional and additive, including for one-milestone and skip-only metapackage reports.

## Why

The E2E command previously understood only executable guides connected through hard dependencies. Selecting a path therefore tested neither its ordered milestones nor a path-level dependency on another metapackage, preventing scheduled E2E workers from treating a learning path as one execution root.

Courses were considered and deliberately excluded because their ordering is curatorial rather than an execution prerequisite and adds no unique leaf-guide coverage. Milestone ordering is likewise kept separate from `depends`: guides run in the declared learning order, but an unrelated milestone failure does not suppress later coverage.

## How to verify

1. Create a local path package with no cover `content.json`, two guide milestones, and a repository index beside the package tree. Run `pathfinder-cli e2e --package <path-directory> --repository <repository.json> --verbose` and confirm the plan contains the two leaf guides in milestone order.
2. Make the first milestone fail while leaving the second independent, then confirm the second still runs. Add a hard `depends` relationship from the second to the first and confirm it is instead reported as skipped because its prerequisite failed.
3. Resolve a published cloud path from a local-tier run with `--output <report.json>`. Confirm the skip-only multi-guide report contains `selection.id`, `selection.type`, and a structured pre-run skip for the root.
4. In a repository-wide run, make one guide depend on an unavailable package and include a separate independent guide. Confirm only the affected dependency chain receives `prerequisite_failed` while the independent guide remains in the execution plan.

## Breaking changes

None. The `selection` field is optional and additive to the open-world `1.0.0` multi-guide report contract. Existing guide runs and older multi-guide consumers continue to use leaf-guide or pre-run-skip identity.

## Reversibility

Reverting this PR stops path expansion and stops new reports from emitting `selection`. Reports already written with the optional field remain valid JSON for open-world consumers; there is no storage migration, persisted browser-state change, or cleanup requirement.

## Out of scope

- Course execution: courses are presentation-only compositions and do not add executable guide prerequisites.
- Implicit path expansion during `--remote`: repository sweeps remain leaf-guide populations; metapackages require explicit `--package` selection.
- Population-wide chain deduplication: shared-root optimization remains separate follow-up work.
- E2E-platform rollout: after merge, publish an immutable runner image, update the platform consumer, and refresh its imported schema provenance and image pins from that runner.

### DIFF
--- a/docs/developer/E2E_TESTING.md
+++ b/docs/developer/E2E_TESTING.md
@@ -15,8 +15,10 @@ This is the canonical implementation-backed reference for E2E CLI behavior. Veri
 ## Source map for agents
 
 - `src/cli/commands/e2e.ts` — Commander options, input resolution, dependency planning, pre-flight orchestration, clean-stack resets, cloud routing, and per-guide Playwright invocation.
+- `src/cli/e2e/e2e-local-package.ts` — local path/journey manifest validation, repository loading, milestone expansion, target gating, and guide hydration.
 - `src/cli/e2e/e2e-runner-contract.ts` — environment-variable contract between the CLI process and Playwright runner.
 - `src/cli/e2e/e2e-package.ts` — remote package and repository resolution, content fetch, schema validation, side-effect classification, and pre-run skip reasons.
+- `src/cli/e2e/guide-chains.ts` — pure package graph planning across hard dependencies, capabilities, and recursive milestones, followed by leaf-guide hydration.
 - `src/cli/e2e/e2e-targets.ts` — manifest `testEnvironment` to concrete target URL or skip reason.
 - `src/cli/e2e/cloud-provisioning.ts` and `src/cli/e2e/cloud-stack-pool-manager.ts` — shared-stack service-account isolation and pool-manager isolated stack leasing.
 - `tests/e2e-runner/guide-runner.spec.ts` — browser-side guide loading, pre-flight checks, DOM discovery, step execution, and result file writing.
@@ -62,7 +64,7 @@ npx pathfinder-cli e2e [options] [files...]
 | `--always-screenshot`                      | Capture screenshots on success and failure                                                                                                               | `false`                           |
 | `--clean`                                  | Run against an isolated docker-compose stack (project `pathfinder-e2e`, Grafana on `:3010`). Resets between dependency chains and tears down at the end. | `false`                           |
 | `--clean-ready-timeout-ms <ms>`            | How long to wait for the isolated Grafana to become healthy after a `--clean` reset                                                                      | `120000`                          |
-| `--package <dirOrId>`                      | Test a local package directory, or — when not an existing directory — a bare package ID resolved remotely via the recommender                            | None                              |
+| `--package <dirOrId>`                      | Test a local or remote guide, path, or journey package. Local paths/journeys also require `--repository` so milestone IDs resolve.                       | None                              |
 | `--tier <tier>`                            | Current environment tier (`local` or `cloud`); `cloud` guides are skipped on a `local` environment                                                       | `local`                           |
 | `--remote`                                 | Resolve and test every package from the CDN repository index                                                                                             | `false`                           |
 | `--repo-url <url>`                         | CDN base URL for `--remote`                                                                                                                              | Public package repository         |
@@ -81,8 +83,8 @@ The CLI accepts these input formats:
 1. **File paths**: `npx pathfinder-cli e2e ./my-guide.json ./another.json`
 2. **Bundled flag**: `npx pathfinder-cli e2e --bundled` (tests all guides in `src/bundled-interactives/`)
 3. **Bundled by name**: `npx pathfinder-cli e2e bundled:welcome-to-grafana`
-4. **Local package directory**: `npx pathfinder-cli e2e --package ./my-package/` (reads `content.json` + `manifest.json`)
-5. **Remote package ID**: `npx pathfinder-cli e2e --package alerting-101` (resolved via the recommender; see [Remote package-aware testing](#remote-package-aware-testing))
+4. **Local package directory**: `npx pathfinder-cli e2e --package ./my-package/` (reads `content.json` + `manifest.json`; add `--repository <path>` for a path or journey)
+5. **Remote package ID**: `npx pathfinder-cli e2e --package alerting-101` (guides, paths, and journeys resolve via the recommender; see [Remote package-aware testing](#remote-package-aware-testing))
 6. **Remote repository**: `npx pathfinder-cli e2e --remote` (every package in the CDN index)
 
 ## Exit codes
@@ -227,6 +229,7 @@ Key contract fields:
 - `errorCode`: structured failure code present on non-passing reports. Notable values: `TIER_MISMATCH` (guide requires a different environment tier), `SKIPPED_PREREQ` (a prerequisite guide failed), `REPORT_MISSING` (Playwright exited but wrote no results file), `AUTH_EXPIRED`, `NO_CAPACITY`, `PLAYWRIGHT_SPAWN_FAILED`.
 - `guide.contentDigest`: SHA-256 digest of the exact guide content executed
 - `guide.sourceUrl`: remote package source URL when available
+- `selection`: for an explicitly selected path or journey, the multi-guide report records the root package `id` and `type` separately from its executable leaf-guide reports
 
 ### Report validation
 
@@ -249,7 +252,7 @@ The emitted schema carries a stable `$id` (`https://grafana.com/schemas/pathfind
 
 The `e2e-report` and `e2e-multi-report` schemas are **open-world**: the exported JSON Schema does not include `additionalProperties: false`. This means additive optional fields introduced in a newer runner version are non-breaking — an orchestrator validating reports from a newer runner against an older schema copy will not reject the report. Consumers should configure their validators accordingly (for example, ajv's default behavior already allows extra fields unless explicitly set to strict mode). The `guide`, `manifest`, and other non-e2e schemas remain strict.
 
-Runs that execute more than one guide write a multi-guide report with aggregate summary fields plus the individual per-guide reports.
+Runs that execute more than one guide, or execute an explicitly selected path/journey with one milestone, write a multi-guide report with aggregate summary fields plus the individual per-guide reports.
 
 The dedicated `Dockerfile.e2e-runner` image contains the matching Playwright runner and Chromium, runs as a non-root user, and is published as a signed immutable `ghcr.io/grafana/pathfinder-e2e-runner:commit-<sha>` tag. Cloud Run Jobs should pin the resulting image digest rather than a mutable tag. The deterministic `always-passes` and `always-fails` package fixtures under `tests/e2e-runner/fixtures/` exercise the contract and artifact paths.
 
@@ -302,6 +305,14 @@ Before running, the CLI builds an execution plan from a `repository.json` index 
 - **Virtual capabilities**: a `depends` target may be a capability name; it resolves to whichever guide `provides` it.
 - **Failure propagation**: if a prerequisite fails, its dependents in the same chain are marked skipped (`prerequisite failed`) and not run; the runner continues with the next chain.
 - Only `depends` forms a chain. `recommends` and `suggests` are advisory and do not affect ordering. A `depends` cycle is a configuration error.
+  An explicitly selected `path` or `journey` recursively expands its `milestones` into executable leaf guides:
+
+- Milestones run in declared order and share one environment, including nested paths/journeys.
+- A metapackage or milestone can declare normal CNF `depends`, including a dependency on another path/journey or a virtual capability provider.
+- Milestone order is not an implicit hard dependency. If a milestone fails, later milestones continue unless their resolved `depends` includes the failed guide or metapackage.
+- A hard dependency on a metapackage requires all of that metapackage's executable leaves to pass before the dependent package can run.
+- Missing milestones, incompatible targets, and cycles crossing `depends` and `milestones` fail before Grafana provisioning or Playwright execution.
+- The metapackage cover `content.json` is not executed; only leaf guides are sent to Playwright.
 
 This ordering applies to every run. `--clean` additionally isolates each chain in its own environment (see below).
 
@@ -488,8 +499,8 @@ When the failure is not clearly a runner or contract bug, avoid changing Pathfin
 
 The CLI can resolve published guides instead of reading local files, then test them against the configured Grafana instance.
 
-- **By ID** (`--package <id>`): when the `--package` value is not an existing local directory, it is treated as a bare package ID and resolved through the recommender (`--resolver-url`, default `https://recommender.grafana.com`). The runner fetches the package's `content.json` and `manifest.json`, validates the content, and runs it.
-- **Whole repository** (`--remote`): fetches the CDN `repository.json` (`--repo-url`, default the public package repository) and tests every package in the index. Dependency-aware chaining still applies, driven by the remote index.
+- **By ID** (`--package <id>`): when the `--package` value is not an existing local directory, it is treated as a bare package ID and resolved through the recommender (`--resolver-url`, default `https://recommender.grafana.com`). Guides run directly. Paths and journeys expand recursively from the CDN `repository.json`, then fetch and run their leaf guides.
+- **Whole repository** (`--remote`): fetches the CDN `repository.json` (`--repo-url`, default the public package repository) and tests every leaf guide in the index. Dependency-aware chaining still applies. Metapackages are not expanded implicitly in a repository sweep; select one explicitly with `--package`.
 
 Guides are routed by their manifest's `testEnvironment.tier`:
 
@@ -519,7 +530,6 @@ Pool-backed run behavior:
 ### Known gaps and follow-up
 
 - Interactive SSO/Okta login (driving the identity provider's login UI) is not supported.
-- Path/journey (`milestones`) expansion is not yet implemented; `path` and `journey` packages are skipped as an unsupported type.
 
 ### Package outcomes
 
@@ -535,7 +545,7 @@ In remote modes a package can end in one of these states. `failed`, `provisionin
 | `skipped_unsafe_shared_stack` | Cloud guide requires an isolated stack, but none is configured | No            |
 | `resolution_failed`           | Recommender returned 404 or a network error                    | No            |
 | `fetch_failed`                | Could not fetch `content.json` from the CDN                    | No            |
-| `unsupported_type`            | Package is a `path` / `journey` (milestone expansion TODO)     | No            |
+| `unsupported_type`            | Repository sweep encountered a non-guide composition package   | No            |
 | `prerequisite_failed`         | A required prerequisite could not be resolved or run           | No            |
 | `skipped_prereq`              | A prerequisite in the same dependency chain failed             | No            |
 | `validation_failed`           | Fetched `content.json` failed guide schema validation          | **Yes**       |
@@ -545,6 +555,9 @@ With `--output`, pre-run skips are recorded under a `preRunSkipped` array, and e
 ```bash
 # Resolve and test a single published guide against local Grafana
 npx pathfinder-cli e2e --package alerting-101
+
+# Resolve and test every milestone in a published path
+npx pathfinder-cli e2e --package prometheus-lj
 
 # Test the whole published repository (local-tier guides run, cloud guides skip)
 npx pathfinder-cli e2e --remote --output results.json
--- a/src/cli/commands/e2e.ts
+++ b/src/cli/commands/e2e.ts
@@ -6,28 +6,30 @@
  */
 
 import { existsSync } from 'fs';
-import { dirname, isAbsolute, join, resolve } from 'path';
+import { join } from 'path';
 
 import { Command, Option } from 'commander';
 
 import { validateGuideFromString, toLegacyResult } from '../../validation';
-import {
-  loadGuideFiles,
-  loadBundledGuides,
-  loadRepositoryIndex,
-  bundledRepositoryPath,
-  type LoadedGuide,
-} from '../utils/file-loader';
+import { loadGuideFiles, loadBundledGuides, type LoadedGuide } from '../utils/file-loader';
 import { planGuideExecution, type ExecutionPlan } from '../e2e/guide-chains';
-import { type E2EErrorCode, type E2EExecutionOutcome } from '../e2e/schemas/e2e-report.schema';
+import { type E2EErrorCode, type E2EExecutionOutcome, type ExecutionSelection } from '../e2e/schemas/e2e-report.schema';
 import {
   contentDigest,
   createMinimalResultsData,
+  generateMultiGuideReport,
   generateReport,
+  writeMultiGuideReport,
   writeReport,
   type TestResultsData,
 } from '../e2e/e2e-reporter';
 import { checkTier, loadManifestFromDir, runManifestPreflight, type CurrentTier } from '../e2e/manifest-preflight';
+import {
+  loadLocalRepositorySource,
+  LocalMetapackageResolutionError,
+  resolveLocalMetapackage,
+  type LocalRepositorySource,
+} from '../e2e/e2e-local-package';
 import {
   resolveRemotePackage,
   resolveRemoteRepository,
@@ -59,7 +61,7 @@ import {
   printSummary,
   writeJsonReport,
 } from '../e2e/e2e-console-reporter';
-import type { ManifestJson, RepositoryEntry, RepositoryJson } from '../../types/package.types';
+import type { ManifestJson, RepositoryEntry } from '../../types/package.types';
 
 import { randomUUID } from 'crypto';
 import { createCloudAuthPolicy, type CloudAuthPolicy } from '../e2e/cloud-auth';
@@ -115,12 +117,6 @@ const DEFAULT_GRAFANA_URL = 'http://localhost:3000';
 const DEFAULT_RESOLVER_URL = 'https://recommender.grafana.com';
 const DEFAULT_CLOUD_URL = 'https://learn.grafana.net/';
 
-/** Repository source for dependency-aware planning (local index or remote CDN index). */
-interface RepoSource {
-  repository: RepositoryJson;
-  loadGuideById: (id: string, entry: RepositoryEntry) => LoadedGuide | null;
-}
-
 /**
  * Everything the run pipeline needs, resolved from CLI inputs regardless of
  * source (local files/bundled, a remote package, or the remote repository).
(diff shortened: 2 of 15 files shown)


leslie-heinzen:
@cursoragent review

leslie-heinzen:
Addressed all Cursor review findings in 242682f1:

**Blocking items resolved:**
- Added `ExecutionSelection` schema tests: absent case on non-metapackage runs, `type: 'journey'` validation, and invalid enum rejection
- Added JSDoc to `ExecutionSelectionSchema` documenting when `selection` is present vs absent and what `'path'`/`'journey'` mean; added inline doc on the field in `MultiGuideReportSchema`

**Suggestions addressed:**
- Added empty milestones edge-case test (`'has no milestones'` error)
- Added pure milestone-only cycle test (two path metapackages cycling without any `depends` edges)
- Added algorithm design rationale comment to `planPackageExecution` (leaf projection, ranking, two-phase resolution, milestone adjacency)
- Updated `CONTEXT_INDEX.md` `E2E_TESTING.md` one-liner to mention package-aware testing, milestone expansion, and report selection metadata

leslie-heinzen:
Addressed Bugbot finding in d433220a.

**Fix:** local path/journey expansion now calls `resolveTarget` per leaf rather than hardcoding `options.grafanaUrl`. Leaves whose `testEnvironment` does not match the current tier (e.g. cloud-tier milestones on a `--tier local` run) are pushed to `preRunSkipped` with a structured `skipped_tier_mismatch` result and excluded from the execution plan and guide list. Instance-specific and auth-gated milestones are handled by the same `resolveTarget` paths used by remote resolution.

A new test covers the concrete case: a path with a local and a cloud milestone on a local environment — the cloud milestone appears in `preRunSkipped` and is absent from `packageMetaById`.

leslie-heinzen:
Addressed both Bugbot findings in 2d6fc6e1.

**Finding 1 (high) — skipped milestones still executed:**
The unfiltered `executionPlan` was returned, so `runChains` still iterated over tier-skipped leaves and ran them against `options.grafanaUrl` because they had no `packageMetaById` entry. Fixed: build a `filteredExecutionPlan` that removes all skipped IDs from each chain and drops empty chains; this is what the command now executes.

**Finding 2 (medium) — skipped deps not cascaded:**
Hard dependents of a skipped leaf were not themselves skipped. Fixed: the single pass now checks each guide's `dependencies` before target resolution; any guide whose hard dep is already in `skippedIds` gets a `prerequisite_failed` pre-run skip and is excluded from the filtered plan.

New test (`cascades a pre-run skip to hard dependents of a skipped leaf`) covers: cloud-tier leaf skipped → hard dependent cascaded to `prerequisite_failed` → independent guide still runs → filtered execution plan contains only the independent guide.

src/cli/e2e/e2e-package.ts
```diff
@@ -290,8 +307,170 @@ function resolutionFailedSkip(target: ResolvedRemoteGuide, message: string): Ski
...
+          id,
+          reason: 'prerequisite_failed',
+          message: `Required guide(s) did not resolve: ${skipped.map((item) => item.id).join(', ')}`,
+          tier: rootTarget.tier,
+        },
+      ],
+      repository,
+      selection
+    );
+  }
```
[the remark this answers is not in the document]
leslie-heinzen: This is intentional and distinguishes pre-execution package eligibility from runtime dependency propagation.

An explicit remote `--package <path-or-journey>` selection is one package-health target. Before provisioning, every required milestone must resolve, validate, and route to the root-compatible target. `skipped_tier_mismatch` or `skipped_no_auth` means the selected environment cannot execute the requested package completely, so the root is reported non-runnable rather than producing partial path health.

The “only hard `depends` failures block downstream” rule applies once a complete runnable plan enters execution: an unrelated milestone failure does not suppress later milestones. It does not make an incomplete/unroutable remote package eligible for provisioning. This atomic boundary is also documented in the implementation plan and prevents a partially exercised path from appearing as a valid root execution.

src/cli/commands/e2e.ts
```diff
@@ -874,6 +916,133 @@ function remoteLoadGuideById(
...
+    return {
+      mode: 'local',
+      guides,
+      executionPlan: filteredExecutionPlan,
+      selection,
+      repoSource,
+      preRunSkipped,
+      packageMetaById,
+      localPackageDir: options.package,
+    };
```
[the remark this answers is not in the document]
leslie-heinzen: The premise here conflates the optional cover content with package-level metadata.

`content.json` is the optional path/journey cover page and is not executed. `manifest.json.testEnvironment` belongs to the selected metapackage itself: tier, minimum Grafana version, plugins, and instance are root-package requirements. Milestones do not inherit those values, but both constraints apply: the root package must be runnable and each leaf must independently be runnable.

Remote metapackage resolution enforces the same invariant by resolving `rootEntry.testEnvironment` before leaf expansion, then checking every leaf against the root-compatible target. Removing local root preflight would make local and remote semantics diverge and would ignore path-level plugin/version requirements.

If root preflight fails, reporting the selected root outcome rather than accumulated leaf skips is intentional: the package never became eligible to execute.

leslie-heinzen:
Refactored the local metapackage tests/implementation into the established E2E module pattern in 6d8ca7fa:

- Extracted local manifest/repository loading, planning/hydration, guide validation, target eligibility, and atomic root outcomes to `src/cli/e2e/e2e-local-package.ts`.
- Moved command-level tests to co-located `src/cli/e2e/e2e-local-package.test.ts`.
- Reduced `src/cli/commands/e2e.ts` to a private adapter that chooses the local resolver and feeds the common pipeline.
- Shared `loadLocalRepositorySource` from the E2E module for ordinary local dependency planning.

Behavior is unchanged from b980ab34. Validation: 88 focused tests pass, typecheck passes, all 152 governance tests pass, and `git diff --check` passes.

src/cli/e2e/e2e-local-package.ts
```diff
@@ -0,0 +1,229 @@
...
+
+    const packageMetaById = new Map<string, PackageMeta>();
+    const preRunSkipped: GuideRunResult[] = [];
+    for (const planned of executionPlan.chains.flat()) {
+      const entry = repoSource.repository[planned.id];
+      const target = resolveTarget(entry?.testEnvironment ?? {}, {
+        grafanaUrl: options.grafanaUrl,
+        currentTier: options.currentTier,
+        cloudUrl: options.cloudUrl,
+      });
```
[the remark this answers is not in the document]
leslie-heinzen: The observed limitation is real, but the comparison to local single-guide cloud execution is not accurate and the suggested partial fix would be unsafe.

`resolveRunInputs` returns for `mode === "local"` before constructing `CloudAuthPolicy` or `CloudStackPoolManagerConfig`. Consequently, `--cloud-instance-admin-token` and pool-manager flags are currently not wired for any local directory package, including single-guide packages. A local cloud guide may only work when `--grafana-url` already targets an independently authenticated environment; it does not use the CLI cloud provisioning path.

Passing capabilities into `resolveLocalMetapackage` alone would mark leaves runnable without supplying the provisioned runner token, cloud cleanup lifecycle, or correctly targeted manifest preflight. Supporting unpublished local cloud paths requires an end-to-end change: build cloud policy/config before the mode split, thread them through local resolution and `RunInputs`, provision targets, and run root preflight against the resolved target.

PR #1474 intentionally keeps the existing boundary: local directory packages are the local authoring/testing path; cloud metapackages use remote package-ID resolution plus the supported cloud/pool-manager flow. I’ll track local-directory cloud provisioning as separate follow-up work rather than add a partial auth/routing change here.

src/cli/e2e/guide-chains.ts
```diff
@@ -94,59 +103,99 @@ function clauseCandidates(clause: string | string[]): string[] {
...
     return true;
   };
 
-  // Resolve one clause to a concrete provider: prefer an alternative already in
-  // the run set (so chains reuse it), otherwise the first resolvable candidate.
-  // Returns null when nothing satisfies the clause.
   const resolveClauseProvider = (candidates: string[]): string | null => {
     for (const candidate of candidates) {
-      const reused = resolveCandidateProviders(candidate, repository, providesIndex).find((p) => runSet.has(p));
+      const reused = resolveCandidateProviders(candidate, repository, providesIndex).find((id) => packageSet.has(id));
```
Jayclifford345, line 320: [blocking] `correctness-and-reliability` / `guide-schema-and-contracts` — OR-group `depends` resolution isn't invariant to which other roots are batched into the same `planPackageExecution` call, so combining two independently-valid roots can manufacture a `depends` cycle that neither root produces alone.

**Why:** `packageSet` is seeded with *every* requested root id up front (line 183, before any `depends` resolution runs). `resolveClauseProvider` (here) then prefers whichever OR-group candidate is already present in `packageSet` — but for a union call that preference is contaminated by sibling roots that happen to be requested in the same batch, not just by the current root's own dependency closure.

**Confirmed by live execution** (3 independent repro fixtures across 3 skeptic sub-agents, all reproduced, none refuted) — e.g.:
```
pkgx:  depends: [['pkgp', 'pkgq']]
pkgp:  (no depends)
pkgq:  depends: [['pkgz', 'pkgx']]
pkgz:  (no depends)
pkgy2: depends: ['pkgq']   // forced
```
`planPackageExecution({rootIds:['pkgx']})` → `errors: []` (resolves pkgx→pkgp).
`planPackageExecution({rootIds:['pkgy2']})` → `errors: []` (resolves pkgy2→pkgq→pkgz).
`planPackageExecution({rootIds:['pkgx','pkgy2']})` → `errors: ['Cycle in depends chain: pkgx → pkgq → pkgx']` — the union of two individually-valid roots fails, order-independently.

This directly contradicts this file's own documented design intent ("Two-phase resolution... ensures provider choice is deterministic regardless of selection order" — the determinism only holds within a *fixed* root set, not across different unions of roots), and it's reachable in production: see the linked comment on `e2e-package.ts` for the blast radius. OR-group `depends` clauses referencing multiple providers of the same capability are a documented, intended authoring pattern (see `docs/design/package/dependencies.md`), not a contrived edge case — and the schema/authoring-validation layer has no cycle-safety check that would catch this at authoring time.

**Suggested action:** make OR-group provider selection invariant to sibling roots — e.g. resolve each root's own dependency closure independently first (mirroring the per-root validation pass callers already do) and carry that choice forward as a pin when planning the union, rather than letting "already in the ambient `packageSet`" pick up state seeded by unrelated roots. Add a `guide-chains.test.ts` case that plans two individually-valid roots separately and then together, asserting the union introduces no new errors — this scenario has no coverage today.

src/cli/e2e/e2e-package.ts
```diff
@@ -434,6 +677,71 @@ export async function resolveRemoteRepository(options: RemoteResolveOptions): Pr
...
+      id: guide.id,
+      reason: 'prerequisite_failed',
+      message: `Package planning failed: ${rootPlan.errors.join('; ')}`,
+      sourceUrl: guide.sourceUrl,
+      tier: guide.tier,
+      sideEffects: guide.sideEffects,
+    });
+  }
+  const plannable = runnable.filter((guide) => !structurallyUnavailable.has(guide.id));
+  const packagePlan = planPackageExecution({ rootIds: plannable.map((guide) => guide.id), repository });
```
Jayclifford345, line 696: [blocking] `reversibility-and-one-way-door` — this is the production blast radius of the planner bug flagged on `guide-chains.ts`'s `resolveClauseProvider`: when the aggregate `planPackageExecution` call over all `plannable` roots reports `errors.length > 0`, `resolveRemoteRepository` returns `{runnable: [], ...}` for the **entire** repository sweep (see the block below), discarding every guide — including ones that individually passed the per-root `structurallyUnavailable` check just above. Since the per-root check and the union call can now disagree (per the linked finding), a single incidental OR-group interaction between two unrelated, otherwise-healthy packages can silently zero out an entire `--remote` CI run with an opaque cycle error naming packages that look unrelated to whoever is debugging it. Once the planner fix lands, worth adding a regression test here specifically (not just in `guide-chains.test.ts`) asserting that two individually-valid roots surviving the per-root filter still produce a non-empty `runnable` set after the union plan.

src/cli/e2e/e2e-local-package.test.ts
```diff
@@ -0,0 +1,189 @@
...
+    });
+    consoleSpy = jest.spyOn(console, 'log').mockImplementation(() => undefined);
+  });
+
+  afterEach(() => {
+    consoleSpy.mockRestore();
+    rmSync(root, { recursive: true, force: true });
+  });
+
+  it('expands a valid path without requiring cover content.json', () => {
```
Jayclifford345, line 53: [suggestion] `testing-and-verification` — none of this file's 4 tests cover the PR's own "How to verify" step 1 scenario end-to-end: two (or more) `local`-tier milestones that **both** resolve successfully, asserting the final `guides` array / `executionPlan.chains` has the right length and preserves declared milestone order. This test (single milestone) and the two below it (which deliberately include a `cloud`-tier milestone to exercise the skip path) don't exercise `resolveLocalMetapackage`'s own `executionPlan.chains.flatMap(...)` aggregation with more than one *successful* guide — a regression there (dedup, reorder, off-by-one) would ship undetected. Worth adding a sibling test with 2+ local-tier milestones that all succeed.

src/cli/e2e/e2e-local-package.ts
```diff
@@ -0,0 +1,229 @@
...
+    );
+  }
+
+  try {
+    const repoSource = loadLocalRepositorySource(options.repositoryPath);
+    const rootEntry = repoSource.repository[manifest.id];
+    if (!rootEntry) {
+      throw new Error(`Root package "${manifest.id}" is missing from the repository index.`);
+    }
+    if (rootEntry.type !== manifest.type) {
```
Jayclifford345, line 144: [suggestion] `testing-and-verification` — this new manifest-vs-repository type-mismatch guard (and its remote counterparts — see the linked comments on `e2e-package.ts`) has no test coverage. Given this is exactly the kind of defensive branch this PR added to reconcile a manifest against a possibly-stale repository index for the new path/journey flow, a regression here (wrong message, dropped `selection`, wrong reason code) wouldn't be caught by CI. Suggest a test that sets up a local `path`/`journey` manifest whose `type` disagrees with its repository entry's `type` and asserts the thrown `LocalMetapackageResolutionError` message + attached `selection`.

src/cli/e2e/e2e-package.ts
```diff
@@ -290,8 +307,170 @@ function resolutionFailedSkip(target: ResolvedRemoteGuide, message: string): Ski
...
+      selection
+    );
+  }
+  if (rootEntry.type !== type) {
+    return skippedResolution(
+      [
+        {
+          id,
+          reason: 'resolution_failed',
+          message: `Root package type mismatch: recommender declared "${type}", repository declared "${rootEntry.type}"`,
```
Jayclifford345, line 344: [suggestion] `testing-and-verification` — three related type-reconciliation branches in this file have no test coverage (verified: no test constructs any of these scenarios):
1. Here — `resolveRemoteMetapackage`'s root type mismatch ("recommender declared X, repository declared Y").
2. Line ~594 — `resolveRemotePackage`'s `resolution.manifest.type` vs `indexedEntry.type` disagreement → `resolution_failed`.
3. Line ~537-544 — CDN index unreachable *and* the manifest declares a `path`/`journey` type → the metapackage-specific "...milestones cannot be resolved" skip message (the existing "index unreachable" tests only use `type: 'guide'` manifests, so this branch never runs).

Same concern as the sibling comment on `e2e-local-package.ts`: these are new defensive branches for the exact contract the PR calls out ("pre-run failures retain the selected root"), and a regression in message/reason/`selection` wouldn't be caught today.

src/cli/commands/e2e.ts
```diff
@@ -418,94 +429,81 @@ function resolveValidGuides(files: string[], options: E2ECommandOptions): Loaded
...
-    loadGuideById = (id: string, entry: RepositoryEntry): LoadedGuide | null => {
-      const rel = entry.path || `${id}/`;
-      const contentPath = rel.endsWith('.json') ? join(repoBaseDir, rel) : join(repoBaseDir, rel, 'content.json');
-      return loadGuideFiles([contentPath])[0] ?? null;
-    };
-  }
-
-  const plan = planGuideExecution({ guides: valid, repository, loadGuideById });
-
+function finalizeExecutionPlan(
```
Jayclifford345, line 432: [nit] tech-debt (non-blocking) — `finalizeExecutionPlan`'s auto-included banner + verbose chain-printing block here is near-duplicated by `printPlan` in `src/cli/e2e/e2e-local-package.ts:102-115` (identical "print auto-included count, then in verbose mode print `🔗 Execution plan: N chain(s)` + one line per chain" logic), and this function's guide-validation block duplicates the shape of `e2e-local-package.ts`'s `validatePlannedGuides` (validate → collect errors → format message), differing only in how the two callers report the failure (`failCommand` vs throwing `LocalMetapackageResolutionError`). Since `e2e.ts` is the hottest file in this subsystem, worth extracting a shared `printExecutionPlan(plan, verbose)` / `validateGuides(guides, verbose)` pair (e.g. into `guide-chains.ts` or a small shared module) that both call sites use, rather than maintaining two copies of the same console output format.

docs/developer/E2E_TESTING.md
```diff
@@ -17,6 +17,7 @@ This is the canonical implementation-backed reference for E2E CLI behavior. Veri
 - `src/cli/commands/e2e.ts` — Commander options, input resolution, dependency planning, pre-flight orchestration, clean-stack resets, cloud routing, and per-guide Playwright invocation.
 - `src/cli/e2e/e2e-runner-contract.ts` — environment-variable contract between the CLI process and Playwright runner.
 - `src/cli/e2e/e2e-package.ts` — remote package and repository resolution, content fetch, schema validation, side-effect classification, and pre-run skip reasons.
+- `src/cli/e2e/guide-chains.ts` — pure package graph planning across hard dependencies, capabilities, and recursive milestones, followed by leaf-guide hydration.
```
Jayclifford345, line 21: [nit] doc-drift (non-blocking) — this PR added the `guide-chains.ts` bullet here but left out the new `src/cli/e2e/e2e-local-package.ts` (229 lines), which owns local path/journey manifest resolution, milestone planning, and pre-run-skip construction — the local-directory counterpart to `e2e-package.ts` just above. Since this doc explicitly bills itself as "the canonical implementation-backed reference," worth adding a line for it here too.

src/cli/commands/e2e.ts
```diff
@@ -976,7 +1006,7 @@ export const e2eCommand = new Command('e2e')
   )
   .option(
     '--repository <path>',
```
Jayclifford345, line 1008: [nit] the `(default: bundled index)` parenthetical here is misleading for the new local path/journey flow: `resolveLocalMetapackage` throws immediately if `--repository` is omitted for a local `path`/`journey` `--package` (no bundled-index fallback applies there), so a user reading `--help` could reasonably expect `--package ./my-path/` to work without `--repository` and then hit a `CONFIGURATION_ERROR` at runtime. The runtime behavior and its docs in `E2E_TESTING.md` are correct — just the CLI help text over-promises. Maybe: `'... (default: bundled index; required for a local path/journey --package)'`.

Jayclifford345 (changes requested):
Principal-engineer-level review via the repo's `/review` orchestrator: routed against `docs/design/CONCERNS.md` (classification: `mixed` — new CLI product behavior + contract/schema addition + heavy test changes), ran always-on concerns (`security`, `correctness-and-reliability`, `testing-and-verification`, `reversibility-and-one-way-door`, `cross-cutting-architecture`) plus the activated subsystem concerns (`cli-and-e2e-runner`, `guide-schema-and-contracts`) as independent reviewers, then ran a 3-skeptic adversarial-verification pass on every medium-or-higher finding before reporting, plus a tech-debt scan and a doc-drift check scoped to the changed files.

**Headline finding (blocking):** the milestone/dependency planner in `guide-chains.ts` isn't invariant to which other roots are requested in the same call — batching two individually-valid roots together can manufacture a `depends` cycle that neither root produces alone, and the repository-wide caller in `e2e-package.ts` then discards the *entire* batch (including guides that individually passed validation) when that happens. This directly undercuts the PR's own headline claim ("Repository-wide runs also keep independent guides runnable when another root has an unavailable or structurally invalid prerequisite"). Confirmed by live execution against the actual PR code with 3 independently-constructed repository fixtures (3/3 skeptics reproduced it, not refuted) — see inline comment for the concrete repro.

A few test-coverage gaps around the new type-mismatch/reconciliation branches and the local multi-milestone success path are flagged as non-blocking but worth closing given this is new, security-adjacent-ish CLI infrastructure. One initially-suspected security finding (an unguarded `new URL()` in the new metapackage target-compatibility check) was investigated and **dropped** after adversarial verification: 2 of 3 skeptics (and my own independent trace) showed the risky comparison only executes when tier+instance already match between guide and root, which forces the two `targetUrl` strings to be byte-identical — and the guide's copy has already survived an equivalent `new URL()` parse earlier in `buildGuideOrSkip`. So the crash is provably unreachable today; not reporting it as a finding, though hardening it defensively wouldn't hurt given how implicit that invariant is.

**Tech debt** (non-blocking): `e2e.ts`'s `finalizeExecutionPlan` and `e2e-local-package.ts`'s `validatePlannedGuides`/`printPlan` implement near-identical plan-printing and guide-validation logic — worth extracting into a shared helper (see inline comment).

**Doc drift** (non-blocking): `E2E_TESTING.md`'s "Source map for agents" list was updated in this PR to add `guide-chains.ts` but omits the new `e2e-local-package.ts` (see inline comment).

Everything else (schema additivity/backward-compat, content-identity checks on fetched remote content, metapackage leaf trust boundary, Node-safe imports, exit-code/skip-vs-fail semantics, preflight ordering) reviewed clean — no findings.

src/cli/e2e/guide-chains.ts
leslie-heinzen: Fixed in cf46a4d1. Multi-root planning now resolves each root closure independently, preserving forced-before-OR reuse within a root, then deterministically merges/deduplicates the leaf plans. The exact pkgx/pkgp/pkgq/pkgz/pkgy2 fixture now asserts both roots pass separately and together, with pkgx pinned to pkgp and pkgq pinned to pkgz.

src/cli/e2e/e2e-package.ts
leslie-heinzen: Added the resolver-level regression in cf46a4d1 using the same adversarial provider graph. resolveRemoteRepository now produces no aggregate error, retains all five runnable guides, and returns the expected non-empty execution plan.

src/cli/e2e/e2e-local-package.test.ts
leslie-heinzen: Added a local all-success multi-milestone test in cf46a4d1. It verifies declared guide order, execution-plan order, package metadata aggregation, and an empty pre-run skip set.

src/cli/e2e/e2e-local-package.ts
leslie-heinzen: Added local manifest/repository root type-mismatch coverage in cf46a4d1, including the specific mismatch message and attached path selection.

src/cli/e2e/e2e-package.ts
leslie-heinzen: Covered all three branches in cf46a4d1: metapackage root type mismatch with its specific message/selection, generic recommender-vs-index type disagreement with repository metapackage selection, and unreachable-index path resolution retaining selection with the milestones-specific message. I also routed metapackage manifest disagreements through resolveRemoteMetapackage so its dedicated root mismatch guard is reachable.

src/cli/commands/e2e.ts
leslie-heinzen: Agreed this duplication is worth consolidating. I left it unchanged in this correctness-focused update to avoid broadening the patch into command-output refactoring; the shared validation/printing extraction can be handled as follow-up tech debt.

docs/developer/E2E_TESTING.md
leslie-heinzen: Updated the E2E source map in cf46a4d1 to include e2e-local-package.ts and its local metapackage responsibilities.

src/cli/commands/e2e.ts
leslie-heinzen: Clarified the --repository help text in cf46a4d1: the bundled default applies to guide dependency lookup, while local path/journey milestone resolution requires an explicit repository path.

src/cli/e2e/e2e-local-package.ts
```diff
@@ -0,0 +1,229 @@
...
+        packageId: planned.id,
+        tier: target.tier,
+        instance: target.instance,
+        targetUrl: target.targetUrl!,
+        sourceUrl: planned.guide.path,
+        startingLocation: entry?.startingLocation ?? '/',
+        sideEffects: classifyGuideSideEffectsFromString(planned.guide.content),
+        ...(entry?.testEnvironment?.plugins?.length ? { plugins: entry.testEnvironment.plugins } : {}),
+      });
+    }
```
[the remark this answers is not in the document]
leslie-heinzen: Fixed in 5c4e3a91. Local metapackage resolution now resolves and gates the selected root target before planning, preserves the root selection and target skip reason when it is ineligible, and rejects an otherwise-runnable leaf when tier, instance, or target origin differs from the root. Added regressions for a cloud root in a local environment and for a runnable root/leaf instance mismatch.

leslie-heinzen:
Reviewed the TruffleHog scan comment: both unverified URI detections are deliberate literals in `src/lib/telemetry/url.test.ts` (URL normalization/userinfo test fixtures). GitHub’s PR files API confirms that file is not modified by #1474, so I did not add unrelated `trufflehog:ignore` changes to this PR.

Jayclifford345 (approved):
All issues resolved and LGTM.

PR #1477, opened 2026-07-30
fix(e2e-runner): pass with 0 steps for markdown-only guides

leslie-heinzen:
## Summary

Guides with no interactive blocks (markdown, image, video only) previously caused the runner to wait 15 s for `[data-testid^="interactive-step-"]` elements that never appear, then exit without writing a results file. The CLI reported `infrastructure_error / REPORT_MISSING`, which cascaded into `SKIPPED_PREREQ` for every subsequent milestone with a hard dependency.

53% of the interactive-tutorials guide population (280 of 529) is markdown-only. This affected all learning path runs that include bookend milestones (business-value intros, end-journey wrap-ups).

## What to look at

- [`tests/e2e-runner/utils/guide-runner/static-analysis.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/d7eef6b0cab0a81e4b6a11e9cd7ed5f5e3c2e28e/tests/e2e-runner/utils/guide-runner/static-analysis.ts) — `countInteractiveBlocks()` walks the full nested block tree (section, conditional both branches, assistant) before Playwright starts.
- [`tests/e2e-runner/guide-runner.spec.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/d7eef6b0cab0a81e4b6a11e9cd7ed5f5e3c2e28e/tests/e2e-runner/guide-runner.spec.ts) — immediately after reading the guide JSON, check the count; if zero, write a passed result with 0 steps and return without touching the DOM wait or step-discovery path. Pre-flight checks (auth, plugin) still run.

## Why

The runner was designed for guides with at least one interactive step. Learning paths expose a large class of read-only milestone pages that are correct content but have no actionable steps. Failing them as infrastructure errors and blocking all downstream milestones is wrong; 0 steps passed is the correct outcome.

## How to verify

Run the CLI against any markdown-only milestone in a local learning path:

```bash
node dist/cli/cli/index.js e2e \
  --package /path/to/interactive-tutorials/alloy-configuration-basics-lj/configuration-file-anatomy \
  --tier local --output /tmp/result.json
```

Confirm the result is `outcome: "passed"` with `summary.total = 0` rather than `infrastructure_error / REPORT_MISSING`.

## Breaking changes

None. The fix is additive: guides with interactive steps follow the existing code path unchanged.

### DIFF
--- a/tests/e2e-runner/utils/guide-runner/static-analysis.ts
+++ b/tests/e2e-runner/utils/guide-runner/static-analysis.ts
@@ -1,22 +1,7 @@
-const INTERACTIVE_TYPES = new Set([
-  'interactive',
-  'multistep',
-  'guided',
-  'code-block',
-  'terminal',
-  'terminal-connect',
-  'grot-guide',
-]);
+const INTERACTIVE_TYPES = new Set(['interactive', 'multistep', 'guided']);
 
-/**
- * Count interactive blocks in a guide recursively.
- *
- * Walks the full nested block tree, including section, assistant, and
- * conditional containers (both whenTrue and whenFalse branches). The count
- * lets the runner skip the DOM wait entirely for markdown-only guides rather
- * than timing out after 15 s waiting for elements that never appear.
- */
-export function countInteractiveBlocks(guide: unknown): number {
+/** Count blocks discoverable through the runner's `interactive-step-*` selector. */
+export function countDiscoverableSteps(guide: unknown): number {
   let count = 0;
 
   function walk(node: unknown): void {
--- a/tests/e2e-runner/utils/guide-runner/static-analysis.ts
+++ b/tests/e2e-runner/utils/guide-runner/static-analysis.ts
@@ -1,7 +1,6 @@
-const INTERACTIVE_TYPES = new Set(['interactive', 'multistep', 'guided']);
+import { isInteractiveBlockType } from '../../../../src/constants/json-guide-classification';
 
-/** Count blocks discoverable through the runner's `interactive-step-*` selector. */
-export function countDiscoverableSteps(guide: unknown): number {
+export function countInteractiveBlocks(guide: unknown): number {
   let count = 0;
 
   function walk(node: unknown): void {
@@ -18,7 +17,7 @@ export function countDiscoverableSteps(guide: unknown): number {
 
     const block = node as Record<string, unknown>;
 
-    if (typeof block.type === 'string' && INTERACTIVE_TYPES.has(block.type)) {
+    if (isInteractiveBlockType(block.type) || block.type === 'snippet-ref') {
       count++;
     }
 
--- a/tests/e2e-runner/utils/guide-runner/static-analysis.ts
+++ b/tests/e2e-runner/utils/guide-runner/static-analysis.ts
@@ -1,0 +1,47 @@
+const INTERACTIVE_TYPES = new Set([
+  'interactive',
+  'multistep',
+  'guided',
+  'code-block',
+  'terminal',
+  'terminal-connect',
+  'grot-guide',
+]);
+
+/**
+ * Count interactive blocks in a guide recursively.
+ *
+ * Walks the full nested block tree, including section, assistant, and
+ * conditional containers (both whenTrue and whenFalse branches). The count
+ * lets the runner skip the DOM wait entirely for markdown-only guides rather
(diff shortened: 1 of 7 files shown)


leslie-heinzen:
@cursoragent review

leslie-heinzen:
Addressed Cursor review feedback in d390f579.

**Selector alignment:** renamed the helper to `countDiscoverableSteps` and restricted its set to `interactive`, `multistep`, and `guided` — the only block types that render the `interactive-step-*` test ID contract consumed by `discoverStepsFromDOM`. Guides containing only `code-block`, `terminal`, `terminal-connect`, or `grot-guide` now take the zero-discoverable-step path instead of waiting for selectors they never render.

**Expanded tests:**
- all included block types (interactive, multistep, guided)
- excluded/unsupported block types (code-block, terminal, terminal-connect, grot-guide)
- conditional true/false branches
- assistant wrappers
- deeply nested mixed content
- null, undefined, malformed, and empty guides

Validation: 100/100 adjacent runner tests pass, typecheck passes, Prettier passes.

tests/e2e-runner/utils/guide-runner/static-analysis.ts
```diff
@@ -0,0 +1,32 @@
+const INTERACTIVE_TYPES = new Set(['interactive', 'multistep', 'guided']);
```
moxious: The PR is sound (and I knew you'd eventually run into this one, glad you found it) but this line here is the key to the matter.... this set is incorrect, incomplete, and duplicative of what we've got elsewhere in the code.

Send an agent to read this:  requiresGrafanaUi (src/components/docs-panel/utils/requires-grafana-ui.ts:48) is the app's mature block-type classifier

Ideal fix: 

hoist definitions of INTERACTIVE_TYPES in a spot like `src/constants` where it can be reused in both places, rewire requiresGrafanaUi to work with that constant, and just import the constant in static-analysis.ts that you need right here.   It would need to go in src/constants so e2e is still safe and avoiding any react imports.

Acceptable fix:  study requiresGrafanaUi to derive the right list (this one isn't it as authored) and put an explicit comment note back to the other location that this set is sensitive to additions/deletions of interactive blocks over time.

**Please note** -- the wild & wooly extra options in `requiresGrafanaUi` is also likely to be a treasure map to unhandled edge cases in e2e code, so it's very worthy of attention.

moxious (changes requested):

tests/e2e-runner/utils/guide-runner/static-analysis.ts
leslie-heinzen: Thanks david - cursoragent caused some confusion for me here with the above review. It suggested _reducing_ the entries in INTERACTIVE_TYPES to just those three, because they are currently the only ones that are supported by the E2E playwright guide runner's [selector discovery logic](https://github.com/grafana/grafana-pathfinder-app/blob/main/tests/e2e-runner/utils/guide-runner/discovery.ts#L53) (the const referenced there will match only interactive, multistep, and guided steps). Originally I had the bigger list of all interactive types.

Its suggestion isn't really addressing the problem at hand, just kind of partially masking a separate gap in functionality. If a guide has a non-skippable quiz, for example, AT ALL in it, the guide will fail currently anyway. Obviously, we should be able to handle all of the interactive elements in a guide, but that's not related to what this PR is trying to fix.

I like the ideal solution you mentioned, so I'm going to roll with that, and disregard cursoragent's proposal.

leslie-heinzen: **(Agent)** Addressed in 2fbb433c. The complete Pathfinder interactive block type set is now hoisted to the Node-safe `src/constants/json-guide-classification.ts` and reused by both `requiresGrafanaUi` and E2E static analysis.

The zero-step decision counts all ten interactive block types, including types the runner does not execute yet, so unsupported types continue exposing runner gaps rather than being mislabeled as read-only. A surviving `snippet-ref` also counts fail-safe as potentially interactive.

`requiresGrafanaUi` keeps its existing behavior inside the shared family: action-dependent blocks inspect their actions, `code-block` always requires Grafana, and self-contained quiz/input/terminal/challenge/grot blocks return false. Added exact membership and consumer regression tests.

Validation: 128 focused tests, 152 governance tests, typecheck, Prettier, and `git diff --check` pass.

moxious (approved):

PR #1482, opened 2026-07-31
feat(telemetry): add masked Faro session replay

Jayclifford345:
## Summary

Adds [Grafana Cloud session replay](https://grafana.com/docs/grafana-cloud/observe-and-act/monitor-applications/frontend-observability/session-replay/) to Pathfinder's existing Faro instance, so a guide that goes wrong can be watched back instead of reconstructed from events — which element the user clicked instead of the highlighted one, whether the sidebar was covering the target, where they gave up.

Recording is fully masked: every text node and every input type, with canvas, fonts, inline images, inline stylesheets and cross-origin iframes all off. On top of that, `replay-scrub.ts` closes the gap the SDK leaves open.

## What to look at

**`src/lib/telemetry/replay-scrub.ts`** — the privacy-load-bearing file. rrweb masks text but never *attributes*, and Grafana puts real content in them (`data-testid="Panel header <panel title>"`, `aria-label`, `title`, `alt`, `placeholder`); dashboard URLs carry `var-*` template variables. The scrubber walks meta hrefs, full-snapshot node trees and mutation `adds`/`attributes`, keeping a structural allowlist and routing URL attributes through the new `stripUrlSecrets`. Allowlist rather than denylist because with all text already masked, an unrecognised attribute is pure downside — adding to `SAFE_ATTRIBUTES` means asserting the attribute cannot carry user-authored text.

**`ContinuousReplayInstrumentation` in `replay.ts`** — a workaround, and the least obvious thing here. `faro-core`'s `setSession()` calls `metas.remove(metaSession)` before `metas.add(...)`, and both halves notify metas listeners, so every `setFaroSessionAttributes` call publishes one notification with no session at all. `ReplayInstrumentation` re-checks recording on each metas change, reads `sessionId === null` as "not sampled", tears rrweb down, then restarts on the next notification — emitting a fresh full-DOM snapshot. Because Pathfinder stamps `surface` on every sidebar open and close, this fragmented recordings into a clip per toggle. Confirmed against production data before fixing (four `faro.session_recording.started` in one session, one per surface transition). `setView` in the same file already adds-before-removing to avoid exactly this gap; **this belongs upstream in `faro-web-sdk`** and the subclass should be deleted once `setSession` matches.

**Activation timing** — the recorder is added via `faro.instrumentations.add()` on first Pathfinder open, not in the `instrumentations` array. Starting at page load would put rrweb's opening snapshot on the wrong side of `passesActivityGate`, leaving mutations with nothing to apply them to.

## Why

Pathfinder's telemetry can say *that* a step failed but not *what the user saw*. Everything else here follows from wanting that without recording anything we can't defend.

## Cloud-only, and default-on

No new environment gate: replay is registered inside `initFaro`, after the `resolveFaroEnvironment()` early return, so a self-hosted or OSS Grafana never reaches it — it doesn't construct a Faro instance at all, and the rrweb chunk is never fetched. Three cases in `faro.test.ts` pin that (OSS, Enterprise, analytics opted out).

`pathfinder.session-replay` defaults to `true` in both the registry and the call site, matching `pathfinder.frontend-telemetry` — it is a kill switch, not a rollout gate. Two consequences called out in the docs:

1. Grafana core ships its own recorder behind `FlagKeys.FaroSessionReplay`. Two rrweb instances on one page means double DOM serialisation per mutation, and rrweb proxies `CSSStyleSheet.prototype.insertRule` globally — Emotion's hot path in core. **Set this flag false wherever core's goes true.**
2. Playback needs the private-preview feature enabled on the stack. Already on for the ops stack Pathfinder reports to.

## How to verify

`npm run check` — 386 suites, 6128 tests.

Bundle split, against a real production build: `module.js` has zero rrweb, the Faro SDK chunk (97 KB) has zero rrweb, and rrweb lands alone in a 194 KB chunk fetched only when the flag is on *and* Pathfinder opens.

Locally: dev-mode Grafana, `localStorage.setItem('pathfinder.faro.local','true')`, reload. With Pathfinder closed no replay chunk loads; open the sidebar and `faro.session_recording.started` posts. Inspect a `faro.session_recording.event` payload — text should be asterisks, no `aria-label`/`data-testid`/`title`/`alt`/`placeholder` keys anywhere, hrefs free of query strings, `#icon-…` refs intact.

Three test files carry the load: `replay-scrub.test.ts` (12 cases), `replay-lifecycle.test.ts` (real SDK + real rrweb, asserts one `started` across three surface stamps — fails without the subclass), and additions to `faro-sdk-contract.test.ts` pinning session-id stability.

## Breaking changes

None. `initFaro()` gains an optional options argument.

## Reversibility

Set `pathfinder.session-replay` false in MTFF and no further sessions start recording, without a release. Reverting the PR removes the dependency and the chunk.

Two limits on that, both now written down in `TELEMETRY.md` under [Stopping a recording](docs/developer/TELEMETRY.md#stopping-a-recording) rather than left implied:

- **Both flags are bootstrap snapshots.** A flip reaches a tab only on its next page load — a tab already recording carries on until it is reloaded or closed. Re-evaluating mid-session would mean polling OFREP or tearing rrweb down live, and a torn-down recorder leaves a mutation stream with no snapshot to apply it to.
- **Recordings already collected are not undone by the flip.** Removing them is a collector-side deletion request against Frontend Observability, not something a flag or a release can do.

## Known limitations

- Two browser tabs are two sessions and therefore two recordings. Inherent to `sessionTracking: { persistent: false }`, which exists to avoid colliding with core Grafana's session key.
- The 60s inactivity threshold pauses recording and resumes on the next interaction, producing legitimate gaps within a session.
- Dropping unrecognised attributes can cause minor visual drift in playback for any UI whose CSS keys off an attribute not on the allowlist. Masking `content` and custom-property strings adds a second, smaller source of the same: a theme that puts a quoted string in a custom property plays back with that value asterisked. Escape sequences are preserved, so icon glyphs still render.
- `SessionInstrumentation`'s beforeSend hook deep-clones every sampled transport item via `JSON.parse(JSON.stringify(item))`. That was cheap for typed events; replay payloads are large strings at high volume. Worth watching before this goes wide.


Jayclifford345:
@cursoragent review

Jayclifford345:
Thanks — three of the four addressed in 4aea99c9, one disputed.

### Fixed

**Medium: URLs in CSS bypass sanitization.** Real, and wider than described. CSS reaches the collector by *two* routes, and neither was covered:

- the `style` and `_cssText` attributes, both on the allowlist and passing through verbatim;
- the text inside `<style>`, which rrweb deliberately leaves unmasked (`serializeTextNode` skips masking when the parent is `STYLE` — masking it would strip the replay of all styling) and which carries a serialized stylesheet whenever the element has a live sheet, **regardless of `inlineStylesheet`**.

rrweb also runs `absolutifyURLs` over stylesheet text before we see it, so a relative `url(logo.png?sig=…)` arrives fully qualified with the query intact. Both routes now go through `scrubCssUrls`, which rewrites each `url()` target via `stripUrlSecrets` while leaving in-document `url(#filter)` references alone. `style`/`_cssText` moved out of `SAFE_ATTRIBUTES` into their own `CSS_ATTRIBUTES` set so there is one home per attribute.

**Low: exotic URL schemes.** `stripUrlSecrets` now keeps only `http:`/`https:`. `javascript:`, `vbscript:`, `blob:` and `mailto:` collapse to `''` — none of them mean anything on playback, so there is no rendering cost to refusing to round-trip them.

**Low: terminal blocking is fragile.** Agreed, and the fix is better than pinning the id: the selector now also matches `.xterm`, xterm's own root class. Renaming our test id can no longer silently unblock the terminal, and it stays two cheap selectors — `isBlocked` runs `matches()`/`closest()` per mutated node.

Five new scrubber cases and four new `stripUrlSecrets` cases cover these.

### Disputed

**Critical: R1 subscription leak.** I don't think this one holds.

`startSessionReplayOnFirstOpen` unsubscribes the moment recording starts. The only case where it doesn't is "Pathfinder is never opened for the life of the page", which leaves exactly one closure in a module-level `Set` that lives as long as the document.

The suggested remediation — return the unsubscribe so `initFaro` can guarantee cleanup — has nowhere to land. `initFaro` returns `Promise<void>`, is fire-and-forget from `module.tsx`, and has no teardown path; there is no "module shuts down" event here. The `stampSurface` listener registered two lines above it is likewise never unsubscribed, and has been since that code was written. Adding a returned disposable that nobody can call would be ceremony rather than a fix.

Happy to be wrong if there's a teardown path I've missed — but a real fix would mean giving the whole telemetry adapter a lifecycle, which is a separate change.

### Not addressed

TruffleHog's two hits are false positives by construction: `url.test.ts` lines 17 and 48 are the assertions that userinfo (`https://user:password@…`) gets stripped. Line 17 predates this PR.

`npm run check` green — 386 suites, 6137 tests.

src/lib/telemetry/replay-scrub.ts
```diff
@@ -0,0 +1,274 @@
...
+      }
+    }
+    if (Array.isArray(data.attributes)) {
+      for (const mutation of data.attributes) {
+        if (isRecord(mutation) && isRecord(mutation.attributes)) {
+          scrubAttributes(mutation.attributes);
+        }
+      }
+    }
+  }
```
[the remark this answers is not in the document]
Jayclifford345: Confirmed and fixed in d624f9a4 — and it was broader than the two sources named.

Checked every incremental source against `@grafana/rrweb-types` for CSS-bearing payloads. Five routes were unscrubbed:

- **`StyleSheetRule` (8)** — `adds[].rule`, which is Emotion's `insertRule` hot path in Grafana, plus `replace` / `replaceSync` for wholesale sheet replacement.
- **`AdoptedStyleSheet` (15)** — `styles[].rules[].rule`.
- **`StyleDeclaration` (13)** — `set.value`, a single inline property write.
- **mutation `texts`** — how `<style>` content changes arrive. Everything else in `texts` is masked to asterisks, so the scrub only ever bites on CSS and is a no-op elsewhere.
- **the `style` attribute in an attribute mutation** — this one wasn't in the report and is easy to miss: a mutation carries `style` as rrweb's `styleOMValue` diff, a property map whose values are either a string or a `[value, priority]` pair, not the flat string a snapshot carries. My previous fix only handled `typeof value === 'string'`, so the diff form fell straight through.

Also handled `Font` (10): `collectFonts` is off so it shouldn't arrive, but `fontSource` is a bare URL rather than CSS — `scrubCssUrls` would have passed it through untouched since there's no `url(` wrapper, so it takes `stripUrlSecrets` directly. That default isn't this module's to rely on.

The dispatch is now a `switch` on `data.source`, which also fixes a latent trap: `adds` means `addedNodeMutation[]` under `Mutation` and `styleSheetAddRule[]` under `StyleSheetRule`, so the shapes have to be keyed off the source rather than the key name.

On the hot path — `scrubCssUrls` short-circuits on any string without `url(`, so Emotion's rule traffic costs one `includes` rather than a regex.

Seven new cases, one per route. `npm run check` green: 386 suites, 6144 tests.

src/lib/telemetry/replay-scrub.ts
```diff
@@ -0,0 +1,368 @@
...
+const CSS_URL_PATTERN = /url\(\s*(['"]?)([^'")]*)\1\s*\)/gi;
+
+function scrubCssUrls(css: string): string {
+  // Emotion rewrites rules constantly, so the common case has to stay cheap.
+  if (!css.includes('url(')) {
+    return css;
+  }
+  return css.replace(CSS_URL_PATTERN, (match, _quote, target: string) =>
+    target.startsWith('#') ? match : `url("${stripUrlSecrets(target)}")`
+  );
```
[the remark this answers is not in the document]
Jayclifford345: Correct — fixed in aa22a950.

`[^'")]*` is the right class for a *bare* `url()`, where CSS forbids unescaped parens, quotes and whitespace. It's wrong for a quoted one, where parens are legal: `url("a(1).png?sig=…")` and quoted `data:` SVGs both terminated the capture early, `\1` then failed to find its closing quote, and the whole match was skipped — so the value passed through untouched.

Now three arms rather than one class:

```
/url\(\s*(?:"([^"]*)"|'([^']*)'|([^'")\s]*))\s*\)/gi
```

A quoted value runs to its closing quote; a bare one keeps the stricter class, which is what stops it running past the end of the declaration. No nested quantifiers, so it stays linear.

Five new cases: double- and single-quoted with parens, whitespace padding inside `url( … )`, a quoted `data:` SVG containing `)` (collapses to `data:`), and multiple `url()`s in one rule to pin that the global flag still advances correctly past the longer matches.

`npm run check` green: 386 suites, 6149 tests.

src/lib/telemetry/url.ts
```diff
@@ -19,3 +19,37 @@ export function normalizeTelemetryUrl(url: string): string {
...
+    // playback and all of which are better not round-tripped.
+    if (parsed.protocol !== 'http:' && parsed.protocol !== 'https:') {
+      return '';
+    }
+    parsed.search = '';
+    parsed.hash = '';
+    parsed.username = '';
+    parsed.password = '';
+    const isAbsolute = /^[a-z][a-z0-9+.-]*:/i.test(url) || url.startsWith('//');
+    return (isAbsolute ? parsed.href : `${parsed.pathname}`).slice(0, MAX_TELEMETRY_URL_LENGTH);
```
moxious, line 64: **[Medium · privacy] Pathname retains user-authored titles — the one place "all text is masked" provably fails.**

Keeping the pathname means Grafana's title-bearing routes leak: `/d/<uid>/<slug>` and `/dashboards/f/<uid>/<slug>` carry the slugified dashboard/folder title. The on-screen title is masked to asterisks everywhere, but the same title rides through here — in the replay `Meta` event `href` and every `href`/`src` attribute — so a board titled "Acme Corp Q3 Revenue Confidential" becomes `/d/xxx/acme-corp-q3-revenue-confidential` on essentially every recording.

Bounded to Medium because Faro is Grafana-internal, but it's customer-derived text crossing the boundary systematically, and it contradicts the PR's headline guarantee. Suggest an explicit decision: either document it as a known residual in the privacy notes (the slug arguably aids replay navigability), or redact the title segment (keep `uid`, drop `slug`). It shouldn't be silent.

src/lib/telemetry/replay-lifecycle.test.ts
```diff
@@ -0,0 +1,71 @@
...
+// fresh full-DOM snapshot and a new clip instead of one continuous recording.
+describe('session replay lifecycle across session-attribute stamps', () => {
+  const transport = new CaptureTransport();
+  let faro: Faro;
+
+  const recordingEvents = () =>
+    transport.items
+      .filter((item) => item.type === 'event')
+      .map((item) => (item.payload as { name?: string }).name ?? '')
+      .filter((name) => name.startsWith('faro.session_recording.') && name !== 'faro.session_recording.event');
```
moxious, line 32: **[Medium · testing] The masking guarantee is never asserted against real rrweb output.**

This is the only test driving the real SDK + real rrweb, and it deliberately filters *out* the `faro.session_recording.event` payloads here — so the recorded DOM is never inspected. Everything else is structural: `faro.test.ts` asserts the *options* are set (`maskTextSelector === '*'`, `maskAllInputs`), and `replay-scrub.test.ts` runs the scrubber on hand-built synthetic events. Nothing verifies those options actually yield masked output.

For a default-on, privacy-load-bearing recorder that's the exact regression the suite should catch: a future `@grafana/faro-instrumentation-replay`/rrweb bump that renamed an option or changed `maskTextSelector:'*'` semantics would pass every test and silently ship an unmasked recording.

Cheap to close — `CaptureTransport` already receives the payloads. Render a node with a known secret + an `<input>` with a known value, take the real snapshot, then assert the captured `faro.session_recording.event` contains asterisks / no plaintext. One test covers text masking, input masking, and the `beforeSend → scrubReplayEvent` wiring end-to-end.

src/lib/telemetry/replay.ts
```diff
@@ -0,0 +1,98 @@
...
+  maskAllInputs: true,
+  maskInputOptions: MASK_EVERY_INPUT,
+  maskTextSelector: '*',
+  blockSelector: BLOCK_SELECTOR,
+  recordCanvas: false,
+  inlineImages: false,
+  inlineStylesheet: false,
+  collectFonts: false,
+  recordCrossOriginIframes: false,
+  samplingRate: 1,
```
moxious: **[Medium · perf/volume] `samplingRate: 1` + default-on + full-page + indefinite, with only a binary control lever.**

This records 100% of Pathfinder-engaged sessions, capturing the *entire* Grafana page (not just the sidebar) for the rest of the tab's life after first open. The only volume control is the kill-switch flag — there's no percentage-sampling lever. Real mitigations exist (records engaged sessions only, 60s inactivity pause), but an SRE who opens Pathfinder once on a long-lived auto-refreshing dashboard ships a full multi-hour session at 100% — and auto-refresh means the inactivity pause rarely fires.

Suggest making `samplingRate` (and ideally a max-duration / idle-teardown) remotely configurable rather than hardcoded, so volume can be dialed during rollout without an all-or-nothing flip.

src/utils/openfeature.ts
```diff
@@ -118,6 +118,22 @@ const pathfinderFeatureFlags = {
...
+   * recording rides on. Defaults to true, so this is a remote off-switch
+   * rather than an opt-in: set it false to stop recording on a stack. Grafana
+   * core ships its own replay recorder behind `FlagKeys.FaroSessionReplay` —
+   * two rrweb instances must not run on one page, so this has to go false
+   * wherever core's goes true.
+   */
+  'pathfinder.session-replay': {
+    valueType: 'boolean',
+    values: [true, false],
+    defaultValue: true,
```
moxious, line 134: **[Product sign-off] Default-`true` for a brand-new full-DOM recorder fails open.**

Since `defaultValue` is the fallback when MTFF is unreachable, default-on means an MTFF outage/misconfig **fails open into recording**. This matches the `frontend-telemetry` precedent and respects the analytics-opt-out lever (`config.analytics.enabled === false` → no Faro → no replay), but full masked-DOM capture is a materially broader data class than the existing counter/error stream, and a revert or flag-flip does **not** undo recordings already collected (irreversible without cleanup).

Not a code defect — but recommend a conscious sign-off that default-on + fail-open is intended for the initial ship, and confirm a staged MTFF ramp is actually in place. Otherwise consider defaulting `false` and ramping up via MTFF.

docs/developer/FEATURE_FLAGS.md
```diff
@@ -33,6 +33,47 @@ The plugin uses the [OpenFeature](https://openfeature.dev/) standard with the OF
...
+**Default**: `true` (recording happens if the flag is not set, and if MTFF is unreachable)
+
+**Behavior**:
+
+- **`true`**: the recorder is registered the first time Pathfinder is opened in any surface, and runs for the rest of the page — including after Pathfinder is closed again
+- **`false`**: neither the replay module nor the rrweb bundle is fetched
+
+**Important**: this is an off-switch, not an opt-in — recording is the default state on every Cloud stack where telemetry is enabled. Two consequences worth holding onto:
+
+1. **Never let this run alongside Grafana core's own recorder.** Core ships one behind `FlagKeys.FaroSessionReplay`. Two rrweb instances on one page means double DOM serialisation on every mutation, and rrweb proxies `CSSStyleSheet.prototype.insertRule` globally while recording — Emotion's hot path in core Grafana. Because both flags now default toward recording, nothing prevents the collision automatically: set `pathfinder.session-replay` to `false` on any stack where core's flag goes true.
```
moxious: **[Medium note · doc accuracy] "both flags now default toward recording" is imprecise about core's default.**

Core's `faroSessionReplay` toggle is `@default false` (private preview) in `@grafana/data` — so there is no *automatic* default-state collision; it requires an operator to deliberately enable core's flag on the same stack. The mechanism you describe is real (verified: `@grafana/rrweb` globally proxies `CSSStyleSheet.prototype.insertRule`, not idempotent-guarded, so two recorders compound it on Emotion's hot path), but the trigger isn't "both default on." Suggest rewording to reflect that collision needs core's flag deliberately enabled.

Adversarial verification downgraded this from High to a doc/ops note precisely because of this — see the companion suggestion on `module.tsx`.

src/module.tsx
```diff
@@ -66,7 +66,11 @@ try {
   if (getFeatureFlagValue('pathfinder.frontend-telemetry', true)) {
     // Session enrichment (identity, surface, experiment cohorts) is owned by initFaro.
     const { initFaro } = await import('./lib/faro');
-    initFaro().catch((e) => logger.exception(e, { source: 'Faro init' }));
+    // Session replay is a second remote switch on top — also default-on, so a
+    // missing flag means recording. It captures the whole page, masked, from
+    // the first time Pathfinder is opened.
+    const sessionReplay = getFeatureFlagValue('pathfinder.session-replay', true);
```
moxious: **[Low · optional defense-in-depth] Make the core-recorder collision self-healing instead of runbook-dependent.**

Verification confirmed `config.featureToggles.faroSessionReplay` is a typed, first-class field in `@grafana/data`, and this repo already reads `config.featureToggles` idiomatically (`src/requirements-manager/checks/env.ts:15`, `src/utils/interactive-guides-api.ts:21`). A one-liner here — gate `sessionReplay` on `&& config.featureToggles?.faroSessionReplay !== true` — makes the safe state automatic rather than leaving it to an operator remembering to flip the flag.

Caveat: a private-preview toggle may not always be surfaced to the frontend (the read would return `undefined`), so treat this as belt-and-suspenders on top of the doc mitigation, not a replacement for it. Non-blocking.

src/lib/telemetry/replay-scrub.ts
```diff
@@ -0,0 +1,375 @@
...
+
+function scrubNode(node: unknown): void {
+  if (!isRecord(node)) {
+    return;
+  }
+  if (isRecord(node.attributes)) {
+    scrubAttributes(node.attributes);
+  }
+  scrubStyleText(node);
+  if (Array.isArray(node.childNodes)) {
```
moxious, line 437: **[Low · defense-in-depth] Comment-node text is unscrubbed — an allowlist-model asymmetry against `maskTextSelector: '*'`.**

`scrubNode` handles attributes, `<style>` text, and recurses `childNodes`, but never touches a comment node's `textContent`. Verified against dep source: rrweb serializes `COMMENT_NODE.textContent` verbatim (masking is TEXT_NODE-only), and neither faro's `buildRecordOptions` nor `REPLAY_OPTIONS` sets `slimDOMOptions`, so comments aren't slimmed either.

Adversarial verification dropped this from Medium → Low because there's no realistic user-text path: Pathfinder's `sanitizeDocumentationHTML` (DOMPurify, `ALLOW_COMMENTS` unset) strips comments, and React comment markers are data-free. Residual risk is third-party panel plugins emitting content-bearing comments into the full-page recording. Cheap optional fix: blank comment `textContent` (`node.type === 5 → textContent.replace(/\S/g, '*')`) before the `childNodes` recursion so both snapshot and mutation-add paths are covered.

src/lib/telemetry/replay-scrub.ts
```diff
@@ -0,0 +1,375 @@
...
+// CSS is the one allowlisted value that can still smuggle a URL, and rrweb
+// absolutifies those before we see them: a relative `url(logo.png?sig=…)`
+// arrives fully qualified with the query intact.
+//
+// The three arms are not interchangeable. A quoted URL may legally contain
+// parentheses (`url("a(1).png")`, or a data: SVG full of them), so it has to
+// run to its closing quote rather than to the first `)`. An unquoted one may
+// not contain parens, quotes or whitespace at all, so there the stricter class
+// is what keeps the match from running past the end of the value.
+const CSS_URL_PATTERN = /url\(\s*(?:"([^"]*)"|'([^']*)'|([^'")\s]*))\s*\)/gi;
```
moxious: **[Low · defense-in-depth] CSS url() scrub only matches the literal `url(` token.**

Bare-string resource references evade it — `image-set("a.png?sig=…" 1x)` and `@import "sheet.css?token=…"` are both valid CSS with no `url(`, so the query string passes through intact (the `css.includes('url(')` early-out also skips them entirely). Low severity because page CSS originates from Grafana's own theme/Emotion, so at worst a Grafana-origin cache-buster leaks, not user secrets. Optional pattern widening or a defensive note.

package.json
```diff
@@ -119,6 +119,7 @@
     "@grafana/assistant": "^0.1.25",
     "@grafana/data": "13.1.1",
     "@grafana/e2e-selectors": "13.1.1",
+    "@grafana/faro-instrumentation-replay": "^2.8.2",
```
moxious: **[Nit] Consider tilde-pinning `~2.8.2`.**

`ContinuousReplayInstrumentation` relies on undocumented internal invariants of `faro-instrumentation-replay` (`addListener` called only in `initialize()`; `removeListener` never called). Correct today and guarded by the real-SDK lifecycle test, but a caret minor bump could regress the internal coupling silently. Tilde-pinning makes that ride in only on deliberate bumps. (Also: the docs use British "serialisation" — repo/Grafana standard is US "serialization.")

moxious (changes requested):
Principal-level review (routed via `docs/design/CONCERNS.md`; 6 concern reviewers + 6 adversarial skeptics, grounded against the installed rrweb/faro dependency sources). This is a well-built PR — the scrubber's allowlist model is sound, the `ContinuousReplayInstrumentation` metas monkeypatch is correct and regression-tested, terminal credential blocking is triple-covered, and the lazy-chunk boundary holds. No critical/high code defect survived verification.

Requesting changes on three items I'd want resolved before this goes wide, all inline:

1. **Slug retention** (`url.ts`) — the one place the "all text is masked" guarantee provably fails: dashboard/folder titles leak slugified through the path. Needs an explicit decision (document as residual, or redact the slug segment).
2. **No end-to-end masking test** — coverage asserts *config*, never *behavior against real rrweb*. A dep bump that changed mask semantics would ship an unmasked recording with all tests green. Cheap to close.
3. **Default-on posture + `samplingRate: 1`** — full-page, 100%, indefinite, fail-open-on-MTFF-outage. Wants a conscious sign-off + confirmation of a staged MTFF ramp, and ideally a configurable sampling lever.

Rest are non-blocking (collision doc/guard, comment-node + CSS defense-in-depth, dep pin). Everything else — correctness, lifecycle, terminal blocking, graceful degradation, tech-debt, doc-drift — came back clean. Full breakdown in the inline comments.

src/lib/telemetry/replay-scrub.ts
```diff
@@ -0,0 +1,375 @@
...
+
+function scrubCssUrls(css: string): string {
+  // Emotion rewrites rules constantly, so the common case has to stay cheap.
+  if (!css.includes('url(')) {
+    return css;
+  }
+  return css.replace(CSS_URL_PATTERN, (match, doubleQuoted?: string, singleQuoted?: string, bare?: string): string => {
+    const target = doubleQuoted ?? singleQuoted ?? bare ?? '';
+    return target.startsWith('#') ? match : `url("${stripUrlSecrets(target)}")`;
+  });
```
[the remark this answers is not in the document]
Jayclifford345: Correct — fixed in 4b7d32da. Self-inflicted: I added the `includes('url(')` short-circuit as a fast path for Emotion's rule traffic and made it case-sensitive while the regex it guards is `/i`, so `URL(...)` and `Url(...)` were skipped before the regex ever ran.

Now a case-insensitive `/url\(/i.test(css)`, deliberately not `toLowerCase().includes(...)` — that would copy the whole stylesheet on every call just to decide there is nothing to do. Non-global, so it carries no `lastIndex` state between calls. Three new cases (`URL`, `Url`, `uRl`).

Jayclifford345:
### Sampling rate is now remotely configurable

Added `pathfinder.session-replay-sampling-rate` (number, default `1`) as a volume dial on top of the on/off switch, read through a new `getNumberFlagValue` sync helper next to the boolean in `module.tsx`. Variants are `[0, 0.1, 0.25, 0.5, 1]`; the decision is a deterministic hash of the session id, so a session either has a recording for its whole life or never does.

The reason this is worth a flag: every rrweb event becomes a Faro event carrying a JSON DOM payload, and a Grafana dashboard mutates continuously. If volume becomes a problem, the dial is a better first move than the switch.

**The rate is range-checked at the point of use rather than trusted.** `resolveSamplingRate` falls back to `1` for anything that is not a finite number in `[0, 1]` — `100` meant as a percentage, a string from a mistyped MTFF value, `NaN`, `Infinity`, a negative. This is deliberate history: [#1275](https://github.com/grafana/grafana-pathfinder-app/pull/1275) deleted an earlier Faro sample-rate flag outright rather than clamp it, because a clamp couldn't tell a deliberate remote `0` from a fat-fingered one. Failing to the default rather than to zero is what makes this one safe to reintroduce — a mistyped value over-records, which is visible and recoverable, instead of silently recording nobody. An explicit `0` is still honoured, because it really is the number 0.

Ten new cases on `resolveSamplingRate`, plus registry assertions. `FEATURE_FLAGS.md` and `TELEMETRY.md` updated.

### Still open

The **R1 subscription leak** from the first review is the one finding I haven't actioned, and I still don't think it holds — [reasoning here](https://github.com/grafana/grafana-pathfinder-app/pull/1482#issuecomment-5143042696). Short version: the only un-unsubscribed path is "Pathfinder never opened", which leaves one closure in a module-level `Set` that lives as long as the document; and the suggested remediation has nowhere to land, because `initFaro` is fire-and-forget with no teardown and the `stampSurface` listener two lines above it has never been unsubscribed either. Happy to do it if someone wants the telemetry adapter to grow a lifecycle, but that's a separate change from this PR.

Everything else across all four review rounds is addressed. `npm run check` green: 386 suites, 6163 tests.

src/lib/telemetry/url.ts
Jayclifford345: Fixed in 47fe85ae. You and Bugbot found this independently, which is a fair signal it was the real one.

Agreed it contradicts the headline guarantee, so I redacted rather than documented. `stripUrlSecrets` now drops the title segment and keeps the uid — `/d/<uid>/<slug>` → `/d/<uid>`, same for `/d-solo/` and `/dashboards/f/`. The uid identifies the board without describing it, which is what the replay actually needs for navigability.

Confirmed end-to-end rather than by inspection: the new `replay-masking.test.ts` drives real rrweb over a real DOM and the captured payload shows `"href":"https://foo.grafana.net/d/abc123"` in both the Meta event and the `<a>` attribute, with the slug gone.

One residual I should flag rather than leave for you to find: **Faro's own `page_id` meta still carries the slug**, and always has — I saw `page_id: /d/*/welcome-to-learn-grafana` in the Loki data while debugging the session split. That is core Faro deriving page meta from `location`, not anything replay introduced, so it predates this PR and affects the existing event stream too. Out of scope here, but it means the title is still reaching the collector by that route and probably wants its own fix.

src/lib/telemetry/replay-lifecycle.test.ts
Jayclifford345: fixed in 47fe85ae with `src/lib/telemetry/replay-masking.test.ts`.

You were exactly right about the gap: `faro.test.ts` asserted the options, `replay-scrub.test.ts` ran the scrubber over hand-built events, and the lifecycle test filtered the payloads out. Nothing connected "options set" to "output masked", so an SDK bump that renamed an option or changed `maskTextSelector: '*'` semantics would have gone green.

The new test renders a fixture carrying nine distinct known secrets — heading text, paragraph text, input `value`, `aria-label`, `data-testid`, `title`, `placeholder`, an HTML comment, and a signed image query — drives the real recorder, and asserts none of them appear in what reaches `CaptureTransport`. It also asserts the recording is non-empty and contains the fixture id first, so the suite fails loudly rather than passing vacuously if the recorder ever stops producing.

Two things it caught that I would otherwise have shipped: comment-node text was unmasked (your separate finding below), and the dashboard slug was surviving in `href`. Both now covered by this test rather than only by unit tests over synthetic input.

src/lib/telemetry/replay.ts
Jayclifford345: Already landed in 4b7d32da, which crossed with your review — `pathfinder.session-replay-sampling-rate` (number, default `1`) is now a remote flag alongside the switch, with variants `[0, 0.1, 0.25, 0.5, 1]`. The decision is a deterministic hash of the session id, so a session either has a recording for its whole life or never does.

It is range-checked at the point of use rather than trusted: `resolveSamplingRate` falls back to `1` for anything that is not a finite number in `[0, 1]`. That is deliberate history — #1275 deleted an earlier Faro sample-rate flag outright rather than clamp it, because a clamp could not distinguish a deliberate remote `0` from a fat-fingered one. Failing to the default rather than to zero is what makes it safe to reintroduce.

Registered in MTFF in grafana/deployment_tools#672847 along with the two booleans, none of which existed there before.

**Your auto-refresh point stands and is not fixed.** The 60s inactivity pause keys off pointer/scroll/key/input, so a ticking dashboard does not defeat it — but the user watching one does not either, and a long-lived tab still ships a long recording. There is no max-duration or idle-teardown option on the instrumentation, so capping it means either tearing the recorder down ourselves on a timer or an upstream change. I would rather land that as a follow-up with a real volume number behind it than guess a cap now; the sampling dial is the lever for the initial ramp.

src/utils/openfeature.ts
Jayclifford345: Confirming this is a conscious decision, not a default that slipped through.

Chose default-on, on the grounds that an opt-in flag collects nothing until someone remembers to target a stack, and that the MTFF flag is meant as a kill switch rather than a rollout gate. That matches how `pathfinder.frontend-telemetry` already ships.

On the staged ramp: the flags did not exist in MTFF at all until grafana/deployment_tools#672847, which is open now. Worth being precise about what that PR does — since the plugin defaults to `true`, absence from MTFF already meant "on", so registering them changes no behaviour. What it changes is that the switch becomes reachable at all; before it, the only way to stop recording was a plugin release. If you want a staged ramp instead, that PR is the one-line place to do it (narrow the wave list, serve `false` to the rest) and I have called that out in its description.

Your irreversibility point is in the PR body under Reversibility, explicitly: flipping the flag stops collection but does not undo recordings already taken.

docs/developer/FEATURE_FLAGS.md
Jayclifford345: Correct, and my wording was wrong — fixed in 47fe85ae. Verified the same way you did: `faroSessionReplay?: boolean` with `@default false` in `@grafana/data`, so there is no automatic default-state collision.

Reworded to say the collision needs core's toggle deliberately enabled on a stack that also has ours on, while keeping the mechanism — which your verification and mine agree on: rrweb's global `CSSStyleSheet.prototype.insertRule` proxy is not idempotent-guarded, so two recorders compound on Emotion's hot path. Also took the companion suggestion; see the `module.tsx` thread.

src/module.tsx
Jayclifford345: Taken — in 47fe85ae. `config.featureToggles?.faroSessionReplay === true` now gates `sessionReplay` in `module.tsx`, so the safe state is automatic.

Your caveat is the important part and I kept it rather than treating this as the fix: a private-preview toggle may not be surfaced to the frontend, in which case the read is `undefined` and this does nothing. The comment at the call site and the `FEATURE_FLAGS.md` note both say to still set `pathfinder.session-replay` false on any stack where core's goes true. Belt-and-braces, as you framed it.

Confirmed the field is typed in the installed `@grafana/data` rather than assuming, and used the direct typed read rather than the `as Record<string, boolean>` cast the two existing call sites use, since this one has a real type.

src/lib/telemetry/replay-scrub.ts
Jayclifford345: Real, and now fixed in 47fe85ae rather than left as optional — the new real-rrweb test made it cheap to prove.

Your read of the dep source matches mine: masking is TEXT_NODE-only and neither `buildRecordOptions` nor `REPLAY_OPTIONS` sets `slimDOMOptions`, so comment `textContent` was serialized verbatim. `scrubNode` now blanks non-whitespace in `type === 5` nodes before recursing, so both the snapshot and mutation-add paths are covered.

Agreed on the severity reasoning — DOMPurify strips comments from our own content and React markers are data-free, leaving third-party panel plugins as the only realistic writer. That is precisely why I would rather not have to reason about it per plugin. The masking test now asserts a comment containing a known address does not survive.

src/lib/telemetry/replay-scrub.ts
Jayclifford345: Fixed in 47fe85ae. Both cases you named now scrub: `@import "…?token=…"` and `image-set("…?sig=…" 1x)`.

The early-out was the sharper half of this — `css.includes('url(')` skipped those strings entirely, so widening the pattern without widening the guard would have fixed nothing. The guard is now `/url\(|@import|image-set\(/i`, still a single non-global test with no string copy, so Emotion's rule traffic stays cheap.

Kept it targeted rather than scrubbing every quoted string in a stylesheet: `content: "…"` is a legitimate quoted string that has nothing to do with resources, and mangling it would corrupt playback for no privacy gain. Agreed on severity — page CSS is Grafana-origin, so the realistic worst case was a cache-buster.

package.json
Jayclifford345: Both taken in 47fe85ae — tilde-pinned to `~2.8.2`, and the British spellings switched to US across the code and docs I added.

Your reasoning for the pin is the right one and worth restating for whoever bumps it next: `ContinuousReplayInstrumentation` swaps `this.metas` for a filtering facade around `super.initialize()`, which holds only because the base class calls `addListener` exactly once from `initialize()` and never calls `removeListener`. Neither is documented. `replay-lifecycle.test.ts` runs against the real SDK so a regression would fail rather than pass silently, but a caret minor would still have shipped the bump before anyone looked at it.

src/lib/telemetry/url.ts
```diff
@@ -19,3 +19,37 @@ export function normalizeTelemetryUrl(url: string): string {
...
+  } catch {
+    return '';
+  }
+}
```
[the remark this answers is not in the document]
Jayclifford345: Same finding as @moxious's on `url.ts:51` — fixed in 47fe85ae.

`stripUrlSecrets` now drops the title segment from the three title-bearing Grafana route shapes and keeps the uid: `/d/<uid>/<slug>` → `/d/<uid>`, likewise `/d-solo/` and `/dashboards/f/`. Both locations you flagged are covered, since `replay-scrub.ts` routes every URL attribute and the Meta `href` through that same function.

Verified against real rrweb output rather than by inspection — the new `replay-masking.test.ts` asserts the captured payload contains `/d/abc123` and not the slug.

src/lib/telemetry/replay-scrub.ts
```diff
@@ -0,0 +1,411 @@
...
+    for (const child of node.childNodes) {
+      scrubNode(child);
+    }
+  }
+}
+
+// Mutates in place. rrweb hands over the event it has just built and keeps no
+// reference to it, and a full snapshot of a Grafana page is far too large to
+// deep-clone on every emit.
+export function scrubReplayEvent(event: ReplayEvent): ReplayEvent {
```
moxious, line 447: The scrub only reaches the rrweb payload — Faro sets `item.meta.page.url` from `location.href` independently, so the slug and query survive here. A live capture of `/d/abc123/acme-corp-quarterly-revenue?orgId=1` still emits both in the transport metadata of every session-replay event. Please run the scrubbed URL through `stripUrlSecrets` on the transport meta too (or set a page-meta override), otherwise the "all text masked" guarantee has a systematic hole one layer out from where it was fixed.

src/lib/telemetry/replay-masking.test.ts
```diff
@@ -0,0 +1,107 @@
...
+      isolate: true,
+      globalObjectKey: 'pathfinderReplayMasking',
+      batching: { enabled: false },
+      dedupe: false,
+    });
+    await activateSessionReplay(faro);
+
+    recorded = transport.items
+      .filter(
+        (item) => item.type === 'event' && (item.payload as { name?: string }).name === 'faro.session_recording.event'
```
moxious: This asserts against `item.payload.attributes` but never inspects `item.meta` — which is exactly where the slug/query residual lives. Extending the assertion to reject the slug and query in `item.meta.page.url` would have caught it and will guard the fix.

docs/developer/TELEMETRY.md
```diff
@@ -60,7 +61,15 @@ Privacy protection is split between enforced normalization and caller discipline
 - **Errors** in typed facade events use low-cardinality classifications such as `recordSequenceActionError`. `logger.error`, `logger.exception`, and direct `pushFaroError` calls retain the exception message, so callers must not include selectors, echoed input, or user-derived text.
 - **Attributes** passed through `stringifyAttributes`—including event, user-action, and session attributes—are stringified and truncated to 500 characters. Measurement and exception contexts must use small, typed values at the call site.
 - Never add high-cardinality or user-derived free-text attributes; new user-derived fields need privacy review (`analytics-and-telemetry` concern).
+- **DOM capture** is a different privacy surface from the attribute rules above, and session replay is the only thing that does it. rrweb records the whole page, Grafana core included; there is no subtree scoping, and no way to unmask a carve-out (`maskTextSelector` resolves through `closest()`, and this rrweb fork has no `unmaskTextSelector`). What that buys and what it does not:
```
moxious, line 64: If the `meta.page.url` slug/query is going to be accepted rather than scrubbed, it needs to be called out here as a known residual — the notes currently claim URL secrets are closed by `replay-scrub.ts`, which isn't true for the transport metadata path.

src/module.tsx
```diff
@@ -66,7 +67,21 @@ try {
...
+    // the switch, range-checked in lib/telemetry/replay.
+    //
+    // Yielding to core's own recorder is automatic rather than a runbook step:
+    // two rrweb instances on one page double DOM serialization per mutation and
+    // compound its global CSSStyleSheet.insertRule proxy, which is Emotion's
+    // hot path. The toggle is private-preview and may not be surfaced to the
+    // frontend at all, so this is belt-and-braces over the flag, not a
+    // replacement for it.
+    const coreIsRecording = config.featureToggles?.faroSessionReplay === true;
+    const sessionReplay = !coreIsRecording && getFeatureFlagValue('pathfinder.session-replay', true);
```
moxious: `true` / `true` / `1` here plus fail-open access means an OFREP outage or missing flag records everyone at full rate. That's a defensible launch choice but it's a product decision — please confirm it's intended, and confirm `grafana/deployment_tools#672847` lands and is verified in OFREP before this ships.

moxious (dismissed):
Thanks for the thorough turnaround — most of the first round is genuinely closed. Verified against head 47fe85a: the real-rrweb masking test, the comment-node scrub, the CSS `image-set`/`@import` coverage, the `FEATURE_FLAGS` wording, the self-healing core-recorder gate, and the tilde-pin + spelling all check out. I also agree the cursor "R1 subscription leak" is a non-issue — it's bounded lifetime retention, not an accumulating leak, and there's no teardown path for a disposer to hang off.

Two things still hold this at changes-requested, both refinements of the original blockers rather than new scope:

1. **Slug/title still crosses the boundary via Faro's page metadata.** The rrweb `Meta` `href` and URL attributes are correctly scrubbed now, but Faro builds `item.meta.page.url` directly from `location.href`, and `beforeSend` / `scrubReplayEvent` only touches the rrweb payload — so a live capture of `/d/abc123/acme-corp-quarterly-revenue?orgId=1` still ships the full slug **and** query in the transport metadata of every `faro.session_recording.event`. This is the exact "all text is masked" exception from the first round; it just moved one layer out. Needs either the transport meta scrubbed too, or an explicit documented residual in `TELEMETRY.md`. Right now it's neither.

2. **Rollout posture needs a conscious decision, and the remote controls don't exist yet.** The `samplingRate` lever is correctly wired and I confirmed an explicit `0` actually stops recording — good. But the switch and dial only become reachable once `grafana/deployment_tools#672847` lands, and that PR currently serves replay `true` / rate `1` in every wave with no ramp. Combined with default-on + fail-open (an OFREP outage falls open *into* recording, indefinitely, and already-collected recordings aren't undone by a later flip), I want a deliberate sign-off that this posture is intended for initial ship — or default it off and ramp via MTFF. Please also land #672847 first; shipping this ahead of it leaves an interval where recording is on everywhere at 100% with no remote remediation.

Everything else is clear. Happy to re-approve once the metadata residual is closed (scrubbed or documented) and the rollout decision + landing order are settled.

src/module.tsx
Jayclifford345: Yes, this is defensible as we do not want to lose observability over pathfinder due to OFREP outage. On is better than off in this case

src/lib/telemetry/replay-scrub.ts
Jayclifford345: Fixed in 5e490377 — `redactPageUrl` runs on every item in `beforeSend`, dropping the query from `meta.page.url`.

The path stays deliberately: it's what makes a replay navigable, and a board name isn't a secret to anyone who can already read this telemetry. The query is a different class — `var-*` values, Explore queries — so that goes.

Worth knowing: Grafana Cloud's collector explodes that query into `page.attributes`, so `page_attr_var_member`, `page_attr_var_ticker` and similar are live in the ops stack today, predating replay. Of 100 sampled streams with `page_attr_*`, 100 had a query in `page_url` and 0 didn't — so removing it at source removes the derived fields too.

src/lib/telemetry/replay-masking.test.ts
Jayclifford345: Done in 5e490377 — the test now reads `item.meta` as well as `item.payload`, asserting a page URL is present with the board path and that none contain a `?`.

Checked it isn't vacuous: removing the `beforeSend` makes it fail and the other 14 pass.

docs/developer/TELEMETRY.md
Jayclifford345: Updated in 5e490377, as a split rather than a blanket residual: the path is kept (stated as a deliberate trade — the masking guarantee covers rendered text and DOM attributes, page URLs are outside it), the query is stripped on both channels, and the collector's query-to-`page.attributes` expansion is documented since that's what makes the metadata channel matter.

src/lib/telemetry/url.ts
```diff
@@ -19,3 +19,49 @@ export function normalizeTelemetryUrl(url: string): string {
...
+    // javascript:/vbscript:/blob: and friends, none of which mean anything on
+    // playback and all of which are better not round-tripped.
+    if (parsed.protocol !== 'http:' && parsed.protocol !== 'https:') {
+      return '';
+    }
+    parsed.search = '';
+    parsed.hash = '';
+    parsed.username = '';
+    parsed.password = '';
+    parsed.pathname = redactTitleSlug(parsed.pathname);
```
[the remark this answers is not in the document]
Jayclifford345: Good catch, and more relevant since we decided to keep the path — fixed in a25cb8db.

A title slug isn't a secret, but these are: the segment after `/public-dashboards/` or `/dashboard/snapshot/` is the whole credential. Both are now redacted; the rest of the path still stands.

src/lib/telemetry/filtering.ts
```diff
@@ -148,9 +148,10 @@ function redactEmbeddedUrls(text: string): string {
 // Pathfinder, and PerformanceInstrumentation's resource entries must be
 // attributable to a domain we actually fetch from. Page-wide navigation
 // timing (the whole page's load, not a specific resource) is always dropped.
-// Everything else — our own pushed events/user-actions/measurements — only
-// ever originates from this isolated instance's own API, so it passes
-// through unfiltered.
+// Everything else — our own pushed events/user-actions/measurements, plus the
+// faro.session_recording.* stream — only ever originates from this isolated
+// instance's own API, so it passes through unfiltered. Replay payloads are
+// scrubbed at their own source, in telemetry/replay-scrub.
```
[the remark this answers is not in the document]
Jayclifford345: Same finding as @moxious's on `replay-scrub.ts` — this reviewed 864a0041, one commit before the fix.

Fixed in 5e490377: `redactPageUrl` now runs on every item in `beforeSend`, stripping the query from `meta.page.url`. `TELEMETRY.md` no longer claims URLs are closed generally — it states that paths are kept deliberately and only the query is removed. `replay-masking.test.ts` asserts against `item.meta` too now, and share tokens in the path are redacted in a25cb8db.

src/lib/telemetry/replay-scrub.ts
```diff
@@ -0,0 +1,411 @@
...
+// Same case-insensitivity as the patterns it guards, and deliberately not
+// `toLowerCase().includes(...)` — that would copy the whole stylesheet on
+// every call just to decide there is nothing to do. Non-global, so it carries
+// no lastIndex state between calls.
+const HAS_CSS_RESOURCE = /url\(|@import|image-set\(/i;
+
+function scrubCssUrls(css: string): string {
+  // Emotion rewrites rules constantly, so the common case has to stay cheap.
+  if (!HAS_CSS_RESOURCE.test(css)) {
+    return css;
```
leslie-heinzen, line 320: **[High · security] Non-URL CSS text bypasses the masking guarantee.** rrweb serializes `<style>` text, `_cssText`, inserted rules, and declaration values without text masking, while this fast path returns CSS with no resource syntax unchanged. A third-party panel can therefore emit customer-derived text through `content: "alice@example.com"` or a text-bearing custom property even though `TELEMETRY.md` says page URLs are the one exception to ‘all text is masked.’ Parse and mask/drop text-bearing CSS values across snapshot and CSSOM routes, with real-recorder coverage; otherwise explicitly document and approve this residual before launch. Already-ingested recordings are not reversible.

src/module.tsx
```diff
@@ -66,7 +67,21 @@ try {
...
+    const sessionReplaySamplingRate = getNumberFlagValue('pathfinder.session-replay-sampling-rate', 1);
```
leslie-heinzen: **[High · rollout prerequisite] Ensure the remote controls exist before these default-on values ship.** Both evaluations fall back to replay enabled at rate `1`, but `grafana/deployment_tools#672847` is still open, so shipping this first creates an interval where every eligible new page records at full rate with no reachable switch or dial. Make that PR’s merge and OFREP verification a release prerequisite, preferably serving replay off or at a reduced initial rate.

docs/developer/TELEMETRY.md
```diff
@@ -56,11 +57,23 @@ Span helpers (`withFaroUserAction`, `setFaroUserActionAttributes`), explicit err
...
+Faro initializes only when `resolveFaroEnvironment()` resolves: Grafana Cloud with analytics enabled, on `.grafana.com` / `.grafana.net` / `.grafana-ops.net` / `.grafana-dev.net` hosts, and only when the default-on `pathfinder.frontend-telemetry` flag is set. Local development sends nothing unless `localStorage['pathfinder.faro.local'] = 'true'` in a dev build. The activity gate drops everything except errors until Pathfinder is opened, so collector sessions mean "used Pathfinder or Pathfinder errored", not "loaded a Grafana page".
+
+Session replay adds a second remote switch (`pathfinder.session-replay`, also default-on) plus a volume dial (`pathfinder.session-replay-sampling-rate`, default `1`, range-checked at the point of use), but no new environment gate: it is registered from inside `initFaro`, after the `resolveFaroEnvironment()` early return, so a self-hosted or OSS Grafana never reaches it — such an instance does not construct a Faro instance in the first place, and the rrweb chunk is never fetched. The open that latches the activity gate is also what starts the recording, and once started it runs for the rest of the page, including after Pathfinder is closed again.
```
leslie-heinzen, line 78: **[High · rollback] A flag flip does not stop active recordings.** Replay values are read once during bootstrap, OFREP visibility refresh is disabled, and the replay instrumentation has no removal path. Long-lived tabs therefore keep recording after the advertised kill switch changes or the plugin is reverted; only later page loads see the new value. Add a live refresh/teardown path, or narrow the rollback contract and runbook to require reloads and define deletion handling for already-ingested recordings.

src/lib/telemetry/faro-adapter.ts
```diff
@@ -109,6 +121,31 @@ export async function initFaro(): Promise<void> {
...
+// The same open that latches passesActivityGate starts the recording, so the
+// first thing rrweb emits is already past the gate. The trailing start() covers
+// the surface having reported itself while the SDK chunk was still loading.
+function startSessionReplayOnFirstOpen(faro: Faro, samplingRate?: number): void {
+  let started = false;
+  const start = () => {
+    if (started || !isPathfinderOpen()) {
+      return;
+    }
+    started = true;
```
leslie-heinzen: **[Medium · correctness] Do not consume the one start trigger before activation succeeds.** `started` is latched and the listener is removed before the dynamic import and `activateSessionReplay()` resolve; the catch then swallows rejection. One transient chunk or SDK failure permanently disables replay for that tab and produces no stable signal. Track `activating` separately, latch/unsubscribe on success, retain a bounded retry path, and add a rejection/reopen test.

docs/developer/TELEMETRY.md
```diff
@@ -56,11 +57,23 @@ Span helpers (`withFaroUserAction`, `setFaroUserActionAttributes`), explicit err
...
+
+Two consequences of it being default-on: recordings are only viewable on a stack where Grafana has switched on the private-preview feature, and Grafana core ships its own replay recorder behind `FlagKeys.FaroSessionReplay` — nothing stops both from running on the same page, so `pathfinder.session-replay` must be set false wherever core's flag goes true.
```
leslie-heinzen: **[Low · contract/test] This now contradicts the implementation and `FEATURE_FLAGS.md`.** `module.tsx` automatically suppresses Pathfinder replay when `config.featureToggles.faroSessionReplay` is surfaced as true, but no semantic bootstrap test exercises that ownership decision. Update this text to describe the guarded-but-belt-and-braces behavior and add a test proving the core-enabled case passes `sessionReplay: false` to `initFaro`.

leslie-heinzen (changes requested):
Principal-level follow-up review at `a25cb8d`, routed through the repository concern registry and checked against all prior review threads. Five medium-or-higher candidates were independently challenged by three refuters each; none was dropped.

Requesting changes on three blockers:

1. Non-URL stylesheet text (`content:`, text-bearing custom properties, and CSSOM values) remains outside rrweb masking and outside the URL-only scrubber.
2. The default-on/full-rate code must not ship before `grafana/deployment_tools#672847` registers and verifies the remote controls. That PR remains open.
3. The advertised kill switch only affects subsequent page loads; active tabs continue recording because flags are startup snapshots and there is no recorder removal path.

Two current code/verification issues also need attention:

- A failed lazy import or replay registration permanently consumes the one start trigger and emits no diagnostic.
- The core-recorder exclusion guard exists, but has no semantic bootstrap test and `TELEMETRY.md` states the opposite behavior.

I independently confirmed the existing `image-set()` query-secret escape and am not duplicating its inline thread: https://github.com/grafana/grafana-pathfinder-app/pull/1482#discussion_r3702468513

Prior findings now verified fixed include incremental CSS sources, page metadata query redaction, capability-path tokens, real-rrweb masking coverage, sampling control, terminal fallback, and dependency pinning. The earlier subscription-leak report remains a non-finding: it is bounded page-lifetime retention, not an accumulating leak. Dashboard slug retention and the default-on/fail-open posture are explicit product decisions; those decisions do not resolve control availability or active-tab rollback.

Validation: 7 focused suites / 268 tests pass at this head; current GitHub checks and bundle comparison are green; replay remains lazy-loaded; no replay-package high/critical advisory delta was found.

Non-blocking instrumentation note: activation failure and invalid-sampling fallback have no low-cardinality stable outcome signal. Tech-debt scan also flags continued bootstrap accumulation in `module.tsx` and comment-heavy internal API documentation in `openfeature.ts`. No agent-guidance drift was found.

src/lib/telemetry/replay-scrub.ts
```diff
@@ -0,0 +1,411 @@
...
+      return target.startsWith('#') ? match : `url("${stripUrlSecrets(target)}")`;
+    })
+    .replace(CSS_IMPORT_PATTERN, (_match, keyword: string, _quote, target: string) => {
+      return `${keyword}"${stripUrlSecrets(target)}"`;
+    })
+    .replace(CSS_IMAGE_SET_PATTERN, (_match, keyword: string, args: string) => {
+      // url() entries inside image-set were already handled above; this only
+      // catches its bare-string form.
+      return `${keyword}${args.replace(CSS_QUOTED_STRING, (_s, _q, target: string) => `"${stripUrlSecrets(target)}"`)})`;
+    });
```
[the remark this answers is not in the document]
Jayclifford345: Half right, and the half that is right is fixed in fed09671.

**The truncation claim does not hold.** `CSS_IMAGE_SET_PATTERN` did stop at the first `)`, but the replacement re-appended one — so `image-set(url("a.png")` was rewritten to itself, byte-identical. Density descriptors were never severed. I checked by running the old pattern over `image-set(url(/a.png?sig=s) 1x, "/b.png?sig=s" 2x)` rather than reading it.

**The escape is real, and wider than the bare-string case you named.** Stopping at the first `)` lands *inside* the argument list whenever an earlier entry contains a parenthesis, so everything after it went unscrubbed — a preceding `url()` entry does it, and so does a filename like `a(1).png` or a query holding a paren. Same class as the gap already closed for `url()`, as you say.

Fixed by walking to the balanced close rather than matching to the first `)`: `scrubImageSets` finds `image-set(`, tracks quote state and paren depth to the real end, and scrubs every entry in between. A `type("image/avif")` descriptor is now skipped explicitly — the balanced walk reaches it where the old pattern stopped short, and running a MIME type through `stripUrlSecrets` would rewrite it into an absolute URL.

Three new cases: a bare string following a `url()` entry, a first entry with a paren in the filename, and the `type()` descriptor.

src/lib/telemetry/replay-scrub.ts
Jayclifford345: Confirmed and fixed in fed09671. This one was real, and I checked it rather than reasoning about it.

I added a `content:` declaration and a text-bearing custom property to the real-recorder fixture in `replay-masking.test.ts`. Both secrets reached the transport verbatim before the fix — two channels, exactly as you describe: `<style>` text (rrweb's `serializeTextNode` skips masking when the parent is `STYLE`) and the `style` attribute.

The fix masks string literals in the two declarations that can put author text on screen — `content` and custom properties — on every route CSS takes to the collector: snapshot `<style>` text, `_cssText`, inserted rules, `replace`/`replaceSync`, adopted sheets, the styleOMValue diff, and single CSSOM writes. The last two carry the property name separately from the value, so those get it passed in rather than pattern-matched out of a declaration.

Two deliberate limits:

- **Escape sequences survive the mask.** `content: "\e900"` still draws its glyph. Masking those would turn every icon-font pseudo-element into asterisk soup for no privacy gain.
- **A resource-bearing value is left to the URL pass.** `content: url("…?sig=…")` is already scrubbed there, and masking its quoted target would break the image to close a leak that is already closed.

One residual I would rather name than have you find: quoted strings *outside* those two declarations are untouched — `font-family: "Inter"` and friends. They cannot carry customer text, and masking them would change layout in playback. The drift runs the other way instead: a theme that puts a quoted string in a custom property plays back with it asterisked. Both now in `TELEMETRY.md` under the DOM-capture bullet.

I did **not** take the "parse CSS" route. A tokenizer on Emotion's `insertRule` hot path is a cost I do not want to carry for a channel that only leaks if a third-party plugin interpolates customer text into a stylesheet. The guard regex still fast-exits an ordinary rule, and its `content` arm is boundary-anchored so `align-content` and `justify-content` — which are everywhere in Grafana's flex layouts — do not pay even that.

src/lib/telemetry/faro-adapter.ts
Jayclifford345: Real, and fixed in fed09671 — the latch was on the attempt, not on the activation.

`started` now closes only in the `.then` after `activateSessionReplay` resolves. A rejection resets `activating` and leaves the subscription in place, so the next open retries. Bounded at three attempts, because a failure that is not transient would otherwise re-import the chunk on every sidebar toggle for the life of the page; the subscription is dropped once they are spent.

Both halves emit now: `pathfinder_session_replay_activation_failed` carries `exhausted: true|false`, so a transient blip and a dead chunk are distinguishable in the same counter.

Your non-blocking note about the invalid-sampling fallback is covered too, and without a new import: `activateSessionReplay` returns the rate it actually used, and the adapter reports `pathfinder_session_replay_sampling_fallback` with `reason: out_of_range | not_a_number` when that differs from what it asked for. Emitting it from inside `replay.ts` would have needed a static import back into `faro-adapter`, which the cycle check in `architecture.test.ts` would have flagged for a diagnostic.

Three new cases: retry-on-reopen, give-up-after-three, and latch-once-resolved. The first fails against the old code.

src/module.tsx
Jayclifford345: Agreed as a release prerequisite rather than a code change: grafana/deployment_tools#672847 lands and is verified in OFREP before this ships. It is open and we will get it merged — I am treating merge order as the mitigation rather than defaulting the plugin off, because the default-on posture is a deliberate call (confirmed with @moxious in the thread above): an OFREP outage should not cost us observability of Pathfinder.

Worth being precise about what #672847 does, because it reads like a rollout gate and is not one. The plugin already defaults to `true`, so absence from MTFF *already* means "on". Registering the flags changes no behaviour — what it changes is that the switch becomes reachable at all. Before it, the only way to stop recording is a plugin release. That is exactly why landing it first is the right sequencing, and it is the one-line place to serve a narrower initial rate if we decide we want one.

One thing this PR does change on this line: the core-recorder gate moved out of `module.tsx` into `resolveSessionReplayOptions`, so the decision is unit-tested rather than inline in the bootstrap. Same behaviour.

docs/developer/TELEMETRY.md
Jayclifford345: You are right on the mechanism, and I am taking the hit rather than fixing it — narrowing the contract, which is the second branch of your own suggestion.

Re-evaluating the flag mid-session means either polling OFREP or tearing the recorder down live, and a torn-down rrweb leaves a mutation stream with no snapshot to apply it to. That is more machinery than this failure mode justifies, so the honest move is to stop advertising something the code does not do.

`TELEMETRY.md` gains a **Stopping a recording** section in fed09671 saying it plainly:

- both replay flags are read once at plugin bootstrap; a flip reaches a tab only on its next page load
- the kill switch is "no new recordings", not "recording stops now" — budget for the tail of long-lived tabs, worst case a dashboard left open overnight
- recordings already ingested are not undone by the flip; removing them is a collector-side deletion request against Frontend Observability, not something a flag or a release can do
- if a recording must stop immediately on a known stack, the only in-band lever is a plugin release plus a forced reload; otherwise it is the flag flip plus natural page turnover

`FEATURE_FLAGS.md` carries the same warning on `pathfinder.session-replay` and links to it, so nobody reaching for the switch reads it as immediate.

docs/developer/TELEMETRY.md
Jayclifford345: Both taken in fed09671.

The text was left over from before the guard existed. It now says what the code does: `resolveSessionReplayOptions` yields automatically when `config.featureToggles.faroSessionReplay` reads `true`, so the safe state is not a runbook step — with the caveat kept, that a private-preview toggle may never be surfaced to the frontend, in which case the read is `undefined`, the guard does nothing, and the remote flag is the only lever. Belt-and-braces, not a replacement. `FEATURE_FLAGS.md` now names the same function rather than `module.tsx`.

On the missing test: the decision was inline in the plugin bootstrap, which is why it had none. It is extracted to `resolveSessionReplayOptions` in `faro-adapter.ts` and covered four ways — records when the flag is on and core is not recording, yields when core's toggle is `true`, leaves the flag in charge when the toggle is not surfaced, and passes the rate through untouched for range-checking at the point of use.

Jayclifford345:
Round five addressed in fed09671. `npm run check` green — 387 suites, 6237 tests.

### Fixed

**Non-URL CSS text (High).** Real, and I proved it before fixing it: adding a `content:` declaration and a text-bearing custom property to the real-recorder fixture made both secrets reach the transport verbatim. `content` and custom-property string literals are now masked on every route CSS takes — snapshot `<style>` text, `_cssText`, inserted rules, `replace`/`replaceSync`, adopted sheets, the styleOMValue diff, and single CSSOM writes. Escape sequences survive so icon glyphs still render, and a resource-bearing value is left to the URL pass rather than masked twice. No tokenizer: the guard regex fast-exits an ordinary Emotion rule, and its `content` arm is boundary-anchored so `align-content`/`justify-content` do not pay for it.

**Activation consumes the start trigger (Medium).** The latch was on the attempt rather than on the resolved activation. It now closes in the `.then`, a rejection leaves the subscription in place so the next open retries, bounded at three. `pathfinder_session_replay_activation_failed` carries `exhausted: true|false`. The invalid-sampling signal is covered too — `activateSessionReplay` returns the rate it used, so the adapter can report a fallback without a static import back into itself.

**Doc contradicts the core-recorder guard, and the guard has no test.** The decision moved out of the plugin bootstrap into `resolveSessionReplayOptions`, which is now covered four ways; `TELEMETRY.md` and `FEATURE_FLAGS.md` describe the automatic-but-belt-and-braces behaviour instead of the pre-guard runbook step.

**`image-set()` escape** (bugbot, [r3702468513](https://github.com/grafana/grafana-pathfinder-app/pull/1482#discussion_r3702468513) — thanks for flagging it rather than duplicating). Now walks to the balanced close, which also exposed a `type("image/avif")` descriptor the old pattern never reached and would have rewritten into a URL.

### Taken as-is

**A flag flip does not stop active recordings (High).** Correct, and I am taking the hit rather than building a live teardown — polling OFREP or removing rrweb mid-session both cost more than this failure mode is worth, and a torn-down recorder leaves mutations with no snapshot to apply them to. So the contract narrows instead: `TELEMETRY.md` gains a **Stopping a recording** section stating that both flags are bootstrap snapshots, that the switch means "no new recordings" rather than "recording stops now", that long-lived tabs are the tail to budget for, and that already-ingested recordings need a collector-side deletion request rather than a flag or a release. `FEATURE_FLAGS.md` links to it from the switch itself.

**Rollout prerequisite (High).** Accepted as a release prerequisite, not a code change. grafana/deployment_tools#672847 is open and will be merged and OFREP-verified before this ships. Not defaulting the plugin off: default-on is the deliberate posture, on the grounds that an OFREP outage should not cost us observability of Pathfinder. Worth restating that #672847 changes no behaviour — absence from MTFF already means "on" — it makes the switch reachable, which is precisely why it goes first.

### Not addressed

Nothing outstanding from this round. The earlier `resolveSamplingRate` observation about long auto-refreshing sessions still stands as a follow-up: there is no max-duration or idle-teardown option on the instrumentation, and I would rather cap it against a real volume number than guess one now. The sampling dial is the lever for the initial ramp.

src/lib/telemetry/replay-scrub.ts
Jayclifford345: Follow-up on my own fix, in a951939c: the first cut captured the declaration value as \`[^;}]*\`, and both characters are legal inside a quoted CSS string — \`content: "Acme; Inc"\` ended the capture mid-string, so nothing got masked. The value pattern now consumes quoted runs before the bare class sees them, and two cases cover it. Same class of "stop at the first delimiter" mistake as the image-set one, found by writing the test rather than by reading the regex.

src/lib/telemetry/replay-scrub.ts
```diff
@@ -0,0 +1,505 @@
...
+// through the URL pass, and masking its quoted target would break the image
+// for a leak the URL pass has already closed.
+function maskCssTextValue(value: string): string {
+  return HAS_CSS_RESOURCE.test(value)
+    ? value
+    : value.replace(
+        CSS_QUOTED_STRING,
+        (_match, quote: string, text: string) => `${quote}${maskCssString(text)}${quote}`
+      );
+}
```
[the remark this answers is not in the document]
Jayclifford345: Right, and this one is on me — fixed in 1499558d.

`content: <image> / <string>` is the standard alt-text form, so a real URL and a real string live in the same value routinely; bailing out of the whole declaration on sight of a `url(` shipped the string beside it. `content: "prefix" url(x)` is the same shape from the other side.

The skip is now per-span rather than per-value: `maskCssTextValue` walks the value, steps over each `url(...)` / `image-set(...)` to its balanced close, and masks the quoted strings in between. Resource targets are still untouched — they came through the URL pass and asterisking one would break the image — but nothing else in the value rides on their presence any more.

Two cases added: alt text after a `url()`, and a string before an `image-set()`.

src/lib/telemetry/replay-scrub.ts
```diff
@@ -0,0 +1,505 @@
...
+// Every quoted string in an image-set argument list is a resource reference
+// except the MIME type in `type("image/avif")`, which a URL rewrite would
+// mangle into an absolute path.
+const CSS_IMAGE_SET_ENTRY = /(type\(\s*)?(["'])([^"']*)\2/g;
+
+// `content` and custom properties are the two declarations that can put
+// author-supplied text on screen, and rrweb masks neither: stylesheet text is
+// exempt from maskTextSelector (masking it would strip the page of styling),
+// and CSSOM writes never pass through the text path at all.
+const CSS_TEXT_DECLARATION = /(^|[^\w-])((?:content|--[\w-]+)\s*:\s*)([^;}]*)/gi;
```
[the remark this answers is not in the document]
Jayclifford345: Correct, and already fixed in a951939c — I found the same thing writing the test for the commit you reviewed, so this landed a few minutes before your comment.

`;` and `}` are both legal inside a quoted CSS string, and `[^;}]*` ended the capture mid-string so the mask had nothing complete to work on. The value pattern now consumes quoted runs before the bare class sees them: `(?:[^;}"']|"[^"]*"|'[^']*')*`. The three arms cannot match the same first character, so there is no backtracking cost. Two cases cover it, one per delimiter.

src/lib/telemetry/faro-adapter.ts
```diff
@@ -109,6 +121,66 @@ export async function initFaro(): Promise<void> {
...
+        activating = false;
+        const exhausted = attempts >= MAX_REPLAY_ACTIVATION_ATTEMPTS;
+        if (exhausted) {
+          unsubscribe();
+        }
+        pushFaroEvent(TELEMETRY_EVENTS.sessionReplayActivationFailed, { exhausted: String(exhausted) });
+      });
+  };
+  const unsubscribe = onPathfinderSurfaceChange(start);
+  start();
```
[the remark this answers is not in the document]
Jayclifford345: Good catch on the assumption — the comment asserted a coupling the code did not actually guarantee. Fixed in 1499558d.

You are right about the mechanism: `passesActivityGate` samples `isPathfinderOpen()` at item time, so the latch depends on something being pushed *during* the open window. In practice an open pushes a Faro user action through `beforeSend` synchronously, which is why this has not bitten — but "in practice" is doing the load-bearing work there, and replay is the payload most exposed to it, since its chunk is fetched on the very open that would have latched the gate.

The latch is now driven by the transition rather than sampled: `markPathfinderActive()` runs in `initFaro`'s surface listener, which fires on every report including the initial one. The lazy check in `passesActivityGate` stays for the cold-start case where a surface is already open at init and never changes.

Worth noting the scope: this is pre-existing gate behaviour rather than something replay introduced, and it applies to the whole telemetry stream, not just recordings. Two lines and a test that fails without them — a surface that opens and closes with nothing pushed in between now leaves the gate open.

Jayclifford345:
Bugbot re-reviewed the fix and found two real gaps in it, plus one I had already closed. Both fixed — a951939c and 1499558d.

- **CSS alt text beside a resource.** My first cut skipped a declaration value entirely when it contained a `url(`, which also skipped the string next to it — and `content: url(i.png) / "alt text"` is the standard alt-text form. Resource spans are now stepped over individually rather than gating the value.
- **`;` and `}` inside a quoted string.** `content: "Acme; Inc"` ended the value capture mid-string, so nothing got masked. Found independently while writing the test; the value pattern now consumes quoted runs first.
- **Activity gate is sampled, not latched.** Not a replay bug and not new, but the comment in `faro-adapter.ts` asserted a coupling the code did not guarantee: `passesActivityGate` only latches when an item happens to be processed while a surface is open. `markPathfinderActive()` now runs in the surface listener, so the transition itself latches it. Applies to the whole telemetry stream, not just recordings.

`npm run check` green — 387 suites, 6242 tests. Every new case fails against the code it replaced; I checked each rather than assuming.

src/lib/telemetry/replay-scrub.ts
```diff
@@ -0,0 +1,528 @@
...
+
+function maskCssString(text: string): string {
+  return text.replace(CSS_ESCAPE_OR_GLYPH, (token) => (token.startsWith('\\') ? token : '*'));
+}
+
+function maskQuotedStrings(value: string): string {
+  return value.replace(
+    CSS_QUOTED_STRING,
+    (_match, quote: string, text: string) => `${quote}${maskCssString(text)}${quote}`
+  );
```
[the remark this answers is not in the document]
Jayclifford345: Real, and worse than described — fixed in 4dc4230c.

You framed it as escaped quotes; the same `[^"']*` also excludes the *other* delimiter, and that case leaks more. I checked both before fixing:

| input | old | new |
|---|---|---|
| `content: "alice\"secret@acme.com"` | `"*****\"secret@acme.com"` | `"*****\"***************"` |
| `content: "it's alice@acme.com"` | *no match at all — untouched* | `"**** **************"` |

The apostrophe case is the sharper one: the pattern fails to match anywhere, so the whole string ships verbatim. And it needs no escaping to reach us — an apostrophe in rendered text is ordinary.

Your version is the one that is guaranteed to occur, though: CSSOM's "serialize a string" wraps in `"` and escapes an inner `"` as `\"`, so any `content` carrying a quote arrives in exactly that form on the `_cssText` and inserted-rule routes.

The fix is one shape applied to every quoted-string pattern in the file — `(?:\\[\s\S]|(?!\N)[^\\])*` — which is what a CSS string actually is: runs to the next *unescaped* delimiter of its own kind, holds the other kind freely. `CSS_QUOTED_STRING`, `CSS_IMPORT_PATTERN`, `CSS_IMAGE_SET_ENTRY`, the declaration-value arms of `CSS_TEXT_DECLARATION`, and the two quoted arms of `CSS_URL_PATTERN` all get it — the last one was not leaking text but would have skipped the rewrite on a URL with an escaped quote, which is a query-retention hole of the same shape. Two alternation arms that cannot match the same first character, so no backtracking. `findClosingParen` was already escape-aware and needed nothing.

Two scrubber cases plus two more secrets in the real-recorder fixture; three of the four fail against the previous commit.

Worth saying plainly: that is three delimiter bugs in this masking code across three rounds — first `)`, then `;`/`}`, now the quotes. All the same mistake in different clothes, and none of them found by reading. If a fourth turns up I would rather replace the pattern set with a small CSS value tokenizer than keep patching arms, but the three known forms are now covered and each has a test that fails without it.

leslie-heinzen (approved):

PR #1483, opened 2026-07-31
feat(docs-panel): support multiple guide editor tabs

dgehring-grafana:
## Summary

Adds first-class support for multiple guide editor tabs. Each editor owns an isolated local draft and backend binding, preserves its identity across sidebar, floating, and fullscreen surfaces, and prevents duplicate tabs from binding the same remote guide.

## What changed

- **Multi-editor tab model** — Create Guide opens a distinct editor tab; tabs can be focused, closed, restored, and handed between surfaces by stable tab ID.
- **Per-tab persistence** — Draft state and remote tracking share `pathfinder-block-editor-state:<tabId>`, with debounced saves, explicit transition flushes, dirty/status chrome, and discard confirmation.
- **Legacy recovery** — Singleton-era draft and backend-tracking keys migrate to the legacy editor tab and force that tab open so existing work is not orphaned.
- **Guide identity** — New local guides receive unique provisional IDs. The first title commit may replace the provisional ID and lock it; remote guides and saved resources remain locked. Template/import IDs avoid backend and sibling-draft collisions.
- **Surface synchronization** — Floating/fullscreen restore the complete tab workspace before applying a pending handoff. Existing handoffs focus the restored tab by ID, and the sidebar force-refreshes when ownership returns so editor changes percolate across models.
- **Remote deduplication** — Loading a library resource already owned by another editor focuses its existing tab rather than creating a second binding.

## Why

A singleton editor tab could not safely support comparing drafts, keeping a remote guide open while starting another, or moving multiple editor documents between panel surfaces. The previous handoff ordering could also overwrite the persisted strip with only the active tab. This change makes the tab workspace—not an individual surface—the durable unit.

## How to verify

1. Open Pathfinder and create two guides. Confirm each tab keeps an independent title, content, and provisional guide ID across reload.
2. Edit both guides, enter fullscreen or pop out one tab, then return. Confirm every tab remains and edits/title/status updates are preserved.
3. Save or load a remote guide, then attempt to load it from another editor tab. Confirm focus moves to the existing owner.
4. Close an editor with unsaved blocks. Confirm dismiss keeps it and discard removes its per-tab storage. Empty editors close without confirmation.
5. Seed singleton-era editor storage, mount Pathfinder, and confirm the recovered editor tab opens with its draft and remote binding.
6. Import or load a template in multiple editor tabs and confirm generated guide IDs do not collide.

## Storage and rollback

The editor storage shape changes from singleton keys to per-tab unified documents. Existing singleton draft/backend state is migrated to `pathfinder-block-editor-state:editor`; the old keys are removed after migration. Reverting this PR will not read the new per-tab keys, though they remain harmless localStorage entries.

## Out of scope

- JSON-mode buffer dirty tracking remains based on applied guide content.
- The unused WYSIWYG return-to-editor banner is unchanged.
- Tab navigation remains model/prop based rather than introducing a new global navigation context.


docs/sources/block-editor/_index.md
```diff
@@ -67,7 +67,7 @@ The fastest way to learn the editor is to build a small guide end to end.
 
 ### 1. Name your guide
 
-Click the title at the top and type a name. Press **Enter** or click away to confirm. The first time you save, the editor creates a unique guide ID by slugifying the title and appending a four-character random suffix (for example, `setting-up-dashboards-a3f9`). The ID is locked after the first save — renaming the title later does not change it.
+Click the title at the top and type a name. Press **Enter** or click away to confirm. The first time you commit a title, the editor creates a unique guide ID by slugifying the title and appending a four-character random suffix (for example, `setting-up-dashboards-a3f9`). The ID is locked after that — renaming the title later does not change it.
```
dgehring-grafana, line 70: This behavior already existed, edited to reflect that - though, templates inject ID (unchanged in this PR) and unedited "New Guide" would save to remote with that ID causing conflicts. I've made a change in this PR to mint a unique id on save if the user does not change the title on first pass. LMK if not desired.

src/components/docs-panel/components/DocsPanelContentArea.tsx
```diff
@@ -261,7 +271,7 @@ export function DocsPanelContentArea(props: DocsPanelContentAreaProps): React.Re
 
           return (
             <div className={isDocsLikeTab(activeTab.type) ? styles.docsContent : styles.journeyContent}>
-              {/* Return to Editor Banner - only shown for WYSIWYG preview */}
```
dgehring-grafana, line 264: AI flagged this WYSIWYG as stale and it does appear to be - should we remove this?

src/components/block-editor/BlockEditor.tsx
```diff
@@ -650,22 +658,6 @@ function BlockEditorInner({ initialGuide, onChange, onCopy, onDownload }: BlockE
     backendGuides.refreshGuides();
   }, [backendGuides]);
 
-  // Handle new guide with smart warning logic
```
dgehring-grafana, line 653: Removed "new guide" flow as I believe it's an artifact of the singleton pattern. Now we can create new guides while maintaining the current one at will.

Spidy88 (changes requested):
Reviewed `ui/add-multiple-guide` (6 commits, 52 files, +1796/−734) against `main`, focused on backwards compatibility, tab-id/save-data lifetime, and async collisions.

The tab model is sound. The problem is its data lifecycle: this branch has five distinct paths that destroy or strand customer authoring work, and the enabling defect — no owner for the per-tab key outside two call sites — is common to all of them. Guide drafts are hours of a customer's work with no server copy. Losing them silently is not a scope decision, it's a defect, and it's exactly the surprise our principle of least surprise exists to prevent. The findings below are ordered on that basis.

---

## Blocking — customer data loss

### 1. Upgrade discards every in-progress draft, and the UI erases the last pointer to it

The PR body documents this as intentional: `pathfinder-block-editor-state` is replaced by `pathfinder-block-editor-state:<tabId>` with no shim, so drafts under the old key are not shown after upgrade. That is a customer opening Grafana on Monday to find Friday's unsaved guide gone, with no warning, no explanation, and no recovery path in the product. Ship-blocking on its own.

It also plays out worse than "not shown." A pre-upgrade persisted editor tab has `id: 'editor'`, which `getCanonicalPersistedTabType` accepts as a non-singleton editor id (`tab-storage-restore.ts:71-76`). The tab restores, finds nothing at `pathfinder-block-editor-state:editor`, and the title effect from #8 immediately renames it "New Guide" and persists that. The customer's named guide tab becomes an empty untitled draft. The content is still on disk under the unsuffixed key, but nothing in the UI points at it any more.

The migration is small because `parseEditorStoredState` already accepts the legacy `{guide, blockIds, viewMode, savedAt, version}` document unchanged — it just yields `remote: undefined`. Rehoming the key on first editor-tab creation covers it:

```ts
// in createEditorTab, when no editor tab exists yet
const legacy = localStorage.getItem(StorageKeys.BLOCK_EDITOR_STATE);
if (legacy && !localStorage.getItem(editorTabStorageKey(tabId))) {
  localStorage.setItem(editorTabStorageKey(tabId), legacy);
  localStorage.removeItem(StorageKeys.BLOCK_EDITOR_STATE);
}
```

Also: `pathfinder-block-editor-backend-tracking` is now dead but never deleted, and its `StorageKeys` entry was removed (`storage-keys.ts`, `storage-keys.test.ts:53`), so nothing downstream can name it for cleanup. Delete both legacy keys once the migration reads them.

### 2. Gate closure deletes unsaved drafts with no confirmation and no undo

`pruneGatedTabs` (`docs-panel.tsx:288-304`) calls `clearEditorTabStorage` on every editor tab it removes. A dev-mode flip or a role downgrade to Viewer therefore permanently destroys unsaved authoring work — no prompt, no recovery. On `main`, prune left the singleton draft intact, so restoring the role restored the draft.

`closeTab` already gates the identical destruction on `editorTabHasUnsavedWork` (`docs-panel.tsx:644-651`). Prune must apply the same check and decline to delete when work is unsaved. A stranded key is a cleanup problem; a deleted draft is a customer's lost afternoon.

This also needs to be in the PR's Breaking changes section — it is a behavior regression against `main` that the description does not mention.

### 3. Prune can resurrect the key it just deleted, stranding the draft permanently

Unlike `closeTab`, prune never calls `flushEditorDraft`, and it clears storage *after* `setState` (line 295 vs 300). With a pending 1 s autosave debounce in flight:

1. `setState({ tabs: pruned.tabs })` — tab removed from model state.
2. `clearEditorTabStorage(tab.id)` — key removed.
3. React re-renders, unmounts `BlockEditor`; the unmount cleanup calls `flush()` (`useBlockPersistence.ts:161-163`).
4. `pendingSaveRef.current` is still `true` → `writeEditorDraftState` → key recreated, now with no owning tab and no path to deletion or display.

Fix by flushing before the clear, and by having `clearEditorTabStorage` also `editorDraftFlushers.delete(key)` so any later unmount flush is a no-op — that closes it for future callers too. The `discardConfirmed: true` branch of `closeTab` (`docs-panel.tsx:644`) has the same shape; it's safe today only because the earlier deferred call already flushed and reset `pendingSaveRef`, and no test pins that cross-call invariant.

### 4. Two tabs can bind the same backend resource, and the later save silently overwrites the earlier

This one destroys *saved* library content, not just local drafts. `focusEditorTabForResource` guards the library-load path (`BlockEditor.tsx:637-640`); `orchestrateSave` has no equivalent.

1. Tab A loads library guide `my-guide` → `remote.resourceName = 'my-guide'`.
2. Tab B is a fresh draft whose title slugifies to `my-guide` (or whose id is set directly), and it saves.
3. The overwrite-confirm modal fires (`useBackendSaveFlow.ts:283`) naming the *library* guide — nothing indicates a sibling tab already holds it.
4. Confirm binds B to `my-guide` as well. A's `lastSyncedJson` is now stale.
5. A saves later and overwrites B's work. No conflict detection, no `resourceVersion` check, no notification.

`findEditorTabIdByResourceName` already answers this in one call. Route the overwrite-confirm through it so the modal can say the guide is open in another tab, and refuse the silent second binding.

### 5. Cross-device: the tab list is server-synced, the drafts are not

`tabStorage` goes through `createUserStorage()` (`user-storage.ts:258`) — the Grafana user-storage backend, synced server-side across devices, with localStorage as a write-through cache. Editor drafts go to raw `localStorage` under `pathfinder-block-editor-state:<tabId>` (`editor-tab-storage.ts:39`). Different scope, different lifetime, no reconciliation.

- **Second browser or machine.** The strip restores editor tab `tab-1754…-a3f9` from the server; the draft key doesn't exist locally, so the editor opens empty. `onGuideTitleChange` then fires with the mount-time default, rewrites the tab title to "New Guide", and `saveTabsToStorage()` pushes that to the server — overwriting the real title on the original machine.
- **Closing that tab on machine B.** `clearEditorTabStorage` removes B's nonexistent key. Machine A's real draft is unreachable: no tab points at it, nothing will delete it.

The singleton key made this survivable — one tab, one always-reused key, so divergence showed a stale draft at worst. Per-tab keys turn every divergence into a loss. Either move the draft into `tabStorage` alongside the tab record, or make drafts explicitly device-local and have the restore path reconcile (#7). What can't stand is a server-synced tab list advertising drafts the local store doesn't have.

### 6. Handoff-created tab can be wiped, with its draft, by the restore that follows it

`FullScreenPanel.tsx:99/134` and `FloatingPanelManager.tsx:143/188` both run `consumePendingGuideOnMount` before `restoreTabsAsync`, and the restore gate is `hasOnlyNonContentTabs(...)`, which counts `editor` as non-content. The handoff creates editor tab `X`; `restoreTabsAsync` then overwrites `tabs` wholesale with the storage snapshot (`docs-panel.tsx:256`).

This survives only because the source surface persisted `X` first — via `setActiveTab` firing `saveTabsToStorage()` unawaited against an async user-storage write. Pop out immediately after creating a tab and the restore reads the pre-`X` snapshot, drops the tab, and strands its key. On `main` the same race was harmless: the constant `'editor'` id meant the restored tab and the dropped tab shared a draft key.

Merge rather than replace in `restoreTabsAsync` (preserve in-memory editor tabs absent from the snapshot), or await the source-side persist before `locationService.push`.

### 7. No orphan reconciliation — the defect underneath #1, #3, #5, and #6

`clearEditorTabStorage` has two production call sites: `closeTab` (`docs-panel.tsx:671`) and `pruneGatedTabs` (`docs-panel.tsx:300`). Every other way an editor tab leaves the list strands its key:

- `restoreTabsAsync` replaces wholesale (`docs-panel.tsx:256`).
- `restoreTabsFromStorage` drops entries failing `getCanonicalPersistedTabType`.
- `tabStorage.setTabs` truncates to the last 50 tabs (`user-storage.ts:877`), or the last 25 on `QuotaExceededError`. Tabs sliced off the front keep their drafts.
- Reset of the server-synced tab list, plus #5.

Nothing enumerates localStorage for the `pathfinder-block-editor-state:` prefix. Guides with many blocks are not small, and this shares the ~5 MB origin budget with the interactive-step prefixes in `user-storage.ts` — so the eventual `QuotaExceededError` will most likely surface as a failure in an unrelated subsystem, including progress tracking.

Add a reconciliation pass in `restoreTabsAsync` once the final tab set is known: enumerate keys under `${StorageKeys.BLOCK_EDITOR_STATE}:` and drop any whose suffix isn't in the restored editor-tab id set. That is also the natural home for #1's migration and closes the stranded-key half of #3, #5, and #6.

---

## High

### 8. Every editor mount transiently renames its own tab to "New Guide"

The title-sync effect (`BlockEditor.tsx:248-250`) is declared above `useBlockPersistence` (`BlockEditor.tsx:568`), so on mount it runs first with `state.guide.title` still at `DEFAULT_GUIDE_METADATA`. It calls `updateEditorTabTitle(tabId, 'New Guide')` → tab mutation → `saveTabsToStorage()` → server write. The restore effect then loads the real title and triggers a second update and second write.

On its own that's a flicker and a redundant write on every tab switch and surface handoff. It stops being cosmetic when the restore produces nothing: it is the mechanism that makes #1 erase the pointer to the legacy draft and makes #5's cross-device title clobber destructive rather than merely wrong-looking. Guard on the restore having settled, or skip the update while the title still equals the default.

### 9. Tab-strip chrome re-parses every draft on every autosave

`notifyEditorTabChromeListeners()` fires from `writeRaw` on every draft write (`editor-tab-storage.ts:120,124`) — roughly once per second while typing. The single `useSyncExternalStore` in `DocsPanelTabBar` re-renders the whole strip, and each editor tab calls `getEditorTabChromeStatus` (`editor-tab-storage.ts:194`): `localStorage.getItem` + `JSON.parse` of the entire guide document + a `JSON.stringify` of `{id,title,blocks}` to compare. Rendered twice per tab, strip and overflow dropdown.

At 4–5 open editor tabs with a large guide that's ~10 full parse/stringify round-trips per second of synchronous main-thread work on the typing path. Cache status per key with write invalidation, or store a precomputed `dirty`/`status` pair in the document.

Related: `useBackendSaveFlow` calls `readEditorStoredState(storageKey)` during render (`useBackendSaveFlow.ts:161`) whenever the guides list hasn't resolved the current resource — another full-document parse on the render path.

---

## Medium

### 10. `confirmPendingClose` can strand `pendingCloseTabId`

`closeTab` returns early at `docs-panel.tsx:640` when the tab isn't found — before the `pendingCloseTabId` reset at line 677. If the pending tab disappears while the modal is open (prune, a competing restore, another surface), `confirmPendingClose()` no-ops and leaves `pendingCloseTabId` set. The modal hides because `pendingCloseTab` resolves to `null`, so it's invisible rather than broken, but the state stays stuck. Clear the id in the early-return branch.

---

## Low

- **Tab badge lost its i18n wrapper.** The strip badge moved from `t('docsPanel.editorDraftBadge', 'Draft')` to literal strings in `editorStatusBadge` (`editor-tab-storage.ts:139-186`), text and tooltips both. With 21 locales in the tree these belong back at the component layer.
- **`useBlockPersistence.clear()` is unreferenced** in production after the New-guide removal, and does a plain `removeItem` (`useBlockPersistence.ts:145`) that drops `remote` along with the draft — inconsistent with the merge-write discipline elsewhere. Delete it or make it symmetric before it gets picked back up.
- **Docs left behind by the "New guide" removal.** `docs/developer/components/block-editor/README.md:159` still says `Click "New Guide"`; `docs/developer/CUSTOM_GUIDES.md:65` still tells users to "start a new guide" to resolve an id collision. The kebab item is gone (`HeaderKebab.tsx`). `/prevent-doc-drift` in the same PR catches these.
- **Removing "New guide" is a user-facing change** not listed under Breaking changes.

---

## Verified clean

Areas I went after specifically and found correct — no second look needed:

- **Draft/remote write isolation.** `writeEditorDraftState` and `writeEditorRemoteState` each read-modify-write and preserve the other's half. I traced autosave-during-save and save-during-autosave; neither wipes the other. `writeRaw`'s empty-document → `removeItem` branch is correct.
- **Dirty-baseline stability across load.** `guideSyncSnapshot` normalizes key order, and `loadGuide`/`getGuide` (`useBlockEditor.ts:1144-1178`) pass block objects by reference without normalization, so a freshly loaded library guide reads clean rather than immediately "(modified)". This was the failure mode I most expected to find; it isn't there.
- **Singleton id reservation.** `getCanonicalPersistedTabType` keeps `SINGLETON_TAB_IDS` reserved in both directions while opening editor ids up; a malformed content record still can't acquire privileged behavior.
- **Normal close/discard flush ordering.** `closeTab` flushes before the dirty check and before clearing, and `pendingSaveRef` is correctly false by the time the editor unmounts. #3's failure mode does not reach this path.
- **Tab-id generation.** `tab-${Date.now()}-${random}` with same-tick synchronous Scenes `setState` — double-invoking `createEditorTab` yields distinct ids and both survive.
- **Test coverage of intended paths.** `docs-panel.tab-restore-guard.test.ts` and the close-tab discard tests cover what they set out to. The gaps map exactly onto the unintended paths above: orphan reconciliation, prune-with-pending-flush, and handoff/restore ordering have none.

---

**Blocking merge:** #1 through #7. Each one loses or strands work the customer cannot get back, and #4 destroys content already saved to the library. #7 is the cheapest of them and makes #1 and #5 substantially easier, so it's the sensible starting point. #8 should go with them since it's the amplifier on #1 and #5.

dgehring-grafana:
@Spidy88 


#1 Fixed via the addition of a legacy migration flow, found in `panel.recoverLegacyEditorTab();` that runs on `useTabRestoration` - it's gross but hopefully easily removable after the initial gaps goes away. It also does have the side effect of removing the original content body and relocating it to the new one, though I think this is probably good practice.

#2 Somewhat intentional? This mirrors original behavior where the permissions gate would discard in-progress work. This PR is massive so even though I noticed this, I did er on the side of leaving OG behavior in and hopefully fixing it and/or improving it later.

#3 Added a behavior to the gate prune that doubles down on the above, prune will flush all editor drafts and remove them in a loop. Again, still considering architecture to allow recovery, but this matches current behavior.

#4 Noticed and this concerned me - I realized that internal ID should be tracked as an actual identifier and not just a stub, so I made it unique by default (i.e. `new-guide-xxx`) and track if it's a locked ID with a flag `idIsFinalized` so the user can still edit it initially via a title update, same as before. Annoyingly, I can't use prefixing to check this condition anymore due to the possibility they use "New Guide 2" or something, thus the flag. That removes the way I imagine this mistake could be made in 99 percent of cases. Unfortunately, what I didn't think of is that we can edit the ID with JSON. Conflict in remote already has an overwrite notice, attempting to open from remote with an open guide of the same id just navigates to the editor.

Addressing with some hooks on the JSON field that detect ID changes. They check for any other tabs in local that have the same ID, and warn the user before committing. If they do - it will discard the old version in local and change the ID so no collision can happen.

#6 Fixed with new flush/restore flow

#8 Mostly fixed, but issue 5 might mess it up

The rest are unaddressed, plan on getting to them at a later point and maybe breaking this PR into something more.... modular.

Jayclifford345 (commented):
## `/review` — routed PR review

**Classification:** `mixed` (product-runtime + contracts-and-schemas). **Coverage confidence:** high.
**Reviewers run:** security, correctness-and-reliability, testing-and-verification, reversibility-and-one-way-door, block-editor-authoring, floating-panel + surface-handoff, guide-schema-and-contracts, docs-retrieval-and-rendering, performance-and-bundle, plus tech-debt and doc-drift passes and an adversarial verification round.

**Verification:** `npm run test:ci --coverage=false` → **397 suites / 6322 passed**, 18 skipped, 5 todo. `tsc --noEmit` → clean. No `.skip`/`.todo` introduced.

**Disposition: comment.** This is well-built for its size — the per-tab document with merge helpers, the restore-before-handoff inversion, and the migration's write-before-delete ordering are all right. Two things I'd want fixed before this leaves draft (inline), then a tier of medium items, then nits.

Full inline comments below. The rest, grouped:

### Also worth fixing

- **No GC for orphaned `pathfinder-block-editor-state:<tabId>` keys.** Cleanup fires only on `closeTab` and the gate-close prune. A restore-rejected record, `clearPersistedTabs()`, the `MAX_PERSISTED_TABS: 50` slice in `user-storage.ts:877`, and a revert all drop tabs without touching their keys. Orphans are load-bearing too: `editorGuideIdExists` scans the whole prefix regardless of ownership, so an invisible orphan reserves its guide slug forever. A sweep of `${BLOCK_EDITOR_STATE}:*` against the restored tab-id set at the end of restore would close both.
- **Migration drops `backendStatus`.** `parseLegacyBackendTracking` copies `resourceName` + `lastPublishedJson` but not `backendStatus`, which main did persist. Confirmed still open — this is [the unresolved cursor[bot] thread](https://github.com/grafana/grafana-pathfinder-app/pull/1483#discussion_r3716904774), not a new finding. Worth adding: the header self-corrects once `backendGuides.guides` loads (`useBackendSaveFlow.ts:159-163` prefers the live list), but the strip badge reads storage only, so the two disagree until the next save.
- **Test coverage gaps tied directly to the new semantics.** Three, all cheap: (1) the remote-dedup invariant is the early return in `BlockEditor.handleLoadFromLibrary` — delete it and every test still passes while two tabs bind one `resourceName`; a repo-wide grep for `onFocusExistingGuide|onFindGuideIdCollision|onDiscardGuideIdCollision` in tests returns zero. (2) The collision-confirm path is destructive (it discards another tab's draft) and nothing asserts confirm-discards / cancel-doesn't / unmount-resolves-false. (3) The five `flushEditorDraft`-before-handoff call sites have no coverage of any kind — unmount and close-tab flushes are both covered, handoff isn't, and handoff is the one that exists because the receiving surface can read storage before the outgoing one unmounts.

### Nits and smaller items

- **`'editor'` lost its reserved-ID protection, and the test that pinned it was deleted.** `SINGLETON_TAB_IDS` no longer contains `'editor'`, but `'editor'` is still `LEGACY_EDITOR_TAB_ID`. A persisted `{id:'editor', type:'docs'}` now restores; `recoverLegacyEditorTab`'s `hasLegacyChrome` and `createEditorTab`'s existence check are both type-qualified, so a second tab with the same id gets pushed — and `closeTab`'s `filter(t => t.id !== tabId)` then removes both. Tamper-only, so low, but the deleted "Disguised editor" case is exactly what would have caught it.
- **QC4 — `collectKeysByPrefix` already exists.** `editorGuideIdExists` hand-rolls the `for i < localStorage.length` → `key(i)` → `startsWith` sweep that `src/lib/storage/key-utils.ts:25` was written to consolidate; its own docblock says it replaced 7+ copies. This is the eighth.
- **`migrateLegacyEditorTabStorage()` runs on every `readEditorStoredState`** with no latch, so every read pays two `getItem` probes for keys that are permanently absent after the first run. It also means a storage write can happen during React's render phase (reachable from `getEditorTabChromeStatus` in `DocsPanelTabBar`'s render body and from `readInitialGuideIdFinalization` in a `useState` initialiser). A module-level `let migrated = false`, or dropping the call and relying on the explicit one in `recoverLegacyEditorTab`, fixes both.
- **i18n regression.** The strip badge was `t('docsPanel.editorDraftBadge', 'Draft')`; it now renders the hardcoded English from `editorTabStatusBadge`. (The header side was already untranslated, so this is levelling down rather than a new class of problem.) Separately, four new `t()` keys — `docsPanel.editorTabModified`, `discardDraftTitle`, `discardDraftBody`, `discardDraftConfirm` — were never extracted into `src/locales/`.
- **`pruneGatedTabs` flushes a pending draft and then deletes it** on the next line (`docs-panel.tsx:319-320`). The flush is unreachable work and reads as a preservation step, which makes it easy to miss that this is the one editor-tab teardown path with no discard confirmation.
- **Legacy recovery can steal the restored active tab.** When the legacy tab already exists in the strip, `createEditorTab` falls through to `setActiveTab(existing.id)` and `saveTabsToStorage()`, overwriting the `activeTabId` that `restoreActiveTabFromStorage` just resolved — persisted, on first load after upgrade.
- **A stale `pendingGuide` now carries `tabId`.** There's no TTL and no clear-on-mode-change, and the sidebar never consumes pending guides. A handoff abandoned by browser-Back survives in memory; a later unrelated pop-out consumes it and `setActiveTab(staleTabId)` silently steals focus. Pre-PR the same stale record produced a visible duplicate tab — self-explaining. Clearing `_pendingGuide` in `applyModeChange` when the committed mode is `'sidebar'` is ~3 lines.
- **N editor tabs all named "New Guide"** give every close button the accessible name "Close New Guide". Distinguishing tabs non-visually is a first-class part of this feature, not polish. Also `docs-panel.tsx:811,861` and `tab-storage-restore.ts:145` say `'New Guide'` while `block-editor/constants.ts:222` says `'New guide'` — sentence case per AGENTS.md, and reading all three from `DEFAULT_GUIDE_METADATA.title` stops the drift.
- **Hardcoded `color: '#888'`** added at `BlockEditor.tsx:1002` (copied from the pre-existing one at `:1023`). Fails contrast on the light theme; `theme.colors.text.secondary` via `useStyles2`.
- **No forward-compat tolerance.** `parseEditorStoredState` rebuilds from a fixed allowlist and every writer is read-modify-write over the stripped object, so any field a newer build adds is destroyed by the next autosave from an older tab. An `...unknownRest` passthrough, or refusing to write over a higher `version`, would cover it.
- **`open_editor` kept its name but changed what it counts** (`TabBarActions.tsx:123`) — it used to conflate create-with-focus; re-focusing now emits nothing. Any dashboard trending it sees an unannounced step change at deploy. A distinct `create_editor_tab` value keeps the series separable.
- **`tab-kinds.ts` lost `Set<LearningJourneyTabType>`** on both new sets (the type-only import was dropped), so a typo in `'devtools'` is no longer a compile error — and `NON_CONTENT_TAB_KINDS` is exported with zero consumers outside its own file.
- **Duplicated assertion** — `useModalManager.test.ts:13-14` asserts `isOpen('import')` twice; the first line replaced the removed `'newGuideConfirm'` case.
- **z-index** moved `navbarFixed` (1000) → `dropdown` (1030) in `docs-panel.styles.ts:231`. The comment justifies clearing the editor header, but the value also clears Grafana's `sidemenu` (1020) and ties `typeahead`. `navbarFixed + 1` matches the stated intent.
- **QC8** — a handful of new JSDoc lines restate their signature (`types.ts:80,122,128`, `docs-panel.tsx:680,859`). The load-bearing comments in this diff are genuinely good and should stay: `editor-tab-storage.ts:5-11`, `DocsPanelContentArea.tsx:171-172`, and the stacking-context notes in `docs-panel.styles.ts`.

### Tech debt

Non-blocking, from a scan scoped to the changed files:

- **B1 — the discard-draft `ConfirmModal` is triplicated** across `docs-panel.tsx:1507`, `FloatingPanelManager.tsx:415`, `FullScreenPanel.tsx:449`, character-identical apart from the model variable. Three copies of a destructive confirmation is a drift hazard.
- **B1 — the `<BlockEditor>` surface wiring is triplicated** too (`DocsPanelContentArea.tsx:172`, `FloatingPanelManager.tsx:387`, `FullScreenPanel.tsx:424`). `BlockEditorProps` went 4 → 9 props in this PR; every future prop now has to land in three places or a surface silently loses the behavior. A `useEditorTabWiring(model, tab)` hook returning the bundle would collapse it.
- **A2 — `docs-panel.tsx` 1494 → 1609 LOC, `BlockEditor.tsx` 939 → 1049.** Both were already oversized. The eight new editor-tab model methods are a cohesive slice (`EditorTabController`), and `useGuideIdFinalization(storageKey, backendGuides)` would lift the id state machine out of `BlockEditorInner`.
- **E4 — `editorDraftFlushers` is an exported mutable `Map`** with no owner, no `clear()`, and its only writer in another file's effect; the tests reach in and hand-`set`/`delete` entries. A `registerEditorDraftFlusher(key, fn): () => void` returning its own disposer keeps it private. (The `chromeVersion`/`chromeListeners` pair next to it is the canonical `useSyncExternalStore` shape — that half is fine.)
- **Bundle boundary is intact but only a comment enforces it.** All five new static imports use the deep path `block-editor/editor-tab-storage` rather than the barrel, so the `lazy(() => import('../../block-editor'))` boundary holds and the module's only import is `StorageKeys`. But that invariant lives in a docblock; one future import inside `editor-tab-storage.ts` drags the editor into the main chunk. Moving it to Tier 1 (`src/lib/storage/`) — which is where a cross-surface storage + listener registry belongs anyway — or adding a boundary tripwire would make it structural.
- Credit where due: this PR **deletes** `resetGuide`, `handleNewGuide`, the new-guide confirm modal, `BACKEND_TRACKING_STORAGE_KEY`, `hasOnlyNonContentTabs`, `getKeyboardCycleTabs`, and the `newGuideButton` test ID rather than orphaning them, and consolidates the duplicated `slugifyTitle`/`generateUniqueId` pair. Every removal was checked for dangling references — there are none in `src/`, `tests/`, or `docs/`.

### Documentation drift

`docs/sources/block-editor/_index.md` got one line, but the user-facing docs still describe a singleton Editor tab that no longer exists — `:33`, `:44`, `:46`, `:222` ("### The Editor tab isn't visible") plus `administrators-reference/_index.md:47` and `upgrade-notes/_index.md:31`. The troubleshooting section now tells users to look for a tab that isn't there. The anatomy table's **New** row describes the kebab item this PR deletes, and nothing documents that closing an editor tab discards its local draft.

Developer docs needing the same treatment: `docs/developer/CUSTOM_GUIDES.md` (entry point is now **More options** → **Create guide**; per-tab document; backend tracking moved under `remote`), `docs/developer/components/block-editor/README.md` (per-tab documents + legacy migration; the "New Guide" step), and `docs/developer/components/docs-panel/README.md` (`pendingCloseTabId` on `CombinedPanelState`, plus the six new model methods).

One flag on `docs/design/CONCERNS.md` itself: `block-editor-authoring`'s `one_way_doors` names `BLOCK_EDITOR_STORAGE_KEY` — a symbol that doesn't exist (it's `StorageKeys.BLOCK_EDITOR_STATE`) — and this PR *is* the shape change that entry was guarding. Its `trigger_paths` also only cover `src/components/block-editor/**`, while the editor-tab lifecycle now lives in `src/components/docs-panel/**`; those paths routed here only through `reversibility-and-one-way-door`. Worth an owner edit.

I'd skip one doc-drift suggestion that came out of that pass: it proposed canonising the `filter({ hasText: 'Draft' })` selector into `E2E_TESTING_CONTRACT.md`. Two other reviewers independently flagged that selector as the problem — don't document it, replace it.

### Instrumentation

New behavior is largely covered by existing channels. One gap and one wrinkle: `createEditorTab` has no `reportAppInteraction`, so the headline "how many guides do people edit at once" question isn't answerable; and the `open_editor` denominator change above should be handled as a rename rather than silently.

### Not found

No F1–F6 security findings. The diff adds no DOM sinks, no raw-HTML rendering, no URL construction, and no new security-rule `eslint-disable`. `editorTabStorageKey` always emits a fixed prefix so a crafted tab id can't reach another namespace; `slugifyTitle`/`toResourceName` collapse to `[a-z0-9-]` and `itemUrl` wraps in `encodeURIComponent`, so no path traversal or resource-name spoofing from a crafted title. Import now regenerates the guide id, which actually removes a pre-existing way for an imported JSON to claim an existing backend resource name.

### Findings dropped in adversarial verification

Recorded so a human can re-open them if they disagree:

- **"`idIsFinalized` desyncs on undo and strands the guide on a provisional id."** The desync is real — the title commit is two `updateGuideMetadata` calls with no `coalesceKey`, so it's two undo entries, and undo restores `guide.id` while `idIsFinalized` stays `true` in state and storage. But the consequence collapses: undo reverts the id to exactly `DEFAULT_GUIDE_METADATA.id`, which is the value `orchestrateSave` re-mints on (`useBackendSaveFlow.ts:259-265`), so no provisional id reaches a backend resource name. Residual is a nit — the first Ctrl+Z after a title commit looks like it did nothing, and the preview-progress key stays `block-editor://preview/new-guide`. Worth a shared `coalesceKey`.
- **"Revert-then-roll-forward destroys work via the unconditional legacy-key delete."** Kept below at reduced severity rather than dropped — see the inline comment.
- **"Migrated `lastSyncedJson` never matches the new `guideSyncSnapshot` baseline."** My own; wrong. `getGuide()` returns `{id, title, blocks}` in exactly that key order on both branches, so the migrated baseline compares byte-equal.

src/components/block-editor/BlockEditor.tsx
```diff
@@ -253,8 +278,18 @@ function BlockEditorInner({ initialGuide, onChange, onCopy, onDownload }: BlockE
     setGuideLintResult(guideLint);
   }, [guideLint, setGuideLintResult]);
 
-  // Modal state - useModalManager handles metadata, newGuideConfirm, import, githubPr, tour
+  // Keep the owning editor tab's title in sync with the working guide.
+  useEffect(() => {
+    if (!initialGuide && !hasLoadedFromStorage.current && readEditorStoredState(storageKey)?.guide) {
```
Jayclifford345, line 283: [blocking] **A newly created editor tab's strip title stops updating ~1 second after it opens.**

`hasLoadedFromStorage.current` is set in exactly one place — inside `onLoad` (`:669-670`) — and `onLoad` only fires when storage already held a guide *at mount*. For a brand-new tab it is `false` for the component's entire life. So once the debounced autosave has written the blank draft (`AUTO_SAVE_DELAY = 1000`), `readEditorStoredState(storageKey)?.guide` is truthy and this guard short-circuits on every subsequent run.

Repro: create an editor tab → wait ~1s → type a title and blur → `onGuideTitleChange` is called **zero** times, `model.updateEditorTabTitle` never runs, and the strip still reads "New Guide". It only recovers after switching away and back (that remount takes the `onLoad` path and flips the ref).

This lands on the headline feature — two or three new drafts are indistinguishable in the strip, which is the thing multi-tab is for.

Suggest making the suppression one-shot rather than a standing predicate, e.g. capture `const hadStoredGuideAtMount = useRef(Boolean(readEditorStoredState(storageKey)?.guide))` once and gate on `hadStoredGuideAtMount.current && !hasLoadedFromStorage.current`, so it clears after the restore pass instead of re-arming as soon as autosave writes.

Two smaller things in the same effect: `onGuideTitleChange` is an inline lambda at all three call sites (`DocsPanelContentArea.tsx:176`, `FloatingPanelManager.tsx:391`, `FullScreenPanel.tsx:428`), so this re-runs on every parent render — worth memoizing. And `BlockEditor.persistence.test.tsx:112` covers only the restore direction; the create-then-rename direction is what's missing.

src/components/block-editor/editor-tab-storage.ts
```diff
@@ -0,0 +1,440 @@
...
+  try {
+    const state = parseEditorStoredState(localStorage.getItem(editorTabStorageKey(LEGACY_EDITOR_TAB_ID)));
+    return state?.guide !== undefined || state?.remote !== undefined;
+  } catch {
+    return false;
+  }
+}
+
+/** Pending debounced draft writers, keyed by storage key. */
+export const editorDraftFlushers = new Map<string, () => void>();
```
Jayclifford345, line 124: [blocking] **Recording state is still a process-wide singleton while the editor went N-instance — and this PR removed the one path that cleared it.**

`RECORDING_STATE_STORAGE_KEY` (`constants.ts:215`) is one unscoped key, and `useRecordingPersistence` takes no tab id — its mount effect restores whatever is there, guarded only by a per-instance `hasRestoredRef`. Since `DocsPanelContentArea.tsx:174` sets `key={activeTab.id}`, every editor-tab switch fully remounts `BlockEditor` and re-runs that restore.

Failure path:
1. Tab A: author starts recording into section `block-…-a3f`. Recording deliberately spans navigation, so the session is persisted.
2. Author switches to tab B (or closes A — `closeTab` clears only A's per-tab document; the recording key survives).
3. B mounts, adopts the orphaned session, and `BlockEditor.tsx:355-372` calls `actionRecorder.startRecording()` — the overlay reappears over an unrelated guide.
4. Author stops in B → `addBlockToSection(block, 'block-…-a3f')` → `useBlockEditor.ts:454` `if (sectionIndex === -1) { return prev; }`. Every recorded step is silently dropped, and `onClear()` then deletes the session, so returning to A recovers nothing.

This is a regression, not pre-existing: before, the singleton editor always re-adopted its own recording (the fixed `EDITOR_TAB_ID` plus restore-side validation guaranteed it), and `handleNewGuide` — deleted in this PR — called `recordingPersistence.clear()`.

Cheapest fix is to scope the recording document to the owning tab the same way the draft was scoped — derive it from the `storageKey` the editor already receives — and delete it inside `clearEditorTabStorage(tabId)` so close, gate-prune, and collision-discard all clear it as one unit. No migration needed; a recording session is transient, so the legacy key can just be removed alongside the others. Falling short of that, at minimum stamp the owning tab id into the payload and ignore a mismatch on restore.

src/components/docs-panel/hooks/useFullScreenHandoff.ts
```diff
@@ -86,8 +87,10 @@ export function useFullScreenHandoff(model: FullScreenModel, isSessionActive: bo
         });
         // Remember where we came from so explicit Exit can land back on the
         // user's prior Grafana page instead of the plugin home.
+        const saveTabs = model.saveTabsToStorage();
```
Jayclifford345, line 90: [suggestion] **Two of the surface-exit paths cede editor ownership without `flushEditorDraft`.**

This PR establishes "flush the debounced draft → persist → flip the mode" as the handoff invariant and applies it to five exits (`FloatingPanelManager.tsx:243,276,296`, `FullScreenPanel.tsx:247,276`). The two sidebar-origin exits don't: this one, and `usePopOutHandoff.ts:54-61` — both call `saveTabsToStorage()` and `setModePersisted(...)` with no flush.

The write still happens today, but only as a by-product of the sidebar `BlockEditor`'s unmount cleanup, which has to commit before the receiving surface's `initializePanelTabsOnMount` → `restoreTabsAsync` → mount-time `readEditorStoredState`. Those are two independent React roots (`module.tsx` renders the floating manager into its own `createCompatRoot`) with independent update queues, so nothing orders them. If the receiving root commits first, the new editor hydrates the pre-keystroke draft, the user keeps typing from stale content, and its own autosave writes that lineage back over the flushed newer one — up to 1s of typing, on every pop-out or fullscreen entry.

Two one-line additions match the five paths that already do it. `full-screen-autodock.ts:113,127-131` has the same gap and additionally flips mode inside a `setTimeout(0)` with no `saveTabsToStorage()` — its comment at `:124-126` ("All surfaces share `tabStorage`") predates the per-tab draft key.

Worth hoisting the pair into a `cedeSurfaceOwnership(model, activeTab)` helper so a seventh exit can't be added without it. Two independent reviewers landed on this one, and it currently has no test coverage of any kind.

src/components/block-editor/editor-tab-storage.ts
```diff
@@ -0,0 +1,440 @@
...
+    if (!existing) {
+      const draft = parseEditorStoredState(legacyDraftRaw) ?? {};
+      const remote = draft.remote ?? parseLegacyBackendTracking(legacyTrackingRaw);
+      const migrated: EditorTabStoredState = remote ? { ...draft, remote } : { ...draft };
+      if (migrated.guide !== undefined || migrated.remote !== undefined) {
+        writeRaw(targetKey, migrated);
+      }
+    }
+
+    localStorage.removeItem(LEGACY_BLOCK_EDITOR_STATE_KEY);
```
Jayclifford345, line 103: [question] **The rollback note in the PR description understates this.**

The write-before-delete ordering is right — a `setItem` quota throw propagates to the outer `catch` at `:106` and these `removeItem` calls are skipped. Two narrower things:

**1. The PR says the stranded per-tab keys "remain harmless localStorage entries." After migration they're the user's only copy.** Migration fires from `recoverLegacyEditorTab` (called by every owning surface on mount) and from every `readEditorStoredState` — so it runs on plugin load, not on editor open. A user who never opens the editor still gets their singleton draft moved and the bare key deleted. If the release is then rolled back, they open the editor and see a blank "New Guide"; the content is in localStorage but no shipped build reads it. Not deleted, but not recoverable without DevTools either.

**2. When the target already exists, the legacy key is deleted without merging.** That's deliberate and test-locked (`editor-tab-storage.test.ts:114`), and I'll grant the counterargument: reaching it needs an old bundle to have rewritten the bare key, which means a rollback window in which that build's editor was already blank. So this doesn't *create* the loss so much as decline to rescue work created inside an already-broken window.

Both collapse to one option, if a rollback is plausible for this release: **don't delete the legacy keys in this PR.** Copy-and-leave, delete in a follow-up once rollback is off the table. That makes the whole change cleanly reversible for a one-line cost.

Two smaller notes on the same function: a legacy draft that fails `JSON.parse` takes the `?? {}` path, writes nothing, and is then deleted anyway — partially salvageable bytes gone. And whichever way you go, the PR description's rollback section should say what actually happens.

src/components/docs-panel/components/DocsPanelTabBar.tsx
```diff
@@ -99,23 +115,28 @@ export function DocsPanelTabBar({
 
       <div className={styles.tabList} ref={tabListRef} data-testid={testIds.docsPanel.tabList}>
         {guideTabs.map((tab) => {
+          const editorStatus = tab.type === 'editor' ? getEditorTabChromeStatus(tab.id) : null;
```
Jayclifford345, line 118: [react] **R18 — the strip re-reads, re-parses, and re-serialises every editor tab's whole document in the render body, at roughly 1 Hz while typing.**

`getEditorTabChromeStatus` does `getItem` + `JSON.parse` of the entire per-tab document (which, since `remote` now shares the draft key, is the whole guide — plus `jsonModeState.json` and `.originalJson`, two more full serialized copies, whenever the tab is in JSON view), then `guideSyncSnapshot(guide)` → `JSON.stringify({id,title,blocks})`. It's called once per editor tab here and again at `:216` for the overflow list, unmemoized. Add the two probe `getItem`s from the unlatched migration and it's 3 `getItem` + a full parse + a full stringify, per editor tab, per render.

The `useSyncExternalStore` at `:95` sets the floor: `chromeVersion` is a single global counter, so a debounced autosave in *any* tab (`writeRaw` → `notifyEditorTabChromeListeners`) re-renders the whole strip and re-runs the entire sweep. Five tabs holding ~200 KB guides is ~2 MB of synchronous JSON work per second, on the keystroke path, to produce one badge and one boolean per tab.

Pre-PR this spot was `<Badge text={t('docsPanel.editorDraftBadge', 'Draft')} …>` — a prop check with zero I/O.

The `chromeVersion` counter is already the exact invalidation signal, so memoizing `getEditorTabChromeStatus` behind a `Map<storageKey, {version, status}>` is a contained fix with no API or storage-format change. Better still, compute `hasUnsyncedChanges` at write time and persist the boolean, so the read side never re-serializes. Making the notification per-key would also stop tab A's autosave from invalidating B–E.

(Same shape, lower frequency: `useBackendSaveFlow.ts:161` calls `readEditorStoredState` in the hook body on every render whenever the backend guide list hasn't resolved — which is a steady state, not a transient, if the backend is unavailable or the bound resource was deleted server-side.)

src/components/block-editor/hooks/useGuideOperations.test.ts
```diff
@@ -0,0 +1,59 @@
...
+    })
+  );
+
+  return { result, loadGuide };
+}
+
+describe('generateUniqueId', () => {
+  beforeEach(() => localStorage.clear());
+
+  it('avoids ids already present in existingNames', () => {
```
Jayclifford345, line 28: [suggestion] **This test passes whether or not the guard it's testing exists.**

30 draws of a 4-char base36 suffix come from a ~1.68M space, so `new Set(existing).size === 30` holds with probability ~0.9997 regardless of `existingNames`. Verified by mutation: deleting `!existingNames.includes(candidate) &&` from `generateUniqueId` leaves all three tests green.

The sibling-draft test right below it is the good one — it mocks `Math.random` twice and asserts both calls. Mirroring that technique for `existingNames` would pin the branch that actually matters: the one stopping a fresh draft from minting an id that overwrites an existing published library guide on first save.

Also uncovered: the 20-iteration exhaustion fallback `` `${base}-${Date.now().toString(36).slice(-6)}` ``, which performs no uniqueness check at all and is the one path that can return a colliding id.

tests/helpers/block-editor.helpers.ts
```diff
@@ -55,8 +55,9 @@ export async function openBlockEditor(page: Page): Promise<void> {
   await moreOptions.click();
   await page.getByRole('menuitem', { name: /create guide/i }).click();
 
-  // Wait for the editor strip tab and block editor content
-  const editorTab = page.getByTestId(testIds.docsPanel.tab('editor'));
+  // Editor tabs use generateTabId() (`tab-<ts>-<rand>`), not a fixed `editor` id.
```
Jayclifford345, line 58: [suggestion] **This locator can't detect the bug class the PR introduces, and it's ambiguous by construction.**

Three separate problems:

1. **The prefix matches containers.** `docs-panel-tab-bar`, `docs-panel-tab-list`, `docs-panel-tab-dropdown`, `docs-panel-tab-overflow-button` all start with `docs-panel-tab-` (`src/constants/testIds.ts:29-35`), and `hasText` matches subtree text — so the tab-bar container matches whenever *any* editor tab shows a Draft badge. `.last()` then resolves to whatever happens to be last in DOM order, and `toBeVisible()` on that is close to a tautology.
2. **`beforeEach` clears `pathfinder-block-editor-state*` but not `grafana-pathfinder-app-tabs`.** Since `onCreateEditorTab` now always mints a new tab, a `describe.serial` suite with ~10 `openBlockEditor()` calls accumulates ~10 editor tabs and this still passes.
3. **It's coupled to the literal `'Draft'`**, which flips to `'Published'` the moment the tab is bound to a published resource, and matches any guide whose *title* contains "Draft". `E2E_TESTING_CONTRACT.md` exists specifically so tests don't depend on text content.

`readBlockEditorStoredState` (`:130-142`) has the same shape — it returns the first `pathfinder-block-editor-state:` key that `localStorage.key(i)` enumerates, so `waitForAutoSave` asserts against an arbitrary draft rather than the active tab's. Harmless with one editor tab; a flake the moment a spec opens two.

The underlying gap is that nothing in the DOM now says "this strip element is an editor tab" — editor tabs went from a fixed `docs-panel-tab-editor` to `generateTabId()`. Adding `data-tab-type={tab.type}` to the strip button (and pinning it in `docs-panel.contract.test.tsx` next to the existing `tab('devtools')` assertion) gives E2E a stable handle, and lets this become `expect(page.locator('[data-tab-type="editor"]')).toHaveCount(1)` — which would actually catch a double-create.

Worth adding one spec for the headline capability while you're here: open two editor tabs, type distinct content in each, reload, assert both drafts survive under their own keys. Nothing covers that today.

src/components/block-editor/hooks/useBlockPersistence.ts
```diff
@@ -96,46 +75,93 @@ function restoreJsonModeState(value: unknown): JsonModeState | undefined {
...
+        idIsFinalized,
         blockIds,
         viewMode,
         jsonModeState: viewMode === 'json' ? (jsonModeState ?? undefined) : undefined,
         savedAt: new Date().toISOString(),
         version: STORAGE_VERSION,
-      };
-      localStorage.setItem(storageKey, JSON.stringify(stored));
+      });
       lastGuideRef.current = JSON.stringify(guide);
```
Jayclifford345, line 105: [nit] **A failed storage write now reports success, and stops logging.**

`writeEditorDraftState` swallows silently (`editor-tab-storage.ts:347-358`, no logger call), so this always reaches `lastGuideRef.current = ...`, `pendingSaveRef.current = false`, and `onSave?.()` → `editor.markSaved()` → `isDirty: false`. Pre-PR, `localStorage.setItem` was called directly inside this `try`, so a `QuotaExceededError` skipped `onSave?.()` *and* logged.

Small in practice — it needs a write failure (quota, Safari private mode), and the "All changes saved to local storage" indicator only renders when `!isBackendAvailable` (`BlockEditorHeader.tsx:139`), a state where the work is unsaveable anyway. The real regression is that the user and the logs are both no longer told.

Having `writeEditorDraftState` return a boolean (or rethrow) and gating `lastGuideRef` / `pendingSaveRef` / `onSave` on it restores both. Worth noting the quota path becomes more reachable given the unbounded per-tab key growth.

src/components/docs-panel/utils/tab-storage-restore.ts
```diff
@@ -67,12 +68,12 @@ function getCanonicalPersistedTabType(data: PersistedTabData): LearningJourneyTa
   if (type === 'devtools') {
     return data.id === DEVTOOLS_TAB_ID ? type : null;
   }
-  if (type === 'editor') {
-    return data.id === EDITOR_TAB_ID ? type : null;
+  if (SINGLETON_TAB_IDS.has(data.id)) {
```
Jayclifford345, line 71: [nit] **`'editor'` is still a reserved identity, but it's no longer reserved here.**

The docstring three lines up still promises the invariant — "Reserved IDs and types must therefore agree in both directions so a malformed content record cannot acquire privileged singleton behavior" — and `'editor'` is still `LEGACY_EDITOR_TAB_ID`, the migration target. But `SINGLETON_TAB_IDS` is now `{recommendations, devtools}` only, so a persisted `{id: 'editor', type: 'docs'}` passes validation and restores as a docs tab holding the reserved id.

From there: `recoverLegacyEditorTab`'s `hasLegacyChrome` requires `type === 'editor'` → false; `createEditorTab`'s existence check is `t.id === tabId && t.type === 'editor'` → also false; so it pushes a **second** tab with `id: 'editor'`. Duplicate React keys, ambiguous `setActiveTab`, and `closeTab`'s `filter(t => t.id !== tabId)` removes both — closing the recovered editor silently closes the user's docs tab, and the reduced list is written straight to `tabStorage`.

Only reachable from tampered or corrupt storage, hence a nit — but the `{id:'editor', type:'docs'}` "Disguised editor" case deleted from `tab-storage-restore.test.ts` in this same change is exactly what pinned it. (The companion deletion, "Content claiming editor privileges", is clearly right — editor is legitimately multi-instance now.)

Either keep `'editor'` reserved for non-editor kinds and restore that case, or make `createEditorTab` id-collision-safe by matching on `t.id === tabId` regardless of type.

src/components/block-editor/editor-tab-storage.ts
```diff
@@ -0,0 +1,440 @@
...
+  } catch {
+    return null;
+  }
+}
+
+/** True when another per-tab draft already uses this guide id. */
+export function editorGuideIdExists(guideId: string): boolean {
+  try {
+    const prefix = `${STATE_BASE_KEY}:`;
+    for (let index = 0; index < localStorage.length; index++) {
```
Jayclifford345, line 203: [nit] **QC4 — this sweep already exists as a Tier 1 helper.**

`src/lib/storage/key-utils.ts:25` `collectKeysByPrefix(storage, prefix)` is the same three-step pattern, and its docblock says it was written because the sweep "is duplicated in 7+ call sites across the codebase". This is the eighth. (Ascending vs descending iteration doesn't matter here since nothing is removed while iterating.)

Separately on cost: this scans the entire `localStorage` key space and `JSON.parse`s every matching document, and `generateUniqueId` calls it inside a 20-iteration retry loop. In practice it exits on the first draw, but `BlockEditor.tsx:254` runs `generateUniqueId` in a `useState` initialiser on *every* mount — and with `key={activeTab.id}`, that's every editor-tab switch — then discards the result whenever the tab restores a stored draft. Making that call lazy (only mint when `readEditorStoredState(storageKey)` has no guide) would skip the scan entirely on the common path.

Scoping `editorGuideIdExists` to the known editor tab ids would also stop orphaned keys from permanently reserving guide slugs — the signature is already available, `findEditorTabIdByGuideId` right below takes an explicit `editorTabIds`.

src/components/docs-panel/docs-panel.tsx
```diff
@@ -278,10 +306,22 @@ class CombinedLearningJourneyPanel extends SceneObjectBase<CombinedPanelState> i
...
+    const pruned = this.pruneUnauthorizedGatedTabs(before, this.state.activeTabId);
     if (!pruned.didPrune) {
       return;
     }
+
+    if (pruned.gateClosed) {
+      const keptIds = new Set(pruned.tabs.map((t) => t.id));
+      for (const tab of before) {
+        if (tab.type === 'editor' && !keptIds.has(tab.id)) {
+          flushEditorDraft(editorTabStorageKey(tab.id));
```
Jayclifford345, line 319: [question] **These two lines state opposite intents, and this is the one editor-tab teardown path with no confirmation.**

`flushEditorDraft` writes the pending draft; `clearEditorTabStorage` deletes the key it just wrote. The flush is unreachable work — but it reads as a preservation step, which makes it easy to miss that an observed `allowEditor` open→closed transition silently and irrecoverably deletes every local editor draft. Contrast `closeTab`, which now defers into `pendingCloseTabId` and asks.

The `didGateClose` machinery is careful and I don't think this misfires in practice — `isCurrentUserEditor()` derives from static `config.bootData.user`, so a true→false transition inside one model instance is hard to reach. So: mostly a readability question. Is discarding drafts on a role downgrade the intended policy? If so, dropping the flush and saying so in a line would make it obvious. If not, pruning the tabs while leaving the drafts in place would let a re-granted role recover the work.

PR #1504, opened 2026-08-05
feat(scripts): upload a learning path via the backend API

Jayclifford345:
## Summary

Adds `scripts/upsert-learning-path.sh` — the authoring counterpart to #1408. The `InteractiveGuide` kind gained `spec.manifest` ([grafana-pathfinder-backend `kinds/interactiveguide.cue`](https://github.com/grafana/grafana-pathfinder-backend/blob/main/kinds/interactiveguide.cue)), so a learning path is now expressible as a backend resource — but nothing could author one. `upsert-guide.sh` uploads a single content-only guide.

The script walks an [interactive-tutorials](https://github.com/grafana/interactive-tutorials) package directory (`manifest.json` + `content.json`), resolves each id in `milestones` to the subdirectory whose manifest declares it, and uploads the milestones **before** the path's cover page so the path never references a guide that isn't there. Subdirectories absent from `milestones` are ignored, keeping website-only prelude pages (`business-value-*`) out. Re-running updates in place.

## Depends on #1408

This is the **write side**; #1408 is the read side (`app-platform-resolver.ts`, `app-platform-paths.ts`, the `/custom-guide-repository` proxy). #1408's own verify step is _"author a path package by POSTing an `InteractiveGuide` with `spec.manifest`"_ — this script is what makes that reproducible instead of hand-rolled curl.

They're independent code paths and this one doesn't touch `src/` at runtime, so it can merge in either order. Without #1408 deployed, an uploaded path stores correctly but renders no milestones in the sidebar.

## What to look at

- **`scripts/lib/app-platform.sh`** — the slug rule that decides a resource's identity, hostname acceptance, and credentialed request construction, in one place. All three were duplicated across the two scripts, and resource naming is the very contract the tool warns callers about.
- **`scripts/upsert-learning-path.sh`** — the manifest mapping is a strict whitelist because the CRD's structural schema rejects unknown fields under `spec`. `depends` is widened from the flat `string[]` used on disk to the CNF `string[][]` the CRD requires, `author` is projected to `{name, team}`, and everything without a typed home is swept into `additionalFields` rather than dropped.
- **`scripts/tests/`** — the behavioural suite (below). It is the reason to trust any of the claims here.

## Three ways this can lose data, and what stops each

**A write replaces `spec` wholesale and the API has no revision verb.** Resource names are slugified package ids from a third-party content repo, so a milestone id like `getting-started` can land on a hand-authored guide of that name. The tool LISTs the collection before writing — draining `metadata.continue`, since this API truncates a one-page read silently — and refuses the whole run on a name it didn't write, before touching anything. `--overwrite` opts in. Because that listing is a snapshot, each write also re-checks ownership against the same GET whose `resourceVersion` it sends (`upsert-guide.sh --require-annotation`), so a resource created or detached in between is refused rather than replaced.

**`spec.id` must already be slug-shaped.** `manifest.milestones` and the catalogue key on `spec.id`, while the resolver and `backend-guide:` URLs GET by `metadata.name` (the slugified `spec.id`). Diverge and every milestone 404s with nothing surfaced in the UI. The script refuses to upload.

**Undeclared block fields are silently pruned** — 200 on the write, field gone on the next GET. The script warns per resource; `--strict-blocks` makes it fatal. Worth knowing that this gap moves in *both* directions: an earlier revision of this PR carried an allowlist that had fallen 26 fields behind the CUE after the backend caught up, so `--strict-blocks` was rejecting content the CRD accepts and every `autoCollapse` warning was a false positive. The remaining gap is `defaultValue` alone. `src/validation/upsert-script-crd-fields.test.ts` now fails if the app's `KNOWN_FIELDS` drifts ahead of the allowlist again.

## Credential handling

`--stack` is validated as one DNS hostname with an optional port before the token is attached to anything — a brace expansion (`{trusted,attacker}`) or userinfo (`trusted@attacker`) would otherwise hand a live service-account token to whatever curl resolved — and every request carries `--globoff --proto '=https'`.

The token also stays out of the process table, where any same-user process could read it for the length of a multi-resource run: it reaches the child script through `$PATHFINDER_SA_TOKEN` rather than argv, reaches curl in a `0600` config file rather than `-H`, and payloads go out with `--data-binary @file`. A test asserts a sentinel token never appears in captured child argv.

## How to verify

`scripts/tests/upsert-learning-path.test.sh` runs the real script against a stubbed `curl` — no stack needed — and asserts on what does and does not reach the wire. 64 cases covering argument handling and usage codes, hostname rejection, credential non-disclosure, package translation, content shape, the block-field allowlist in both directions, milestone-before-cover ordering, collision refusal and `--overwrite`, the detached-after-pre-flight race, listing pagination and fail-closed behaviour, partial failure, and signal cleanup:

```bash
npm run test:scripts
```

That runs `bash -n` over every tracked script, `shellcheck --severity=warning` over this surface, and the suite. It is wired into `npm run check` and a `Test scripts` CI job (which installs shellcheck), so this is a real gate rather than a one-off transcript. My earlier claim that "no shell test infrastructure exists in this repo and this PR doesn't add one" was accurate at the time and is no longer.

Also verified end to end against `pathfinder.grafana-dev.net` (`stacks-9239`, Grafana 13.2.0), where `spec.manifest` and the #1408 proxy are deployed: 8 resources created, `business-value-log-metrics` correctly excluded, all 7 milestone ids resolvable by resource name, `metadata.name == spec.id`, `targeting` preserved under `additionalFields`; re-run → 8 updated, 0 created. All 120 packages in interactive-tutorials dry-run clean.

## Breaking changes

None. New script; `upsert-guide.sh` gains `--require-annotation` and `$PATHFINDER_SA_TOKEN`, both additive, and its error path now surfaces the K8s `Status` message on create as well as update.

## Reversibility

Reverting this PR removes the authoring command. It does not undo a command that already ran: a run performs persistent creates and wholesale `spec` replacements, never deletes a milestone dropped from the manifest, and the API keeps no revision history.

To unwind a run: retry it to converge after a partial failure (writes are idempotent); `DELETE` resources it created or that the manifest no longer lists; restore anything it replaced from a pre-write export. `docs/developer/EXTERNAL_API.md` carries this as a table. Take an export first if the namespace holds guides you cannot recreate — `--dry-run` plus the collision refusal is what makes that rarely necessary, not an ability to undo.


moxious (changes requested):
Requesting changes on four items. The upload logic itself holds up well — `shellcheck` is clean, it behaves identically on bash 3.2.57 and 5.3, the milestone-before-cover ordering is right, and the `spec.id`-must-be-slug guard closes a genuinely nasty silent-404 mode. Everything below is about the dry run, the docs, and the error path.

Findings 1 and 2 undercut claims the PR itself makes, so they're the ones I'd most want fixed here rather than in a follow-up.

---

## 1. `--dry-run` exits 0 when validation failed

`scripts/upsert-learning-path.sh:377-380`

The dry-run `exit 0` sits **above** the `FAILED > 0` check at `:383`, and the root cover-page call at `:371-374` — unlike the milestone loop at `:366` — has no `exit 1` of its own. Reproduced with no special flags, root `content.json` `.id` mismatching `manifest.json`:

```
fx2/mypath/content.json .id is "My Path With Spaces" but manifest.json declares "mypath"

Dry run complete: 2 resources would be uploaded.
EXIT=0
```

Also fires for `--dry-run --continue-on-error` with an unresolvable milestone id, where the count is wrong too — "3 resources would be uploaded" while one has no directory at all. The non-dry-run path is correct; this is dry-run-scoped.

Why it matters: the dry run is this PR's stated verification strategy, and `EXTERNAL_API.md` frames this as the CI / Terraform surface. As written the preview is strictly weaker than the real run — the identical fault (bad cover `id`, `--strict-blocks` violation) is fatal live but silent in dry run. Wired as a CI gate, that means green gate → real run publishes the milestones → cover page then fails, leaving a half-published path on a live stack.

**Fix:** move the `FAILED` check above the dry-run branch, and report `TOTAL - FAILED`:

```bash
if [[ "$FAILED" -gt 0 ]]; then
  echo "Failed:${FAILED_NAMES}" >&2
  exit 1
fi
if [[ "$DRY_RUN" == true ]]; then
  echo "Dry run complete: ${TOTAL} ${RESOURCES} would be uploaded."
  exit 0
fi
```

---

## 2. The docs understate the CRD block-field gap by 14 fields

`docs/developer/EXTERNAL_API.md:129-133`

The doc names six undeclared fields (`autoCollapse`, `start`, `end`, `code`, `language`, `command`). I diffed `KNOWN_FIELDS` on `main` against `#Block`/`#NestedBlock`/`#Step` in `grafana-pathfinder-backend/kinds/interactiveguide.cue`. The real count is **20**:

> `authorNote, autoCollapse, buttonText, code, collapsed, command, datasourceFilter, defaultValue, end, language, openGuide, screens, shuffle, snippetId, start, targetstate, vmApp, vmScenario, vmTemplate, welcome`

To be clear about where the error is: **the script is right, the doc is wrong.** `def BLOCK` (51 entries) and `def STEP` (11) are an exact transcription of the CUE — symmetric difference zero in both directions. Every warning it emits is a true positive. I initially suspected the allowlist of over-warning and it does not.

The sharpest omission is `targetstate`. `src/bundled-interactives/loki-grafana-101/content.json` uses it, so uploading that guide through this path silently degrades a toggle step from drive-to-state to blind-click — 200 OK, field gone on the next GET. An author reading the six-item list has no reason to expect that, and `targetstate` is the one field on the list whose loss changes runtime behaviour rather than just rendering.

**Fix:** list all 20, or derive the list. Worth calling out `targetstate` specifically given the behavioural consequence.

---

## 3. The script can overwrite an unrelated pre-existing guide in the namespace

`scripts/upsert-learning-path.sh:311-329`, `scripts/upsert-guide.sh:187-208`

Resource names are slugified package ids, and the block editor derives names with a byte-identical rule (`useBackendGuides.ts:133-136` vs `upsert-guide.sh:106-110`). `upsert-guide.sh` does GET-200 → PUT with a wholesale `spec` replacement and no ownership, provenance, or annotation check. The duplicate-id guard only compares the package's own subdirectories against each other — it never consults what already exists on the stack.

So a tutorials package containing a milestone id like `getting-started` replaces a hand-authored guide of that name. There's no copy in the source repo to restore from and no revision verb in the API, so reverting this PR doesn't help. `--dry-run` can't warn either, since it never contacts the stack (`:331-340` is skipped).

Two things genuinely soften this, and I don't want to overstate it: the editor's create path POSTs to the collection URL, so it 409s rather than clobbering; and the per-resource line prints `updated` rather than `created`, which is a real if weak signal. `GET → PUT` is also pre-existing on `main`. What's new is applying it in bulk, keyed on ids from a third-party content repo.

**Fix:** LIST the collection once before the first write and refuse to overwrite a name this tool didn't write, unless `--overwrite` is passed. Having `--dry-run` perform that LIST would also make the preview actually preview. Cheaper version if that's too much: distinguish `overwrote existing` from `updated` in the summary so the operator at least learns after the fact.

---

## 4. `write_resource` is wired into PUT only, so create still discards the error body

`scripts/upsert-guide.sh:221`

The create branch still runs `curl -sSf`, which emits `curl: (22) The requested URL returned error: 422` and throws away the K8s `Status` body naming the rejected field. The PR description says the change surfaces that message for writes; it does so only for updates.

That's backwards from where it's needed. The first upload of a `spec.manifest`-bearing guide to a stack whose CRD predates the field is a **create** — precisely the case the new hint at `upsert-learning-path.sh:265-266` is written for, and it tells the operator _"if it names `manifest`"_, referring to a message the create path structurally cannot produce.

**Fix:** route the POST branch through `write_resource` too. Same three-line change as the PUT branch.

---

Separately, one bug worth folding in while you're in the file, since it contradicts the header's own promise at `:33-35` that non-milestone subdirectories are ignored: `sub_id=$(jq -r '.id // empty' "${dir}manifest.json")` at `:314` is a bare assignment under `set -e`, so a `business-value-*` prelude directory with an unparseable `manifest.json` aborts the entire run with `EXIT=5` — outside the documented `0/1/64/66/127` contract, and with no message at all if the file is valid JSON but not an object. A `|| { echo "warning: skipping ${dir}" >&2; continue; }` covers it.

There are a handful of smaller items (unanchored `grep -q '422'` matching namespace digits; `author.*` extra keys dropped into neither the typed field nor `additionalFields`; the "re-running is idempotent" wording being additive-only given there's no delete; root id excluded from the duplicate guard). Happy to drop those in as a separate comment or leave them for a follow-up — none of them block.

moxious (changes requested):
Re-reviewing at 7f9b0be9. The four blocking items from the last round are addressed, and three are cleanly closed — I verified the dry-run exit gate (`EXIT=1`, count now `TOTAL−FAILED`), the `write_resource` POST wiring, and the `HTTP 422` anchoring by running them. The provenance design in finding 3 is better than what I asked for: pre-flight-before-any-write with the comment explaining why the per-resource check isn't enough is exactly right, and the `($existingAnn + $ann)` merge is a real improvement over a base that wrote no annotations at all.

Requesting changes on a much smaller surface. Three items, all the same shape as the one I flagged last time — a guard that is correct in structure but fails open or arrives too late — plus four cheap ride-alongs.

Two things from my earlier pass you should ignore: I flagged a `CONTEXT_INDEX.md` narrowing that turned out to be a two-dot diff artifact (the widened row landed on `main` in #1580 after you branched — your hunk doesn't touch it, don't "fix" it), and a LIST pagination finding I've withdrawn as scope-stretched.

---

## Gate

### 1. Only the collision check was hoisted above the writes

`scripts/upsert-learning-path.sh:222-271` vs `:496-500`

The rationale in your commit message — a cover-page refusal must not leave the milestones already replaced — is right, and it applies to five more checks that stayed lazy: `content.json` readability, `.id`/`.title`/`.blocks` presence, content↔manifest id agreement, slug-shape, and `--strict-blocks`. All are pure local-file checks with no network dependency. Against a stubbed writer:

```
[1/3]   m-a  created
[2/3]   m-b  created
.../lp/content.json .id is "My Path" but manifest.json declares "my-path"
Done. 2 created, 0 updated, 1 failed.        EXIT=1
```

Two milestones live, no cover page sequencing them. Fires regardless of `--continue-on-error`. `--dry-run` catches it, so the documented CI path is safe — a direct run isn't.

**What satisfies the gate:** split `upsert_package` into `validate_package` (everything from `:222` through `:283`) and `write_package` (`:301` onward), then call `validate_package` for every package inside the loop that already walks the names at `:462`. That loop is already positioned correctly and already `exit 1`s; it just needs to validate as well as check provenance. The write loop then calls `write_package` only.

### 2. Ctrl-C deletes the temp dir and keeps uploading

`scripts/upsert-learning-path.sh:154`

```bash
trap 'rm -rf "$TMP_DIR"' EXIT INT TERM
```

No `exit` in the handler, so bash runs it and resumes. I confirmed with a minimal repro: SIGINT fires, `TMP_DIR` is removed, execution continues — and the remaining spec files then can't be written. On a tool whose premise is "a write replaces spec wholesale and the API keeps no revisions," the abort button currently produces the worst available outcome.

**What satisfies the gate:**

```bash
trap 'rm -rf "$TMP_DIR"' EXIT
trap 'rm -rf "$TMP_DIR"; exit 130' INT TERM
```

Add 130/143 to the exit-code contract at `:44-50`.

### 3. `set -e` is suspended inside `upsert_package`, so a jq failure silently disables `--strict-blocks`

`scripts/upsert-learning-path.sh:264`, `:277`, `:282` — call sites `:489`, `:497`

Both call sites are `if ! upsert_package …`, and bash disables `errexit` for the whole body of a function invoked in a condition context. Verified:

```
f called bare:         set -e aborts at the failed assignment       EXIT=1
f called as 'if ! f':  failure swallowed, execution continues       EXIT=0
```

So the reasoning in your comment at `:363-366` holds at `:367` (top level) but not for the structurally identical assignment at `:264`:

```bash
unknown=$(jq -r "$UNKNOWN_BLOCK_FIELDS_JQ" "$content")
```

If that jq fails, `unknown` is empty, the warning is skipped, and `--strict-blocks` passes — the one flag an operator sets specifically to guarantee no lossy upload fails open with no signal. The `jq … > "$spec_file"` writes at `:277`/`:282` are unprotected the same way, though those degrade to a counted failure rather than a bad write.

**What satisfies the gate:** `|| { echo "…" >&2; return 1; }` on all three, matching the guard you already wrote at `:367`.

---

## Cheap enough to take in the same pass

### 4. The dry run reports success having checked nothing, in the default invocation

`:422` / `:434`, doc claim in `EXTERNAL_API.md`

The `warning: could not list existing guides` sits inside `if [[ -n "$NAMESPACE" ]]`. `--namespace` is optional and omitted from every doc example, so an unreachable stack skips the whole block, warning included:

```
curl: (6) Could not resolve host: bogus.invalid.example
Package:   p (path, 2 resources)          <- no "Namespace:" line
Dry run complete: 2 resources would be uploaded.        EXIT=0
```

The doc promises "if the stack is unreachable the collision check is skipped with a warning." A typo'd `--stack` or CI egress rules reduce the gate to a JSON linter while reporting success — the scenario your own comment at `:516-518` warns against.

**Fix:** move the warning out of the namespace guard, and print `Collisions: not checked` on stdout next to `Mode: dry run` so the preview is self-describing.

### 5. Exit codes 5 and 2 escape the documented contract

`:335-336`, `:463`

```
$ upsert-learning-path.sh --dry-run … --package <malformed root manifest.json>
jq: parse error: Invalid numeric literal at line 1, column 4
EXIT=5
```

No package name, no remediation, no summary. `PKG_TYPE=`/`PKG_ID=` are bare command substitutions at top level; `existing_owner`'s awk exits 2 on a newline-bearing id. This is the exact bug class you diagnosed and guarded at `:363-366` — for subdirectories, but not for the mandatory root manifest.

**Fix:** guard both root reads as `:367` does, and validate `PKG_ID` is slug-shaped right after `:336`. That second half also closes an annotation-injection path: `PKG_ID` is stamped onto every milestone at `:305` before the root's own validation runs last, so a value containing a newline plus `=` can override the `managed-by` key the overwrite guard depends on.

### 6. The block-field severity sentence is wrong

`EXTERNAL_API.md`, the "`targetstate` is the one to know about" paragraph

"Every other field on the list costs you rendering or authoring metadata" isn't true of `setupCommands`, `setupScript`, `successCriteria`, `brief`, or the vm* family — `successCriteria` and `brief` are required on `challenge`, so a pruned challenge block fails app-schema validation outright. Worth knowing: `targetstate` appears **0 times** in the uploadable `interactive-tutorials` corpus, while `vmTemplate`/`vmApp` appear in `enable-coda` — the sentence elevates the less reachable field.

Keep the punchy framing, just widen it: name the behavioural fields before the metadata claim.

Separately, the published list omits six fields the script _will_ warn on — `buttonText`, `code`, `language`, `screens`, `snippetId`, `welcome` (from `terminal-connect`, `code-block`, `grot-guide`, `snippet-ref`). Four are the required payload of their block type. Three independent derivations agree on this set. Since the doc already names `--dry-run` as the live check, the durable fix is probably to delete the enumeration rather than maintain a second copy.

### 7. The annotation-preservation claim is one-directional

`EXTERNAL_API.md`: "Annotations set by other tools are preserved across an update."

True of `upsert-guide.sh`. False of the other two writers of these resources — `useBackendGuides.ts` (`saveGuide`/`publishGuide`/`unpublishGuide`, each builds `metadata` as `{name, namespace, resourceVersion}`) and `finalize.ts:151-153`. A PUT is a whole-object replace, so one **Unpublish** click on a script-uploaded milestone erases `managed-by`, and the next run's pre-flight refuses the entire package — with run-global `--overwrite` as the only offered unblock.

I'm not asking you to fix the editor here. Just say it in the doc, because an operator hitting that refusal has no way to tell a self-inflicted detachment from a genuine foreign guide, and `--overwrite` can't tell either.

Worth flagging while you're in there: `saveGuide` also drops `spec.manifest` (`JsonGuide` has no `manifest` field, so the editor can't round-trip it). Since this script is effectively the only producer of `spec.manifest` resources, editing a path cover in the editor silently collapses it into a flat guide — `app-platform-resolver` falls back to `{id, type:'guide'}` and warns only on a malformed manifest, not a missing one. That's an editor bug, not yours, but it's worth an issue and a sentence in `CUSTOM_GUIDES.md` next to the paragraph that tells authors to load and update.

---

## Not blocking — follow-ups

- Add a **pre-contract candidate** row to `CONCERNS.md` for the `managed-by`/`source-package` ownership vocabulary. Not an anchor: only one of three writers honors it today, so there's no coherent contract to anchor yet. Promote it once the editor and MCP preserve annotations.
- Add `scripts/upsert-*.sh` to `guide-schema-and-contracts` `trigger_paths` — right now this 522-line tool activates no subsystem reviewer at all. (`build-and-ci` already sets the precedent with `scripts/e2e-runner*`.)
- `jq '.items[]?'` exits 0 on any JSON body, so a degenerate 2xx (a 302 — there's no `-L` — or a `Status` body) sets `COLLISION_CHECK=true` on empty data and bypasses `--overwrite` entirely, reporting the write as `updated`. One line: assert `.items | type == "array"` before setting the flag.
- Extract the duplicated `slugify` into `scripts/lib/slugify.sh`. It's the resource-identity rule the script itself warns about at `:239-241`.
- Replace the `grep -q '^Creating '` coupling with a machine-readable marker from `upsert-guide.sh`.
- A committed fixture package plus `--dry-run` in CI — ~15 lines, `jq` is preinstalled on `ubuntu-latest`. It would have caught items 1, 4, and 5. Your "no shell test infrastructure exists" note is accurate; I verified it (no bats, no shellcheck in CI, prettier and lint-staged globs exclude `.sh`, and none of the five existing shell scripts is tested, including the one CI actually executes). Not a gap I'd hold this PR for.
- `COMMANDS.md` documents neither upsert script.

The upload logic itself continues to hold up. Milestone-before-cover ordering, the slug-equality guard, bash 3.2 compatibility, and the pre-flight design are all sound — every item above is a guard arriving late or failing quiet, not a design problem.

moxious:
Filed the two editor-side follow-ups I mentioned in the review, so they are tracked independently of this PR:

- #1599 — `saveGuide` deletes `spec.manifest`, collapsing a path into a flat guide
- #1600 — all three editor writes (and MCP `finalize.ts`) drop `metadata.annotations`, detaching script-managed guides from their provenance

Neither blocks this PR. #1600 is the one worth knowing about while you finish item 7 in the review — it is the mechanism behind the doc qualifier I asked for.

Jayclifford345:
All seven items addressed, plus both `CONCERNS.md` follow-ups. Rebased onto `main` at `dddbd33c`, so the branch now carries #1580 — I left the `BACKEND_PROXY_PATTERN.md` row alone as you said, and verified the rebase kept your widened version rather than reinstating the narrow one.

## Gate

**1. Only the collision check was hoisted.** Split into `validate_package` (`:237`) and `write_package` (`:322`). The loop that already walked the names now validates each package *and* aggregates clashes (`:558`); a second loop writes (`:582`). Your repro:

```
.../lp/content.json .id is "My Path" but manifest.json declares "lp"
Done. 0 created, 0 updated, 1 failed.        EXIT=1   writes=0
```

Fires the same way with `--strict-blocks`, a malformed root manifest, and an unresolvable milestone id — all now zero writes rather than "milestones live, no cover page". Under `--continue-on-error` the cover page is skipped and the milestones still land, which is that flag's documented trade, but the run reports `EXIT=1` instead of claiming success.

**2. Ctrl-C.** Three traps, `exit 130` / `exit 143`, both added to the exit-code contract. Worth noting for anyone re-testing: SIGINT can't be exercised from a background job — bash sets it `SIG_IGN` for `&` commands, and a signal ignored at entry can't be trapped, so `kill -INT` on the script pid does nothing regardless of the handler. I tested SIGTERM against a control:

```
old shape (no exit):  wrote 1 (tmpdir yes) / wrote 2 (tmpdir NO) / wrote 3 (tmpdir NO) / reached end   EXIT=0
new shape:            wrote 1                                                                          EXIT=143
```

**3. `set -e` suspended inside the function.** Guarded all three. Confirmed load-bearing with content whose `.id`/`.title`/`.blocks` read fine but whose scan errors (`blocks: ["oops"]` → `keys` on a string):

```
unguarded + --strict-blocks:  [1/3] m-a created  [2/3] m-b created  [3/3] lp created   EXIT=0  writes=3
guarded   + --strict-blocks:  could not scan .../m-a/content.json ...                  EXIT=1  writes=0
```

I also folded the three `.id`/`.title`/`.blocks` reads into one guarded `@tsv` call, so a malformed milestone reports as malformed instead of "has no .id".

## Ride-alongs

**4.** The LIST attempt moved out of the `-n "$NAMESPACE"` guard, so the warning always fires, and the header prints `Collisions: not checked` in both modes. I took the `.items | type == "array"` follow-up here too, since it's the same fail-open shape — a `Status` envelope, a redirect body, and `{"items":null}` are all rejected now instead of reading as an empty namespace.

**5.** Both root files are parse-checked up front, so a malformed root manifest is `EXIT=1` naming the file rather than `EXIT=5` with jq's raw error. `PKG_ID` is now checked for emptiness (there was no check at all) and slug shape before the first write, which closes the annotation-forge path you spotted — a newline-plus-`=` id is refused before it reaches `--annotation`.

**6.** Widened the paragraph to lead with the behavioural fields: `brief`/`successCriteria` are `.min(1)` on the challenge schema, so a pruned challenge fails app-side validation outright; then `setupCommands`/`setupScript` and the vm\* trio, then `code`/`language`/`screens`/`snippetId` as the whole payload of `code-block`/`grot-guide`/`snippet-ref`. **Deleted the enumeration** rather than fix it — your read is right that a second copy of a set that moves with both schemas will just rot again. The regeneration procedure stays, and `--dry-run` is named as the authority.

**7.** Scoped the claim to `upsert-guide.sh` and stated what the other writers do. Confirmed before writing it: `saveGuide` builds `spec` as `{id, title, schemaVersion, blocks, status}`, so `spec.manifest` is dropped, and all three editor paths rebuild `metadata` as `{name, namespace, resourceVersion}`. `CUSTOM_GUIDES.md` gained a paragraph on the cover-page case, including that `--overwrite` is the unblock and that neither it nor the message can tell a detachment from a real collision.

## Follow-ups now included

Both `CONCERNS.md` items are in, as a separate commit. Worth flagging that the first attempt at the `trigger_paths` row conflicted: the `requirements-manager` row directly above it gained paths and keywords in #1578, so on the un-rebased branch the hunk couldn't auto-merge — and a careless resolution at merge time would have reverted your additions. I pulled it back out, then rebased and re-added it against the current file. The ownership vocabulary went in as a **pre-contract candidate**, not an anchor, for the reason you gave.

Still open, deliberately: the `slugify` extraction, the machine-readable marker replacing `grep -q '^Creating '`, the CI fixture, and `COMMANDS.md`.

## Verification

19-case regression suite against a stub writer covering every error path above plus `--no-manifest`, `--status draft`/bogus, a standalone `type: guide` package, and both `--continue-on-error` branches. All pass. `shellcheck` is back to the three pre-existing SC2001 hints and nothing else; `bash -n` clean; prettier clean on all four docs. One gap in my verification versus yours: both `bash` and `/bin/bash` on this machine are 3.2.57, and there's no bash 5 installed, so I only re-ran the 3.2 leg.

moxious (changes requested):
Independent review run against this repo's own review orchestrator. The counterpart claim checks out: `spec.manifest` is present in the deployed backend manifest (group `pathfinderbackend.ext.grafana.app`, `$ref: #/components/schemas/Manifest`), the source CUE declares `manifest?: #Manifest` at `kinds/interactiveguide.cue:42-45`, the read side already consumes it, and an offline dry run resolved all 120 current `interactive-tutorials` packages. Requesting changes on the writer implementation.

**[security][blocking] Validate the stack host before attaching credentials.** Both scripts strip an optional scheme and trailing slash, then interpolate `STACK` into a credentialed curl URL without parsing a hostname, rejecting userinfo/path/query/glob characters, disabling URL globbing, or restricting the protocol. `https://{trusted.invalid,attacker.invalid}/apis/test` issues two authenticated transfers, and `https://trusted.invalid@attacker.invalid/apis/test` resolves to `host=attacker.invalid` — the `Authorization` header rides along in both cases. A mis-pasted or hostile `--stack` therefore discloses a live service-account token, usable until rotated. Centralise endpoint construction in one helper, accept exactly one DNS hostname plus optional numeric port, and add `--globoff --proto '=https'` to every invocation. (`scripts/upsert-learning-path.sh:151-153,454-475`; `scripts/upsert-guide.sh:119-122,180-190`)

```suggestion
if [[ ! "$STACK" =~ ^[A-Za-z0-9]([A-Za-z0-9.-]*[A-Za-z0-9])?(:[0-9]{1,5})?$ ]]; then
  echo "--stack must be one hostname with an optional port" >&2
  exit 64
fi
```

**[security][blocking] Keep the token out of process arguments.** `--token` is forwarded to `upsert-guide.sh` as an argument for every resource and interpolated into curl's `-H`, so any same-user process can read it from the process table for the duration of a multi-resource run; the payload also rides in curl's `-d`. Pass the secret through a protected file descriptor or token file, keep it out of script-to-script argv, give curl its sensitive header via a protected config input, and send JSON with `--data-binary @-`. Worth a test asserting a sentinel secret never appears in captured child argv. (`scripts/upsert-learning-path.sh:123,332-335`; `scripts/upsert-guide.sh:187-190,207`)

**[blocking] The collision guard is neither pagination-complete nor atomic with the write.** The preflight performs one LIST, accepts any `items` array as a complete collision index, and ignores `metadata.continue` — with curl stubbed to return `{"items":[],"metadata":{"continue":"next-page"}}` the run printed `Dry run complete: 8 resources would be uploaded.` and never reported that collisions were unchecked. `docs/design/BACKEND_PROXY_PATTERN.md:25-30` records that this API silently truncates one-page reads. Separately, the helper re-GETs, adopts the current `resourceVersion`, and PUTs without re-checking ownership against that same response (`scripts/upsert-guide.sh:207-234`), so a resource created or detached after preflight can be replaced without `--overwrite` — and per this PR's own docs no API revision can restore it. Prefer exact-name GET preflight per intended resource: POST when absent so a concurrent create returns 409, PUT with the preflight `resourceVersion` so a concurrent change returns 409. If LIST remains, drain `metadata.continue` with a bound and fail closed on any page error. (`scripts/upsert-learning-path.sh:473-485,559-572`)

**[blocking] Commit the behavioural suite and run it in CI.** This adds 588 lines of branching credentialed shell and changes the shared writer, with no test file in the diff and no `shellcheck`, `bash -n`, or script invocation anywhere in `package.json` or `.github/workflows`. The 32 green checks provide no evidence for argument handling, package translation, schema parity, pagination, ownership races, overwrite opt-in, partial failure, signal cleanup, host rejection, or credential non-disclosure — and several review rounds have already found fail-open branches in this file. The 19-case stub suite described in a comment is not on the branch and did not run. Committing it plus one package command wired into CI would cover most of the above without introducing a new framework.

**[blocking] Use one package/schema publish gate.** The new shell validator forks the established contract and is incorrect in both directions. False rejection: the hard-coded `BLOCK` and `STEP` arrays omit 26 Block fields and two Step fields that are present in the current deployable backend manifest, so `--strict-blocks` fails 53 of 120 current packages — `drilldown-logs-lj` fails six of eight resources solely on `autoCollapse` — while `docs/developer/EXTERNAL_API.md:130-168` still calls this script the authoritative check. False acceptance: `validate_package` checks only nonempty stringified `id`/`title`, that `blocks` is an array, slug agreement, and unknown field names, so it accepted `{"id":"invalid-guide","title":42,"blocks":[{}]}` with exit 0 — a payload the production CRD rejects, potentially after earlier milestones have already been created or replaced. `src/validation/validate-package.ts:67-197` already applies `ContentJsonSchema`, `ManifestJsonSchema`, `validateGuide` and cross-file semantics; invoke it before resource composition and leave only backend-specific manifest translation here. (`scripts/upsert-learning-path.sh:170-186,237-302`)

**[blocking] State the rollback contract accurately.** The body's "fully reversible" describes source rollback, not the effect this PR introduces. A run performs multiple persistent creates and wholesale spec replacements, never deletes removed milestones, and has no revision API; an injected failure on the second write produced `Done. 1 created, 0 updated, 1 failed.` Reverting the PR prevents future runs but undoes none of that. (`scripts/upsert-learning-path.sh:578-588`; help at `:95-97`; `docs/developer/EXTERNAL_API.md:170-176,212-214`)

```suggestion
## Reversibility

Reverting this PR removes the authoring command, but it does not undo a command that already ran. Runs create or replace persistent InteractiveGuide resources, and the API has no revision history. Retry a failed run to converge; manually delete newly created or removed-from-manifest resources, and restore overwritten resources from a pre-write export.
```

**[blocking] Do not add a second owner for resource naming and HTTP.** `slugify` is byte-for-byte duplicated across both scripts (`cmp` exit 0) despite being the resource-identity contract, and the 588-line entry point now owns CLI/credential setup, package discovery, schema transformation, collision inventory and ordered write orchestration, plus its own stack normalisation and authenticated namespace access — while the body states there is one credentialed HTTP implementation. The URL, pagination and schema defects above are that drift already showing rather than theoretical cleanup: a maintainer now has to reason across 851 shell lines plus the existing TypeScript publish gate to change one authoring flow safely. Extract one sourced helper for resource naming and credentialed App Platform requests (including endpoint validation and conditional ownership writes), and leave this script responsible for milestone ordering, manifest projection, and summary/exit behaviour.

**[blocking] Return the documented usage error for a missing option value.** `scripts/upsert-learning-path.sh --stack` prints `scripts/upsert-learning-path.sh: line 122: $2: unbound variable` and exits 1, while `:44-50` reserves exit 64 for usage errors — so automation cannot distinguish invalid usage from an upload or backend failure. Guard every value-taking option before reading `$2`. (`scripts/upsert-learning-path.sh:120-127`)

**[question] One item looks outside this PR's scope to me. Your call whether to file it, defer it, or drop it — I have not filed anything.** Writer ownership is still a pre-contract candidate in `docs/design/CONCERNS.md:122-129`, and other writers' manifest/annotation preservation gaps are tracked separately. Worth deciding whether this PR is the right moment to establish that contract anchor rather than leaving a third writer under the same open question.

Jayclifford345:
Pushed as c4925640. Eight of the nine blocking items are fixed; one I'm pushing back on in part, and the `question` I've answered rather than acted on. Everything below is verified by `npm run test:scripts` — a committed suite of 64 cases that runs the real script against a stubbed `curl` and asserts on what does and does not reach the wire.

## Fixed

**Validate the stack host before attaching credentials.** Done, and thank you — this was the finding worth the whole round. `--stack` now has to be one DNS hostname with an optional port, checked before `ap_auth_init` runs, and every request carries `--globoff --proto '=https'`. Both of your reproductions are pinned as tests, along with a path, a query string, and an embedded space:

```
ok  --stack rejects "{trusted.invalid,attacker.invalid}" before attaching the token
ok  --stack rejects "trusted.invalid@attacker.invalid" before attaching the token
```

Endpoint construction is centralised in `scripts/lib/app-platform.sh` (`ap_validate_stack`, `ap_curl`, `ap_guides_url`), so there is one place to get this wrong rather than four.

**Keep the token out of process arguments.** Done, all three ways. The child gets it through `$PATHFINDER_SA_TOKEN` instead of argv; curl gets the `Authorization` header from a `0600` config file instead of `-H`; payloads go out with `--data-binary @file`. Unquoted in the config on purpose — curl takes the rest of the line verbatim, and quoting would subject a token to backslash escaping. The sentinel-argv test you asked for exists:

```
ok  the token never reaches curl's argv
```

Both scripts still accept `--token` for interactive use, with the process-table caveat in `--help` and in `EXTERNAL_API.md`.

**Pagination, and ownership atomic with the write.** Both fixed, and they were two separate bugs.

`collect_existing()` drains `metadata.continue` with a 100-page bound and returns non-zero on any page error, so a partial listing can never set `COLLISION_CHECK=true`. Your `{"items":[],"metadata":{"continue":"next-page"}}` case and a page-two 500 are both tests, as is a collision that only appears on page two.

For the race, I went with the smaller of your two options rather than restructuring the pre-flight: `upsert-guide.sh` gained `--require-annotation key=value`, checked against the same GET whose `resourceVersion` the PUT sends. That makes ownership and write one optimistic transaction without giving up the up-front refusal, which is what stops a cover-page rejection from leaving milestones replaced. Creates already 409 on a concurrent create, since they POST to the collection. The regression case — listing says ours, live GET says detached — is a test:

```
ok  a resource detached after the pre-flight is still refused
ok  --overwrite drops that guard too
```

**Return the documented usage error for a missing option value.** Fixed via `ap_require_value` on every value-taking option in both scripts. `--stack` with no value now prints `--stack requires a value` and exits 64.

**State the rollback contract accurately.** Corrected, close to your suggested wording, in the PR body and as a table in `EXTERNAL_API.md` (created → `DELETE`; updated → re-upload previous contents; replaced → restore from a pre-write export; dropped milestone → `DELETE`; partial failure → re-run to converge). The body no longer says "fully reversible".

**Do not add a second owner for resource naming and HTTP.** Done — `scripts/lib/app-platform.sh` owns `ap_slugify`, stack normalisation and validation, the credentialed `ap_curl`, the collection URL, and namespace detection. `slugify` was the case that mattered: it is the resource-identity rule the script itself warns about, and it was byte-duplicated.

**The block-field allowlist.** You were right, and this is the correction I'd most want on the record: `_blockFields`/`#Block`/`#NestedBlock`/`#Step` at `kinds/interactiveguide.cue` HEAD declare 77 block fields and 13 step fields; the script had 51 and 11. So the backend caught up after your first round — where the symmetric difference genuinely was zero — and the transcription then fell behind it. `--strict-blocks` was rejecting content the CRD accepts, and every `autoCollapse` warning had become a false positive. Refreshed to 77/13. Computed both sides mechanically rather than by eye; the remaining app-side gap is `defaultValue` on `input`, alone.

Since a hand transcription has now drifted in both directions, `src/validation/upsert-script-crd-fields.test.ts` pins it: it fails when the app's `KNOWN_FIELDS` gains a block field the `BLOCK` array lacks, with `PRUNED_BY_CRD` as the deliberate exception list. Mutation-checked by deleting `autoCollapse` from the array — the test fails. It can't see the backend repo, so cross-repo drift still needs the CUE and the array updated together, and the doc now says so. The `EXTERNAL_API.md` severity paragraph you flagged in the previous round is gone rather than widened — every field it named as behaviourally load-bearing is now declared, which is exactly why an enumeration in a doc was the wrong shape.

**False acceptance in `validate_package`.** Fixed. Your `{"id":"invalid-guide","title":42,"blocks":[{}]}` now fails, along with an empty `blocks`, a non-object block, and a block with no `.type` — all before the first write. Four tests.

## Partly pushing back

**"Commit the behavioural suite and run it in CI."** The suite is committed and wired in, so the substance is done: `scripts/tests/upsert-learning-path.test.sh` plus `scripts/tests/curl-stub.sh`, run by `npm run test:scripts`, which is in `npm run check` and in a new `Test scripts` CI job that installs shellcheck and gates `ci-gate`. `bash -n` runs over every tracked script; `shellcheck --severity=warning` over this surface only, because the older scripts carry warnings of their own and cleaning those up belongs to a PR about them.

What I'm not doing is treating the absence of that suite as having been a merge blocker. Your first round assessed the same gap and concluded "not a gap I'd hold this PR for", having verified there was no bats, no shellcheck in CI, and no test for any of the five existing shell scripts including the one CI executes. Nothing about the diff changed between those two rounds. If a 588-line credentialed script needs committed behavioural coverage to merge — and I now think it does — that's a standard worth stating rather than one round's judgement reversing the next's, because the same standard would block the scripts already on `main`.

**"Use one package/schema publish gate."** Half accepted. The false-rejection and false-acceptance defects were real and are fixed above. Calling `src/validation/validate-package.ts` from the script, I'm declining: the script's contract is `curl` + `jq`, it runs from CI shells and operator laptops with no built `node_modules`, and `validate-package.ts` is Tier 1 app code reached today through `dist/cli/`. Making an upload depend on a TypeScript build to validate JSON trades a real dependency for a duplication that is now test-pinned in the one direction a single repo can check. The doc points authors at `npm run validate:packages` as the schema gate and at `--dry-run` as the backend-shape gate, which is the honest division.

## Answered, not acted on

**Whether this PR should establish the writer-ownership contract anchor.** Leaving it a pre-contract candidate, for the reason you gave in the previous round: only one of three writers honours the annotations. `useBackendGuides.ts` (`saveGuide`/`publishGuide`/`unpublishGuide`) and `finalize.ts` still rebuild `metadata` as `{name, namespace, resourceVersion}`, so an **unpublish** click still detaches a script-uploaded milestone, and neither the refusal nor `--overwrite` can tell that from a genuine foreign guide. Anchoring a contract two of its three writers violate would make the next Coda-adjacent PR's evolution scan report a conformance violation against the editor for behaviour this PR didn't introduce. `CONCERNS.md` keeps the candidate row and `CUSTOM_GUIDES.md` warns authors; promote it when the editor and MCP preserve annotations.

Also unchanged from last round, and still worth an issue in this repo rather than a fix here: `saveGuide` drops `spec.manifest` entirely, so editing a path cover page in the editor silently collapses it into a flat guide.

## Previous round, for completeness

Items 1–3 of the gate (validation hoisted above the writes, `trap … exit 130/143`, the three `jq` guards inside a function called in condition context) and cheap items 4–5 (the warning out of the namespace guard, `Collisions: not checked`, the guarded root reads and `PKG_ID` slug validation) were in `7f9b0be9`/`a4e1c4e1` and are now covered by tests rather than by transcript — including the malformed-root-manifest case that used to exit 5, and a malformed `business-value-*` prelude that must warn and continue. Non-blocking follow-ups taken: `scripts/upsert-*.sh` plus `scripts/lib/**` and `scripts/tests/**` in `guide-schema-and-contracts` `trigger_paths`; `.items | type == "array"` asserted before the flag is set; `slugify` extracted; the `grep '^Creating '` coupling replaced with an `action=create` marker; both scripts documented in `COMMANDS.md`; a committed fixture package running under `--dry-run` in CI.

moxious (approved):
Thanks for the detailed response. I re-verified the prior findings at the requested commit and the small follow-up now at the live head. The 64-case shell suite passes locally, the dedicated CI job is a required gate, the original stack/token reproductions are pinned, pagination fails closed, and `--require-annotation` closes both the update and concurrent-create race paths. I also mechanically confirmed the backend/script field sets are exactly 77 block fields and 13 step fields.

Your objection on the round-over-round testing standard is valid, as are the decisions to keep the operator script curl+jq-only and to leave writer ownership as a pre-contract candidate. Current CI is green. Approved.

PR #1508, opened 2026-08-05
feat(docs-panel): show Dev Tools as a strip tab and lift overflow z-index

dgehring-grafana:
## Summary

Dev Tools is a normal closable strip tab instead of a chrome-less singleton view, so leaving it no longer feels like leaving the panel or flipping Dev Mode off. The tab bar also stacks above panel chrome so the overflow menu stays usable when the editor header is open.

## What to look at

- **Tab taxonomy** ([`tab-kinds.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/22521e678bfc2d99a4d0805aaf25dedec8545ef2/src/components/docs-panel/utils/tab-kinds.ts)) — `getGuideStripTabs` excludes only recommendations; Dev Tools claims a strip slot. `hasOnlyNonContentTabs` still treats Dev Tools as non-content so restore is not blocked by an open Dev Tools tab alone.
- **Close adjacency + keyboard cycle** ([`docs-panel.tsx`](https://github.com/grafana/grafana-pathfinder-app/blob/22521e678bfc2d99a4d0805aaf25dedec8545ef2/src/components/docs-panel/docs-panel.tsx), [`keyboard-shortcuts.hook.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/22521e678bfc2d99a4d0805aaf25dedec8545ef2/src/components/docs-panel/keyboard-shortcuts.hook.ts)) — closing a neighbor can land on Dev Tools, and Ctrl/Cmd+Tab includes it. That looks like a regression against the old “skip Dev Tools” tests, but it is the intended model now that the tab has a visible active marker.
- **Overflow stacking** ([`docs-panel.styles.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/22521e678bfc2d99a4d0805aaf25dedec8545ef2/src/styles/docs-panel.styles.ts)) — tab bar `zIndex` moves from `navbarFixed` to `dropdown` so the overflow menu is not trapped behind later `navbarFixed` chrome (editor header / health bar).

## Why

Without a strip tab, Dev Tools felt like a dead-end page: the exit path was unclear and users could mistake “leave this view” for “turn Dev Mode off.” Giving it a real tab also deletes the specialized “strip-excluded / skip in keyboard cycle” treatment.

An alternate close affordance on the Dev Tools page was considered; a strip tab is preferable because it removes that special-case logic and makes Dev Tools navigable like other strip tabs when that is useful.

## How to verify

1. Enable Dev Mode, open Dev Tools from the overflow menu — confirm it appears as a closable strip tab with a normal active marker.
2. Open a guide alongside Dev Tools, then close the guide — focus should move to Dev Tools (not skip over it).
3. From recommendations, press Ctrl/Cmd+Tab — Dev Tools should be in the cycle; Shift+Ctrl/Cmd+Tab should reverse through it.
4. Close the Dev Tools tab with the strip close control (and with Ctrl/Cmd+W while it is active) — confirm you return to another strip tab or recommendations without toggling Dev Mode.
5. With the guide editor open, open the overflow menu — confirm the dropdown paints above the editor header, not underneath it.

## Breaking changes

None. This change is additive and fully reversible. Existing persisted Dev Tools tabs remain valid; they now compete for strip width like other strip tabs.

## Out of scope

- Editor tab status badges and the badge-shrink fix that keeps the close control reachable when the badge overflows — those land in the follow-up chrome-lifecycle branch (`ui/editor-tab-chrome-lifecycle`), not this diff (this PR only lifts tab-bar z-index).
- Fullscreen/floating tab-storage overwrite on dock-back — follow-up (`ui/fix-surface-tab-storage`).

### DIFF
--- a/src/components/block-editor/BlockEditor.tab-chrome.test.tsx
+++ b/src/components/block-editor/BlockEditor.tab-chrome.test.tsx
@@ -1,0 +1,124 @@
+/**
+ * The editor tab's strip title and badge are owned by the docs panel, which
+ * renders outside this component. These cover the two channels that keep them
+ * in sync: the `onGuideTitleChange` callback and the published chrome status.
+ */
+
+import React from 'react';
+import { fireEvent, render, screen } from '@testing-library/react';
+
+import { StorageKeys } from '../../lib/storage-keys';
+import type { JsonGuide } from './types';
+import { BlockEditor } from './BlockEditor';
+import { DEFAULT_GUIDE_METADATA } from './constants';
+import {
+  getEditorChromeStatus,
+  resetEditorChromeStatus,
+  subscribeEditorChromeStatus,
+  type EditorChromeStatus,
+} from './editor-chrome-status';
+
+function storeGuide(guide: JsonGuide) {
+  localStorage.setItem(
+    StorageKeys.BLOCK_EDITOR_STATE,
+    JSON.stringify({ guide, blockIds: [], savedAt: new Date().toISOString(), version: 2 })
+  );
+}
+
+describe('BlockEditor tab chrome', () => {
+  beforeEach(() => {
+    localStorage.clear();
+    resetEditorChromeStatus();
+  });
+
+  it('reports the working guide title on mount and after a rename', () => {
+    const onGuideTitleChange = jest.fn();
+    render(<BlockEditor onGuideTitleChange={onGuideTitleChange} />);
+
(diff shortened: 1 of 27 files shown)


dgehring-grafana:
Opened against the wrong stacked branch (ui/editor-tab-chrome-lifecycle). Re-opening from ui/devtools-strip-visibility.

PR #1510, opened 2026-08-05
feat(docs-panel): sync editor tab chrome title and status across surfaces

dgehring-grafana:
## Summary

There were several cases where quickly switching tabs, changing views, or remounting the editor could leave tab chrome (title + status badge) out of sync with the working guide. This suite locks that lifecycle down so the strip and header stay aligned across mount, edit, save, and surface handoffs.

Stacked on #1509 (`ui/devtools-strip-visibility`).

## What to look at

- **Chrome readiness** (`BlockEditor.tsx`) — title/badge publish waits on `isChromeReady`; `try/finally` around load so a restore failure cannot leave the strip frozen. Unmount resets strip chrome after the persistence draft flush so the long-lived tab bar does not keep a stale badge.
- **Shared status** (`editor-chrome-status.ts`) — single source for header + strip badge mapping. Backend tracking with no local draft reports unsynced (false-clean fix), not a clean published badge over an empty editor.
- **Title sync** — `onGuideTitleChange` from the editor into the owning tab on sidebar, floating, and fullscreen.
- **Persistence** — draft flush on unmount so a mid-debounce surface switch does not drop the latest edit.
- **Strip UI** — italic modified title + status badge styles; tab bar badge shrink so the close control stays reachable.

## Alternatives considered

Originally coupled with minting a unique id for guides still on the `new-guide` placeholder — useful as an extension into multi-tab work, and because an id should almost always be unique. Dropped here: it goes beyond the scope of this lifecycle fix (the editor is still a singleton), and without a persisted id-lock it gets wobbly. Keep the existing title-commit mint + overwrite prompt instead.

## Gotchas

None immediately obvious.

## How to verify

1. Open Create Guide — strip title is “New guide”; header and strip badges match (not saved / draft).
2. Rename the guide — strip title updates without remounting the editor.
3. Edit content, then quickly switch away and back (or dock floating ↔ sidebar) — badge still reflects unsynced vs published; no frozen “clean” badge.
4. Save as draft / publish — header and strip badges both update.
5. Close the editor tab, reopen — chrome resets from persistence; no stale badge from the previous session.
6. With a mid-edit draft, pop out or go fullscreen before the auto-save debounce fires — after docking back, the latest edit is still present (unmount flush).
7. Backend tracking present but local draft cleared/missing — strip shows unsynced, not a false clean published badge.

## Breaking changes

None.

## Out of scope

- Fullscreen/floating overwriting sibling tabs on dock-back — #1511 (`ui/fix-surface-tab-storage`).
- Multi-editor unique guide id minting / persisted id locks.


src/components/block-editor/utils/guide-id.ts
```diff
@@ -0,0 +1,32 @@
...
+ * Guide ID helpers for title → id derivation in the singleton editor.
+ *
+ * The ID is what `toResourceName` turns into the backend resource name. Title
+ * commit mints once while the guide is still on the default placeholder
+ * (`isIdLocked`); after that the id stays put. Multi-editor collision handling
+ * is intentionally out of scope here.
+ */
+
+/** Converts a guide title to a URL-safe kebab-case slug. */
+export function slugifyTitle(title: string): string {
```
Jayclifford345: This module is a straight lift of `slugifyTitle` / `generateUniqueId` out of `BlockEditor.tsx` plus a default `existingNames = []`, and nothing in the tab-chrome lifecycle touches it — the only caller is still the one `generateUniqueId(title, existingNames)` site at `BlockEditor.tsx:765`, unchanged. The PR body says the unique-id minting work was deliberately dropped from this PR, so this looks like the leftover of it.

Two small consequences:

- `slugifyTitle` is now exported but only used inside this file, so it reads as public API that nothing consumes.
- The extraction lands with no `guide-id.test.ts`, whereas `generateUniqueId`'s collision loop and the `Date.now()` fallback are exactly the kind of thing that was easier to leave untested while it was a private function in a big component.

Either fold the move into whichever PR actually needs the shared module, or keep it here and drop the `export` on `slugifyTitle`. Not blocking either way.

Jayclifford345 (approved):
Approving. The two-channel design (live publish + persisted fallback) is the right shape for chrome that outlives the component that owns the state, and the false-clean fix is a real bug caught.

**⚠️ CI has not run on this PR.** `.github/workflows/ci.yml` triggers on `pull_request: branches: [main, master]`, so a PR based on `ui/devtools-strip-visibility` gets no Lint / Typecheck / Test / Build / Prettier jobs — only the always-on scanners. Compare the check list here against #1509. I ran the missing ones locally on `8673e599`:

- `tsc --noEmit` — clean
- `npm run test:ci` — 396 suites, 6303 passed, 0 failed
- `npm run prettier-test` — clean
- `npm run lint` — 0 errors (24 pre-existing `no-deprecated` warnings)

Not a change request, just so it's on the record that the green checkmarks above don't mean what they usually mean. Same applies to #1511.

**Verified**

- `isChromeReady` can't deadlock. `readStoredEditorGuide()` and `useBlockPersistence`'s mount-restore effect read the same key and both bail on the same conditions, so if the former returns a guide the latter always reaches `onLoad` — and the `try/finally` you added unlocks even when `loadGuide` throws on a `blocks`-less draft. The `it.each` restore test plus the malformed-draft test pin both halves.
- The unmount ordering claim holds: `useBlockPersistence` is called earlier in `BlockEditorInner`'s body than the `resetEditorChromeStatus` effect, so React runs the draft flush cleanup first. It's an implicit coupling — a hook reorder silently breaks it — but the comment says so and `wakes the strip after the unmount draft flush, not before` fails if it regresses. Fine as is.
- `editor-chrome-status.ts` importing only `StorageKeys` does keep the lazy `BlockEditor` chunk out of the docs-panel main chunk, and the dependency runs the right way (`constants.ts` → `editor-chrome-status`, never the reverse), so no cycle.
- `flush()` on unmount is correctly bounded by `pendingSaveRef`: the no-change branch clears it, `clear()` clears it, and `autoSave: false` never sets it. The `writes nothing on unmount when no save is pending` and `cancels a pending write` tests cover both.
- Title sync is not per-keystroke — `HeaderTitleRow` only calls `onTitleCommit` on blur/Enter, so `updateEditorTabTitle` → `saveTabsToStorage()` fires once per rename, not once per character. Worth stating explicitly since `saveTabsToStorage` hits user storage.

**Nits, non-blocking**

- `docsPanel.editorTabModified` isn't in `src/locales/en-US/grafana-pathfinder-app.json`. Neither was the `editorDraftBadge` key you're replacing, so this is pre-existing drift rather than something you introduced, and `npm run check` doesn't gate it — but a `npm run i18n-extract` on this branch would pick up both the new key and the removal.
- Inline below on `utils/guide-id.ts`.

src/components/block-editor/utils/guide-id.ts
dgehring-grafana: I'll drop for now as I'd like to re-evaluate how we do internal ID management fundamentally. It might end up being the same thing, but less bloat is better for now.

src/components/full-screen/FullScreenPanel.tsx
```diff
@@ -244,12 +235,15 @@ function FullScreenPanelRenderer(_props: SceneComponentProps<FullScreenPanel>) {
...
-  const handleExitToSidebar = useCallback(() => {
+  const handleExitToSidebar = useCallback(async () => {
     reportAppInteraction(UserInteraction.FullScreenExit, {
       destination: 'sidebar',
       guide_url: guideUrl || '',
       guide_title: title,
     });
+    // Flush before the mode flip: the sidebar force-restores from tabStorage
+    // when it takes the workspace back, so anything unsaved here is lost.
+    await panel.saveTabsToStorage();
```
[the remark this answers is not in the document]
dgehring-grafana: Addressed with a gate that only attempts restore if tabs > 0 from fullscreen context. Closetab() already calls the sync, so if we truly do have no tabs, state will already reflect.

PR #1511, opened 2026-08-05
fix(docs-panel): stop surface handoffs from overwriting sibling tabs

dgehring-grafana:
## Summary

A prior deficient behavior: popping out the panel or switching to fullscreen could overwrite shared tab storage with the single active tab, clearing every other open tab. This change keeps the full tab workspace intact across surface handoffs so sibling tabs survive and come back when ownership returns.

Stacked on #1510 (`ui/editor-tab-chrome-lifecycle`).

## What to look at

- **Restore-before-open** (`pendingGuideRouter.ts` → `initializePanelTabsOnMount`) — consume pending synchronously for in-flight gating, restore the full strip, then apply the handoff. Opening first used to skip restore and let the next save erase siblings.
- **`tabId` on handoffs** — after restore the target tab already exists; focus it instead of appending a duplicate. Skip `setActiveTab` when already active so the restore-time load is not double-fetched.
- **Outbound flush** — await `saveTabsToStorage()` before sidebar/floating/fullscreen mode changes (including close/dock).
- **Sidebar reclaim** (`useTabRestoration.ts`) — `{ force: true }` when returning from floating/fullscreen; empty-strip-only hydrate on initial sidebar mount.
- **Taxonomy** — restore gating moves from `hasOnlyNonContentTabs` to `getGuideStripTabs` (empty strip).

## Alternatives considered

Not really. Without a larger refactor of how surfaces own tab state, the receiving view should align with the passed workspace. This change reinforces that: when state is handed back, or when the active surface mutates tabs, saves include the sibling context other views still need.

## Gotchas

Restore/hydrate gating migrates from `hasOnlyNonContentTabs` to “fetch/use guide strip tabs” (`getGuideStripTabs`). The original non-content flag is still useful for “no URL to fetch” rendering (e.g. remote URL load paths), but it was being conflated with legacy “permanent / chrome-only tabs.” Editor and Dev Tools are strip members now, so they are live strip state — not a reason to treat the strip as empty for restore.

## How to verify

1. Sidebar: open two guides (+ optional editor). Enter fullscreen on guide A — both guides remain in storage/strip context; A is focused once (no duplicate A tab).
2. In fullscreen, open or close a sibling (or rename) — Exit to sidebar — strip matches what fullscreen had (force restore), not the pre-fullscreen snapshot.
3. Fullscreen → floating and floating → fullscreen — same strip; no duplicate of the active guide.
4. Floating dock / close — sidebar reclaim shows floating’s tabs, not the pre-pop-out snapshot.
5. Pop out from sidebar with multiple tabs open — floating shows the active guide; dock back — siblings still present.
6. Home / auto-launch into floating (no `tabId`) — still opens a new tab after restore (genuine launch path unchanged).
7. Editor-only handoff into fullscreen — editor focuses via singleton path; siblings from storage still restored underneath.

## Breaking changes

None.

## Out of scope

- Multi-editor `createEditorTab({ tabId })` from `ui/add-multiple-guide`.
- Making `saveTabsToStorage` fail the mode flip on write errors (still best-effort / logged).

### DIFF
--- a/src/components/docs-panel/hooks/useTabRestoration.ts
+++ b/src/components/docs-panel/hooks/useTabRestoration.ts
@@ -1,8 +1,11 @@
 /**
- * Triggers tab restoration from storage when the sidebar instance is
- * actually responsible for owning the tab surface — gated by `panelMode`
- * and by the "only non-content tabs" predicate so we don't overwrite
- * user-opened guide/docs tabs on a remount (editor chrome alone is OK).
+ * Keeps the sidebar model aligned with the shared tab workspace.
+ *
+ * Initial mount restores only an empty strip. Returning from floating or
+ * fullscreen force-refreshes even a populated strip, because those surfaces
+ * own separate models and may have opened, closed, or renamed tabs while
+ * they held the workspace — without the refresh the sidebar keeps its
+ * pre-handoff snapshot and their work looks discarded.
  *
  * Why the dep array is `[panelMode]` only (preserved verbatim — Pattern J
  * boundary touching the `_hasRestoredTabs` instance guard, deferred to a
@@ -13,16 +16,17 @@
  *     re-fire on every tab open/close — but the in-class `_hasRestoredTabs`
  *     guard makes those re-fires no-ops, so it'd be wasted work, not a
  *     bug.
- *   - The non-content check reads `tabs` via closure capture from
+ *   - The strip-empty check reads `tabs` via closure capture from
  *     the render that registered the effect. That's safe because the
  *     restoration trigger only matters at the boundary where `panelMode`
  *     flips away from `'fullscreen'`. The `tabs` snapshot at that moment
  *     is whatever the renderer last rendered.
  *
- * Full-screen mode skip (preserved verbatim):
- *   When the full-screen panel owns the session, the sidebar instance
+ * Non-sidebar mode skip:
+ *   While floating or full screen owns the session, the sidebar instance
  *   must NOT call restoreTabsAsync — otherwise both instances race on
- *   tabStorage and drift the saved tab content.
+ *   tabStorage and drift the saved tab content. Those surfaces flush
+ *   storage before they hand the workspace back.
  *
  * Restore-once guard (preserved — Pattern I, deferred):
  *   `_hasRestoredTabs` lives on the model instance, not the hook. The
@@ -31,20 +35,17 @@
  *   guard into a hook would change StrictMode and fullscreen-remount
  *   semantics — see the deferred-work note in the refactor plan.
  *
- * Contract surfaces preserved (Pattern J — pinned by
- * docs-panel.tab-restore-guard.test.ts and utils/tab-storage-restore.test.ts):
- *   - hasOnlyNonContentTabs predicate (does NOT touch tabStorage)
- *   - model.restoreTabsAsync() entry point (unchanged)
- *   - `_hasRestoredTabs` guard semantics (untouched on the class)
+ * The model's restore-once guard still protects the initial StrictMode
+ * replay; `{ force: true }` is reserved for a surface ownership transition.
  */
 import * as React from 'react';
-import { hasOnlyNonContentTabs } from '../utils';
+import { getGuideStripTabs } from '../utils';
 import type { LearningJourneyTab, CombinedPanelState } from '../../../types/content-panel.types';
 import type { PanelMode } from '../../../global-state/panel-mode';
 
 interface TabRestorationModel {
   state: CombinedPanelState;
-  restoreTabsAsync(): Promise<void>;
+  restoreTabsAsync(options?: { force?: boolean }): Promise<void>;
 }
 
 export interface UseTabRestorationArgs {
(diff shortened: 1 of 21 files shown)


src/components/docs-panel/hooks/useTabRestoration.ts
```diff
@@ -54,17 +55,27 @@ export interface UseTabRestorationArgs {
...
+    previousModeRef.current = panelMode;
+
+    if (panelMode !== 'sidebar') {
+      return;
+    }
+
+    // Floating/fullscreen own separate models. Reconcile their saved strip
+    // when this long-lived sidebar model becomes the owner again.
+    if (previousMode !== 'sidebar') {
+      void model.restoreTabsAsync({ force: true });
```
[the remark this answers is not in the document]
dgehring-grafana: Added a `pendingTabsStorageWrite` flag to the docs-panel - now, if a write/save is in-flight for the tabs storage, the read mechanism will await it before giving results back. This should prevent a case where we can't await the side effect writes before switching to a new context - before, we'd swap immediately, now we still can, but we don't provide the new view's state until old manipulators finish their job.

Jayclifford345, line 72: Worth naming the cost of this explicitly somewhere, because it isn't in the PR description's verify list.

`restoreTabsAsync` replaces `state.tabs` wholesale with the objects `restoreTabsFromStorage` builds, and those carry `content: null`. So a force restore doesn't just reconcile the tab *set* — it drops the loaded content of every tab the sidebar already had. `initializeRestoredActiveTab()` refetches the active one; the siblings stay blank until the user clicks them, at which point `setActiveTab` triggers a fresh `loadTab`.

Functionally that's fine (progress lives in its own storage, so it's a spinner rather than data loss) and it's the same thing a page reload does. But it means every fullscreen/floating round trip now costs a refetch of the active guide plus a lazy refetch of each sibling, where before the sidebar just kept its snapshot. On a slow docs fetch that's a visible regression in perceived speed on a path users will hit repeatedly.

If that's an accepted trade, a line here saying so would stop the next person from treating it as a bug. If it's not, the narrower fix is to merge by id — keep `content` from the existing in-memory tab when the restored tab has the same `id` and `currentUrl` — rather than reconstructing the whole array.

Jayclifford345 (approved):
Approving. Restore-before-open is the right fix, and pairing it with `tabId` on the handoff is what makes it safe — reordering without the focus-by-id guard would have traded tab loss for tab duplication.

**⚠️ CI has not run on this PR** either (see #1510 — `ci.yml` only triggers on PRs targeting `main`). Ran the missing jobs locally on `3406db31`:

- `tsc --noEmit` — clean
- `npm run test:ci` — 396 suites, 6312 passed, 0 failed
- `npm run test:coverage` — thresholds met
- `npm run lint` — 0 errors (24 pre-existing `no-deprecated` warnings)
- `npm run prettier-test` — clean

**Verified**

- The `pendingTabStorageWrite` barrier can't deadlock a restore: `writeTabsToStorage` wraps everything in `try/catch` and `tabStorage.setTabs` / `setActiveTab` swallow their own errors, so the promise always settles. The identity check in `.finally` correctly avoids an older write retiring a newer one.
- `initializePanelTabsOnMount` keeps the consume synchronous with mount — `consumePendingGuide()` and `markInFlight()` both run before the first `await`, so the empty-state fallback still can't fire on top of an incoming guide. That was the load-bearing property of the old declaration-order arrangement and it survives the move.
- The gate swap from `hasOnlyNonContentTabs` to `getGuideStripTabs(...).length === 0` is strictly narrower, so the initial-mount path can't newly clobber anything. The lone-editor case it stops covering is picked up by the force path instead.
- The `tabId` early return in `openPendingGuide` skips `openDocsPage`, which means it also skips `_recordAutoLaunchSource(source)`. That looked like an alignment-prompt regression at first, but it isn't: `initializeRestoredActiveTab` tags the load `browser_restore`, and `browser_restore`, `fullscreen_handoff` and `floating_panel_dock` are all in `ALIGNED_BY_CONSTRUCTION_SOURCES`, so the classification is unchanged. Might be worth a line in the code saying so — the next reader will have the same question.
- Outbound flushes are snapshot-correct: `writeTabsToStorage` builds `tabsToSave` synchronously before its first `await`, so calling `saveTabsToStorage()` before `setPendingGuide` and awaiting after doesn't race the state it's capturing.

**Worth flagging**

1. **Stack is stale.** #1509's tip is `1faa1711` (a merge of `main`), but this branch and #1510 both fork from `22521e67`, the commit *before* that merge — so `git merge-base --is-ancestor origin/ui/devtools-strip-visibility origin/ui/editor-tab-chrome-lifecycle` is false. All three are also 2 commits behind current `main`. Concretely, my local run here was missing four test files that exist on #1509 (`toggle-state`, `toggle-click`, `StepEditor`, `content-renderer.prop-forwarding`). Restack before merging, and given the CI gap above it's worth pushing the restacked branches and re-running locally rather than assuming.

2. **Force restore discards loaded content for every sidebar tab**, not just the reconciled ones — inline below.

3. **The write barrier only covers the newest in-flight write.** `await pendingTabStorageWrite` evaluates the module variable once; a save issued *after* the restore starts awaiting isn't waited on. Every path that matters awaits its own save before flipping mode, so the barrier is only load-bearing for the two fire-and-forget cases you call out in the comment — and it covers those. Just noting the guarantee is "the newest write at the moment restore began", which is narrower than the docblock's "a handover can never restore the pre-handoff strip" reads.

src/components/docs-panel/hooks/useTabRestoration.ts
dgehring-grafana: I think a merge is a non-invasive and fine behavior to add - will append via a commit, along with some more specific docs changes on the pendingTabsStorageWrite

PR #1514, opened 2026-08-06
fix(settings): preserve provisioned jsonData fields on plugin settings save

Jayclifford345:
## Summary

- Fixes a bug where saving plugin settings would clear provisioned fields like `stackId` from Grafana Cloud
- Ensures all existing `jsonData` fields are preserved when saving settings from any configuration tab

## Problem

When saving plugin settings, the frontend was losing provisioned fields (like `stackId`) that are set by Grafana Cloud. This happened because:

1. `getConfigWithDefaults()` only returns fields defined in `DocsPluginConfig`
2. Grafana's plugin settings API (`POST /api/plugins/{pluginId}/settings`) does a **full overwrite** of `jsonData`, not a merge
3. Any field not explicitly included in the save payload gets deleted from the server

### Impact

Without `stackId`, the backend plugin's OBO (On-Behalf-Of) token exchanger fails to initialize, causing:
- Private guides to become unavailable (`obo-unavailable` error)
- Users needing to wait for settings to be re-provisioned or manually intervene

## Solution

Spread the original `jsonData` first to preserve ALL existing fields before applying defaults and form-specific overrides:

```typescript
const newJsonData = {
  ...(jsonData || {}),                        // 1. Preserve ALL existing fields (incl. provisioned)
  ...getConfigWithDefaults(jsonData || {}),   // 2. Apply defaults for known fields  
  // 3. Override with form-specific values...
};
```

This pattern was applied to all 5 locations where settings are saved:
- `ConfigurationForm.tsx` (3 locations)
- `TermsAndConditions.tsx` (1 location)
- `InteractiveFeatures.tsx` (1 location)

## References

- [Grafana Plugin Tools: Add authentication for app plugins](https://grafana.com/developers/plugin-tools/how-to-guides/app-plugins/add-authentication-for-app-plugins)
- [Grafana source: pkg/api/plugins.go](https://github.com/grafana/grafana/blob/main/pkg/api/plugins.go)

### DIFF
--- a/src/components/AppConfig/ConfigurationForm.tsx
+++ b/src/components/AppConfig/ConfigurationForm.tsx
@@ -155,6 +155,7 @@ const ConfigurationForm = ({ plugin }: ConfigurationFormProps) => {
         enabled,
         pinned,
         jsonData: {
+          ...(jsonData || {}),
           ...getConfigWithDefaults(jsonData || {}),
           enableAssistantDevMode: newValue,
         },
@@ -265,6 +266,7 @@ const ConfigurationForm = ({ plugin }: ConfigurationFormProps) => {
           enabled,
           pinned,
           jsonData: {
+            ...(jsonData || {}),
             ...getConfigWithDefaults(jsonData || {}),
             codaRegistered: true,
             codaApiUrl: apiUrl,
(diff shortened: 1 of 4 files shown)


jdbaldry (approved):
The regression test shows that this seems to fix at least one situation in which this can happen. I _think_ it depends on the jsonData of the current page load and not the live server data but I'm not sure if there's any precedent for a compare-and-swap-like behavior that other plugins have used. Might be worth investigating after this workaround is in place.

PR #1529, opened 2026-08-07
fix(openfeature): validate remote experiment variant before enrolling

### LINKED ISSUE #1520: Unchecked cast on remote feature-flag variant enrolls users into nonexistent experiment arms
## Bug description

`HighlightedGuideConfig.variant` is read from a remote feature-flag payload and cast without a membership check. A typo'd or renamed remote variant value (say `"treament"`) is not `'excluded'`, so the user is enrolled into an experiment arm that does not exist, the sidebar auto-opens, a once-per-browser localStorage marker is written for the bogus arm, and analytics reports it verbatim.

## Root cause

`src/utils/openfeature.ts:551-577`. Note that `docType`, **eleven lines above**, is correctly validated against a `ReadonlySet` — and `variant`, from the same untrusted payload, is not:

```ts
const VALID_DOC_TYPES: ReadonlySet<HighlightedGuideDocType> = new Set([...]);
const docType =
  typeof record.docType === 'string' && VALID_DOC_TYPES.has(record.docType as HighlightedGuideDocType)
    ? (record.docType as HighlightedGuideDocType)   // ← checked
    : undefined;

return {
  variant: record.variant as HighlightedGuideConfig['variant'],   // ← :570, not checked
  ...
};
```

The guard at `:557` only establishes `typeof record.variant === 'string'`, not that it is one of `'excluded' | 'control' | 'treatment'`.

Downstream, `getActiveExperiments()` (`:591`) branches on `config.variant === 'excluded' ? [] : [...]`, so an unrecognized value falls into the enrolled branch. `src/utils/experiments/highlighted-guide-orchestrator.ts` then auto-opens the panel and writes `StorageKeys.HIGHLIGHTED_GUIDE_AUTO_OPEN_PREFIX`.

## Suggested fix

Mirror the `docType` treatment directly above it:

```ts
const VALID_VARIANTS: ReadonlySet<HighlightedGuideConfig['variant']> =
  new Set(['excluded', 'control', 'treatment']);

if (!VALID_VARIANTS.has(record.variant as HighlightedGuideConfig['variant'])) {
  return null;   // caller falls back to DEFAULT_HIGHLIGHTED_GUIDE_CONFIG
}
```

Roughly four lines in one file, fully mechanical. Worth adding a unit test covering an unrecognized variant string.

## Notes

The general shape — an unchecked `as` on a value that just crossed a trust boundary — is the pattern epic #603 Phase 7 targets. This instance is unusual in that the correct pattern is already present in the same function, which makes it a clean fix rather than a design question.

## How this was found

Part of a five-agent type-design investigation commissioned against epic #603.

moxious:
## Summary

A remote `pathfinder.highlighted-guide-experiment` payload carrying an unrecognized `variant` — a typo'd `treament`, a renamed arm, an empty string — enrolled the user into an experiment arm that does not exist. `validateHighlightedGuideValue` membership-checked `docType` against a `ReadonlySet` but cast `variant` from the same untrusted payload with a bare `as` eleven lines later, so the value reached `getActiveExperiments()`, which only special-cases `'excluded'`. Everything downstream then fired: the sidebar auto-opened, a once-per-browser localStorage marker was written for the bogus arm, and the bogus arm rode along verbatim on the `experiments` property of every analytics event and on the Faro session's `experiments` attribute.

The fix is a `VALID_VARIANTS` set plus an early `return null`, so a rejected remote payload falls back to `DEFAULT_HIGHLIGHTED_GUIDE_CONFIG` (`variant: 'excluded'`).

## What to look at

- [`src/utils/openfeature.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/ab3a4cfd/src/utils/openfeature.ts) — the rejection is **whole-payload, not field-level**, which is the opposite of how `docType` is handled directly above it: an invalid `docType` is dropped and the rest of the payload survives, whereas an invalid `variant` discards the whole config. The asymmetry is deliberate. A bad `docType` is cosmetic — it only picks which card type the Featured slot renders, so falling back to `findDocPage`'s inference is strictly better than discarding a working experiment. A bad `variant` is an enrollment decision, and there is no safe way to partially honor a payload whose arm assignment is unintelligible. Whole-payload rejection also matches the two rejection paths already in the function. The `ReadonlySet<HighlightedGuideConfig['variant']>` annotation is load-bearing rather than decorative — it makes a typo inside the set itself a compile error (introducing `'treatmnt'` produces TS2322), which is what keeps the hand-written list from drifting from the union. It does not catch an *omission* from the set, which is tracked separately.
- **The two sources fall back differently, and this is the thing to get right.** A rejected *remote* payload yields the default config. A rejected *localStorage override* is **ignored in favor of the remote MTFF value** — it does not fall back to the default. Locally there is no MTFF provider so the client returns the default we pass it, which is why this looked like the same thing; on a Cloud stack with a live flag, a QA engineer with a typo'd variant gets the *remote* arm's guide auto-launched instead of theirs. The fallthrough itself is pre-existing behavior shared with the two older rejection paths and is left alone; what this PR adds is a warn so it is discoverable.
- **Rejection now logs**, once per source per page load, with a low-cardinality `reason` (`unknown_variant` | `invalid_shape`) and never the raw remote string. `getActiveExperiments` re-reads this flag on every `reportAppInteraction`, so the guard is not optional — an unguarded warn would flood the console and the Faro log stream. Channel is `logger.warn` rather than a new typed facade op: `TELEMETRY.md`'s decision rule reserves `pushFaroEvent`-backed ops for multi-tier degradation ladders, this is a single binary rejection, and the logger bridge already mirrors every warn into Faro as a queryable log.
- [`docs/developer/FEATURE_FLAGS.md`](https://github.com/grafana/grafana-pathfinder-app/blob/ab3a4cfd/docs/developer/FEATURE_FLAGS.md) and [`EXPERIMENT_TESTING.md`](https://github.com/grafana/grafana-pathfinder-app/blob/ab3a4cfd/docs/developer/EXPERIMENT_TESTING.md) — the operator-facing consequence: renaming an arm in MTFF turns it off rather than half-enrolling its cohort under a bogus label. Both now distinguish the two fallback paths, and both note that `resetCache` is discarded along with the rest of a rejected payload — `handleHighlightedGuideResetCache` reads it *before* any variant check, so "rename an arm and set `resetCache: true`" clears nothing.

**Exposure events are not part of this.** An earlier draft of this description claimed a rejected payload now produces no exposure event. That was wrong: `reportFeatureFlagExposure` only tracks `control` and `treatment` (`openfeature-tracking.ts:44`), so an unrecognized arm never produced a `pathfinder_feature_flag_evaluated` event before validation either. There is no delta here, which also means **exposure counts are not a signal that a flag payload is broken** — the new warn is. What the fix does stop is the bogus arm reaching the `experiments` analytics property and the Faro session attribute, both of which carried it verbatim.

## How to verify

1. Reproduce the old behavior on `main`: in DevTools, run `__pathfinderExperiment.setOverride('pathfinder.highlighted-guide-experiment', { variant: 'treament', pages: ['/a/grafana-irm-app*'], guideId: 'bundled:<some-guide>' })`, then reload onto a matching page. The sidebar auto-opens and `__pathfinderExperiment.config.variant` echoes the typo.
2. On this branch, repeat with the same misspelled variant. On a local OSS stack nothing happens; on a Cloud stack whose flag is live you get the *remote* arm instead. Either way the console shows `[OpenFeature] Rejected the override payload for 'pathfinder.highlighted-guide-experiment'` with `{ reason: 'unknown_variant' }`, exactly once no matter how much you click around. `__pathfinderExperiment.showOverrides()` still shows the override is stored — it is being read and rejected, not dropped.
3. Confirm the once-per-browser marker was not written for the bogus arm: check `localStorage` for a key under the highlighted-guide auto-open prefix keyed `{hostname}:{guideId}`.
4. Swap `variant` to `'treatment'` and reload — auto-open, the Featured-slot injection, and the exposure event should all behave exactly as before this change, and nothing should warn.

## Breaking changes

None for any valid configuration; a dedicated test asserts each of the three arms passes through byte-identical. Behavior changes only for payloads that were already invalid, where an unrecognized variant is now a no-op instead of an enrollment. Two consequences for an already-invalid payload worth naming: `resetCache: true` is no longer honored (it is discarded with the rest), and `window.__pathfinderExperiment.config` now reports `excluded` rather than echoing the bad variant — the warn replaces that as the diagnostic. Operators should know that renaming an arm in MTFF disables it rather than silently relabeling it.

## Reversibility

Fully reversible by revert — no storage format, event schema, or flag contract changed. One asymmetry worth naming: the localStorage auto-open markers this change prevents from being written are not cleaned up retroactively, so any browser that already recorded a marker under a bogus arm keeps it. Because that marker is keyed `{hostname}:{guideId}` with no variant component, a cohort already auto-opened under a bogus arm will not auto-open again if the operator fixes the typo while keeping the same `guideId`. This is pre-existing damage from the bug rather than something the fix introduces, and `resetCache: true` clears all markers for the hostname if an operator wants a clean slate.

Fixes #1520


moxious:
## Reviewer note: this PR makes an implicit decision about localStorage overrides

An independent annotation on #1520 raised a scope detail that this PR resolves silently. Flagging it so the decision is made deliberately rather than inherited.

`validateHighlightedGuideValue` is shared by **two** callers: the remote flag value and the localStorage override used for QA and demos. The new variant check sits in that shared validator, so it applies to both.

- **Remote path** — invalid variant returns `null`, caller falls back to `DEFAULT_HIGHLIGHTED_GUIDE_CONFIG`. This is the intended fail-closed fix and has no effect on valid configurations.
- **Override path** — invalid variant returns `null`, and the caller then *falls through to the live OpenFeature provider*. So a QA engineer who typos `treatement` in their override does not get an error or the default; they silently get the real remote assignment.

That fallthrough is pre-existing behavior, and `openfeature.test.ts` already enshrines it for a missing `guideId` (`getHighlightedGuideConfig should NOT fire exposure when override is invalid and falls through`). This PR extends the same behavior to a new rejection reason rather than inventing one, which is the conservative choice.

The wrinkle: `docs/developer/EXPERIMENT_TESTING.md` states that overrides "bypass MTFF entirely." Silently falling back to the remote assignment is not a bypass. The doc and the code disagreed before this PR; this PR widens the set of inputs that hit the disagreement.

**Options, if a reviewer wants to change it:**
1. Keep as-is (current state of the PR) — consistent with the existing test, doc stays inaccurate.
2. Make invalid overrides fail closed to the default instead of falling through — arguably matches the doc's intent, but changes existing tested behavior and needs the test updated.
3. Keep as-is and fix the doc.

I would take (1) for this PR and (3) as a separate docs change, since (2) is a behavior change beyond the stated zero-functional-change scope. But it should be a choice, not an accident.

## Minor correction to the linked issue's impact claim

The issue body overstates the analytics blast radius. `reportFeatureFlagExposure` independently allowlists `control` and `treatment`, so a bogus arm does **not** emit `pathfinder_feature_flag_evaluated` or write its exposure-dedup marker. The bad value reaches general analytics/Faro experiment enrichment via `getActiveExperiments`, and the localStorage marker actually written by the product flow is the separate highlighted-guide auto-open marker. The bug and this fix are unchanged; only the described consequence was too broad.

tomglenn (dismissed):
Variant-enrollment hole closed and fails safe (unknown arm → excluded, no cohort attached). Nit for a follow-up: `pages[]` elements are still uncast.

tomglenn (approved):

Issue #1530, opened 2026-08-07
json-guide-format.md omits the challenge and snippet-ref block types


moxious:
## Summary

`docs/developer/interactive-examples/json-guide-format.md` is the authoritative schema reference for the guide format, but its "Block Types Summary" table and per-type `####` sections omit **`challenge`** and **`snippet-ref`** entirely.

Both have been real, parseable block types in `JsonBlockSchema` for a while (since #875 and the snippet work respectively) — this predates the block-type totality ratchet and is not fallout of it.

## Why it became more urgent

The block-type ratchet PR fixed a bug where `challenge` was absent from `BLOCK_TYPE_GROUPS` and therefore **unreachable in the block-editor palette**. Now that authors can actually insert a challenge block, they will encounter a block type with no entry in the schema reference.

So the ordering is now backwards: the block is discoverable in the UI but undocumented in the reference authors are pointed at.

## What is needed

Two per-type sections plus two summary-table rows.

**`challenge`** — the larger of the two, ~14 fields. Worth getting right because several are non-obvious:
- `mode` (`'coda' | 'standard'`) — this is the field that determines everything else. `'standard'` runs against the learner's own Grafana and provisions nothing; `'coda'` provisions a VM and provides a terminal. Note `ChallengeBlockForm` defaults new blocks to `'standard'`.
- `successCriteria` — a requirement token, typically `coda-exit-zero:<command>` in coda mode, any Pathfinder requirement in standard mode.
- `vmTemplate` / `vmScenario` / `vmApp` — ignored when `mode` is `'standard'`.
- `setupScript` vs `setupCommands` — `setupCommands` is deprecated; prefer `setupScript`. Worth stating explicitly so authors do not copy the deprecated form.
- `hintLevels`, `failureMessage`, `brief`, `title`, plus the common `requirements` / `objectives` / `skippable`.

**`snippet-ref`** — small: `snippetId` plus the common fields. Worth documenting the constraint that snippets cannot contain refs to other snippets (enforced by `NonRecursiveBlockSchemaNoRef` in `json-guide.schema.ts`).

Source of truth for both is `src/types/json-guide.schema.ts` — the `.describe()` strings on each field are already written and can largely be lifted directly.

## Note on the underlying drift

[...]

Issue #1523, opened 2026-08-07
Make block-conversion registries total over BlockType (closes the class behind #619)


moxious:
## Summary

Direct follow-up to the block-type totality ratchet (see epic #603 analysis comment). That PR made `KNOWN_FIELDS`, `BLOCK_TYPE_METADATA`, `FORM_COMPONENTS` and the palette groups total over `JsonBlock['type']`. Two registries in `block-conversion.ts` were **deliberately left out** because making them total requires per-type design judgment rather than mechanical fallout.

## The gap

`src/components/block-editor/utils/block-conversion.ts:51,74`:

```ts
const CONTENT_FIELDS: Partial<Record<BlockType, string>> = { ... };
const REQUIRED_DEFAULTS: Partial<Record<BlockType, Record<string, unknown>>> = { ... };
```

`Partial<>` means a missing key is legal, and a missing key is a runtime `undefined` that surfaces as a failed `safeParse` and an error toast — with no compile-time signal.

This is the exact root cause of #619 (`image`/`video` → `terminal-connect` threw because `REQUIRED_DEFAULTS` had no `terminal-connect` entry). It was fixed by adding the one missing key, which leaves the shape that produced it intact.

`CONVERTIBLE_TYPES` (`:26`) is a third hand-maintained list — a new block type simply never appears as a conversion target, silently.

`getConversionWarning` (`:125-131`) has a related early return:

```ts
const targetKnownFields = KNOWN_FIELDS[targetType];
if (!targetKnownFields) {
  return null;   // no warning computed
}
```

Before the ratchet PR this branch was reachable for `challenge`, silently disabling data-loss warnings when converting to it. It is now unreachable for real block types, but the shape remains.

## Why it wasn't mechanical

Making these total means answering, per block type, questions that are genuine product decisions:

- What is the "content field" of a `section` or `grot-guide`? Probably none — but that should be an explicit `null`, not an omission.
- What is a sensible `REQUIRED_DEFAULTS` entry for `assistant`? Is `assistant` convertible at all?
- Should `snippet-ref` be convertible? It is currently in neither `CONVERTIBLE_TYPES` nor `CONTAINER_TYPES` — it just falls through.

## Suggested approach

Use the explicit-null pattern so "no content field" is a stated decision rather than an omission:

```ts
const CONTENT_FIELDS: Record<BlockType, string | null> = { ... };
[...]

Issue #1522, opened 2026-08-07
Malformed guide content makes "start" a silent dead click (unvalidated parse in prepareGuideLaunch)


moxious:
## Bug description

`prepareGuideLaunch` parses guide content with a bare cast and immediately walks `guide.blocks`. On the content tiers that are not schema-validated, a malformed guide throws inside an async function whose rejection isn't handled at that site — the caller commits no surface and **the docs panel shows a spinner indefinitely with no error text**.

## Root cause

`src/components/docs-panel/utils/prepare-guide-launch.ts:80-93`:

```ts
let guide: JsonGuide;
try {
  guide = JSON.parse(rawContent.content) as JsonGuide;   // ← :82
} catch { ... }
// comment at :84 claims "fetchContent validates native JSON guides"
const { guide: expandedGuide, unresolvedSnippetIds } = await inlineSnippetRefsInGuideWithStatus(guide);
```

The `try/catch` catches a `JSON.parse` throw. It does not catch **syntactically valid but structurally wrong** content, which is the actual failure mode.

The comment's claim holds for only one of the fetch tiers. `src/docs-retrieval/content-fetcher.ts:230` validates only when `isNativeJson === true`. The other paths do not:

- `wrapContentAsJsonGuide()` (`content-fetcher.ts:35-56`) synthesizes a guide object with no validation.
- `injectJourneyExtrasIntoJsonGuide()` (`content-fetcher/cover-page.ts:121-144`) re-parses with its own bare cast and re-emits a spliced object.
- `content-fetcher/bundled.ts:49-62` returns raw localStorage bytes after parsing only far enough to sniff `.title` — this is the E2E-injection and WYSIWYG-preview path.

If `blocks` is absent, `inlineSnippetRefsInGuideWithStatus(guide)` throws on the walk. `requiresGrafanaUi(expandedGuide)` at `:93` has the same exposure.

## Suggested fix

`JsonGuideSchema` is already imported transitively on this path and the function already has a typed failure return the caller renders, so the fallback is free:

```ts
const validated = JsonGuideSchema.safeParse(safeJsonParse(rawContent.content));
if (!validated.success) {
  logger.error('[PrepareGuideLaunch] Guide failed schema validation', {
    url,
    issue: validated.error.issues[0]?.message,
  });
  return { ok: false, error: 'Guide content is malformed and cannot be opened' };
}
const guide = validated.data as JsonGuide;
```

[...]

Issue #1521, opened 2026-08-07
Upstream Coda 401 surfaces as 500 on /sample-apps and /alloy-scenarios (error classification keyed on message substrings)


moxious:
## Bug description

An expired Coda token surfaces to the browser as **500 Internal Server Error** instead of 401 on `/sample-apps` and `/alloy-scenarios`, so the frontend cannot tell "you need to re-register" from "the backend is broken".

## Root cause

Handler-level error classification is done by substring-matching a formatted error message. `pkg/plugin/resources.go:319` and `:346`:

```go
if strings.Contains(err.Error(), "authentication failed") {
    a.writeError(w, err.Error(), http.StatusUnauthorized)
} else {
    a.writeError(w, err.Error(), http.StatusInternalServerError)
}
```

`CreateVM` produces a message that matches, because it special-cases the status (`pkg/plugin/coda.go:296`):

```go
if resp.StatusCode == http.StatusUnauthorized {
    return nil, fmt.Errorf("authentication failed: token may be invalid or expired, please re-register")
}
```

`ListSampleApps` (`coda.go:559`) and `ListAlloyScenarios` (`coda.go:529`) have **no such branch**. They fall through to the generic:

```go
if resp.StatusCode != http.StatusOK {
    bodyBytes, _ := io.ReadAll(resp.Body)
    return nil, fmt.Errorf("unexpected status code %d: %s", resp.StatusCode, string(bodyBytes))
}
```

`"unexpected status code 401: ..."` does not contain `"authentication failed"`, so the handler misclassifies it as a 500.

**Nuance:** a *local* token-refresh failure does return `"authentication failed: %w"` via `setAuthHeader` and is classified correctly. It is specifically the **upstream 401** path on these two routes that is misrouted.

## The underlying design problem

`strings.Contains(err.Error(), "authentication failed")` appears **six times** in `resources.go` (`:218`, `:242`, `:269`, `:292`, `:319`, `:346`), plus `"not found"` at `:240`. Control flow — and therefore the HTTP status the browser sees — is keyed on the exact prose of a `fmt.Errorf` string defined in another file. Rewording any of those messages silently changes status codes.

The same shape appears in `isSSHAuthError` / `isSSHRetryableError` (`stream.go:262-290`) and `isDeadSessionError` (`coda_exec.go:222-236`), which match against third-party SSH error text.

## The fix already exists in this repo

[...]

PR #1531, opened 2026-08-07
fix(block-editor): make the challenge block reachable and ratchet block-type registries total

moxious:
## Summary

The `challenge` block could not be inserted from the block-editor palette. It had a schema, metadata, and a working form — but it was missing from `BLOCK_TYPE_GROUPS`, and `BlockPalette` renders by mapping that list, so it was unreachable **from the palette** with no other symptom. (It was not wholly unreachable: `CONVERTIBLE_TYPES` includes `challenge`, so "Switch type" on any non-container block already produced one. See "Behavior changes" below — that route lands in the *worse* default.)

Root cause is not the missing entry. The block-type union has several hand-maintained projections and nothing forced them to stay total, so members fell out silently: `challenge` was absent from `KNOWN_FIELDS`, `VALID_BLOCK_TYPES`, and `BLOCK_TYPE_GROUPS`; `assistant` was absent from the editor's hand-written `BlockType`. This PR derives those projections from the schema union and adds compile-time totality checks, then fixes what the checks caught.

## What to look at

- **`src/types/json-guide.schema.ts`** — `KNOWN_FIELDS` gains a trailing `satisfies Record<JsonBlock['type'] | meta, ReadonlySet<string>>`. The public type stays `Record<string, …>` on purpose: callers index it with an unvalidated `block.type` off untrusted JSON, so narrowing it would force casts at every call site. The `satisfies` clause buys totality without that churn, matching the pattern already used throughout `package.schema.ts`. It ratchets both directions — a missing key *and* a typo'd extra key are both compile errors. `VALID_BLOCK_TYPES` is now derived from `KNOWN_FIELDS` rather than restated.
- **`src/components/block-editor/constants.ts`** — the load-bearing change. `BLOCK_TYPE_ORDER` is no longer hand-maintained; it is `BLOCK_TYPE_GROUPS.flatMap(...)`, so palette membership has exactly one source. `_palettePartitionIsTotal` is a compile-time assertion that every `BlockType` is either grouped or in `PALETTE_EXCLUDED_BLOCK_TYPES`; when it fails, the error message names the offending block type.
- **`src/components/block-editor/BlockFormModal.tsx`** — `FORM_COMPONENTS` becomes total with an explicit `assistant: null`. The `!FormComponent` guard that consumes it already existed and was already live at runtime for `assistant` blocks (the `Record<BlockType, …>` annotation lied, because the hand-written `BlockType` omitted `assistant`, so the lookup returned `undefined`). This PR makes the type honest; the behavior is unchanged.
- **`src/cli/utils/block-registry.ts`** — `challenge` added to `CLI_EXCLUDED_BLOCK_TYPES`. This is the ratchet forcing a decision, not an oversight: the CLI registry-completeness test previously anchored on a `VALID_BLOCK_TYPES` that was itself missing `challenge`, so it never saw the gap it exists to catch. The recorded reason is `hintLevels` — `z.array(z.object(...))` is `unsupported` to the schema-options bridge and gets dropped, so `add-block challenge` would silently produce a hintless challenge. An `add-hint` sibling command is the prerequisite for registering it; every other field projects fine. Tracked in #1540.
- **`src/validation/type-coupling.test.ts`** — `verifyFieldsFromEffects` unwrapped a single `.refine()` layer and silently skipped when it found no object shape, so it could pass green while asserting nothing. It now unwraps recursively and throws. Worth a close look: this is the same failure mode the rest of the PR is about, and it was worse than it looked — the single-unwrap predicate skipped **all eight** schemas routed through it (`markdown`, `interactive`, `guided`, `quiz`, `assistant`, `input`, `grot-guide`, `_step`). That half of the suite was asserting nothing at all. It passes now, so `KNOWN_FIELDS` happened to be correct for those types.

## Why

Documented constraints drift; compiler errors do not. A registry keyed by a union but typed `Record<string, …>` or `Partial<Record<…>>` accepts an incomplete projection silently, and every consumer then needs a runtime guard that fails open. `detectUnknownFields` is the clearest case — `if (!knownFields) { return warnings; }` meant a typo'd optional field on a `challenge` block passed `validate --strict` with no warning at all.

The failure this PR is really aimed at is the second-order one: a guard existed for exactly this class and was blind to it, because it was anchored to another hand-maintained list. Deriving beats restating.

## Behavior changes

An earlier revision of this description claimed no behavior change. That was wrong. Giving `challenge` a `KNOWN_FIELDS` entry changes two runtime paths in `block-conversion.ts` that were previously short-circuited by `if (!targetKnownFields) { return null; }` (`:129`) and `if (targetKnownFields)` (`:205`). Both are repairs of the silent-data-loss class #1523 and #804 describe, arrived at incidentally rather than by design:

1. **Switching any block type → `challenge` now raises a data-loss confirmation** where it previously switched silently. Verified: `quiz → challenge` now reports `lostFields: ['choices', 'multiSelect']`; before, `getConversionWarning` returned `null` and `handleTypeSwitchRequest` took the no-confirmation branch.
2. **`convertBlockType(*, 'challenge')` now preserves fields shared with the target**, notably `id` — which `resolveStepId` uses for completion tracking, so the old behavior could orphan learner completion records — and `authorNote`. Previously the shared-field copy was skipped entirely.

Both now have tests in `block-conversion.test.ts`; neither did before.

3. **Cosmetic:** `ALLOWED_BRANCH_BLOCK_TYPES` (`BranchBlocksEditor.tsx:298`) is derived from `BLOCK_TYPE_ORDER`, so the option *order* in the conditional-branch "Block type" combobox changed. Same 14 members, regrouped. The palette itself is unaffected — it orders by `BLOCK_TYPE_GROUPS`, and I diffed the rendered sequence before and after: identical apart from the new Challenge entry.

Worth noting the new palette route is the *safer* one. A block created from the palette gets `mode: 'standard'` (`ChallengeBlockForm.inferInitialMode`, `!block → 'standard'`), which runs against the learner's own Grafana with no VM. The pre-existing "Switch type" route produces a block with no `mode` and no Coda fields, which `inferInitialMode` resolves to **`'coda'`**. Making the block palette-insertable moves the default in the safe direction.

No stored data, schema shape, or wire format is altered; `satisfies` and the partition assertion erase at runtime. `challenge` becoming insertable is the bug fix.

## How to verify

1. On `main`, open the block editor and open the block palette. Confirm there is no **Challenge** entry under any group heading.
2. On this branch, repeat. Confirm **Challenge** now appears under **Interactive**, and that selecting it opens the challenge form defaulted to **standard** mode.
3. Confirm nothing else moved in the palette: the group headings, their order, and the blocks within each group should be unchanged apart from the new Challenge entry. (`BLOCK_TYPE_ORDER` changed from a hand-written list to a derived one — this step is what confirms the derivation reproduces the previous palette ordering.)
4. Exercise the palette ratchet: remove `'challenge'` from the `interactive` group in `constants.ts` and run a typecheck. Expect an error naming `"challenge"` in the `_palettePartitionIsTotal` assertion. Revert.
5. Exercise the registry ratchet, both directions: delete the `challenge` entry from `KNOWN_FIELDS` and typecheck (expect `Property 'challenge' is missing`); then add a bogus key and typecheck (expect `'bogusTypo' does not exist in type`). Revert.
6. Author a guide containing a `challenge` block with a typo'd optional field (e.g. `hintLevls`) and run `validate --strict`. Expect an unknown-field warning; before this change the check returned early and reported nothing.
7. Run the CLI `add-step` command with `--id`, then `validate --strict` on the result. Expect no `Unknown field 'id'` warning — the schema has always accepted step ids, but `KNOWN_FIELDS._step` never listed one. (Latent: no guide in the repo currently has a step `id`.)
8. Open an existing non-container block, use **Switch type** → **Challenge**, and confirm the data-loss dialog now appears and that the block's `id` survives the conversion.

## Breaking changes

None. See "Behavior changes" for the two runtime deltas — both replace silent data loss with a confirmation or a preserved field, so no existing guide is invalidated and no author flow is removed.

## Out of scope

- **`block-conversion.ts` totality** (`CONTENT_FIELDS`, `REQUIRED_DEFAULTS`, `CONVERTIBLE_TYPES` are still `Partial<Record<…>>` / hand-maintained). Deliberately excluded — making them total requires per-type product judgment (is `assistant` convertible? what is the content field of a `section`?) rather than mechanical fallout. Tracked in #1523, which also documents the hardcoded-count test that pins the current incomplete list (`block-conversion.test.ts:45-48`) and the `image`/`video`/empty-`markdown` → `challenge` conversions that currently throw.
- **`json-guide-format.md` schema sections for `challenge` and `snippet-ref`.** Both predate this change; `challenge` became more urgent because it is now insertable, and `docs/developer/components/block-editor/README.md` now routes developers there for per-type schemas. Reviewed and accepted as a fast-follow rather than a blocker: the new `docs/sources/block-editor/_index.md` content covers the author who clicks the palette entry. Tracked in #1530.
- **Gating the Coda VM mode option on `enableCodaTerminal`** (#1541) — pre-existing, and this PR moves the default away from it. **`BranchBlocksEditor` silently producing a markdown block when Challenge is picked** (#1542) — pre-existing, `challenge` was already in the old hand-written `BLOCK_TYPE_ORDER`. **`ChallengeBlockForm` dropping five fields on save** (#1543).
- Other latent bugs found by the same analysis are filed separately (#1519, #1520, #1521, #1522) rather than bundled here.

Refs #603, #1523, #1530, #1540, #1541, #1542, #1543
Relates to #619, #804


moxious:
## Reviewer note: one unflagged ordering change

Checked this branch against the independent annotation added to #1523 after the PR was opened. One of its warnings applies here, and one side effect is not covered by the PR body.

**Not violated — the important one.** The annotation warns: *"Do not derive conversion targets from palette membership/order. The policies intentionally differ: `html` is palette-excluded but currently offered as a conversion target, while `snippet-ref` is in the palette but is not a target."* This branch does **not** touch `block-conversion.ts`; `CONVERTIBLE_TYPES` is untouched and conversion policy is unchanged. Confirmed by diff.

**Unflagged side effect.** The review round replaced the hand-maintained `BLOCK_TYPE_ORDER` with `BLOCK_TYPE_GROUPS.flatMap(...)`. Membership is identical (same 17 types), but **the order changed**:

```
before: markdown image video section collapsible conditional interactive multistep
        guided terminal terminal-connect challenge code-block quiz input grot-guide snippet-ref

after:  markdown image video code-block | interactive multistep guided input quiz
        terminal terminal-connect challenge | section collapsible conditional grot-guide | snippet-ref
```

For `BlockPalette` this is inert — it renders by mapping the groups and uses `BLOCK_TYPE_ORDER` only as a membership filter. But there is a second consumer:

```ts
// src/components/block-editor/forms/BranchBlocksEditor.tsx:298
const ALLOWED_BRANCH_BLOCK_TYPES: BlockType[] = BLOCK_TYPE_ORDER.filter(
  (t) => t !== 'section' && t !== 'collapsible' && t !== 'conditional'
);
```

Same 14 members before and after, but presented in a different order in the conditional-branch block picker. Cosmetic, and arguably an improvement since it now follows the same grouping logic as the palette — but the PR body says **"Breaking changes: None"** and step 3 of *How to verify* only asks a reviewer to check palette ordering.

Suggested: either add a verify step for `BranchBlocksEditor`'s picker, or state the reordering explicitly. It should not be discovered by a user.

## Also relevant

The annotation on #1523 notes that `getAvailableConversions` gates *sources* only by `CONTAINER_TYPES`, so `collapsible`, `assistant` and `snippet-ref` are source-eligible at the utility boundary even though none is a valid target — `convertBlockType` can produce e.g. an `image` while silently discarding `blocks` or `snippetId`. That is latent today only because those forms do not render `TypeSwitchDropdown`. It is out of scope here (this branch does not touch conversion), but it is worth knowing that **`assistant` joining `BlockType` in this PR does not widen that exposure** — `assistant` has no standalone form and `FORM_COMPONENTS.assistant` is explicitly `null`.

tomglenn (approved):
Ratchet verified real (injecting a bogus type triggers compile-time + runtime failures); challenge block fully wired; Node-safety intact.

PR #1533, opened 2026-08-07
fix(coda): classify Coda errors by upstream status, not message prose

### LINKED ISSUE #1521: Upstream Coda 401 surfaces as 500 on /sample-apps and /alloy-scenarios (error classification keyed on message substrings)

moxious:
## Summary

An expired Coda token surfaced to the browser as HTTP `500` on `/sample-apps` and `/alloy-scenarios`, so the frontend could not tell "you need to re-register" apart from "the backend is broken". Handler classification substring-matched `"authentication failed"` against a message built in `coda.go`; `ListSampleApps` and `ListAlloyScenarios` were the only two Coda calls with no upstream-401 branch, so their `401` fell through to the generic `"unexpected status code 401: <body>"` and was read as a server error. `CodaClient` now returns typed errors carrying the upstream status, and handlers branch on `errors.As` instead of prose — the two routes get the correct `401` for free.

## What to look at

- **`pkg/plugin/coda.go`** — the new `codaUpstreamError{status, msg, detail}` and `codaAuthSetupError{err}`, and the `isCodaAuthError` / `isCodaNotFoundError` classifiers. This deliberately mirrors `appPlatformUpstreamError` in `app_platform_client.go` rather than introducing a second pattern. The messages are preserved verbatim on purpose: `codaAuthSetupError.Error()` reproduces the old `fmt.Errorf("authentication failed: %w", err)` byte-for-byte, and the `429` message keeps its "quota" / "maximum number" wording so the existing match at `stream.go:562` still fires.
- **`isVMNotFoundError` is now typed-only** — the `strings.Contains(err.Error(), "VM not found")` branch is **deleted**, not kept alongside the typed check. It was not redundant, and it was not belt-and-braces: on an upstream `404` the typed check already fires, so the string branch's only reachable contribution was a `5xx` whose body happens to mention a missing VM being read as a destroyed VM — which makes `stream.go:414` clear the cache and reprovision, burning one of the user's three quota slots and adding a ~3-minute wait. Both callers (`stream.go:414`, `:588`) see only `GetVM` errors, whose sole `"VM not found: %s"` producer is the typed 404 branch, so deleting it is safe. `TestIsVMNotFoundError` in `coda_test.go` had to be rewritten to construct typed errors — it previously asserted substring matching as correct behavior in a case literally named `"partial match"`, which is the proposition #603 exists to retire.
- **`pkg/plugin/resources.go`** — all six `strings.Contains(err.Error(), "authentication failed")` sites plus the `"not found"` check are now typed, and the five identical classify-and-write blocks collapse into one `writeCodaError` helper. Routes affected: `POST /vms`, `GET /vms`, `GET /vms/{id}`, `DELETE /vms/{id}`, `GET /sample-apps`, `GET /alloy-scenarios`. `writeCodaError` also logs the upstream detail that the response withholds (see below). The `"invalid enrollment key"` check in `handleCodaRegister` is deliberately left on `strings.Contains` — **for scope discipline: it is not one of the six sites in #1521.** Nothing technical blocks converting it (`Register()` could return a `codaUpstreamError` like any other caller); it is simply a different ratchet and belongs in its own change. Tracked in #1535.
- **`docs/developer/CODA.md`** — per-status table for the Coda-backed routes, the two prose-matching deltas below, and a note not to reintroduce `strings.Contains` over `err.Error()`.

## Why

The objective was a bug fix plus type safety with **minimal functional change** — three status deltas total, all enumerated below, two of them on paths only reachable by accident. Every error path through the six classification sites was enumerated and its status recorded before editing, then re-verified after.

Verification method: a logging-only probe test was compiled against a pristine `origin/main` tree and against this branch, and the full **6 routes × 12 upstream outcomes = 72 cells** of `(HTTP status, response body)` were diffed mechanically. Eleven cells differ, in the four classes below. Every other cell is byte-identical, including the `503` unregistered path on all six routes, the local token-refresh `401` on all six routes, all success codes, and the `POST /vms` `429`/`409` special cases.

| Path                                                                   | Before | After     |
| ---------------------------------------------------------------------- | ------ | --------- |
| Upstream `401` on `/sample-apps`, `/alloy-scenarios`                   | `500`  | **`401`** |
| Upstream `401` on `/vms`, `/vms/{id}` (all methods)                    | `401`  | `401`     |
| Local token-refresh failure (all six routes)                           | `401`  | `401`     |
| `GET /vms/{id}` upstream `404`                                         | `404`  | `404`     |
| `POST /vms` upstream `429` / `409`                                     | `500`  | `500`     |
| `POST /vms` local quota guard                                          | `429`  | `429`     |
| Unregistered instance (all six routes)                                 | `503`  | `503`     |
| Request construction, transport, decode failures                       | `500`  | `500`     |
| Any other upstream status                                              | `500`  | `500`     |
| Upstream non-401 whose **body** contains `"authentication failed"`     | `401`  | **`500`** |
| `GET /vms/{id}` upstream non-404 whose **body** contains `"not found"` | `404`  | **`500`** |

The three deltas, all improvements:

1. **The fix** — upstream `401` on the two list routes now returns `401` instead of `500`.
2. An upstream non-401 whose response _body_ happened to contain the literal text `"authentication failed"` used to produce `401` and now correctly produces `500`. Applies to all six routes.
3. On `GET /vms/{id}`, an upstream non-404 whose response _body_ happened to contain `"not found"` used to produce `404` and now correctly produces `500`. The same reasoning applies to `isVMNotFoundError`, whose prose branch is deleted for the same reason.

**On the 401 response body**, scoped precisely: a `401` returns a fixed message and never echoes the upstream body. This is a change on **two of the six routes, not all six** — `POST /vms`, `GET /vms`, `GET /vms/{id}` and `DELETE /vms/{id}` already returned that exact fixed message on `main`. Only `/sample-apps` and `/alloy-scenarios` previously echoed the upstream body, and they did so with a `500`.

Suppressing the body must not cost debuggability, so `newCodaUpstreamError` reads the body — bounded by `maxCodaErrorBodyBytes`, mirroring the `maxBytes` enforcement in `app_platform_client.go` — into a `detail` field that `Error()` omits and `writeCodaError` logs. Without this, discarding the 401 body would have removed it from the plugin log as well as the response, and the log line is what an operator uses to tell `"instance not enrolled"` from `"jti revoked"` from clock skew. Bounding the read also fixes a pre-existing unbounded `io.ReadAll` and drains the connection for reuse, which the old 401 early-return skipped.

## How to verify

1. Reproduce the original bug on `main`: point a registered instance at a Coda API that returns `401` (or let a refresh token expire), open the block editor and add a terminal-connect block. The sample-app / alloy-scenario dropdown fails, and the network tab shows `500` on `/api/plugins/grafana-pathfinder-app/resources/sample-apps`.
2. On this branch, repeat: the same request returns `401` with body `{"error":"authentication failed: token may be invalid or expired, please re-register"}`, the response no longer echoes the upstream body, and the plugin log carries the upstream body under `Coda upstream detail withheld from response`.
3. Confirm the non-regression cases by hand: `GET /vms/{unknown-id}` still returns `404`, `POST /vms` past the quota still returns `429` from the local guard, and an unregistered instance still returns `503` on every Coda-backed route. All three are now also pinned by tests.
4. Exercise terminal reconnect end to end — connect a terminal, let the VM be destroyed out from under the session, and confirm reprovisioning still works. This is the `isVMNotFoundError` path in `stream.go` that the typed 404 has to keep satisfying; `TestIsVMNotFoundErrorAgainstRealGetVM` pins it against a real `GetVM` response rather than a hand-built error.

## Tests

`pkg/plugin/coda_error_classification_test.go` is new. Before this PR, `pkg/plugin/` had **zero** handler-level tests for any Coda route — `newTestApp` returns an `App` with `coda == nil`, so only the `503` path was reachable. #1521 was structurally uncatchable by the suite. The new coverage:

- `TestCodaHandlerStatusClassification` — table-driven over all six routes × {upstream `401`, `500`, `503`, local auth-setup failure, success} plus the `POST /vms` `429`/`409` and `GET /vms/{id}` `404` extras, asserting real HTTP status through the real mux.
- `TestGetVMNotFoundBranchDoesNotShadowOtherStatuses` — `handleGetVM` is the one route whose error path is not solely `writeCodaError`, so it is the one place a future edit could regress silently. Pins that its 404 branch neither swallows an upstream `401` nor fires on not-found body prose.
- `TestNewCodaUpstreamErrorCapturesBodyAsDetailOn401` / `...DoesNotLeakBodyOn401` / `...BoundsBodyRead` — the detail is recoverable for logging, absent from the message, and truncated at the cap.
- `TestCreateVMLocalQuotaGuardReturns429` — the local guard was the only `429` producer on any route and had no test, which is why the local-vs-upstream `429` inconsistency went unnoticed.
- `TestCodaHandlersUnavailableWithoutClient` — pins the `503` unregistered path on all six routes.

## Breaking changes

None for callers that treat non-2xx as failure. `/sample-apps` and `/alloy-scenarios` now return `401` where they previously returned `500` for an expired token — that is the fix. No frontend consumer branches on the distinction: `grep '401'` across `src/` returns zero production hits, there is no shared HTTP helper, interceptor, retry wrapper or error normalizer, `useCodaOptions` has a bare `error()` handler that never reads status, and `package-recommendations-client.ts` collapses every failure to one session-sticky boolean.

One consequence found in review and worth knowing: **Grafana core runs a global 401 handler on every `getBackendSrv()` request**, gated only on `isLocalUrl` (`backend_srv.ts:499`), which a plugin resource URL satisfies. Because `useCodaOptions` passes no `retry`, a `401` now costs one spurious `GET /api/login/ping` plus one request retry per failed catalog load. Not user-visible and not a blocker — filed as #1539 with the full analysis. It does **not** trigger a sign-out in practice: that needs `/api/login/ping` to itself 401, and a dead Grafana session already 401s these routes at Grafana's own auth middleware before plugin code runs. The user-visible toast improves, from a red error reading `unexpected status code 401: {"error":"token expired"}` to a yellow warning reading `authentication failed: token may be invalid or expired, please re-register`.

## Reversibility

Fully reversible — this is a backend HTTP status contract change with no storage, telemetry, or manifest component, and no frontend consumer reads the distinction yet. A revert restores the old `500` and nothing else has to be undone.

## Out of scope

- The frontend half of #1521 — no consumer distinguishes retryable from terminal failures by type. The issue explicitly scopes that out.
- The SSH and session text matchers in `stream.go` (`isSSHAuthError`, `isSSHRetryableError`) and `coda_exec.go` (`isDeadSessionError`), and `categorizeConnectionError` in `terminal.go`. They match third-party SSH and WebSocket library text, which is a different problem with no typed status to key on. All were verified to still receive the text they expect: none of them ever sees a Coda REST error.
- Upstream `429` from `CreateVM` still maps to `500` while the local quota guard returns `429` for the same condition. Preserved deliberately because `stream.go:562` matches on the message prose, and verified to still fire. Tracked in #1535, which also needs to unify the two messages' differing wording, not just the status.
- `setAuthHeader` reporting upstream availability failures (a `5xx` from Coda's own `/auth/refresh`) as `401`. This predates the PR — the old `strings.Contains` matched the same wrapped message and also returned `401` — and is preserved on purpose so this PR moves exactly the documented cells. Tracked in #1535; the `codaAuthSetupError` doc comment points there rather than certifying the classification as correct.
- Unvalidated upstream JSON decodes in `coda.go` — tracked separately. (The missing response-body size limit on the error path is fixed here, since the read was centralized into one function.)
- Duplicated `pkg/plugin/coda.go` file-inventory rows in `.cursor/rules/coda.mdc` and `.cursor/rules/systemPatterns.mdc`. Neither is falsified by this change — neither makes any claim about error classification — and `AGENTS.md` already directs agents touching `pkg/` to load `docs/developer/CODA.md` alongside `coda.mdc`, so the new contract is on the prescribed path.

Fixes #1521
Refs #603, #1535, #1539


moxious:
## Consistency check against the annotated issue

An independent annotation was added to #1521 after this PR was opened. Checked this branch against all four of its constraints — **three are satisfied, two decisions are made implicitly and should be confirmed by a reviewer.**

**Satisfied — the one that would have been a regression.** The annotation warned that replacing all six string checks with an upstream-only `errors.As` would regress local refresh failures from 401 to 500, because the local `setAuthHeader` path is not an upstream error. This branch handles it: `codaAuthSetupError` is a distinct type (`coda.go:96`) and `isCodaAuthError` (`:106`) accepts both sources. The before/after table in the PR body confirms local token-refresh failure stays 401 on all six routes, verified empirically against unmodified `main`.

**Satisfied.** No Go test enshrined the erroneous 500 on either catalogue endpoint; this branch adds coverage for upstream 401, local refresh failure, and non-auth upstream failure.

**Satisfied.** The change is backend-contract-only. `useCodaOptions` still discards every error identically and exposes only `isLoading`, so no re-registration UX appears from this alone — as the annotation notes, that is fine as long as the PR is not described as fixing user-visible behavior. It is not.

**Implicit decision 1 — no retry semantics.** An upstream 401 means the cached access token was rejected despite `getAccessToken`'s one-minute refresh buffer considering it valid (revocation, clock skew, changed upstream validation). The annotation asks whether to surface 401 and instruct re-registration, or invalidate the cached token and attempt one refresh-and-retry first. This branch takes the first option, matching what Create/Get/Delete/List already do. That is the right call for a scope-limited fix — retry would be a broader behavior change — but it is worth a reviewer confirming rather than inheriting.

**Implicit decision 2 — the 404 path keeps its prose coupling.** This branch migrated the `"not found"` check at `resources.go:240` to the typed classifier, but `isVMNotFoundError` retains a `strings.Contains` fallback specifically because `TestIsVMNotFoundError/partial match` feeds it plain `errors.New` prose. The annotation's position is that this test enshrines the same string-coupling being retired and should be replaced when the 404 path migrates. The compatibility choice here is defensible and its two callers were verified to operate only on `GetVM` errors — but it does leave one prose match alive in code whose stated purpose is removing them. Either replace that test in this PR, or note it as deliberate residue.

Neither implicit decision blocks merge in my view; both are worth an explicit line from a reviewer.

## Follow-ups already filed

- #1535 — `setAuthHeader` classifies upstream availability failures (e.g. a 5xx from `/auth/refresh`) as 401. Pre-existing, not introduced here; this PR's typed error is what makes it cheap to fix. Also covers the upstream-429-from-`CreateVM` case that was preserved for the same prose-matching reason.

tomglenn:
Closing: the Coda backend is being moved/replaced, so this error-classification fix would land on code that's going away. The fix itself is sound — reopen if that timeline slips.

Jayclifford345:
This is a Coda plugin issue. The mechanism this PR built landed in `grafana/grafana-coda-app` in a
stronger form: typed sentinels plus a closed-set wire `code`, so no upstream body ever reaches a
consumer on any status rather than being suppressed on `401` only, and the "do not reintroduce
`strings.Contains` over `err.Error()`" note is now a repo non-negotiable that a grep confirms holds.
The two deltas this PR called out as accidental-path improvements hold by construction there, and
`isVMNotFoundError`'s prose branch is gone with no fallback.

One property of this PR did not survive the extraction and is being carried over: the `detail` field
that kept the upstream `401` body out of the response but in the plugin log. The new backend discards
the body on every classified status, so an operator has no way to tell "instance not enrolled" from
"jti revoked" from clock skew — which is the reason this PR added it. The 72-cell before/after
verification method is recorded there as the standard for future status changes.

https://github.com/grafana/grafana-coda-app/issues/71

Tracked in `grafana/grafana-coda-app`, owned by @Jayclifford345 and the Coda plugin maintainers.

Triaged as part of the Coda extraction (grafana/grafana-pathfinder-app#1468).

Issue #1519, opened 2026-08-07
Guide identity for var-* checks rides a race-prone window singleton


moxious:
## Bug description

`var-<name>:<value>` requirements — a shipped authoring feature — resolve against the guide id `'default'` for **every guide**, so guide response variables are effectively global instead of scoped per guide. Two guides that both store an answer under the same variable name will read each other's values.

## Root cause

`src/requirements-manager/checks/vars.ts:13-25` reads a window global that **nothing ever writes**:

```ts
/**
 * Get the current guide ID from the global context set by ContentRenderer.
 * Falls back to 'default' if not available.
 */
function getCurrentGuideId(): string {
  try {
    const guideId = (window as any).__DocsPluginGuideId;
    if (guideId && typeof guideId === 'string') {
      return guideId;
    }
    return 'default';
  } catch {
    return 'default';
  }
}
```

`grep -rn "__DocsPluginGuideId" src/` returns **exactly one hit** — the read above. The docstring describes a contract with `ContentRenderer` that does not exist. Either it was never wired up, or the writer was removed and the reader was left behind.

Because the fallback is silent and returns a usable string, there is no error, no warning, and no failing test. The feature appears to work.

## Related: `__DocsPluginContentKey`

The sibling global has the same shape — `src/global-state/content-key.ts:6-12` claims `content-renderer.tsx` sets it, but the only writers in the tree are tests. `content-key.ts` survives because it has a three-level fallback chain, but the documented contract is equally stale.

## Why this class of bug is invisible

`(window as any).__Foo` is a contract TypeScript cannot see. There is no declaration, so there is no compile error when a writer disappears, and no way for an agent editing `ContentRenderer` to know that a Tier-2 requirements check depends on a global it sets.

## Suggested fix

1. **Fix the bug** — pass the guide id through the requirements-check context rather than a window global. `guideVariableCheck` is reached via `routeUnifiedCheck`, which already has call-site context available.
2. **Prevent the class** — declare the remaining `__DocsPlugin*` globals in one `declare global` block with typed accessor pairs, so a reader without a writer is greppable from a single place. There are six such globals today.

## How this was found

Part of a five-agent type-design investigation commissioned against epic #603. See the analysis comment on that epic for full context and the exemplar PR.

moxious:
## Triage: this is NOT a zero-functional-change fix

[...]

PR #1537, opened 2026-08-07
fix(source): replace raw NUL bytes that hid two files from grep

### LINKED ISSUE #1536: Literal NUL byte makes content-renderer.tsx invisible to grep and rg
## Summary

`src/components/content-renderer/content-renderer.tsx` contains a **literal NUL byte** at line 596. `file` therefore classifies the file as `data` rather than text, and `grep -r` / `rg` skip it silently unless forced with `-a` / `--text`.

The file is 1000+ lines of the content-rendering hot path. It is currently invisible to the default search every reader reaches for first.

## Location

Line 596 builds a cache key by joining two values with a delimiter that cannot collide with content:

```
const overlayKey = `${html}<NUL>${baseUrl}`;
```

where `<NUL>` is a **raw 0x00 byte in the source**, not a two-character escape. The intent is reasonable; the encoding is the one form that poisons the file.

## Why this is worth fixing

It already produced a false bug report. #1519 was filed claiming `__DocsPluginGuideId` had a reader and no writer, because:

```
$ grep -rn  "__DocsPluginGuideId" src/     # 1 hit  - the read only
$ grep -ran "__DocsPluginGuideId" src/     # 2 hits - read AND write (this file, line 395)
```

The same false negative hides the `__DocsPluginContentKey` writer at line 343. Any search touching this file — an agent tracing a symbol, a developer grepping for a prop, a reviewer checking whether something is still referenced — gets a confidently wrong answer with no indication anything was skipped.

Note the failure mode: search does not error or warn, it returns a _shorter_ result set. That is the most dangerous shape a false negative can take, because a short result reads as "I have found everything."

## Fix

Replace the raw byte with an escape sequence. Semantically identical at runtime, and the file becomes text again:

```
const overlayKey = `${html}�${baseUrl}`;
```

Verify with:

```
$ perl -ne 'print "line $.\n" if /\x00/' src/components/content-renderer/content-renderer.tsx
$ file src/components/content-renderer/content-renderer.tsx   # should report "text", not "data"
```

## Preventing recurrence

This is squarely the kind of mechanical constraint epic #603 exists for — a repo property that silently degrades every future reader's ability to reason about the code.

A cheap guard: a structural test (or lint rule) asserting no file under `src/` contains a raw NUL. That is a few lines, runs in CI, and its failure message can name the file and point at the `�` escape. Worth adding in the same PR so the class cannot return.

Consider widening it slightly to any control character outside tab / newline / carriage return, since the same binary-classification behavior applies to those.

## Related

- #1519 — the false bug report this produced, now retracted and rescoped
- Epic #603 — mechanical constraint enforcement

moxious:
## Summary

Two source files each carried a literal `0x00` byte as a string delimiter, which made `file` classify them as `data` and made `grep -r` / `rg` skip them silently. Both delimiters are now written as the `\x00` escape — the identical character at runtime, so the cache key and memo key they build are byte-identical. A CI guard now fails if any tracked file grows a raw control byte again, and it runs at commit time as well as in CI.

## What to look at

- [`src/components/content-renderer/content-renderer.tsx`](https://github.com/grafana/grafana-pathfinder-app/blob/aa5051362d1688cd2c174a996417419f2419b943/src/components/content-renderer/content-renderer.tsx) and [`src/components/block-editor/LintBadge.tsx`](https://github.com/grafana/grafana-pathfinder-app/blob/aa5051362d1688cd2c174a996417419f2419b943/src/components/block-editor/LintBadge.tsx) — one character each. `overlayKey` still joins `html` and `baseUrl` with U+0000, and `pathKey` still joins path segments with U+0000. `LintBadge.tsx` renders as a binary diff in the GitHub UI because the _pre-image_ was binary; `git diff --text` shows the single changed line.
- [`src/validation/control-bytes.test.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/aa5051362d1688cd2c174a996417419f2419b943/src/validation/control-bytes.test.ts) — the guard. It enumerates tracked paths via `git ls-files` but reads **working-tree content**, so an unstaged edit to a tracked file is caught immediately; a brand-new file is caught once it is added. It skips a denylist of binary asset extensions rather than allowlisting text extensions, so a new file type is scanned by default instead of being silently exempt. The failure message is the teaching surface, and it has two arms:
  - **Source arm** — names file, line, column (in characters, matching what an editor shows), an escape for **every** distinct offending byte, plus the doc: issue #1536 and the AGENTS.md section.
  - **Binary-asset arm** — when a file's control-byte density and absolute count both indicate a blob, the message says so and names `BINARY_EXTENSIONS` and the file holding it. Without this, committing a `.bmp` produced twenty lines of byte offsets and advice to write an ESC escape into TypeScript. Density is derived from content rather than a second "source extension" list, which would reintroduce the hand-maintained projection the guard exists to remove.
- [`package.json`](https://github.com/grafana/grafana-pathfinder-app/blob/aa5051362d1688cd2c174a996417419f2419b943/package.json) — **developer-workflow change.** The guard joins `test:governance`, which `.husky/pre-commit` runs, so a raw byte fails at commit time rather than in CI. `--passWithNoTests` is dropped from that script: all four listed paths resolve to real suites, and the flag meant a fully-broken path list would pass green — the same vacuity class this guard exists to prevent. See the caveat under **Known gap** below.
- [`src/cli/e2e/guide-chains.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/aa5051362d1688cd2c174a996417419f2419b943/src/cli/e2e/guide-chains.ts) — a third NUL-delimiter site that predates this PR and was already escaped, but with the other spelling. Aligned to `\x00` so the repo teaches one form. Both denote U+0000, so `cycleKey` is byte-identical.
- [`AGENTS.md`](https://github.com/grafana/grafana-pathfinder-app/blob/aa5051362d1688cd2c174a996417419f2419b943/AGENTS.md) — the rule, one line of why, the spelling convention, and a pointer. The full rationale lives in the guard's failure message, which is where an agent actually encounters it.

## Why

The interesting part is not the byte, it is the shape of the failure it causes. A poisoned file does not make search error out or print a warning — it makes search return a **shorter** result set. Every other search failure announces itself; this one is indistinguishable from success, because a short result reads as "I have found everything."

That is exactly how it played out. #1519 was filed claiming `__DocsPluginGuideId` had a reader and no writer, and therefore that every guide's `var-*` responses collapsed onto a shared `'default'` key. The writer was there the whole time, in `content-renderer.tsx` — the file `grep` was skipping. The issue went as far as a triage comment recommending a migrate-on-read fallback before an independent analysis caught it; that recommendation would have leaked one guide's unscoped answers into another. #1519 has been retracted and rescoped.

The guard is scoped wider than the bug: any control byte outside tab, newline, and carriage return, across every tracked file. Those characters produce the same binary classification, the repo is already clean of all of them, and narrowing the rule to NUL alone would leave the next variant of the same trap uncaught.

Worth knowing that the two files were never in the same state. Git's binary heuristic only inspects the first 8000 bytes, and the NUL in `LintBadge.tsx` sits at offset 2176 while `content-renderer.tsx`'s is at 20808. So git treated `LintBadge.tsx` as binary — its patch history has been unreviewable and it would have merge-conflicted wholesale rather than by hunk — while `content-renderer.tsx` was always text to git. `.gitattributes` does not help here: it sets `*.tsx text eol=lf`, but that governs eol normalization only; diff and grep binaryness come from the separate, unset `diff` attribute. For non-git tools the offset is irrelevant and both files were skipped.

## How to verify

1. Reproduce the false negative on `main`: `git checkout main`, then compare `grep -rn "__DocsPluginGuideId" src/` (1 hit) with `grep -ran "__DocsPluginGuideId" src/` (2 hits). The second hit is the writer that #1519 claimed did not exist.
2. On this branch, `file src/components/content-renderer/content-renderer.tsx src/components/block-editor/LintBadge.tsx` reports text rather than `data`, and both `grep` invocations now return 2 hits.
3. Confirm the delimiters are unchanged at runtime — both escape spellings are U+0000, so every key is byte-identical.
4. Confirm the guard bites and is not vacuous. Three probes, each leaving the tree clean:
   - **Source arm:** write a raw NUL and a raw ESC into a scratch file under `src/`, `git add --intent-to-add` it, run `npx jest src/validation/control-bytes.test.ts --coverage=false`. Expect both hits named with correct line and column, and both escapes suggested.
   - **Binary arm:** `cp src/img/img_1.png src/img/probe.bmp`, `git add --intent-to-add` it, re-run. Expect the message to name `BINARY_EXTENSIONS` and this file, not escape advice.
   - **Vacuity:** copy the guard with `REPO_ROOT` repointed at an empty git repo. Expect it to fail loudly on the file-count and reachability assertions rather than pass green.
5. Confirm no other offenders remain: `git ls-files -z | perl -0ne '@f=split /\0/; for $p (@f){ next unless -f $p; open(my $fh,"<:raw",$p) or next; local $/; my $c=<$fh>; close $fh; print "$p\n" if $c =~ /\x00/ }'` should list only PNGs.
6. `npm run test:governance` should show four suites, including the guard.

## Known gap

`.husky/pre-commit` is tracked with mode `100644`, and has been since it was added. Git refuses to execute a non-executable hook — it emits `hook was ignored because it's not set as executable` and continues — so the pre-commit wiring above does not fire until that mode bit is fixed (`git update-index --chmod=+x .husky/pre-commit`) or husky's `prepare` step is re-run to generate its `.husky/_` shims. That is a pre-existing repo condition affecting `lint-staged`, `typecheck`, and `test:governance` equally, and flipping it changes every contributor's commit latency, so it is deliberately left as a separate decision rather than folded in here. The guard is still enforced by `npm run test:ci` and `npm run check`.

## Breaking changes

None. All three delimiters are the same character at runtime, so cache keys, memo keys, and cycle keys are byte-identical and no stored or derived value changes. Nothing persists either key — both are compared in memory only, and neither file touches storage, logging, or telemetry.

## Out of scope

- The genuine defect underneath the false one — `__DocsPluginGuideId` is a process-wide window singleton written from a passive effect, so a child requirement check can read the previous guide's id and two mounted renderers can overwrite each other. That is tracked on the rescoped #1519 and needs a plumbing decision this PR deliberately does not make.
- Trojan Source / bidi-override characters (CVE-2021-42574). They are all above `0x20` once UTF-8 encoded, so this guard's byte-level predicate cannot match them. Different predicate, different threat model, and it needs its own allowlist design for a future RTL locale — filed as #1547.
- Guard coverage stops at the tracked tree. Generated output under `dist/` and vendored `node_modules/` are not scanned, and should not be.
- No ESLint rule. In-editor feedback was considered and rejected: the test reaches `.md`, `.go`, `.json`, and `.yml`, which ESLint cannot, and the pre-commit wiring captures most of the latency benefit.

Fixes #1536


tomglenn (approved):
Verified `\x00` is byte-identical to the old raw NUL at all three call sites (no cache-key drift), and the CI guard fires on an injected NUL.

PR #1538, opened 2026-08-07
docs: document the challenge and snippet-ref guide block types

### LINKED ISSUE #1530: json-guide-format.md omits the challenge and snippet-ref block types

moxious:
## Summary

The JSON guide format reference documented 17 of the 19 `JsonBlock` variants — `challenge` and `snippet-ref` had neither a summary-table row nor a per-type section. Both are now documented, and a coverage test reads the block-type list out of the schema union at runtime so the next new block type cannot land undocumented.

**Round-2 note.** The first pass drafted the prose from the schema's `.describe()` strings and claimed it was verified against runtime source. A field-by-field review found four statements that were wrong rather than merely imprecise, and they are corrected in `61b19688`. Treat the earlier "verified against runtime source" claim as overstated — the schema descriptions turned out not to be a trustworthy source, which is now the main argument against generating this reference (see **Why**).

## What to look at

- [`docs/developer/interactive-examples/json-guide-format.md`](https://github.com/grafana/grafana-pathfinder-app/blob/61b196881374d2b5fa37a210e67aed716b2c9f6a/docs/developer/interactive-examples/json-guide-format.md) — two new per-type sections plus two summary rows (`challenge` is Interactive, `snippet-ref` is Reusable, matching the block-palette grouping once #1531 lands). The prose deliberately describes several quirks rather than smoothing them over: `mode` resolves to `coda` when omitted from JSON but to `standard` for a block created in the editor; `hintLevels` is `{ text: string }[]`, not a string array; `setupScript` beats the deprecated `setupCommands` at runtime and the editor migrates the latter on save.

  The four corrections in `61b19688`, each traced to source:

  - **`snippet-ref` "the parser never sees this block" was inverted.** The guide schema accepts `snippet-ref` (`json-guide.schema.ts:761`), validation runs before expansion (`content-fetcher.ts:230`), and `json-parser.ts:315-321` carries a dedicated `case 'snippet-ref':` arm. It is the _renderer_ that never sees it. The table cell and summary row said "at parse time" and are now consistent with the prose.
  - **`setupScript` and `setupCommands` do not "do the same job".** `setupScript` is one remote invocation with a 120s budget; each `setupCommands` entry is a separate invocation with its own 30s budget and no shared shell state, so `cd`, `export` and heredocs do not carry between entries (`challenge-block.tsx:265-297`). An author who trusted "same job" and wrote `["cd /etc/prometheus", "sed -i … prometheus.yml"]` would get the `sed` in the wrong directory with setup reporting success.
  - **The coda-mode gate was understated.** It is `isDevMode && pluginConfig.enableCodaTerminal` (`docs-panel.tsx:1430`) — per-user dev mode as well as the plugin setting, so not a single admin toggle. And when the gate is closed the block neither errors nor hides: it sits on "Provisioning challenge VM…" indefinitely (#1541). Since an omitted `mode` resolves to `coda`, that is the usual way in, so the `mode` note is now a warning rather than a neutral statement.
  - **`requirements`, `objectives` and `skippable` are schema-accepted but runtime-inert.** `convertChallengeBlock` maps 11 props and none of these three (`json-parser.ts:987-999`), while `convertCodeBlockBlock` immediately below it does map them. So a challenge with `"requirements": ["is-admin"]` renders for everyone, `objectives` never credits, and `skippable` shows no skip control — with no validation error anywhere. The rows stay (the schema really does accept them) but are now labelled as not yet honoured. Converter fix tracked in #1544; the editor-side half is #1543.

  Four smaller trap-closers in the same commit: `failureMessage` _replaces_ the requirement checker's own error text rather than adding to it; hints are unreachable outside the `ready` and `failed-check` states, so a learner stuck on provisioning cannot see them; an unresolved ref reports `network-error` even for an unknown ID (#1546); and a ref may not sit inside a `collapsible`.

- [`src/validation/json-guide-format-docs.test.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/61b196881374d2b5fa37a210e67aed716b2c9f6a/src/validation/json-guide-format-docs.test.ts) — the guard walks the `JsonBlockSchema` union at runtime rather than reading a hand-maintained list, because the hand-maintained lists are exactly what drifts (`VALID_BLOCK_TYPES` on `main` is itself missing `challenge`; #1531 fixes that separately). Worth tracing: the summary-row check is scoped to the "Block Types Summary" table only, and the heading check refuses a match that a longer block type already explains, so `terminal` is not satisfied by the `terminal-connect` heading.

  `61b19688` replaces `toBeGreaterThan(15)` with an exact count of 19. The loose bound tolerated silently losing up to three union arms, and on a Zod-internals break it surfaced as `` `.each` called with an empty Array `` rather than naming the cause. The `expect(new Set(blockTypes)).toEqual(VALID_BLOCK_TYPES)` cross-check was considered and left out: it fails today, because that set is still the hand-written 18-entry list missing `challenge`. It becomes worth adding once #1531 merges and derives the set from `KNOWN_FIELDS`.

- [`README.md`](https://github.com/grafana/grafana-pathfinder-app/blob/61b196881374d2b5fa37a210e67aed716b2c9f6a/README.md) — the docs bullet hand-copied a block-type list this change would have made stale; it now points at the reference instead of duplicating it. The adjacent demo link also pointed at a path that no longer exists after the package restructure, and claimed the demo covered "all block types" when it contains 9 of 19.

## Why

Fixes #1530. Both block types have been parseable for a while, but `challenge` became urgent because #1531 makes it insertable from the block-editor palette — authors can now reach a block the reference never describes.

Generating the reference from the CLI schema walker was considered and rejected. The original reasons still hold: `challenge` and `snippet-ref` are both outside the CLI authoring registry, and `schema-options.ts` skips arrays of objects such as `hintLevels`. The stronger reason emerged from the round-2 review — two `.describe()` strings are themselves wrong (`snippetId` says "resolve at parse time"; `mode` says "'coda' (default)" on a field with no `.default()`), so generation would have propagated both defects into the reference instead of catching them. #1545 tracks fixing them at the source, since the CLI and MCP surface that text. The guard therefore checks coverage only and leaves the prose hand-written.

## How to verify

1. Read the two new sections against `src/types/json-guide.schema.ts` for field names, requiredness and types. Do **not** treat the `.describe()` strings as the oracle for behaviour — check `challenge-block.tsx` and `ChallengeBlockForm.tsx` for that, and see #1545.
2. Confirm the guard bites, both modes. Delete the `` | `challenge` `` row from the "Block Types Summary" table and run `npx jest src/validation/json-guide-format-docs.test.ts --coverage=false`; then restore it and rename the `#### Challenge block` heading instead. Each should fail naming the block type and the file to edit:
   - `Block type "challenge" has no summary-table row. Add one to the "Block Types Summary" table in docs/developer/interactive-examples/json-guide-format.md.`
   - `Block type "challenge" has no "#### " section. Add one to docs/developer/interactive-examples/json-guide-format.md describing its fields.`
3. Confirm the exact-count assertion is live: set `EXPECTED_BLOCK_TYPE_COUNT` to 20 and check it fails with `Expected length: 20 / Received length: 19`.
4. Confirm the guard is not satisfied by a row outside the table: delete the real `` | `challenge` `` row and plant a matching one next to the `#### Challenge block` heading. The row check should still fail.
5. Every fenced `json` example in the two new sections parses under `JsonBlockSchema.safeParse` — worth re-checking mechanically rather than by eye, since authors copy them.
6. Copy the standard-mode challenge example into a guide and load it — the brief and "Check my work" should appear immediately with no provisioning step.

## Breaking changes

None. Documentation plus one test; no runtime behaviour changed, and no source file is touched. The quirks described above are intentional and covered by existing tests, and were documented as-is rather than altered.

## Reversibility

Fully reversible — reverting restores the previous prose and drops the guard. The four behavioural defects the review surfaced are filed separately (#1544, #1545, #1546) and are not fixed here, so a revert loses documentation only, not fixes.

## Out of scope

- **All source-code changes.** The four defects this review found in the challenge and snippet-ref runtimes are documented as current behaviour and filed as #1544, #1545 and #1546. #1541 and #1543 were filed independently by the #1531 fix pass.
- `docs/developer/components/block-editor/README.md` and `docs/sources/block-editor/_index.md` are owned by in-flight #1531 and deliberately untouched.
- The two new `####` headings are sentence case per `AGENTS.md`; the existing 17 are title case. Left mismatched rather than expanding this into a repo-wide heading sweep.
- `authorNote` is undocumented for all 19 block types in this file, and the summary table's row order does not match the prose order. Both predate this change.
- The "TypeScript Types" section of the reference hand-lists a subset of the exported types and guards (missing `JsonChallengeBlock`, `JsonSnippetRefBlock`, and others). That drift predates this change and the new guard does not cover it.

Fixes #1530

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,12 +75,12 @@ Editors and admins can author custom interactive guides in two ways:
 
 ### Documentation
 
-- **[JSON Guide Format Reference](docs/developer/interactive-examples/json-guide-format.md)** — Complete reference for the JSON guide structure and all block types (markdown, image, video, section, conditional, interactive, multistep, guided, code-block, terminal, terminal-connect, grot-guide, quiz, input, assistant)
+- **[JSON Guide Format Reference](docs/developer/interactive-examples/json-guide-format.md)** — Complete reference for the JSON guide structure and every block type, with a summary table and a per-type section
 - **[Interactive Types](docs/developer/interactive-examples/interactive-types.md)** — Action types (`highlight`, `button`, `formfill`, `navigate`, `hover`, `noop`, `popout`) and when to use each
 - **[Requirements Reference](docs/developer/interactive-examples/requirements-reference.md)** — Available requirements for controlling when interactive elements are accessible
 - **[Selectors Reference](docs/developer/interactive-examples/selectors-reference.md)** — How to target DOM elements with the enhanced selector engine and the resilience pipeline
 
-See the [JSON Guide Demo](src/bundled-interactives/json-guide-demo.json) for a complete example of all block types.
+See the [JSON Guide Demo](src/bundled-interactives/json-guide-demo/content.json) for a worked example of the most common block types.
 
 ## For developers
 
--- a/docs/developer/interactive-examples/json-guide-format.md
+++ b/docs/developer/interactive-examples/json-guide-format.md
@@ -1013,6 +1013,99 @@ A choose-your-own-adventure decision tree where each screen offers options that 
 
 Each screen has `id`, `title`, `body` (markdown), and an `options[]` array. Each option has a `label` and a `next` screen ID. A screen with no `options` ends the tree. The block editor includes a YAML import flow for converting Grot Guide YAML directly into JSON.
 
+#### Challenge block
+
+A capture-the-flag style task: a title, a markdown brief, optional progressive hints, and a "Check my work" button that evaluates a single requirement token. The `mode` field picks the execution model, and most of the other fields follow from it.
+
+| Mode         | Behavior                                                                                                                                                                                                           |
+| ------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `"standard"` | Runs against the learner's own Grafana. Nothing is provisioned, and the brief and "Check my work" are available immediately. `successCriteria` is any Pathfinder requirement, such as `has-datasource:prometheus`. |
+| `"coda"`     | Provisions a Coda VM with a terminal, runs the setup script, then reveals "Check my work". `successCriteria` is typically `coda-exit-zero:<command>`.                                                              |
+
+`mode` is optional, and it resolves to a different value depending on how the block reaches you:
+
+- **Omitted in JSON** — the runtime treats the block as `"coda"`. Challenges predate `"standard"`, so an existing guide with no `mode` keeps its original Coda behavior.
+- **Created in the block editor** — the challenge form writes `mode` explicitly and seeds a brand-new block with `"standard"`, the cheaper authoring path. Opening a legacy block that has no `mode` infers `"coda"`, either from the presence of a VM or setup field or as the safe fallback.
+
+Both defaults are intentional. Set `mode` explicitly in hand-written JSON so the block does not depend on which path it came through.
+
+Standard mode:
+
+```json
+{
+  "type": "challenge",
+  "mode": "standard",
+  "title": "Connect your first data source",
+  "brief": "Add a Prometheus data source so the rest of this guide has something to query.",
+  "successCriteria": "has-datasource:prometheus",
+  "failureMessage": "No Prometheus data source yet — check the connections page.",
+  "hintLevels": [
+    { "text": "Start from Connections > Data sources." },
+    { "text": "The Add new data source button is in the top right." }
+  ]
+}
+```
+
+Coda mode:
+
+```json
+{
+  "type": "challenge",
+  "mode": "coda",
+  "title": "Fix the broken scrape config",
+  "brief": "The Prometheus config in `/etc/prometheus` has a bad scrape target. Repair it.",
+  "vmTemplate": "vm-aws",
+  "setupScript": "set -euo pipefail\nsed -i 's/localhost:9999/localhost:9090/' /etc/prometheus/prometheus.yml.broken",
+  "successCriteria": "coda-exit-zero:promtool check config /etc/prometheus/prometheus.yml"
+}
+```
+
+| Field             | Type                     | Required | Default   | Description                                                                       |
+| ----------------- | ------------------------ | -------- | --------- | --------------------------------------------------------------------------------- |
+| `mode`            | `"coda"` \| `"standard"` | ❌       | see above | Execution model                                                                   |
+| `title`           | string                   | ✅       | —         | Short title shown above the brief                                                 |
+| `brief`           | string                   | ✅       | —         | Markdown problem statement                                                        |
+| `successCriteria` | string                   | ✅       | —         | Requirement evaluated when the user clicks "Check my work"                        |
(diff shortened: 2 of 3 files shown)


tomglenn (approved):
Docs match the schema; the coverage test is a real runtime guard that fails on an undocumented variant.

PR #1551, opened 2026-08-09
fix: validate pages array elements in HighlightedGuideConfig

### LINKED ISSUE #1534: Unchecked element cast on highlighted-guide 'pages' (sibling of #1520)
## Summary

Follow-up to #1520 (fixed in #1529), found while reviewing that fix. The **same function** has a second unchecked cast on the same untrusted payload — this time on array *elements* rather than a scalar.

## The gap

`src/utils/openfeature.ts`, `validateHighlightedGuideValue`:

```ts
// :556 — container is checked
if (typeof record.variant !== 'string' || !Array.isArray(record.pages) || typeof record.guideId !== 'string') {
  return null;
}
...
// :571 — elements are not
pages: record.pages as string[],
```

`Array.isArray` establishes only that `pages` is an array. A remote payload of `{ pages: [1, 2] }` or `{ pages: [{}, null] }` satisfies the guard and reaches consumers typed `string[]`.

Consumer: `src/utils/experiments/highlighted-guide-orchestrator.ts:127` → `matchesHighlightedGuidePage(config.pages, path)` → the URL pattern matcher. Non-string elements will either throw or silently fail to match depending on how the matcher handles them, so the observable symptom is either a thrown error during auto-open evaluation or an experiment that silently never fires — both hard to trace back to a malformed flag payload.

## Suggested fix

Element-wise validation in the same guard, consistent with how #1529 handled `variant`:

```ts
if (!Array.isArray(record.pages) || !record.pages.every((p) => typeof p === 'string')) {
  return null;
}
```

That also removes the need for the `as string[]` cast at `:571`.

## Two adjacent design questions worth settling here

Both surfaced while fixing #1520 and are cheap to resolve in the same PR:

1. **`ReadonlySet.has()` does not narrow**, so the `variant` check added in #1529 still needs a redundant `as` at `:574`. A small type-predicate helper would remove that cast at both the `variant` and `docType` checks. Worth doing once rather than accumulating casts each time a field is validated.

2. **The rejection policy is inconsistent within this one function.** An invalid `docType` degrades gracefully (falls back to URL inference); an invalid `variant` now discards the entire payload. Both are defensible in isolation, but the asymmetry appears to be accidental rather than chosen. Decide the rule — probably "fields that gate enrollment reject; presentational fields degrade" — and state it in a comment so the next field added follows it.

## Zero-functional-change note

The `pages` fix is behavior-preserving for any well-formed payload; it only changes what happens for a payload that is already malformed. Same reasoning as #1520.

## Related

- #1520 / #1529 — the sibling `variant` cast
- Epic #603 — mechanical constraint enforcement; this is the same "unchecked `as` immediately after a trust boundary" shape

Nikshaan:
## What this fixes

Fixes #1534 .

validateHighlightedGuideValue checks that record.pages is an array but not that its contents are strings, before trying to cast it to string[]. So a malformed payload like { pages: [1, 2] } would validate as okay.

Downstream, matchesHighlightedGuidePage calls pattern.trim() on each pattern, which requires it to be a string so an all-non-string pages array throws. so an all-non-string pages array would throw inside the auto-open orchestrator (uncaught), and a mixed array could silently mask the bug if one of the patterns was a string ( .some() short-circuits).

## The fix

Adds .every((p) => typeof p === 'string') to the existing shape guard, so that a malformed pages array would be caught the same way as other shape mismatches would be, by falling back to the default, rather than throwing. Same guard-and-early-return pattern is used elsewhere in this file.

## Scope of the fix

Only the check for the pages element. The variant already has an equivalent check for being a scalar in the same guard, and docType is checked separately and falls back to default, degrading gracefully rather than rejecting the shape (unrelated to this fix). Does not include the extra logging part of #1529 , which is separate and not merged in yet.

## Testing

Tests were added to openfeature.test.ts by example, in the same vein as the existing indirect tests for these functions through getHighlightedGuideConfig :

all pages are non-strings to fall back to default

pages are mixed to fall back to default (the masking case)

pages are all strings to no change

These were verified by removing the fix and checking that the two negative cases now failed for the expected reason, and then putting it back and checking that all 33 tests are still passing.

tsc --noEmit , eslint, and prettier --check all pass.

## Update

Also addressed the two adjacent design questions from the issue:

- **Type-predicate helpers** (`isValidDocType`, `isHighlightedGuideVariant`) remove the redundant `as` casts at the `docType`/`variant` call sites. One cast remains inside `isValidDocType` itself,  `ReadonlySet.has()` doesn't narrow, so the cast is required to call it at all; the predicate removes casts at the *branching* sites, not that one.
- `variant`'s fallback path (`highlightedGuideVariantFromString`) still uses `as` for unrecognized values, preserving `main`'s current passthrough behavior for unknown variants no rejection was added here, since that's `#1529`'s scope, still unmerged. Once #1529 lands and starts rejecting invalid `variant` in the shape guard, that fallback branch becomes unreachable and can be deleted.
- Documented the rejection-policy rule in a comment above the shape guard: enrollment-gating fields reject the whole payload; presentational fields degrade gracefully.

### DIFF
--- a/src/utils/openfeature.test.ts
+++ b/src/utils/openfeature.test.ts
@@ -682,26 +682,27 @@ describe('openfeature', () => {
       jest.isolateModules(() => {
         const mockOF = createMockOpenFeature();
         const mockReact = createMockReactSdk();
-        mockOF.mockClient.getObjectValue.mockReturnValue({
-          variant: 'excluded',
-          pages: [],
-          guideId: '',
-        });
+        const remoteConfig = {
+          variant: 'control',
+          pages: ['/connections/datasources*'],
+          guideId: 'bundled:remote-guide',
+        };
+        mockOF.mockClient.getObjectValue.mockReturnValue(remoteConfig);
         jest.doMock('@openfeature/web-sdk', () => mockOF);
         jest.doMock('@openfeature/react-sdk', () => mockReact);
 
-        const {
-          setFlagOverride,
-          getHighlightedGuideConfig,
-          DEFAULT_HIGHLIGHTED_GUIDE_CONFIG,
-        } = require('./openfeature');
+        const { setFlagOverride, getHighlightedGuideConfig } = require('./openfeature');
         setFlagOverride('pathfinder.highlighted-guide-experiment', {
           variant: 'treatment',
           pages: [1, 2],
           guideId: 'bundled:my-guide',
         });
 
-        expect(getHighlightedGuideConfig()).toEqual(DEFAULT_HIGHLIGHTED_GUIDE_CONFIG);
+        expect(getHighlightedGuideConfig()).toEqual({
+          ...remoteConfig,
+          autoOpen: true,
+          resetCache: false,
+        });
         expect(mockReportFeatureFlagExposure).not.toHaveBeenCalled();
         expect(mockOF.mockClient.getObjectValue).toHaveBeenCalled();
       });
@@ -711,26 +712,27 @@ describe('openfeature', () => {
       jest.isolateModules(() => {
         const mockOF = createMockOpenFeature();
         const mockReact = createMockReactSdk();
-        mockOF.mockClient.getObjectValue.mockReturnValue({
-          variant: 'excluded',
-          pages: [],
-          guideId: '',
-        });
+        const remoteConfig = {
+          variant: 'control',
+          pages: ['/connections/datasources*'],
+          guideId: 'bundled:remote-guide',
+        };
+        mockOF.mockClient.getObjectValue.mockReturnValue(remoteConfig);
         jest.doMock('@openfeature/web-sdk', () => mockOF);
         jest.doMock('@openfeature/react-sdk', () => mockReact);
 
-        const {
-          setFlagOverride,
-          getHighlightedGuideConfig,
-          DEFAULT_HIGHLIGHTED_GUIDE_CONFIG,
-        } = require('./openfeature');
+        const { setFlagOverride, getHighlightedGuideConfig } = require('./openfeature');
         setFlagOverride('pathfinder.highlighted-guide-experiment', {
           variant: 'treatment',
           pages: ['/explore', 1],
           guideId: 'bundled:my-guide',
         });
 
-        expect(getHighlightedGuideConfig()).toEqual(DEFAULT_HIGHLIGHTED_GUIDE_CONFIG);
+        expect(getHighlightedGuideConfig()).toEqual({
+          ...remoteConfig,
+          autoOpen: true,
+          resetCache: false,
+        });
         expect(mockReportFeatureFlagExposure).not.toHaveBeenCalled();
         expect(mockOF.mockClient.getObjectValue).toHaveBeenCalled();
       });
--- a/src/utils/openfeature.test.ts
+++ b/src/utils/openfeature.test.ts
@@ -677,6 +677,89 @@ describe('openfeature', () => {
         expect(mockOF.mockClient.getObjectValue).toHaveBeenCalled();
       });
     });
+
+    it('getHighlightedGuideConfig should fall through when pages contain only non-string elements', () => {
+      jest.isolateModules(() => {
+        const mockOF = createMockOpenFeature();
+        const mockReact = createMockReactSdk();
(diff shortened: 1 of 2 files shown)


moxious (changes requested):
Thanks for this, @Nikshaan — and thanks for signing the CLA. This is a well-scoped fix with real tests, and the core of it is exactly right.

The first commit (`ed65e5bf`) is good and we want it.

The changes below are all in the second commit. This is our fault, not yours. The two "design questions" in #1534 were written assuming #1529 had already merged — it's still open. So when the issue said "the `variant` check added in #1529 still needs a redundant `as`," it was describing a check that isn't on `main` yet. You did the reasonable thing with what the issue told you.

### 1. Please drop the second commit (`6468d704`)

The problem is this pair:

```ts
function highlightedGuideVariantFromString(value: string): HighlightedGuideConfig['variant'] {
  return isHighlightedGuideVariant(value) ? value : (value as HighlightedGuideConfig['variant']);
}
```

Both branches return `value`, so this is an identity function — byte-identical in behavior to the `as` cast it replaced. The narrowing from `isHighlightedGuideVariant` is computed and then discarded.

That matters because of the comment added just above the guard:

> Enrollment-gating fields (e.g. variant, pages, guideId) reject the whole payload when invalid.

For `pages` and `guideId` that's now true. For `variant` it isn't — the guard still only checks `typeof record.variant !==
'string'`. So a GOFF payload with a capitalizationeatment"` still validates. Every enrollment gatetests the complement (`!== 'excluded'`), so the user gets auto-open and Featured injection — but `TRACKED_EXPERIMENT_VARIANTS`
doesn't contain `"Treatment"`, so no exposure event` falls through to `'excluded'`, stamping everydownstream analytics event as excluded. Enrolled users, zero A/B denominator, contradictory variant label.

A comment that documents a guarantee the code doesn't provide is worse than no comment, because the next person adding a gating
field will trust it and ship a `typeof`-only checkclass #1520 exists to close.

The good news is you don't need to fix this. #1529 already does it properly it adds a `VALID_VARIANTS` set, validates `variant` against it, and hoists `VALID_DOC_TYPES` to module scope with a line identical to yours. It's in review now. So the two design questions from #1534 will get settled there, including the rejection-policy comment, at which point it'll be accurate.

Dropping the commit also avoids a guaranteed merge conflict — you and #1529 both delete the same inline `VALID_DOC_TYPES` block and both edit the same guard, so whichever lands second conflicts, and a naive resolution produces a duplicate identifier.

### 2. While you're there, drop the `as string[]` cast

#1534 called this out and it's the natural payoff ry element is checked, the cast has nothing left to hide:

```ts
if (
  typeof record.variant !== 'string' ||
  !Array.isArray(record.pages) ||
  !record.pages.every((p): p is string => typeof p
  typeof record.guideId !== 'string'
) {
  return null;
}
```

and then at the return:

```ts
pages: record.pages,
```

The explicit `(p): p is string` form is worth preferring over the plain predicate: the plain one only compiles because `Array.isArray` widens to `any[]`, which would keep silently accepting things if the element type ever changes.

### 3. Small test clarity fix (optional, but cheap)

In the two rejection tests, the mocked remote cliecluded', pages: [], guideId: '' }` — which isdeep-equal to `DEFAULT_HIGHLIGHTED_GUIDE_CONFIG`. So `expect(getHighlightedGuideConfig()).toEqual(DEFAULT_HIGHLIGHTED_GUIDE_CONFIG)` passes whether the override was rejected and the *remote value* won (the actual documented behavior) or the hard default was returned. Making the remote mock a distinct valid config and asserting *that* comes back would make the tests prove what their names claim.

Nikshaan:
Thanks for the detailed review, all makes sense. Removed the second commit entirely.

Left two minor things on top of the original fix applied - the explicit `(p): p is string` predicate, and the removal of `as string[]` cast, both checked to be safe with `tsc --noEmit` on the real project config, no errors at the return sites. Also fixed two rejection test cases to assert against a different mocked remote config than the one that was used as a default, so that they actually verify that the fallback returns the remote value, not the hardcoded default.

Left `variant`/`docType` for the moment, will rebase on #1529 when it lands.

moxious (approved):
Many thanks for the contribution!

PR #1464, opened 2026-07-30
feat(learning-paths): rework My Learning layout and restrict Discover More to paths

moxious:
## Intent

Address all important review findings on PR #1464 so the My Learning rework is merge-worthy. My Courses must contain only started paths with progress strictly between 0% and 100%, Completed must contain 100% paths, and untouched 0% paths must remain eligible for Discover more rather than being excluded from both surfaces. Preserve the existing path-only Discover more behavior and launch flow. Correct milestone terminology and localize the pending Opening state, add behavioral coverage for 0%/partial/100% partitioning plus singular/plural milestone labels, document the new stable My Learning test IDs, and trim the flagged restatement comments. The telemetry degradation event and splitting the shared testIds registry were explicitly judged non-blocking follow-up work and are out of scope for this merge-readiness fix.

## What Changed

- Rebuilt the My Learning tab as a two-column layout, decomposing the monolithic `MyLearningTab.tsx` into dedicated section components (`HeroStats`, `MyCoursesSection`, `BadgesSection`, `CompletedSection`, `DiscoverMoreSection`) plus `BadgeGridItem`/`BadgeDetailCard` badge tiles, and partitioned paths by progress so My Courses holds only in-progress paths (strictly between 0% and 100%), Completed holds 100% paths, and untouched 0% paths stay eligible for Discover More.
- Added the `useDiscoverMore` hook to restrict Discover More to learning paths (dropping guide-typed and URL-less entries, honoring `excludeTitles` and a count cap), corrected milestone terminology with singular/plural i18n keys, and localized the pending "Opening…" launch state.
- Added behavioral coverage for 0%/partial/100% partitioning and singular/plural milestone labels, introduced stable My Learning test IDs in `constants/testIds.ts`, and documented the new hook and test IDs in the learning-paths and E2E testing references.

## Risk Assessment

✅ Low: A well-bounded, thoroughly tested layout refactor whose course/completed/0% partition, path-only Discover more, launch flow, milestone localization, and documented test IDs all conform to the stated intent with no reachable defects found.

## Testing

[...]

Issue #1463, opened 2026-07-30
My Learning page: layout rework — two-column, restyled badges, Discover More, Completed


moxious:
## Scope

PR #1464 delivered this rework. Four of the five implementation slices shipped as written. The fifth shipped deliberately inverted, and this issue is rewritten to record that rather than to request work.

**Delivered**

- Slice 1 — the monolithic tab is split into `HeroStats`, `MyCoursesSection`, `BadgesSection`, `DiscoverMoreSection` and `CompletedSection`, and the layout is a container-query two-column scaffold (`src/components/LearningPaths/MyLearningTab.styles.ts:59-71`).
- Slice 2 — the static learning-altitude stat, and the streak tile rendered unconditionally (`src/components/LearningPaths/sections/HeroStats.tsx:42,50`).
- Slice 3 — badges as centered-icon tiles with the title beneath and a thin bottom progress bar (`src/components/LearningPaths/BadgeGridItem.tsx:62,81`).
- Slice 5 — Discover more driven by the upstream package listing, path-typed entries only, excluding titles shown elsewhere, capped at five (`src/learning-paths/useDiscoverMore.ts:100-110`).
- The Completed section, with a Done badge and no dates (`src/components/LearningPaths/sections/CompletedSection.tsx`).

**Superseded: slice 4**

This issue asked for My Courses to hold paths with progress strictly between 0 and 100, leaving untouched paths to Discover more. The shipped rule is `progress < 100` with no lower bound, so untouched paths stay in the list and are suppressed from Discover more (`src/components/LearningPaths/MyLearningTab.tsx:86,97`).

That was decided on purpose. It is the final commit on the #1464 branch, and both human approvals landed after it.

The reason is structural rather than a matter of taste. Discover more reads only the upstream package index and matches by exact title. The paths list reads the bundled catalogue plus App Platform paths. Those are different sets, so the rule this issue asked for can drop an untouched bundled path from the list without any guarantee it reappears in Discover more. The bundled catalogue is two paths, so a first-run user would see an empty section.

`src/components/LearningPaths/MyLearningTab.test.tsx:285-308` asserts the shipped rule directly. Implementing this issue as originally written would fail that test.

Two smaller divergences, both intentional: the section heading reads "My paths" rather than "My Courses", and the issue's claim that `useLearningPaths` already filtered on `0 < progress < 100` was never true — that expression was a sort key in the tab component, not a filter.

**What remains**

[...]

Issue #1462, opened 2026-07-30
My Learning Page - Design Iteration


moxious:
Desired eventual design for My Learning is going in this direction:

<img width="708" height="726" alt="Image" src="https://github.com/user-attachments/assets/2cf3d18c-4e7c-4683-aa0d-fbf57d81a8c8" />

Requested changes over time:
* Added My Courses for items assigned to the user
* Side by side my courses and Badges
* Moved learning paths down and renamed to Discover More
* Added Completed Section
* Added default learning mode to the hero at the top

This is filed as an epic, because it will require multiple tickets of work, and will require other features to be merged in order to enable some of the desired sections.

moxious:
## 🧭 Implementation readiness

**Stakes:** No code basis applies here. This issue is a design direction, not a task an agent can implement directly.

**Sanity check:** The issue's own body says "This is filed as an epic, because it will require multiple tickets of work." It carries no labels and no comments.

**Severity:** N/A — not an implementation issue.

**Labels:** Applied `agent/excluded`.

**Verdict:** ⚪ Excluded. Concrete sub-issues that implement pieces of this direction (for example #1463) get prepped on their own. (tracking: `agent/excluded`)

_Posted by the `prep-issues` tool, not written by a person._

PR #1581, opened 2026-08-10
feat(learning-paths): finish My Learning layout against the epic design

Spidy88:
## Summary

Finishes the My Learning layout against the [epic design](https://github.com/grafana/grafana-pathfinder-app/issues/1462), picking up the eight gaps the layout rework in #1464 left open: the columns now split 7:5 instead of 50/50, badges sit in a fixed three-column grid, both columns scroll in place behind a bottom fade instead of a "view all" toggle that dragged the sibling column down, Discover more gains its description and a disclosure chevron and adopts the My paths card chrome, and badges earned reads as a single accent-coloured count rather than a fraction.

Two behaviour fixes ride along. Card launch buttons now show a disabled state — launch is single-flight, so during the multi-second prepare step every card's button was correctly click-blocked but still rendered at full accent fill with a pointer cursor.

And the path cards are now operable by keyboard. Both card headers carried `role="button"` around real nested controls; that role is Children Presentational, so assistive tech was told to ignore Continue, Restart and the chevron. It also caused a double-toggle — Enter on the chevron fired the header's key handler *and* the button's activation click, cancelling out, so the cards could not be expanded from the keyboard at all. The disclosure semantics now live on the chevron, which was always a real focusable button.

my-learning-redesign

## What to look at

- **`src/components/LearningPaths/` — the layout.** The 7:5 split is a container query, so it resolves against the panel's own width and still stacks in the docked sidebar. Two style decisions look redundant but are load-bearing: `containerType` sits on a `columnsContainer` wrapper rather than the page container because `container-type` implies layout containment, which would make the container a containing block for `position: fixed` descendants and trap the [`BadgeDetailCard`](https://github.com/grafana/grafana-pathfinder-app/blob/472c79a8/src/components/LearningPaths/BadgeDetailCard.tsx) overlay inside the panel; and `gridAutoRows: 'min-content'` on the badge grid is required because as a flex child the grid gets a definite height, and default auto rows then compress below their intrinsic size and clamp two-line badge names to one line.
- **The fade mechanics.** `paddingBottom` deliberately lives on `scrollRegion`, not alongside the mask in `scrollRegionFaded`: padding counts toward `scrollHeight`, so a fade that supplied its own padding would keep its own overflow condition true and latch itself on. `flexShrink: 0` on the region's children is likewise required — without it the bounded height squashes every card to fit instead of scrolling.
- **Card accessibility (`LearningPathCard`, `DiscoverMoreCard`).** Neither header is a `role="button"` any more, so the nested controls keep their own semantics; the chevron carries `aria-expanded` + `aria-controls`, and the collapsed region is `aria-hidden` so a screen reader doesn't read content the toggle reports as collapsed. Removing the role also removed the *cause* of the double-toggle, so the keydown guard that previously worked around it is gone rather than retained. The header keeps a plain `onClick` as mouse convenience.
- **`src/hooks/useVerticalOverflow.ts`.** Returns a *callback* ref rather than a ref object — Discover more renders a skeleton first, so the measured list attaches on a later render, and an effect reading `ref.current` once on mount would leave the hook permanently dead for that section. It measures `scrollHeight - scrollTop - clientHeight`, i.e. "is there content below", not total overflow: a fade means "more below", so keying it on overflow alone dimmed the last row once the user reached the end.
- **`src/constants/testIds.ts` and `docs/developer/E2E_TESTING_CONTRACT.md`.** Adds `discover-more-expand-${id}` and removes the two "view all" ids along with the buttons. See Breaking changes.

## Why

The epic design is the target for My Learning, and #1464 delivered the structure but not the proportions or the scroll behaviour. The "view all" toggles were the most visible problem: expanding either column grew its panel and pushed the sibling column's content down the page, so the two-column layout only held together in one state.

The fade is applied conditionally rather than always because an unconditional mask dims content with nothing below it — with only two courses in a panel stretched to match a taller Badges column, and again at the bottom of any list once scrolled to the end. CSS cannot express either condition, which is why this needs a measured hook rather than a pure style.

## How to verify

1. Open My Learning at `/a/grafana-pathfinder-app/` in a wide panel. Confirm My paths and Badges split 7:5 (not 50/50), the badge grid is exactly three columns, and "Badges earned" shows a single number in the accent colour rather than `n/total`.
2. With only one or two paths in progress, confirm My paths shows **no** fade, its last card is at full opacity, and its panel bottom aligns with the Badges panel bottom.
3. Expand a My paths card until the list overflows. Confirm the fade appears at the panel's bottom edge, then scroll to the end and confirm the fade clears rather than dimming the last row.
4. Confirm Badges scrolls independently — the panel height must not change and the sibling column must not move.
5. In Discover more, confirm the description under the title, that cards match the My paths card styling, and that the chevron expands to reveal the package description.
6. Keyboard: Tab through a path card. Focus should go Continue → chevron, never the header itself. Press Enter on the chevron — the card should expand (before this change it did nothing, because the toggle fired twice and cancelled out).
7. Start a guide from any card and watch the other cards during the prepare step: every launch button should dim and take a `not-allowed` cursor rather than staying fully lit.
8. Click a badge tile. Confirm the detail overlay covers the **whole viewport**, not just the My Learning panel — this is the regression guard for the container-query containment noted above.
9. Narrow the panel below ~640px (or dock it in the sidebar). Confirm the columns stack and neither column becomes its own nested scroller inside the already-scrolling page.

## Breaking changes

Two `data-test-*` values are removed along with the buttons they identified: `learning-paths-show-all` and `learning-paths-show-all-badges`. Neither was part of the documented My Learning testid contract in `E2E_TESTING_CONTRACT.md`, and neither is referenced anywhere in `src`, `tests`, or `docs` after this change — but any out-of-tree selector targeting them will no longer match, and there is no replacement because the affordance itself is gone. The new `discover-more-expand-${id}` is additive and documented.

## Reversibility

Fully reversible. The change is presentational plus one new hook; it touches no storage format, completion or progress semantics, analytics payload, or backend contract, so a revert restores prior behaviour with no migration or manual recovery. The only public-contract surface is the two removed testids described above, which a revert would restore verbatim.

## Out of scope

- Assignments, due dates, and per-item completion dates from the epic design remain deferred — there is no data basis for them yet, as scoped in #1463.
- Hover shadows on cards and badge tiles are clipped at the scroll regions' edges, because `overflow-y: auto` forces `overflow-x` to compute to `auto`. Cosmetic, and the obvious fix (a negative-margin gutter) would perturb the fade geometry — worth its own change if it bothers anyone.
- Two i18n keys (`myLearning.pathsDescription`, `myLearning.learningPaths`) were already unused before this branch and are left in place to keep the diff focused. The new `en-US` keys are absent from the other 20 locales, which have no `myLearning` block at all and fall back to the English defaults.

Addresses #1462. Part of #1463.


moxious (dismissed):
On the specific things you asked to look into - test-id selectors are not a concern.  I checked, nothing in this repo references showAllPathsButton/showAllBadgesButton or the removed i18n keys, and neither was ever listed in E2E_TESTING_CONTRACT.md's documented contract

Scrollbars / expanding-content concern: this is a legitimate design trade-off, not a defect - OK with it.

---

1. Medium / high confidence — a11y inconsistency the PR itself proves is fixableLearningPathCard.tsx's header div still carries role="button" tabIndex={0} around three real, nested <button> elements (Continue, Restart, chevron) — the "Children Presentational" ARIA anti-pattern (WCAG 4.1.2 / axe nested-interactive rule) that hides those buttons' semantics from assistive tech. In this same PR, the new sibling DiscoverMoreCard.tsx explicitly identifies and avoids this exact pattern with a comment ("Deliberately not role="button"... would hide the nested Start and chevron from assistive tech"), relying instead on the chevron — already an unconditional, focusable <button> in both components — as the sole keyboard path. The PR is already editing LearningPathCard's onKeyDown on this identical div (to fix the double-toggle bug) and patches around a symptom of this same structural problem rather than removing the cause. Adversarial verification (3/3 skeptics) confirmed this survives: the fix transfers cleanly (drop role/tabIndex/onKeyDown from the header, keep onClick as mouse convenience, the chevron already carries aria-expanded duty) — this isn't scope creep, it's the same PR solving the identical problem two ways.Suggested action: apply the same fix to LearningPathCard.tsx's header before merge, or file a same-day follow-up if you want to keep this PR focused.
2. Low-medium / medium confidence — untested dynamic-children path in useVerticalOverflow
The hook's MutationObserver → observeSubtree() re-observe cycle (fires when a list's children change while the container stays mounted — e.g., a new badge appended) has zero test coverage; every growth/shrink test in useVerticalOverflow.test.ts drives the mocked ResizeObserver callback directly, never mutates real children. Confirmed independently by both the testing-and-verification reviewer and the tech-debt scan (category D2), and confirmed still true against the current head (the observer logic is unchanged by the later commits). Not blocking — the code is short and visibly correct — but it's the one meaningfully untested branch in an otherwise thorough test file, and it's the mechanism a "badge count updates while the panel is open" scenario actually depends on.

3. Low / reduced confidence (disagreement noted) — overlay-containment test may not test what it claims
A reviewer flagged that the test named for keeping BadgeDetailCard outside the columnsContainer containment boundary never actually triggers selectedBadge/renders BadgeDetailCard, so it wouldn't catch a future regression nesting the overlay inside the container. Adversarial verification split 2 refuted / 1 not-refuted — but one "refuted" vote checked the wrong git ref and wrongly concluded the code didn't exist at all, which isn't a valid rebuttal. The underlying JSX is verified correct today (the overlay renders as a sibling, unaffected by this diff), so this is a residual test-completeness note, not a live bug — flagging at low severity given the split verdict.

Jayclifford345 (approved):

Issue #1262, opened 2026-07-07
Reduce excessive padding in numbered list containers


Jayclifford345:
## Context

The numbered list container inside content panels ("Explore your data" and similar) has noticeably oversized padding that should be tightened up.

> "Why the ginormous padding here? let's reduce"
> "here as well, ginormous."

<img width="1091" height="845" alt="Image" src="https://github.com/user-attachments/assets/dff83a38-3547-4331-a5e1-200d43ec7b3c" />

## Code pointers

- `src/components/interactive-tutorial/interactive-section.tsx:1344` — `<ol className="interactive-section-content">` renders the guide's numbered steps
- `src/styles/interactive.styles.ts` (lines 368-426) — `.interactive-section-content` has `padding: theme.spacing(2)` (16px), plus `li[data-numbered="true"]` adds `paddingLeft: theme.spacing(4)` (32px) for the counter number, stacking to ~48px of left offset. There's an existing inline comment here referencing #841 documenting the alignment math — read that before changing the values, since it explains why the padding was set this way.

---
Part of the Pathfinder UX review epic — [Figma board](https://www.figma.com/design/MuEIAafuIdrJOljvIy7E0T/Pathfinder-UX-review?node-id=1-4124)

moxious:
Design decision needed: specify the target padding values from the Figma board before making any changes — the current values carry an explicit alignment invariant that must be respected.

The padding values in `src/styles/interactive.styles.ts` (lines 368-426) are documented with an alignment constraint originally introduced in issue #841: `paddingLeft: theme.spacing(4)` on `li[data-numbered]`, the `top: theme.spacing(2)` on the `::before` counter, and the extra `paddingLeft` on `data-step="false"` items all interlock to keep step numbers, card titles, and plain-text list items horizontally aligned. Changing any one value without adjusting the others will break this alignment.

The issue asks to "reduce" padding but does not specify target values, and the Figma board is referenced but its measurements are not described in the issue body.

**Implementation basis (once decision is made)**
All changes are confined to `src/styles/interactive.styles.ts:368-426`. Once target spacing values are confirmed from Figma, the implementation is a mechanical substitution — but the inter-dependent values (`paddingLeft`, `::before top`, step-card `marginTop` / border widths) must be updated together to preserve alignment.

dgehring-grafana:
[...]

PR #1583, opened 2026-08-11
ui: tighten guide margins and consolidate feedback chrome (#1262)

dgehring-grafana:
## Summary

- Align guide spacing and layout with the Figma spec for Pathfinder guides, addressing oversized numbered-list padding called out in [#1262](https://github.com/grafana/grafana-pathfinder-app/issues/1262).
- Goes a bit outside the ticket’s narrow scope: the Figma margin changes look inconsistent unless the same rhythm is applied across related guide chrome (requirement / error / quiz / challenge / code / terminal feedback, action-button padding, section vs standalone cards).
- Pulls repeated feedback-box styling out of per-component aberrations into `interactive.styles.ts` (shared `.interactive-feedback-box` + variants), keeping only layout or intentional accents local. Prefers flex where block layout was brittle.
- Adds a DevTools-only style-validation spray guide for manual visual QA while we don’t have Storybook. Happy to remove on request.

## Alternatives considered

Ticket comments suggested keeping changes inside `interactive.styles.ts` only. That wasn’t enough once the “complete previous step” / feedback chrome was in play — several components duplicated or owned that styling themselves. Centralizing the shared shell and wiring classNames was cleaner and more consistent than restyling each aberration in isolation.

`GUIDE_ACTION_BUTTON_SELECTOR` is a bit blunt, but Figma specifies specialized guide button padding. A single scoped selector is more targeted, easier to revert, and less invasive than tagging every button individually.

## Test plan

- [ ] Open a real user guide (e.g. first-dashboard): check section numbering, card gaps, completed/skipped padding
- [ ] Enable Dev Mode (`?page=configuration&dev=true`), open DevTools → **Style validation**, walk the spray board (§1–10)
- [ ] Confirm requirement / error / quiz / challenge / code / terminal feedback boxes share the compact shell
- [ ] Confirm spray guide is **not** in `index.json` (no end-user recommendations)
- [ ] CI: bundled-repository + loader consistency pass with new manifest / `repository.json` entry

## Gotchas

- Spray board + DevTools opener are out of scope for #1262; local fixture only (`category: reference`, not in `index.json`).
- Visual blast radius is **all** ContentRenderer-mounted guides (sidebar / floating / guide-reader / block preview), not just the ticket’s numbered list.
- Not every component has been altered, and some inconsistencies exist, though minor. Examples are code blocks and standalone inputs, which have retained their margins. Will happily address here if it's deemed necessary, or in another pass.
- The step number has moved from the list item to inside the card. It's absolutely positioned in the same way it was before, but it makes misalignment if the first item happens to deviate from standard height more noticeable.


src/styles/interactive.styles.ts
```diff
@@ -497,9 +529,6 @@ const getInteractiveComponentStyles = (theme: GrafanaTheme2) => ({
   // because the paused state is recoverable via the prompt above.
   '.interactive-section-alignment-banner': {
     margin: `${theme.spacing(1.5)} ${theme.spacing(2)} 0`,
-    padding: '8px 12px',
-    background: theme.colors.info.transparent,
-    border: `1px solid ${theme.colors.info.border}`,
     borderLeft: `3px solid ${theme.colors.info.main}`,
```
tomglenn, line 532: This 3px accent no longer survives. `.interactive-section-alignment-banner` is declared at line 530, but the banner now also carries `interactive-feedback-box interactive-feedback-box--info`, and those rules land later in the same object at equal specificity (`.css-hash .interactive-feedback-box` vs `.css-hash .interactive-section-alignment-banner` — both one class deep). So `border: 1px solid transparent` and then `borderColor: info.border` win the cascade for `border-left-width`/`border-left-color`, and the intentional info-blue accent renders as an ordinary 1px border.

Nesting the accent under the variant (or otherwise raising specificity) keeps it:

```suggestion
    borderLeft: `3px solid ${theme.colors.info.main}`,
    borderRadius: '4px',

    '&.interactive-feedback-box': {
      borderLeftWidth: '3px',
      borderLeftColor: theme.colors.info.main,
    },
```

Same cascade hazard applies to `challenge-block.tsx`'s `hint` (`borderLeft: 3px solid info.border` against the base box's `border: 1px solid transparent`). That one happens to win today only because the component's emotion class is inserted after the panel's during render — worth not relying on.

src/styles/interactive.styles.ts
```diff
@@ -382,47 +400,81 @@ const getInteractiveComponentStyles = (theme: GrafanaTheme2) => ({
...
+    '& > li[data-step="true"] > *': {
+      marginTop: 0,
+      marginBottom: 0,
+    },
+
     '& > li[data-numbered="true"]': {
       counterIncrement: 'step-counter',
+    },
+
+    '& > li[data-numbered="true"][data-step="true"] > :first-child': {
```
tomglenn, line 416: Moving the gutter and the `::before` counter from the `<li>` onto the card makes the number's horizontal position depend on the card's border width, because an absolutely positioned `::before` resolves `left: 0` against its containing block's **padding** box — i.e. inset by `border-left-width`.

Concretely, in one mixed numbered list:

- `.interactive-step` family (step, guided, multi-step, terminal, code-block) — `border: 2px solid transparent` → number and title both at `li.x + 2`
- quiz (`interactive-quiz.tsx:585`) and challenge (`challenge-block.tsx:96`) — `border: 1px solid border.weak` → `li.x + 1`
- plain markdown (`data-step="false"`, line 457) — no border, `::before` still on the `<li>` → `li.x + 0`

So a section mixing prose, a step, and a quiz gets three different left edges for the numbers, off by 1–2px. That's exactly what the removed `calc(${theme.spacing(4)} + ${theme.spacing(2)} + 2px)` comment was compensating for in #841.

Keeping `::before` on the `<li>` (which has no border, so it's a stable reference) and giving the card only the horizontal inset would avoid per-block-type compensation entirely.

src/styles/interactive.styles.ts
```diff
@@ -382,47 +400,81 @@ const getInteractiveComponentStyles = (theme: GrafanaTheme2) => ({
       listStyle: 'none',
     },
 
+    // Interactive blocks keep local margins for standalone guide-root and
+    // inline-conditional mounts. Inside a section, this flex container's gap
+    // is the single vertical rhythm, so suppress those component-owned margins.
+    // Scope to data-step=true to preserve authored prose and media spacing.
+    '& > li[data-step="true"] > *': {
```
tomglenn, line 407: This zeroes margins on **every** child of the `<li>`, but the gutter inset below only lands on `> :first-child`. Previously the `<li>` itself owned `paddingLeft`, so all children were inset regardless of how many there were.

`data-step="true"` is set for any child whose `type` isn't a string (`section-numbering.tsx:59`) — that's "is a React component", not "renders a single padded card". A component returning a fragment with sibling roots gets only its first root inset; the rest sit flush under the number. Same for any component that isn't a self-padded card — the default `::before { top: theme.spacing(2) }` assumes `spacing(2)` of card padding, so content starting at 0 renders 16px above its own number.

Not necessarily reachable with today's block set, but the assumption is now load-bearing and invisible. Worth either scoping the inset to `> *` to match this rule, or noting the single-padded-root contract where `isStep` is computed.

src/styles/interactive.styles.ts
```diff
@@ -326,6 +336,11 @@ const getInteractiveComponentStyles = (theme: GrafanaTheme2) => ({
     },
   },
 
+  // Per Figma design, buttons in guides have specialized padding
+  [GUIDE_ACTION_BUTTON_SELECTOR]: {
```
tomglenn, line 340: Two things about the breadth here:

1. `> button` matches regardless of `size`, so a future `size="md"` or `"lg"` button in any of these seven containers silently gets `sm` horizontal padding. Every call site is `size="sm"` today, so it's inert — but it'll be a confusing override to track down later.
2. It only matches `<button>` children of those containers, which misses the hand-rolled buttons in the same chrome — `interactive-requirement-retry-btn`, `interactive-guided-fix-btn`, `interactive-lazy-retry-btn`, `interactive-requirement-ai-fix-btn` — each of which keeps its own padding. So "guide buttons have specialized padding" isn't actually uniform after this.

Not blocking, but if the goal is one Figma padding for guide buttons, those four are the remaining gap.

src/components/interactive-tutorial/interactive-guided.tsx
```diff
@@ -836,7 +836,7 @@ export const InteractiveGuided = forwardRef<{ executeStep: () => Promise<boolean
         {(uiState === STEP_STATES.REQUIREMENTS_UNMET || (uiState === 'checking' && checker.explanation)) &&
           checker.explanation && (
             <div className={`interactive-guided-requirements${checker.isChecking ? ' rechecking' : ''}`}>
-              <div className="interactive-guided-requirement-box">
+              <div className="interactive-feedback-box interactive-feedback-box--neutral interactive-requirement-box interactive-guided-requirement-box">
```
tomglenn, line 839: `interactive-requirement-box` isn't defined anywhere — not in `interactive.styles.ts`, not in a component stylesheet, and nothing selects it in tests or e2e. It's added at four sites (here, `interactive-step.tsx:1220`, `interactive-multi-step.tsx:915`, `interactive-section.tsx:1315`) and is a no-op at all four.

If it's meant as the shared hook for requirement-flavoured boxes, it needs a rule; otherwise drop it so the class lists stay honest:

```suggestion
              <div className="interactive-feedback-box interactive-feedback-box--neutral interactive-guided-requirement-box">
```

tomglenn (approved):
Nice consolidation — `.interactive-feedback-box` + variants is the right shape, and pulling the duplicated shells out of five components makes the chrome much easier to reason about. Typecheck and the interactive-tutorial / package-engine / validation suites pass locally (the one failure, `import-graph.test.ts › isNodeBuiltin › node:test`, is unrelated — `import-graph.ts` is untouched here).

**Worth fixing before this leaves draft:**

- `interactive.styles.ts:532` — `.interactive-section-alignment-banner`'s `borderLeft: 3px solid info.main` is declared at line 530, but `.interactive-feedback-box` (665) and `--info` (686) set `border`/`borderColor` later at equal specificity, so the deliberate left accent collapses to a plain 1px border.
- `interactive.styles.ts:416` — moving `paddingLeft` and `::before` from the `<li>` onto the card makes the number gutter depend on each card's border width: 2px for the `.interactive-step` family, 1px for quiz/challenge, 0 for plain markdown. That's the alignment the deleted `calc(spacing(4) + spacing(2) + 2px)` comment was compensating for (#841).

**Smaller:**

- `interactive-requirement-box` is added to four components and defined nowhere — drop it or define it.
- `GUIDE_ACTION_BUTTON_SELECTOR` is size-agnostic and skips the non-Grafana `<button>`s in the requirement/lazy chrome.

On the spray guide: I'd keep it. It's consistent with `e2e-framework-test` / `json-guide-demo` / `block-editor-tutorial` (in `repository.json`, absent from `index.json`), all its `reftarget`s are self-referential, and it's the only way to eyeball this blast radius without Storybook.

tomglenn (changes requested):
Correcting my own verdict — I approved this a moment ago, but two of the inline findings are behavioural regressions rather than suggestions, so requesting changes is the honest state. The inline comments already on the PR stand as-is; nothing new to read.

Blocking:

- `interactive.styles.ts:532` — the alignment banner's 3px info accent is lost to the later `.interactive-feedback-box--info` rule at equal specificity.
- `interactive.styles.ts:416` — the numbered-list gutter now varies by card border width (0/1/2px), reintroducing the #841 misalignment.

The two smaller items (`interactive-requirement-box` dead class, `GUIDE_ACTION_BUTTON_SELECTOR` breadth) are non-blocking, and I still think the spray guide is worth keeping. Everything positive in the previous review holds — the `.interactive-feedback-box` consolidation is the right shape.

PR #1587, opened 2026-08-11
fix(custom-guides): count catalogue hard failures, not just soft-200s

### LINKED ISSUE #1562: Custom-guide catalogue hard failures (503/network) emit no frontend telemetry (#1408 follow-up)
Follow-up from the second-round review of PR #1408 (https://github.com/grafana/grafana-pathfinder-app/pull/1408). Companion to https://github.com/grafana/grafana-pathfinder-app/issues/1555 (the remaining drain `*_ms` + locked-milestone signals); filed separately because it is a one-file fix an agent can land independently.

## Severity

**Low-medium** (observability gap, no data pollution). Confirmed by 3 adversarial skeptics — all three independently concluded the gap is an internal inconsistency with the PR's own stated purpose for the telemetry op, not telemetry-maximalism.

## Analysis

PR #1408 added `recordCustomGuideCatalogueUnavailable(reason)` (`src/lib/telemetry/facade.ts:108-110`) with an in-file rationale citing the stackId-wipe incident: "the Faro event is the countable, alertable signal" for the invisible-empty-catalogue class. But it only fires on the **soft-200** envelope (`capability.available === false`, `src/lib/custom-guide-repository-client.ts:78-86`).

Hard failures never reach it: the chain is `requestCatalogue().then(cacheFill).catch(() => []).finally(...)`, and the `.catch` at `custom-guide-repository-client.ts:123` maps every rejection to `[]` with no `logger` call and no Faro op. The request also passes `showErrorAlert: false` (`:76`), so Grafana's backendSrv surfaces nothing, and the Faro adapter has no fetch instrumentation to compensate (`src/lib/telemetry/faro-adapter.ts:76-84` enables only Errors/Session/View/Performance; resource timings carry no status codes and are filtered to docs/recommender hostnames).

The backend's own comment documents exactly the mode this misses: a wrong CAP token or unreachable auth-api "503s this route indefinitely" (`pkg/plugin/custom_guide_repository.go:165-173`, transient branch → `http.StatusServiceUnavailable`). Under that mode the surface renders empty forever, a dashboard on the new op reads zero, and the only signal is a backend Info log in Loki. Callers cannot log it either — `fetchCustomGuideRepository` never rejects.

## Implementation

1. In the `.catch` at `src/lib/custom-guide-repository-client.ts:123`, before returning `[]`: `logger.warn('[custom-guides] catalogue fetch failed', { reason: 'transport-error' })` and `recordCustomGuideCatalogueUnavailable('transport-error')`. If the error carries an HTTP status (backendSrv `FetchError.status`), use a bounded `http-<status>` reason instead, mirroring the backend's `upstream-<status>` pattern. Keep the reason out of RudderStack (the op is Faro-only by design — see the comment at `facade.ts:102-107`).
2. Update the two existing rejection tests (`src/lib/custom-guide-repository-client.test.ts` ~`:83` "returns an empty array and swallows errors on fetch failure" and ~`:109` "does not cache failures") to also assert the logger + op calls.

## Links

- Sibling telemetry scope: https://github.com/grafana/grafana-pathfinder-app/issues/1555
- Policy: `docs/developer/TELEMETRY.md` rule 1 (a degradation ladder needs a countable, alertable event).

tomglenn:
## Summary

The custom-guide catalogue fetch has two failure paths and only one was observable. The soft-200 path (backend returns HTTP 200 with `capability.available: false`) logs and emits `pathfinder_custom_guide_catalogue_unavailable`. The hard path — the request itself rejecting on a 503, a network drop, or a timeout — ended in a bare `.catch(() => [])` that discarded the error: no log, no telemetry.

This makes the hard path emit the same pair of signals, split by cause:

- the rejection carries an HTTP status → `http-<status>` (e.g. `http-503`)
- no status at all (network failure, DNS, abort, timeout) → `transport-error`

Status extraction follows the existing in-repo pattern at `src/utils/fetchBackendGuides.ts:47-51` (`status` / `statusCode` / `data.statusCode`), bounded to the 100-599 range so the token stays low-cardinality.

Fixes #1562

## What to look at

- `src/lib/custom-guide-repository-client.ts` — `classifyRequestFailure`, and the `.catch` in `fetchCustomGuideRepository` that now logs and records before returning `[]`.
- `src/lib/custom-guide-repository-client.test.ts` — the two tests that pinned the silent behaviour are replaced/extended; new coverage for a status-bearing rejection, the nested `data.statusCode` shape, and a status-less rejection.
- `src/lib/telemetry/facade.ts` — comment only. The op's rationale said "a soft-200 with a machine `reason`"; it now covers both causes. No signature or `TELEMETRY_EVENTS` change.

Telemetry policy (`docs/developer/TELEMETRY.md`): the Faro event attribute gets only the bounded token — never `err.message`, never the response body. The raw detail goes to `logger.warn` (not `logger.error`, which escalates to a Faro exception, and which would be the wrong severity for a best-effort listing). `recordCustomGuideCatalogueUnavailable` is `pushFaroEvent`-backed and Faro-only; there is no `reportInteraction` path anywhere in `src/lib/telemetry/`, so nothing here can reach RudderStack.

## Why

The backend genuinely returns hard 503s — a wrong CAP token or an unreachable auth-api 503s the route indefinitely (`pkg/plugin/custom_guide_repository.go`, transient branch). The request passes `showErrorAlert: false`, so Grafana's own toast is suppressed, and the Faro adapter has no fetch instrumentation to compensate. Under that mode the Custom Guides surface renders empty forever, a dashboard on `pathfinder_custom_guide_catalogue_unavailable` reads zero, and the only trace is a backend log. Neither caller can compensate, because `fetchCustomGuideRepository` never rejects.

Splitting by cause is the operational point: it separates "our backend is down" (a page-worthy `http-5xx` cluster) from "their connection died" (`transport-error`, expected background noise).

## How to verify

```bash
eval "$(fnm env --shell bash)" && fnm use 24.18.0
npx jest src/lib/custom-guide-repository-client.test.ts src/lib/telemetry --coverage=false
npm run typecheck
```

The four new/updated assertions fail on `main` (verified by stashing the source change: 4 failed, 8 passed) and pass with it (12/12). Focused run over the client plus the whole telemetry directory: 172 passed. `typecheck`, `eslint`, and `prettier --check` are clean on all three changed files.

Manually: on a stack where the catalogue route 503s, open the Custom Guides surface and confirm a `[custom-guides] catalogue fetch failed` warning in the console with `reason: http-503`, plus one `pathfinder_custom_guide_catalogue_unavailable` event with the same reason.

## Breaking changes

None. `fetchCustomGuideRepository` still resolves to `[]` on every failure and still never rejects — two places document a dependency on that (the function's own docblock and `useCustomGuideCatalogueOnOpen.ts:26-28`). Failures are still not cached, so a transient error does not stick for the TTL. The only new outputs are one log line and one Faro event per failed fetch.

## Reversibility

Runtime-reversible; telemetry history remains. Revert the commit and the runtime behaviour returns to exactly what it was — no schema, storage, or API surface is touched, and `TELEMETRY_EVENTS` is unchanged, so no dashboard or alert breaks on a revert (the event simply stops carrying `http-*` / `transport-error` reasons). What a revert cannot undo is the Faro events and logs already ingested: those survive until retention expires.

## Out of scope

A 200 response with `capability.available: true` but a missing or non-array `guides` field also returns empty silently (`custom-guide-repository-client.ts:89`, pinned by the "returns an empty array when the response is malformed" test). Same symptom — empty surface, no signal — but a different cause: a malformed success envelope rather than a transport or capability failure, so it wants a different reason token and arguably a different severity. Deliberately left alone here; filed as #1591.

### DIFF
--- a/src/lib/custom-guide-repository-client.test.ts
+++ b/src/lib/custom-guide-repository-client.test.ts
@@ -121,56 +121,62 @@ describe('fetchCustomGuideRepository', () => {
     expect(result).toEqual([]);
   });
 
-  it('records an http-<status> reason when the request rejects with a status', async () => {
-    mockGet.mockRejectedValue({ status: 503, statusText: 'Service Unavailable', message: 'upstream drain failed' });
-
-    const result = await fetchCustomGuideRepository('stacks-123');
-
-    expect(result).toEqual([]);
-    expect(recordCustomGuideCatalogueUnavailable).toHaveBeenCalledWith('http-503');
-    expect(logger.warn).toHaveBeenCalledWith('[custom-guides] catalogue fetch failed', {
+  // The reason lands on a Faro event attribute and in the bridged log context,
+  // so every rejection shape must collapse to a token from the closed set.
+  it.each([
+    { shape: 'a top-level status', err: { status: 503, statusText: 'Service Unavailable' }, reason: 'http-503' },
+    { shape: 'a top-level statusCode', err: { statusCode: 418 }, reason: 'http-418' },
+    { shape: 'a nested data.statusCode', err: { data: { statusCode: 502 } }, reason: 'http-502' },
+    {
+      shape: 'status ahead of both statusCode shapes',
+      err: { status: 503, statusCode: 404, data: { statusCode: 500 } },
       reason: 'http-503',
-      message: 'upstream drain failed',
-    });
-  });
-
-  it('reads the status off the nested statusCode shapes too', async () => {
-    mockGet.mockRejectedValue({ data: { statusCode: 502 } });
-
-    await fetchCustomGuideRepository('stacks-123');
-
-    expect(recordCustomGuideCatalogueUnavailable).toHaveBeenCalledWith('http-502');
-  });
-
-  it('records transport-error when the rejection carries no status', async () => {
-    mockGet.mockRejectedValue(new Error('network error'));
+    },
+    {
+      shape: 'statusCode ahead of data.statusCode',
+      err: { statusCode: 404, data: { statusCode: 500 } },
+      reason: 'http-404',
+    },
+    { shape: 'one below the low bound', err: { status: 99 }, reason: 'transport-error' },
+    { shape: 'the low bound', err: { status: 100 }, reason: 'http-100' },
+    { shape: 'the high bound', err: { status: 599 }, reason: 'http-599' },
+    { shape: 'one above the high bound', err: { status: 600 }, reason: 'transport-error' },
+    { shape: 'a wildly out-of-range status', err: { status: 99999 }, reason: 'transport-error' },
+    { shape: 'a negative status', err: { status: -503 }, reason: 'transport-error' },
+    { shape: 'a string status', err: { status: '503' }, reason: 'transport-error' },
+    { shape: 'a non-integer status', err: { data: { statusCode: 503.0000001 } }, reason: 'transport-error' },
+    { shape: 'no status at all', err: new Error('network error'), reason: 'transport-error' },
+  ])('records $reason for $shape', async ({ err, reason }) => {
+    mockGet.mockRejectedValue(err);
 
     const result = await fetchCustomGuideRepository('stacks-123');
 
     expect(result).toEqual([]);
-    expect(recordCustomGuideCatalogueUnavailable).toHaveBeenCalledWith('transport-error');
-    expect(logger.warn).toHaveBeenCalledWith('[custom-guides] catalogue fetch failed', {
-      reason: 'transport-error',
-      message: 'network error',
-    });
+    expect(recordCustomGuideCatalogueUnavailable).toHaveBeenCalledWith(reason);
+    expect(logger.warn).toHaveBeenCalledWith('[custom-guides] catalogue fetch failed', { reason });
   });
 
-  // A malformed error shape must not mint an unbounded Faro attribute value.
-  it('keeps the reason bounded when the status is out of range or not an integer', async () => {
-    mockGet.mockRejectedValueOnce({ status: 99999 });
-    await fetchCustomGuideRepository('stacks-123');
-    expect(recordCustomGuideCatalogueUnavailable).toHaveBeenLastCalledWith('transport-error');
+  // logging.ts sanitizes the log context but does not strip it, so anything put
+  // there reaches Faro — an error message would be a user-derived free-text
+  // attribute, which docs/developer/TELEMETRY.md forbids.
+  it('never forwards the error message into the bridged log context', async () => {
+    const sentinel = 'c0ffee-user-derived-detail';
+    mockGet.mockRejectedValue({ status: 503, message: `upstream drain failed for ${sentinel}` });
 
-    mockGet.mockRejectedValueOnce({ data: { statusCode: 503.0000001 } });
     await fetchCustomGuideRepository('stacks-123');
-    expect(recordCustomGuideCatalogueUnavailable).toHaveBeenLastCalledWith('transport-error');
+
+    expect(logger.warn).toHaveBeenCalledWith('[custom-guides] catalogue fetch failed', { reason: 'http-503' });
+    expect(JSON.stringify((logger.warn as jest.Mock).mock.calls)).not.toContain(sentinel);
   });
 
   // Both callers document that this never rejects — a throwing sink must not break that.
-  it('still resolves to an empty array when the observability calls throw', async () => {
+  it.each([
+    { sink: 'the logger', mock: () => logger.warn as jest.Mock },
+    { sink: 'the telemetry facade', mock: () => recordCustomGuideCatalogueUnavailable as jest.Mock },
+  ])('still resolves to an empty array when $sink throws', async ({ mock }) => {
     mockGet.mockRejectedValue(new Error('network error'));
-    (logger.warn as jest.Mock).mockImplementationOnce(() => {
-      throw new Error('logger blew up');
+    mock().mockImplementationOnce(() => {
+      throw new Error('observability blew up');
     });
 
     await expect(fetchCustomGuideRepository('stacks-123')).resolves.toEqual([]);
--- a/src/lib/custom-guide-repository-client.test.ts
+++ b/src/lib/custom-guide-repository-client.test.ts
@@ -155,6 +155,27 @@ describe('fetchCustomGuideRepository', () => {
     });
   });
 
+  // A malformed error shape must not mint an unbounded Faro attribute value.
+  it('keeps the reason bounded when the status is out of range or not an integer', async () => {
+    mockGet.mockRejectedValueOnce({ status: 99999 });
+    await fetchCustomGuideRepository('stacks-123');
+    expect(recordCustomGuideCatalogueUnavailable).toHaveBeenLastCalledWith('transport-error');
+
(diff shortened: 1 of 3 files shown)


tomglenn:
Adversarial review pass. The design holds up — behaviour is unchanged (still resolves `[]`, still no caching of failures), the reason token never carries error text, and the event rate is bounded by mount-guarded callers plus the in-flight dedupe. Two gaps found and pushed as a follow-up commit.

**1. The `.catch` could now reject.** Moving `logger.warn` + `recordCustomGuideCatalogueUnavailable` into the catch body means a throw from either propagates, and `.finally` passes it through — so `fetchCustomGuideRepository` rejects. Confirmed with a probe: with a throwing `logger.warn`, the promise rejected rather than resolving to `[]`. Neither sink can actually throw today (the sanitizer is internally try/caught, and the bridge sinks are `guardTelemetry`-wrapped), so this is not a live bug — but the bare `.catch(() => [])` was structurally incapable of it, and three callers lean on that, one of them (`fetchAppPlatformLearningPaths`) with no handler at all. The pair now sits behind `reportCatalogueFetchFailure`, which swallows its own failures on the same principle as `guardTelemetry`.

**2. The cardinality bound let non-integers through.** `{ data: { statusCode: 503.0000001 } }` produced `http-503.0000001`. Since `data.statusCode` is read off the response body rather than the fetch layer, the bound is the only thing keeping that attribute low-cardinality per `docs/developer/TELEMETRY.md`. Added `Number.isInteger`.

**On test coverage:** the four listed cases all discriminate — reverting to the bare catch fails 4, dropping the nested `data.statusCode` read fails 1, breaking the message extraction fails 2, caching the failure fails 1. But the range check itself was untested: replacing `status >= 100 && status <= 599` with a bare `typeof status === 'number'` left all 12 tests green, so the guard the telemetry-policy argument rests on could have been deleted silently. Both guards are now pinned.

Everything else checked out. `err.message` cannot carry the response body (that is `err.data`), the log context is strictly more conservative than the repo-wide `logger.warn('...', { error })` convention, and the facade.ts comment edit is an accuracy fix on a comment that was going stale.

Full suite green: 417 suites, 6637 passed / 18 skipped / 5 todo. `typecheck`, `eslint`, and `prettier --check` clean.

moxious (changes requested):
Thanks for splitting the hard-failure path out — the classifier, caching behavior, and the "never rejects" contract all check out. One privacy/telemetry issue needs fixing before this lands, plus a couple of smaller items.

**[blocking]** `src/lib/custom-guide-repository-client.ts:92` — The typed event's `reason` is bounded, but the raw error message also reaches Faro through the `logger.warn` context (`logging.ts:50-53` only sanitizes/truncates it, it doesn't strip it). That conflicts with `docs/developer/TELEMETRY.md:60-63`, which forbids high-cardinality or user-derived free-text attributes. Please keep the warning but pass only `{ reason }`, and add a test asserting an error-message sentinel is not forwarded. If more diagnostic detail is needed, classify it into another closed token rather than sending `err.message` verbatim.

**[suggestion]** `src/lib/custom-guide-repository-client.test.ts:137-176` — Could we pin the complete new contract with table-driven cases for the top-level `statusCode` branch, precedence when more than one status shape is present, the boundary values `99/100/599/600`, and a negative/string value? As written, the suite wouldn't catch removal of the `statusCode` branch or an off-by-one in the valid range. Also, "observability calls throw" currently only exercises the `logger.warn` throw path, not `recordCustomGuideCatalogueUnavailable` throwing.

**[nit]** PR description, Reversibility — the runtime change is cleanly revertible, but the new Faro events/logs already ingested survive a code revert until retention expiry. Might be worth softening "fully reversible" to "runtime-reversible; telemetry history remains."

**[question]** `src/lib/custom-guide-repository-client.ts:116` — Agreed this can stay out of scope here (the backend contract always emits an array and the guard is fail-safe), but could you file/link the promised follow-up issue for the malformed-`guides` case before merge? Today that schema-drift case is cached as a successful empty catalogue with no signal, which is operationally indistinguishable from "no guides authored."

tomglenn:
All four items addressed in 4bad2ef. Point by point:

**[blocking] raw error message reaching Faro — fixed.** You're right, and the path is exactly as you describe: `logger.warn`'s context goes to `pushFaroLog` via `sanitizeContext`, which maps each value through `sanitizeForLogging` and truncates — it never drops a key. So `message: err.message` arrived as a user-derived free-text attribute, which `docs/developer/TELEMETRY.md` forbids. The warning stays and now carries only `{ reason }`, the same bounded token as the typed event. `requestFailureMessage` had no other caller, so it's gone rather than left dead. I did not try to salvage any of the detail into a second token — nothing in the message is classifiable into a closed set without inventing a parser over upstream error text, and the status branch already carries the part that is.

The new test builds a rejection whose message embeds a sentinel, asserts the exact context is `{ reason: 'http-503' }`, and separately scans every `logger.warn` call for the sentinel. Re-adding the `message` field fails it (and 14 others).

**[suggestion] test table — done, and the gap was real.** The status cases are now a 14-row table covering: the top-level `status` branch, the top-level `statusCode` branch, the nested `data.statusCode` branch, precedence between all three (`status` ahead of both, `statusCode` ahead of `data.statusCode`), the boundaries 99/100/599/600, a wildly out-of-range value, a negative value, a string value, a non-integer, and no status at all. Each row asserts the resolved token on both sinks — the facade call and the log context.

Your two specific worries both confirmed by mutation: deleting the top-level `statusCode` branch previously left the suite green and now fails 2 rows; the `100–599` bound now fails 4 rows when removed, and `Number.isInteger` fails 1. Inverting the precedence order fails 2. The whole discrimination pass:

| mutation | failing tests |
| --- | --- |
| re-add the `message` field to the log context | 15 |
| drop `Number.isInteger` | 1 |
| drop the `100–599` bound | 4 |
| drop the top-level `statusCode` branch | 2 |
| drop the nested `data.statusCode` branch | 1 |
| invert branch precedence | 2 |
| remove the try/catch around the observability calls | 2 |
| revert to the bare `.catch(() => [])` | 16 |
| cache the failure result | 1 |

The throwing-sink test is now a two-row table, so `recordCustomGuideCatalogueUnavailable` throwing is covered alongside `logger.warn` throwing — removing the try/catch fails both, where before it only failed one.

The three invariants are unchanged and still pinned: resolves `[]` (every table row asserts it), never rejects (the two throwing-sink rows), failures not cached (the retry test, which fails if a failure is written to the cache).

**[nit] Reversibility — reworded.** Now "Runtime-reversible; telemetry history remains", spelling out that a revert restores the runtime behaviour exactly but does not unsend the Faro events and logs already ingested — those sit until retention expires. No other section of the description changed.

**[question] follow-up issue — filed and linked.** #1591 covers the malformed-`guides` case: a 200 with `capability.available: true` and a missing or non-array `guides` field, cached as a successful empty catalogue with no signal, operationally indistinguishable from "no guides authored". The "Out of scope" section now links it.

Verification on the pushed commit, Node 24.18.0: full frontend suite `npm run test:ci` — **417 suites passed, 6649 passed / 18 skipped / 5 todo, 6672 total** (up 12 from 6637, matching the 14→26 growth in this file). `npm run typecheck` clean; `eslint` and `prettier --check` clean on both changed files. Note the "How to verify" section of the description still quotes the pre-review counts (12/12, 172) — the numbers above supersede it.

moxious (approved):

PR #1588, opened 2026-08-11
fix(learning-paths): clear the path's own completion keys on reset

### LINKED ISSUE #1560: resetPath never clears path-level backend-guide:<pathId> completion keys for App Platform paths (#1408 follow-up)
Follow-up from the second-round review of PR #1408 (https://github.com/grafana/grafana-pathfinder-app/pull/1408).

## Severity

**Medium**, `irreversible_without_cleanup` — the leaked rows persist in synced user storage and the feature's own reset affordance can never clear them. Confirmed by 3 adversarial skeptics; one ran the real storage modules in a scratch Jest harness (write under the path key, run the reset branch's exact clearing calls, both milestone slugs and the 100% journey value survive).

## Analysis

`resetPath`'s no-url branch — the branch App Platform paths take, since they are synthesized without a `url` (`src/learning-paths/app-platform-paths.ts:59-77`) — clears per-**member** state only:

- `src/learning-paths/learning-paths.hook.ts:445-467` iterates `path.guides` (member ids), clearing `bundled:<memberId>` / `backend-guide:<memberId>` content keys and bare-id `completedGuides`. It never calls `milestoneCompletionStorage.clear` at all, and clears `journeyCompletionStorage` only for member keys.

But the path-level state is written under the path's **own** base key `backend-guide:<pathId>`:

- `markMilestoneDone` writes the milestone slug set via `milestoneCompletionStorage.markCompleted(journeyBaseUrl, slug)` (`src/docs-retrieval/learning-journey-helpers.ts:506`), where `journeyBaseUrl = learningJourney.baseUrl` = the path manifest's resolution `backend-guide:<pathId>` (`src/docs-retrieval/content-fetcher/package-content.ts:274-283`, `src/package-engine/app-platform-resolver.ts:143`).
- `docs-panel.tsx:555-561` writes the journey percentage under the same key (the `bundled:`-only check at `learning-journey-helpers.ts:415` gates only the `markGuideCompleted` bridge, not the storage set at `:412`).
- The url-based branch for bundled paths DOES clear path-level state (`learning-paths.hook.ts:413-414`), so the asymmetry is specific to the new branch.

Consequences after a user hits Reset on a completed App Platform course card (Reset renders when `isCompleted` — `LearningPathCard.tsx:138`): the milestone slug set and journey percentage survive; the all-milestones threshold (`learning-journey-helpers.ts:529-531`) stays satisfied, so after a reload (the completion recorder's dedup is an in-memory, intra-session Set — `completion-recorder.ts:35-42`) completing any single member immediately re-emits a whole-journey completion. Skeptics trimmed two consequence claims: `LearningPathTableOfContents` is not production-rendered, and App Platform paths don't surface as `type: 'package'` recommendation cards — so the stale keys' UI blast radius is narrower than first reported; the storage leak and journey-trigger re-fire stand.

## Implementation

1. In the no-url branch of `resetPath` (`src/learning-paths/learning-paths.hook.ts:445-467`), also clear the path's own base keys: `milestoneCompletionStorage.clear('backend-guide:' + path.id)` and the `journeyCompletionStorage` entry for the same key (add `bundled:' + path.id` for symmetry with the url-based branch), and evict that content cache entry alongside the member evictions already there.
2. Add a test: using the real storage modules (see the harness in `learning-journey-helpers.completion-boundary.test.ts`), write a milestone slug + journey percentage under `backend-guide:<pathId>`, run `resetPath`, assert both are gone and that a subsequent single `markMilestoneDone` does not satisfy the journey threshold.

## Links

- Root PR: https://github.com/grafana/grafana-pathfinder-app/pull/1408 (the no-url branch was extended for App Platform paths in commit `d79b019d`)
- Related umbrella: https://github.com/grafana/grafana-pathfinder-app/issues/1423

### LINKED ISSUE #1590: Course reset loses most of its own deletions: concurrent clear() calls race on one shared record
## Symptom

Resetting a learning path clears far less than it claims. Member guides keep their completion percentages, so a "reset" course still shows progress against its guides.

Only **one** deletion survives per store, per reset — whichever write happens to land last. A path with ten members drops nine of its ten intended deletions.

## Mechanism

Every entry for a store lives under a **single** storage key. `clear(key)` is therefore a read-modify-write of the whole record:

`src/lib/storage/bounded-record-storage.ts:81-90`

```ts
async clear(key: string): Promise<void> {
  try {
    const storage = createStorage();
    const data = (await storage.getItem<Record<string, number>>(storageKey)) || {};
    delete data[key];
    await storage.setItem(storageKey, data);
  } catch (error) {
    logger.warn(`Failed to clear ${label}`, { error });
  }
},
```

Fire N of those concurrently and they all read the same snapshot, each deletes its own key from its own copy, and each writes the whole thing back. Last write wins; the other N-1 deletions are silently reverted.

`resetPath` in `src/learning-paths/learning-paths.hook.ts` does exactly that, in both branches:

- **`:433-440`** (paths with a `url`) — one `Promise.all` fans `interactiveCompletionStorage.clear(key)` across every milestone key, and `journeyCompletionStorage.clear(key)` across every journey key.
- **`:449-457`** (paths without a `url` — all App Platform and bundled courses) — one `Promise.all` fans `interactiveCompletionStorage.clear` and `journeyCompletionStorage.clear` across `guides.length × 2` content keys.

Confirmed empirically: seed three keys, clear them concurrently, two survive. Clear them sequentially and none do.

For contrast, `handleResetProgress` in `src/components/LearningPaths/MyLearningTab.tsx` already loops sequentially and is correct.

## Why no test catches it

`resetPath` has no coverage that uses real storage — the existing `learning-paths.hook.test.ts` mocks the storage module wholesale, so every `clear()` is a `jest.fn()` that records a call and races with nothing. The assertions check *that clear was called*, which is true, rather than *that the data is gone*, which is not.

## Suggested fix

Await the clears sequentially, or give the store a batch API (`clearMany(keys: string[])`) that does one read-modify-write for the whole set — preferable, since it is also N-times fewer round trips.

`interactiveStepStorage.clearAllForContent` is a different store and may be unaffected; check before folding it into any batch.

Needs a test using the real storage modules, asserting the resulting record is empty rather than asserting call counts.

## Provenance

Not related to the completion-records work — `gh pr diff --name-only` on #1433 and #1434 shows neither touches this file. The concurrent-clear shape traces to #1408 (`bd5dae6b`), with earlier layers from #909 and #587.

Adjacent but distinct from #1560 / PR #1588, which fixes a *different* defect in the same function (the path's own keys never being cleared at all) and deliberately scoped this out. #1588's own new clears are sequential and correct.

Found during review of PR #1588.

tomglenn:
## Summary

`resetPath` never cleared the path's own completion keys for paths that carry no `url` — the branch App Platform courses always take. Resetting a completed course cleared the visible state but left the "all milestones done" checklist behind, so the next single guide completion re-crossed the whole-course threshold and the app re-declared the course complete. Nothing in the product could clear that record.

Review then surfaced two more defects in the same reset flow, both reproduced: member completion records were being lost to a read-modify-write race, and a path's own key can hold interactive progress that reset ignored. Both are fixed here.

This clears the path-level milestone checklist, journey percentage and interactive progress; batches every bounded-record delete so none are lost; and adds a backfill so users already in the bad state can get out.

Fixes #1560
Closes #1590

## What to look at

- **`src/lib/storage/bounded-record-storage.ts` — new `clearMany(keys)`.** Every entry in one of these stores lives under a single storage key, so `clear(key)` is a whole-record read-delete-write. Firing N of those concurrently means they all read the same snapshot and the last write restores what its siblings deleted. `clearMany` does one read, N deletes, one write, with the same error handling as `clear`.
- **`src/learning-paths/learning-paths.hook.ts`** — both branches of `resetPath` now route their bounded-record deletes through `clearMany`:
  - the no-`url` branch clears the path's own key under both `backend-guide:<pathId>` and `bundled:<pathId>` — milestone checklist, journey percentage, interactive completion and interactive step state — alongside the member keys it already handled;
  - the `url` branch collects its milestone and journey keys first and issues one batched delete per record, replacing the per-key concurrent clears that had the same race.
- `src/lib/user-storage.ts` — new `milestoneCompletionStorage.clearAll()`, following the existing `clearAll` pattern on the other storage helpers.
- `src/components/LearningPaths/MyLearningTab.tsx` — "Reset all learning progress" now calls it. That surface previously never touched milestone records at all.
- `src/learning-paths/reset-path-completion.test.ts` and `src/lib/storage/bounded-record-storage.test.ts` — new, and deliberately using the **real** storage modules.

Scope note: the two branches are still not unified. Both now share `clearMany`, but their key-discovery differs — the `url` branch recovers milestone keys by prefix scan, the other derives them from `path.guides` — and collapsing that is a refactor this PR isn't.

## Why

App Platform paths are built from the backend catalogue with no `url` (`src/learning-paths/app-platform-paths.ts`), so they always fall into the second branch of `resetPath`. Path-level data is nonetheless written under `backend-guide:<pathId>` — by the resolver's `contentUrl`, by `markMilestoneDone`, and by the journey percentage write in `docs-panel.tsx`. The branch only ever iterated `path.guides`, so it cleared members and left the cover key untouched, and it never called `milestoneCompletionStorage` at all.

Two corrections from review, both worth recording because the original scope cuts were wrong:

- **Interactive state under the path key is not hypothetical.** `CustomGuidesSection` opens a path directly as `backend-guide:<pathId>`, and `AppPlatformPackageResolver` renders that resource's `spec.blocks` under the same content key. Nothing in the schema forbids interactive blocks on a path cover, and an interactive step seeded at 100% under that key survived a reset. If interactive path covers are meant to be impossible, that belongs in schema validation rather than being assumed here.
- **The member-clear race was real and reproducible.** Seeding six member journey keys for a three-guide path and running `resetPath` left five behind, because the surrounding `Promise.all` issued one whole-record rewrite per key. The same shape existed in the `url` branch, tracked as #1590 and fixed here rather than deferred — shipping a reset that clears its cover key but still loses most member deletions would have been worse than either state alone.

`resetPath` had zero direct coverage. The existing `learning-paths.hook.test.ts` mocks the whole storage module — including `milestoneCompletionStorage: { clear: jest.fn() }` — so it structurally cannot observe which keys a reset touches, and a mocked `clear` races with nothing. The new files are modelled on `src/lib/user-storage.test.ts` and assert against real storage.

The backfill exists because the bad record is already on disk for affected users and no product button reached it.

## How to verify

```
npm run test:ci
```

Seven `resetPath` cases plus the `clearMany` unit tests. Each was checked by breaking the behaviour it covers and confirming failure:

- reverting the hook change fails the four cover-key cases;
- restoring the old `Promise.all(map(clear))` in either branch fails the two "in one pass, without restoring siblings" cases — the race reproduces on demand;
- excluding cover keys from the interactive clears fails the cover-progress case in isolation.

Each race test seeds an unrelated sentinel key and asserts it survives, so a fix that over-deletes fails too.

Manually: complete every guide in an App Platform course, reset it from My Learning, then complete one guide. The course should read as 1-of-N, not complete.

## Breaking changes

None. Reset now deletes additional storage keys for the path's own cover — milestone, journey and interactive state — and the global reset deletes milestone records it previously left behind. Both are the intended semantics of "reset". `clearMany` is additive; no existing caller changes behaviour.

## Reversibility

Fully reversible; revert the commit. No schema, no migration, no server-side state. The only irreversible effect is per-user and intended: records deleted by a reset are gone, which is what the button says it does.

## Ordering

This should land before #1433/#1434. Today the surviving record is local only; once those land, the same stale threshold gets written to the server as a durable completion record.

Also for #1434, not handled here: a reset does not currently discard writes already sitting in the durable completion-write queue. `src/completion-records/` is untouched by this PR.

### DIFF
--- a/src/components/LearningPaths/MyLearningTab.test.tsx
+++ b/src/components/LearningPaths/MyLearningTab.test.tsx
@@ -16,6 +16,7 @@ import { render, screen, waitFor, act, fireEvent } from '@testing-library/react'
 import { MyLearningTab } from './MyLearningTab';
 import { prepareGuideLaunch, type PrepareGuideLaunchResult } from '../docs-panel/utils/prepare-guide-launch';
 import { testIds } from '../../constants/testIds';
+import { milestoneCompletionStorage } from '../../lib/user-storage';
 
 jest.mock('../docs-panel/utils/prepare-guide-launch', () => ({
   prepareGuideLaunch: jest.fn(),
@@ -92,6 +93,7 @@ jest.mock('../../lib/user-storage', () => ({
   journeyCompletionStorage: { getAll: jest.fn(async () => ({})), clear: jest.fn() },
   interactiveStepStorage: { clearAll: jest.fn() },
   interactiveCompletionStorage: { clearAll: jest.fn() },
+  milestoneCompletionStorage: { clearAll: jest.fn() },
 }));
 jest.mock('../../global-state/completion-store', () => ({ evictAllContentCaches: jest.fn() }));
 
@@ -495,5 +497,28 @@ describe('MyLearningTab — App Platform guide launch', () => {
       source: 'home_page',
       packageInfo: undefined,
     });
+  });
+});
+
+describe('MyLearningTab — reset all learning progress', () => {
+  it('drops milestone checklists so a single later completion cannot re-complete a course', async () => {
+    const confirmSpy = jest.spyOn(window, 'confirm').mockReturnValue(true);
+
+    render(<MyLearningTab onOpenGuide={jest.fn()} />);
+    fireEvent.click(screen.getByTestId(testIds.learningPaths.resetProgressButton));
+
+    await waitFor(() => expect(milestoneCompletionStorage.clearAll).toHaveBeenCalledTimes(1));
+    confirmSpy.mockRestore();
+  });
+
+  it('leaves milestone checklists alone when the confirmation is declined', async () => {
+    const confirmSpy = jest.spyOn(window, 'confirm').mockReturnValue(false);
+
+    render(<MyLearningTab onOpenGuide={jest.fn()} />);
+    fireEvent.click(screen.getByTestId(testIds.learningPaths.resetProgressButton));
+
+    await waitFor(() => expect(confirmSpy).toHaveBeenCalled());
+    expect(milestoneCompletionStorage.clearAll).not.toHaveBeenCalled();
+    confirmSpy.mockRestore();
   });
 });
--- a/src/components/LearningPaths/MyLearningTab.tsx
+++ b/src/components/LearningPaths/MyLearningTab.tsx
@@ -25,6 +25,7 @@ import {
   journeyCompletionStorage,
   interactiveStepStorage,
   interactiveCompletionStorage,
+  milestoneCompletionStorage,
 } from '../../lib/user-storage';
 import { evictAllContentCaches } from '../../global-state/completion-store';
 import type { EarnedBadge } from '../../types';
@@ -233,6 +234,11 @@ export function MyLearningTab({ onOpenGuide }: MyLearningTabProps) {
       for (const url of Object.keys(completions)) {
         await journeyCompletionStorage.clear(url);
       }
+
+      // Milestone checklists outlive a per-path reset for paths that predate the
+      // path-key fix, so a global reset has to drop them too — otherwise the next
+      // single completion re-crosses the all-milestones threshold.
+      await milestoneCompletionStorage.clearAll();
 
       // Clear all interactive guide step and completion state
       // This prevents guides from instantly re-completing when reopened
--- a/src/learning-paths/learning-paths.hook.ts
+++ b/src/learning-paths/learning-paths.hook.ts
@@ -409,81 +409,57 @@ export function useLearningPaths(): UseLearningPathsReturn {
       }
 
       if (path.url) {
-        // URL-based path: clear milestone tracking and journey completion
         await milestoneCompletionStorage.clear(path.url);
-        await journeyCompletionStorage.clear(path.url);
 
-        // Remove milestone slugs from completedGuides (path.guides contains fetched slugs)
         if (path.guides.length > 0) {
           await learningProgressStorage.removeCompletedGuides(path.guides);
         }
 
-        // Clear interactive steps for milestone URLs (iterate over completed milestones)
-        // Since we don't have the full milestone URLs stored, clear by prefix pattern
-        // The content keys for milestones start with the path URL
         const [completions, journeyCompletions] = await Promise.all([
           interactiveCompletionStorage.getAll(),
           journeyCompletionStorage.getAll(),
         ]);
 
+        // Milestone content keys aren't stored anywhere, so recover them by prefix.
         const normalizedUrl = path.url.replace(/\/+$/, '');
-
-        // Batch clear operations for better performance with many milestones
         const milestoneKeys = Object.keys(completions).filter((key) => key.startsWith(normalizedUrl));
-        await Promise.all([
-          ...milestoneKeys.map((key) =>
-            Promise.all([interactiveCompletionStorage.clear(key), interactiveStepStorage.clearAllForContent(key)])
-          ),
-          ...Object.keys(journeyCompletions)
-            .filter((key) => key.startsWith(normalizedUrl))
-            .map((key) => journeyCompletionStorage.clear(key)),
-        ]);
-        // Drop the completion store's in-memory cache for each milestone
-        // we cleared. Without this, any open guide in the path would
-        // still surface its prior completion snapshot via the store.
+        const journeyKeys = [path.url, ...Object.keys(journeyCompletions).filter((k) => k.startsWith(normalizedUrl))];
+
+        await Promise.all(milestoneKeys.map((key) => interactiveStepStorage.clearAllForContent(key)));
+        await interactiveCompletionStorage.clearMany(milestoneKeys);
+        await journeyCompletionStorage.clearMany(journeyKeys);
+
         milestoneKeys.forEach((key) => evictContentCache(key));
       } else {
         // No base URL: either a static bundled path (`bundled:<id>`) or an App
         // Platform path whose members are `backend-guide:<id>`. We can't tell
         // them apart from `path.guides` alone, so clear both content schemes.
+        const pathKeys = [`bundled:${path.id}`, `backend-guide:${path.id}`];
+        const contentKeys = [
+          ...pathKeys,
+          ...path.guides.flatMap((guideId) => [`bundled:${guideId}`, `backend-guide:${guideId}`]),
+        ];
 
-        // The path's OWN key, not just its members: milestone checklists and the
-        // journey percentage are stored under the cover key. Sequential because
-        // both helpers read-modify-write one shared record.
-        for (const pathKey of [`bundled:${path.id}`, `backend-guide:${path.id}`]) {
+        for (const pathKey of pathKeys) {
           await milestoneCompletionStorage.clear(pathKey);
-          await journeyCompletionStorage.clear(pathKey);
         }
 
-        await Promise.all(
-          path.guides.flatMap((guideId) =>
-            [`bundled:${guideId}`, `backend-guide:${guideId}`].map((contentKey) =>
-              Promise.all([
-                interactiveStepStorage.clearAllForContent(contentKey),
-                interactiveCompletionStorage.clear(contentKey),
-                journeyCompletionStorage.clear(contentKey),
-              ])
-            )
-          )
-        );
-        path.guides.forEach((guideId) => {
-          evictContentCache(`bundled:${guideId}`);
-          evictContentCache(`backend-guide:${guideId}`);
-        });
+        await Promise.all(contentKeys.map((key) => interactiveStepStorage.clearAllForContent(key)));
+        // Batched, not one clear per key: each helper read-modify-writes a single shared record.
+        await interactiveCompletionStorage.clearMany(contentKeys);
+        await journeyCompletionStorage.clearMany(contentKeys);
 
-        // Remove guide IDs from completedGuides (bare ids, matching how App
-        // Platform + bundled completions are keyed).
+        contentKeys.forEach((key) => evictContentCache(key));
+
         await learningProgressStorage.removeCompletedGuides(path.guides);
       }
 
-      // Dispatch event to notify UI components to refresh
       window.dispatchEvent(
         new CustomEvent(StorageEvents.InteractiveProgressCleared, {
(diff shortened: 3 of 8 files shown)


moxious (changes requested):
The cover-key fix itself is correct, but the same reset branch still has a reproducible read-modify-write race for member completion records, and the declared scope cut for path-cover interactive state doesn't hold for directly openable App Platform paths.

**[blocking]** `src/learning-paths/learning-paths.hook.ts:458` — These member clears still race on the shared records. `journeyCompletionStorage.clear()` and `interactiveCompletionStorage.clear()` each perform a whole-record get/delete/set (`bounded-record-storage.ts:81-86`), but the surrounding `Promise.all` starts one call per bundled/backend member key. Every call can read the same snapshot, after which the last write restores the keys deleted by its siblings. I reproduced this with the real storage modules: seeding six member journey keys for a three-guide path and calling `resetPath` left five of the six entries behind. Please serialize calls within each bounded storage helper, or add an atomic `clearMany(keys)` that reads and writes each record once. The regression test should seed all bundled/backend member keys plus an unrelated sentinel, then assert every member key is removed and the sentinel remains. (The new cover-key loop itself is safely sequential — the dangerous concurrency is in the unchanged member loop immediately below it.)

**[blocking]** `src/learning-paths/learning-paths.hook.ts:453` — The path's own key can also hold interactive progress, so clearing only milestone and journey state leaves Restart incomplete. `CustomGuidesSection` directly opens a path as `backend-guide:<pathId>`, and `AppPlatformPackageResolver` renders that path resource's `spec.blocks` under the same content key — no schema rule forbids interactive blocks on a path cover. I seeded an interactive step at 100% completion under `backend-guide:<pathId>` and confirmed it survived `resetPath`. Please also clear `interactiveStepStorage.clearAllForContent(pathKey)` and `interactiveCompletionStorage.clear(pathKey)` (and evict `pathKey` from the completion cache) for both path-key schemes, with a real-storage cover-progress test. If interactive path covers are meant to be impossible, that should be enforced in schema validation rather than assumed.

**[nit]** `src/learning-paths/reset-path-completion.test.ts:1` — The file header preserves ticket/review history that the test names and assertions already express (QC8.3). Please trim it, along with the similar narrating comments at `learning-paths.hook.ts:412`, `:479`, and `:486`. A short note on the actual same-helper read-modify-write invariant is worth keeping once the race above is fixed.

Once the member-record race and path-cover cleanup are addressed with real-storage tests, this and the global backfill should be safe to merge — ahead of #1433/#1434, as already noted in the PR's own Ordering section.

tomglenn:
Thanks @moxious — both blockers reproduced as described, and both are fixed in c93f055. Point by point:

**Blocker 1 — member clears race on a shared record (`learning-paths.hook.ts:458`)**

Fixed with the atomic option rather than serialization. `createBoundedRecordStorage` now exposes `clearMany(keys)`, which does one read, N deletes, one write, with the same try/catch + `logger.warn` shape as `clear`. Both `resetPath` branches route through it — the `url` branch at ~:433-440 had exactly the same race across milestone and journey keys, so **this closes #1590 as well**.

Evidence, against the real storage modules:

- New test `clears every member key in one pass, without restoring siblings, and spares unrelated content` seeds all six member keys (`bundled:` + `backend-guide:` for three guides) in both `journeyCompletionStorage` and `interactiveCompletionStorage`, plus a sentinel `backend-guide:unrelated-guide`. After `resetPath`: zero member keys remain in either record, sentinel still 100.
- Discrimination: reverting the else branch to `Promise.all(contentKeys.map((key) => …clear(key)))` fails 4 of the 6 tests in that file, including the new one — the surviving-siblings race reproduces in jsdom exactly as you saw.
- New test `clears every milestone key in one pass … and spares other journeys` covers the `url` branch: three milestone URLs under the path prefix plus an unrelated journey key. Discrimination: restoring the old concurrent per-key clears leaves all three milestone journey keys behind (`https://…/collect-logs/`, `/define-rules/`, `/route-alerts/`), which is #1590 reproduced and then fixed.

**Blocker 2 — path cover can hold interactive progress (`learning-paths.hook.ts:453`)**

Agreed, the scope cut doesn't hold. Both path-key schemes (`bundled:<pathId>` and `backend-guide:<pathId>`) are now folded into the same `contentKeys` list as the members, so each gets `interactiveStepStorage.clearAllForContent`, `interactiveCompletionStorage` clearing (via `clearMany`, so no new race), and `evictContentCache`.

Evidence: new test `clears interactive progress recorded against the path cover itself` seeds two completed steps plus a 100% interactive completion under both cover keys and warms the count cache. After reset, step storage is empty, completion is 0, `countAllCompleted` is 0, and `evictContentCache` was called with each cover key. Discrimination: excluding just the cover keys from the interactive clears (leaving milestone/journey handling intact) fails that test alone.

Ordering note: the new cover clears go through `clearMany` alongside the members, so they can't reintroduce the race; the two `milestoneCompletionStorage.clear` calls stay sequential.

**Nit — comments**

Dropped the `reset-path-completion.test.ts` file header and the narrating comments at `:412`, `:479`, `:486`, plus a couple of adjacent ones in the same function under trim-on-touch. Kept one note on the invariant, as the doc line on `clearMany` in `BoundedRecordStorage` ("Concurrent `clear` calls share one record, so each write restores the keys its siblings deleted") and one short line at the call site pointing at it.

**Verification**

Full suite green: 418 suites, 6644 passed / 18 skipped / 5 todo. `npm run typecheck`, `eslint`, and `prettier --check` all clean on the changed files. Every new test was verified to fail when its behaviour is broken and pass when restored; the 7 pre-existing tests from this PR were re-checked the same way and still discriminate.

tomglenn:
Description updated — no code changes. All three points addressed:

**Stale scope note.** Removed. The body now states that interactive state under the path key *is* cleared, for both key schemes, and records why the original scope cut was wrong: `CustomGuidesSection` opens a path directly as `backend-guide:<pathId>` and the resolver renders that resource's `spec.blocks` under the same key, so an interactive step seeded there survives a reset. Your point about enforcing it in schema validation rather than assuming it is carried over verbatim.

**Sequential-per-key explanation.** Removed, and replaced with the batch API. "What to look at" now leads with `clearMany(keys)` on `bounded-record-storage.ts` — one read, N deletes, one write — and describes both branches routing through it, including that the `url` branch collects its milestone and journey keys first. The "How to verify" section now lists the actual per-test discrimination results, including that restoring the old `Promise.all(map(clear))` in either branch reproduces the race on demand, and that each race test seeds a sentinel so an over-deleting fix fails too.

**#1590 won't close.** Fixed — the body now carries `Closes #1590` alongside `Fixes #1560`. GitHub confirms both are linked: `closingIssuesReferences` returns `[1560, 1590]`.

Also corrected two things you didn't flag, since the description is the merge-time record: the "two branches are not unified" note now says they share `clearMany` but keep separate key-discovery (prefix scan vs `path.guides`), and Breaking changes now lists the full set of keys a reset deletes rather than the pre-revision two.

Ready for another look.

moxious (approved):

Issue #1566, opened 2026-08-10
Test-coverage gaps from #1408 second-round review (additive tests only)


moxious:
Follow-up from the second-round review of PR #1408 (https://github.com/grafana/grafana-pathfinder-app/pull/1408). Additive test-only gaps — no production code changes. (The locked-milestone fixture gaps and the repository-normalization suppression tests are tracked with their owning fixes, not here.)

## Severity

**Low** — each guards an already-correct behavior whose silent regression would be invisible to the current suite.

## Items

1. **Completion-records `obo-unavailable` branch: 0% coverage.** `pkg/plugin/completion_records.go:578-580` (the nil-exchanger gate in `resolveCompletionBackend`) has zero coverage — verified with `go tool cover`. Its custom-guide twin is tested (`TestCustomGuide_NoOBOCredentialsStructurallyUnavailable`, `pkg/plugin/custom_guide_repository_test.go:378`). Clone that test's shape for the completions route: unprovisioned exchanger → capability=false with reason `obo-unavailable`, no upstream call.

2. **"Cheap declines stay untagged" is never asserted against the real resolver.** The composite's caching decision keys entirely on `PackageResolutionFailure.repository`, and the two pre-attempt decline paths (no namespace `src/package-engine/app-platform-resolver.ts:129`, GAP toggle off `:136`) must stay untagged to preserve static-tier negative caching. The decline tests (`app-platform-resolver.test.ts:55-75`) assert `ok: false` but not the absence of `repository`; the composite-side test mocks the resolver. Add `expect(result.repository).toBeUndefined()` to both decline tests.

3. **`buildManifest` repository-forcing has no wrong-value test.** `app-platform-resolver.ts:100-104` forces `repository` last because it is "the sole input to the durable completion key" — but no test feeds a persisted manifest carrying a _wrong_ `repository` (e.g. `'interactive-tutorials'`) and asserts the forced value wins. The id-forcing half has exactly this test (`app-platform-resolver.test.ts:194`); mirror it.

[...]

PR #1589, opened 2026-08-11
test: close the App Platform coverage gaps from the #1408 second-round review

tomglenn:
## Summary

Additive, test-only coverage for the App Platform gaps flagged in the #1408 second-round review (#1566). No production code changes.

- **Completion-records read path.** `resolveCompletionBackend`'s nil-exchanger gate (`pkg/plugin/completion_records.go`) reported `obo-unavailable` with nothing asserting it. New `TestMyCompletions_NoOBOCredentialsStructurallyUnavailable` mirrors the custom-guide twin. Distinct from the write resolver (`resolveCompletionWriteBackend`) covered by the open PR #1433, so expect a trivial rebase.
- **Cheap declines stay untagged.** The two pre-network declines in `app-platform-resolver.ts` now assert `result.repository` is absent. That absence is load-bearing: `composite-resolver.ts` uses `result.repository` as the sole trigger for cache eviction, so tagging a cheap decline would repeal negative caching for every tier.
- **`buildManifest` repository forcing.** Every existing case already passed the correct `'app-platform'`, so deleting the force broke nothing. Added a case feeding `'interactive-tutorials'` (the schema's public-CDN default) and asserting the forced value wins.
- **Go group-version pin.** `TestAggregationToggleMatchesGroup` derives the expected toggle *from* `customGuideGroupVersion`, so editing the Go group would move both sides together and stay green while TypeScript kept pinning its own literal. One literal assertion closes that.
- **Completion write to read round-trip.** New suite exercising the real localStorage-backed storage layer: completing the members of a synthesized App Platform path through `markMilestoneDone` must read back through `getPathProgress`.

## What to look at

- `src/learning-paths/learning-paths.completion-round-trip.test.ts` — the only new file. It mocks `@grafana/runtime` and nothing else, so the catalogue fetch, path synthesis, badge coordinator and progress storage all run for real. The existing `learning-paths.hook.test.ts` is untouched: it documents itself as a deliberately narrow App-Platform-ingestion suite and stubs storage away on purpose.
- The `beforeEach` in that file seeds an explicit empty progress record rather than relying on an empty `localStorage`. On a storage miss `learningProgressStorage.get()` hands back a shared module-level `DEFAULT_PROGRESS` object that `badge-coordinator.markGuideCompleted` mutates in place via `push`, so completions leak between cases. Benign in production (the miss happens once per session) but worth a look — it is a latent shared-mutable-default, and out of scope for a test-only PR.

## Why

Each of these guards an already-correct behaviour whose silent regression would be invisible to the current suite. The round-trip is the highest value: the write side (`getMilestoneSlug` stripping the `backend-guide:` scheme) and the read side (`calculatePathProgress` matching bare ids in `LearningPath.guides`) are only coupled by convention, and every test on either side mocked the other away. If they drift, My Learning silently reports 0% for private paths.

## How to verify

```
go test ./pkg/...
npx jest src/package-engine/app-platform-resolver.test.ts src/learning-paths/learning-paths.completion-round-trip.test.ts --coverage=false
npm run typecheck && npm run lint:go
```

Every new assertion was checked to discriminate — the covered behaviour was removed, the failure observed, and the code restored:

| Test | Break applied | Result |
| --- | --- | --- |
| `TestMyCompletions_NoOBOCredentialsStructurallyUnavailable` | removed the `a.oboExchanger == nil` gate | fails (nil-pointer panic on the doomed upstream call) |
| `TestAggregationToggleMatchesGroup` (new pin) | `customGuideGroupVersion` → `.../v1beta1` | fails on the pin only; the derived-toggle cases stay green, which is exactly the drift the pin exists for |
| both untagged-decline assertions | made `decline()` tag `repository` | 2 failures |
| repository-forcing case | dropped `repository` from the forced overwrite | fails, received `interactive-tutorials` |
| round-trip suite | removed the `backend-guide:` strip in `getMilestoneSlug` | 3 of 4 fail |
| round-trip suite | broke `calculatePathProgress`'s completed count | the 100% case fails |

Full suite: 418 suites / 6638 tests passing.

## Breaking changes

None. Tests only.

## Reversibility

Fully reversible — revert the commit. Nothing outside test files changes.

## Note on the issue

Refs #1566, deliberately not `Fixes`. **Item 4 was rescoped.** The issue asks for two toggle-name tests that already exist: the Go comment naming the TypeScript counterpart (`pkg/plugin/app_platform_client_test.go`, present since #1408) and the TypeScript literal assertion (`src/utils/interactive-guides-api.test.ts`, from #1430). Neither was re-added. The one real gap — the Go side deriving its expectation from its own constant — is closed here. Please review the issue before closing it.

### DIFF
--- a/pkg/plugin/app_platform_client_test.go
+++ b/pkg/plugin/app_platform_client_test.go
@@ -230,4 +230,12 @@ func TestAggregationToggleMatchesGroup(t *testing.T) {
 	if customGuideAggregationToggle == pathfinderBackendAggregationToggle {
 		t.Error("custom-guide and completion-records must gate on distinct aggregation toggles")
 	}
+
+	// The cases above derive the expected toggle FROM the Go group constants, so
+	// editing a group would move both sides together and stay green while the TS
+	// literal (APP_PLATFORM_API_VERSION, src/utils/interactive-guides-api.ts)
+	// kept pinning the old group. Pin the Go side to the same literal.
+	if customGuideGroupVersion != "pathfinderbackend.ext.grafana.app/v1alpha1" {
+		t.Errorf("customGuideGroupVersion = %q, but src/utils/interactive-guides-api.ts pins pathfinderbackend.ext.grafana.app/v1alpha1", customGuideGroupVersion)
+	}
 }
--- a/pkg/plugin/completion_records_test.go
+++ b/pkg/plugin/completion_records_test.go
@@ -712,6 +712,27 @@ func TestMyCompletions_NoAppURLStructurallyUnavailable(t *testing.T) {
 	}
 }
 
+// An unprovisioned stack has no way to authenticate as the caller upstream, so
+// the read route reports itself unavailable rather than attempting a call that
+// can only 401. No lister override here: that override short-circuits ahead of
+// the exchanger gate, so this is the real resolve path.
+func TestMyCompletions_NoOBOCredentialsStructurallyUnavailable(t *testing.T) {
+	withFrozenTime(t, time.Unix(1_700_000_000, 0))
+	resetCompletionRecordsCache()
+	t.Cleanup(resetCompletionRecordsCache)
+
+	rr, body := doMyCompletionsReq(t, completionRequest(t, "/completion-records/my", "user:1"))
+	if rr.Code != http.StatusOK {
+		t.Fatalf("status = %d, want 200", rr.Code)
+	}
+	if body.Capability.Available || body.Capability.Reason != reasonOBOUnavailable {
+		t.Fatalf("expected capability=false %q without provisioned credentials, got %+v", reasonOBOUnavailable, body.Capability)
+	}
+	if len(body.Completions) != 0 {
+		t.Errorf("expected no completions, got %d", len(body.Completions))
+	}
+}
+
 // --- Capability probe --------------------------------------------------------
 
 func doCapability(t *testing.T, sub string) (*httptest.ResponseRecorder, completionCapability) {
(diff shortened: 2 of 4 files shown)


moxious (approved):
Verified this is genuinely test-only (four test files, 166 additions, zero deletions, no production changes) and checked all five coverage gaps from #1566 individually — all five are real and closed, including confirming the two "toggle-name" tests the issue asked for already exist elsewhere so re-adding them would've been redundant. Issue #1566 can close once this merges.

One non-blocking item worth tracking separately, not a issue with this PR:

**[suggestion]** `src/learning-paths/learning-paths.completion-round-trip.test.ts:71` — This seed correctly isolates the suite, but it also documents a reproducible production reset bug: `learningProgressStorage.get()` returns the shared `DEFAULT_PROGRESS` object, `markGuideCompleted()` mutates it in place, and `clear()` only removes the persisted key. After a first-session completion, **Reset progress** can therefore read the mutated singleton and resurrect the old completion — I reproduced this directly with a temporary regression test. Could you file/link a follow-up before merging? The runtime fix can stay out of this test-only PR; a fresh-default factory plus a complete → clear → read regression would close it.

**[question]** `pkg/plugin/app_platform_client_test.go:204-240` — the "expect a trivial rebase" note against #1433 undersells the overlap a bit: #1433 edits this exact test and changes the toggle invariant from distinct `.com`/`.app` toggles to a shared `.app` toggle, plus adds its own literal routing pins. Worth confirming landing order and whether #1433 should retain this assertion or deduplicate against its own routing test.

Approving — solid additive coverage, nothing here blocks merge.

PR #1593, opened 2026-08-11
feat(my-learning): split private paths into their own section

Jayclifford345:
## Summary

App Platform paths — the org's own guides from the namespace custom-guide catalogue — were merged into the same flat `paths` list as the curated bundled/CDN ones ([`learning-paths.hook.ts:174`](https://github.com/grafana/grafana-pathfinder-app/blob/main/src/learning-paths/learning-paths.hook.ts#L174)), and `MyLearningTab` split that list purely by progress. A private org path therefore landed in **My paths** indistinguishable from a Grafana-authored one.

This adds a **Private paths** container above the My paths ∥ Badges row, shown only when the namespace actually has one.

## What to look at

- `src/types/learning-paths.types.ts` — new `isPrivate?: boolean` on `LearningPath`. Previously the only signal was `manifest` presence, which is typed `Record<string, unknown>`, so the UI tier would have needed a cast to read provenance.
- `src/learning-paths/app-platform-paths.ts` — the adapter stamps `isPrivate: true`. Every entry it synthesises is namespace-scoped by construction (the client force-stamps `repository: 'app-platform'`), so there's no condition. This is the one place where the typed `CustomGuideManifest` is still in scope.
- `src/components/LearningPaths/sections/PrivatePathsSection.tsx` — new section. Deliberately a near-copy of `MyCoursesSection` using the identical `section` / `columnSection` / `pathsGrid` style keys, so it reads as the same object. Differences: `lock` icon, and an early `return null` on an empty list instead of an empty state (precedent: `CustomGuidesSection.tsx:119`). No new entries in `MyLearningTab.styles.ts`.
- `src/components/LearningPaths/MyLearningTab.tsx` — the `courses` memo became a shared `inProgress` list, partitioned into `privatePaths` / `courses`. **`excludeTitles` now sources from `inProgress`** so Discover more still suppresses private titles — worth a look, it's the one place the split could have silently regressed something.

## Why

Orgs publishing their own paths had no way to tell them apart from Grafana's curated ones on the learning hub. Naming the distinction on the type also stops future callers from sniffing `manifest` to infer provenance.

## Design decisions

- **Placement**: full-width, above the two-column row, rather than stacked inside the left column — keeps the 7:5 grid and the 440px column cap intact.
- **Naming**: "Private paths". Note this is a second term for the same packages the recommendations tab calls "Custom guides". Deliberate — the learning hub frames them by visibility, the context panel by authorship — but flagging it so it isn't read as drift later.
- **Completed private paths** keep flowing to the shared **Completed** section, mirroring how My paths behaves. Private paths shows only not-yet-complete ones.

## How to verify

```bash
npm run check
```

Green locally (418 suites / 6641 tests, Go included), plus `npm run build` clean.

Three new cases in `MyLearningTab.test.tsx` cover the split: private path routes into Private paths and out of My paths, section absent when there are none, and a completed private path moves to Completed. The routing test was mutation-checked (dropping the `!path.isPrivate` filter fails it) so it isn't a false pass.

**Not yet verified — needs a reviewer with a real stack.** The live render against a namespace with published App Platform paths couldn't be exercised locally (no namespace / OBO token). Worth eyeballing:
1. Private paths renders above My paths with the org's path/journey cards.
2. Continue on a private card still opens the member with milestone chrome (toolbar, next/prev, back-to-cover) — `handleOpenGuide` is unchanged and still threads the path manifest as `packageInfo`, but that path is untested end-to-end here.
3. Panel below 640px — the section should grow and let the page scroll, same as My paths when the columns stack.

## Breaking changes

None. `isPrivate` is optional and only the App Platform adapter sets it; orgs with no custom guides see an unchanged page. No backend, catalogue, or launch-path changes.

## Reversibility

Fully reversible — revert the commit. No storage keys, analytics contracts, or persisted state touched.

## Contract note

Extends the My Learning App Platform ingestion established by #1408 (RFC CUSTOM-GUIDE-PACKAGES.md §6.11) **presentationally**: it adds a named provenance field and a surface, without changing what the catalogue returns or how packages launch.

## Note for reviewers

`npm run i18n-extract` turned out to be destructive in this repo — it deletes the entire `helpFooter` namespace, resets translated strings back to English across all 20 non-en-US locales, and flattens plural keys. I reverted that and added the two new keys by hand, so the locale diff here is two lines. The extractor config has drifted from the checked-in locale files; that looks worth a separate issue.


Spidy88 (approved):
LGTM

Issue #1586, opened 2026-08-11
Journey completion counter never prunes stale milestone ids (split out of #1565 item 3)


tomglenn:
Split out of #1565 item 3, which was previously believed to be folded into #1559. #1559 closed having fixed only the **denominator**; the numerator described here is untouched, so this was left tracked nowhere.

## The problem

The set of completed milestone slugs stored under `backend-guide:<pathId>` only ever grows. Nothing prunes it when a path's members change.

The completion gate compares that set's size against the *current* unlocked milestone count:

`src/docs-retrieval/learning-journey-helpers.ts:531`

```ts
if (completed.size >= totalMilestones) {
```

`#1559` corrected `totalMilestones` at all four `markMilestoneDone` call sites so the denominator is now the live unlocked count (`link-handler.hook.ts:534`, `LearningJourneyMilestoneToolbar.tsx:165`, `DocsPanelContentArea.tsx:444`, `useLastMilestoneAutoComplete.ts:56`). The numerator was not touched.

So if an author renames or removes milestones a user had partial progress on, the stale slugs still count toward a smaller live target. The journey can be declared complete — badge awarded, `journey_completed` fired — before the user has actually finished.

## Why this needs a decision, not just a patch

#1565's stated fix does **not** compile against the real signature:

```ts
milestones.filter(m => !m.isLocked && completed.has(getMilestoneSlug(m.url))).length >= threshold
```

`markMilestoneDone` (`src/docs-retrieval/learning-journey-helpers.ts:497-502`) receives only `(journeyBaseUrl, milestoneSlug, totalMilestones?, context?)`. There is no `milestones` array in scope, and `CompletionContext` (`{packageManifest?, guideTitle?, pathId?}`) does not carry one.

Options for getting the live milestone list to the counter:

1. Extend `CompletionContext` to carry the milestone list.
2. Add an explicit parameter to `markMilestoneDone`.
3. Compute the intersected count at each of the four call sites and pass it in.

(3) is the smallest diff but duplicates the logic four ways — the exact shape that caused #1559. (1) and (2) are cleaner but change a signature with four callers.

## Acceptance

- Completing a path whose milestones were renamed or removed mid-progress does not fire the completion gate early.
- A regression test that fails against the current `completed.size` comparison.

_Split out of #1565. Root: #1408._

moxious:
## ✅ This looks fixed already

[...]

Issue #1567, opened 2026-08-10
Journey percentage is never recorded as 100 for backend-guide: paths (the 100% write is bundled:-gated)


moxious:
Split out of https://github.com/grafana/grafana-pathfinder-app/issues/1565 (item 4) during triage of the https://github.com/grafana/grafana-pathfinder-app/pull/1408 follow-up set, with a corrected diagnosis. Verified at PR head `16e5bc4b`.

## Summary

The persisted journey-completion percentage is never recorded as 100 for an App Platform path, because the only 100% write is gated on the `bundled:` scheme.

`#1565` originally framed this as "`getJourneyProgress` uses a locked-inclusive display count, so a partially-published path caps below 100 while the unlocked-count gate declares it complete." That framing is wrong. The real cause is narrower and unconditional:

- `src/components/docs-panel/components/DocsPanelContentArea.tsx:430-432` — the only call to `setJourneyCompletionPercentage(baseUrl, 100, …)` is guarded by `baseUrl?.startsWith('bundled:')`.
- App Platform paths key on `backend-guide:<pathId>` (`src/package-engine/app-platform-resolver.ts:143` → `src/docs-retrieval/content-fetcher/package-content.ts:274-283`), so they never take that branch.
- The only other write is the per-milestone-load one at `src/components/docs-panel/docs-panel.tsx:555-561`, which stores `getJourneyProgress` = `currentMilestone / totalMilestones` (`src/docs-retrieval/learning-journey-helpers.ts:127`).

Net: the stored value is always the last-loaded milestone's ordinal ratio — and it moves **down** when a user revisits an earlier milestone. A fully-published, fully-completed App Platform path still never reaches 100.

## Severity

**Low.** The disagreement is real and unconditional, but there is probably no rendering surface today: App Platform paths don't surface as `type: 'package'` recommendation cards, and `LearningPathCard`'s `isCompleted` comes from `calculatePathProgress` (`src/learning-paths/learning-paths.hook.ts:70`), not `journeyCompletionStorage`. Treat this as correcting a store that lies before something starts reading it, not as a live UI bug. Note `LearningPathTableOfContents` is not production-rendered.

## Implementation

Fix the gate at `DocsPanelContentArea.tsx:430`, not the write site at `docs-panel.tsx:558`. Clamping at the load-time write would put completion-gate knowledge into a component that has none, and would encode the wrong model.

[...]

PR #1594, opened 2026-08-11
fix: small correctness/hardening follow-ups from #1408 (items 1, 2, 5)

### LINKED ISSUE #1565: Small correctness/hardening follow-ups from #1408 second-round review
Follow-up from the second-round review of PR #1408 (https://github.com/grafana/grafana-pathfinder-app/pull/1408). Grouped small correctness/hardening items — each is independent, low severity, with an exact fix; pick off individually.

> **Scope (updated 2026-08-11):** this issue is items **1, 2 and 5** only. Items 3 and 4 have moved out — see "Moved out" below. The three remaining items are settled and ready to implement.

## Items

1. **Prototype-chain leak through the published-member filter** (hardening; low). `src/learning-paths/app-platform-paths.ts:52` writes `guideMetadata[entry.id] = …` and `:72` filters with `(id) => id in guideMetadata`, where both ids are CR-authored strings. `'constructor' in {}` is true via the prototype chain, so a milestone id like `constructor`/`toString` passes the published filter with no metadata behind it; an entry id of `__proto__` swaps the local object's prototype. Contained (no `Object.prototype` pollution) — worst case is garbage rendering, and a member that can never complete, which pins the path below 100% (`learning-paths.hook.ts:76` divides by `path.guides.length`). Fix: build `guideMetadata` as `Object.create(null)` and gate with `Object.hasOwn(guideMetadata, id)`. Note there are further plain-object sites feeding the same consumer path — `app-platform-paths.ts:30`, `learning-paths.hook.ts:38` and `:203`, and `fetch-path-guides.ts:75`.

2. **Rejected-promise cache eviction lands one microtask late** (latent; low). `src/package-engine/composite-resolver.ts:57-65` registers success-eviction via `promise.then(...)` and rejection-eviction via `.catch` on the _derived_ promise, so a caller that retries synchronously inside its own rejection handler gets the still-cached rejected promise back once. No current caller retries this way; the adjacent comment ("the next call re-fetches fresh") is untrue for exactly that case. Fix: use the two-argument form so both evictions register on the first-hop reaction — `promise.then(onFulfilledEvict, () => this.cache.delete(cacheKey))` — and add a reject-once-then-retry test asserting the inner resolver is called twice (the existing rejection test at `composite-resolver.test.ts:268-274` only asserts propagation).

5. **TS type overclaims an unvalidated wire field** (low). `CustomGuideManifest.type` is declared as a required narrowed union (`src/lib/custom-guide-repository-client.ts:23`), but the Go side serializes the persisted string verbatim with no validation and no omitempty (`pkg/plugin/custom_guide_repository_client.go:44`) — it can be `""` or anything. Current consumers compare against `'path'`/`'journey'` literals so behavior is safe; the type just lies.

   **Fix on the TypeScript side — validate at the boundary** in `requestCatalogue()` (`custom-guide-repository-client.ts:96-98` already has a single shaping return, so one guard mapping an unknown `type` to `undefined` slots in). Also delete `depends` from `CustomGuideManifest`; it has zero readers in `src/` and the wire data still travels downstream as `Record<string, unknown>` regardless.

   **Do not "fix" the Go side.** PR #1580 merged golden-fixture contract tests that deliberately assert the wire is wider than the TS mirror — see `src/validation/backend-api-contract.test.ts:332-361` ("custom-guide manifest is wider than its mirror") and the committed fixture `pkg/plugin/testdata/contract/custom-guide-repository.wire-widened-manifest.json`. Adding validation or `omitempty` in Go would regenerate that golden and break a test that landed deliberately. The loose decode is intentional. If the TS mirror widens, touch up the wording in that test title and the doc comment on `src/types/backend-api.schema.ts:80-91` so they don't go stale.

## Moved out

- **Item 3** (append-only completion numerator) → **#1586**. Previously believed folded into #1559, but #1559 fixed only the denominator. Its stated one-line fix does not compile against the real `markMilestoneDone` signature, so it needs an API-shape decision and is tracked separately.
- **Item 4** (journey-progress store disagrees with the completion gate) → **#1567**.

## Links

- Tech-debt roundup (adjacent, larger extractions): https://github.com/grafana/grafana-pathfinder-app/issues/1557

tomglenn:
## Summary

Items **1, 2 and 5** of #1565 — three independent low-severity follow-ups from the #1408 second-round review, one commit each.

1. **Prototype-chain leak in the published-member filter.** Guide-metadata lookups are now built with `Object.create(null)` and gated with `Object.hasOwn`, so a milestone id like `constructor` / `toString` / `__proto__` can no longer test as a published member with no guide behind it.
2. **Cache eviction now registers on the resolution's first-hop reaction.** `CompositePackageResolver.resolve` passes both handlers to a single `.then` instead of hanging rejection-eviction off a trailing `.catch` on the derived promise.
3. **`CustomGuideManifest.type` no longer overclaims the wire.** `requestCatalogue()` narrows the unvalidated wire string to a `PackageType` or `undefined` in its existing shaping pass; `depends` (zero readers) is deleted from the interface.

Items **3 and 4 are out of scope here** and are *not* fixed by this PR, despite the `Fixes` line closing the umbrella issue: item 3 (append-only completion numerator) is tracked as **#1586** and needs an API-shape decision; item 4 (journey-progress store vs the completion gate) is tracked as **#1567**.

## What to look at

- `src/learning-paths/app-platform-paths.ts` — the `Object.hasOwn` membership gate and the null-prototype `guideMetadata`. Also converted: `EMPTY_RESULT`, `EMPTY_APP_PLATFORM_RESULT` (`learning-paths.hook.ts`), and the map built in `fetch-path-guides.ts`.
- `src/learning-paths/learning-paths.hook.ts` — the App Platform consumer needed no change (its truthy check is already correct against a null-prototype map, verified by test). One site beyond the issue's list is hardened: the final fallback tier reads bundled `paths.json` metadata, which keeps `Object.prototype`, so a guide id of `toString` resolved to a function and rendered untitled. It is the last hop of the same lookup chain, so it is fixed and tested here rather than left behind.
- `src/package-engine/composite-resolver.ts` — the two-argument `.then`. **Read the commit message**: the accompanying test pins the observable contract (a rejected resolution is evicted, so a retry reaches the resolver again) but does **not** discriminate the two forms. `resolve()` is an `async` method, so its wrapper already interposes a microtask hop and the old derived-promise eviction still landed before any caller's rejection handler. The change is robustness and clarity — it would matter the moment `resolve()` stops being `async` — not a live bug fix, and the misleading "the next call re-fetches fresh" framing in the adjacent comment is corrected.
- `src/lib/custom-guide-repository-client.ts` — `narrowPackageType` plus `shapeEntry`, which now carries the `repository` forcing and the `type` narrowing in one place. The response is typed with `WireManifest` / `WireEntry` so the decode site no longer claims a narrowed `type` it hasn't validated.
- `src/types/backend-api.schema.ts` and `src/validation/backend-api-contract.test.ts` — **comment and doc wording only**. The Go side is deliberately untouched: the #1580 golden fixtures assert the wire is wider than this mirror, so adding validation or `omitempty` in Go would regenerate a golden and break a test that landed on purpose. Assertions, titles, and the committed fixture are unchanged; only prose that would have gone stale (the mirror's `type` is now narrowed at the boundary, and it declares no `depends` field at all) was refreshed.

## Why

Item 1's impact is not merely cosmetic. A phantom member reaches `LearningPath.guides`, which is the progress denominator — `calculatePathProgress` divides by `path.guides.length` — so the member can never complete and the path can never reach 100%. It also renders untitled, since there is no metadata behind the id. An entry id of `__proto__` swapped the local object's prototype (contained; no `Object.prototype` pollution).

Item 2 removes a latent ordering dependency and a comment that promised behaviour the code did not deliver for the retry-in-rejection-handler case.

Item 5 removes a type that lied: the Go proxy serializes the stored manifest type verbatim with no validation and no `omitempty`, so the wire can carry `""` or anything else, while the TS interface declared a required narrowed union. Current readers compare against `'path'` / `'journey'` literals, so behaviour was already safe — this makes the declaration true at the boundary instead of relying on every downstream reader being careful.

## How to verify

```bash
npm run check      # full gate: exits 0
npm run test:ci    # 419 suites, 6684 passed / 18 skipped / 5 todo
go test ./pkg/...  # unchanged, passes
```

Every new test was discrimination-checked by breaking the behaviour it covers and confirming failure:

| Test | Behaviour broken | Result |
| --- | --- | --- |
| 5 hostile milestone ids + `__proto__` entry id (`app-platform-paths.test.ts`) | restore `{}` and `in` | all 6 fail |
| 3 inherited names (`fetch-path-guides.test.ts`) | restore `{}` | all 3 fail |
| id-titled member (`learning-paths.hook.test.ts`) | restore the unguarded fallback read | fails (`title` is `undefined`) |
| reject-once-then-retry (`composite-resolver.test.ts`) | drop rejection eviction entirely | fails |
| reject-once-then-retry | restore the derived-promise `.catch` form | **still passes** — see above; reported as a contract guard, not as coverage of the timing change |
| 5 unrecognized `type` values (`custom-guide-repository-client.test.ts`) | pass the wire string through | all 5 fail |
| 3 recognized `type` values | drop every type | all 3 fail |

## Breaking changes

None at runtime. One type-declaration change: `CustomGuideManifest.type` is now `PackageType | undefined` rather than a required `PackageType`, and `depends` is gone. Both are internal to `src/`; the only reader of `type` compares it against `'path'` / `'journey'`, and `depends` had no readers. The wire contract and the Go structs are unchanged.

## Reversibility

Fully reversible. Three independent commits, revertable individually with no shared state, no migrations, and no persisted-format change.

Fixes #1565

### DIFF
--- a/src/learning-paths/app-platform-paths.test.ts
+++ b/src/learning-paths/app-platform-paths.test.ts
@@ -97,6 +97,45 @@ describe('fetchAppPlatformLearningPaths', () => {
     expect(result.guideMetadata['draft-guide']).toBeUndefined();
   });
 
+  // Milestone ids are CR-authored. An inherited built-in name must not pass the
+  // published-member gate: it would render untitled and, because it can never
+  // complete, pin the path's progress denominator below 100%.
+  it.each(['constructor', 'toString', 'hasOwnProperty', '__proto__', 'valueOf'])(
+    'drops the milestone id %s, which has no published guide behind it',
+    async (hostileId) => {
+      mockFetchCustomGuideRepository.mockResolvedValue([
+        {
+          id: 'fe-alerting-path',
+          title: 'Alerting enablement',
+          status: 'published',
+          manifest: { type: 'path', milestones: [hostileId, 'fe-alerting-01'] },
+        },
+        { id: 'fe-alerting-01', title: 'Alerting module 1', status: 'published' },
+      ]);
+
+      const result = await fetchAppPlatformLearningPaths('stacks-123');
+
+      expect(result.paths[0]!.guides).toEqual(['fe-alerting-01']);
+      expect(result.guideMetadata[hostileId]).toBeUndefined();
+    }
+  );
+
+  it('records an entry id of __proto__ as an own metadata key instead of swapping the prototype', async () => {
+    mockFetchCustomGuideRepository.mockResolvedValue([
+      { id: '__proto__', title: 'Hostile', status: 'published' },
+      { id: 'fe-alerting-01', title: 'Alerting module 1', status: 'published' },
+    ]);
+
+    const result = await fetchAppPlatformLearningPaths('stacks-123');
+
+    expect(Object.hasOwn(result.guideMetadata, '__proto__')).toBe(true);
+    expect(result.guideMetadata['fe-alerting-01']).toEqual({
+      title: 'Alerting module 1',
+      estimatedMinutes: 5,
+      url: 'backend-guide:fe-alerting-01',
+    });
+  });
+
   it('falls back to id when a path manifest has no description', async () => {
     mockFetchCustomGuideRepository.mockResolvedValue([
       {
--- a/src/learning-paths/app-platform-paths.ts
+++ b/src/learning-paths/app-platform-paths.ts
@@ -27,7 +27,7 @@ export interface AppPlatformPathsResult {
   guideMetadata: Record<string, GuideMetadataEntry>;
 }
 
-const EMPTY_RESULT: AppPlatformPathsResult = { paths: [], guideMetadata: {} };
+const EMPTY_RESULT: AppPlatformPathsResult = { paths: [], guideMetadata: Object.create(null) };
 
 /**
  * Fetches the namespace's custom-guide catalogue and splits it into
@@ -47,7 +47,9 @@ export async function fetchAppPlatformLearningPaths(namespace: string): Promise<
     return EMPTY_RESULT;
   }
 
-  const guideMetadata: Record<string, GuideMetadataEntry> = {};
+  // Null-prototype: ids are CR-authored, so `constructor`/`toString`/`__proto__`
+  // must not resolve through Object.prototype in the membership gate below.
+  const guideMetadata: Record<string, GuideMetadataEntry> = Object.create(null);
   for (const entry of published) {
     guideMetadata[entry.id] = {
       title: entry.title || entry.id,
@@ -69,7 +71,7 @@ export async function fetchAppPlatformLearningPaths(namespace: string): Promise<
       // in a published path's manifest must not leak into My Learning (it has no
       // guideMetadata entry, so it would render titled by its raw id and inflate
       // the denominator). Matches the resolver's published-only gate.
-      guides: (entry.manifest?.milestones ?? []).filter((id) => id in guideMetadata),
+      guides: (entry.manifest?.milestones ?? []).filter((id) => Object.hasOwn(guideMetadata, id)),
       badgeId: '',
       // Carried so the My Learning launch can render members with milestone
       // chrome (packageInfo) rather than as standalone guides.
--- a/src/learning-paths/fetch-path-guides.test.ts
+++ b/src/learning-paths/fetch-path-guides.test.ts
@@ -134,6 +134,22 @@ describe('fetchPathGuides', () => {
     );
   });
 
+  // resolveGuideMetadata looks this map up by arbitrary guide id, so an id that
+  // happens to name an Object.prototype member must miss, not hand back a function.
+  it.each(['constructor', 'toString', 'hasOwnProperty'])(
+    'resolves no metadata for the inherited name %s',
+    async (inheritedName) => {
+      mockFetch.mockResolvedValueOnce({
+        ok: true,
+        json: async () => SAMPLE_INDEX_JSON,
+      });
+
+      const result = await fetchPathGuides('https://grafana.com/docs/learning-paths/linux-server-integration/');
+
+      expect(result!.guideMetadata[inheritedName]).toBeUndefined();
+    }
+  );
+
   it('returns null on fetch failure', async () => {
     mockFetch.mockResolvedValueOnce({
       ok: false,
--- a/src/learning-paths/fetch-path-guides.ts
+++ b/src/learning-paths/fetch-path-guides.ts
@@ -72,7 +72,10 @@ export async function fetchPathGuides(pathUrl: string, signal?: AbortSignal): Pr
     const validItems = data.filter((item) => !item.params?.grafana?.skip);
 
     const guides: string[] = [];
-    const guideMetadata: Record<string, GuideMetadataEntry> = {};
+    // Null-prototype: slugs are remote-authored, and consumers look this up by
+    // arbitrary guide id, so `constructor`/`toString` must miss rather than
+    // resolve through Object.prototype.
+    const guideMetadata: Record<string, GuideMetadataEntry> = Object.create(null);
 
     for (const item of validItems) {
       const relpermalink: string = item.relpermalink || '';
--- a/src/learning-paths/learning-paths.hook.test.ts
+++ b/src/learning-paths/learning-paths.hook.test.ts
@@ -114,6 +114,26 @@ describe('useLearningPaths — App Platform path ingestion', () => {
     );
   });
 
+  // The bundled paths.json metadata keeps Object.prototype, and the App Platform
+  // tier is null-prototype, so a member id naming a built-in must fall through to
+  // the id-titled default rather than resolving to Object.prototype.toString.
+  it('titles a member named after an Object.prototype member by its id', async () => {
+    mockFetchAppPlatformLearningPaths.mockResolvedValue({
+      paths: [
+        { id: 'fe-alerting-path', title: 'Alerting enablement', description: '', guides: ['toString'], badgeId: '' },
+      ],
+      guideMetadata: Object.create(null),
+    });
+
+    const { result } = renderHook(() => useLearningPaths());
+
+    await waitFor(() => expect(result.current.paths.map((p) => p.id)).toContain('fe-alerting-path'));
+
+    const [member] = result.current.getPathGuides('fe-alerting-path');
+    expect(member!.title).toBe('toString');
+    expect(member!.url).toBeUndefined();
+  });
+
   it('does not fetch when no namespace is available', async () => {
     mockNamespace = undefined;
 
--- a/src/learning-paths/learning-paths.hook.ts
+++ b/src/learning-paths/learning-paths.hook.ts
@@ -35,7 +35,7 @@ import { fetchPathGuides, type FetchedPathGuides } from './fetch-path-guides';
 import { fetchAppPlatformLearningPaths, type AppPlatformPathsResult } from './app-platform-paths';
 import { markGuideCompleted as coordinatorMarkGuideCompleted } from './badge-coordinator';
 
-const EMPTY_APP_PLATFORM_RESULT: AppPlatformPathsResult = { paths: [], guideMetadata: {} };
+const EMPTY_APP_PLATFORM_RESULT: AppPlatformPathsResult = { paths: [], guideMetadata: Object.create(null) };
 
 // ============================================================================
 // CONSTANTS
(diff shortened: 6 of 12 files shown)


moxious (approved):
Reviewed with `/review` — routed concerns: security, correctness, testing, reversibility, cross-cutting, plus `package-engine` and `guide-schema-and-contracts`. Contract-evolution gate fired for `guide-schema-and-contracts`; verdict `coherent_extension`, no finding. CI green.

Approving — nothing here blocks merge. Nice PR: the discrimination table earned its keep, I re-ran four of the six mutations independently and they all hold, and the honesty about the composite-resolver test not discriminating saved me from filing it as a coverage gap. One item I'd like to see fixed before this merges if you're willing; the rest are nits.

---

**[suggestion] `resolveGuideMetadata` still has one ungated lookup, and it's the one that throws** — `learning-paths.hook.ts:192`

The PR hardens the read at `:209` and the body says that's "the last hop of the same lookup chain." The hop 15 lines above is still open:

```ts
const scoped = dynamicGuideData[pathId]?.guideMetadata[guideId];
```

`dynamicGuideData` is a plain `{}` (`:93`, refilled from `{}` at `:117`), and `?.` guards only the first hop. So when `pathId` names an inherited member, `dynamicGuideData['constructor']` is a truthy function, `.guideMetadata` is `undefined`, and `undefined[guideId]` throws. Reproduced at head:

```
TypeError: Cannot read properties of undefined (reading 'fe-alerting-01')
  at src/learning-paths/learning-paths.hook.ts:192:63
  at resolveGuideMetadata → getPathGuides
```

Unlike the member-id case, this needs nothing but the catalogue listing — `spec.id` is free-form and `:174` merges App Platform paths in unfiltered, and the new `Object.hasOwn` member filter vets only the _member_ ids, so a path with a hostile id still has a non-empty `guides` array. It surfaces in two places: `MyCoursesSection` (contained by `MyLearningErrorBoundary`, though "Try again" re-throws, so it's a persistent dead surface) and `UserProfileBar` via `context-panel.tsx:1149`, which has no local boundary and falls through to `PluginErrorBoundary`.

Two caveats in your favour: it is **pre-existing** (identical at base — not a regression you introduced), and it was **not** in #1565's list — that issue's `:203` points at the App Platform branch, not `:192`. So this is genuinely adjacent-but-unscoped work, and a follow-up issue is a defensible call. My preference is to fix it here, because it's a one-line change inside a function you're already editing, it fails worse than the case you did fix, and shipping the sibling unfixed leaves the review record reading as "this class is closed" when it isn't:

```ts
const [dynamicGuideData, setDynamicGuideData] = useState<
  Record<string, FetchedPathGuides>
>(() => Object.create(null));
// and `const results = Object.create(null)` at :117
```

Both maps are keyed by trusted bundled `path.id`, so nothing else changes. Worth also parameterizing the new hook test over the **pathId** position — the current one puts the hostile string in the guideId position, which is exactly why it passes today.

While you're there: `:203` (`appPlatformData.guideMetadata[guideId]`) is still a bare bracket read while `:209` got the gate. You're right that it's safe today — I verified every producer is null-prototype — but `Record<string, GuideMetadataEntry>` doesn't carry that invariant, and two existing test doubles in `learning-paths.hook.test.ts` (~`:80`, `:103`) already pass plain `{}`. Gating it removes the invariant instead of documenting it, which would also let `EMPTY_RESULT` / `EMPTY_APP_PLATFORM_RESULT` go back to plain `{}`.

---

**[nit] the composite-resolver test comment asserts the counterfactual your PR body correctly retracts**

`composite-resolver.test.ts` says a caller "would otherwise be handed the still-cached rejected promise and never reach the resolver." Your body already explains why that never held. I measured it: the old form evicts at hop 1 with the caller reacting at hop 2, and the new form is the same, because the eviction reaction is registered synchronously on `promise` while a caller can only register on the promise `async resolve()` produces by adoption. Keep the two-argument form — it's tighter and drops the dependence on `resolve()` staying `async` — but the comment should say _that_, and the test title should claim the caller-visible retry contract it actually pins. (I also checked the one real semantic loss: a throw in the fulfillment handler is no longer caught by a trailing `.catch`. Unreachable — `result.repository`, `Set.has` and `Map.delete` can't throw, and `resolveUncached` already dereferences `result.ok`.)

---

**[nit] the refreshed contract prose overclaims on `depends`**

`backend-api-contract.test.ts:328-332` still says "Field names line up," but removing `depends` from `CustomGuideManifest` means they no longer do — Go still declares `Depends []json.RawMessage`. And "the client closes the gap at its own boundary" is true for `type` but not for `depends`: `shapeEntry` spreads `...manifest`, so the wire value still rides through onto a `CustomGuideManifest`-typed object and into `packageManifest: Record<string, unknown>`. It's _undeclared_, not closed. Same at `backend-api.schema.ts:80` — "wider **here than in** `CustomGuideManifest`" no longer describes a field that has no counterpart at all. Worth naming the two mechanisms separately, since this comment is the only record a future reviewer has for this hand-mirrored boundary.

---

**[nit]** `custom-guide-repository-client.ts:22-26` says `requestCatalogue` drops non-`PackageType` values — the drop is in `narrowPackageType`/`shapeEntry`; `requestCatalogue` just calls `guides.map(shapeEntry)`. The rationale is also restated in the `shapeEntry` block. Cut the JSDoc to the one invariant the type can't express (the wire can carry `""`).

**[nit]** `narrowPackageType(undefined)` is the case the doc comment promises and the only one untested — every fixture carries a `type` key. Behaviour is correct (`z.enum` rejects `undefined` without throwing, verified), so it's one more `it.each` row. It matters if anyone later switches to `.parse()`, or Go gains `omitempty`: that becomes a throw inside `requestCatalogue`, swallowed into an empty catalogue with a `transport-error` token — the silent-empty-surface failure this client was hardened against.

**[nit]** The three new `fetch-path-guides.test.ts` cases assert only `toBeUndefined()` with no positive control, so they'd also pass if `guideMetadata` were empty or ingestion broke entirely — `app-platform-paths.test.ts` gets this right by asserting a real id still resolves in the same map.

---

**Follow-up, not this PR:** narrowing membership makes the derived My Learning denominator disagree with the _persisted_ one. `journeyCompletionStorage` stores `currentMilestone/totalMilestones` from `resolvePackageMilestones`, which keeps an unresolvable milestone as a locked placeholder rather than dropping it — so My Learning can reach 100% while the stored percentage stays (N−1)/N, and that stored value is what the recommendation card reads back. Nothing irreversible fires (App Platform paths carry `badgeId: ''`, and `markMilestoneDone` gates on `countUnlockedMilestones`, not `path.guides`), and it does not touch #1586 — that numerator is slug-keyed and a phantom milestone can never enter it. Worth its own issue, since aligning it changes a persisted value's meaning.

**Tech debt, non-blocking:** the `Object.create(null)` + own-key idiom is now open-coded at six production sites across three files with no shared helper, and the repo carries two spellings of the lookup half (`Object.hasOwn` in `package-engine/resolver.ts` vs `Object.prototype.hasOwnProperty.call` in `cli/utils/cli-validators.ts` and `validation/normalize-guide-aliases.ts`). `Object.create(null)` also returns `any`, so each construction site launders an untyped value into a `Record<string, GuideMetadataEntry>` slot. A branded `GuideMetadataMap` built only by `createGuideMetadataMap()` would make the ungated branches provably safe rather than safe-by-convention. Separately, `getPathsData().guideMetadata` is indexed at three sites and this PR gated one — `MyLearningTab.tsx:172` and `useNextLearningAction.ts:105` survive only via `?.title`/`?.url`.

**Docs:** one gap worth closing in the same PR — the tier-spanning rule isn't recorded anywhere. Suggested addition to `docs/developer/learning-paths/README.md` after the content-system paragraph: _Guide ids are author-supplied — remote slugs, or milestone ids from a custom-guide custom resource — so `guideMetadata` lookups gate on own keys: the remote and App Platform maps are built with `Object.create(null)`, and the bundled JSON maps keep `Object.prototype`, so the hook's fallback read tests `Object.hasOwn` first._

---

**Verified so you don't have to re-derive it:** `PackageTypeSchema` is a bare `z.enum` with no coercion, transform or catch; `safeParse` rejects `''`, `'PATH'`, `'journey '` and a NUL-containing value, and returns failure rather than throwing for `undefined`. The Go struct really is `Type string` with a `json:"type"` tag and no `omitempty`, and the committed golden really carries `"type":""` — so leaving the Go side alone is correct, and #1580's deliberate-widening claim checks out. The contract test and fixture changes really are comment-only. The new `src/lib/` → `src/types/package.schema` import edge is Tier 1 → Tier 0 and does not widen the Node-reachability closure. `depends` has no reader.

Contract-evolution detail, for the record: the client was already the designated boundary-shaping owner (the `repository` forcing since #1408), so naming that pass `shapeEntry` and adding `type` narrowing inside it is a coherent extension rather than a second owner competing with the zod wire mirror — which parses nothing at runtime by design. No anchor row required; recording one for the custom-guide wire mirror would be a reasonable optional follow-up.

<sub>Reviewed with `/review` (10 concern reviewers + contract-evolution scan + 8 adversarial verifications). Two candidate findings were dropped by verification: a shared-singleton test gap (3/3 skeptics refuted — the right fix is gating `:203`, which makes the constant conversions unnecessary) and one further item I'll raise off-thread.</sub>

PR #1595, opened 2026-08-11
fix(custom-guides): signal a malformed catalogue instead of caching it empty

### LINKED ISSUE #1591: Malformed guides field in the catalogue response is cached as a successful empty catalogue with no signal
## Problem

`requestCatalogue()` in `src/lib/custom-guide-repository-client.ts` guards the `guides` field defensively:

```ts
const guides = Array.isArray(response.guides) ? response.guides : [];
```

The guard is fail-safe and correct as far as it goes. What it does not do is *say anything*. If the backend ever returns HTTP 200 with `capability.available: true` and a `guides` field that is missing, null, or not an array, this silently yields an empty catalogue — which is then **cached for the full TTL** as a successful response.

Operationally that is indistinguishable from "this stack has no custom guides authored". The Custom Guides surface renders empty, the catalogue-unavailable metric stays at zero, and nothing appears in the console.

## Why it's worth signalling

This is the one remaining silent-empty path in this client. The two neighbours both report now:

- capability unavailable (soft 200) → `logger.warn` + `recordCustomGuideCatalogueUnavailable(reason)`
- hard failure (503, network, timeout) → same pair, with a `http-<status>` / `transport-error` reason (PR #1587)

Schema drift between the Go proxy and this client is exactly the class of bug #1526 and PR #1580 exist to catch, and this is the runtime half of it.

## Suggested fix

When `response.guides` is present but not an array, emit the same pair with a distinct bounded reason (e.g. `malformed-response`), and consider not caching that result — a schema mismatch is not a state worth holding for the whole TTL.

Note the distinction between *absent* and *malformed*: an absent `guides` on an available capability is plausibly a legitimate empty catalogue, whereas a non-array value is definitely wrong. Only the latter clearly warrants a signal.

## Scope note

Raised at review request on PR #1587, which deliberately left this out of scope — that PR fixes the hard-failure path only. The existing behaviour is pinned by the "response is malformed" test in `src/lib/custom-guide-repository-client.test.ts`, which asserts the empty-array result; that test should be extended rather than replaced.

dchaudhari7177:
Closes #1591. Follows PR #1587, which fixed the hard-failure path and deliberately left this one out of scope.

## What

`requestCatalogue()` guarded `guides` with `Array.isArray` and fell back to `[]`. Fail-safe, but silent — and the empty result was then **cached for the full TTL**, so a schema mismatch was indistinguishable from "this stack has no guides authored": empty surface, catalogue-unavailable metric at zero, nothing in the console.

Now a present-but-non-array `guides` emits the same pair as its two neighbours, with a bounded `malformed-response` reason, and the result is **not cached**.

The no-cache half is the part I'd flag for review: drift may well be fixed by the next backend deploy, and caching would keep the surface empty for the whole TTL *after* it recovered. `requestCatalogue` now returns `{ entries, cacheable }` so the caller decides; the capability-unavailable path keeps its existing caching behaviour, unchanged.

## The absent/malformed distinction

The issue calls this out and it turned out to matter more than it first looks. **`null` is treated as absent, not malformed** — the proxy is Go, and `json.Marshal` of a nil `[]guide` emits `null`, so a stack with no guides at all is expected to send exactly that. Signalling on it would fire the metric on the most ordinary case there is.

So:

| `guides` | treated as | signal |
| --- | --- | --- |
| omitted | empty catalogue | silent |
| `null` | empty catalogue | silent |
| `[]` | empty catalogue | silent, cached |
| object / string / number / boolean | drift | `malformed-response`, not cached |

If you'd rather `null` were also treated as drift, that's a one-word change — but I think the Go nil-slice encoding argues against it.

## Tests

`custom-guide-repository-client.test.ts` extended, not replaced, as the issue asks — the existing "response is malformed" case (which actually exercises the *absent* path) still passes untouched. Added 8:

- `guides` omitted / `null` → asserts **no** warn and **no** metric
- `guides` as object / string / number / boolean → asserts `malformed-response` on both the logger and the metric
- a malformed response is not cached: second call re-fetches and picks up the recovered catalogue
- a legitimately empty `[]` **is** still cached (one `get`, not two)

With `src/lib/custom-guide-repository-client.ts` reverted to `main`, **5 of the new tests fail**; the 3 that pin deliberately-unchanged behaviour pass either way, which is what I'd want them to do.

## Checks run locally

| gate | result |
| --- | --- |
| `jest src/lib/custom-guide-repository-client.test.ts` | 34 passed (26 existing + 8 new) |
| `npm run typecheck` | pass |
| `npm run lint` | 0 errors (24 pre-existing warnings elsewhere) |
| `prettier --check` | clean |
| `jest src/validation` | 818 passed, 4 failed |

Those 4 failures are in `src/validation/import-graph.test.ts` and are **identical on pristine `main`** on my machine — they look like Windows path-separator assumptions, not something this PR touches. I didn't run `npm run check` in full (no Go toolchain here), so `lint:go` / `test:go` / coverage thresholds are CI's.

Comments kept to the AGENTS.md keep-list — the only one carrying real weight is the Go nil-slice note, since that's the bit that isn't obvious from the code.

### DIFF
--- a/src/lib/custom-guide-repository-client.test.ts
+++ b/src/lib/custom-guide-repository-client.test.ts
@@ -150,6 +150,68 @@ describe('fetchCustomGuideRepository', () => {
     expect(result).toEqual([]);
   });
 
+  // An omitted or null `guides` on an available capability is an ordinary empty
+  // catalogue, not drift — Go's json.Marshal of a nil slice emits `null`, so the
+  // proxy sends exactly this for a stack with no guides. Signalling here would
+  // fire on the most common case there is.
+  it.each([
+    { shape: 'omitted', response: { capability: { available: true } } },
+    { shape: 'null', response: { capability: { available: true }, guides: null } },
+  ])('stays silent when guides is $shape', async ({ response }) => {
+    mockGet.mockResolvedValue(response);
+
+    const result = await fetchCustomGuideRepository('stacks-123');
+
+    expect(result).toEqual([]);
+    expect(recordCustomGuideCatalogueUnavailable).not.toHaveBeenCalled();
+    expect(logger.warn).not.toHaveBeenCalled();
+  });
+
+  // A present-but-non-array `guides` cannot be a catalogue under any encoding,
+  // so it is schema drift and must be countable rather than silently empty.
+  it.each([
+    { shape: 'an object', guides: { 'fe-alerting-path': {} } },
+    { shape: 'a string', guides: 'fe-alerting-path' },
+    { shape: 'a number', guides: 3 },
+    { shape: 'a boolean', guides: true },
+  ])('reports malformed-response when guides is $shape', async ({ guides }) => {
+    mockGet.mockResolvedValue({ capability: { available: true }, guides });
+
+    const result = await fetchCustomGuideRepository('stacks-123');
+
+    expect(result).toEqual([]);
+    expect(recordCustomGuideCatalogueUnavailable).toHaveBeenCalledWith('malformed-response');
+    expect(logger.warn).toHaveBeenCalledWith('[custom-guides] catalogue malformed', {
+      reason: 'malformed-response',
+    });
+  });
+
+  it('does not cache a malformed response for the TTL', async () => {
+    mockGet.mockResolvedValueOnce({ capability: { available: true }, guides: { nope: true } });
+
+    await expect(fetchCustomGuideRepository('stacks-123')).resolves.toEqual([]);
+
+    // The backend recovers on the very next call; a cached empty would hide it.
+    mockGet.mockResolvedValueOnce({
+      capability: { available: true },
+      guides: [{ id: 'fe-alerting-path', title: 'Alerting enablement', status: 'published' }],
+    });
+
+    const recovered = await fetchCustomGuideRepository('stacks-123');
+
+    expect(recovered).toHaveLength(1);
+    expect(mockGet).toHaveBeenCalledTimes(2);
+  });
+
+  it('still caches a legitimately empty catalogue', async () => {
+    mockGet.mockResolvedValue({ capability: { available: true }, guides: [] });
+
+    await fetchCustomGuideRepository('stacks-123');
+    await fetchCustomGuideRepository('stacks-123');
+
+    expect(mockGet).toHaveBeenCalledTimes(1);
+  });
+
   // The reason lands on a Faro event attribute and in the bridged log context,
   // so every rejection shape must collapse to a token from the closed set.
   it.each([
--- a/src/lib/custom-guide-repository-client.ts
+++ b/src/lib/custom-guide-repository-client.ts
@@ -101,6 +101,20 @@ function reportCatalogueFetchFailure(err: unknown): void {
   }
 }
 
+interface CatalogueResult {
+  entries: CustomGuideRepositoryEntry[];
+  cacheable: boolean;
+}
+
+// Bounded token, same closed vocabulary as the reasons above (TELEMETRY.md).
+const MALFORMED_REASON = 'malformed-response';
+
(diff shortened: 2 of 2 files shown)


src/lib/custom-guide-repository-client.ts
```diff
@@ -108,7 +122,14 @@ async function requestCatalogue(): Promise<CustomGuideRepositoryEntry[]> {
...
+    return { entries: [], cacheable: true };
+  }
+  if (isMalformedGuides(response.guides)) {
+    // Schema drift between the Go proxy and this client would otherwise render
+    // as "no guides authored", with the metric at zero and a silent console.
+    logger.warn('[custom-guides] catalogue malformed', { reason: MALFORMED_REASON });
+    recordCustomGuideCatalogueUnavailable(MALFORMED_REASON);
+    return { entries: [], cacheable: false };
   }
   const guides = Array.isArray(response.guides) ? response.guides : [];
```
tomglenn, line 167: Now that `isMalformedGuides` already narrows `response.guides` to `undefined | null | array` by this point, the `Array.isArray` check here is redundant. Purely optional:

```suggestion
  const guides = response.guides ?? [];
```

src/lib/custom-guide-repository-client.test.ts
```diff
@@ -121,6 +121,68 @@ describe('fetchCustomGuideRepository', () => {
...
+    expect(logger.warn).not.toHaveBeenCalled();
+  });
+
+  // A present-but-non-array `guides` cannot be a catalogue under any encoding,
+  // so it is schema drift and must be countable rather than silently empty.
+  it.each([
+    { shape: 'an object', guides: { 'fe-alerting-path': {} } },
+    { shape: 'a string', guides: 'fe-alerting-path' },
+    { shape: 'a number', guides: 3 },
+    { shape: 'a boolean', guides: true },
```
tomglenn, line 176: Optional: these only cover truthy non-array values. Worth adding the falsy-but-present shapes too, since a truthiness-based (instead of `!== undefined && !== null`) implementation would pass all of these while still misclassifying `false`/`0`/`''` as "absent":

```suggestion
    { shape: 'an object', guides: { 'fe-alerting-path': {} } },
    { shape: 'a string', guides: 'fe-alerting-path' },
    { shape: 'a number', guides: 3 },
    { shape: 'a boolean', guides: true },
    { shape: 'false', guides: false },
    { shape: 'zero', guides: 0 },
    { shape: 'an empty string', guides: '' },
```

tomglenn (dismissed):
Thanks for taking this on — nice, clean fix, and it closes #1591 exactly as scoped. Appreciated the discipline of running your own "reverted to main" check and putting the result in the description; I independently reproduced it and it holds exactly.

Two tiny optional nits inline, nothing blocking. Thanks for the contribution!

tomglenn:
@dchaudhari7177 looks like there is a merge conflict that needs to be resolved. Please could you resolve this and then we'll be happy to merge. :)

dchaudhari7177:
Rebased onto `main` (now at f86637b) and force-pushed — `mergeable` is back to `MERGEABLE`.

The conflict was a real semantic one, not just adjacent lines: `main` gained `shapeEntry`/`narrowPackageType` (#1600-ish), which does the `repository: 'app-platform'` forcing this branch had inlined, plus `type` narrowing this branch didn't have. I dropped my inline version and kept `shapeEntry` — the malformed-catalogue change now just wraps it, returning `{ entries: guides.map(shapeEntry), cacheable: true }`, so the two changes compose rather than one overwriting the other.

`npx jest src/lib/custom-guide-repository-client.test.ts --coverage=false` — 42/42 passing on the rebased head.

moxious:
Thanks for the contribution. My quick review seemed like things are OK here, but you have a failing unit test. Recommend rebasing to latest main and looking into fixing the test.

Jayclifford345 (approved):

PR #1596, opened 2026-08-11
fix(storage): gate guide-response lookups on own keys

moxious:
## Intent

Make the guide-response storage lookups gate on own keys.

Guide-response records in `guideResponseStorage` (`src/lib/user-storage.ts`) are keyed by an author-supplied guide id derived from the content URL, and by author-supplied input-block variable names. Every keyed lookup should therefore gate on `Object.hasOwn`: an id or variable name that happens to name an inherited `Object.prototype` member must miss rather than resolve through the prototype chain.

This follows the own-key convention already used in `src/package-engine/resolver.ts`. PR #1594 (in flight, not depended on here) applies the same idiom to the learning-path guide-metadata maps.

## What changed

- The four keyed lookups in `guideResponseStorage` — `getForGuide`, `getResponse`, `setResponse`, `deleteResponse` — now gate on `Object.hasOwn` before reading the responses container. `setResponse` builds the per-guide record from an own-key hit (or a fresh object) before writing; `deleteResponse` operates on that resolved record.
- Added a `guideResponseStorage — own-key lookups` suite in `src/lib/user-storage.test.ts` covering the ordinary round-trip plus inherited-member cases for both guide ids and variable names.

12 changed lines of production code plus focused tests.

## Deliberately out of scope

These are decisions, not omissions:

- `clearForGuide` is unchanged — `delete all[guideId]` only removes own properties and never walks the prototype chain, so it is already own-key by construction.
- `getAll`, `GuideResponsesSchema` and `DEFAULT_GUIDE_RESPONSES` are unchanged. Own-key gating at the lookup sites is sufficient; changing the container shape or the schema was ruled out as unnecessary scope.
- No shared own-key or null-prototype helper was extracted, and the other storage records in this file were not converted.
- No new methods, no signature changes, no event or telemetry changes, no schema changes.
- Two adjacent gaps are deferred to follow-up issues rather than fixed here: the UI read path for the same data (`GuideResponseContext`, `variable-substitution.ts`, `input-block.tsx`), which never touches storage; and `getAll()` returning the shared module-level `DEFAULT_GUIDE_RESPONSES`, which `setResponse` then mutates. The new tests seed an empty record into storage in `beforeEach` so they stay order-independent without a production change for the latter.

## How to verify

```bash
npx jest src/lib/user-storage.test.ts --coverage=false   # 53 passed
npm run check                                             # full gate
```

Green locally: `tsc --noEmit`, `eslint` on both changed files, the focused suite, and the full `npm run check` gate (419 suites, 6670 tests). Mutation-checked: reverting only the production change fails exactly the four new own-key tests, while the baseline round-trip test passes either way.

No UI surface is involved, so no browser or screenshot verification applies.

## Reversibility

Fully reversible. One commit, one module, no persisted-format change and no migration.

### DIFF
--- a/src/lib/user-storage.test.ts
+++ b/src/lib/user-storage.test.ts
@@ -2,6 +2,7 @@ import { getAppEvents } from '@grafana/runtime';
 
 import {
   __resetQuotaWarningForTests,
+  guideResponseStorage,
   interactiveCompletionStorage,
   interactiveStepStorage,
   journeyCompletionStorage,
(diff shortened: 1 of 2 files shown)


leslie-heinzen:
## ✅ Approved

**Verdict: approve.** No blocking findings remain after verification.

### Routing
Classification is `product-runtime` and `mixed`. Five always-on concerns ran. No conditional concern reached `min_signals`, because the diff touches one runtime file. Coverage confidence is high.

### The fix corrects a real fault
I confirmed the base-commit behavior. I did not infer it.
1. Base `setResponse('__proto__', name, value)` ran `all['__proto__'][name] = value`. This wrote onto the live `Object.prototype`, and the value leaked to unrelated objects.
2. Base `deleteResponse('__proto__', 'toLocaleString')` erased the built-in method.
3. One production consumer reaches this code: `guideVariableCheck` (`src/requirements-manager/checks/vars.ts:47`). It backs the `var-<name>:<value>` requirement.
4. At base, a guide with `var-toString:*` resolved to an inherited function. The requirement passed, but the user never answered the input. At head it fails correctly.
5. The guide schema accepts `toString` as a variable name. An author can write it, so the fault is reachable.
6. Tests 3 and 4 fail against the base implementation. They are real regression guards.

The change also closed a latent XSS path. At base, an authored `defaultValue` and a guide path of `__proto__` can persist an attacker string onto `Object.prototype`. The `subHtml` function escapes own keys only, through `Object.entries`. An inherited value then reached `dangerouslySetInnerHTML` without escape. Head removes the write.

**Note:** the escape loop is the last defense here, not the storage layer. If a pollution primitive returns, that loop does not stop it.

Reversibility is clean. The `z.record` schema in Zod strips a top-level `__proto__` from parsed payloads. No such record can persist. The base damage stayed in memory and cleared on reload. A revert needs no migration.

### One small point (low)
The last line of `setResponse` is `all[guideId] = guideRecord`. The read paths have a gate. This line does not. For `guideId === '__proto__'` the line reparents the container and creates no own key, so the write is a silent no-op. I confirmed this against the real module. The round trip returns `{}`.

This is still better than base. The trigger needs a guide served at exactly `/__proto__`. It is not a reason to block. But the PR body describes the write path as gated, and the test passes whether the write persists or vanishes. You can assert durability, or use `Object.defineProperty` for the write.

### Adversarial verification
Reviewers raised five findings of medium severity or higher. All five dropped. Two or more of three skeptics refuted each one.
1. **Storage and context `hasResponse` disagree** — dropped. No production code calls `guideResponseStorage.hasResponse`.
2. **`variable-substitution.ts` resolves inherited members** — dropped. The file is untouched and the effect is cosmetic.
3. **`bounded-record-storage.ts` lacks the guard** — dropped. The file is untouched. The finding was also wrong on one fact: `milestoneCompletionStorage` is hand-written and reads through `Object.entries`.
4. **`setResponse` write path at high severity** — dropped at high severity. It remains as the low point above.
5. **Contract-evolution verdict `contract_missing`** — dropped on two grounds. The cited sites solve different problems. Also, §3b says not to run the gate for always-on concerns, and this concern is always-on. That scan was my error.

### Tech debt
One high-confidence finding. It is pre-existing, and this PR did not cause it. It sits on the lines the PR rewrote.

`getAll()` returns the module-level constant `DEFAULT_GUIDE_RESPONSES` by reference on the empty path and the validation-failure path. Then `setResponse` and `deleteResponse` mutate the object they receive. I confirmed the effect against the real module with an ordinary guide ID:

```
await setResponse('my-guide', 'region', 'us-east-1');
await clearAll(); localStorage.clear();
await getAll();  // → {"my-guide":{"region":"us-east-1"}}  — expected {}
```

Cleared responses come back for the rest of the page session. Two fixes work: freeze the constant with `Object.freeze`, and return a fresh `{}` from the fallback paths. One more point: the `beforeEach` block in the new test seeds storage to avoid this path, so the tests never exercise first-run object identity.

A weaker point: the own-key guard appears three times in three shapes. A small `getOwnGuideRecord(all, guideId)` helper keeps a fourth call site from dropping the guard.

### Documentation drift
None. The PR adds no subsystem, script, skill, route, flag, or storage key.

### Instrumentation
I skipped this check, per the routing in `TELEMETRY.md`. This PR fixes existing behavior and adds no feature behavior.

### Follow-ups (separate PRs, none blocking)
1. Correct the `DEFAULT_GUIDE_RESPONSES` aliasing fault. It loses and revives user data.
2. Add own-key guards to `GuideResponseContext.tsx`, `variable-substitution.ts`, and `bounded-record-storage.ts:73`, in one consistency pass.

leslie-heinzen (approved):

PR #1602, opened 2026-08-12
fix(validation): stop the JSONC stripper eating comments inside strings

Jayclifford345:
## Summary

`readJsoncFile` in `src/validation/import-graph.ts` stripped JSONC comments with two blind regexes:

```ts
text.replace(/\/\*[\s\S]*?\*\//g, '').replace(/^\s*\/\/.*$/gm, '')
```

Neither is string-aware, so any string *value* containing comment syntax gets corrupted. The concrete case: a tsconfig `include` of `"src/cli/**/*"` comes back as `"src/cli*"` — the `/**/` is matched as an empty block comment and removed — silently narrowing the glob to a path that matches nothing.

Replaced with a string-aware scan that tracks quoting and backslash escapes and only strips comments outside string literals, extracted as `stripJsonComments` so the string cases are directly testable.

## Why now

Latent today — the existing callers (`loadTsconfigPaths`, `architecture.test.ts`) only read `compilerOptions.paths` and `baseUrl`, which contain no `/**/`. It surfaced while writing a test that reads `tsconfig.cli.json`'s `include`, where the corruption made the expansion silently return nothing rather than fail.

Split out of #1597 to keep that PR scoped to the data-check step type.

## What to look at

- `stripJsonComments` — the escape handling (`\\` inside a string consumes the next char, so `\"` doesn't end the string) and the unterminated-comment path, which truncates rather than throwing.

## How to verify

`npx jest src/validation/import-graph.test.ts` — 130 pass. The `keeps recursive globs intact` case fails against the old implementation, which is the regression this locks in.

Full `npm run test:ci`: 424 suites / 6739 tests pass. `npm run typecheck` and `npm run lint` clean (0 errors).

## Breaking changes

None. Comment stripping is unchanged for every input that doesn't embed comment syntax in a string.

## Reversibility

Fully reversible — one self-contained function, no persisted state or external contract.

## Follow-up not included here

The failure that led here was a `Dockerfile.cli` COPY list that didn't cover the CLI's compile closure — invisible locally, fails minutes into the CLI publish Docker build. Fixed for the immediate case in #1597; a guard test for the general case is still open (#1603) because the version I wrote destabilised the jest run.

### DIFF
--- a/src/validation/import-graph.test.ts
+++ b/src/validation/import-graph.test.ts
@@ -40,6 +40,7 @@ import {
   collectSourceFiles,
   loadTsconfigPaths,
   readJsoncFile,
+  stripJsonComments,
   resolveImportToRelative,
   resolvePathAlias,
   toPosixPath,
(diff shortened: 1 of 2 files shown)


leslie-heinzen (approved):
LGTM :shipit: 

No blocking concerns. The fix is correct, well tested, and fully reversible.

### Non-blocking notes

1. The `tolerates an unterminated comment rather than throwing` test only pins the case where tolerance changes nothing, since the comment starts after the object already closed. The new code does shift from throwing to quietly truncating on malformed input: `{"a":1} /* unterminated` threw before and now parses. Consider either adding a mid-document case that asserts `JSON.parse` still rejects the result, or renaming the test to say "trailing comment after a complete document" so it is not read as a broader promise.

2. `tsconfig.cli.json` is not read through `readJsoncFile` anywhere except the new test. The real build runs `tsc -p tsconfig.cli.json`, so the compiler parses it. That makes `keeps recursive globs intact` a check on a path production does not use, and it will fail if someone legitimately edits that `include` array. The hermetic `stripJsonComments` case already covers the same string. Minor, but a one-line comment explaining the coupling would help whoever triages it later.

3. `jsonc-parser` is already a dependency and exports `stripComments`, which I confirmed is a byte-identical drop-in here. I am not asking you to switch. Three independent checks came down against it: the existing `jsonc-parser` use in `json-position.ts` maps AST offsets for Monaco markers rather than stripping comments, so there is no second copy of this logic to keep in sync, and `parse()` recovers from errors, which would loosen tsconfig validation unless the caller also wired up error checking. A short comment saying why this module stays dependency-free would save the next reviewer from asking the same question.

leslie-heinzen:
@Jayclifford345 noticed this one in the "needs approval" swim lane on your new board. I approved it before the board existed, so I moved it to the "approved" swim lane. 

Will let you handle merging still. 👍🏻

Issue #1568, opened 2026-08-10
Verify the inbound X-Grafana-Id against JWKS before trusting its namespace (OBO fails open)


moxious:
Filed during triage of the https://github.com/grafana/grafana-pathfinder-app/pull/1408 follow-up set. This is the **code half** of https://github.com/grafana/grafana-pathfinder-app/issues/1563 (which corrects the doc prose describing the same boundary). Today JWKS verification is named in `docs/developer/CODA.md` as future hardening with no tracked issue behind it.

## Summary

The App Platform proxies trust the inbound `X-Grafana-Id` header without verifying its signature. `parseIDToken` (`pkg/plugin/app_platform_identity.go:45-73`) validates only JWT _structure_ and `exp` — no signature check, no issuer check, no audience check. The parsed namespace then selects the tenant the OBO exchange mints an access token for.

## Why the boundary is weaker than the docs imply

Verified against grafana/grafana source during triage of https://github.com/grafana/grafana-pathfinder-app/issues/1563:

- `ForwardIDMiddleware.applyToken` calls `req.SetHTTPHeader(...)` only when `requester.GetIDToken() != ""`. It never calls `DeleteHTTPHeader` — so it **overwrites**, it does not strip.
- `pkg/api/plugin_resource.go` builds `CallResourceRequest{Headers: req.Header}` from the raw incoming HTTP request _before_ any middleware runs.
- `ClearAuthHeadersMiddleware`'s strip-list (`GetAuthHTTPHeaders`: `Authorization`, `X-Grafana-Device-Id`, plus configured JWT / auth-proxy header names) never includes `X-Grafana-Id`.

So a client-supplied `X-Grafana-Id` survives to the plugin on any request whose authenticated requester has no ID token — API-key and service-account auth being the realistic cases.

## Why it matters: the design fails open

Combined with structure-only inbound validation, the OBO design fails **open** rather than closed: if the inbound trust boundary were ever broken, a forged identity yields a real minted access token that stays valid for up to its configured lifetime (600s) before anything catches it. That window is the concrete argument for local verification, and it's the reason this shouldn't stay a doc footnote.

## Severity

**Medium.** Not a demonstrated exploit — the vector requires an auth mode where the requester has no ID token, and the mint fails closed on an empty ID token. But it is the single load-bearing trust assumption under every App Platform proxy route (custom-guide catalogue _and_ completion records), and it is currently unenforced in code while the docs describe it as airtight.

## Implementation

[...]

PR #1604, opened 2026-08-12
fix(backend): verify the inbound Grafana ID token against the stack's JWKS

moxious:
## Intent

Close PR 1604's JWKS revocation blocker by bounding the App-level ID-token verifier lifetime to five minutes, while retaining authlib key caching within that window and preserving immediate unknown-kid refresh behavior. Add a deterministic same-App-URL rotation test using the existing timeNow seam that proves an old key remains accepted only inside the documented refresh window, is rejected at expiry, and the replacement key is accepted. Update the backend proxy contract, canonical Coda trust-boundary documentation, and agent-facing backend file maps to document the bounded revocation window and the new auth verifier owner. Validate the committed change and update the existing PR 1604 branch; do not create a separate PR.

## What Changed

- Replaced the App Platform proxies' structural ID-token check with cryptographic verification: a new `pkg/plugin/auth/id_token.go` verifier fetches the stack's JWKS from `{appURL}/api/signing-keys/keys` (bounded 5s HTTP timeout, key fetch detached from caller cancellation), checks signature/`typ`/`exp`, rejects `exp`-less tokens, and returns `sub` verbatim. The App caches one verifier per app URL for at most 5 minutes so a key removed from the live JWKS cannot stay trusted indefinitely, while authlib's key cache — including its immediate re-fetch on an unknown `kid`, which merges into rather than replaces the cached set — still applies inside that window.
- Split the identity gate's boolean into an `identityStatus` shared by all three proxy routes: verified, rejected, unverifiable (no resolvable app URL → new `identity-unverifiable` capability reason), and signing-keys-down (retryable → §7's 503 + `Retry-After` via new `writeCompletionUnavailable` / `writeCustomGuideUnavailable` helpers, not a soft-200 envelope). `deriveCompletionUserID`, `validIDToken`, and `subjectFromIDToken` became `App` methods and now return that status.
- Documented the bounded revocation window and the new `pkg/plugin/auth` verifier owner in `BACKEND_PROXY_PATTERN.md`, `CODA.md`, and the agent-facing backend file maps in `.cursor/rules/`; extended the Go test suites with same-app-URL rotation coverage over the `timeNow` seam (old key accepted inside the window, rejected at expiry, replacement key accepted) plus signing-keys-URL and detached-fetch tests, and refreshed the frontend capability-reason docs and a stale line-number reference in `backend-api.schema.ts`.

Review left four informational notes, all deliberate or doc-level: an unreachable `cfg == nil` guard in `idTokenVerifier`, a `kid` negative-caching nuance in `CODA.md` (the document phase committed a clarifying follow-up), the bounded inline JWKS fetch that authlib's shape implies (matching Grafana core's own ExtendedJWT client), and the intentional identity-before-backend-resolution ordering that shifts telemetry vocabulary on stacks where the aggregation toggle is off.

## Risk Assessment

✅ Low: The change is a well-bounded security hardening of one shared identity gate: it satisfies every source-verifiable clause of the stated intent, is covered by a deterministic rotation test plus a full token-rejection matrix, keeps every failure mode fail-closed, and the only substantiated findings are informational (dead defensive branch, one slightly optimistic doc sentence, and two acknowledged tradeoffs of the new inline JWKS dependency).

## Testing

I exercised the real GET /completion-records/my handler across a same-app-URL signing-key rotation with the timeNow seam frozen and advanced, capturing the HTTP status, Retry-After, and JSON body a browser would receive at each step: genuine token serves data, a forged token with the same kid and sub is refused with a soft-200 identity-unavailable envelope, the retired key is still accepted at t+1m and t+4m59s, is rejected at exactly t+5m, the rotated key is accepted, and an unreachable JWKS returns 503 with Retry-After; the transcript's fetch counter also shows key caching survives inside the window. Alongside that I ran the committed identity, verifier, and auth-package tests plus go build for the backend, added a focused test for the documented unknown-kid immediate-refresh behavior that nothing covered, and mutation-checked the bound by deleting the max-age condition, which makes the committed rotation test fail as expected. No screenshot is included because the change is backend-only (the frontend diff is comment text), so the reviewer-visible surface is the proxy's HTTP responses, captured as the transcript artifact.

<details>
<summary>Evidence: HTTP transcript: JWKS revocation window on GET /completion-records/my</summary>

```text
=== RUN   TestEvidence_JWKSRevocationWindow

### 1. genuine token, key published by the stack → data served
t+0s      GET /api/plugins/grafana-pathfinder-app/resources/completion-records/my
          X-Grafana-Id: eyJhbGciOiJFUzI1Ni…SyqzK3wA
HTTP 200
{
  "asOf": "2023-11-14T22:13:20Z",
  "capability": {
    "available": true
  },
  "completions": [
    {
      "count": 1,
      "guideCategory": "interactive",
      "guideId": "loki",
      "guideSource": "bundled",
      "guideTitle": "Get started with Loki",
      "latestCompletedAt": "2026-08-01T10:00:00Z",
      "latestSource": "objectives",
      "maxCompletionPercent": 100,
      "pathId": ""
    }
  ],
  "userId": "user:1"
}
(JWKS fetches so far: 1)

### 2. forged token: same kid, attacker's key, same sub → refused (#1568)
t+0s      GET /api/plugins/grafana-pathfinder-app/resources/completion-records/my
          X-Grafana-Id: eyJhbGciOiJFUzI1Ni…2D2q4-1A
HTTP 200
{
  "capability": {
    "available": false,
    "reason": "identity-unavailable"
  },
  "completions": []
}
(JWKS fetches so far: 1)

--- operator rotates JWKS: stack-key-1 removed, stack-key-2 published ---

### 3. retired key, inside the 5-minute window → still accepted (cached keys)
t+1m0s    GET /api/plugins/grafana-pathfinder-app/resources/completion-records/my
          X-Grafana-Id: eyJhbGciOiJFUzI1Ni…SyqzK3wA
HTTP 200
{
  "asOf": "2023-11-14T22:13:20Z",
  "capability": {
    "available": true
  },
  "completions": [
    {
      "count": 1,
      "guideCategory": "interactive",
      "guideId": "loki",
      "guideSource": "bundled",
      "guideTitle": "Get started with Loki",
      "latestCompletedAt": "2026-08-01T10:00:00Z",
      "latestSource": "objectives",
      "maxCompletionPercent": 100,
      "pathId": ""
    }
  ],
  "userId": "user:1"
}
(JWKS fetches so far: 1)

### 4. brand-new key, same window → accepted immediately (unknown-kid refetch)
t+1m0s    GET /api/plugins/grafana-pathfinder-app/resources/completion-records/my
          X-Grafana-Id: eyJhbGciOiJFUzI1Ni…jVjmtscA
HTTP 200
{
  "asOf": "2023-11-14T22:13:20Z",
  "capability": {
    "available": true
  },
  "completions": [
    {
      "count": 1,
      "guideCategory": "interactive",
      "guideId": "loki",
      "guideSource": "bundled",
      "guideTitle": "Get started with Loki",
      "latestCompletedAt": "2026-08-01T10:00:00Z",
      "latestSource": "objectives",
      "maxCompletionPercent": 100,
      "pathId": ""
    }
  ],
  "userId": "user:1"
}
(JWKS fetches so far: 2)

### 5. retired key, one second before the bound → still accepted
t+4m59s   GET /api/plugins/grafana-pathfinder-app/resources/completion-records/my
          X-Grafana-Id: eyJhbGciOiJFUzI1Ni…SyqzK3wA
HTTP 200
{
  "asOf": "2023-11-14T22:13:20Z",
  "capability": {
    "available": true
  },
  "completions": [
    {
      "count": 1,
      "guideCategory": "interactive",
      "guideId": "loki",
      "guideSource": "bundled",
      "guideTitle": "Get started with Loki",
      "latestCompletedAt": "2026-08-01T10:00:00Z",
      "latestSource": "objectives",
      "maxCompletionPercent": 100,
      "pathId": ""
    }
  ],
  "userId": "user:1"
}
(JWKS fetches so far: 2)

### 6. retired key, at the 5-minute bound → revoked, no data
t+5m0s    GET /api/plugins/grafana-pathfinder-app/resources/completion-records/my
          X-Grafana-Id: eyJhbGciOiJFUzI1Ni…SyqzK3wA
HTTP 200
{
  "capability": {
    "available": false,
    "reason": "identity-unavailable"
  },
  "completions": []
}
(JWKS fetches so far: 3)

### 7. rotated key, after the rebuild → data served
t+5m0s    GET /api/plugins/grafana-pathfinder-app/resources/completion-records/my
          X-Grafana-Id: eyJhbGciOiJFUzI1Ni…jVjmtscA
HTTP 200
{
  "asOf": "2023-11-14T22:18:20Z",
  "capability": {
    "available": true
  },
  "completions": [
    {
      "count": 1,
      "guideCategory": "interactive",
      "guideId": "loki",
      "guideSource": "bundled",
      "guideTitle": "Get started with Loki",
      "latestCompletedAt": "2026-08-01T10:00:00Z",
      "latestSource": "objectives",
      "maxCompletionPercent": 100,
      "pathId": ""
    }
  ],
  "userId": "user:1"
}
(JWKS fetches so far: 3)

### 8. JWKS unreachable → transient 503 + Retry-After, not a capability envelope
t+5m0s    GET /api/plugins/grafana-pathfinder-app/resources/completion-records/my
          X-Grafana-Id: eyJhbGciOiJFUzI1Ni…jVjmtscA
HTTP 503
Retry-After: 30
{
  "error": "completion-records-unavailable"
}
(JWKS fetches so far: 3)
--- PASS: TestEvidence_JWKSRevocationWindow (0.00s)
PASS
ok  	github.com/grafana/grafana-pathfinder-app/pkg/plugin	0.302s
```
</details>
<details>
<summary>Evidence: Revocation-window transcript (key steps, abridged)</summary>

```text
### 2. forged token: same kid, attacker's key, same sub → refused (#1568)
t+0s GET /api/plugins/grafana-pathfinder-app/resources/completion-records/my
HTTP 200
{ "capability": { "available": false, "reason": "identity-unavailable" }, "completions": [] }
(JWKS fetches so far: 1)

--- operator rotates JWKS: stack-key-1 removed, stack-key-2 published ---

### 4. brand-new key, same window → accepted immediately (unknown-kid refetch)
t+1m0s HTTP 200 capability.available=true userId=user:1 (JWKS fetches so far: 2)

### 5. retired key, one second before the bound → still accepted
t+4m59s HTTP 200 capability.available=true userId=user:1 (JWKS fetches so far: 2)

### 6. retired key, at the 5-minute bound → revoked, no data
t+5m0s HTTP 200
{ "capability": { "available": false, "reason": "identity-unavailable" }, "completions": [] }
(JWKS fetches so far: 3)

### 7. rotated key, after the rebuild → data served
t+5m0s HTTP 200 capability.available=true userId=user:1

### 8. JWKS unreachable → transient 503 + Retry-After, not a capability envelope
t+5m0s HTTP 503
Retry-After: 30
{ "error": "completion-records-unavailable" }
```
</details>
<details>
<summary>Evidence: Mutation check: the 5-minute bound is load-bearing</summary>

```text
$ # temporarily dropped the max-age condition from idTokenVerifier()
$ go test ./pkg/plugin -run 'TestVerifyIDToken_KeySetRefreshBoundsRetiredKeyTrust' -count=1
--- FAIL: TestVerifyIDToken_KeySetRefreshBoundsRetiredKeyTrust (0.00s)
app_platform_identity_test.go:493: retired token after refresh: status = 0, want rejected
FAIL
$ # source restored; git diff clean for pkg/plugin/app_platform_identity.go
```
</details>

## Pipeline

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>
<summary>⚠️ **Review** - 4 infos</summary>

- ℹ️ `pkg/plugin/app_platform_identity.go:131` - The `cfg == nil` guard in `idTokenVerifier` is unreachable: `config.GrafanaConfigFromContext` (plugin-SDK v0.296.1, config/grafana_cfg.go:39) returns `NewGrafanaCfg(nil)` for both a missing context value and a stored nil, never a nil pointer. The `TestVerifyIDToken_UnverifiableFailsClosed` subcase named &#34;no grafana config on context&#34; therefore passes through the `cfg.AppURL()` error branch, not the branch its name implies. Dropping the nil check (and renaming that subcase to what it actually covers) removes dead code without changing behavior — `identityUnverifiable` is still returned via the `AppURL()` error path.
- ℹ️ `docs/developer/CODA.md:195` - &#34;Unknown `kid`s still trigger authlib&#39;s immediate re-fetch, so a newly published key need not wait for that interval&#34; holds only for the first request carrying that `kid`. `DefaultKeyRetriever.Get` calls `setEmptyCacheItem` when a `kid` is still absent after the refetch, so every later token with the same `kid` is rejected with no refetch until the verifier is rebuilt. Concretely: if a token signed with rotated-in key K2 reaches the plugin in the window before K2 appears in the stack&#39;s JWKS, K2 is negative-cached and all K2 tokens fail closed for the remainder of that verifier&#39;s ≤5-minute lifetime, not &#34;immediately&#34; on the next request. The bound is still five minutes, so only the doc sentence needs tightening (e.g. &#34;the first unknown `kid` triggers a re-fetch; if the key is not published yet it is negative-cached until the next rebuild&#34;).
- ℹ️ `pkg/plugin/auth/id_token.go:82` - The identity gate now performs an inline JWKS fetch before any signature is checked: authlib&#39;s `Verify` order is parse → `typ` → `getKeyID` → `keys.Get`, and `keys.Get` fetches on a cache miss. A caller who can set `X-Grafana-Id` (the same narrow population this branch&#39;s own CODA.md note describes — an authenticated requester whose ID token `ForwardIDMiddleware` does not overwrite) can therefore drive one HTTP GET to `{appURL}/api/signing-keys/keys` plus one negative-cache entry keyed by an attacker-chosen `kid` string per request, using tokens with no valid signature at all. Repeats of the same `kid` are free (negative cache), and the whole cache is discarded on each 5-minute rebuild, so growth is bounded by ~5 minutes of distinct-`kid` traffic against a local, unauthenticated endpoint. Noting the bound rather than recommending a change: this is authlib&#39;s shape, matching Grafana core&#39;s own ExtendedJWT client, and mitigating it would mean adding a rate limit or `kid` shape check that is a product decision, not a review fix.
- ℹ️ `pkg/plugin/custom_guide_repository.go:121` - The identity gate runs before `resolveCustomGuideBackend`/`resolveCompletionBackend`, so stacks where the aggregation feature toggle is off — i.e. every stack the feature is not rolled out to — now resolve signing keys and can answer a JWKS blip with a 503 instead of the stable `feature-toggle-disabled` soft-200 envelope. User-visible impact is nil (the catalogue client returns `[]` on both paths), but the telemetry vocabulary shifts: `reportCatalogueFetchFailure` records `http-503` instead of `recordCustomGuideCatalogueUnavailable(&#39;feature-toggle-disabled&#39;)`, so a signing-keys blip on an un-enabled stack reads as a proxy failure rather than &#34;not rolled out&#34;. Deliberate ordering (identity before anything that could leak reachability detail), so flagging for awareness only.

</details>

<details>

✅ No issues found.
- `go test ./pkg/plugin -run 'TestVerifyIDToken|TestValidIDToken|TestIdentityStatus|TestIdentityGate|TestDeriveCompletionUserID' -v`
- `go test ./pkg/plugin/auth/... -run 'TestSigningKeysURL|TestVerify_KeyFetchDetachedFromCallerCancellation|TestSigningKeysUnavailable' -v`
- <code>`go test ./pkg/plugin/... -count=1` (backend packages only)</code>
- `go build ./...`
- <code>Manual end-to-end HTTP transcript: temporary harness driving `app.handleMyCompletions` through genuine / forged / retired-key / rotated-key / JWKS-down cases against a live mutable JWKS server, frozen clock advanced across `idTokenVerifierMaxAge` (harness removed after capture)</code>
- <code>Mutation check: removed the `now.Before(idVerifierCreatedAt + idTokenVerifierMaxAge)` condition and confirmed `TestVerifyIDToken_KeySetRefreshBoundsRetiredKeyTrust` fails, then restored the file</code>
- <code>Added and ran `go test ./pkg/plugin -run TestVerifyIDToken_NewKeyAcceptedMidWindow -v` (new focused test for unknown-kid immediate refresh inside the window)</code>
- <code>`gh pr view 1604 --json headRefName` to confirm PR 1604 already tracks `fix/1568-jwks-id-token-verification` (no separate PR needed)</code>

</details>

<details>
<summary>🔧 **Document** - 1 issue found → auto-fixed ✅</summary>

- ℹ️ `docs/developer/CODA.md:195` - Judgment call left to the author: CODA.md says &#34;Unknown `kid`s still trigger authlib&#39;s immediate re-fetch, so a newly published key need not wait for that interval.&#34; A reader could infer that re-fetch also replaces the cached key set and therefore revokes a retired key immediately, which would contradict the preceding five-minute bound. The uncommitted `TestVerifyIDToken_NewKeyAcceptedMidWindow` asserts the opposite (the re-fetch merges into the cached set, so the retired key still verifies mid-window). Adding a clause like &#34;that re-fetch merges into the cached set rather than replacing it, so it never shortens the window&#34; would close the misreading. I did not make the edit because confirming the additive-merge behaviour requires running that Go test, which this pass is not permitted to do; the sentence as written is not incorrect.

🔧 Fix: clarify JWKS re-fetch merges into cached key set
✅ Re-checked - no issues remain.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>


moxious:
@cursoragent review

docs/developer/CODA.md
```diff
@@ -156,14 +156,48 @@ All routes are prefixed by Grafana as `/api/plugins/grafana-pathfinder-app/resou
...
+authenticate an identity), and `exp`/`nbf` with go-jose's one-minute leeway. `exp` **presence** is
+additionally required here, because go-jose validates expiry only when the claim is present and an
+`exp`-less token would otherwise verify as non-expiring. The `sub` claim is extracted verbatim
+only on routes that serve per-user data (`pkg/plugin/app_platform_identity.go`).
+
+Because the signature is checked, none of this depends on Grafana's server→plugin forwarding to
+keep the header honest — which matters, since `X-Grafana-Id` is **not** on
+`ClearAuthHeadersMiddleware`'s strip-list and `ForwardIDMiddleware` overwrites rather than
+deletes, so a client-set value can survive to the plugin whenever the authenticated requester has
+no ID token of its own.
```
tomglenn: **Must fix — this overstates what verification buys.**

A checked signature proves the token was *issued* by this stack. It does not prove the caller presenting it is the subject. A copied, still-unexpired token replayed on `/completion-records/my` verifies and returns the victim's records, because nothing binds the token to the requester.

The replay exposure itself is pre-existing (it was reachable before this branch too, alongside forgery), so it isn't yours to close here. But this sentence is new, and as written it tells the next reader that forwarding is no longer load-bearing — which is exactly the belief that would let someone skip an auth check on a new proxy route. The old wording was honest about depending on forwarding; the new wording drops that without replacing it.

Worth noting you already handle this well for the namespace claim, which is explicitly flagged as tracked separately. Same candour applied here fixes it:

```suggestion
Because the signature is checked, the header's *authenticity* no longer depends on Grafana's
server→plugin forwarding — which matters, since `X-Grafana-Id` is **not** on
`ClearAuthHeadersMiddleware`'s strip-list and `ForwardIDMiddleware` overwrites rather than
deletes, so a client-set value can survive to the plugin whenever the authenticated requester has
no ID token of its own. Verification proves the token was **issued** by this stack, not that the
caller presenting it is its subject: a copied, still-unexpired token replayed on a per-user route
still verifies. Binding the token to its presenter is tracked separately.
```

docs/developer/CODA.md
```diff
@@ -156,14 +156,48 @@ All routes are prefixed by Grafana as `/api/plugins/grafana-pathfinder-app/resou
...
+  dark past the end of the outage.
+
+All three routes take that decision from one shared `identityStatus`
+(`pkg/plugin/app_platform_identity.go`), so they cannot classify a failure differently.
+
+Authlib caches fetched keys for the lifetime of one verifier. Pathfinder reuses that verifier for
+at most **five minutes**, then rebuilds it against the same signing-keys URL, so steady-state
+verification avoids per-request network calls while a key removed from JWKS remains trusted for
+no more than five minutes. Unknown `kid`s still trigger authlib's immediate re-fetch, so a newly
+published key need not wait for that interval; that re-fetch merges into the current verifier's
```
tomglenn: **Must fix — factually wrong, and it's the one your own pipeline flagged.**

"Unknown `kid`s still trigger authlib's immediate re-fetch" holds only for the **first** request bearing that `kid`. When the key is still absent after that re-fetch, authlib calls `setEmptyCacheItem`, and `getCachedItem` then returns that empty entry with `ok=true` — so every later request with the same `kid` is rejected with **no** re-fetch, until your five-minute rebuild clears it.

Demonstrated: publish key A → send a token signed by unpublished key B (rejected, fetch count 1→2) → publish B → send the same token five more times → **zero** further fetches, still rejected. A fresh verifier accepts B immediately.

The informational note in your PR body describes this accurately, and says a clarifying commit handled it — but `c4f5e91a` corrected the adjacent merge-vs-replace clause instead. That clause is right, and it stays right; this one is the one still standing.

```suggestion
no more than five minutes. The **first** request bearing an unknown `kid` triggers an immediate
re-fetch; if the key is still unpublished, authlib negative-caches that `kid` and later requests
carrying it are rejected without re-fetching until the rebuild. That re-fetch merges into the current verifier's
```

**Please don't "fix" this in code** — see the note in the review summary. The two obvious code fixes are both worse than the current behaviour, and one of them is impossible.

docs/developer/CODA.md
```diff
@@ -173,10 +207,11 @@ reach the plugin), and never the ID token itself, which no hop on the outbound p
...
 
-The single future-hardening item is cryptographic verification of the ID token against
-Grafana's JWKS via `github.com/grafana/authlib`; it is not wired today because it needs runtime
-key-endpoint configuration. Do not re-argue this trade-off per PR — this section is the
-canonical statement for all App Platform proxies.
+Do not re-argue any of this per PR — this section is the canonical statement for all App Platform
+proxies. One deliberate omission: `aud` is not validated, because an ID token's audience is
+`org:<orgID>`, which tells a plugin nothing it can act on. This mirrors Grafana's own ExtendedJWT
+client. Binding the token's `namespace` claim to the plugin-context namespace is tracked
+separately and is not required for the signature to make `sub` unforgeable.
```
tomglenn: Same root as the comment on line 171 — `unforgeable` is carrying more weight than the signature actually delivers. Replay isn't forgery, so `sub` being unforgeable doesn't make it bound to the caller.

```suggestion
separately and is not required for the signature to make `sub` unforgeable — though unforgeable is
not the same as bound to the presenter; see the replay caveat above.
```

pkg/plugin/auth/id_token.go
```diff
@@ -0,0 +1,106 @@
...
+// ExtendedJWT client (grafana/pkg/services/authn/clients/ext_jwt.go).
+func NewIDTokenVerifier(appURL string) (*IDTokenVerifier, error) {
+	keysURL, err := signingKeysURL(appURL)
+	if err != nil {
+		return nil, err
+	}
+
+	keys := authn.NewKeyRetriever(
+		authn.KeyRetrieverConfig{SigningKeysURL: keysURL},
+		authn.WithHTTPClientKeyRetrieverOpt(&http.Client{Timeout: signingKeysFetchTimeout}),
```
tomglenn: **Should fix — the JWKS fetch follows redirects anywhere, including off-host.**

This client sets `Timeout` but no `CheckRedirect`, so Go's default follows up to 10 redirects. If the stack's own signing-keys path ever redirects off-origin, the key set is taken from wherever it lands. Demonstrated: a 302 to an attacker origin serving `{"keys":[{"kid":"rogue-kid",...}]}` → a token signed by the rogue key verifies and returns `sub="user:victim"`.

Calibrating the severity honestly, because I initially had this higher than it deserves: the URL has **no** caller-controlled input (operator config plus a compile-time constant path, no query string), so there's no open-redirect gadget to point at it. Exploiting it means already owning the front door, proxy, or DNS — at which point the on-behalf-of access token is equally exposed. And before this branch there was no signature check at all, so forgery was free and unconditional; this is strictly a large improvement either way.

Raising it because the surface is new here and the fix is small and local. Same-*hostname* rather than same-origin, since I confirmed a legitimate http→https redirect is followed today and changes host:port — a naive same-origin rule would break those deployments:

```suggestion
		authn.WithHTTPClientKeyRetrieverOpt(&http.Client{
			Timeout: signingKeysFetchTimeout,
			CheckRedirect: func(req *http.Request, via []*http.Request) error {
				if req.URL.Hostname() != via[0].URL.Hostname() {
					return fmt.Errorf("signing-keys redirect left the stack host: %s", req.URL.Hostname())
				}
				return nil
			},
		}),
```

pkg/plugin/app_platform_identity.go
```diff
@@ -1,82 +1,157 @@
...
+
+// identityStatus is the verdict of the identity gate. Three distinct failures,
+// because BACKEND_PROXY_PATTERN.md §7's transient/structural split cuts across
+// them: only one of the three is retryable, and reporting a retryable outage as
+// a standing condition darkens the surface past the end of the outage.
+type identityStatus int
+
+const (
+	// identityVerified: the caller carries a cryptographically verified token.
+	identityVerified identityStatus = iota
```
tomglenn: **Should fix — `identityVerified` is the iota zero value, so an unset status reads as authenticated.**

No production path hits this: every return is explicit, and all three route switches deny via `default:`. But it isn't purely hypothetical either — your own test tables already rely on it. Four cases across two tables express "verified" by *omitting* `wantStatus`, so `{name: "verified token", token: ...}` asserts success only by accident of the zero value. A future case that forgets the field silently asserts "authenticated" instead of failing.

Making the zero value an explicit non-verdict fixes both the latent fail-open and the test-table opacity:

```suggestion
	// identityUnknown is the zero value, so a status left unset can never be
	// mistaken for a verified caller.
	identityUnknown identityStatus = iota

	// identityVerified: the caller carries a cryptographically verified token.
	identityVerified
```

This is not a one-liner — it breaks four tests, and a complete green fix measured at roughly 14 insertions / 9 deletions, mostly making those tables state their expected status explicitly. `capabilityReason()`'s `default:` arm returning `""` for an unrecognised status is worth a look in the same pass.

pkg/plugin/app_platform_identity.go
```diff
@@ -1,82 +1,157 @@
...
-		Sub string `json:"sub"`
-		Exp int64  `json:"exp"`
+// idTokenVerifier returns this stack's ID-token verifier, building it on first
+// use. The signing-keys URL derives from the per-request Grafana config, so the
+// verifier cannot be built in NewApp. It is reused briefly so authlib's key
+// cache is shared across requests, then rebuilt to bound how long a key removed
+// from the live JWKS remains trusted.
+func (a *App) idTokenVerifier(ctx context.Context) (*auth.IDTokenVerifier, error) {
+	cfg := config.GrafanaConfigFromContext(ctx)
+	if cfg == nil {
```
tomglenn: Minor: this branch is unreachable, and the docs describe it as a real cause.

`config.GrafanaConfigFromContext` returns `NewGrafanaCfg(nil)` for both a missing context value and a stored nil — never a nil pointer. So `cfg == nil` never fires, and the `TestVerifyIDToken_UnverifiableFailsClosed` subcase named "no grafana config on context" actually exercises the `cfg.AppURL()` error path instead.

Your PR body already identifies this as informational note 1, so you may have decided to keep it as defensive depth — fine either way, but two things follow if it stays: the subcase name should say what it really covers, and `CODA.md:181` plus the matching line in the backend-proxy pattern doc currently list "no Grafana config on the request" as a genuine cause of the unverifiable outcome, which reads as reachable. Worth reconciling one way or the other so code and prose agree.

src/lib/custom-guide-repository-client.ts
```diff
@@ -44,10 +44,13 @@ export interface CustomGuideRepositoryEntry {
...
  * a machine `reason` when the proxy can't serve (the response is still a
  * soft-200 in that case). Reasons: `identity-unavailable`,
- * `grafana-config-unavailable`, `feature-toggle-disabled`, `namespace-unavailable`,
- * `app-url-unavailable`, `obo-unavailable` (no provisioned on-behalf-of token —
- * check this first when the surface is unexpectedly empty), `backend-unavailable`,
- * or `upstream-<status>` for an upstream error.
+ * `identity-unverifiable` (the stack exposes no ID-token signing keys to verify
+ * against; an unreachable one is a transient 503 instead),
+ * `grafana-config-unavailable`, `feature-toggle-disabled`,
+ * `namespace-unavailable`, `app-url-unavailable`, `obo-unavailable` (no
```
tomglenn: Minor: `app-url-unavailable` is no longer reachable, so documenting it here sends the next debugger down a dead path.

The identity gate now runs before `resolveCustomGuideBackend` and consumes the no-app-URL case itself, so that condition surfaces as `identity-unverifiable`. Your own `custom_guide_repository_test.go:361` comment concedes the reason is now unreachable.

```suggestion
 * `namespace-unavailable`, `obo-unavailable` (no
```

(`grafana-config-unavailable` a line above is dead too, but that predates this branch — leave it or sweep it, your call.)

pkg/plugin/auth/id_token.go
```diff
@@ -0,0 +1,106 @@
...
+// stack that issues the ID token also serves the key that verifies it.
+const SigningKeysPath = "/api/signing-keys/keys"
+
+// signingKeysFetchTimeout bounds one JWKS fetch. authlib's key retriever
+// defaults to http.DefaultClient, which has no timeout, and the fetch runs
+// inline in the identity gate of every proxy route. It also bounds Verify's
+// detached context, which is a different thing: the client timeout covers one
+// HTTP round trip, the deadline covers the whole wait including another
+// caller's in-flight fetch.
+const signingKeysFetchTimeout = 5 * time.Second
```
tomglenn, line 22: Comment density in this file is around 40% (roughly 42 of 107 lines). Several earn their place and should stay: `ErrMissingExpiry`, the reuse-but-replace invariant on the struct, `signingKeysURL`'s doubled-slash warning, and `Verify` returning `("", nil)` as success. Those are all real invariants a reader can't recover from the code.

A few are the shapes the comment rule asks us to drop:

- This block's "It also bounds Verify's detached context, which is a different thing…" argues a distinction in-line, and the same point is made again at lines 78-81.
- `SigningKeysUnavailable` carries eight comment lines above a one-line function, mostly justifying the classification choice.
- `NewIDTokenVerifier`'s comment restates the constructor and then repeats the audience paragraph verbatim from `CODA.md`, where it already lives.
- In `app_platform_identity.go`, "which is exactly the attack #1568 closed" is a ticket reference that the commit message already carries, and the file header grew from 12 to 16 lines of design argument.

Design rationale of that kind reads better in the PR body than in the source, since it stops being true silently.

tomglenn (changes requested):
Requesting changes, but the core of this is sound and I want to be clear about that up front: I tried hard to break the verification and could not. Attempts that failed include `alg: none`, an empty signature, HS256 signed with the marshalled public key as the HMAC secret, an ES384 attacker key, a correct `kid` with the wrong key, self-supplied `jwk` / `jku` / `x5u` header injection, a missing `kid`, `at+jwt` type confusion, a missing `exp`, a future `nbf`, and duplicate `sub` keys. Pinning ES256 at parse kills that whole class. Cross-stack acceptance is inert by construction, since both the signing-keys URL and the LIST namespace come from plugin context rather than the token. Verifier-cache poisoning does not work either — the map key is the exact app URL string that the fetch derives from, so two origins cannot share a verifier. The forged-`sub` hole from the linked issue is genuinely closed.

Concurrency and resources hold up too: 64 goroutines by 20 verifications across forced rebuilds produced no races and exactly one fetch per window, the mutex is not held across the fetch, the five-minute bound is frozen at construction rather than sliding, and 300 rebuilds leaked no goroutines. Twelve targeted mutations were caught by ten of your tests, with messages that name the right cause — including a total auth bypass, which tripped fifteen. Your slash-safe URL join has a strict exact-string test behind it, which is more than the competing branch could say.

Everything below is edges around that core. The two doc items are the ones I would genuinely block on, because they are sentences this branch wrote that will mislead the next reader.

## Must fix

1. **The trust-boundary prose now overstates the guarantee.** A checked signature proves issuance, not possession — a copied unexpired token replayed on a per-user route still verifies. The replay exposure predates this branch, so it is not yours to close; the sentence asserting that forwarding is no longer load-bearing is new, and that is what needs the caveat. Inline on `CODA.md:171` and `:214`.

2. **The unknown-`kid` sentence is wrong.** Immediate re-fetch applies only to the first request bearing that `kid`; after that authlib negative-caches it and later requests are rejected with no fetch until the rebuild. Reproduced with fetch counts. Your PR body describes this correctly as informational note 2 and says a follow-up commit handled it, but that commit corrected the adjacent merge-versus-replace clause instead. Inline on `CODA.md:195`.

**Please fix this one in prose only.** Both obvious code fixes are worse than what you have. Reclassifying unknown-`kid` as transient would hand a genuinely forged token a 503 plus `Retry-After`, inverting your deliberate narrow classifier. Injecting a shorter-TTL cache cannot work at all — `setEmptyCacheItem` passes its ten-minute TTL explicitly per `Set`, and `LocalCache.Set` forwards it, so the config value is ignored. The only variant that does work is a cache that drops empty entries, and that turns one repeated bad `kid` from a single fetch into fifty. The rotation fix and the fetch-volume fix are mutually exclusive; resist doing both.

For what it is worth, the production severity here is low. Grafana commits the key insert and invalidates the JWKS cache before returning the signer, so ordinary and month-boundary rotation are race-free. The precondition is a Grafana core bug — `GetJWKS` reads a remote cache with a twelve-hour TTL and no lock or singleflight, so an interleaved read-then-write can serve a stale key set for hours, with invalidation failures swallowed at debug level. During that window verification fails regardless of the negative cache, so this branch's marginal contribution is capped at five minutes. I will raise that upstream separately; it is not this repo's to fix.

## Should fix

3. **The JWKS fetch follows redirects off-host.** No `CheckRedirect` on the client, so a redirect away from the stack's origin hands over the key set. Demonstrated end to end. Severity is genuinely modest — the URL has no caller-controlled input, so there is no gadget to aim at it, and exploitation implies already owning the front door — but the surface is new here and the fix is small. Use same-hostname, not same-origin: a legitimate http-to-https redirect changes host:port and a naive rule breaks it. Inline on `id_token.go:58` with a patch.

4. **`identityVerified` is the iota zero value.** No live path returns it implicitly, but your own test tables already depend on it — four cases across two tables express "verified" by omitting `wantStatus`. Making the zero value an explicit non-verdict closes a latent fail-open and makes those tables state what they assert. Around 14 insertions and 9 deletions, not a one-liner. Inline on `app_platform_identity.go:42`.

## Smaller items

5. `app-url-unavailable` is now unreachable but still documented to frontend consumers in `custom-guide-repository-client.ts:50`. Inline.
6. The `cfg == nil` branch at `app_platform_identity.go:131` cannot fire, and two docs list it as a real cause. Inline.
7. Comment density in `id_token.go` is around 40%, with several blocks arguing design in-line or duplicating prose that already lives in `CODA.md`. Inline, with the keep-list called out so nothing useful gets swept.
8. Two agent-facing files still describe the old world, and neither is in this diff: the backend section of `AGENTS.md` still presents `pkg/plugin/auth` as on-behalf-of minting only, when it now also owns inbound verification; and `CONTEXT_INDEX.md:96` routes the backend-proxy pattern doc by a non-recursive `pkg/plugin/*.go` glob, which never matches the new `pkg/plugin/auth/` file. Both rules-file maps did get rows, so this is just the last two.
9. Two test gaps worth closing while you are in here: the unverifiable state is asserted at HTTP level for two routes but not for the completion-records capability route (one table row), and nothing drives the verifier cache concurrently, so `-race` passing over it is currently vacuous. Optionally pin the one-minute expiry leeway too — it is correct and matches what Grafana core accepts, and it is documented here, but it comes from a transitive dependency's default and could drift without a test noticing.

## Things I chased and want to save you from re-chasing

- **Fetch volume on novel `kid`s is not an amplification vector.** Measured factor is 1.00, and the signing-keys endpoint takes no auth, so anyone can hit it directly at the same rate without going through the plugin. Routing through it costs an attacker a valid session and buys nothing. The header also only survives client-set for a requester with no ID token of its own, which is a narrower and more privileged population than "any signed-in user".
- **The expiry leeway is fine.** One minute is the library default, Grafana core's own verifiers accept the same window, and matching the issuer is right. It is documented in this diff. Zeroing it is not even reachable through the config surface.
- **The empty-key-set and 404 classifications are both defensible.** A 404 on your own JWKS path is an operational fault, and a loud retryable 503 beats darkening silently; an empty key set only happens when no key was ever created, in which case no verifiable token exists.
- **The merge-versus-replace clause is correct** and should stay as written.

Verification run at `c4f5e91a`: `go build ./...`, `go vet ./pkg/...`, `npm run lint:go`, `npm run test:go`, and `go test -race -count=1 ./pkg/...` all pass, as do `-count=2` and `-shuffle=on`.

<sub>Reviewed with parallel security, correctness, test-quality, and documentation passes, then an adversarial round that attempted to refute each finding. Three of my own initial findings did not survive that round and were dropped rather than reported; the severity calibration on item 3 was corrected downward the same way.</sub>

tomglenn (approved):

tomglenn:
Heads-up: this doesn't work on Grafana Cloud, and it's currently breaking dev.

`{appURL}/api/signing-keys/keys` returns `{"keys":null}` on every Cloud stack we checked — `pathfinder.grafana-dev.net`, `learn.grafana.net`, `snapshots.raintank.io`, `play.grafana.org`. With an empty key set no `kid` ever resolves, authlib returns `ErrInvalidSigningKey`, and every App Platform proxy route serves `{"capability":{"available":false,"reason":"identity-unavailable"}}`.

Jay hit it on dev yesterday: `[custom-guides] catalogue unavailable {reason: 'identity-unavailable'}`. Completion records go through the same gate, so those are down there too — silently, since that path logs nothing when the header is absent.

Prod looks fine only because it's still on v2.16.0, whose gate is the old structural check. Dev and prod are both Grafana 13.3.0, so the plugin version is the only difference. #1694 would ship this everywhere.

The premise looks like the issue. `pkg/plugin/auth/id_token.go:14-17` says "the stack that issues the ID token also serves the key that verifies it" — true for self-hosted, but on Cloud the stack doesn't sign these tokens; auth-api does. Its JWKS lives at `http://api-lb.auth.svc.cluster.local/v1/keys`, described in `deployment_tools` (`ksonnet/lib/mobile-app/mobile-api.libsonnet:49-50`) as exactly "used to validate the X-Grafana-Id header". It's the same host the plugin already mints OBO tokens against, so reachability is proven by prod working today.

So the fix looks like: verify against auth-api, fall back to the stack endpoint for self-hosted, where the local signer does populate it. Full signature verification either way — nothing given up.

I'm having that built and I'd value your review on it, since it's your area. One thing worth flagging: if the host or path is wrong the fetch fails rather than returning empty, which maps to a hard 503 on every proxy route — worse than today's soft 200. So it wants a dev smoke test before it goes near ops.

One question I couldn't answer from here: does auth-api validate the subject token's signature during the OBO exchange? authlib's alternative `Subject` field needs a strictly higher scope than `SubjectToken`, which hints the token path is validated. If it is, the exposure #1568 describes is narrower than stated.

PR #1609, opened 2026-08-13
fix(backend): verify inbound Grafana ID tokens against the stack's own JWKS

### LINKED ISSUE #1568: Verify the inbound X-Grafana-Id against JWKS before trusting its namespace (OBO fails open)

tomglenn:
## Summary

`parseIDToken` in `pkg/plugin/app_platform_identity.go` only structurally decoded the inbound `X-Grafana-Id` header — split on `.`, base64-decode, check `exp` — with no signature check. A caller with no real Grafana ID token (API-key or service-account auth, where `ForwardIDMiddleware` never sets or clears the header) could forge the `sub` claim. On a warm completion-records cache hit, the handler serves `idx.byUser[<unverified sub>]` with no mint or upstream call in between, so a forged `sub` reads another user's completion records.

## What

- New `pkg/plugin/auth.IdentityVerifier`, backed by `authlib`'s `authn.NewIDTokenVerifier` + `authn.NewKeyRetriever` against the stack's own unauthenticated `{appURL}/api/signing-keys/keys`. One verifier per app URL, lazily built and cached.
- Explicit rejection of tokens with no `exp` claim at all, layered on top of authlib's check — go-jose's own validation only checks expiry when `exp` is present, so this guard has to stay.
- `validIDToken` / `subjectFromIDToken` / `deriveCompletionUserID` are now `*App` methods so they can reach `cfg.AppURL()`. `parseIDToken` / `decodeJWTSegment` are deleted.
- The identity gate still runs before the rest of config resolution (`docs/design/BACKEND_PROXY_PATTERN.md` §4) — it resolves only the app URL it needs internally and fails closed on `reasonIdentityUnavailable` if that isn't available yet, rather than being reordered.
- `docs/developer/CODA.md` and `docs/design/BACKEND_PROXY_PATTERN.md` §3 updated: both described JWKS verification as a "future hardening" item; now they say what the code actually does.

## Why

Algorithm confusion, fail-open behavior, and stale documentation were all checked directly rather than assumed:

- authlib hardcodes `jose.ES256` as the only accepted signing algorithm in `ParseSigned`, so `alg: none` and algorithm-confusion attacks are rejected before key selection ever runs.
- Every failure branch — no app context, unresolved app URL, empty token, JWKS unreachable, unknown `kid`, forged signature, wrong `typ`, expired, missing `exp` — was traced and fails closed; none pass through.
- The per-app-URL verifier cache uses a single mutex across the full check-then-construct-then-store sequence, with no lock held across the underlying JWKS network fetch (authlib's own retriever handles that with singleflight). `go test -race` is clean.
- `BACKEND_PROXY_PATTERN.md` §3 still named JWKS verification a "future hardening item; do not re-argue it per PR" — now false and contradicted the correctly-updated `CODA.md` paragraph from the same change, so it's fixed here too.

## How to verify

```bash
go build ./...
npm run lint:go
npm run test:go
go test -race ./pkg/plugin/... ./pkg/plugin/auth/...
```

New tests in `pkg/plugin/auth/identity_verifier_test.go` cover: valid token accepted, forged signature rejected, unknown `kid` rejected, wrong `typ` rejected, expired rejected, missing `exp` rejected, JWKS-unreachable fails closed, per-app-URL caching. `app_platform_identity_test.go` rebuilt around a real ES256 signer plus a shared `httptest` JWKS server.

## Breaking changes

None at the API level. Two existing tests (`TestCustomGuide_NoAppURLStructurallyUnavailable`, `TestMyCompletions_NoAppURLStructurallyUnavailable`) now expect `reasonIdentityUnavailable` instead of the old app-URL-unavailable reason, since the identity gate now fails first and correctly — verified against the actual handler code, not just adjusted to match.

## Reversibility

Fully reversible — two commits, no migrations, no persisted-format change. `go-jose/v4` moved from an indirect to a direct `go.mod` dependency (it's imported directly by test code to hand-sign JWTs); no new module added.

Fixes #1568


Jayclifford345 (commented):
## Review — routed via `/review` (`docs/design/CONCERNS.md`)

**Disposition: request changes**, on one blocking defect. The security design underneath is sound: I confirmed the forged-`sub` hole from #1568 is genuinely closed — `getCompletionIndex` has exactly two callers and both sit behind the verified gate, so no warm-cache read escapes it. authlib's ES256-only `tokenSignAlgs` does reject `alg: none` and algorithm confusion at parse, as the PR body claims.

Concerns activated: `security` (+ standalone `secure` lens), `go-backend`, `correctness-and-reliability`, `testing-and-verification`, `reversibility-and-one-way-door`, `cross-cutting-architecture`.

---

## 🔴 Blocking — the JWKS URL is unreachable on every real Grafana

`pkg/plugin/auth/identity_verifier.go` builds the endpoint by raw concatenation:

```go
SigningKeysURL: appURL + signingKeysPath
```

Grafana's `parseAppUrlAndSubUrl` (`pkg/setting/setting.go`) **unconditionally appends a trailing slash** to `root_url` before exporting it as `GF_APP_URL` — `if appUrl[len(appUrl)-1] != '/' { appUrl += "/" }`. The SDK's `AppURL()` returns it verbatim with no trimming. So the built URL is `https://<stack>//api/signing-keys/keys`.

An operator cannot opt out of this: Grafana appends the slash back regardless of what `root_url` / `GF_SERVER_ROOT_URL` is set to. That line predates 2021, so no version at or above the `>=12.3.0-0` `grafanaDependency` floor escapes it.

Reproduced end to end against a real container (`grafana/grafana:latest`, and independently on `grafana/grafana:12.3.0`), calling authlib's own `NewKeyRetriever(...).Get()`:

| appURL | built URL | result |
| --- | --- | --- |
| `http://host:3011` | `…/api/signing-keys/keys` | `invalid token: unrecognized signing key` — **JWKS fetched and decoded fine** |
| `http://host:3011/` | `…//api/signing-keys/keys` | `unable to fetch signing keys: unable to decode response` |

The error-class difference is the proof. Grafana returns **302 → `/login`** for the double-slash path (its router strips only one leading `/`, so the empty first segment never matches), and `http.DefaultClient` follows the redirect to a 200 HTML page, which then fails the JWKS JSON decode.

**Impact:** `verifyIDToken` returns `("", false)` for every request, so `/completion-records/*` and `/custom-guide-repository` serve `capability.available: false, reason: "identity-unavailable"` for every user, permanently. It's a soft-200, so nothing in HTTP monitoring fires. It fails closed, so this is an outage rather than a vulnerability — but it is a total, silent one.

This repo already knows about the trailing slash — `buildAppPlatformURL` at `pkg/plugin/app_platform_client.go:49` does `strings.TrimRight(appURL, "/")`, and `app_platform_client_test.go` pins exactly that input shape. The new code is the only app-URL consumer that skips it.

### The test suite structurally cannot catch this

- `httptest.Server.URL` never carries a trailing slash, so every new test supplies the one shape Grafana never emits.
- The shared harness in `app_platform_identity_test.go` serves JWKS from `http.NewServeMux`, which **301-cleans** `//api/…` → `/api/…`, and `http.DefaultClient` follows it. `identity_verifier_test.go`'s bare `http.HandlerFunc` answers on *any* path.

I verified this masking directly: appending `/` to the test server URL leaves the entire suite green.

### Suggested fix (validated)

```go
SigningKeysURL: strings.TrimRight(appURL, "/") + signingKeysPath
```

I applied this plus a strict-path regression test — a bare `http.HandlerFunc` asserting `r.URL.Path == "/api/signing-keys/keys"` exactly, driven with an appURL ending in `/`. It **fails before the change and passes after**, with `go test ./pkg/...` green either way. A test written against `ServeMux` cannot fail, so it has to be the strict form.

Worth also normalizing the `verifiers` map key, so `https://x` and `https://x/` don't get two verifiers with two independent key caches.

---

## Medium

**1. The new in-request network hop is unbounded.** `NewKeyRetriever` defaults to `http.DefaultClient` (no `Timeout`), and Grafana's `CallResource` sets no context deadline — so a hung signing-keys endpoint parks handler goroutines with no ceiling, inside a plugin whose caller is the same Grafana that is waiting on it. authlib singleflights on a fixed `"fetch"` key, so the first caller's cancellation fails verification for every concurrent waiter. Every other outbound hop here is explicitly bounded (`appPlatformUpstreamTimeout`, `customGuideAggregateDeadline`, `completionAggregateDeadline`). Suggest `authn.WithHTTPClientKeyRetrieverOpt(&http.Client{Timeout: …})`.

Related: a *failed* fetch is never negatively cached, so a JWKS outage re-attempts on every request — the exact shape `completionFailureCooldown` already exists to prevent one layer down.

**2. The identity gate now runs ahead of the cheap local checks.** Both handlers call the gate before `resolve*Backend`, so a stack with the aggregation toggle **off** still pays a JWKS round trip on every request to do nothing. §4 of the pattern doc orders identity first for information-disclosure reasons, but a toggle-off check leaks nothing about the caller — worth a deliberate call either way.

**3. The namespace claim is decoded and thrown away.** The type comment justifies skipping audience validation because "the namespace claim serves that role instead" — and then nothing reads it. `VerifySubject` returns only `claims.Subject`; `claims.Rest.Namespace` is discarded. Grafana's own `ext_jwt` client, which this comment names as the model, does **both** halves — empty `AllowedAudiences` *and* `errExtJWTDisallowedNamespaceClaim`. The trusted counterpart is already in hand as `PluginConfigFromContext(...).Namespace`, and `authlib/types.NamespaceMatches` is already in `go.mod`. Either check it, or drop the clause — as written, the rewritten trust-boundary docs claim a stronger guarantee than the code delivers.

**4. The capability `reason` vocabulary collapsed.** Because the gate now resolves the app URL itself, `reasonAppURLUnavailable` and the `cfg == nil` arm of `reasonBackendUnavailable` are **unreachable**. Those tokens were split deliberately — the const block says the shared token was "undiagnosable from the capability envelope alone (the only signal available without backend log access)". `reason` is also the attribute on the alertable `recordCustomGuideCatalogueUnavailable` Faro event, so a JWKS outage is now indistinguishable from "nobody is signed in".

A related sharpening: `handleCompletionCapability`'s own comment says "a probe that flips available=false during a 30-second blip would grey out UI for everyone" — a transient JWKS blip now takes exactly that `capability=false` branch instead of the transient 503 + `Retry-After` path. A distinct token (e.g. `identity-unverifiable`) for verification failures would fix the diagnosis *and* make the blocking defect above detectable on the first stack rather than via user reports.

**5. The new gate logs nothing.** All four failure causes are discarded unlogged, and `VerifySubject` goes to the trouble of producing a classified error that nothing consumes. This contradicts §9 and the repo's own `credentialDiagOnce` precedent, written for precisely this incident shape. Log the error class and resolved JWKS URL — never the token, `sub`, or key material. (Confirmed clean on leakage today: authlib's errors are static sentinels, and the one token-adjacent log site already redacts to a boolean.)

**6. QC8 — comment volume.** `identity_verifier.go` is ~33% comment lines, `app_platform_identity.go` ~38%. Four earn their place and should stay: the go-jose `exp` gotcha (a real external-behavior workaround, correctly pinpointed), the no-audience-check rationale, the leeway note, and the wall-clock-expiry trap in the tests. The rest fail `AGENTS.md`'s bar — `NewIdentityVerifier`'s comment restates its own constructor, `VerifySubject`'s seven-clause failure list restates "returns an error", `verifyIDToken` spends 7 of 10 lines arguing the design against an imagined alternative (that belongs in the PR body), and the same 5-line block is pasted verbatim above both `*_NoAppURLStructurallyUnavailable` tests to paper over names that no longer describe what they assert.

---

## Low

- **Wrong doc citation**: two comments cite "BACKEND_PROXY_PATTERN.md §4" for an ordering rule. §4 is *Cache*; §3 is *Caller identity* — and §3 doesn't state that rule either (it says "before spending an upstream call", which is a different constraint). This PR already edits §3, so it's the natural place to add it.
- **Signing keys are cached with `NoExpiration`** — rotation forward works (a new `kid` misses and refetches), but revocation never takes effect until process restart. Separately, an unknown `kid` is negative-cached for 10 minutes and Grafana's `kid` is month-scoped, so a month-boundary race could darken both routes for up to 10 minutes. `WithKeyRetrieverCache` at the endpoint's advertised `max-age=3600` would honour the issuer's own bound.
- **The "never nil" invariant isn't enforced**: 2 of the 4 `&App{}` sites (`coda_exec_test.go:124`, `stream_test.go:94`) leave `identityVerifier` nil, and a nil receiver would **panic** in `verifierFor`'s `mu.Lock()` rather than fail closed. Latent — neither reaches the gate today.
- **PR body correction**: authlib was **not** bumped. Base and head pin the same digest (`20260723132736-22c8f38b300f`); the go.sum diff only removes stale lines for a previously-superseded one. go-jose moves indirect→direct because of *test* imports only. `govulncheck` is clean for both (the one advisory, `GO-2026-5932` in `x/crypto` openpgp, is pre-existing and off this path).
- **Stale prose left behind**: `resolveCustomGuideBackend`'s doc comment still lists "no app URL" as a reachable cause; `BACKEND_PROXY_PATTERN.md` still says "A namespace-global catalogue proxy validates structure and forwards" two bullets below the one this PR rewrote; `handleCompletionCapability` is still described as "a cheap probe" though it now makes a network call; and `src/lib/custom-guide-repository-client.ts:48` still documents `app-url-unavailable` to frontend consumers.
- **Test hygiene**: the shared `sync.Once` JWKS server is never closed (the sibling package correctly uses `t.Cleanup(srv.Close)`), and the two flipped `NoAppURL` tests now duplicate `TestValidIDToken_UnresolvedConfigFailsClosed` while retaining an inert `withFrozenTime`.

---

## Residual worth recording

Verification proves **issuance**, not **possession by the requester**. Forgery is closed; replay is not — the `X-Grafana-Id` header is still chosen by an API-key/service-account caller, and anyone holding a copy of another user's unexpired ID token can present it on `/completion-records/my`. Binding `claims.Rest.Username` against `PluginConfigFromContext(...).User.Login` would close it. Since this PR is rewriting exactly the CODA.md paragraph that used to name JWKS as *the* residual, that section is the natural place to name this one in its stead — otherwise the doc now reads as though the header is fully trustworthy.

---

## Doc drift (non-blocking)

Three agent-guidance files still describe the replaced mechanism and were not updated:

- `AGENTS.md` — the Backend section describes the `pkg/plugin/auth` seam as OBO-minting only; it now also owns inbound verification.
- `.cursor/rules/coda.mdc:91` and `.cursor/rules/systemPatterns.mdc:186` — both still say "fail-closed **structural** JWT validation", and neither file map has a row for the new `pkg/plugin/auth/identity_verifier.go`.
- `docs/developer/CONTEXT_INDEX.md` — the `BACKEND_PROXY_PATTERN.md` routing glob is `pkg/plugin/*.go`, which doesn't match the new subdirectory.

(The dated `#1398`/`#1400` conformance appendix in `BACKEND_PROXY_PATTERN.md` still naming JWKS as future work is a frozen historical record and should stay as-is.)

**No contract anchor row is required.** This strengthens a contract already written in `BACKEND_PROXY_PATTERN.md` §3 and `CODA.md`, and updates both in the same PR — it doesn't establish or replace one. Contract-evolution scan for `go-backend` returned `coherent_extension`, no finding.

---

## Verification run (at head `b0f087e`)

| Command | Result |
| --- | --- |
| `go build ./...` / `go vet ./pkg/...` | pass |
| `npm run lint:go` | 0 issues |
| `npm run test:go` | pass |
| `go test -race -count=1 ./pkg/...` | pass |

All of the PR's own verification claims hold — which is exactly why the blocking defect matters: a fully green suite coexists with a total feature outage.

<sub>Automated review pass (Claude Code `/review`). Findings above `low` went through adversarial verification; the blocking one survived three independent refutation attempts and was reproduced against live Grafana containers.</sub>

tomglenn:
Closing as duplicate of #1604

PR #635, opened 2026-02-28
Phase 3: Package engine for bundled runtime resolution

moxious:
## Summary

Implements the `package-engine` Tier 2 engine that resolves bundled packages at runtime using the `repository.json` migrated in Phase 2. This completes the local end-to-end cycle: the plugin can now load and resolve packaged content.

- **Resolution types** (`PackageResolution` discriminated union, `PackageResolver` interface, `ResolveOptions`) defined at Tier 0 in `src/types/package.types.ts` for broad importability
- **`BundledPackageResolver`** reads bundled `repository.json`, resolves bare package IDs to `bundled:` URLs, with optional content loading via `loadContent: true`
- **Package loader** (`loader.ts`): loads `content.json` + `manifest.json` from bundled directories, supports legacy single-file fallback; intentional transitional duplication with `docs-retrieval` to avoid lateral coupling
- **Structural dependency resolver** (`dependency-resolver.ts`): `getProviders()`, `getTransitiveDependencies()` with cycle-safe DFS, `getRecommendedBy()`, `getDependedOnBy()`, `flattenDependencyList()`, `buildProvidesIndex()`
- **Barrel export** with stable public API surface — consumers import through `package-engine/` barrel
- Registered in `TIER_MAP` and ESLint tier boundary config; architecture ratchet unchanged (vertical=4, lateral=9, barrel=0)

### Files changed

| File | Change |
|------|--------|
| `src/types/package.types.ts` | Added resolution types (Tier 0) |
| `src/types/index.ts` | Export package types from barrel |
| `src/validation/import-graph.ts` | Register `package-engine` in TIER_MAP |
| `eslint.config.mjs` | Register `package-engine` in TIER_2_ENGINES |
| `src/package-engine/index.ts` | Barrel export |
| `src/package-engine/resolver.ts` | BundledPackageResolver implementation |
| `src/package-engine/loader.ts` | Content/manifest loader |
| `src/package-engine/dependency-resolver.ts` | Structural dependency queries |
| `src/package-engine/resolver.test.ts` | Resolver tests (mocked + integration) |
| `src/package-engine/loader.test.ts` | Loader tests against real bundled content |
| `src/package-engine/dependency-resolver.test.ts` | Dependency resolver pure unit tests |
| `docs/design/PACKAGE-IMPLEMENTATION-PLAN.md` | Phase 3 marked complete with decisions |

## Test plan

- [x] 72 Layer 2 unit tests pass (`npx jest --testPathPatterns=src/package-engine/`)
[...]

PR #1610, opened 2026-08-13
fix(package-engine): gate URL-only app-platform resolution on publish status

### LINKED ISSUE #1561: App-platform resolver URL-only mode vouches for unverified ids — drafts open by id, not-found error class regressed (#1408 follow-up)
Follow-up from the second-round review of PR #1408 (https://github.com/grafana/grafana-pathfinder-app/pull/1408).

## Severity

**Medium.** Confirmed by 3 adversarial skeptics with honest reachability trims (below). Namespace-scoped (viewers in the namespace can read drafts via the API anyway), so this is a product published-only gate bypass and an error-UX regression, not a tenant-isolation hole — but the PR's own comment at `app-platform-resolver.ts:176-179` names "every namespace viewer" as the audience the gate exists for.

## Analysis

`AppPlatformPackageResolver.resolve` with `loadContent` falsy returns **unconditional success** — no existence check, no `spec.status` check (`src/package-engine/app-platform-resolver.ts:146-156` returns `{ok: true, contentUrl: 'backend-guide:' + id}` before any fetch). Both sibling resolvers validate even in URL-only mode: bundled checks index membership (`src/package-engine/resolver.ts:47-49`), CDN returns not-found for absent entries (`online-cdn-resolver.ts:53-55`). `PackageResolutionSuccess` is documented as "the resolver found the package" (`src/types/package.types.ts:196-199`), so this breaks the composite's success ⇒ exists contract for the one mutable tier.

Downstream, `fetchBackendInteractive` (`src/docs-retrieval/content-fetcher/backend-guide.ts:46-59`) has **no status check** — its `BackendGuideResource` interface doesn't even model `status` — and it calls the aggregated single-resource GET (`src/utils/interactive-guides-api.ts:29`), which serves drafts (published-only is purely a client-side LIST filter, `src/utils/fetchBackendGuides.ts:43`; the block editor loads/saves drafts through the same endpoint). So:

1. **Drafts open fully by id**, bypassing the published gate this same PR added one branch up (`app-platform-resolver.ts:174-180`), which renders the same guide locked in every list surface. Reachable vectors (verified): a published guide's `recommends`/`suggests` fallback nav links carry a packageId with empty contentUrl (`context-panel.tsx:170`) and route through `docs-tab-loader.ts:25` → `fetchPackageById` (`package-content.ts:334`, URL-only mode); restored tabs (`docs-panel.tsx:821-824`); any future caller of the exported resolution API. (The raw `?doc=api:<id>` deep link renders drafts too, but that path is byte-identical to merge-base — pre-existing, not introduced here.)
2. **Error-class regression:** a nonexistent id previously produced `errorType: 'not-found'` ("Document not found") from the composite's lastFailure (`package-content.ts:340-341`); it now resolves ok, fires a doomed request, and surfaces as `'other'` ("Failed to load custom guide") via `backend-guide.ts:82-88`.

## Implementation

Preferred single-chokepoint fix — enforce status at the content layer, which also covers the pre-existing deep-link vector:

1. In `fetchBackendInteractive` (`src/docs-retrieval/content-fetcher/backend-guide.ts`), add `status` to `BackendGuideResource` and fail resolution when `spec.status !== 'published'` (match the resolver's treatment at `app-platform-resolver.ts:174-180`; return a not-found-shaped error so the UX reads "not available" rather than a generic failure). If author preview of drafts is desired later, that's an explicit follow-up, not a default.
2. Alternatively (or additionally), make the app-platform resolver's URL-only branch do a metadata-only existence+status probe instead of returning unconditionally — but note the branch's current test enshrines the hole (`app-platform-resolver.test.ts:41`, "resolves URLs without hitting the backend"), so that test changes either way.
3. Restore the `not-found` error class for nonexistent ids (map the 404 from the single-resource GET to `errorType: 'not-found'`).
4. Tests: draft-by-id (assert locked/unavailable, not rendered), nonexistent-id (assert `not-found` class), and published-id happy path unchanged.

## Links

- The published-only invariant this bypasses was added by the same PR per review finding "draft guides render as available path members".

tomglenn:
## Summary

`AppPlatformPackageResolver.resolve()` returned an unconditional success in URL-only mode (`loadContent` falsy) with no publish-status check anywhere in the stack — the Go proxy passes `Status` through unfiltered, and every existing gate was client-side and only ran on the content-loading path. A draft or nonexistent guide id resolved as if it existed whenever a caller only needed the URL (deep links, bare-id milestone navigation).

## What

- Factored the existing fetch-and-status-check into a shared `probePublishedGuide()` in `src/package-engine/app-platform-resolver.ts`.
- Added `verifyPublished?: boolean` to `ResolveOptions` (`src/types/package.types.ts`), wired into exactly the one URL-only call site that needed it — `fetchPackageById` in `src/docs-retrieval/content-fetcher/package-content.ts`. The baseUrl-hydration call in the same file, which runs on every milestone content fetch, deliberately stays probe-free to avoid a network round-trip on a hot, unrelated path.
- Failed probes return `attemptedFailure()` (repository-tagged), not `decline()` — `composite-resolver.ts` keys cache eviction strictly on that tag, so a draft-then-published guide re-resolves correctly on the next call instead of staying cache-locked as missing.
- A 403 maps to the same `not-found` result as a 404, so an unauthorized caller can't use the response code as an existence oracle for a resource it can't read.
- Fixed during review: `CompositePackageResolver`'s cache key didn't include `verifyPublished`, so a probed and an unprobed call for the same draft guide, issued close enough together, could collide on one cache entry — silently reproducing the exact bug this PR closes, under a race instead of always. Folded `verifyPublished` into the cache key and added a regression test that fails without the fix.

## Why fixed this way, not a content-layer gate

The alternative — gating status inside the shared content-loading function — was rejected because that function also serves the canonical share-link scheme (`src/utils/find-doc-page.ts`): it's what "copy workshop link" emits and what auto-dock restore replays. Gating there would break share links and live workshops for any guide not currently published, including a workshop where the presenter sets the guide to draft mid-session. The resolver-tier probe adds the check in the one affected path only, leaving share-link behavior untouched — confirmed by rerunning `backend-guide.test.ts` unmodified (all 12 pass) and grepping it for any `resolve`/`ResolveOptions`/`verifyPublished` reference (none).

## How to verify

```bash
npm run typecheck
npm run lint:fix
npm run test:ci -- --coverage=false
```

Discrimination-checked twice independently: `probePublishedGuide()` was temporarily forced to always report published, which failed 10 of 18 tests in `app-platform-resolver.test.ts` (the three new draft/404/403 cases plus several pre-existing ones — confirming the gate is load-bearing, not just additive), then restored with a clean diff. The cache-key fix has its own dedicated regression test in `composite-resolver.test.ts`.

## Breaking changes

None. `verifyPublished` is optional and defaults to unset (existing behavior) everywhere except the one call site that now sets it explicitly. Note the shared `probePublishedGuide()` also changes 403 handling on the pre-existing content-loading paths (`loadContent: true` / `'metadata-only'`) from a generic `network-error` to `not-found` — no test pinned the old behavior, so nothing breaks, but it's a slightly wider surface than the URL-only path alone.

## Reversibility

Fully reversible — two commits, no migrations, no persisted-format or cache-shape change outside the process-local resolver cache.

## Scope — deliberately narrower than #1561

This gates **`fetchPackageById`** only: the context-panel `recommends`/`suggests` fallback links, which resolve by bare id with no URL. Two other routes to a draft stay open **on purpose**:

- `?doc=api:<id>` / `backend-guide:<id>` — the canonical share-link scheme
- restored tabs, which always replay a `backend-guide:<id>` URL

Both go through `fetchBackendInteractive`, and gating there was considered and rejected: it would break "copy workshop link" and auto-dock restore for any unpublished guide, including a live workshop whose author flips a guide to draft mid-session. That loader now carries a comment recording this.

Worth being explicit that this is a **product visibility gate, not an authorization boundary**. `itemUrl` targets the App Platform API directly from the browser under the user's own session, and Kubernetes RBAC there is namespace-scoped rather than object-scoped, so any namespace member can already read a draft directly. The remaining vectors are pre-existing rather than anything this PR introduces.

**Does not close #1561** — that issue names restored tabs as in-scope, and this covers one of the three routes. Leaving it open to track the remainder against the share-link decision above.

Refs #1561


Jayclifford345 (commented):
## Review — routed via `/review` (`docs/design/CONCERNS.md`)

**Disposition: approve with changes.** Nothing here blocks merge. The second commit's cache-key fix is real and well-tested — I mutation-verified it: reverting *only* the `cacheKey` line to the base form fails exactly the new test at `composite-resolver.test.ts` (`Expected false, Received true`), and nothing else. Good catch, and the ordering assumption holds for the right reason (`resolve` runs get → `resolveUncached` → set with no `await` between, so the first call's promise is installed synchronously).

The main concern is that the PR closes a narrower hole than its title and comments claim.

Concerns activated: `package-engine`, `docs-retrieval-and-rendering`, `guide-schema-and-contracts`, `security`, `correctness-and-reliability`, `testing-and-verification`, `reversibility-and-one-way-door`.

---

## Medium — the gate isn't at the chokepoint the issue asked for

`verifyPublished: true` is set at exactly one call site, and `loadDocsTabContentResult` only reaches `fetchPackageById` when `packageInfo` is set **and the URL is empty**:

```ts
if (packageInfo) {
  if (normalizedUrl) {
    return fetchPackageContent(normalizedUrl, ...);   // ← everything else lands here
  }
  if (packageInfo.packageId) {
    return fetchPackageById(packageInfo.packageId, ...);  // ← the only gated branch
  }
```

In practice that's the context-panel `recommends`/`suggests` fallback links. Every other entry point carries a `backend-guide:<id>` URL and goes `fetchContent` → `fetchBackendInteractive`, which this PR doesn't touch and whose `BackendGuideResource` interface doesn't even declare `status`.

So drafts still open through:

- **`?doc=api:<id>`** — the canonical share-link form (`find-doc-page.ts:36-46` maps both `api:` and `backend-guide:` straight to the ungated scheme, with no backend contact)
- **restored tabs** — `initializeRestoredActiveTab` → `loadTab(id, currentUrl || baseUrl)`, always a non-empty `backend-guide:<id>`

#1561 named restored tabs as an in-scope vector, and recommended the `fetchBackendInteractive` chokepoint as option 1 precisely because it "also covers the pre-existing deep-link vector". The shipped gate covers the narrowest of the three.

**Severity recalibrated after adversarial verification, and I agree with the correction.** `itemUrl` targets the App Platform API **directly from the browser under the user's own session** — the Go plugin only proxies the namespace LIST (`/custom-guide-repository`), and its own client comments note "Kubernetes RBAC is namespace-, not object-, scoped". Any namespace member can already `curl` a draft. So this is a **product/UX visibility gate, not an authorization boundary**, and the residual vector is **pre-existing rather than a regression this PR introduces**. That's why it's medium and non-blocking.

Two concrete asks regardless of which way you go:

1. **The new comment is factually wrong.** "this resolve() is the whole answer for a deep link (no content-loading call follows to catch a draft/missing id)" — a content-loading call *does* follow, three lines later (`return fetchPackageContent(resolution.contentUrl, ...)`). The accurate statement is narrower: the follow-on fetch has no *publish-status* gate. As written it will mislead the next reader into thinking the resolver is the only request on this path (it isn't — see below). Same issue with "they'll hit the same gate there" in `app-platform-resolver.ts` and "will fail there" in `package.types.ts` — both name a downstream gate that does not exist.
2. **No test for `?doc=api:<draft-id>`.** The whole `src/package-engine` + `src/docs-retrieval` suite (453 tests) is green with the vector open.

If the narrower scope is deliberate, that's a defensible call — say so in the PR body and leave #1561 open (or split a follow-up) rather than closing it on this evidence.

---

## Medium

**`verifyPublished` fails open by default.** Only `AppPlatformPackageResolver` reads it, out of four leaf resolvers; and `CompositePackageResolver.resolveUncached` returns the **first `ok`** result, so an id that also exists in the bundled or CDN tier returns success without app-platform's probe ever running. The doc comment's "Ignored by resolvers with no draft/published distinction" is factually accurate today — bundled checks `Object.hasOwn` and CDN checks index membership, both unconditionally — but nothing enforces it, and #635's documented `PackageResolutionSuccess` invariant ("the resolver found the package") is now caller-parameterized per tier.

Forgetting the flag is *exactly* how this bug arose. Consider inverting the polarity — verification as the URL-only default, with an explicit `skipVerification` on the two known hot paths (baseUrl hydration, milestone fan-out) — so the safe behaviour is what you get by omission. At minimum, a shared conformance test pinning which resolvers honour it.

**The gated path now GETs the same resource twice.** `probePublishedGuide` fetches the full guide — `spec` carries the entire `blocks` array — reads one string field, and drops `probe.resource` on the floor; `fetchPackageContent` → `fetchBackendInteractive` then issues the byte-identical GET again, sequentially. Nothing dedupes: `UNCACHEABLE_REPOSITORIES` evicts every app-platform entry on settle, and `getBackendSrv().fetch` has no response cache. This is the same cost the PR declares unacceptable one file over ("a network round-trip probe here would be a real perf regression"). Threading `probe.resource` through — or putting the gate in `fetchBackendInteractive`, which fixes both this and the scope issue — collapses it to one request.

**The 403 → `not-found` remap buys no confidentiality and widens past the stated scope.** The caller here is the user's own browser, which already received the raw 403 before this code runs — nothing is concealed. What it does do: a user whose session expired, who lost RBAC, or whose org never granted the role now sees `Failed to resolve package: <id>` with a Retry button that can never succeed, indistinguishable from a deleted guide. And because failures are repository-tagged and evicted on settle, a permanently-403 id re-issues the doomed request on every retry with no negative caching.

It also applies to the pre-existing `loadContent: true` / `'metadata-only'` paths via the shared helper, not just the URL-only path the PR body scopes it to. Suggest keeping 404 → `not-found` and giving 403 its own code, or at minimum logging the underlying status before collapsing it and replacing the existence-leak rationale with the real one (uniform UI).

**Contract evolution (`package-engine`): `contract_missing`, advisory — not blocking.** `status === 'published'` is inlined independently at four sites (`probePublishedGuide`, `usePublishedGuides`, `fetchAppPlatformLearningPaths`, `fetchBackendGuides`), plus a fifth deliberate exemption (`fetchBackendInteractive`) recorded only in the PR body. Within the *resolution* contract this PR is well-behaved — `verifyPublished` lands on the central `ResolveOptions`, failures keep the established repository tag so #1408's eviction still frees a draft-then-published guide, and the shared helper removes a duplicated fetch+status block. No new representation is added, hence no branching condition. The cheapest fix is a one-line comment on `fetchBackendInteractive` stating drafts are served there **on purpose** for share links and workshop replay, so the divergence doesn't have to be rediscovered on every future change.

---

## Low

- **Untested branches** tied directly to the changed semantics: the probe's `network-error` fallthrough (`grep -c 'network-error'` on the test file returns 0 — a previously infallible URL-only path is now fallible), the `!resource?.spec` branch, and 403 on the content/metadata-only paths. The PR body notes no test pinned the old 403 behaviour — which means neither the old nor the new semantics is pinned now.
- **QC8**: ~40 lines of added multi-line comment across ~60 lines of new logic. The 5-line block on `cacheKey`, the 9-line `verifyPublished` JSDoc, and the three test comments could each be one line or dropped (the test names already say it).
- **The `not.toHaveBeenCalledWith` assertion is near-redundant.** Mutation-tested: making the hydration call pass `verifyPublished: true` fails at the *positive* assertion above it (jest's `toHaveBeenCalledWith` is exact structural equality), so line 619 is never reached. It only bites against an *additional* verified call — worth saying that in the comment, or dropping the line.
- **Cache key over-splits**: `verifyPublished` is inert when `loadContent` is truthy (the probe runs regardless), so `{loadContent: true}` and `{loadContent: true, verifyPublished: true}` are semantically identical yet get separate entries, each doing a full content load. No caller does this today. Also note the key split means `fetchPackageById` (`id:false:true`) no longer shares a resolution with the hydration call (`id:false:false`) — free for bundled/CDN, but `RecommenderPackageResolver` has no cache, so a recommender-enabled stack now issues two GETs where one served both.
- **`ProbeResult` discards the narrowing it just established** — `resource.spec` stays optional, so downstream re-guards with `?.` and `buildManifest` takes an optional spec. Typing success as `{ ok: true; spec: NonNullable<...> }` would record the invariant the probe proves.

---

## Checked and clean

- **`itemUrl` F3 encoding claim holds** — both `namespace` and `name` go through `encodeURIComponent`, so separators can't break out of the path segment. (Only residual: `..` isn't encoded, but it degrades to the LIST endpoint whose `spec` is undefined → `not-found`. Pre-existing, no exposure.)
- **Tier fallthrough is not short-circuited** — `resolveUncached` records `lastFailure` and keeps iterating, and app-platform is last anyway. The repository-tagged `attemptedFailure` does drive eviction, so the draft→published re-resolve claim in the PR body is correct.
- **No persisted-state or one-way-door risk.** Read-path only; the composite cache is a process-local `Map<string, Promise>` that dies with the page. No code prunes completion records for unresolvable ids, and milestone locking is derived per render, not stored. A revert restores prior behaviour exactly. Worth stating in the PR's Reversibility section rather than leaving it to be re-derived.
- **No schema obligation** — `package.schema.ts` models only persisted JSON; `ResolveOptions` is an internal call-option type with no validated payload and no zod counterpart to keep in lockstep.

---

## Verification run (at head `d341807`)

| Command | Result |
| --- | --- |
| `npx tsc --noEmit` | clean |
| `eslint` on changed files | clean |
| jest — `src/package-engine` + `src/docs-retrieval` | 32 suites, 453 tests, pass |

The PR's "how to verify" claims hold.

<sub>Automated review pass (Claude Code `/review`). Findings above `low` went through adversarial verification — the scope finding survived, but its security framing was correctly refuted and is reflected in the recalibration above.</sub>

Jayclifford345 (dismissed):
Realised this one can be approvied, there are some minors so i won't merged till you have had a chance to read @tomglenn but can be a followup PR :)

Jayclifford345 (approved):

PR #1611, opened 2026-08-13
fix(config): give plugin-settings distribution a single owner

Jayclifford345:
## Summary

Pathfinder reads its settings from Grafana's `meta.jsonData` snapshot, which can lag a recent save. This switches to the settings API as the authoritative source, and consolidates the four places that wrote `window.__pathfinderPluginConfig` down to one.

The new `hooks/usePathfinderPluginConfig.ts` is the only writer of that global — the readiness signal the e2e runner waits on, and the config source for callers outside React.

## What to look at

**Boot ordering (`src/module.tsx`).** `plugin.init` stays synchronous and publishes from `meta.jsonData` immediately, then kicks off an unawaited refresh. `AppPlugin.init` is typed `void` and Grafana never awaits it, so anything behind an `await` would run after first paint — after scene construction has already read the published config, and after the deep-link and link-interception listeners needed to exist. Worth a look if you disagree that a synchronous first publish is required.

**Entry bundle.** `module.tsx` imports the hook file directly rather than through the `hooks` barrel. The barrel reaches `lib/user-storage` and transitively zod, which took `module.js` from 113 KB to 224 KB raw. Now 114,429 bytes raw / 37,791 gzip with zero zod references. Pinned by a new case in `entry-bundle-boundary.test.ts` plus an `ALLOWED_BARREL_VIOLATIONS` entry so the direct import is a documented decision rather than an oversight.

**Trust boundary on the update event.** `PATHFINDER_CONFIG_UPDATED_EVENT` is dispatched on `document`, so any script sharing the page can fire it. The event therefore carries no payload and subscribers re-read the global; nothing trusts `event.detail`. `event-names.ts` says so now.

**Draft preservation (`ConfigurationForm.tsx`, second commit).** The form seeds its draft from the API read, which unconditionally would discard whatever the admin typed before the response landed — and the draft is what `onSubmit` writes back, so the loss reached durable settings. Now gated on an `isDraftEdited` ref, with all field changes routed through an `editDraft` helper.

## Why

Four components each derived config from context and wrote the global on a last-writer-wins basis, so whichever effect ran last won and stale values could overwrite fetched ones. Each consumer also fetched independently. The refresh is now single-flight (one request per page load), and publishing is value-compared so a redundant publish costs no re-renders.

## How to verify

- `npm run check` passes; 425 suites / 6756 tests green.
- `npm run build` then check `dist/module.js` is ~114 KB raw and contains no `_zod` references.
- Manually: change a setting, save, and confirm the panel picks it up without a full reload. Then start typing in the config form on a slow connection and confirm your edit survives the settings response.

## Reversibility

Fully revertable — no schema, storage, or API-contract change. The two commits are independent and can be reverted separately.

## Open item

This establishes a contract (one owner for `window.__pathfinderPluginConfig`, payload-free invalidation event). Per `AGENTS.md` that wants a contract anchor row in `docs/design/CONCERNS.md`; not added yet — say the word and I'll push it here rather than leaving it for a follow-up.

## Note on provenance

Split out of the `spike/coda-client-real-sdk` branch behind #1607, where it was unrelated to that PR's subject. Rebased onto `main`; the only file both touch is `docs-panel.tsx`, in non-overlapping regions.

### DIFF
--- a/docs/developer/components/App/README.md
+++ b/docs/developer/components/App/README.md
@@ -81,12 +81,9 @@ The root application component that initializes the plugin, handles error bounda
 ```typescript
 function App(props: AppRootProps) {
   const scene = useMemo(() => getSceneApp(), []);
-  const config = useMemo(() => getConfigWithDefaults(props.meta.jsonData || {}), [props.meta.jsonData]);
 
-  // Set global config for module-level utilities
-  useEffect(() => {
-    (window as any).__pathfinderPluginConfig = config;
-  }, [config]);
+  // `window.__pathfinderPluginConfig` is published by `plugin.init` via
+  // `publishPathfinderPluginConfig`, which is its only writer.
 
   // Initialize plugin lifecycle
   useEffect(() => {
--- a/src/components/App/App.tsx
+++ b/src/components/App/App.tsx
@@ -8,7 +8,6 @@ import { homePage } from '../../pages/homePage';
 import { docsPage } from '../../pages/docsPage';
 import { fullScreenPage } from '../../pages/fullScreenPage';
 import { PluginPropsContext } from '../../utils/utils.plugin';
-import { getConfigWithDefaults } from '../../constants';
 import { onPluginStart } from '../../context-engine';
 import { PathfinderFeatureProvider } from '../OpenFeatureProvider';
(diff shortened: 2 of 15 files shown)


Spidy88 (approved):
LGTM

PR #1612, opened 2026-08-13
feat(interactive): roll the data check into the datasource picker

Jayclifford345:
## Summary

Supersedes #1597. Same capability, one less step type: the data check becomes a field on the **existing datasource picker** rather than a block type of its own, enabled per-guide by authors.

https://github.com/user-attachments/assets/a643714d-4632-459c-b8d0-0dff5da8dfae

A guide can teach "build a panel showing container CPU", the user can follow every step against an instance with no container metrics, and end with an empty panel and no idea why. Adding `dataCheckQuery` to a datasource picker lets them confirm up front that the data is really there.

```json
{
  "type": "input",
  "id": "check-container-metrics",
  "inputType": "datasource",
  "prompt": "Pick the data source holding your container metrics.",
  "variableName": "metricsDatasource",
  "datasourceFilter": "prometheus",
  "dataCheckQuery": "container_cpu_usage_seconds_total",
  "dataCheckFailureMessage": "No container CPU metrics here. Pick a data source scraping cAdvisor or kubelet.",
  "dataCheckBlocking": true,
  "skippable": true
}
```

Five flat optional fields, all `"datasource"` inputType only: `dataCheckQuery` (its presence enables the check), `dataCheckFailureMessage`, `dataCheckTimeFrom`, `dataCheckTimeTo`, `dataCheckBlocking`.

## What to look at

- **One JSON block, two element types** — [`json-parser.ts`](../blob/claude/datasource-picker-data-check/src/docs-retrieval/json-parser.ts). This is the decision the whole PR turns on. `parseTypeKey` keys on the *parsed element type*, not on props, so registering `InputBlock` as a tracked step would make **every** input block — text, boolean, plain datasource — count toward section completion. That is exactly what the comment at `content-renderer.tsx:923` warns about ("would inflate the total step count, making 100% completion impossible"). So `convertInputBlock` branches instead: a blocking check emits `datasource-check-step`; everything else emits `input-block`, untouched. Authors still see one block type; the split is internal, and no existing guide changes behaviour.

- **`dataCheckBlocking` is the only axis** — [`datasource-check-step.tsx`](../blob/claude/datasource-picker-data-check/src/components/interactive-tutorial/datasource-check-step.tsx). Unset (default): the check is advisory, rendered inside the passive `InputBlock`, and never gates anything. Set: the block becomes a tracked step that completes only on a pass, or a skip when `skippable`. This is moxious's review point from #1597 taken literally — hard gating is an explicit author decision, never accidental.

- **Validation moved from the filter to the selection** — [`use-data-check.ts`](../blob/claude/datasource-picker-data-check/src/components/interactive-tutorial/use-data-check.ts). #1597 needed `datasourceType` to be a 4-value enum so the picker could not offer a selection no check could evaluate. The picker's `datasourceFilter` is a free substring-matched string and cannot change without breaking existing guides. So the picked data source's real type is resolved through `getNormalizedDatasourceType`; `null` renders a "this check can't run here" notice instead of the button. Strictly better — it validates what actually gets queried, not what was authored.

- **Query execution, ported unchanged** — [`run-data-check-query.ts`](../blob/claude/datasource-picker-data-check/src/lib/datasource/run-data-check-query.ts). Still the only place in the plugin that calls `/api/ds/query`, so the caps live in exactly one place: 15s abort, a 100-point cap **where the query type honours it** (Tempo's `limit`, Loki's `maxLines`), and a default 1h range. Two calibrations from @tomglenn's review: a Prometheus instant query takes no `step`, so `maxDataPoints` does nothing on the primary documented example and its real cost is series cardinality, which nothing here caps; and the time range is a default, not a bound. Worth re-tracing if you did not review #1597: "has data" means ≥1 **row**, not ≥1 frame — Grafana returns a schema-only frame for a query that matched nothing, so counting frames would report every miss as a hit.

- **`pausesSectionRun`** — [`step-type-registry.ts`](../blob/claude/datasource-picker-data-check/src/components/interactive-tutorial/step-type-registry.ts). A narrow slice of #1597's fix, and load-bearing rather than tidy-up: an unrecognised `targetAction` falls through `executeInteractiveAction`'s `default:` branch, which warns and then **reports success**. Without this, "Do Section" would mark the check complete without it ever running. `main` gates this on `isGuided`, which has exactly one runtime consumer, so the swap is four lines. It deliberately leaves the quiz / challenge / terminal instances of the same bug alone.

- **Shared, not duplicated** — `getDatasourceOptions` moved out of `input-block.tsx` into [`datasource-options.ts`](../blob/claude/datasource-picker-data-check/src/components/interactive-tutorial/datasource-options.ts) unchanged, so both hosts agree on filter semantics by construction. This closes the duplication I flagged on #1597, and this time without changing what `InputBlock` persists.

## Why

The stored value stays the data source **name**, not the uid #1597 used. Every existing consumer of a picker's variable expects a name — `{{variable}}` substitution, `var-` requirements, and reftarget selectors like `button:contains({{myDatasource}})` all match the visible name. The uid is resolved internally at query time.

All five fields are flat scalars because [`schema-options.ts`](../blob/claude/datasource-picker-data-check/src/cli/utils/schema-options.ts) reports nested objects as `unsupported` and silently skips them; flat keeps CLI and MCP authoring working for free.

The check runs on click alone. A `requirements` token would be re-evaluated by five independent mechanisms — a 2s per-step subscription, a 3s heartbeat, a 5s section recheck, a retry harness, and `SequentialRequirementsManager` on any context change. On a metered backend that is billable.

The Zod refinement rejects `dataCheck*` on a non-datasource input, and any check field other than the query without the query. Both would otherwise be silently inert, which reads to an author as a check that never runs.

## How to verify

`npm run check` is green: 432 suites, 6856 tests, coverage thresholds met, Go lint/tests and the review-contract check included.

Test-quality note: every new suite passed on first run, so I mutated the two invariants that matter most and confirmed they bite. The first mutation only broke 1 of 2 stale-pick tests — jsdom's `<select>` cannot hold a value that is not among its options, so asserting on the rendered value alone passes even with the validation removed. Those tests now assert that no check can *run* against a pick the picker never offered, and both fail against the mutation. Same treatment for the two `GuideResponseProvider` fixes, which fail 2 tests when reverted.

**Browser checks — the gap #1597 never closed.** Open `json-guide-demo` and scroll to "Data check demo", which ships one advisory and one blocking example:

1. **Pass** — pick a Prometheus data source, **Run check** → advisory shows "Data available"; blocking completes the step.
2. **Fail, advisory** — point the query at a nonsense metric → the failure message shows, and you can still save and move on.
3. **Fail, blocking** — same on the blocking block → step stays incomplete, "Run check again" appears, the section will not complete.
4. **Skip** — `skippable` releases the blocking step; confirm the badge reads "Skipped", not "Data available".
5. **Persistence** — reload; the pick is restored and `{{blockingDatasource}}` still substitutes the data source **name**.
6. **Change after a pass** — switch data source; the step returns to unchecked.
7. **Unsupported type** — set `datasourceFilter` to `mysql` and pick one → the "can't run here" notice renders, no button, and no query is issued.
8. **None of the type** — a filter matching nothing warns; Skip appears only with `skippable`.
9. **Cost caps** — network tab: exactly one `/api/ds/query` per press, `maxDataPoints: 100`, and the authored range.
10. **No regression** — a plain `inputType: 'datasource'` block with no `dataCheckQuery` still renders passive and uncounted, with Save/Reset unchanged.

Local Grafana has no Prometheus data source by default — provision one first.

## Undisclosed scope, now disclosed

`markStepCompleted` (`completion-store.ts`) now refuses to overwrite a `'skipped'` entry until a reset clears it. That is **not** data-check-specific — it applies to every step type, and it is why a section reporting a skipped step complete with the generic `'manual'` can no longer relabel it as one the user passed. It is store-level and covered by a store-level test using a plain `step-1` / `section-x`, not a data check.

## Breaking changes

None. Every field is additive and optional, and a guide with no `dataCheckQuery` parses to exactly the element it did before. No `@grafana/assistant` bump, and `EXPECTED_BLOCK_TYPE_COUNT` is unchanged because there is no new block type.

## Reversibility

Mostly reversible. The **one-way door is the capability, not the code**: this is still the first time the plugin executes a data source query. The mitigations are load-bearing and should not be relaxed without a deliberate decision — bounded range and result size, abort timeout, a check only ever running against a data source the picker currently offers, and user action as the only trigger. `docs/design/CONCERNS.md` records these as the `data-check` contract anchor, which this PR **establishes and replaces #1597's version of** rather than extending.

Partially reversible in the usual way: steps completed under a blocking check persist in the completion store and would be orphaned if the field were removed.

## Out of scope

- **The AI check mode.** Deferred pending @moxious's open question on #1597 about whether an AI verdict may create durable completion state. The orchestrator, verdict schema, assistant query tool, and the `@grafana/assistant` bump are all left behind; adding them later is additive.
- **#1597's other generic fixes**, which are real bugs but unrelated to the picker and worth their own PR so they are not lost when it closes: the `use-step-reset` / `step-redo-button` refactor across challenge / code-block / terminal / terminal-connect / quiz; `pausesSectionRun` on quiz and challenge; the terminal / terminal-connect `refTarget: 'none' → 'multiStepRefs'` + `isMultiStep: true` fix; and the `!isEnabled && !isCompleted` fix that keeps a completed step's Redo clickable.
- **`requirements` is dead on input blocks.** Declared in `InputBlockProps`, threaded all the way from JSON, and then omitted from the component's destructuring, so it silently does nothing. The new step *does* honour it via the checker, leaving a documented inconsistency. Separate fix.
- **Guide authors cannot discover this yet** — the format specifics live in the separate `grafana/interactive-tutorials` repo, same as for #1597.
- Pre-existing and untouched: `defaultValue` on a datasource input is dropped by the block-editor round trip; `BranchBlocksEditor` hardcodes `inputType: 'text'` so the datasource variant is unreachable **via the block editor** (raw JSON reaches it, and a blocking check inside an inline conditional parses to a real tracked step — the resulting count skew reproduces identically on the base commit with a quiz or challenge, so it is pre-existing); `getDataSourceSrv` deprecation warnings need a Grafana ≥13.1 floor.


src/components/interactive-tutorial/datasource-check-step.tsx
```diff
@@ -0,0 +1,376 @@
...
+
+  const markSkipped = checker.markSkipped;
+  const handleSkip = useCallback(async () => {
+    reportAppInteraction(UserInteraction.DataCheckSkipped, {
+      datasource_type: supportedType ?? 'unknown',
+      step_id: renderedStepId,
+    });
+    await markSkipped?.();
+    markComplete('skipped');
+  }, [markSkipped, markComplete, supportedType, renderedStepId]);
```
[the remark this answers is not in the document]
Jayclifford345: Fixed in bded7cb8. You were right on both halves, and there was a third defect sharing the root cause.

`handleSkip` now calls `reset()` first, which aborts the controller — so the query stops, `handleRedo` is no longer inert while `state === 'checking'`, and the late `handleRun` path can't reach `markComplete()`.

The pass-over-skip you flagged had a partial mitigation I'd have missed without your trace: the ported completion-store fix makes `markStepCompleted` refuse to overwrite a `'skipped'` entry, so the standalone store was already safe. But `onStepComplete` and `onComplete` fired regardless, so a section-owned step was still told it completed. The abort closes that.

The third defect: `run()` returned a boolean, so an aborted check was indistinguishable from a failed one and reported `data_check_failed`. It returns a `DataCheckOutcome` now — `'aborted'` reports nothing and completes nothing.

Four tests cover it: the signal is aborted on skip, the skip is what gets recorded when the abandoned request lands after it, no outcome event is reported for the abandoned check, and Redo works immediately. Removing the `reset()` fails them.

src/components/interactive-tutorial/data-check-controls.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const styles = useStyles2(getStyles);
+  const isChecking = state === 'checking';
+
+  return (
+    <>
+      {state === 'failed' && (
+        <div className={styles.failure} role="status" aria-live="polite" data-testid={failureTestId}>
+          <Icon name="exclamation-triangle" />
+          <div>{failureMessage || failureDetail || 'The data is currently not available.'}</div>
+        </div>
```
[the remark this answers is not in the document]
Jayclifford345: Fixed in bded7cb8. This is the sharpest finding on the PR, because the wrong copy would have been acted on rather than just misread — a blocking check plus `skippable` means a user who believes their metric is absent will skip a step whose real problem is their connection or permissions.

Root cause was in `useDataCheck`, not the banner: it collapsed every non-pass into one `'failed'` state, so the banner had no way to tell the two apart. `runDataCheckQuery` already distinguished them — `{ok: true, hasData: false}` vs `{ok: false, error}` — and the hook was throwing that away.

`'no-data'` and `'error'` are now separate states. The author's `dataCheckFailureMessage` speaks only for the first; an error renders "The check could not run." plus the underlying detail, and never the author's copy. They're also distinguished in telemetry via an `outcome` attribute, so "the data wasn't there" and "we couldn't look" stop being one number.

This also answers a follow-up @moxious asked for on #1597 — distinguishing "the check ran and found no matching data" from "the check could not run" — so it's recorded as an invariant on the `data-check` contract anchor rather than left as behaviour, and `json-guide-format.md` now tells authors to write the message for the empty-result case only.

src/components/content-renderer/content-renderer.tsx
```diff
@@ -1263,10 +1266,32 @@ function renderParsedElement(
...
           datasourceFilter={element.props.datasourceFilter}
+          dataCheckQuery={element.props.dataCheckQuery}
+          dataCheckFailureMessage={sub(element.props.dataCheckFailureMessage)}
+          dataCheckTimeFrom={element.props.dataCheckTimeFrom}
+          dataCheckTimeTo={element.props.dataCheckTimeTo}
         >
           {renderChildren(element.children)}
         </InputBlock>
       );
+    case 'datasource-check-step':
```
tomglenn, line 1277: **Must fix — this arm can return `null` and the entire suite stays green.**

I mutated this case to render nothing and ran the full suite: 432 suites, 6886 tests, all passing. The feature is completely dead in the app and CI says nothing.

The gap is that nothing joins the two halves. `json-parser-datasource-check.test.ts` asserts the parser emits a `datasource-check-step` element; `datasource-check-step.test.tsx` renders the component directly with props. No test takes guide JSON through the parser and into the renderer, so the wiring between them is unpinned.

One test that parses a guide containing a blocking check and asserts the step renders would close it. Worth having for a capability that is otherwise well covered — this is the one place where a green pipeline would let the whole thing ship inert.

Secondary, same case arm: this is the only tracked case of nine that does not forward `stepIndex` / `totalSteps` (compare the arms above at 1110-1239). `DatasourceCheckStep` declares both props and never reads them, so the section-supplied position is dropped on both ends. Wire them or drop the props.

src/lib/datasource/run-data-check-query.ts
```diff
@@ -0,0 +1,160 @@
...
+ * Per-type query model. Prometheus and Loki share `expr`; Tempo takes TraceQL
+ * under `query`, and Pyroscope needs a profile type plus a label selector,
+ * which authors write as `<profileTypeId>|<labelSelector>`.
+ */
+function buildQueryModel(type: SupportedDatasourceType, query: string): Record<string, unknown> {
+  switch (type) {
+    case 'prometheus':
+      return { expr: query, instant: true, range: false };
+    case 'loki':
+      return { expr: query, queryType: 'range' };
```
tomglenn: **Should fix — Loki is the one query type sent with no result bound.**

Tempo gets `limit`, Prometheus is `instant: true` so it returns a single sample per series, Pyroscope is a profile query. Loki alone goes out as a range query with no `maxLines` and no `limit`, so the only thing bounding it is the authored time range — which is itself only a default, not a cap (see my comment on the schema).

An authored `dataCheckTimeFrom: "now-10y"` against a Loki source is a ten-year log-selector scan from one button press. Server-side `max_entries_limit_per_query` and the datasource's own `maxLines` still apply, so this is a footgun rather than a hole, and guide content is first-party or stack-editor authored. But you already own a limits constant, and this is the one call site not using it:

```suggestion
      return { expr: query, queryType: 'range', maxLines: DATA_CHECK_QUERY_LIMITS.maxDataPoints };
```

Worth calibrating one claim while you are here: the PR body and `CONCERNS.md` both present `maxDataPoints: 100` as bounding result size generally. For the primary documented Prometheus example it does nothing at all — an instant query takes no `step`, and its cost is series cardinality, which nothing here caps. `dataCheckQuery: '{__name__=~".+"}'` is one unbounded high-cardinality query against a metered tenant. No client-side fix exists for that, so this is a wording correction rather than code.

src/types/json-guide.schema.ts
```diff
@@ -467,27 +467,80 @@ export const JsonQuizBlockSchema = z
...
+    placeholder: z.string().optional().describe('Placeholder text for text input'),
+    checkboxLabel: z.string().optional().describe('Label shown next to a boolean checkbox'),
+    defaultValue: z.union([z.string(), z.boolean()]).optional().describe('Default value (string or boolean)'),
+    required: z.boolean().optional().describe('Whether the input must be provided to continue'),
+    pattern: z.string().optional().describe('Regex pattern the value must match (text inputs only)'),
+    validationMessage: z.string().optional().describe('Message shown when validation fails'),
+    requirements: z.array(RequirementTokenSchema).optional().describe('Prerequisite conditions'),
+    skippable: z.boolean().optional().describe('Allow user to skip this block'),
+    datasourceFilter: z.string().optional().describe('Filter for datasource input (e.g., loki, prometheus)'),
+    dataCheckQuery: z
```
tomglenn, line 498: **Should fix — an empty or whitespace-only query is accepted, then silently does nothing.**

`dataCheckQuery: ""` and `dataCheckQuery: "   "` both pass validation. The parser then gates on `Boolean(block.dataCheckQuery?.trim())` and emits a plain passive `input-block`, so the author gets exactly the outcome your refinement was written to prevent: a block that reads as a check and never runs one.

Your own refinement already calls `.trim()` a few lines down, so the schema knows the distinction — it just is not applied to the field itself:

```suggestion
    dataCheckQuery: z
      .string()
      .trim()
      .min(1)
      .optional()
```

Two related gaps in the same block, both cheap:

- `dataCheckBlocking: true` with no `id` validates. `json-guide-format.md` tells authors to give every blocking check an explicit `id`, and completion records persist against it, so an author who omits it gets records that orphan on the next edit. A refinement requiring `id` when `dataCheckBlocking` is set would match the doc.
- The query has no maximum length.

src/types/json-guide.schema.ts
```diff
@@ -467,27 +467,80 @@ export const JsonQuizBlockSchema = z
...
+      .string()
+      .optional()
+      .describe(
+        "Query run against the picked data source to confirm it holds this guide's data. Its presence enables the check. Datasource inputs only, and only Prometheus, Loki, Tempo, and Pyroscope can be checked."
+      ),
+    dataCheckFailureMessage: z.string().optional().describe('Message shown when the check finds no data'),
+    dataCheckTimeFrom: z.string().optional().describe('Check query range start (defaults to now-1h)'),
```
tomglenn, line 508: The time fields are unvalidated free strings — `dataCheckTimeFrom: "banana"` validates and fails only at query time, as does an inverted range (`from: now`, `to: now-6h`).

I would not add Grafana time-expression parsing to Zod for this; the expression grammar is wide (relative expressions, ISO timestamps, epochs) and the cost/benefit is poor. But the description here says "defaults to now-1h", which is accurate, while `CONCERNS.md:122` lists "bounded range" alongside two genuinely enforced numbers. Those two readings disagree. A default is not a bound — worth making the contract row say "default range" so the anchor matches the code it is anchoring.

src/components/interactive-tutorial/datasource-check-step.tsx
```diff
@@ -0,0 +1,385 @@
...
+        setUncontrolledName(name);
+        return;
+      }
+      if (name) {
+        responseContext.setResponse(variableName, name);
+      } else {
+        responseContext.deleteResponse(variableName);
+      }
+    },
+    [responseContext, variableName, reset, persistReset, checkerResetStep]
```
tomglenn: **Must fix — a passed check is only retracted when the user drives this picker directly.**

`selectedDatasource` is derived from the guide response on every render (line 130), but this handler is the only thing that calls `reset()` / `persistReset()`. Any other writer of the same `variableName` changes the picked data source underneath a completed check and nothing retracts it.

Demonstrated with three jest reproductions:

1. **Stale pass, the serious one.** Pick Prometheus, run, pass — `onStepComplete` fires and the store records the completion. Another writer sets the variable to `Prometheus staging`. The picker re-renders showing the new value, the badge still reads "Data available", and neither `onStepReset` nor `resetStep` is called. The section goes on counting a completion for a check that never ran against the data source now selected.
2. Same shape with `hasData: false` leaves a stale failure banner standing against a different data source.
3. A `clearForGuide` empties the picker while "Data available" remains rendered.

This is reachable through the PR's own two-host design: an advisory `input` datasource block sharing a `variableName` with a blocking step is exactly the pairing the demo guide ships.

Also in this handler: `onStepReset` is never called, only `persistReset()` — and `persistReset` is a no-op when the step is section-owned (178-182). So even the visible path only retracts a completion for a standalone step; inside a section, changing the data source by hand leaves the completion standing too.

Suggested shape: key the retraction off the resolved name changing rather than off this handler, and call `onStepReset` alongside `persistReset` so section-owned steps retract as well.

src/components/interactive-tutorial/input-block.tsx
```diff
@@ -440,8 +459,27 @@ export function InputBlock({
...
+            {dataCheckQuery && (
+              <DataCheckControls
+                state={dataCheck.state}
+                failureDetail={dataCheck.failureDetail}
+                failureMessage={dataCheckFailureMessage}
+                canRun={dataCheck.canRun}
+                isUnsupportedType={Boolean(selectedDatasource) && !dataCheck.supportedType}
+                onRun={handleRunDataCheck}
+                runTestId={testIds.dataCheck.runQueryButton(variableName)}
+                failureTestId={testIds.dataCheck.failure(variableName)}
```
tomglenn: **Should fix — one test-id factory, two key spaces.**

Here the ids are built from `variableName`; in `datasource-check-step.tsx:357-358` the same two factories are called with `renderedStepId`. Both produce `datasource-check-run-<x>`, so an advisory block whose `variableName` happens to equal another step's rendered id collides, and a test-id lookup can match the wrong host.

Pick one key for both call sites — `stepId` reads correct given the factory's own parameter name.

Separately, `testIds.interactive.datasourcePicker` and `testIds.dataCheck.datasourcePicker` now both exist and one of them is redundant.

src/constants/testIds.ts
```diff
@@ -165,13 +165,23 @@ export const testIds = {
...
     lazyScrollRetryButton: (stepId: string) => `interactive-lazy-retry-${stepId}`,
   },
 
+  // Datasource picker with a data check attached
+  dataCheck: {
+    step: (stepId: string) => `datasource-check-step-${stepId}`,
+    datasourcePicker: (stepId: string) => `datasource-check-picker-${stepId}`,
+    runQueryButton: (stepId: string) => `datasource-check-run-${stepId}`,
+    skipButton: (stepId: string) => `datasource-check-skip-${stepId}`,
+    failure: (stepId: string) => `datasource-check-failure-${stepId}`,
```
tomglenn, line 182: Minor, but it feeds the collision noted on `input-block.tsx:473`: these factories name their parameter `stepId`, and one of the two call sites passes a `variableName` instead. Either rename the parameter to reflect that it is any stable key, or fix the call site — right now the signature documents an invariant the code does not keep.

docs/design/CONCERNS.md
```diff
@@ -72,32 +73,33 @@ Column key — **cat**: `AO`=always-on · `sub`=subsystem · `xcut`=cross-cuttin
...
...                                                                                                                                                                                                                                                                     | New `/api/ds/query` callers; changes to `DATA_CHECK_QUERY_LIMITS`; running a check from an effect rather than a handler; removing the offered-options validation of a restored pick; making `input` tracked for any shape other than a blocking check                                                                                                                                                                                                                                                                                                                    | `npm run test:ci`; the 11-point browser checklist in the PR that established this concern                                                                                                                                                          | `interactive-engine`, `requirements-manager`, `guide-schema-and-contracts`, `analytics-and-telemetry`, `reversibility-and-one-way-door` |
```
tomglenn: The `purpose` text and its impact questions are genuinely good — "a query that could not run is never reported as absent data" is exactly the invariant Bugbot's finding was about, and it is now written down where the next change will trip over it.

Three corrections so the anchor matches the code on the commit that establishes it:

1. **"bounded"** overstates the time range. `dataCheckTimeFrom` / `dataCheckTimeTo` are unvalidated strings and the constant is only substituted when they are empty, so it is a default, not a bound.
2. The `data-check` row also asserts a check runs "on click alone". That holds for this component, but the interactive engine's "Do it" resolves an author's `reftarget` document-wide and clicks every match, including these run buttons — the plain-text form (`Run check`) is the one the schema description recommends. I confirmed three checks firing from one unrelated multistep. This is a pre-existing engine behaviour that already reaches Insert, quiz check and terminal connect, so it is not yours to fix here and I would not gate on it — but an anchor claiming user-action-only should say "user action, including an engine-driven click" rather than imply the surface is closed.
3. An authored `defaultValue` leaves the advisory run button enabled on mount with no user pick at all (`input-block.tsx:156-163`). Worth a line, since "only against a data source the picker currently offers" is true while "the user chose it" is not.

docs/design/CONCERNS.md
```diff
@@ -113,10 +115,11 @@ Rules:
...
+| `data-check`             | #1597 → this PR | `runDataCheckQuery` (`src/lib/datasource/run-data-check-query.ts`) is the sole owner of query execution and its cost caps (100 data points, 15s abort, bounded range). `useDataCheck` owns the check lifecycle and its `AbortController`. A check runs on user action only, never on a polling cadence, and only against a data source the picker currently offers. A query that ran and found nothing (`'no-data'`) and one that could not run (`'error'`) are distinct states, and the author's failure message speaks only for the first. Abandoning a check aborts it, and an aborted check reports no outcome and completes nothing. `dataCheckQuery` alone enables the check; `dataCheckBlocking` alone decides whether a failure gates a section, and is the only shape under which an `input` block becomes a tracked step. Invariants in `purpose`. |
```
tomglenn: Two things on this row:

- **`#1597 → this PR`** will not resolve once merged. Peer rows use numbers (`#1298 → #1300`); this wants `#1597 → #1612`.
- **"100 data points, 15s abort, bounded range"** — the first is inert on the Prometheus path (instant queries take no `step`; cost is cardinality) and the third is a default rather than a cap. The 15s abort is real and correctly implemented. Suggest "15s abort, 100-point cap where the query type honours it, default 1h range".

Worth noting your own rule at line 118 warns that rows arriving in the contract-establishing PR miss the fracture window. This PR adds both rows, which is unavoidable for a new capability — but it does mean these two are the version everything downstream will be checked against, so they are worth getting exactly right now.

tomglenn (changes requested):
Requesting changes, but the design here is sound and most of what follows is small. Worth saying what held up first, because I attacked this hard and most of it did not move.

**The split is right, and I verified the premise rather than taking it.** `STEP_TYPE_PARSE_KEYS` really is built from parse keys and the counters really do read element type only, so registering `InputBlock` would have counted every text and boolean input. The new type lands in both denominators and sections re-register their real count. Flip-safety works in both directions: an advisory block gets no `stepId` at all, so advisory→blocking cannot be satisfied by a stale record, and blocking→advisory orphans self-heal through `reconcileSection`. `EXPECTED_BLOCK_TYPE_COUNT` is unchanged for the right reason — the guard counts union discriminants and `unwrap()` follows `innerType`, so the new `.superRefine` does not hide anything.

**The lifecycle is solid.** `useDataCheck` has a real generation guard, not a bare abort — I tested second-resolves-first, a late error arriving mid-second-run, unmount mid-flight, and a datasource swap mid-flight, and none of them wrote stale state. Both Bugbot fixes genuinely hold: `reset()` aborts synchronously before `markSkipped`, so Redo is not inert and a late run cannot reach `markComplete`; `'aborted'` never enters state or telemetry; the author's copy is reachable only on `'no-data'`. Hydration restores the name rather than the uid, a durable pass renders with zero queries, and a skip reads "Skipped".

**Security largely holds.** The abort is real and propagates, nothing runs a check on mount, hydration, redo or section-run, concurrency is guarded, `pausesSectionRun` genuinely stops Do Section, and F1-F6 are clean — the failure messages render as JSX text nodes. Telemetry carries no query text, no error string and no data source name. The cost caps cannot be overridden from authored JSON.

**Test coverage is strong in the places that usually rot.** Twenty-six mutations, twenty caught with messages naming the right cause — including rows-vs-frames, both parser-split directions, the `'no-data'`/`'error'` separation, skip-aborts, and the unsupported-type gate. The jsdom `<select>` trap you found and called out is genuinely closed: both stale-pick tests now assert the button is disabled and that no query runs.

Four things I would fix before this merges, all new in this diff.

## Must fix

1. **The feature can ship entirely dead with a green pipeline.** Making the `datasource-check-step` arm of `content-renderer.tsx` return `null` leaves all 432 suites and 6886 tests passing. The parser test asserts the element type and the component test renders the component directly; nothing joins them. One parse-then-render test closes it. Inline at `content-renderer.tsx:1277`.

2. **A passed check is retracted only when the user drives the picker directly.** Any other writer of the same `variableName` swaps the data source under a completed check and nothing resets it — the badge still reads "Data available" and the section keeps counting the completion. Three reproductions inline, and it is reachable through this PR's own two-host pairing. `handleDatasourceChange` also never calls `onStepReset`, and `persistReset` is a no-op for a section-owned step, so even the visible path only retracts for a standalone one. Inline at `datasource-check-step.tsx:214`.

3. **The cost caps are not pinned as a contract.** Every assertion compares against the constant itself, so only presence is tested and never the value. `maxDataPoints` 100 → 100000 passes. The 15s abort → 900s passes. `defaultFrom` `now-1h` → `now-5y` passes. This PR states the caps are load-bearing and should not be relaxed without a deliberate decision, and `CONCERNS.md` lists changing them as a fracture — right now a one-line edit relaxes any of them silently. Assert the literals.

4. **`requirements` on a datasource input now behaves two different ways depending on `dataCheckBlocking`.** The dead prop is pre-existing — declared in `InputBlockProps`, never destructured, identical on the base commit — so that half is not yours. But `DatasourceCheckStep` honours it through `useStepChecker`, so identical authored `requirements` gates a blocking check and is silently inert on an advisory one. That inconsistency is new here, and it is an authoring trap with no signal. The PR body says this leaves "a documented inconsistency", but `json-guide-format.md:763` still describes the field as working and nothing anywhere says it is inert. Either honour it in both hosts or document it where authors will read it.

## Should fix

5. `pausesSectionRun` is the change you call load-bearing, and its only real consumer — the `|| stepInfo.pausesSectionRun` at `interactive-section.tsx:813` — has no test. Removing it there leaves all 56 relevant suites green. Only the schema constant is pinned. Two of my passes flagged this independently.

6. An empty or whitespace-only `dataCheckQuery` validates, then degrades to a plain passive input — exactly the "reads as a check that never runs" outcome the refinement exists to prevent. One-line Zod fix inline, plus two related gaps in the same block.

7. Loki is the only query type sent with no result bound. One-line `maxLines` inline.

8. Test-id factories are called with `variableName` in one host and `renderedStepId` in the other, so one factory has two key spaces and can collide.

## Smaller

- `CONCERNS.md` wording: "bounded range" is a default not a bound; "100 data points" is inert on the Prometheus path; `#1597 → this PR` will not resolve post-merge. All inline.
- "unreachable inside `conditional` blocks" is true of the block editor and false of raw JSON — I parsed a blocking check inside an inline conditional and it emitted a real tracked step. The resulting count skew (a completion outside any denominator, through the one percentage path with no clamp) reproduces identically on the base commit with a quiz or challenge, so it is genuinely pre-existing and not yours to fix. The claim is the only part that needs changing: "unreachable via the block editor".
- Making `'skipped'` sticky in `completion-store.ts:355` applies to every step type, not just data checks. It is probably more correct, but it is undisclosed scope and wants a non-datasource test.
- `getDatasourceOptions` in the new shared module has no consumer outside its own test. The extraction itself did work — both hosts import `filterDatasourcesByType` and `toDatasourceOptions` — so the duplication really is closed; this is just a third export nothing calls.
- The `'checking'` state has no test in either host, so the spinner and the double-click guard are unpinned. The advisory host is the thinner half generally: its stale-pick and `'aborted'` guards are both mutation-survivable, though the tracked twins are covered.
- One hydration test is vacuous — `does not carry one guide's pick into the next` unmounts and remounts, so it never exercises the guard; it passes with the fix reverted. Related: your note that the `GuideResponseProvider` fixes fail 2 tests when reverted is 1 for that half.
- The `clears the stored pick when the user clears the picker` test exercises a path the real UI cannot reach — `Combobox` only emits `null` when `isClearable` is passed, and the source does not pass it.
- Telemetry: a 15s-abort network operation with no `*_ms` measurement, and timeout and HTTP error both collapse into `outcome: 'error'` — so the distinction your second commit exists to make is invisible in the data.
- `interactive-types.md:260` heading wants sentence case and "data source"; `datasource-check-step.tsx:186`'s "Handle reset trigger from parent section" restates the code. Comment density is otherwise fine at ~3.4%.

## Chased and cleared, so nobody re-opens them

- **The engine "Do it" fan-out can fire these checks** — an author's `reftarget` resolves document-wide and clicks every match, and `Run check` is exactly the plain-text form the schema description recommends. Three checks fired from one unrelated multistep. But this is a pre-existing engine class that already reaches Insert, quiz check and terminal connect; excluding plugin DOM from engine targeting would break guides that teach Pathfinder itself, and the content marker is not set in sidebar mode anyway. Not this PR's to fix — noted on the `CONCERNS.md` row so the anchor does not overclaim.
- **Unbounded time ranges** are near-harmless on Prometheus, since an instant query ignores `from` entirely. Only Loki is exposed, hence the one-line fix rather than schema validation.
- **No cardinality cap** is real but not expressible client-side; tenant limits are the only defence.
- The 42 lines removed from `assistant-integration/tools/types.ts` are a tier-0 relocation of the datasource type helpers, not abandoned AI-check scaffolding — the deferral is clean.

`npm run check` reproduces green here: 432 suites, 6886 tests. `--runInBand --detectOpenHandles` shows no act() warnings and no leaked handles, and three randomized runs showed no flake.

<sub>Reviewed with parallel passes over security, completion accounting, lifecycle, test quality and schema/docs, then an adversarial round that tried to refute each finding. Several of my own initial findings did not survive it and were dropped rather than raised; two severity calls were corrected downward the same way.</sub>

src/components/content-renderer/content-renderer.tsx
Jayclifford345: Fixed in 2994cb5f. You found the one hole where a green pipeline meant nothing, and the reproduction was the convincing part — I ran the same mutation before fixing it and got the same 432/6886.

`content-renderer.datasource-check.test.tsx` takes guide JSON through `parseJsonGuide` and into the renderer, asserting on the DOM: a blocking check reaches it as a tracked step, an advisory one as a passive picker, and a plain picker offers no check at all. Making the arm return `null` now fails 2 of the 4.

It also caught something on the way in. The advisory assertion failed on first run because the two hosts build their run-button id from different keys — the collision you flagged on `input-block.tsx:473`, surfacing as a test failure rather than as an argument. Fixed there.

On `stepIndex` / `totalSteps`: wired rather than dropped. The arm forwards them like the other eight, and `DatasourceCheckStep` now spends them on `buildInteractiveStepProperties`, which is what every other tracked step does with them — so `data_check_run` / `_passed` / `_failed` / `_skipped` carry `current_step`, `total_document_steps`, `completion_percentage` and the section context instead of a bare `step_id`. That also closes half of your telemetry note: the events carry `duration_ms` now, and timeout no longer collapses into `outcome: 'error'`.

src/lib/datasource/run-data-check-query.ts
Jayclifford345: Took the suggestion as written. Loki now goes out with `maxLines: DATA_CHECK_QUERY_LIMITS.maxDataPoints`, so the one call site that owned a limits constant and did not use it does.

The per-type test asserts the literal on both bounded types now — `maxLines: 100` for Loki, `limit: 100` for Tempo — rather than `toMatchObject` against the constant, which is the same weakness you raised at the review level.

The wording correction is right and I have taken it in all three places, because "capped at 100 data points" was doing real work in the PR body and it is not true of the documented example:

- `CONCERNS.md` anchor: "15s abort, 100-point cap where the query type honours it, default 1h range".
- `json-guide-format.md` now says it plainly, with your example: an instant query takes no `step`, so its cost is series cardinality, `{__name__=~".+"}` is one unbounded high-cardinality query against whatever tenant the user picked, and no client-side fix exists — keep the query selective.
- The PR body is updated to match.

src/types/json-guide.schema.ts
Jayclifford345: All three taken.

**Empty query** — your suggestion as written, plus `.max(2000)` for the length gap. `''` and `'   '` are now rejected at the field rather than degrading to a passive picker, which was exactly the outcome the refinement below was written to prevent. Two tests cover it.

**`id` on a blocking check** — added, and worth flagging what it costs, because it is a real divergence rather than a free win: no other tracked type requires one. A quiz or a terminal step with no `id` gets `${sectionId}-quiz-${n}`, which has the same orphan-on-edit problem and is not rejected. I still think requiring it here is right — the surface is new, so nothing breaks, and the doc already told authors to do it — but it does mean data checks are stricter than their siblings for a reason that applies to all of them. An advisory check is exempt: it is not a tracked step and stores nothing.

One consequence worth naming: the parser's derive-a-stepId-in-a-section fallback is now unreachable through validated JSON, so `derives a stepId inside a section when the author gave none` was asserting on a path authors cannot take. It is now a rejection test plus one that pins the authored id surviving into a section.

src/types/json-guide.schema.ts
Jayclifford345: Agreed on not putting Grafana time-expression parsing into Zod — the grammar is wide, and a validator that accepts `now-10y` but rejects `banana` would still not be a bound, so it buys very little.

So this is a wording fix, as you framed it. `CONCERNS.md:122` now reads "default 1h range" alongside the two enforced numbers, and `json-guide-format.md` says outright that `dataCheckTimeFrom` / `dataCheckTimeTo` are defaults rather than bounds and that nothing rejects `now-10y`. The description here was already accurate and is unchanged; it was the anchor claiming more than the code that had to move.

The one place the unbounded range actually bites is Loki, and that is now bounded by `maxLines` per your other comment — so what remains is a documented default, not a silent cap that isn't one.

src/components/interactive-tutorial/datasource-check-step.tsx
Jayclifford345: Fixed in 2994cb5f, and the fix ended up a layer below where you pointed — which I think makes your finding sharper rather than narrower.

Your diagnosis was that this handler is the only thing calling `reset()`. The reason that was possible is that the verdict was free-standing state that happened to be about a data source. So `useDataCheck` now derives it instead: the verdict carries the uid it was produced against, and anything not matching the current pick reads as `'idle'`. No writer has to remember to clear it, because there is nothing to clear.

That alone fixes reproductions 2 and 3 — the stale failure banner and the `clearForGuide` case — **and it fixes them in the advisory host too**, which you did not have to point out but which has the identical bug: `InputBlock`'s picker is updated by the storage-event handler, and only its own `handleDatasourceChange` was calling `reset()`.

Reproduction 1, the serious one, needed the completion store as well, so that is a small effect keyed on the resolved name:

- it fires for **any** writer, not this handler;
- it calls `onStepReset` alongside `persistReset`, so a section-owned step retracts — you were right that even the visible path only worked standalone;
- a name arriving where there was none is hydration, not a change, so a durable pass survives a reload. That guard is the part I would have got wrong without writing your reproduction down first.

`handleDatasourceChange` is now just the write. `handleRedo` and the retraction share one `retract()`.

Two other things fell out. `DatasourceCheckStep`'s `Combobox` did not pass `isClearable`, so the test you flagged at the review level (`clears the stored pick when the user clears the picker`) was exercising a path the UI could not reach — it does now, and clearing retracts through the same effect. And I did not previously have a way to simulate a foreign writer at all; the mock provider now exposes one, which is what all of this is asserted through.

Eight tests: five in the tracked host (retracts a section-counted pass, retracts a standalone one, drops the stale failure, retracts on a guide clear, and leaves a hydrated pass alone), three in the advisory host. Mutations: deleting the `retract()` call fails 5; making the verdict derivation unconditional fails 4.

src/components/interactive-tutorial/input-block.tsx
Jayclifford345: Fixed, though not by picking one key — the two hosts genuinely do not have one. `DatasourceCheckStep` has a `stepId`; `InputBlock` has no step at all, so `variableName` is the only stable key it owns. Forcing both through one factory is what created the shared key space.

So they are now separate id spaces with distinct prefixes:

```ts
dataCheck: {           // the tracked step, keyed by stepId
  runQueryButton: (stepId) => `datasource-check-run-${stepId}`,
  ...
},
advisoryDataCheck: {   // the passive input block, keyed by variableName
  runQueryButton: (variableName) => `input-data-check-run-${variableName}`,
  failure: (variableName) => `input-data-check-failure-${variableName}`,
},
```

A collision is now impossible rather than unlikely, and each factory's parameter name is what it is actually passed — which also settles your `testIds.ts:182` comment without renaming anything.

On the redundancy: `testIds.interactive.datasourcePicker` and `testIds.dataCheck.datasourcePicker` are both live, but for different hosts — the first is the advisory picker (variableName), the second the tracked one (stepId). Both are used exactly once. I have left them, since the split above now makes which-is-which legible; say the word if you would rather see the advisory one moved under `advisoryDataCheck` for symmetry.

This is also the finding that the new parse-then-render test caught independently — its advisory assertion failed on the first run because it looked for `datasource-check-run-<variableName>`.

src/constants/testIds.ts
Jayclifford345: Fixed at the call site rather than by renaming, so the signature's invariant is now one the code keeps.

`testIds.dataCheck.*` is only ever called with a rendered step id — the two call sites that passed a `variableName` moved to a new `testIds.advisoryDataCheck.*` group whose parameter genuinely is named `variableName`. Both factories now document what they are actually handed, and the shared key space that made the collision possible is gone. Detail on the split in the reply on `input-block.tsx:473`.

docs/design/CONCERNS.md
Jayclifford345: All three corrections taken — the row now reads:

> ...a check runs on a click and never on a timer or a requirements token, only against a data source the picker currently offers...

1. **"bounded"** is gone from the anchor; it says "default 1h range" alongside the two numbers that are actually enforced.

2. **The engine fan-out.** Recording it rather than dropping it, since you confirmed it with three checks firing from one unrelated multistep. The anchor now says a check runs "on user action only — including an engine-driven click, which the interactive engine can dispatch document-wide". Agreed it is not this PR's to fix: the class already reaches Insert, quiz check and terminal connect, and excluding plugin DOM from targeting would break guides that teach Pathfinder itself.

3. **`defaultValue`.** The anchor now says the guarantee is "offered", not "chosen" — and the impact questions gained *"Does an authored `defaultValue` now let a check run against a data source the user never picked?"*, so the next change trips over it rather than reading the prose.

I also added one invariant the review turned up, since it is the load-bearing half of the fix for your `datasource-check-step.tsx:214` finding: a verdict and any completion it earned belong to the data source they were run against, and are retracted whenever the resolved pick changes — whichever writer changed it. That was true of nothing before this round.

docs/design/CONCERNS.md
Jayclifford345: Both fixed.

- `#1597 → #1612`, matching the peer rows.
- Caps wording is yours nearly verbatim: **"15s abort, 100-point cap where the query type honours it, default 1h range"**. Loki now honours it too, per your `run-data-check-query.ts:49` comment, so that phrasing covers three of the four types with Prometheus the documented exception.

Your point about the fracture window is the one I have least ability to fix and the most reason to take seriously — these two rows are the version everything downstream gets checked against, and nothing was watching while they were written. So beyond your three corrections I have tightened the anchor where this review proved it was already wrong: the retraction invariant (a verdict belongs to the data source it was run against, and is retracted by any writer of the pick) did not hold in the code until 2994cb5f, and would have been anchored as true.

If you would rather the whole `data-check` anchor land in a follow-up, once the capability has one round of real use behind it, I would take that trade — the routing row at line 67 is the half that needs to exist now, and the contract half is the one that benefits from waiting.

Jayclifford345:
Thanks @tomglenn — everything is addressed in 2994cb5f, and `main` is merged in (the only conflict was the `CONCERNS.md` anchor table, where your `#1616` row and the `data-check` row landed on the same lines).

Each inline comment has its own reply. This covers the four review-level items that had no inline anchor, and the "smaller" list.

## Must fix

**3. The cost caps are not pinned as a contract.** You were right that only presence was tested. There is now a test asserting the literals — `maxDataPoints` 100, `timeoutMs` 15_000, `defaultFrom` `now-1h`, `defaultTo` `now` — so each of your three mutations fails. The per-type test also asserts `maxLines: 100` and `limit: 100` as literals rather than `toMatchObject` against the constant.

**4. `requirements` behaves two ways depending on `dataCheckBlocking`.** Documented rather than honoured in both, because honouring it on `InputBlock` means wiring `useStepChecker` into a passive block that has no step, no completion and no skip semantics — a bigger change than this PR should carry, and one that would want its own review. `json-guide-format.md` now says outright that the field is honoured **only** when `dataCheckBlocking` is set, and that on text, boolean and advisory datasource inputs it validates, parses and then does nothing. The field's own table row says the same, so an author sees it without reading the prose. The dead prop itself stays pre-existing and untouched.

## Should fix

**5. `pausesSectionRun` has no consumer test.** Added to the runner tripwire, mirroring the guided-pause case: a section of plain → data check → plain stops at the check, does not fire `interactive-section-completed`, and — the part that matters — does not credit the check itself. Reverting the `|| stepInfo.pausesSectionRun` fails it. The harness gained a `DatasourceCheckStep` mock alongside the other eight.

## Smaller

- **`'checking'` has no test in either host** — fixed in both: the spinner renders, the button is disabled, and pressing it twice more still issues exactly one query.
- **The vacuous hydration test** — `does not carry one guide's pick into the next` now rerenders the same provider with a new `guideId` instead of unmounting, so it exercises the render-time guard. Your correction was right: it passed with the fix reverted. It now fails, and the `GuideResponseProvider` pair fails 2 rather than 1.
- **The `isClearable` test exercised an unreachable path** — the tracked step's `Combobox` now passes `isClearable`, so it is reachable, and clearing retracts the completion through the new path.
- **`getDatasourceOptions` had no consumer** — deleted, along with its test. The two functions both hosts actually import stay.
- **Telemetry** — events carry `duration_ms`, and `runDataCheckQuery` returns a `failureKind` so a timeout reports `outcome: 'timeout'` rather than collapsing into `'error'`. The distinction the second commit exists to make is now visible in the data.
- **`interactive-types.md:260`** — "Data source picker with a data check". **`datasource-check-step.tsx:186`** — comment deleted.

## Taken as read, and acted on

- **`'skipped'` stickiness is undisclosed scope** — correct, and now disclosed in the PR body with a note that it applies to every step type. The existing test is already store-level and non-datasource (plain `step-1` / `section-x`), so I have left it rather than adding a second one; say the word if you want it exercised through a quiz too.
- **"unreachable inside `conditional` blocks"** — PR body now says "unreachable **via the block editor**", with your finding that raw JSON reaches it and that the count skew is pre-existing.

## One thing to flag

`npm run check` is green here — 437 suites, 6980 tests, thresholds met — **except** `tests/e2e-runner/utils/guide-runner/requirements.test.ts`, which fails 3 tests on `Unexpected testId requested: learning-paths-badge-toast`. I reproduced it on a clean `origin/main` worktree: it is pre-existing from #1617 landing on top of #1615, not from this branch. Worth a separate fix — `main` is currently red.

src/components/block-editor/forms/InputBlockForm.tsx
```diff
@@ -125,6 +132,15 @@ export function InputBlockForm({
...
         ...(inputType === 'datasource' && datasourceFilter.trim() && { datasourceFilter: datasourceFilter.trim() }),
+        // The query is what enables the check, so the rest is dropped without it —
+        // the schema rejects a check field that could never take effect.
+        ...(hasDataCheck && {
+          dataCheckQuery: dataCheckQuery.trim(),
+          ...(dataCheckFailureMessage.trim() && { dataCheckFailureMessage: dataCheckFailureMessage.trim() }),
+          ...(dataCheckTimeFrom.trim() && { dataCheckTimeFrom: dataCheckTimeFrom.trim() }),
+          ...(dataCheckTimeTo.trim() && { dataCheckTimeTo: dataCheckTimeTo.trim() }),
+          ...(dataCheckBlocking && { dataCheckBlocking }),
+        }),
```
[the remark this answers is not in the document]
Jayclifford345: Fixed in 1c37ebda. Correct on both halves, and the severity call is right — this made the authoring path dead, not merely lossy.

Both halves are fallout from requiring an explicit `id` on a blocking check in the previous commit, which I added on review feedback without checking who else writes these blocks. `InputBlockForm` never wrote `id`, so the editor could not produce a schema-valid blocking check at all, and editing an existing one silently dropped the id its completion records are keyed on — that second half was already true before this PR for every input block, and my change turned it from lossy into invalid.

It now carries `initial?.id` through on every submit, so an edit preserves whatever the guide already had. When a blocking check is created without one, it derives `check-<variableName>` from the field that is already a validated identifier. That is deliberately different from the parser's positional fallback, which is the thing the `id` requirement exists to avoid: a derived id written into the JSON is durable across later edits, a positional one is not.

Three tests, all failing without the fix: the emitted blocking check parses against `JsonInputBlockSchema`, an advisory check still emits no `id` (it stores no completion), and an edit keeps the authored id rather than deriving over it.

src/docs-retrieval/json-parser.ts
```diff
@@ -900,13 +900,42 @@ function convertQuizBlock(block: JsonQuizBlock, path: string, stepContext?: Step
...
+          failureMessage: block.dataCheckFailureMessage,
+          timeFrom: block.dataCheckTimeFrom,
+          timeTo: block.dataCheckTimeTo,
+          requirements,
+          skippable: block.skippable ?? false,
+        },
+        children: promptElements,
+      },
+      hasInteractive: true,
+    };
```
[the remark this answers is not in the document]
Jayclifford345: Fixed in 1c37ebda, by rejecting rather than forwarding.

You are right that the field is inert on the blocking host and that the asymmetry is invisible to an author. The question is which way to resolve it, and forwarding is the wrong direction here: a blocking check gates a section and writes durable completion, so seeding the pick would let a step complete against a data source the user never saw. That is a weaker guarantee than the one the contract anchor claims.

So the schema now rejects `defaultValue` when `dataCheckBlocking` is set, with a message saying why. This is the same principle as the refinement immediately above it — a field that would be silently inert is rejected, not ignored — and no existing guide can hit it, since `dataCheckBlocking` is new in this PR. Advisory checks honour `defaultValue` exactly as before.

It also sharpens something @tomglenn raised on `CONCERNS.md:99`: an authored `defaultValue` leaves the advisory run button enabled on mount with no user pick. That is now true of the advisory host only, and the anchor says so — "offered" is the guarantee on an advisory check, "the user chose it" holds on a blocking one. `json-guide-format.md` documents both.

Two tests: the blocking combination is rejected on the `defaultValue` path, and the advisory one still validates.

src/components/block-editor/forms/InputBlockForm.tsx
```diff
@@ -101,14 +107,25 @@ export function InputBlockForm({
...
         .map((r) => r.trim())
         .filter((r) => r.length > 0);
 
+      const isBlockingCheck = hasDataCheck && dataCheckBlocking;
+      // The form has no id field, so an edit would otherwise drop the one the
+      // guide already carries — and a blocking check is schema-invalid without
+      // one, because its completion records are keyed on it. Derived from
+      // `variableName`, which is already a validated identifier, so the id is
+      // written into the JSON and survives later edits.
+      const blockId = initial?.id ?? (isBlockingCheck ? `check-${slugify(variableName.trim())}` : undefined);
```
tomglenn, line 124: Nit, worth taking while you're here: `slugify` is the wrong transform for this input and it is the only thing that makes the derived id non-injective.

`variableName` is already gated on `/^[a-zA-Z_][a-zA-Z0-9_]*$/` (submit is blocked on `hasValidVariableName`), so slugifying can only destroy information. Verified collisions: `metrics` and `_metrics` both give `check-metrics`; `metricsDs` and `metricsds` both give `check-metricsds`; `a_b` and `a__b` both give `check-a-b`; and `variableName: '_'` gives the id `"check-"`, which the schema accepts.

```suggestion
      const blockId = initial?.id ?? (isBlockingCheck ? `check-${variableName.trim()}` : undefined);
```

Safe to take: nothing assumes slug-safety for a block id — `DatasourceCheckStep` emits no `data-step-id`, and there is no anchor or URL use.

**Deliberately not asking for more than the one-liner.** I originally had this as a must-fix on the grounds that two colliding blocking checks share one completion record, which does reproduce — the store keys `contentKey → sectionId → stepId` and the section reducer tests `completedSteps.has(s.stepId)`, so they alias in both. But the authoring input it needs (two blocking checks, same section, variable names differing only by case or underscores, both ids editor-derived) is fuzzing-grade rather than a realistic mistake. And the genuinely reachable duplicate-id path is one click away and untouched here: `useBlockEditor.ts:592`'s `duplicateNestedBlock` deep-clones the `id` verbatim. That is pre-existing on main, and it means the real gap is duplicate-id validation rather than this derivation — worth its own issue rather than growing this PR.

Also: I asked for the explicit-`id` requirement last round without checking who else writes these blocks, which is what made the editor path invalid rather than merely lossy. That one is on me, and your fix is the right shape.

src/test-utils/data-check-stubs.tsx
```diff
@@ -0,0 +1,61 @@
...
+export const loggerStub = {
+  logger: { debug: jest.fn(), info: jest.fn(), warn: jest.fn(), error: jest.fn(), exception: jest.fn() },
+};
+
+export const analyticsStub = {
+  reportAppInteraction: jest.fn(),
+  buildInteractiveStepProperties: (
+    baseProperties: Record<string, string | number | boolean>,
+    stepContext: Record<string, unknown>
+  ) => ({ ...baseProperties, step_id: stepContext.stepId }),
```
tomglenn, line 46: **The one thing I'd fix before merge.** This stub discards the two fields the wiring exists to deliver, so the fix for last round's tripwire finding has its own tripwire.

`buildInteractiveStepProperties` is stubbed to `{ ...baseProperties, step_id: stepContext.stepId }`. `stepIndex` and `totalSteps` never survive it, so nothing asserts that the renderer arm forwards them or that `DatasourceCheckStep` spends them. I removed the forwarding from `content-renderer.tsx` and the suite stayed green — same shape as the `case` returning `null` from last round, one layer in.

Cheapest close: have the stub pass through what it is given (`current_step` / `total_document_steps` from `stepContext`), and assert on one telemetry payload in the tracked host. `prop-forwarding.test.ts` only covers `interactive` blocks today, so it will not catch this either.

Worth it specifically because the value of this wiring is entirely in the telemetry payload — if it silently stops forwarding, nothing breaks visibly and the data just gets thinner.

src/components/interactive-tutorial/datasource-check-step.tsx
```diff
@@ -0,0 +1,422 @@
...
+  className?: string;
+
+  // Unified state management props (passed by parent section)
+  stepId?: string;
+  isEligibleForChecking?: boolean;
+  onStepComplete?: (stepId: string) => void;
+  onStepReset?: (stepId: string) => void;
+  resetTrigger?: number;
+
+  // Step position tracking
```
tomglenn, line 46: Two QC8 banner comments in a new file — `// Unified state management props (passed by parent section)` and `// Step position tracking` — both restate the prop names directly beneath them. Worth dropping since the file is new and the rest of its commenting is genuinely good: the `verdictOwner` hydration-versus-change block, the `resetTrigger` note, and the `'no-data'`/`'error'` rationale in `use-data-check.ts` all state invariants a reader cannot recover from the code, which is exactly the bar.

One trim-on-touch miss in the same family: `input-block.tsx:237`'s `// Handle datasource selection change` sits directly above the function you rewrote this round, with `// Handle reset/clear`, `// Reset to defaults`, `// Handle save` and `// Handle skip` on adjacent declarations.

docs/developer/interactive-examples/interactive-types.md
```diff
@@ -257,17 +257,42 @@ If you specify `"requirements": ["exists-reftarget"]` on a multistep, also set `
 
 See [guided-interactions.md](./guided-interactions.md) for detailed documentation.
 
+### Data source picker with a data check
+
+- **Purpose**: confirms the data source the user picked actually holds the data the guide teaches against.
+- **Behavior**: the user picks a data source and presses **Run check**. The author's query runs once, on that click, and passes when it returns at least one row. Without `dataCheckBlocking` the result is advisory; with it, the block becomes a tracked step that completes only on a pass.
```
tomglenn, line 263: "completes only on a pass" is not true when `skippable` is set — and the example immediately above sets it. `json-guide-format.md` gets this right ("or is skipped if `skippable`"); worth matching it here.

```suggestion
- **Behavior**: the user picks a data source and presses **Run check**. The author's query runs once, on that click, and passes when it returns at least one row. Without `dataCheckBlocking` the result is advisory; with it, the block becomes a tracked step that completes on a pass, or is skipped when `skippable` is set.
```

docs/developer/interactive-examples/json-guide-format.md
```diff
@@ -816,7 +826,48 @@ Collects user responses that can be stored as variables and used elsewhere in th
...
+`dataCheckBlocking` is the one thing that changes the block's behaviour in a guide:
+
+| `dataCheckBlocking` | The block is                | A failing check                                                    |
+| ------------------- | --------------------------- | ------------------------------------------------------------------ |
+| unset (default)     | passive, like any input     | reports what it found; the user carries on                         |
+| `true`              | a tracked, completable step | holds the section up until it passes, or is skipped if `skippable` |
+
+Turn blocking on deliberately. It makes the block count toward the section's step total, so an author who cannot guarantee the data exists on every instance should pair it with `skippable`.
+
+Only Prometheus, Loki, Tempo, and Pyroscope can be queried. If the user picks a data source of any other type, the block says so instead of offering the button rather than silently passing. Pyroscope queries are written as `<profileTypeId>|<labelSelector>`.
```
tomglenn, line 862: This clause has two competing constructions spliced together — "says so instead of offering the button rather than silently passing" — and I flagged it last round without being specific enough to act on, so that one is on me.

```suggestion
Only Prometheus, Loki, Tempo, and Pyroscope can be queried. If the user picks a data source of any other type, the block says so and offers no button, rather than silently passing. Pyroscope queries are written as `<profileTypeId>|<labelSelector>`.
```

Smaller, same file: new prose writes "datasource picker" as one word while the new heading in `interactive-types.md` uses "data source". The style rule wants two words in prose.

docs/design/CONCERNS.md
```diff
@@ -72,32 +73,33 @@ Column key — **cat**: `AO`=always-on · `sub`=subsystem · `xcut`=cross-cuttin
...
...                                                                                                                                                                                                                                                                     | New `/api/ds/query` callers; changes to `DATA_CHECK_QUERY_LIMITS`; running a check from an effect rather than a handler; removing the offered-options validation of a restored pick; making `input` tracked for any shape other than a blocking check                                                                                                                                                                                                                                                                                                                    | `npm run test:ci`; the 11-point browser checklist in the PR that established this concern                                                                                                                                                          | `interactive-engine`, `requirements-manager`, `guide-schema-and-contracts`, `analytics-and-telemetry`, `reversibility-and-one-way-door` |
```
tomglenn, line 99: The `verify` column cites "the 11-point browser checklist in the PR that established this concern" — the PR body's how-to-verify list has 10 numbered checks. Either correct the count or drop the number, since the checklist will keep drifting and an anchor that miscounts it is worse than one that just points at it.

Everything else I asked for on this row is here and reads well — the caps wording, the engine-driven-click caveat, the "offered" versus "chose" split, and the new `defaultValue` impact question. The extra invariant you added about a verdict belonging to the data source it was run against is the right thing to have written down; it is the part of this round that will stop the next change reintroducing the bug.

tomglenn (approved):
Approving. Everything from the last round is genuinely addressed, and I verified it by re-running my own mutations rather than reading the replies — your cited numbers match mine in every case I checked.

## Verified, with the mutation evidence

- **The dead-feature tripwire is closed.** `content-renderer.datasource-check.test.tsx` goes through the real parser (a JSON string into `ContentRenderer`, reaching `parseJsonGuide`), not a hand-built element tree. Making the arm return `null` fails 2 tests, exactly as you said; mutating the parser so a blocking check emits `input-block` fails 6 across 2 suites, so it catches that direction too.
- **The caps are pinned as literals.** All four mutations caught: `maxDataPoints` 100→100000 (3 fail), `timeoutMs` 15s→900s (1), `defaultFrom` `now-1h`→`now-5y` (1), `defaultTo` `now`→`now+1d` (1). None survived. Loki genuinely sends `maxLines` now.
- **The retraction refactor is sound, and better than what I asked for.** I asked you to patch the handler; deriving the verdict against the current uid eliminates the class instead, and it fixed the advisory host's identical bug that I had not spotted. No false-valid reading survived probing: A→B, B→A, A→null, null→A, A→A, unresolvable names, and `undefined`/`null` uid on either side all mask correctly, and the `?? null` normalisation closes the accidental `undefined === undefined` match. Retraction fires exactly once per logical change and the store is idempotent for repeated resets. A foreign write mid-flight aborts the query and the late verdict is masked. It also removes an R14 violation by moving the verdict into render. Your two mutation counts reproduce exactly (5 and 4).
- **`'checking'`, the hydration test, `isClearable`, the test-id split, dead code** — all verified. Removing the `disabled` guard fails 4 tests across both hosts; disabling the render-time guard fails 2, your number. The two id namespaces now have disjoint prefixes, so the collision is structurally impossible rather than merely unlikely — and leaving both `datasourcePicker` entries is right, since merging them would recreate the shared key space.
- **Telemetry carries no leak.** Bounded attributes only — `datasource_type` (5 values), `blocking`, `outcome` (4), `duration_ms`, constant `interaction_location`. No query text, no error string, no data source name.
- **`requirements`: documenting is the right call.** The table row is where an author actually looks, and wiring `useStepChecker` into a block with no step, completion or skip semantics is genuinely a separate change. The honest end state is schema-level rejection on non-blocking inputs, which is its own PR.
- **`defaultValue` rejected rather than forwarded** is the right direction, for the reason you gave: a blocking check writes durable completion, so seeding a pick the user never saw is a weaker guarantee than the anchor claims. Demo guide and all 10 packages still validate.

## Before merge — one item

The stub in `data-check-stubs.tsx` discards `stepIndex` / `totalSteps`, so the fix for last round's tripwire has its own tripwire: I removed the forwarding and the suite stayed green. Inline, with the cheapest close.

## Nits, take or leave

The `slugify` one-liner, the two QC8 banners in the new file, and four prose corrections — all inline. Plus three stale statements in the PR body itself, which is the one place last round's wording fix did not land:

- **Reversibility still says "bounded range and result size"** — the exact phrase we retired. It now contradicts your own body two sections up and `json-guide-format.md`'s "defaults, not bounds; nothing rejects `now-10y`".
- It still credits `getDatasourceOptions` as "moved out of `input-block.tsx` … unchanged". You deleted it this round.
- The how-to-verify list is 10 items; `CONCERNS.md` cites 11.

## Things I raised and then talked myself out of — so nobody reopens them

- **A foreign write un-sticking a skip and rewinding the section tail.** Reproduces, and the tail reset is real and precise. But that is pre-existing Redo semantics — `interactive-section.tsx`'s reset handler is untouched here, so this PR adds a trigger, not the rewind. And re-asking is the defensible side: downstream steps consume `{{var}}`, so a rewritten pick invalidates them whether the step passed or was skipped. Treating a skip differently would be the inconsistency. Withdrawn.
- **A stale pass surviving on remount.** Also reproduces — a durable pass, the variable rewritten while the step is unmounted, and it remounts showing "Data available" with no reset. But two of the writers I assumed do not exist: `GuideResponseProvider` has no `storage` listener (`GuideResponseChanged` is a same-window `CustomEvent`), so cross-tab and cross-device sync cannot reach a mounted provider, and `clearResponses` has no callers. The only real second writer is a sibling `input` block sharing a `variableName`, which your demo deliberately avoids. It is a residual on a brand-new surface rather than a regression, and putting a uid on the durable completion format at round three is disproportionate. Worth one line in the anchor; not worth code.
- **The `defaultValue` asymmetry on the advisory host** — an authored default arms the run button on mount with no pick. Your anchor already scopes the guarantee correctly ("offered" for advisory, chosen for blocking); the only gap is that the author-facing docs never say the button starts enabled. One sentence if you want it.

For what it's worth, two of this round's three real findings trace back to my own review: the explicit-`id` requirement is what made the editor path invalid, and the hydration guard exists because I asked for a durable pass to survive reload. The responses have been consistently better than the asks.

Verified at `61e677c2`: 438 suites / 7014 tests green (`tsc --noEmit` clean, no act() warnings, `--detectOpenHandles` clean). The three `learning-paths-badge-toast` failures in `tests/e2e-runner/utils/guide-runner/requirements.test.ts` are pre-existing on main — I reproduced them on a clean `main` worktree, matching your finding, and `main` is currently red for the same reason.

<sub>Re-reviewed with parallel passes over the four review-level items, the derived-verdict refactor, the block-editor and schema changes, and docs/telemetry — then an adversarial round that tried to refute each new finding. Two of my own findings were withdrawn there and one severity was corrected downward.</sub>

PR #1614, opened 2026-08-13
fix(e2e): harden runner lifecycle handling

leslie-heinzen:
## Summary

Fixes four E2E runner lifecycle gaps that produced false guide failures: skipped steps did not advance plugin state, guided readiness used a fixed five-second deadline, navigation left stale locators, and late no-op completion could hang during scroll.

## What to look at

- The step executor now synchronizes plugin state before it records a requirement-based skip.
- Guided execution uses the step timeout budget, detects terminal states while waiting for overlays, and re-resolves locators after navigation.
- Scroll and late-completion checks now use bounded waits so detached steps cannot consume the full guide timeout.

## How to verify

1. Run `reduce-metrics-cardinality` against a clean cloud stack. Confirm the first skippable step advances plugin state and the next step no longer reports “Complete previous step.”
2. Run `get-started-grafana-assistant-first-prompt`. Confirm the late no-op step exits as pre-completed instead of hanging during scroll.
3. Run a guided step whose first target takes more than five seconds to appear. Confirm the runner waits within the step’s configured budget.
4. Run a guided action that navigates or reloads the page. Confirm the runner does not continue polling the pre-navigation locator.

## Breaking changes

None. The changes are defensive runner behavior and are reversible.

## Out of scope

- Pathfinder badge-celebration overlays are handled in a separate PR.
- The plugin-side `completeEarly` guided-step lifecycle is handled in a separate PR.

### DIFF
--- a/docs/developer/E2E_TESTING.md
+++ b/docs/developer/E2E_TESTING.md
@@ -175,9 +175,12 @@ Requirements met? → Execute step
 
 **Skippable steps** (those with a Skip button) allow the test to continue when requirements cannot be met. **Mandatory steps** cause the test to abort on failure, marking remaining steps as `not_reached`.
 
-An unmet read gets a short settle window before the runner treats it as terminal. This applies whether or not a Fix button is visible. See `Requirements settle window` in the timing table below.
+An unmet read without a Fix button gets a short settle window before the runner treats it as terminal. A visible Fix button is already an actionable settled state. See `Requirements settle window` in the timing table below.
 
 The plugin's requirement check can be mid-transition right after the initial check, or right after a Fix button click. The runner polls briefly for the state to settle, instead of failing on one transient read.
+Skipping a step is a two-part handshake, not just a runner-side decision. The runner clicks the plugin's Skip control (the step's always-available Skip button, or the narrower Skip button inside the requirements-explanation banner, whichever the plugin rendered) and waits for the step to reach a terminal state: `completed`, or a successful detach. Only then does the runner record the step as `SKIPPED`. If the plugin never reaches that terminal state, for example no Skip control is found, the click fails, or it stays `requirements-unmet`, the runner records `FAILED` instead of a false `SKIPPED`. A false skip would leave the plugin gated on "Complete previous step" for every step that follows.
+
+A no-op or objective-based step can complete, or its element can detach, between discovery and this point in execution, before the runner even scrolls to it. The runner rechecks for this right before scrolling and records `PASSED`, the same outcome it records when it observes a step completing via objectives right before clicking "Do it". This keeps one DOM state, attached and `completed`, mapped to one outcome, no matter which check in the runner happens to observe it first.
 
 Overall success requires zero mandatory failures and either at least one verified pass or zero failed steps. A run where every step is skipped cleanly succeeds; a run with no verified pass and any failed skippable step fails.
 
@@ -328,15 +331,19 @@ The environment is reset **between dependency chains**, not between every guide.
 
 ## Timing and timeouts
 
-| Constant                   | Value            | Purpose                                             |
-| -------------------------- | ---------------- | --------------------------------------------------- |
-| Base step timeout          | 30s              | Maximum time for a single step                      |
-| Multistep bonus            | +5s per action   | Added for each internal action in multisteps        |
-| Guided substep bonus       | +30s per substep | Added for each substep in guided blocks             |
-| Button enable wait         | 10s              | Wait for sequential dependencies                    |
-| Fix button timeout         | 10s              | Per fix operation                                   |
-| Max fix attempts           | 3                | Retry limit before giving up                        |
-| Requirements settle window | 1s               | Poll budget before an unmet read counts as terminal |
+| Constant                   | Value            | Purpose                                                                                     |
+| -------------------------- | ---------------- | ------------------------------------------------------------------------------------------- |
+| Base step timeout          | 30s              | Maximum time for a single step                                                              |
+| Multistep bonus            | +5s per action   | Added for each internal action in multisteps                                                |
+| Guided substep bonus       | +30s per substep | Added for each substep in guided blocks                                                     |
+| Button enable wait         | 10s              | Wait for sequential dependencies                                                            |
+| Fix button timeout         | 10s              | Per fix operation                                                                           |
+| Max fix attempts           | 3                | Retry limit before giving up                                                                |
+| Requirements settle window | 1s               | Poll budget before an unmet read with no Fix button counts as terminal                      |
+| Scroll into view           | 5s               | Bounds scrolling a step into view, so a step completing or detaching there can't hang       |
+| Late completion check      | 2s               | Bounds the pre-scroll recheck for a step that completed or detached since discovery         |
+| Skip sync                  | 5s               | Bounds waiting for the plugin to reach a terminal state after the runner clicks Skip        |
+| Guided reload wait         | 15s              | Bounds waiting for `domcontentloaded` after a detected reload or navigation mid-guided-step |
 
 Examples:
 
--- a/docs/developer/E2E_TESTING_CONTRACT.md
+++ b/docs/developer/E2E_TESTING_CONTRACT.md
@@ -564,6 +564,30 @@ if (requirementsState === 'unmet') {
 
 ---
 
+### Skip controls and runner synchronization
+
+The plugin renders two Skip controls that both call the same underlying `markSkipped()` action:
+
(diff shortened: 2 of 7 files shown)


Jayclifford345:
@cursoragent review

moxious (commented):

CI is green on `b8648d7f`. `npx jest tests/e2e-runner/utils/guide-runner --coverage=false` → 7 suites, 47 tests passing; `npm run typecheck` clean.

Four higher-severity candidates were raised and then killed under verification, including a `contract_branching` verdict the policy script graded blocking/high — the six "read step state" sites turn out to read four different attributes with incompatible return shapes, so there was no single contract to branch. Not reporting those in detail; happy to share if useful.

### Medium — `executeStep` is now self-inconsistent about one DOM state

A step already at `data-test-step-state="completed"` when `executeStep` begins, but not flagged `isPreCompleted` at discovery, now returns `skipped` / `pre_completed` at `execution.ts:913-918`. At base it fell through to `checkObjectiveCompletion` and returned `passed`. **That older path is still present and unchanged** at `execution.ts:1013-1043`, so the identical DOM condition yields `skipped` or `passed` depending only on which check observes it first.

This matters because `summarizeResults` (`execution.ts:1478`) computes `success = mandatoryFailed === 0 && (passed > 0 || failed === 0)`, and `guide-runner.spec.ts` asserts on it. A run with one skippable failure whose only other steps late-complete drops `passed` to 0 and flips green → red. Separately, `pre_completed` now carries three meanings (completed at discovery / completed between discovery and scroll / element detached), which the console reporter renders identically at `console-reporter.ts:183`.

Scope check: the published JSON report's guide `outcome` derives from `mandatoryFailed` (`src/cli/e2e/e2e-reporter.ts:170-183`), so this is the in-process summary and report readability, not the guide verdict. No schema or enum change, so fully reversible.

Suggested fix: pick one classification for "already complete on arrival" and apply it at both sites — either return `passed` from the new precheck when the state genuinely reads `completed` and reserve `skipped/pre_completed` for detachment, or delete the now-dead `passed` branch. Either way, add a test for the completed-and-still-attached case; only the detached case is covered today (`execution.test.ts:533-558`).

### Low

- **`clickSkipButtonAndSync` ignores a signal the runner already computes.** It probes only `testIds.interactive.skipButton`, while `detectRequirements` already probes `testIds.interactive.requirementSkipButton` and surfaces `requirements.hasSkipButton` — which the new path never consults. The PR's own test at `execution.test.ts:480-512` sets `hasSkipButton: false` and still expects the click. Guided and multi-step render `interactive-skip-<id>` only in their `idle` blocks and `interactive-requirement-skip-<id>` in their error/cancelled blocks. The regression window is largely empty (sequentially gated steps are discovered `skippable: false` via `computeStepEligibility` + `createBlockedState`), which is why this is low rather than high — but consulting the already-computed flag is free.
- **Negative assertion is weaker than the invariant.** `execution.ts:836` asserts `not.toHaveAttribute('data-test-step-state', 'requirements-unmet')`, which passes on any other value including a transient `checking`. `markSkipped` (`step-checker.hook.ts:814-824`) sets `isCompleted: true` synchronously, so asserting the positive `completed` is available and stronger. Also note the same `timeout` is spent on both the click and the assertion, so worst case is 2× `SKIP_SYNC_TIMEOUT_MS`.
- **New bounds are tighter than the implicit defaults they replace.** `SCROLL_INTO_VIEW_TIMEOUT_MS = 5000` replaces a previously unbounded `scrollIntoViewIfNeeded()`. On a slow cloud stack a docs panel that used to settle within the per-test budget can now fail the step. Same shape for `LATE_COMPLETION_CHECK_TIMEOUT_MS = 2000` and the zero-tolerance `count() === 0` during the remount window the new `waitForLoadState` path creates. Deliberate trades — just worth watching in the first live runs.
- **Test-double discards the attribute name.** The fake `toHaveAttribute` at `execution.test.ts:39-44` takes `_attr` and discards it, reading `data-test-step-state` unconditionally, so the skip-sync tests would pass unchanged if the production code asserted on the wrong attribute. The name is separately pinned by `data-attributes.contract.test.tsx`, so this is fidelity rather than a hole.
- **New exports missing from the barrel.** `waitForGuidedCommentBoxReady`, `runGuidedSubstepLoop`, `clickSkipButtonAndSync`, `GuidedCommentBoxWaitOutcome` and the four new constants are `export`ed but absent from `index.ts:80-96`, unlike the module's other public helpers.
- **`LATE_COMPLETION_CHECK_TIMEOUT_MS = 2000` doesn't reach its own siblings** — `execution.ts:294`, `:450-451`, `:510` keep the bare literal.
- **Fourth near-identical artifact-assembly block.** `executeStep` now has four copies of the `if (artifacts && preScreenshotPath) … else if …` shape; this PR adds the fourth. A three-line `withPreScreenshot(artifacts, preScreenshotPath)` helper collapses all four.

### Positive notes

The `framenavigated` listener closes a real gap — base only re-synced on `urlBefore !== page.url()`, which misses same-URL reloads entirely. Correctly removed in a `finally`. The error-propagation discipline is deliberate and right: `count()` errors are never swallowed into a false "completed". The long comments encode Playwright auto-wait semantics a reader genuinely cannot infer, so they land on the AGENTS.md keep-list.

### Coordination with #1617 — needs a deliberate resolution

Two things, both reproduced by real merge trial rather than inferred:

1. **One content conflict**, in the `button`/`highlight` branch of `runGuidedSubstepLoop` (`execution.ts` ~730-760). This PR wraps the region in a `framenavigated` try/finally; #1617 inserts `await dismissBadgeCelebrations(page)` between `scrollIntoViewIfNeeded()` and `click()`. Resolution: keep this PR's try/finally and put the dismissal inside it, immediately before `target.click()`. Both PRs independently flip `runGuidedSubstepLoop` to `export`; git auto-merges that.
2. **A semantic gap the conflict resolution won't catch.** #1617 precedes every click site it knows about with a dismissal. This PR adds a **new** one — `skipButton.click({ timeout: SKIP_SYNC_TIMEOUT_MS })` at `execution.ts:834` — that #1617 cannot know about. Post-merge it's the only click path in the runner without celebration dismissal, and this PR makes its failure a hard `failed`. A badge toast over the Skip control intercepts the click, the 5s bound expires, and you get a spurious "Skip sync failed" — exactly the class of false failure both PRs exist to remove. **Whoever merges second should add `await dismissBadgeCelebrations(page)` before that click.**

#1615 and #1616 are complementary with no conflict. Worth noting that #1616 makes this PR's `framenavigated` detection _more_ necessary, not less: a `completeEarly` install can reload at the same URL, which is precisely the case base's URL comparison missed.

### Doc drift (non-blocking)

- `types.ts:113` — `pre_completed // Step was already completed before execution` is now stale; it covers three situations including detachment, which is not "already completed".
- `docs/developer/E2E_TESTING.md` — the outcome table should note that late-completing steps land in `skipped` rather than `passed`, and list the four new bounds (`SCROLL_INTO_VIEW_TIMEOUT_MS`, `LATE_COMPLETION_CHECK_TIMEOUT_MS`, `SKIP_SYNC_TIMEOUT_MS`, `GUIDED_RELOAD_LOAD_TIMEOUT_MS`), since they replace Playwright's configured default on paths that previously had none.
- `docs/developer/E2E_TESTING_CONTRACT.md` — this PR encodes a new runner→plugin expectation ("a skippable step in `requirements-unmet` exposes a skip affordance, or the runner reports failed") that currently lives only in runner code.

### One caveat on confidence

The center of gravity here is Playwright _runtime_ semantics — auto-wait, actionability, detachment timing — but `execution.test.ts` fakes `@playwright/test` module-wide. The new tests are real and they pass, but they model the harness, not the browser, and several candidate findings turned on "what does real Playwright actually do here", which the doubles can't answer. Green CI is weaker evidence than usual on this PR. The live runs in your "How to verify" section are the real verification.

leslie-heinzen:
## Review follow-up

Addressed the automated and Cursor review findings:

- Query and navigation errors no longer count as successful detachment.
- Guided readiness reads use absolute deadlines and operation-level bounds.
- Detached steps short-circuit before attribute reads.
- Skip synchronization now requires confirmed completion or detachment and supports both Skip control variants.
- Late attached completion and confirmed detachment now report consistently as passed.
- Late completion reads are bounded and recheck detachment after read failure.
- Runner test doubles, barrel exports, artifact assembly, timing docs, and Skip contract docs were updated.

Validation: typecheck passed, 130 focused tests passed, Prettier passed, and the diff check is clean.

Latest review follow-up commit: `bd3d87bd`.

Jayclifford345 (dismissed):
Approved on David's review and fix

Jayclifford345 (approved):

PR #1615, opened 2026-08-13
fix(e2e): stabilize requirement fix settling

leslie-heinzen:
## Summary

Prevents the E2E runner from reporting “No Fix button available” from a transient requirement state. After a Fix action, the runner now polls briefly for requirements to become met or for the Fix control to return before it declares a terminal failure.

## How to verify

1. Run `adaptive-metrics-rules-exemptions` against a clean cloud stack.
2. Trigger the navigation requirement for the “Add new rule” step.
3. Confirm the runner continues when the requirement becomes met during the settle window.
4. Confirm a genuinely unmet requirement with no Fix control still fails after the bounded wait.

## Breaking changes

None. The change adds a bounded settle period without changing the existing retry count.

### DIFF
--- a/docs/developer/E2E_TESTING.md
+++ b/docs/developer/E2E_TESTING.md
@@ -175,6 +175,10 @@ Requirements met? → Execute step
 
 **Skippable steps** (those with a Skip button) allow the test to continue when requirements cannot be met. **Mandatory steps** cause the test to abort on failure, marking remaining steps as `not_reached`.
 
+An unmet read gets a short settle window before the runner treats it as terminal. This applies whether or not a Fix button is visible. See `Requirements settle window` in the timing table below.
+
+The plugin's requirement check can be mid-transition right after the initial check, or right after a Fix button click. The runner polls briefly for the state to settle, instead of failing on one transient read.
+
 Overall success requires zero mandatory failures and either at least one verified pass or zero failed steps. A run where every step is skipped cleanly succeeds; a run with no verified pass and any failed skippable step fails.
 
 ## Artifacts and reporting
@@ -324,14 +328,15 @@ The environment is reset **between dependency chains**, not between every guide.
 
 ## Timing and timeouts
 
-| Constant             | Value            | Purpose                                      |
-| -------------------- | ---------------- | -------------------------------------------- |
-| Base step timeout    | 30s              | Maximum time for a single step               |
-| Multistep bonus      | +5s per action   | Added for each internal action in multisteps |
-| Guided substep bonus | +30s per substep | Added for each substep in guided blocks      |
-| Button enable wait   | 10s              | Wait for sequential dependencies             |
-| Fix button timeout   | 10s              | Per fix operation                            |
-| Max fix attempts     | 3                | Retry limit before giving up                 |
+| Constant                   | Value            | Purpose                                             |
+| -------------------------- | ---------------- | --------------------------------------------------- |
+| Base step timeout          | 30s              | Maximum time for a single step                      |
+| Multistep bonus            | +5s per action   | Added for each internal action in multisteps        |
+| Guided substep bonus       | +30s per substep | Added for each substep in guided blocks             |
+| Button enable wait         | 10s              | Wait for sequential dependencies                    |
+| Fix button timeout         | 10s              | Per fix operation                                   |
+| Max fix attempts           | 3                | Retry limit before giving up                        |
+| Requirements settle window | 1s               | Poll budget before an unmet read counts as terminal |
 
 Examples:
 
--- a/tests/e2e-runner/utils/guide-runner/constants.ts
+++ b/tests/e2e-runner/utils/guide-runner/constants.ts
@@ -159,6 +159,18 @@ export const REQUIREMENTS_CHECK_TIMEOUT_MS = 10000;
  */
 export const REQUIREMENTS_POLL_INTERVAL_MS = 200;
 
+/**
+ * Maximum time to poll for a settled requirements state before treating
+ * an unmet read (with or without a Fix button) as genuinely terminal,
+ * rather than a transient DOM state.
+ *
+ * The plugin's requirement check can be mid-transition right after the
+ * initial check or right after a Fix button click: a step can read as
+ * unmet even though it settles to met (or to showing a Fix button)
+ * moments later.
+ */
+export const REQUIREMENTS_SETTLE_TIMEOUT_MS = 1000;
+
 // ============================================
 // Fix Button Execution Constants (L3-4B)
 // ============================================
--- a/tests/e2e-runner/utils/guide-runner/requirements.test.ts
+++ b/tests/e2e-runner/utils/guide-runner/requirements.test.ts
@@ -1,0 +1,178 @@
+/**
+ * Unit tests for guide-runner requirements handling.
+ *
+ * Covers the settle-window behavior in attemptToFixRequirements: when a fix
+ * attempt finds no Fix button, the runner polls briefly for the DOM to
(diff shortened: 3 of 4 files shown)


Jayclifford345:
@cursoragent review

moxious (commented):

The change does what it says, verified empirically rather than by reading: reverting only the second commit's hunk makes 2 of the 3 new tests fail, so both commits are load-bearing and pinned. `test:ci` (425 suites / 6737 tests), coverage, governance, typecheck, lint, prettier and build all pass on head; CI is green.

Eight candidate findings died under adversarial verification, including a `contract_branching` verdict the policy script graded **blocking/high** — its load-bearing claim was that on attempts 2/3 the new 1s poll is the only wait guarding the terminal decision, which is false, since `clickFixButton` calls `waitForRequirementsCheckComplete` on every attempt whose click lands. Not reporting the rest in detail; happy to share if useful.

### What I checked and could not break

- **It cannot mask a genuinely-failing requirement.** `handleRequirementsWithFix` re-detects fresh at `:718` and `execution.ts:748` decides pass/fail from _that_ read, not from `FixResult`. A false `success: true` corrupts the fix record and the log line, never the step verdict.
- **The terminal message stays truthful.** `'No Fix button available'` is still emitted only after a full window with no Fix button and not met. Measured with a fake-clock probe: 5 polls × 200ms, elapsed exactly 1000ms, then the terminal break — and it `break`s, so it does not recur per attempt.
- **No interaction with the retry count.** `MAX_FIX_ATTEMPTS` is untouched and the settle consumes no attempt.
- **Measured latency**, fake-clock probes against the real code: 0s when a fix failed and the Fix button remains (the loop exits with zero polls), 1.0s once on the no-Fix-button terminal path. Theoretical worst case ~7.2s/step needs a Fix button appearing on the last poll of every attempt.

### Medium — the bounded wait is bounded _between_ polls, not _within_ one

`Date.now() - startTime < settleTimeoutMs` is only evaluated between iterations (`requirements.ts:481-484`). Each iteration awaits `detectRequirements`, which calls `locator.isEnabled()` (`:99`, `:102`) and `locator.textContent()` (`:122`, `:212`) with **no timeout**. Neither `playwright.config.ts` nor `tests/e2e-runner/playwright.config.ts` sets `actionTimeout`, and nothing calls `setDefaultTimeout`, so Playwright's default of 0 (unbounded) applies. This was reproduced against real Chrome: `isEnabled()` on an element detached after a passing `count()` was still pending at 4000ms.

So a single poll can hang until the guide-level `test.setTimeout`, killing the whole run with an opaque global timeout instead of one attributed step. That matters here specifically because the PR's premise is that the DOM is mid-transition — which is exactly when a detach between `count()` and `isEnabled()` is most likely — and the settle loop adds up to ~5 such reads per call site per attempt.

Honest scoping: every unguarded call is **pre-existing** inside `detectRequirements`; this widens exposure rather than creating it. The repo already knows the hazard — `execution.ts:396` carries the comment _"a bare getAttribute on a detached locator blocks until the global test timeout"_ and defends with `{ timeout: 2000 }` there, but not here.

Fix is one line: `use: { actionTimeout: 2000 }` in `tests/e2e-runner/playwright.config.ts`, or `{ timeout: REQUIREMENTS_POLL_INTERVAL_MS }` on the four calls in `detectRequirements`. Either makes the 1s bound real. Reasonable as a follow-up issue rather than a merge gate, since it predates this PR.

### Nits

- **The docstring overclaims what the loop does.** `requirements.ts:379-387` says a step "can read as unmet (with or without a Fix button) even though it settles to met" — but the `!latest.hasFixButton` guard makes the loop return immediately when a Fix button is present. That short-circuit is intentional and correct (a visible Fix button is a settled, actionable state the retry loop will click); the code is right and the comment is wrong. Same at `:613-615`. `:647` restates the line below it.
- **`REQUIREMENTS_SETTLE_TIMEOUT_MS` is module-local** at `:388` while every sibling timing constant lives in `constants.ts:150-188`, whose header points at `docs/developer/E2E_TESTING.md#timing-and-timeouts`. One-line move.
- **Real coverage gaps:** `:593` (Fix button appears during the settle, then clicked with the settled `fixType`) and `:635-648` (`finalStatus = settledPostFixRequirements.status`) are never executed. The mock hardcodes the spinner locator to `count: 0` (`requirements.test.ts:102`), so `status: 'checking'` is unreachable in the entire suite, and no test uses `maxAttempts > 1` with a failing attempt. File coverage is 58.67% stmts / 45.77% branch.
- **Test-harness fragility.** `createSequencedPage(states, clock = createFakeClock())` defaults the clock and the `Date.now` spy is a per-test convention (`requirements.test.ts:68-70`). Deleting only the spy from test 2 keeps everything **green** while the suite goes 0.42s → 3.9s (a real 1s sleep). Making `clock` required and installing the spy inside the factory would close that. Also `:204`'s `toBeGreaterThan(1)` proves "polled twice", not "honored the 1000ms window".
- **`finalStatus` is inconsistent across terminal paths** — forced to `'met'` at `:566`, left at the `'unmet'` initializer at `:576-589`, assigned verbatim at `:648`. Cosmetic: `finalStatus` has no consumer outside `types.ts:219`.

### Doc drift (non-blocking)

`docs/developer/E2E_TESTING.md:173` — the decision tree's `└─ No fix button → FAILED` now means "→ bounded settle → FAILED", and the `Fix button available → Attempt fix` branch can conclude PASSED with zero Fix clicks. The timing table at `:325` omits the new 1000ms window, though it already omits `POST_FIX_SETTLE_DELAY_MS` and friends, so that's incremental drift rather than novel — worth adding if the constant moves to `constants.ts`.

### Coordination with the sibling PRs

Merge safety verified with `git merge-tree` rather than assumed: this PR merges clean against #1614, #1617, and current `origin/main`. #1616 is `src/`-only, no overlap.

**#1617's insertion point survives.** In the merged tree its `dismissBadgeCelebrations(page)` call lands exactly where intended in `clickFixButton` — this PR only inserts above (the constant) and below (`waitForRequirementsSettle`), never touching that function's body.

**Composed worst case with all four merged: ~116s/step vs ~81s on main today.** Of that ~35s increase, #1617 contributes ~28s and this PR ~7s — about 6% of the composed total, and 0–1.2s measured. The budgets stack but harmlessly, and the reason is worth knowing: `calculateStepTimeout` is **not** a step wall clock. It's passed only to `waitForCompletionWithObjectivePolling`, so the entire requirements phase is unbudgeted at the step level; the only enforcement is the pooled guide-level `test.setTimeout`. So this PR's window is purely additive to #1614's deadline model rather than nested inside it.

One posture wart worth a thought: this PR is deliberately making transient unmet reads _non-terminal_, while #1617's insert at `requirements.ts:411` sits outside `clickFixButton`'s `try` and is terminal — a dismissal failure escapes to `executeStep`'s catch and fails the step with a badge message instead of a requirements message. The two changes point in opposite directions on the same path. Moving that call inside the `try` resolves it for free.

leslie-heinzen:
## Review follow-up

Addressed the automated review findings:

- The settle window now runs after the final Fix click, without consuming another attempt.
- Requirement reads inside the settle loop use operation-level bounds and remaining deadline budget.
- A returning Fix control preserves its `fixType`.
- Final status comes from the settled post-Fix state.
- Added multi-attempt, returning-Fix, final-attempt, and bounded-read coverage.
- The fake clock is deterministic and advances only through runner waits.
- Shared timing constants and E2E timing documentation were updated.

Validation: typecheck passed, 92 focused tests passed, Prettier passed, and the diff check is clean.

Latest review follow-up commit: `5a5a9a1a`.

Jayclifford345 (approved):

PR #1616, opened 2026-08-13
fix(interactive): preserve guided completeEarly cleanup

leslie-heinzen:
## Summary

Fixes an orphaned guided overlay caused by `completeEarly`. The component previously persisted completion before the guided handler installed its listeners, which could auto-collapse and unmount the section before the overlay existed. The stale asynchronous callback then created an overlay with no owner until the step timeout.

The guided loop now starts first. For the final button or highlight action, completion persists from the click listener during document capture, before the application handles the click and changes routes.

## What to look at

- Completion persistence is idempotent and occurs only after the guided action listener is active.
- Cancellation, timeout, and error outcomes do not persist false completion.
- Guided handler cleanup still owns every listener, timer, highlight, and comment box.
- Schema text, author guidance, and the interactive-engine contract anchor describe the same click and non-click lifecycle.

## How to verify

1. Build this branch’s plugin frontend and serve it on a disposable Grafana Cloud stack with Graft.
2. Load `alerting-query-language-install-datasources` with its original `completeEarly: true` guided block.
3. Start the guided install step and click the highlighted QuickPizza install control.
4. Confirm the click completes, the overlay disappears, and Grafana navigates to the QuickPizza dashboards folder.
5. Cancel or time out a guided `completeEarly` action and confirm the step does not persist completion.

## Breaking changes

None. This preserves the intended `completeEarly` contract while correcting its ordering.

## Out of scope

- E2E runner handling for badge celebrations and other transient overlays is handled separately.


moxious (commented):
## Automated review pass — verdict: request changes (narrow)

Reviewed with the repo's `/review` orchestrator (concern routing, contract-evolution gate, parallel concern reviewers, adversarial verification). Flagging up front that this is an agent-assisted pass, not a line-by-line human read — treat the findings as leads worth checking rather than settled conclusions. Posting as a comment rather than a formal block; the "request changes" is a recommendation, not a gate.

**The core change is right.** Starting the guided loop before persisting, and persisting from the capture-phase click listener, genuinely fixes the orphaned-overlay leak — auto-collapse does unmount the step (`interactive-section.tsx:1359` renders children behind `!isCollapsed`), and the `document`-level overlay listener did outlive its owner. Every exit path still routes through `cleanupListeners(true)` / `cancel()`, so this does not trade one leak for another. `tsc --noEmit` clean, prettier clean, `src/validation` 822/822, both changed suites pass.

What's requested is small. The reason it's phrased as request-changes rather than nits: **this PR records a permanent contract anchor, and two of its four sentences assert guarantees the code does not enforce.** Anchors are what future PRs get gated against, so an overstated one ossifies — and an inaccurate anchor is worse than none.

The §3b contract-evolution scan itself came back clean (`coherent_extension` → disposition `none`). The pre-change state was genuinely `contract_missing` — no code modeled `completeEarly`'s lifecycle, just three parallel copies and repeated patches — and establishing the owner plus recording the anchor in the same PR is exactly what §5 requires. The anchor is also routable (checked: `guide-schema-and-contracts.related` includes `interactive-engine`). It's the text that needs work, not the placement.

### M1 — "cancellation, timeout, and error do not persist completion" is reachably false

The completion listener carrying `onActionCompleted` is installed at `guided-handler.ts:164`, but `await this.highlightTarget(...)` at `:177` includes a documented ≥300ms DOM-settling delay. A click landing in that window persists completion synchronously from capture. If `highlightTarget` — or `cleanupListeners(true)` → `clearAllHighlights()` at `:198` — then throws, the `catch` at `:205` returns `'error'`. The component shows Retry while completion is already written.

That's not a hypothetical interleaving: "the target's subtree was just replaced by the click" is precisely the scenario `completeEarly` exists for. The same shape applies to `'cancelled'` — `createCancelListener` and the Escape handler resolve without aborting, so a click in the same tick can persist and still lose the race.

This is asserted on four surfaces now: `docs/design/CONCERNS.md:120`, `json-guide.schema.ts:386-392`, `json-guide.types.ts:396-402`, and both author docs.

Two ways out, and the first is nicer because it makes the claim _true_ rather than narrowing it:

- Move the `try/catch` around `highlightTarget` so a post-persist throw resolves `'completed'`.
- Or soften all four surfaces to "an error _before the final action signal_ does not persist completion."

### M2 — optimistic latch plus unguarded callback can permanently suppress persistence

```ts
const completeStep = () => {
  if (completionPersisted) { return; }
  completionPersisted = true;      // latched BEFORE any work
  persistCompletion();
  ...
```

`interactive-guided.tsx:348-353`. And `guided-handler.ts:897-903` calls `onActionCompleted?.()` with no `try/catch`, from inside a `document` capture listener.

If `persistCompletion` / `onStepComplete` / `onComplete` throws: the latch is already set, so the later `completeStep()` calls at `:383` and `:402` no-op; the race has already resolved `'completed'`; and the throw escapes into DOM dispatch on a different stack, where `executeStep`'s own `try/catch` cannot see it. Net result — the step renders completed, the `critical: true` Faro span stamps `ok`, nothing is persisted, and no error surfaces anywhere.

Two one-line fixes: set `completionPersisted = true` only after the three calls succeed, and wrap the `onActionCompleted?.()` invocation in `try/catch` inside `guided-handler.ts` so a consumer throw can't corrupt engine state.

(Verification note: this finding got fewer independent skeptics than the others due to a concurrency cap, so it's worth a second look. The code claims were re-verified directly against head.)

### M3 — the load-bearing "final action only" gate is untested, and a shipped guide is the uncovered shape

Every `completeEarly` test passes a single-element `internalActions`, so `i === internalActions.length - 1` at `:371-372` and `:382-384` is trivially true at `i === 0`. Mutation-proven: relaxing **both** guards to bare `completeEarly` leaves the full suite green.

Meanwhile `src/bundled-interactives/json-guide-demo/content.json:160-186` ships a `guided` block with `completeEarly: true` and **two** `highlight` steps — precisely the uncovered shape. With the gate removed, the first click would persist the whole block.

One test with a two-action `internalActions`, asserting `onStepComplete` fires only after the second, closes this.

### M4 — timeout/error non-persistence is untested, and one test's title claims what its body can't check

Only cancellation has `expect(onStepComplete).not.toHaveBeenCalled()` (`interactive-guided.test.tsx:377`). Adding a `completeStep()` call to the timeout or error branch breaks nothing.

The PR renamed a test to _"reruns failed actions without persisting the failed run"_ (`:401`) without touching its body — it asserts only `data-test-step-state`, and `deriveGuidedUiState` ranks `hasError` above `isCompleted`, so that attribute **cannot** distinguish "persisted then errored" from "never persisted". Given M1 shows that interleaving is reachable, the gap isn't academic.

Two `expect(onStepComplete).not.toHaveBeenCalled()` lines, plus one for M1's persist-then-error path.

### Low

- **The anchor overstates timer ownership.** `rectUpdateInterval` in `waitForClick` (`guided-handler.ts:906-910`) is a bare `setInterval` never pushed to `pendingTimeouts`, so `cleanupListeners` cannot clear it. Timeout, skip, comment-box cancel and Escape all resolve without aborting `currentAbortController`, and the next step overwrites it — so the interval polls every 100ms until the element detaches. Pre-existing, but the anchor now claims "`GuidedHandler` owns listeners, **timers**, overlays… and cleanup." Either register the interval or drop "timers" from the clause.
- **`E2E_TESTING_CONTRACT.md:402`** — the precedence rule stays correct, but its stated rationale ("keeps the active substep observable when `completeEarly` persists completion before the remaining actions settle") is now true only for **multi-step**, which still uses the old ordering at `interactive-multi-step.tsx:342-353`. It's a contract doc the runner PRs read, so worth fixing here.
- **The block-editor authoring GUI still describes the old semantics.** `GuidedBlockForm.tsx:201-202` reads _"Complete early (mark complete when user performs action early)"_ / _"Marks the block as done if user completes the action before being prompted."_ Pre-existing and false at base too — but this PR rewrote the same field's text on four other surfaces, so the in-product one is now the odd one out.
- **`completeEarly` is now a ~500ms no-op for non-click final actions.** For `hover`/`formfill`/`noop` finals, the only difference from not setting the flag is skipping ahead of the 500ms visual-feedback delay; for a `noop` final it no longer completes early in any user-visible sense. The docs describe the mechanism but not that the flag has gone near-inert for those shapes.
- **The docs say "the click signal", but two paths persist without one.** `guided-handler.ts:940-952`: the 16px-proximity fallback synthesizes `element.click()`, and when `element.isConnected` is false it skips the synthetic click entirely and still calls `complete()`. Pre-existing result semantics; the new prose is just more absolute than the code.

### Dropped under verification

Recording these so they don't get re-raised. A debounced-user-storage-write-lost-on-navigation finding was refuted — `syncFromGrafanaStorage`'s `keysToSync` doesn't include `INTERACTIVE_STEPS_PREFIX`, which is write-only to the backend and always read from localStorage, so a dropped queued write is unobservable. Also refuted: "persistence races unmount" (`cleanup('completed')` resolves before `onActionCompleted`, and cleanup is idempotent), and "capture-phase persistence can fire on a non-target click" (`isTargetOrChild` / `isWithinBounds` gate it, same as the pre-existing non-`completeEarly` path). Security reviewed clean.

### Tech debt (non-blocking)

**`completeEarly` is now triplicated with fresh divergence.** `interactive-step.tsx:470-481` and `interactive-multi-step.tsx:342-353` still carry the pre-emptive `persistCompletion(); … setTimeout(50)` pattern this PR deleted from guided — including the same false-completion-on-failed-navigate hazard. The bug class is fixed in one of three copies. Worth a tracking issue rather than widening this PR.

Smaller: `completeEarly && i === last` is computed twice (`:371`, `:382`) — hoisting to one `isFinalAction` const halves the untested-gate surface from M3. And `executeGuidedStep` is now five positional params with a defaulted one before an optional one; next addition should probably be an options object.

The PR correctly removed two QC8-shaped comments. `// CRITICAL FIX: Attach listener BEFORE highlighting…` at `guided-handler.ts:162` is adjacent to a changed line and is a shouting/historical shape — trim-on-touch candidate.

### Coordination with #1614 / #1615 / #1617

Consistent with them, not papered over by them — this removes the root cause while they harden against residual legitimate cases (SPA reload, correctly-timed auto-collapse, badge toasts). Nothing in them becomes dead weight: post-merge the step still unmounts, just at the final action rather than at start, and this PR's own `AutoCollapseHarness` test proves it. #1614 explicitly names the scenario — one of its tests is titled _"treats a successfully-confirmed detached step as completion (e.g. after a completeEarly navigation)"_.

No file collisions with any of the three.

Two second-order shifts worth stating out loud, both intended: sequential sections will now cascade when a `completeEarly` block times out (no early unblock), and abandoned blocks no longer leave a persisted completion that masked them as `pre_completed` on the next run. **Expect previously-invisible E2E failures to surface** — that's this PR working, not a runner regression.

One open question I couldn't settle: whether merging the runner hardening (#1614/#1617) before this PR risks converting a zero-substep guided run into a quiet `passed`. One analysis pass argued it does; another rebutted it via `deriveGuidedUiState` checking `isExecuting` before `isCompleted`, which means a `completeEarly` step exposes `executing` throughout its run under both versions. The rebuttal looks stronger, but it's worth your judgment since you know the runner's failure modes better than a static read does.

leslie-heinzen:
## Review follow-up

Addressed the lifecycle and contract review findings:

- Added one synchronous arbiter for completion, skip, cancel, timeout, abort, and callback errors.
- Late clicks cannot persist completion after cancel or skip wins.
- Persistence callback failures become observable engine errors.
- Successful completion remains complete through later setup or cleanup faults.
- The idempotence latch is set only after persistence callbacks succeed.
- Connectivity intervals are tracked and cleaned with other guided resources.
- Added two-action final gating, timeout/error non-persistence, cancellation race, callback failure, cleanup failure, and timer cleanup tests.
- Updated schema text, author guidance, block-editor text, E2E contract wording, and the interactive-engine contract anchor.

Validation: typecheck passed, 42 focused tests passed, Prettier passed, and the diff check is clean.

Latest review follow-up commit: `63bbb0e4`.

Spidy88 (approved):
**Review: Approve**

Confirmed this fixes a real race: `completeEarly` used to persist completion (→ `onComplete` → section reacting) before `GuidedHandler` had attached its listener or drawn the highlight overlay, so a section collapse could leave the overlay orphaned in the DOM until `stepTimeout`. The fix reorders things so persistence for a final `button`/`highlight` action fires from inside the click listener itself (capture phase, before the app's own handler), and non-click final actions persist after their result resolves — same as before. The new `GuidedStepArbiter` makes "first terminal outcome wins" explicit instead of a bare `Promise.race`, and it's well covered (click-before-DOM-mutation ordering, callback-throws, cancel/skip-beats-later-click all have direct tests).

Scope check on whether this class of bug lurks elsewhere:
- **`interactive-multi-step.tsx`** has textually identical `completeEarly` code, but it's dead — the field isn't in the multistep Zod schema, the CLI field allowlist, or the type, and `json-parser.ts`'s `convertMultistepBlock` never reads or forwards it. No authored content can trigger it today. Worth a one-line note for whoever eventually adds `completeEarly` authoring support to multistep, so this fix's ordering gets reapplied then.
- **`interactive-step.tsx`** *is* live, but its `do`-mode handlers (button/focus/hover/formfill) never create an overlay in do mode — no equivalent hazard exists there. Tracing it did surface an unrelated, real gap (a `completeEarly` step whose action or `verify` subsequently fails stays recorded as completed, with no reconciliation) — separate issue, doesn't block this PR.

**Suggested follow-up (non-blocking):** the existing unmount-cleanup effect (`interactive-guided.tsx:261-265`, `guidedHandler.cancel()`) has the same race in reverse — it aborts whatever `AbortController` currently exists, but each step mints a fresh one, so a cancel that fires before a step's listener is set up is a no-op. Worth considering a single root `AbortController` created once alongside `guidedHandler` and aborted at most once, on unmount (a good fit, since unmount happens exactly once) — with each step's own per-step controller (unchanged, still fresh per step for cancel/skip/timeout) composed against the root's signal, so either one aborting cancels the step. The codebase already has a ready-made helper for this composition, `composeAbortSignals()` in `src/components/PrTester/github-api.ts:448` (hand-rolled to avoid depending on `AbortSignal.any` runtime support) — reusing it here would mean the follow-up is mostly wiring, not new logic. That would make "the component unmounted" a durable fact any later signal check inherits, rather than something that only wins if it happens to fire in the right order — closing this whole race class structurally instead of case-by-case. Raising as a future idea, not a request to change this PR.

PR #1617, opened 2026-08-13
fix(e2e): dismiss badge celebrations before actions

leslie-heinzen:
## Summary

Prevents Pathfinder badge celebrations from blocking E2E runner interactions. The runner now waits through short queue-render gaps, dismisses up to three known badge toasts, and fails with a bounded diagnostic if the overlay remains.

## What to look at

- The helper targets only Pathfinder’s stable badge-toast test IDs. It does not close arbitrary Grafana modals.
- Dismissal runs before target resolution, runner-controlled clicks, hidden-target reveal, guided hover and form-fill retries, guided Skip, and requirement Fix.
- Queue handling is bounded to the plugin’s current three-celebration maximum.

## How to verify

1. Use a fresh user or stack that earns a badge when a guide completes.
2. Complete a guide that queues one or more badge celebrations.
3. Continue into the next runner-controlled action.
4. Confirm each badge toast is dismissed before the underlying action runs.
5. Force the toast to remain visible and confirm the runner reports a bounded badge-dismissal error.

## Breaking changes

None. This adds narrow E2E handling for an existing Pathfinder overlay.

### DIFF
--- a/docs/developer/E2E_TESTING_CONTRACT.md
+++ b/docs/developer/E2E_TESTING_CONTRACT.md
@@ -68,6 +68,23 @@ A My learning layout or selector refactor must preserve these values or update t
 
 ---
 
+## Badge celebration runner contract
+
+After a guide completes, Pathfinder can show a full-screen badge celebration before the runner starts the next action.
+
+The runner and plugin share these stable test IDs:
+
+- **`learning-paths-badge-toast`** (`testIds.learningPaths.badgeToast`): identifies the visible badge dialog.
+- **`learning-paths-badge-toast-dismiss`** (`testIds.learningPaths.badgeToastDismiss`): identifies the dismiss action inside the current dialog.
+
+The runner uses only these selectors. It does not close generic Grafana modals.
+
+If a refactor changes these values, update `BadgeUnlockedToast.tsx`, the guide runner, the contract test, and this document in the same change.
+
+The source-level tripwire lives in `src/components/LearningPaths/BadgeUnlockedToast.contract.test.ts`.
+
+---
+
 ## Design Principles
 
 ### 1. Semantic Over Syntactic
@@ -420,6 +437,7 @@ Contract tests enforce the stability of E2E attributes at build time, preventing
 - `src/components/interactive-tutorial/data-attributes.contract.test.tsx` - React component attributes
 - `src/interactive-engine/comment-box.contract.test.ts` - DOM-created element attributes
 - `src/components/docs-panel/docs-panel.contract.test.tsx` - Docs panel test IDs (constant values, source reference mapping, auto-derived exhaustiveness, window globals, scroll-restoration)
+- `src/components/LearningPaths/BadgeUnlockedToast.contract.test.ts` - Badge celebration test IDs and source references
 
 ### Pattern: Dual Assertion
 
--- a/src/components/LearningPaths/BadgeUnlockedToast.contract.test.ts
+++ b/src/components/LearningPaths/BadgeUnlockedToast.contract.test.ts
@@ -1,0 +1,18 @@
+import { readFileSync } from 'fs';
+import { join } from 'path';
+
+import { testIds } from '../../constants/testIds';
+
+const source = readFileSync(join(__dirname, 'BadgeUnlockedToast.tsx'), 'utf-8');
+
+describe('Badge unlocked toast E2E contract', () => {
+  it('keeps the guide-runner test IDs stable', () => {
+    expect(testIds.learningPaths.badgeToast).toBe('learning-paths-badge-toast');
+    expect(testIds.learningPaths.badgeToastDismiss).toBe('learning-paths-badge-toast-dismiss');
+  });
+
+  it('uses both guide-runner test IDs in the component source', () => {
+    expect(source).toContain('data-testid={testIds.learningPaths.badgeToast}');
+    expect(source).toContain('data-testid={testIds.learningPaths.badgeToastDismiss}');
+  });
+});
--- a/tests/e2e-runner/utils/guide-runner/badge-celebrations.test.ts
+++ b/tests/e2e-runner/utils/guide-runner/badge-celebrations.test.ts
@@ -18,6 +18,11 @@ interface BadgeHarnessOptions {
   remainsVisible?: boolean;
   initialDelayMs?: number;
   interToastDelayMs?: number;
+  textReadFailure?: {
+    call: number;
+    mode: 'disappear' | 'visible';
+    error: Error;
+  };
 }
 
 function createBadgeHarness(titles: string[], options: BadgeHarnessOptions = {}) {
@@ -26,6 +31,7 @@ function createBadgeHarness(titles: string[], options: BadgeHarnessOptions = {})
   let nextIndex = currentIndex === 0 ? 1 : 0;
   let nextVisibleAt =
     currentIndex === -1 && titles.length > 0 ? (options.initialDelayMs ?? 0) : Number.POSITIVE_INFINITY;
+  let textReadCount = 0;
   const events: string[] = [];
   const toast = {} as Locator;
   const dismissButton = {} as Locator;
@@ -44,9 +50,19 @@ function createBadgeHarness(titles: string[], options: BadgeHarnessOptions = {})
   toast.first = jest.fn(() => toast);
   toast.count = jest.fn(async () => (currentIndex >= 0 ? 1 : 0));
   toast.isVisible = jest.fn(async () => currentIndex >= 0);
-  toast.textContent = jest.fn(async () => {
+  toast.textContent = jest.fn(async (textOptions?: { timeout?: number }) => {
     if (currentIndex < 0) {
       return null;
+    }
+    textReadCount++;
+    if (options.textReadFailure?.call === textReadCount) {
+      elapsedMs += textOptions?.timeout ?? 0;
+      if (options.textReadFailure.mode === 'disappear') {
+        currentIndex = -1;
+        nextVisibleAt =
+          nextIndex < titles.length ? elapsedMs + (options.interToastDelayMs ?? 0) : Number.POSITIVE_INFINITY;
+      }
+      throw options.textReadFailure.error;
     }
     const queueCount = titles.length - nextIndex;
     const queueText = queueCount > 0 ? ` (+${queueCount} more)` : '';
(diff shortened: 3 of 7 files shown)


moxious (commented):

Every medium-or-higher finding raised against this PR was killed under adversarial verification, most 3/3 and several with empirical repro. What remains is genuinely small. `npx jest tests/e2e-runner/utils/guide-runner/` → 5 suites, 60 tests green; `src/validation/architecture.test.ts` passes; prettier clean on all five files.

Three design choices that read as suspicious on first inspection turn out to be deliberate and defensible, so recording the reasoning here:

**The no-badge path costs ~100ms every time, and that's on purpose.** `waitForVisibleToast` (`badge-celebrations.ts:51-60`) always burns 4 × `waitForTimeout(25)` plus ~5 protocol round trips when no badge exists — there is no zero-cost path. Commit `1c3b9155` _removed_ the original zero-cost `isVisible()` check and replaced it with this idle poll, because the toast renders asynchronously (React state → effect → `setTimeout`). So it's a bought robustness tradeoff. Accounting: ~10 invocation sites, a guided substep pays it 2–3× (~200–300ms), roughly +14s across a full bundled-guide sweep. That matters less than it looks — `calculateGuideTimeout` budgets 30s per guided substep (<1% consumption), the runner already sleeps ~1s/substep by design, and root `playwright.config.ts:20` sets `testIgnore: '**/e2e-runner/**'`, so no gating CI job pays it. Only the interactive `pathfinder-cli e2e` tool does.

**The six-plus call sites are justified.** A 5–15s overlay spans many subsequent actions of every kind, so a single choke point wouldn't cover it.

**The test IDs are real and reasonably stable.** `testIds.learningPaths.badgeToast` / `badgeToastDismiss` exist at `src/constants/testIds.ts:302-303` and are rendered by `BadgeUnlockedToast.tsx:82,131`. They're typed constants imported by both sides, so a key rename is a compile error and a value change moves both sides together. The only silent break is removing the attribute from the JSX, which degrades to a no-op early return rather than a failure.

### Nits

- **`.first()` asymmetry on the dismiss locator.** `badge-celebrations.ts:92` uses `.first()` for the toast; `:108` does not — `page.getByTestId(testIds.learningPaths.badgeToastDismiss)` then `.click()`. Playwright strict mode throws if that resolves to more than one element, and two `BadgeUnlockedToast` instances are mountable (the sidebar panel at `docs-panel.tsx:1521` and the full-page route via `CombinedLearningJourneyPanel`), each running its own `useBadgeCelebrationQueue`. Not reachable in today's runner flow, but a guide that navigates to `/a/grafana-pathfinder-app` would make it so, and the failure would be an opaque strict-mode error. One character of risk removal.
- **`MAX_BADGE_CELEBRATIONS = 3` is correct, but the PR body's rationale isn't.** `useBadgeCelebrationQueue.ts:47` slices `newBadges.slice(0, 3)` **per event**, and line 48 _appends_ to the queue — so it isn't a queue maximum. The value is still right, for a different reason: `badge-coordinator.ts:50` gates on `!progress.completedGuides.includes(guideId)` (one event per guide) and `guide-runner.spec.ts` runs exactly one guide per browser context. Importing the constant isn't an option — `tests` is a `NODE_CONTEXT_ROOTS` entry (`src/validation/import-graph.ts:636`), so pulling a value out of a React hook would break the environment-reachability ratchet. The local constant is the only legal shape; it just needs a one-line comment naming the two actual invariants, since nothing would alert someone who changes either. This is squarely the "hidden invariant the type system can't express" case on the AGENTS.md keep-list.
- **Dismissal sits outside `clickFixButton`'s `try`.** `requirements.ts:411` lands between the `if (buttonCount === 0) return false;` guard and the `try {` at `:416` whose `catch { return false }` is the function's failure contract. A throw therefore escapes `attemptToFixRequirements`' retry loop and lands in `executeStep`'s outer catch, failing the step with a badge message instead of a requirements message. Verification pushed back on the _harm_ here — the function was already non-total at base, since `getByTestId(...).count()` also sits outside the try, and a genuinely wedged modal would fail the retries identically — so this is defensible as-is. But moving the line one row down is free and makes the shape match the docstring. Worth doing especially alongside #1615, which is deliberately making transient unmet reads non-terminal on that same path.
- **`errorReason()` defined then re-inlined.** `badge-celebrations.ts:20-22` defines it, uses it at `:33` and `:41`, then `:113` re-inlines the identical `error instanceof Error ? error.message : String(error)`.
- **`(+N more)` is the runner's only text-driven control-flow decision.** `hasQueuedCelebration` (`:14-16`) regexes `/\(\+\d+ more\)/` against product copy at `BadgeUnlockedToast.tsx:102`, and the unit test's fake reconstructs the same string (`badge-celebrations.test.ts:67-69`), so it can't detect drift. Other `textContent()` reads in the runner are diagnostic-only; this one gates behavior. Reported for awareness rather than action — verification showed drift is fail-safe (`expectNextToast=false` makes `waitForToastTransition` return `true` on invisibility, the correct answer, never throw), and one skeptic reproduced it by running the real helper with a deliberately drifted regex: gaps of 0/25/75/100ms still dismissed both toasts cleanly.
- **`E2E_TESTING_CONTRACT.md` doesn't record the new dependency.** The runner now leans on 13 plugin test IDs and exactly one (`docs-panel-container`) appears in a normative contract section; `badgeToast`/`badgeToastDismiss` appear in no contract test anywhere. Pre-existing gap this PR widens rather than creates. A two-line bullet under the existing "Docs panel bootstrap contract" pattern would follow the convention that's already there, including its source-level tripwires.

Also worth knowing: the four timing constants here (`MAX_BADGE_CELEBRATIONS`, `BADGE_TRANSITION_TIMEOUT_MS`, `BADGE_IDLE_TIMEOUT_MS`, `BADGE_TRANSITION_POLL_MS`) are module-local, while every other runner constant lives in `constants.ts` — which #1614 is concurrently extending with 27 new lines. Two concurrent PRs adding runner timing constants to two different places is the start of a split; moving these four during the rebase is nearly free.

### Dropped under verification

Recording the notable one: the contract-evolution scan produced a `contract_branching` verdict that the policy script graded **blocking/high**, on the theory that `dismissGenericHelpMenu` (`bootstrap.ts:101`) is a competing owner of "clear interfering UI". Refuted 3/3 — that function is module-private and unexported, a local one-off inside `openDocsPanelAttempt`'s retry rather than a capability owner, and you can't branch a contract that was never established. The mechanism divergence is also _correct_: `page.keyboard.press('Escape')` is a global `window` listener that would close unrelated Grafana drawers mid-run.

Also refuted: the latency finding above (3/3), a `dismissBadgeCelebrations`-throws-inside-a-best-effort-ladder finding (3/3 — the base `revealGuidedTarget` already ends with a throwing `target.waitFor()`), and an auto-dismiss race (2/2 — the window is one CDP round trip against a 5000ms lifetime, and `badge-celebrations.test.ts:183` and `:203` explicitly cover disappear-during-read and assert _resolve_, not throw).

### The product bug underneath this

This is the one thing I'd not let close with this PR. The "toast" is a genuine full-viewport blocking modal: `position: fixed`, inset 0, `rgba(0,0,0,0.5)` backdrop, `backdropFilter: blur(4px)`, `zIndex: theme.zIndex.modal`, and **no `pointer-events: none`** (`learning-paths.styles.ts:557-570`), with a click-to-dismiss overlay handler.

So a real user mid-guide has their next click **swallowed** for up to 5s per badge (`AUTO_DISMISS_DURATION = 5000`), up to 3 queued back-to-back — and the 500ms inter-toast arm at `useBadgeCelebrationQueue.ts:33` is dead code (the enclosing guard admits only `null`), so successors render at 0ms, i.e. ~15s of continuous overlay. It can fire mid-run via `markMilestoneDone` → `markGuideCompleted(milestoneSlug)`.

The runner fix here is legitimate and necessary — it's faithfully reproducing real user pain. But the product fix is separate, and the plugin already owns `src/interactive-engine/global-interaction-blocker.ts` with section-blocking state that celebrations could be gated on. Worth a product issue.

### One follow-up worth considering

`page.addLocatorHandler(locator, handler)` (Playwright ≥1.42) is purpose-built for exactly this shape: one registration at page setup, fires **only** when an action is actually blocked (a true zero-cost fast path), and closes a TOCTOU window that pre-emptive dismissal structurally cannot — the toast can render _between_ the dismissal check and the click. #1616 actually tightens that window, since completion there now fires inside the runner's own click. Not a blocker for this PR, but it would collapse ten insertion points into one.

### Coordination with #1614

Verified by real merge trials rather than inferred. **#1615 merges clean** with this PR. **#1614 conflicts**, but only once: one region in `execution.ts`, the guided `button`/`highlight` click, where #1614 wraps the click in a `framenavigated` try/finally and this PR inserts the dismissal before it. Resolution is mechanical — put `await dismissBadgeCelebrations(page)` inside #1614's `try`, before `target.click()`. All other insertion points auto-merge correctly (confirmed 10 surviving call sites post-merge, including the one inside #1614's rewritten `waitForFormfillSettle`), and both `export` widenings apply cleanly. #1614 doesn't touch `resolveGuidedTarget` or `revealGuidedTarget`, so the signature change here is unaffected.

**One thing the conflict resolution won't catch:** #1614 adds a **new** click site — `skipButton.click({ timeout: SKIP_SYNC_TIMEOUT_MS })` in `clickSkipButtonAndSync` — that this PR cannot know about. Post-merge it would be the only click path in the runner without celebration dismissal, and #1614 makes its failure a hard `failed`. A badge toast over the Skip control → intercepted click → 5s timeout → spurious "Skip sync failed". Whoever merges second should add the dismissal there.

Suggested order: land #1614 first and rebase this PR onto it. The changes here are one-line additions that are cheap to re-place; #1614's 247-line `execution.ts` rewrite is expensive to re-place. Small PR rebases onto the big one, not the reverse.

leslie-heinzen:
## Review follow-up

Addressed the automated and Cursor review findings:

- Added delayed first-toast and inter-toast stabilization.
- Badge dismissal now runs before target resolution, hidden-target reveal, hover, form-fill retry, guided Skip, and requirement Fix.
- All toast text reads use bounded operation timeouts and remaining transition budget.
- Toast disappearance during a read is handled without clicking a stale control.
- Visible-toast and visibility-recheck failures return bounded diagnostics with their original causes.
- The dismiss locator uses `.first()`, shared error formatting is reused, and the three-attempt invariant is documented accurately.
- Added E2E contract documentation and a source-level contract test for the badge toast test IDs.

Validation: typecheck passed, 69 focused tests passed, lint and docs sync passed, Prettier passed, and the diff check is clean.

Latest review follow-up commit: `b44a626d`.

Jayclifford345 (approved):
LGTM based on David's review and your fixes

PR #1619, opened 2026-08-14
fix(e2e): teach the requirements stub about the badge toast

Jayclifford345:
## Summary

`main` has been red since [#1617](https://github.com/grafana/grafana-pathfinder-app/pull/1617). Three tests in `tests/e2e-runner/utils/guide-runner/requirements.test.ts` fail with `Unexpected testId requested: learning-paths-badge-toast`, and every PR branched from or merged with `main` inherits it — which is how I found it, on [#1612](https://github.com/grafana/grafana-pathfinder-app/pull/1612).

No production code changes.

## Why

#1617 made `clickFixButton` call `dismissBadgeCelebrations(page)` before clicking, which does `page.getByTestId(testIds.learningPaths.badgeToast).first()`.

`requirements.test.ts` builds its `Page` stub from an explicit `locatorsByTestId` map and **throws** on any testId it does not know:

```ts
const locator = locatorsByTestId.get(testId);
if (!locator) {
  throw new Error(`Unexpected testId requested: ${testId}`);
}
```

That throw is deliberate and worth keeping — it is what catches an unstubbed lookup instead of silently returning `undefined`. It just was not updated when #1617 added a new one, so all three `attemptToFixRequirements` settle-window tests now throw before reaching the code they exist to test.

## What to look at

**The stub gains a badge-toast locator** standing for "no celebration on screen", which is the state these tests mean — they are about the requirement settle window, not about celebrations. `count` resolves 0, and `first()` returns the locator itself, since `dismissBadgeCelebrations` calls it.

**One assertion had to become badge-agnostic.** With no toast present, `waitForVisibleToast` polls its full 100ms idle window (four 25ms waits) before every fix click. `settles to met after a Fix click even on the final allowed attempt` asserted `waitForTimeout.mock.calls.length === 2`, which would now be 6. Rather than hard-code 6, it asserts on the requirement waits alone:

```ts
expect(requirementWaits(waitForTimeout)).toEqual([POST_FIX_SETTLE_DELAY_MS, REQUIREMENTS_POLL_INTERVAL_MS]);
```

That is what the test is actually about — one settle delay, then one poll — and it now says so, rather than encoding a total that changes whenever anything else in the click path waits.

## Worth knowing, not fixed here

`dismissBadgeCelebrations` costs a fixed ~100ms whenever no toast is present, which is the common case, and it runs before every fix click, hidden-target hover, and form-fill retry. That is #1617's deliberate design — the toast may not have appeared yet — but it is a real per-action cost in the runner and someone may want to revisit the idle window separately.

## How to verify

```bash
npx jest tests/e2e-runner/utils/guide-runner/requirements.test.ts --coverage=false
```

3 failed / 3 passed on `main`; 6 passed here. Full suite is green: 429 suites, 6856 tests.

The dismissal behaviour this uncovers is not left untested — `badge-celebrations.test.ts:241` already covers "dismisses the toast before a requirement Fix click".

## Breaking changes

None.

## Reversibility

Fully reversible; the diff is one test file.

### DIFF
--- a/tests/e2e-runner/utils/guide-runner/requirements.test.ts
+++ b/tests/e2e-runner/utils/guide-runner/requirements.test.ts
@@ -19,7 +19,7 @@ jest.mock('@playwright/test', () => ({
 import type { Locator, Page } from '@playwright/test';
 
 import { testIds } from '../../../../src/constants/testIds';
-import { REQUIREMENTS_POLL_INTERVAL_MS, REQUIREMENTS_SETTLE_TIMEOUT_MS } from './constants';
+import { POST_FIX_SETTLE_DELAY_MS, REQUIREMENTS_POLL_INTERVAL_MS, REQUIREMENTS_SETTLE_TIMEOUT_MS } from './constants';
 import { attemptToFixRequirements, detectRequirements } from './requirements';
 import type { TestableStep } from './types';
 
@@ -32,6 +32,13 @@ afterEach(() => {
 // ============================================
 
 const STEP_ID = 'test-step-1';
+
+/** Badge dismissal polls at 25ms; every requirement wait is longer. */
+const BADGE_POLL_MS = 25;
+
+function requirementWaits(waitForTimeout: jest.Mock): number[] {
+  return waitForTimeout.mock.calls.map(([ms]) => ms as number).filter((ms) => ms !== BADGE_POLL_MS);
+}
 
 function createTestableStep(overrides: Partial<TestableStep> = {}): TestableStep {
   return {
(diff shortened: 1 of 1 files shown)


moxious (approved):
@leslie-heinzen  test alert strikes again (in a good way) we got notified of the breakage in our build-alert channels

caught quickly, the system is working (for as much as you can call broken main "working" :lolsob: )

Good flake fix. I don't think Leslie merged through a failing build (so nothing further to do with protecting main). Seems like he merged passing tests, and then the same passing tests failed on main (homer d'oh)

leslie-heinzen:
Thank you both. @Jayclifford345 @moxious 

For the record it wasn't a flake, but a matter of not coordinating with myself well enough. The requirements tests that failed were added in #1615, which was one of the other E2E PRs and got merged at the same time. #1615 didn't have adequate prep for the #1617 changes for badge notifications clicks.

PR #1624, opened 2026-08-15
fix(e2e): bound step and browser failures

leslie-heinzen:
## Summary

The E2E runner keeps the existing operation timeout for normal step failures and evidence collection. It adds a later wall-clock backstop for operations that do not settle.

If the backstop expires, or the browser terminates during active execution, the runner closes the page and writes an `infrastructure_error` result. Normal skippable and mandatory failure behavior remains unchanged.

## What to look at

- [Step execution](https://github.com/grafana/grafana-pathfinder-app/blob/f7dfd62e64f962ff715df9611e9b647fe088bc67/tests/e2e-runner/utils/guide-runner/execution.ts) sets the backstop to twice the inner step budget plus 20 seconds. The guide timeout also reserves bounded cleanup and report-publication time.
- [Browser termination monitoring](https://github.com/grafana/grafana-pathfinder-app/blob/f7dfd62e64f962ff715df9611e9b647fe088bc67/tests/e2e-runner/utils/guide-runner/termination-monitor.ts) uses the first unexpected page crash, page close, context close, or browser disconnect event.
- [Guide execution](https://github.com/grafana/grafana-pathfinder-app/blob/f7dfd62e64f962ff715df9611e9b647fe088bc67/tests/e2e-runner/guide-runner.spec.ts) writes the infrastructure report before it fails Playwright. It prefers a drained execution result or uses the completed-step snapshot taken before cleanup.

## Why

One production browser OOM occurred approximately 135 seconds after task start. The runner then waited approximately 920 seconds before it returned.

Another deterministic guide error ran for approximately 24 minutes. The outer guide timeout did not bound a stuck operation inside one step.

This change limits those failure durations without adding a parent process watchdog or a new report contract.

## How to verify

1. Run `npm exec -- jest tests/e2e-runner/utils/guide-runner/execution.test.ts --runInBand --coverage=false`.
2. Make sure that normal skippable failures continue, while a hard runner backstop closes the page and reports infrastructure.
3. Run `npm exec -- jest tests/e2e-runner/utils/guide-runner/termination-monitor.test.ts --runInBand --coverage=false`.
4. Make sure that the first unexpected browser event wins and an expected deadline close does not become an infrastructure result.
5. Kill the Chromium browser during active execution and make sure that the report has `outcome: infrastructure_error` and `errorCode: UNKNOWN`.

## Breaking changes

None. The runner keeps report schema `1.0.0` and the existing error codes.

## Out of scope

- A parent Playwright process watchdog or process-group signaling.
- New runner error codes or a report schema change.
- Corrections for the individual guide-content failures.


moxious (changes requested):
**Classification:** `mixed` (E2E runner tooling + contract docs) · **Router:** `cli-and-e2e-runner` (strong), plus all five always-on concerns · **Coverage confidence:** high

The termination monitor is well-built — pure `EventEmitter` wiring, paired `dispose()`, genuinely tested. The hard deadline is under-budgeted, and that's the one blocking item.

---

## 1. HIGH — the hard deadline has zero headroom over the work it wraps

`tests/e2e-runner/utils/guide-runner/execution.ts:1009,1037`

`executeStep` arms `setTimeout(…, timeout)` with `timeout = options.timeout ?? calculateStepTimeout(step)`, then passes that **same value** into `executeStepWork`, which hands it in full to `waitForCompletionWithObjectivePolling` (`:1288`, `:1316`). That completion wait only begins after a sequential preamble that is never zero:

| stage                                   | bound                                         |
| --------------------------------------- | --------------------------------------------- |
| `checkLateCompletionOrDetachment`       | 2s                                            |
| `scrollStepIntoView` + settle           | 5s + 300ms                                    |
| `handleRequirementsWithFix`             | 10s + up to 3 × (10s fix + 1–2s settle) ≈ 46s |
| `waitForStepActionToAppear` / `Enabled` | 15s / 10s                                     |
| post-click settle                       | 500ms                                         |

So for a simple step the deadline is **30s** while the function's own design budget is 30s of completion wait _plus_ up to ~45s of preamble. The repo already encodes this gap: `calculateGuideTimeout` adds `STEP_OVERHEAD_TIMEOUT_MS = 20000` **per step** on top of `calculateStepTimeout(step)`, precisely because per-step wall clock exceeds the calculated step timeout. The new deadline grants zero of it. The guided path is worse by construction — `runGuidedSubstepLoop` grants each substep a fresh 30s plus 10s target resolution, and then a further full `timeout` completion wait runs afterwards.

Two consequences, both deterministic rather than probabilistic:

- **The pre-existing per-step failure path becomes unreachable.** A step that legitimately exhausts its completion wait always hits the outer deadline first (by preamble + 1s). That path returned real `consoleErrors`, `buildFailureArtifacts(…)`, and `classifyError(errorMsg)` against a still-open page. The synthetic deadline result hardcodes `consoleErrors: []`, omits `artifacts`, sets `classification: 'unknown'`, and `:1537` then suppresses the final screenshot. The most common failure mode loses all of its evidence.
- **Skippable steps now abort the whole guide.** `if (result.deadlineExceeded || (!step.skippable && stopOnMandatoryFailure))` (`:1515`) bypasses both `step.skippable` and `stopOnMandatoryFailure`, and the synthetic result stamps `skippable: false`. At base (`origin/main:1428`) a skippable step whose wait expired reached the `else if (step.skippable)` continue branch and the run could still pass. Downstream this now emits `outcome: failed` / `errorCode: MANDATORY_FAILURE` — indistinguishable from genuine guide-content drift, which is exactly the signal guide-health analysis is built on.

**Suggested action:** give the deadline a preamble allowance rather than reusing the per-operation budget — e.g. `calculateStepTimeout(step) + STEP_OVERHEAD_TIMEOUT_MS`, and roughly 2× the step budget for the guided path, whose loop and completion wait are additive by design. Add a test where a step burns preamble time and then completes inside its completion window; the three new tests only exercise a page that never settles, so nothing currently pins the deadline against the inner budgets. Consider also carrying the drained work's `consoleErrors` and `classification` into the synthetic result.

_Reversibility: reversible._ Three independent skeptic passes were tasked to refute this finding; none could, each independently recomputed the arithmetic and confirmed the base-vs-head skip-vs-fail discontinuity.

---

## 2. LOW — the termination report is the only non-passing report with no `errorCode`

`tests/e2e-runner/guide-runner.spec.ts:352-367`

`terminationResult` sets no `abortReason`, and `writeResultsFile` derives `errorCode: allStepsResult.abortReason ?? (outcome === 'failed' ? 'UNKNOWN' : undefined)` (`:101`). No abort file is written on this path, so `processPlaywrightResults` takes the plain-results branch, and `generateReport`'s fallback (`src/cli/e2e/e2e-reporter.ts:184-187`) is gated on `outcome === 'failed'`. Net result: `outcome: infrastructure_error` with `errorCode: undefined`.

That contradicts `docs/developer/E2E_TESTING.md:236` — "structured failure code **present on non-passing reports**" — in a file this PR edits, and every other producer supplies one (`REPORT_MISSING`, `NO_CAPACITY`, `PROVISIONING_FAILED`, `GRAFANA_UNREACHABLE`, `PLAYWRIGHT_SPAWN_FAILED`).

Kept at low, not medium: verification confirmed no consumer breaks. The only read of `report.errorCode` is `isAuthExpired`, where absence is correct; exit codes derive from the Playwright process result; the schema marks the field optional.

**Suggested action:** set `errorCode: 'UNKNOWN'` (already in `E2EErrorCodeSchema`, so the "no new error codes" claim still holds), or set `abortReason` so the reporter's fallback fires. One line plus one assertion.

---

## 3. LOW — the termination step list is a mutating snapshot, and the crashed step blames the guide

`tests/e2e-runner/guide-runner.spec.ts:352`

The branch reports `completedResults` — the array `onStepComplete` pushes into — snapshotted **after** `settleWithin(execution, 1000)`, while `executeAllSteps` is still running and still mutating it. On a dead browser each remaining Playwright call rejects immediately and `executeStepWork`'s catch-all (`:1345`) converts it into a `failed` step, so the in-flight step always lands in the report as a guide failure reading `Target page, context or browser has been closed`, plus up to ~4 more before the next `validateSession` checkpoint flips the loop into its AUTH_EXPIRED backfill. Meanwhile `execution` has usually already settled with the correct `AllStepsResult` — proper `not_reached` backfill, `abortReason`, `finalScreenshot` — and the branch discards it.

Verification refuted the stronger framing: there is **no** one-result-per-discovered-step contract (`createMinimalResultsData` emits `results: []` for `REPORT_MISSING`), and guide-level verdicts are safe because `isReportSuccess` requires `outcome === 'passed'`. What survives is triage noise — `summarizeSteps` and `generateMultiGuideSummary` sum step statuses with no outcome gate, so spurious rows inflate run-level failure counts.

**Suggested action:** prefer the real result when it exists — if `execution` settled during the drain, report `executionResult.results`; otherwise snapshot `completedResults` _before_ the drain.

---

## Dropped by adversarial verification

Recorded so they don't get re-raised:

- **`expectPageClose()` is a permanent latch that blinds the crash detector** — refuted 3/3. After a deadline the run always aborts, `execution` resolves in microtasks, and `dispose()` runs; the only live suppression window is the ~2s deliberate close-and-drain, which is what the latch is for. Broad suppression is also the _correct_ scope: a deliberate `page.close()` cascades into context-close and disconnect, so narrowing it to the `close` event would manufacture false `infrastructure_error` reports.
- **Multi-guide roll-up pass-count inflation** — refuted; `isReportSuccess` gates on outcome, so a crashed guide already counts as failed and no pass count is inflated.

## Contract evolution scan

Gate fired for `cli-and-e2e-runner` (`history_status: complete`, 23 prior semantic PRs). Scan verdict **`coherent_extension`**, `anchor_violated: false` → policy disposition `none`, no finding. The three failure-encoding sites are a layering that predates this PR (#1372 established the schema-backed report contract; #1477 generalized the per-call-site `outcome` literal), and `processPlaywrightResults` is untouched. The one real contract movement — observable browser death moving out of "the worker's responsibility" and into a runner-written report — is declared in the PR body and documented in both docs in the same commit. Good.

**Advisory:** this is the third-or-later use of this capability with **no anchor row**. Worth recording one in `docs/design/CONCERNS.md` (Contract anchors) naming the three-layer ownership: `executeAllSteps` owns per-step verdicts and the abort decision, `guide-runner.spec.ts` owns the guide-level report outcome, `playwright-runner.ts` owns only process-level fallbacks.

## Tech debt (non-blocking)

- **B1** `settleWithin` is byte-identical in `guide-runner.spec.ts:125` and `execution.ts:981`, and the spec re-inlines `closePageWithin`'s body with a bare `1000` literal at `:350` — three copies of one policy, two of them anonymous. Export the `execution.ts` copy through the barrel the spec already imports from.
- **B1** `StepExecutionOptions` (`:968`) and `executeAllSteps`' inline options (`:1402`) redeclare 5 of 6 fields with copied doc comments, already drifted on `artifactsDir`. `interface AllStepsOptions extends StepExecutionOptions` would collapse it.
- **A3** `executeStepWork`'s `options = {}` default and its `calculatedTimeout` fallback are unreachable — its sole caller always passes a resolved `timeout`. Two computation sites for one policy, which is how the budgets could silently diverge.
- **A1** The 30-line `executeStep` JSDoc (with `@param page` / `@returns`) is now misattached to `interface StepExecutionOptions`; the algorithm it narrates moved to `executeStepWork`, which carries no doc, and it says nothing about the deadline behavior that is now all `executeStep` does.

## Documentation drift (non-blocking)

No `AGENTS.md` / `.cursor/rules/` update needed — `tests/` is outside the tier model and `cli-and-e2e-runner` routing already covers `tests/e2e-runner/**`. Two items, both in docs this PR already touches:

- `E2E_TESTING.md:236` — the `errorCode` line is now false. Preferred fix is finding 2 (in code); if the gap is accepted instead, the line must say the field is optional and name the termination case.
- `E2E_TESTING.md:25` — the `guide-runner/` responsibility bullet should gain "browser-termination monitoring (`termination-monitor.ts`)".

## Instrumentation

Skipped — runner tooling, not product runtime.

## What CI cannot tell you

`tests/e2e-runner/utils/guide-runner/*.test.ts` mock `@playwright/test` module-wide and drive fake timers against a hand-rolled `Page`, so the 38 passing tests (verified green on this head) prove the promise plumbing and nothing about Playwright runtime semantics — which is what runner lifecycle bugs actually turn on. Before merge, someone should run a real guide and confirm: a deliberately hung step trips the deadline and closes a real Chromium page without wedging fixture teardown; `kill -9` on the browser mid-run produces an `infrastructure_error` report; and with `--trace`, `trace.zip` still lands at the path `writeTracePathFile` recorded after the page was closed early.

leslie-heinzen:
Addressed the review findings in `02a0b8ee`:

- The inner step timeout remains the normal operation and evidence budget.
- The hard runner backstop is now `2 × step timeout + 20s`.
- The outer guide timeout includes every step backstop plus 3 seconds for bounded close, drain, and report publication.
- Normal skippable failures still continue. Only the hard runner backstop produces an infrastructure outcome.
- Non-passing reports without a specific code now use `UNKNOWN`.
- Browser termination snapshots completed steps before cleanup and prefers the real execution result when it drains.
- Infrastructure reports are written before Playwright exits nonzero.
- The bounded settlement helper and execution option types are shared.
- The E2E ownership contract and documentation are updated.

Validation:

- 176 focused tests passed.
- Full Jest: 441 suites and 7,047 tests passed.
- Typecheck, CLI build, formatting, and diff checks passed.
- Lint has 0 errors and 24 existing warnings.
- Real Chromium checks passed for page closure, hard-deadline cancellation, and browser-process `SIGKILL`.
- Independent follow-up review returned no findings.

moxious (approved):
Re-reviewed at `f7dfd62e`. Both blockers from my earlier review are properly fixed, no new blocking finding survived verification, all 30 checks green. Approving.

## Prior findings — closed

**H-1 — backstop equal to the inner step budget.** Fixed. `calculateStepDeadline = calculateStepTimeout * 2 + STEP_OVERHEAD_TIMEOUT_MS`, and `calculateGuideTimeout` now sums per-step deadlines plus a 3s cleanup grace, so the outer guide timeout can no longer fire ahead of the per-step backstop. The `2×` is the right shape for guided steps specifically, because `runGuidedSubstepLoop` grants n × 30s and then a further full `timeout` completion wait runs.

**H-2 — deadline path bypassed `skippable` and emitted false content failures.** Fixed. The synthetic result carries `skippable: step.skippable` and `classification: 'infrastructure'`, `executeAllSteps` sets a distinct `infrastructureError` flag with no `abortReason`, and the guide report comes out as `outcome: 'infrastructure_error'` rather than `failed` / `MANDATORY_FAILURE`. A consumer computing content-failure rate now excludes it at both step and report level, which was the point.

**L-1 — missing `errorCode`, contradicting `E2E_TESTING.md`.** Fixed in both writers. I checked the `'skipped'` outcome does not regress: `SKIPPED_PREREQ` arrives via `data.abortReason` and wins ahead of the new `UNKNOWN` fallback.

**L-2 — mutating snapshot of the termination step list.** Fixed, with the drained real result preferred over the snapshot.

The `cli-and-e2e-runner` contract anchor landed in this PR rather than being deferred. Thank you.

## Dropped by verification — recorded so it is not re-raised

I pursued a finding that the backstop grants a flat ~50s of headroom over legitimate inner work for every step shape, while `executeStepWork`'s preamble has a much larger bounded ceiling — so a step could be killed before reaching its own `Requirements not met after N fix attempt(s)` verdict, taking the whole guide with it.

Adversarial verification refuted it, and correctly: that verdict returns from `executeStepWork` **before** the button-appear/enable waits and before the inner completion wait, so preamble cost and inner budget sit on mutually exclusive paths and must not be stacked into one worst case. The residual window is narrow — 3 × ~27s fix attempts plus a 10s initial check ≈ 84–94s against an 80s backstop for a plain step, reachable only if every bounded wait maxes out simultaneously, which is not meaningfully distinguishable from the stall the backstop exists to catch.

If you want belt-and-braces, the cheap version is a single test pinning `calculateStepDeadline(step)` above `REQUIREMENTS_CHECK_TIMEOUT_MS + MAX_FIX_ATTEMPTS × (FIX_BUTTON_TIMEOUT_MS + NAVIGATION_FIX_SETTLE_DELAY_MS + FIX_BUTTON_TIMEOUT_MS / 2 + REQUIREMENTS_CHECK_TIMEOUT_MS)`, so a later bump to `MAX_FIX_ATTEMPTS` cannot silently close the gap. Not a merge condition.

## Non-blocking

1. **The anchor row is slightly under-specified.** It names three owners, but this PR also establishes a fourth invariant in `e2e-reporter.ts` — every non-passed report carries an `errorCode`, `UNKNOWN` when nothing more specific applies — and `generateReport` remains a second `outcome` deriver alongside `guide-runner.spec.ts`. One more clause would keep the next PR from re-fracturing it.

2. **Reversibility is `partially_reversible`.** Reverting restores the code, but reports emitted while this is live carry `infrastructure_error` where they previously carried `failed`. Given the guide-health analysis currently running, anyone querying by outcome across that boundary needs to know the distribution shifts. Worth a line in the PR body.

3. **Dead line.** In `guide-runner.spec.ts`, `terminationMonitor.expectPageClose()` in the terminated branch runs after the `finally` block already called `dispose()`, so the listeners are gone and the call does nothing.

4. **Tech debt, for later.** `settleWithin` and `BoundedSettlement` are generic promise utilities now living in the 1670-line `execution.ts` and re-exported from the package index. Relatedly, `executeStep`'s hand-rolled `settled`-flag race does approximately what `settleWithin` does and could likely collapse into it.

## Coverage note

These unit tests fake `@playwright/test` module-wide and the spec-level termination branch has no unit coverage, so green CI is weaker evidence than usual for this subsystem. The behaviour that matters most — a real browser OOM mid-guide — rests on the manual step 5 in "How to verify".

PR #1625, opened 2026-08-15
docs: correct guide schema descriptions

### LINKED ISSUE #1545: Two guide-schema .describe() strings are wrong and reach authors via CLI help
Two `.describe()` strings on the guide schema say things that are not true of the runtime. Both are surfaced to authors through the CLI `--help` walker (`src/cli/utils/schema-options.ts`) and through the MCP authoring server, so they are author-facing text, not just internal comments.

## 1. `snippetId` — "resolve at parse time"

```ts
// src/types/json-guide.schema.ts:736
snippetId: SnippetIdSchema.describe('Upstream snippet ID to resolve at parse time'),
```

Resolution does not happen at parse time. The guide is Zod-validated first (`src/docs-retrieval/content-fetcher.ts:230`, then `src/docs-retrieval/json-parser.ts:111`), and expansion happens after validation and before render (`src/components/content-renderer/content-renderer.tsx:599-614`). `json-parser.ts:315-321` even carries a dedicated `case 'snippet-ref':` fallback arm whose comment notes that reaching it means resolution failed — which could not happen if refs were gone by parse time.

The block comment at `:729-730` gets it right ("Resolves at parse time — never reaches the renderer" is half right; the second clause is the accurate one). Suggested wording: `'Upstream snippet ID, resolved after validation and before render'`.

## 2. `mode` — "'coda' (default)"

```ts
// src/types/json-guide.schema.ts:544-549
mode: z.enum(['coda', 'standard']).optional().describe(
  "... 'coda' (default) runs in a Coda VM with a terminal ..."
),
```

The field is `.optional()` with no `.default()`, so the schema itself defaults nothing. Three different things happen depending on the path:

- omitted in JSON → the runtime treats it as `'coda'` (`src/components/interactive-tutorial/challenge-block.tsx:180`)
- a new block in the block editor → `'standard'` (`ChallengeBlockForm.tsx:62-64`)
- a legacy block opened in the editor → inferred `'coda'` (`ChallengeBlockForm.tsx:65-83`)

"(default)" without saying whose default is the misleading part, and it is the one an author reading CLI help sees. Suggested wording: name it as the runtime default for JSON that omits the field, and note the editor seeds `'standard'`.

## Why file this rather than fix it in #1538

#1538 is a docs-only PR and #1531 is in flight on the same file. Also worth recording: this is the concrete reason not to generate the reference table from `.describe()` strings, as floated in #1530 — generation would have propagated both of these defects into the reference instead of catching them.

Refs #1530, #1538

lijiayou0728:
## Summary

Correct the author-facing guide schema descriptions for `challenge.mode` and `snippet-ref` resolution.

Snippet references are now described as resolving after guide validation and before rendering. The challenge mode description now distinguishes between the absence of a schema default, the runtime `coda` fallback, and the block editor's `standard` initial value.

## What to look at

- Align the exported schema descriptions, type comments, CLI documentation, and E2E diagnostic wording with the actual validation and expansion order.
- Add regression coverage for the two author-facing schema descriptions.

## How to verify

1. Export the block schema and confirm the `snippetId` description says references resolve after validation and before rendering.
2. Confirm the `challenge.mode` description distinguishes the absent schema default from the runtime `coda` fallback and the editor's `standard` initial value.
3. Compare the descriptions with the challenge and snippet-reference sections of the JSON guide format documentation.

## Local verification

`git diff --check` passed. The focused Jest test and full project checks were not run locally because dependency installation failed with a registry connection reset. This Draft PR will use CI for the complete test result.

## Breaking changes

None. This changes documentation and schema descriptions only; validation and runtime behavior are unchanged.

Fixes #1545

🤖 Generated with [Codex](https://openai.com/codex/)

### DIFF
--- a/docs/developer/CLI_TOOLS.md
+++ b/docs/developer/CLI_TOOLS.md
@@ -337,7 +337,7 @@ The output is a D3-compatible JSON object with `nodes`, `edges`, and `metadata`:
 
 ## Build-snippets command
 
-Generates a snippet catalog (`index.json`) from a directory of snippet bodies. Snippet bodies (`<id>.json`) are the source of truth; the catalog is always regenerated, never hand-edited. The catalog is consumed by the snippet engine, which resolves `snippet-ref` blocks by fetching `<id>.json` at parse time.
+Generates a snippet catalog (`index.json`) from a directory of snippet bodies. Snippet bodies (`<id>.json`) are the source of truth; the catalog is always regenerated, never hand-edited. The catalog is consumed by the snippet engine, which resolves `snippet-ref` blocks by fetching `<id>.json` after guide validation and before render.
 
 ### Basic syntax
 
--- a/src/cli/e2e/side-effects.ts
+++ b/src/cli/e2e/side-effects.ts
@@ -197,7 +197,11 @@ function classifyBlock(block: JsonBlock, path: string): SideEffectClassification
     return reason('unknown', path, `${block.type} block can have side effects outside static guide analysis`);
   }
   if (isSnippetRefBlock(block)) {
-    return reason('unknown', path, 'Snippet content is resolved after validation and must be classified after expansion');
+    return reason(
+      'unknown',
+      path,
+      'Snippet content is resolved after validation and must be classified after expansion'
+    );
   }
   return reason('unknown', path, 'Unknown block type');
 }
--- a/src/cli/e2e/side-effects.ts
+++ b/src/cli/e2e/side-effects.ts
@@ -197,7 +197,7 @@ function classifyBlock(block: JsonBlock, path: string): SideEffectClassification
     return reason('unknown', path, `${block.type} block can have side effects outside static guide analysis`);
   }
   if (isSnippetRefBlock(block)) {
-    return reason('unknown', path, 'Snippet content is resolved at parse time and must be classified after expansion');
+    return reason('unknown', path, 'Snippet content is resolved after validation and must be classified after expansion');
   }
   return reason('unknown', path, 'Unknown block type');
 }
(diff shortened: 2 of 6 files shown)


moxious (changes requested):
Thank you for your contribution!  One change needed and then it's mergeable:

The PR rewrote this exact sentence and carried an error through it unchanged: *"the snippet engine then fetches the snippet by id (CDN-first, bundled fallback)"*. There is no bundled fallback. getSnippetResolver() (src/snippet-engine/caching-snippet-resolver.ts:76-81) composes exactly new CachingSnippetResolver(createOnlineSnippetResolver()); the word "bundled" appears nowhere under src/snippet-engine/; a failed fetch returns { ok: false, code: 'network-error' } with no second source. .cursor/rules/systemPatterns.mdc:127 states the opposite in so many words: "Fetches from CDN … no bundled fallback."

Why it matters: this is the @coupling-designated contract doc for the block type, and the inaccuracy is the one that changes author behavior — it implies a broken ref degrades to a local copy, when the real outcome is the inert placeholder documented at json-guide-format.md:1126. A PR whose entire deliverable is prose accuracy should not leave a falsehood in a line it owns.

Suggested action: drop the parenthetical, or make it (CDN only — a failed fetch yields an inert placeholder). One-clause edit inside a line already in the diff.

---

Non-blocking comments:

Doc drift, outside the diff. .cursor/rules/systemPatterns.mdc:127 says inline-refs.ts splices "before parsing" — true on the launch path, false on the renderer path. The same stale premise sits in two code comments: src/docs-retrieval/json-parser.ts:315 ("Refs are inlined upstream before parsing") and src/snippet-engine/inline-refs.ts:3-4. Natural riders on this PR since it's the same correction, or a maintain-docs backlog entry.

moxious:
hi @lijiayou0728 do you have a chance to return to this?

moxious:
hi @lijiayou0728 -- we're closing this because it's conflicted.  We do appreciate contributions though, and would be very open to submitting new PRs (or pushing to this branch cleanups and re-opening)

PR #1632, opened 2026-08-17
fix(block-editor): default challenge brief on conversion

### LINKED ISSUE #1575: Converting image, video or empty markdown to a challenge block fails (brief is never populated)
## Summary

Converting an `image`, a `video`, or an empty `markdown` block to a `challenge` block fails with a "Conversion failed" toast. Every other source type converts to `challenge` fine.

Split out of #1523, which is a strictly zero-functional-change ratchet and deliberately leaves this behavior as-is.

## Mechanism

`challenge` requires `brief` (`src/types/json-guide.schema.ts:552` — `brief: z.string().min(1)`).

`REQUIRED_DEFAULTS.challenge` (`src/components/block-editor/utils/block-conversion.ts:86-89`) supplies `title` and `successCriteria` but **not** `brief`. `brief` is instead populated indirectly, via `CONTENT_FIELDS.challenge = 'brief'` (`:59`), which copies the source block's primary content field into it.

So the conversion succeeds precisely when the source has a non-empty content field:

| Source | `CONTENT_FIELDS` entry | Result |
| --- | --- | --- |
| `markdown` (with content), `html`, `interactive`, `multistep`, `guided`, `quiz`, `input`, `terminal`, `terminal-connect`, `code-block` | present | ✅ works |
| `image`, `video` | **absent** | ❌ `brief` never set → `safeParse` fails at `:224-228` → throws |
| `markdown` with `content: ''` | present but falsy | ❌ skipped by the guard at `:188-193` → same failure |

This is the same shape as #619, on a newer block type.

## User impact

An author picks "convert to challenge" from the type-switch dropdown on an image or video block and gets a generic error toast with no explanation of why that particular pair is unsupported. The failure surfaces at `useBlockConversionHandlers.ts:246-249` as an `alert-error`, so it is visible but not diagnostic.

## Suggested fix

Add a `brief` default to `REQUIRED_DEFAULTS.challenge`, matching how `image`/`video`/`interactive` already get placeholder defaults at `:79-82`.

**Also decide `mode` while here.** `challenge-block.tsx:180` defaults `mode` to `'coda'`, and `inferInitialMode` also lands on `'coda'` for a block with no VM fields — so a converted challenge silently becomes a **VM-provisioning** challenge. Emitting `mode: 'standard'` on conversion is probably the kinder default, but it is a product call, not an obvious one: it decides which authoring path a converted challenge lands on.

## Not the fix

Do not resolve this by removing `challenge` from the conversion target list. It would break the ten source types that convert successfully today, and `block-conversion.test.ts:349` and `:124` both pin that working behavior.

## Related

- #1523 — the zero-change totality ratchet this was split out of
- #619 — the original instance of this shape
- #804 — open, same file

fly1d:
## Summary

Converting an image, video, or empty markdown block to a challenge now succeeds instead of failing schema validation. The conversion supplies a non-empty placeholder brief only when the source has no usable text; content-bearing sources continue mapping their content into the challenge brief.

## How to verify

1. In the block editor, switch an image block to a challenge and confirm the conversion succeeds with `Complete this challenge` in the brief field.
2. Repeat with a video block and an empty markdown block.
3. Switch a non-empty markdown block to a challenge and confirm its original content is used as the brief instead of the placeholder.

## Breaking changes

## Out of scope

- Challenge `mode` selection remains unchanged; it is a separate product decision per the issue scope.

Fixes #1575

Generated with [OpenAI Codex](https://openai.com/codex/).

### DIFF
--- a/src/components/block-editor/utils/block-conversion.test.ts
+++ b/src/components/block-editor/utils/block-conversion.test.ts
@@ -501,16 +501,14 @@ describe('convertBlockType', () => {
      * Sources with no `CONTENT_FIELDS` entry — `image`, `video`, `collapsible`,
      * `assistant` and `snippet-ref` — carry no text into a target whose required
      * text field has no `REQUIRED_DEFAULTS` fallback, so the conversion throws.
-     * Pinned here rather than fixed, so a fix has to move this list deliberately;
-     * the `image`/`video` symptom is tracked as
-     * https://github.com/grafana/grafana-pathfinder-app/issues/1575.
+     * Pinned here so adding a fallback moves this list deliberately.
      */
     const CONTENTLESS_SOURCE_FAILURES: Partial<Record<BlockType, readonly BlockType[]>> = {
-      image: ['markdown', 'html', 'interactive', 'quiz', 'input', 'terminal', 'challenge'],
-      video: ['markdown', 'html', 'interactive', 'quiz', 'input', 'terminal', 'challenge'],
-      collapsible: ['markdown', 'html', 'interactive', 'quiz', 'input', 'terminal', 'challenge'],
-      assistant: ['markdown', 'html', 'interactive', 'quiz', 'input', 'terminal', 'challenge'],
-      'snippet-ref': ['markdown', 'html', 'interactive', 'quiz', 'input', 'terminal', 'challenge'],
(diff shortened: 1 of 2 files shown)


moxious (approved):
Thanks for the contribution!  LGTM

PR #1633, opened 2026-08-17
fix: forward interactive step props

### LINKED ISSUE #1506: content-renderer silently drops openGuide (and formHint, validateInput) props on top-level JSON steps
## Summary

`content-renderer.tsx` forwards interactive-step parser output to `InteractiveStep` through a hand-maintained explicit prop list. Three fields are parsed but never forwarded, so setting them on a top-level JSON-authored step silently no-ops:

- **`openGuide`** — a navigation feature: a `navigate` step with `openGuide` set should open the named guide after navigating. The prop is consumed by `InteractiveStep` (`src/components/interactive-tutorial/interactive-step.tsx:187` → `data-openguide` at `:970`) but never reaches it, so the behavior silently does nothing.
- **`formHint`** — disables the formfill hint UX.
- **`validateInput`** — disables formfill input validation UX.

These are **pre-existing** drops (not introduced by any recent change). They were surfaced by the prop-forwarding tripwire test added in #1493 (`src/components/content-renderer/content-renderer.prop-forwarding.test.ts`), which currently documents them in a `KNOWN_UNFORWARDED` allowlist rather than fixing them — a deliberate choice to land the tripwire without widening that PR's blast radius.

## Why this matters

`openGuide` is a real navigation feature that authors can set in JSON and reasonably expect to work; today it silently fails with no error, no warning, and passing unit tests. `formHint` / `validateInput` degrade the formfill authoring experience but are lower impact.

## Suggested fix

Forward the three fields from `content-renderer.tsx` (the `interactive-step` case) and remove them from the `KNOWN_UNFORWARDED` allowlist in `content-renderer.prop-forwarding.test.ts` (the tripwire asserts bidirectionally, so un-dropping a field without updating the allowlist will fail the test — update both together). **Prioritize `openGuide`**, since it's a functional navigation feature rather than polish.

## References

- Surfaced by #1493 (adds the tripwire that documents these drops)
- Allowlist: `src/components/content-renderer/content-renderer.prop-forwarding.test.ts`
- Consumer: `src/components/interactive-tutorial/interactive-step.tsx`

mikemikimike:
Closes #1506

## Summary

Forward interactive-step fields that were parsed from top-level JSON guides but silently dropped before reaching the runtime:

- `openGuide`
- `formHint`
- `validateInput`

`openGuide` now works through all relevant execution paths, including direct React actions, section orchestration (`Do Section`), controller/cross-tab commands, and live-tab execution. Explicit `openGuide` takes precedence over the legacy URL `?doc=` fallback, while the fallback remains supported for backward compatibility.

The form target observer now handles inputs that appear after the step mounts, while remaining gated and debounced. The prop-forwarding tripwire and step registry tests were updated, and regression coverage was added for guide fallback precedence.

## What to look at

- `interactive-step.tsx` — the form target resolution moved from a `useMemo` (which cached `null` when the input was not yet in the DOM) to a `useState` + debounced `MutationObserver`. It only observes while a formfill step is armed: `validateInput === true`, enabled, not complete, not disabled.
- `cross-tab.types.ts` — `openGuide` reaches the wire by derivation (`Omit<InternalAction, 'requirements'>`) rather than by hand-listing, so the next field to land there needs no one to remember.
- `content-renderer.prop-forwarding.test.ts` — `KNOWN_UNFORWARDED` is now empty and load-bearing: deleting the `openGuide=` line from the renderer fails the tripwire by name.

## Scope

`openGuide` is accepted on `JsonInteractiveBlockSchema` but deliberately not on `JsonStepSchema`, so a step nested inside a multistep cannot set it. Widening it there is a separate change with its own wire and orchestration questions.

`openGuide` fires during the `navigate` action and only on the internal-navigation branch — external `http(s)` targets go through `window.open` and never consult it. The schema `.describe()` strings for both `openGuide` and `formHint` were corrected to say what the runtime actually does, since those strings feed the CLI and MCP authoring surfaces.

## How to verify

Behavioural coverage for the newly live behaviour is in `interactive-step.test.tsx`:

- non-matching input renders the `formHint` banner, matching input completes the step
- the same, for a target input that is appended to the DOM *after* the step mounts

Both mutations raised in review are killed by these tests — forcing `enabled: false` on the form-validation predicate fails 2 tests, and neutering the `MutationObserver` fails 1.

Local run of the full `npm run check` gate passes: typecheck, lint, prettier, docs terms, `lint:go`, `test:coverage`, review-contract, and script suites. `test:go` reports a stale `struct-tags.json` golden (`json.RawMessage` vs `jsontext.Value`) — that reproduces on unmodified `main` with a Go toolchain newer than the pinned `go 1.26.5`, and is unrelated to this PR.

## Breaking changes

None. Every field added is optional, and the `?doc=` fallback is unchanged for guides that rely on it.

## Reversibility

Straightforward revert — the change is additive prop forwarding plus one localised rewrite of form-target resolution.


moxious (changes requested):
Thanks for closing this out — formHint and validateInput are wired correctly, and it's good to see the tripwire's gap list actually shrinking.

The one thing I'd like changed before merge is KNOWN_UNFORWARDED: string[] = []. That list is a ledger, and emptying it claims all three fields now work. openGuide doesn't. The renderer forwards it, but InteractiveStep only writes it to data-openguide, and the React execution path never reads that attribute:

- InteractiveActionRequest (src/types/interactive.types.ts:56-59) has no openGuide field, so executeInteractiveAction (interactive.hook.ts:395-405) rebuilds InteractiveElementData without it.
- The only reader of data-openguide (dom-utils.ts:68) is reached via the sequence selector .interactive[data-targetaction] (interactive.hook.ts:344), which never matches a React step's interactive-step class.

So navigate-handler.ts:108's data.openGuide stays undefined and only the ?doc= fallback works. I checked this by rendering a navigate step with openGuide set and spying on findDocPage — zero calls; the ?doc= control got one.

Not a regression (it was equally dead before), but it does mean the empty list plus the "keep this list empty" comment is the only written record of the gap, and it's now gone. Issue #1506's suggested fix would read as done while the field still no-ops.

Smaller things, none blocking:

- validateInput / formHint only arm if the target input is already in the DOM when the step mounts — formTargetElement is a useMemo over author-static deps (interactive-step.tsx:397-402) and caches null. For navigate → formfill on the destination page the new behavior never fires. Worth a follow-up issue at least; happy to file it.
- No behavioral test covers what this turns on — I forced the enabled predicate at interactive-step.tsx:433-438 permanently off and all 446 suites stayed green. One RTL test on InteractiveStep (non-matching input → hint shows, matching → completes) would cover it and retire the it.todo at data-attributes.contract.test.tsx:860.
- formHint's schema .describe() (json-guide.schema.ts:167 and :322) says "placeholder text", but the runtime renders it as the validation-failure banner. That string feeds the CLI and MCP authoring server, so it's worth correcting now that the field is live.
- forwardedProps() guards the start index but not the end — if the arm ever becomes self-closing, slice(start, -1) scrapes the rest of the file and the tripwire passes unconditionally. One-line fix while you're in there.

Jayclifford345:
@cursoragent review

mikemikimike:
Added regression coverage in commit c07f53c for NavigateHandler guide fallback precedence:\n\n- An explicit openGuide takes priority over the URL ?doc= parameter.\n- The URL ?doc= parameter is used when openGuide is absent.\n- No guide is opened when neither source is present.\n\ngit diff --check passes. Jest could not be run locally because npm dependencies are not installed. Please re-review the latest HEAD.

mikemikimike:
Addressed the valid Cursor Bugbot finding in commit dd85cff2: openGuide is now preserved in StepInfo and INTERACTIVE_STEP_SCHEMA.toStepInfoExtension, so Do Section execution forwards it through xecuteInteractiveAction. Added a registry regression assertion.\n\nThe controller wire payload and live-tab executor already carry openGuide on the latest HEAD. The form observer is also gated by alidateInput/step state and uses a 50ms debounce.

moxious (commented):
Re-reviewed at 9e104295. The `openGuide` blocker from my last round is properly closed. I'm keeping changes-requested on for the test gap, though — my reasoning is below.

`openGuide` is now wired through every hop I called out: `InteractiveActionRequest` carries it, `executeInteractiveAction` puts it on `InteractiveElementData` (the exact gap), `resolveGuideParam` gives it precedence over the `?doc=` fallback, and section orchestration carries it via `StepInfo` + the registry mapping. I like that cross-tab got fixed by derivation rather than by hand-listing — `Omit<InternalAction, 'requirements'>` means the next field to land on that wire won't need anyone to remember.

I checked the tripwire rather than taking the empty list on faith: deleting the `openGuide=` line from `content-renderer.tsx` fails `content-renderer.prop-forwarding.test.ts` naming the field. The ledger is load-bearing now, which was the whole point. Also confirmed `resolveGuideParam` is only reachable from `handleDoMode`, so "Show me" doesn't open the destination guide.

On the test gap. I flagged this as non-blocking last round on the assumption it was a pre-existing hole. It isn't anymore — this round added new runtime behaviour with no coverage, so the untested surface grew rather than shrank. Two mutations, both on the current head:

- forcing `enabled: false` on the form-validation predicate (interactive-step.tsx:473-482) leaves 854 tests passing, 0 failing
- neutering the new `MutationObserver` so late-appearing inputs never resolve leaves 838 passing, 0 failing

The second one is the new part. The `useMemo` → `useState` + debounced body-observer rewrite is the riskiest code in this PR — it observes `document.body` with `subtree: true` for every armed formfill step — and nothing exercises it. If it regresses, or if someone simplifies it later, CI stays green and the failure shows up as a guide that silently stops validating input. That's the same class of silent-drop bug this PR exists to fix, so I don't want to trade one for the other on the way in.

The `openGuide` half is genuinely well covered — hook, handler, cross-tab, registry, controller mode. It's specifically `formHint` / `validateInput` that's dark, and `it.todo('InteractiveStep: data-test-form-state changes during form validation')` at data-attributes.contract.test.tsx:860 is still sitting there describing exactly the missing test.

One RTL test on `InteractiveStep` — non-matching input shows the hint banner, matching input completes the step — kills both mutations and retires that todo. That's the only thing I'm holding on; ping me when it's up and I'll turn this around quickly.

Two nits while you're in these files, neither blocking:

- `json-guide.schema.ts:330` describes `openGuide` as "Guide ID to open when this block completes", but it fires during the navigate action, only for `targetAction: 'navigate'`, and is silently ignored for external `http(s)` targets since the `window.open` branch never calls `resolveGuideParam`. Same drift you just corrected on `formHint`, and it feeds the CLI and MCP authoring surface.
- `openGuide` is on `JsonInteractiveBlockSchema` but not `JsonStepSchema`, so a step nested in a multistep can't set it. I assume that's deliberate — worth a line in the PR body either way.

moxious (dismissed):
Thanks for the contribution!

It closes a gap between what the authoring service promises and what the experience actually delivers.

Three fields that guide authors could already write — and that the schema happily accepted — did nothing at runtime. Authors wrote them, learners never saw them. This makes all three real:

- A learner who types the wrong value into a form field now gets told what's expected. Before, the step just sat there doing nothing — no error, no hint, no progress. For a product aimed at newcomers, silence is the worst possible response to a typo; it reads as "the guide is broken" and it's where people quit.
- Form validation now works on the page you just navigated to. Previously the input had to already exist the moment the step mounted, which excluded the single most common shape in real guides: go here, then fill this in. So the feature was dead exactly where authors would reach for it.
- A navigation step can hand the learner off to the right guide when they arrive. That's the primitive that lets guides chain into each other instead of dead-ending.

moxious (dismissed):

leslie-heinzen (dismissed):

Jayclifford345 (dismissed):

Jayclifford345:
Howdy @mikemikimike, I would love to get this in; we just unfortunately have some commits which are unsigned. If you can sign them and push again that would be great!

Jayclifford345 (dismissed):

mikemikimike:
I checked the local signing setup. This host has no GPG executable or configured signing key, so I cannot safely rewrite the existing three commits as verified commits without introducing a history rewrite. The PR head is currently `8c07fd0` on `fix/forward-interactive-step-props`.

If GitHub accepts SSH-signed commits for this repository, I can add a follow-up signed commit, but that would not make the existing commits verified. Please confirm whether you prefer that route or a maintainer-side rebase/signing pass.

moxious (dismissed):

moxious:
@mikemikimike it looks like history rewrite is your only option, all commits must be signed; but you'll be rewriting history on your fork, not on ours.  Once again, everything's looking good, but we can't merge without signing.

moxious:
The unsigned commits are:

  - 7914094b91287ee1d8990a746b7acdd260cb152b
    (https://github.com/grafana/grafana-pathfinder-app/commit/7914094b91287ee1d8990a746b7acdd260cb152b)

  - 31ad8935d0f7307f05fe97ea20ff8ca413fd44af
    (https://github.com/grafana/grafana-pathfinder-app/commit/31ad8935d0f7307f05fe97ea20ff8ca413fd44af)

  - 1baf8ddf1f03da072c007e8c8548fe0379a69b43
    (https://github.com/grafana/grafana-pathfinder-app/commit/1baf8ddf1f03da072c007e8c8548fe0379a69b43)

moxious:
Thanks for forwarding the interactive-step props (openGuide / formHint / validateInput) through the runtime — closes #1506 nicely.

CI is failing on this one, and two distinct failures trace to this PR's own diff:

- `Prettier` reports formatting issues in `src/components/content-renderer/content-renderer.tsx` and `src/types/interactive.types.ts` — both files this PR reformatted into single-line signatures that don't match the repo's Prettier config.
- `Test` and `Build CLI, build image, smoke test` both report the same failure: `schema command › exportSchema › exports accurate challenge mode and snippet reference descriptions` in `src/cli/__tests__/schema-command.test.ts`, which still expects the old `snippetId` description ("Upstream snippet ID, resolved after validation and before render") that this PR's `json-guide.schema.ts` change replaced with "Upstream snippet ID to resolve at parse time".

Run `npm run prettier-fix` and update `schema-command.test.ts`'s expected string to match the new description, and CI will rerun; we'll review once it's green.

moxious (changes requested):
Thanks for forwarding `openGuide`, `formHint` and `validateInput` through the renderer into the runtime — that closes the exact gap the `KNOWN_UNFORWARDED` tripwire was left open to record, and the new formfill coverage (including a target that only appears after the step mounts) is the part real guides most needed. CI is green on this head apart from the fork-permission comment write, which is ours, not yours.

Almost all of what follows traces to one mechanical cause rather than to your judgement: the branch's first commit rewrote three files from an older copy instead of editing them, so main's newer content got overwritten. A rebase resolves most of it. One of the four items is my own fault — my 2026-08-31 comment read the schema revert backwards and asked you to flip a test to match it; please disregard that instruction, it's covered in B1.

---

## Merge contract

1. [blocking] **B1 — Stale base overwrites merged schema prose and inverts the tripwire that guarded it** (medium · guide-schema-and-contracts)
   The branch's first commit (8e5780f4) rewrote src/types/json-guide.schema.ts from an older file state rather than editing it, so HEAD is a byte-exact revert of merged PR #1703 (82cf647d): snippetId goes back to 'Upstream snippet ID to resolve at parse time', and the challenge mode description loses 'The schema has no default: JSON that omits mode resolves to coda at runtime, while the block editor seeds a new block with standard. Set mode explicitly.' Both strings reach authors and MCP-driven authoring agents through exportSchema. 'At parse time' is contradicted by src/snippet-engine/inline-refs.ts, which replaces every snippet-ref before the parser runs; and the restored "'coda' (default)" describes a default the schema does not have (z.enum([...]).optional(), no .default()), while the divergence #1703 documented is still live -- challenge-block.tsx defaults an absent mode to 'coda' and ChallengeBlockForm.tsx seeds new blocks 'standard'. All four assertions in src/cli/__tests__/schema-command.test.ts:74-77 are now inverted, so the tripwire actively requires the pre-#1703 text and #1703 cannot re-land without a second flip. The same stale-base commit also deleted JsonDividerBlockSchema and JsonCalloutBlockSchema (added by #1702 immediately before the base), which a later commit patched back, and dropped the two @coupling docblocks. Note: the test inversion followed a reviewer instruction of 2026-08-31 that misread the revert's direction, so it is not the contributor's misjudgement.
   Required: Rebase on current main and take main's side for src/types/json-guide.schema.ts rather than re-applying the branch's copy, then re-apply only the intended formHint description change. Restore the #1703 prose at the mode and snippetId describes, restore the two @coupling docblocks, and un-invert schema-command.test.ts:74-77 back to asserting the corrected strings.

2. [blocking] **B2 — Deletes the handoff-gate suite the contract registry names as required verification** (medium · full-screen-sidebar-handoff)
   HEAD removes describe('Full-screen sidebar handoff gate') (5 tests) and the jest.mock('../global-state/panel-mode') block from src/interactive-engine/interactive.hook.test.ts, while the runtime gate in interactive.hook.ts is byte-identical -- so nothing fails today. docs/design/CONCERN_DETAILS.md names that suite as verification item 5 for this concern, and it now points at a suite that does not exist. Applying the deletion test, three invariants lose their only coverage anywhere in the tree: that executeInteractiveAction forwards the resolved fullScreenFallbackLocation as { targetPath: ... }; that the gate stamps elementData.skipCompletionOnEmptyTarget; and that executeInteractiveAction returns 'error' when elementData.completionSuppressed is set. The surviving { targetPath: ... } assertions at interactive-guided.test.tsx:698 and code-block-step.test.tsx:51 belong to independent call sites in those components, and code-block-step.test.tsx mocks useInteractiveElements outright; interactive.hook.fullscreen-handoff-timing.test.ts asserts handler ordering only; content-renderer.fullscreen-fallback.test.tsx asserts the handoff fired, never its argument. Every surviving skipCompletionOnEmptyTarget and completionSuppressed test injects the flag by hand at the handler level. Two of the three are the guard against marking a step complete whose target was never found. The mock also appears to be gratuitous collateral: it is orthogonal to the new openGuide navigate assertion and could have coexisted with it.
   Required: Re-home the three assertions rather than restoring the whole suite -- interactive.hook.fullscreen-handoff-timing.test.ts already mounts the hook in full screen with a mocked ButtonHandler and is the natural host, and its docblock needs rewriting anyway since it is written as a delta against the deleted suite. Then update verification item 5 in docs/design/CONCERN_DETAILS.md to name the new host.

3. [blocking] **B3 — Reintroduces the vestigial KNOWN_FIELDS._manifest entry, already two fields out of date** (medium · guide-schema-and-contracts)
   HEAD adds _manifest to KnownFieldsMetaKey and an 18-name set to KNOWN_FIELDS in src/types/json-guide.schema.ts. That entry is absent at the base: it was deliberately removed by commit b3e51e18, 'drop vestigial KNOWN_FIELDS._manifest; fix drifted schema table'. This is the same stale-base overwrite as B1, so the removal is silently undone. The reintroduced set is already wrong -- ManifestJsonObjectSchema in src/types/package.schema.ts declares 20 fields and the set omits estimatedMinutes (documented as an author-provided time estimate and read at build-repository.ts:196 and package-content.ts:96) and stats. Nothing reads it: src/validation/unknown-fields.ts consults only KNOWN_FIELDS[blockType], ._step, ._choice and ._guide. Nothing guards it either: type-coupling.test.ts names schemas one at a time with no _manifest case, and upsert-script-crd-fields.test.ts skips every underscore-prefixed key. The satisfies guard forces the entry to exist once the meta key is added, and nothing in this PR's stated scope needs that meta key. ManifestJsonObjectSchema is also a z.looseObject documented 'Loose by design: unknown top-level keys are extension metadata', which is at odds with keeping an unknown-field allowlist for it at all.
   Required: Drop the _manifest entry and its KnownFieldsMetaKey member -- the rebase in B1 removes both. If a manifest field list is genuinely wanted later, land it separately with estimatedMinutes and stats included and a verifyFields(ManifestJsonObjectSchema, '_manifest') case in type-coupling.test.ts so it cannot drift.

4. [blocking] **B4 — Removes the only record of why completionSuppressed is checked twice, at the one site no test covers** (low · correctness-and-reliability)
   src/interactive-engine/interactive.hook.ts loses the rationale before the outcomeFrom callback explaining that completionSuppressed is checked there, not only after the span closes, so a suppressed handler's Faro span is not stamped 'ok' before the suppression is known -- it is set synchronously inside the awaited handler, before that line runs. This is a temporal precondition the type system cannot express, and a whole-tree grep finds it stated nowhere else at HEAD; the comment that survives three lines below documents the sequenceResult classification, not the suppression ordering. This is not covered by B2: the deleted test asserted only outcome === 'error', which is the second site. The outcomeFrom suppression branch had no test at the base and has none at HEAD, so restoring B2's coverage still leaves it unguarded. A future editor pruning the apparently-duplicate check at the outcomeFrom site would pass every test and silently mis-stamp Faro spans 'ok' for suppressed steps.
   Required: Restore the outcomeFrom rationale comment in interactive.hook.ts, or add an assertion that options.outcomeFrom() returns 'error' when completionSuppressed is set, alongside the existing unsuppressed case at interactive.hook.test.ts:298.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1633
Summary: forward interactive step props
Verdict: Request Changes
Results: 4 blockers, 0 non-blocking findings, 0 follow-ups


---

Next step: rebase on current `main`, taking main's side for `src/types/json-guide.schema.ts` and `src/interactive-engine/interactive.hook.test.ts`, then re-apply your intended changes on top — the prop forwarding, the formfill observer, the `openGuide` threading and the `formHint` description are all good and should survive untouched. Push and I'll re-review; happy to take the B2 test re-homing myself in a follow-up if you'd rather not carry it.

moxious (changes requested):
branch has picked up some minor merge conflicts; please address and I'll try to take this up Tuesday for review/merge

mikemikimike:
Rebased the branch onto the current `main` and re-signed the updated commit with the repository-associated SSH signing key.

The review fixes are preserved, including the schema contract restoration, removal of the stale `_manifest` allowlist entry, the relocated full-screen handoff assertions, and the `completionSuppressed` timing rationale.

Verification:
- GitHub reports commit `700c3d9e789598be33114e2e9de49dd81ced3120` as verified.
- Focused Jest tests: 95 passed.
- TypeScript typecheck and Prettier check passed.
- The local full Jest run reached the known Windows path-separator baseline failures in `src/validation/import-graph.test.ts`; the PR CI checks are running on the new head.

Please re-review the updated head.

moxious:
@mikemikimike please make sure all of your commits are signed. I appreciate you sticking with this PR, but adding new commits is creating new blockers, which prevent us from re-reviewing.

mikemikimike:
Rewrote `fix/forward-interactive-step-props` onto the current `main` and SSH-signed every commit in the branch.

The new head is `a0c29a951fa829036d483effddb233e424ced2ef`. GitHub reports all three PR commits as verified, and the `signed-commits`, Semgrep, Cursor Bugbot, and Socket Security checks are passing. Please re-review.

src/interactive-engine/interactive.hook.fullscreen-handoff-timing.test.ts
```diff
@@ -170,4 +176,26 @@ describe('executeInteractiveAction composed with the real requestSidebarHandoffA
...
+    await act(async () => {
+      outcome = await result.current.executeInteractiveAction({
+        targetAction: 'button',
+        refTarget: 'test-target',
+        buttonType: 'do',
+      });
+    });
+
+    expect(outcome).toBe('error');
+  });
```
moxious, line 200: B5, blocking. The end-to-end assertion for this PR's bug is missing here.

At `8b3d748d` this file asserted the executor hop:

```ts
expect(navigateHandler.execute).toHaveBeenCalledWith(
  expect.objectContaining({ openGuide: 'bundled:destination-guide' }),
);
```

It is gone at `a0c29a95`, lost in the `8b3d748d` -> `9551c2d0` rewrite — a novel loss, not a main's-side revert. It was the only assertion covering `interactive.hook.ts` -> `navigateHandler.execute`, which is the exact hop #1506 is about.

`openGuide` appears in seven test files at this head, but each tests a different layer: `navigate-handler.test.ts` exercises `resolveGuideParam` directly, `step-type-registry.test.ts` the registry extension, `cross-tab.types.test.ts` the wire validation, `interactive-step.test.tsx` the component. None asserts that the executor puts `openGuide` on `InteractiveElementData`.

Restore an assertion that `openGuide` reaches the navigate handler through `executeInteractiveAction`. The re-homed timing test is a fine host if you would rather not reopen this file.

Invariant: the renderer-to-executor hop for every prop this PR forwards must be asserted at the executor boundary, not only at the layers either side of it.

src/components/content-renderer/content-renderer.prop-forwarding.test.ts
```diff
@@ -23,7 +23,7 @@ import type { JsonGuide } from '../../types/json-guide.types';
  * JSON guide gets silence. Documented here rather than fixed so this tripwire
  * can land without changing runtime behaviour; shrinking this list is the fix.
  */
-const KNOWN_UNFORWARDED = ['formHint', 'openGuide', 'validateInput'];
+const KNOWN_UNFORWARDED: string[] = [];
```
moxious, line 26: B6, blocking. Emptying this list is correct and required — the suite asserts exact equality, so forwarding the three props without emptying it fails. But it removes the thing that was accidentally protecting line 62.

Line 62 is `source.slice(start, source.indexOf('</InteractiveStep>', start))`. `start` is guarded on line 61; the end index is not. If `</InteractiveStep>` stops following the arm, `indexOf` returns `-1` and `slice(start, -1)` scrapes to the end of `content-renderer.tsx`, so `forwardedProps()` returns a superset of every `element.props.X` in the file.

That unguarded line is byte-identical on `main` and at the merge base, so it is not yours. What changes is the consequence: with three names in this list a widened block fails loudly; with `[]` it passes vacuously in all three cases.

You had it fixed at `8b3d748d`, and I cleared it by name on 2026-09-01:

```ts
const end = source.indexOf('</InteractiveStep>', start);
expect(end).toBeGreaterThan(start);
const block = source.slice(start, end);
```

The rebase took main's side for this file — the third file, which my round-6 `Required` line failed to list. Please restore those two lines, or let me push them.

Invariant: `forwardedProps()` must never widen its scanned block on a failed lookup — every `indexOf` it performs is asserted to have found its delimiter before the result reaches `slice`.

src/types/json-guide.schema.ts
```diff
@@ -376,7 +376,7 @@ export const JsonInteractiveBlockSchema = z
     objectives: z.array(ObjectiveTokenSchema).optional().describe(objectivesDescription('block')),
     skippable: z.boolean().optional().describe('Allow user to skip this block'),
     hint: z.string().optional().describe('Hint text shown if user is stuck'),
-    formHint: z.string().optional().describe('Placeholder text for formfill input fields'),
+    formHint: z.string().optional().describe('Validation-failure message for formfill input fields'),
```
moxious, line 379: B7, blocking, and this one is ours. Anchored here because the line it concerns — `openGuide` at `json-guide.schema.ts:389` — is not part of this diff.

This still reads `'Guide ID to open when this block completes'`, but `openGuide` fires during the `navigate` action and is ignored on the `window.open` branch. The string feeds the CLI and the MCP authoring server, so authors are told the wrong trigger.

I pushed `08d5f48f` ("docs: correct the openGuide schema description") onto this branch on 2026-08-25. Force-push #2 on 08-29 rolled the merge base back four days and destroyed it. You re-authored the correction at `8b3d748d` and force-push #9 destroyed it again. Your PR description's claim that it was corrected was true when you wrote it.

Please restore it here and in `docs/design/AGENT-AUTHORING.md` — `src/validation/agent-authoring-docs.test.ts` fails if the doc's quote does not track this schema. Your `8b3d748d` wording was good: "Guide ID to open during an internal navigate action (ignored for external http(s) targets)".

moxious (changes requested):
Thanks for staying with this. Before the list: most of this PR's length is our process, not your code, and the three items below are all fixes that existed and were destroyed — two of them by rewrites we told you to perform, one of them by an instruction of mine that named the wrong files. I have said so explicitly in the review and recorded B1 through B4 as closed so they cannot be reopened.

---

Round 7. **B1–B4 are all closed at `a0c29a95`, and I am recording them as closed so no later round reopens them.** I have re-verified each against the merge base rather than trusting the vanished anchors: the schema diff is now only the two `formHint` describes (B1), the three handoff invariants are re-homed into `interactive.hook.fullscreen-handoff-timing.test.ts` with `CONCERN_DETAILS.md` re-pointed (B2), `_manifest` is absent (B3), and the `outcomeFrom` rationale is back at `interactive.hook.ts:520-523` (B4). Signing, CLA and 39/40 checks are green; the one `publish-report` failure is our fork-permission comment write, not yours.

Four items are outstanding. **Three of them are fixes that existed and were destroyed by the rewrites we asked you to perform, not work you failed to do** — I go through that below, because the record should say so.

## What we got wrong

Before the list: on 2026-08-27 you asked whether we wanted a fork-side history rewrite or a maintainer-side signing pass. We said rewrite was your only option. That answer is the direct cause of most of this PR's length — every signing round became a force-push, and every force-push orphaned every review anchor. **All 13 heads that were ever reviewed are orphaned from the current head**, so six approvals and a whole `cleared` list are anchored to commits nothing can reach. That is our process failing you, and my 2026-09-11 note framing it as "adding new commits is creating new blockers" had it backwards.

My round-6 `Required` line also under-specified itself. It said take main's side for `json-guide.schema.ts` and `interactive.hook.test.ts`. The rebase reasonably took main's side for a third file too, and that is exactly where item 2 below died. You did what I asked; I asked imprecisely.

---

## Merge contract

Fix these three and I merge. No further rebase is needed or wanted — the branch is `MERGEABLE` and every rebase so far has cost a round and eaten a fix.

1. [blocking] **The end-to-end assertion for this PR's actual bug was lost and is not covered anywhere else** (medium · testing-and-verification)

   At `8b3d748d` this existed in `src/interactive-engine/interactive.hook.test.ts`:

   ```ts
   openGuide: 'bundled:destination-guide',
   // …
   expect(navigateHandler.execute).toHaveBeenCalledWith(
     expect.objectContaining({ openGuide: 'bundled:destination-guide' }),
   );
   ```

   It is absent at `a0c29a95`, and it is the only place that ever asserted the `interactive.hook.ts` → `navigateHandler.execute` hop. `openGuide` still appears in seven test files at this head, but every one of them tests a different layer: `navigate-handler.test.ts` exercises `resolveGuideParam` directly, `step-type-registry.test.ts` the registry extension, `cross-tab.types.test.ts` the wire validation, `interactive-step.test.tsx` the component. None asserts that the executor actually puts `openGuide` on `InteractiveElementData`.

   That hop is the bug. #1506 is precisely "the renderer emits it and the executor never receives it", so with this assertion gone the PR closes a bug whose fix has no integration test. It was lost in the `8b3d748d` → `9551c2d0` rewrite — a novel loss, not a main's-side revert.

   **Required:** restore an assertion that `openGuide` reaches the navigate handler through `executeInteractiveAction`. Anywhere is fine — the re-homed timing test is a reasonable host if you would rather not reopen `interactive.hook.test.ts`. Invariant: **the renderer-to-executor hop for every prop this PR forwards must be asserted at the executor boundary, not only at the layers either side of it.**

2. [blocking] **The `forwardedProps()` end-index guard was cleared in round 6 and lost in the rebase; with `KNOWN_UNFORWARDED` empty it now fails silently instead of loudly** (medium · docs-retrieval-and-rendering)

   `content-renderer.prop-forwarding.test.ts:62` at this head is unguarded:

   ```ts
   const block = source.slice(start, source.indexOf('</InteractiveStep>', start));
   ```

   If `</InteractiveStep>` ever stops following the arm, `indexOf` returns `-1`, `slice(start, -1)` scrapes to the end of `content-renderer.tsx`, and `forwardedProps()` returns a superset of every `element.props.X` in the file.

   To be clear about what is and isn't yours here: **that unguarded line is byte-identical on `main` and at the merge base** — you did not introduce it. What changes is the consequence. On `main`, `KNOWN_UNFORWARDED` holds three names, so a widened block contradicts a three-name expectation and the suite fails loudly. Emptying the list is *required* by this fix — the suite asserts exact equality, so forwarding the three props without emptying it fails — and with it empty the same widening makes all three cases pass vacuously. So the fix correctly removes the thing that was accidentally protecting the helper.

   You had this fixed. At `8b3d748d` it read:

   ```ts
   const end = source.indexOf('</InteractiveStep>', start);
   expect(end).toBeGreaterThan(start);
   const block = source.slice(start, end);
   ```

   I cleared it by name on 2026-09-01, and the rebase took main's side for that file — the third file, the one my `Required` line did not list.

   **Required:** restore those two lines. Invariant: **`forwardedProps()` must never widen its scanned block on a failed lookup — every `indexOf` it performs is asserted to have found its delimiter before the result reaches `slice`.**

3. [blocking] **The `openGuide` schema description fix is missing, and it was my commit that got destroyed** (low · guide-schema-and-contracts)

   `json-guide.schema.ts` still describes `openGuide` as `'Guide ID to open when this block completes'`, while it actually fires during the `navigate` action and is ignored on the `window.open` branch. That string feeds the CLI and the MCP authoring server, so authors get told the wrong trigger.

   The history: I pushed `08d5f48f` ("docs: correct the openGuide schema description") onto this branch on 2026-08-25. Force-push #2 on 08-29 rolled the merge base backwards four days and destroyed it. You then re-authored the correction yourself at `8b3d748d`, and force-push #9 destroyed it again. Your PR description's claim that it was corrected was true when you wrote it. This one is entirely ours.

   **Required:** restore the correction in both `src/types/json-guide.schema.ts` and `docs/design/AGENT-AUTHORING.md` — `src/validation/agent-authoring-docs.test.ts` will fail if the doc's quote does not track the schema. Wording along the lines of "Guide ID to open during an internal navigate action (ignored for external http(s) targets)" is what we want; your `8b3d748d` phrasing was good.

---

## Non-blocking

- `content-renderer.prop-forwarding.test.ts:20-26`'s docblock still reads "These are pre-existing gaps… Documented here rather than fixed… shrinking this list is the fix" above a now-empty list, and the third case (`'keeps the documented-gap list honest'`) degenerates to a `forEach` over `[]`. Worth a sentence saying the list is empty and must stay empty. Not a gate.
- `step-type-registry.ts` adds `fullScreenFallbackLocation` to the multistep, guided and code-block schemas. Those three paths never reach `executeInteractiveAction`, so the field is never read — additive and harmless, and it fixes nothing that #1506 is about. Leave it or drop it, your call.

## What I checked and found clean

Recording this so no later round re-derives it.

- **Every forwarded prop is read at its destination.** `formHint` and `validateInput` reach `useFormElementValidation` and render as the failure banner at `interactive-step.tsx:1286-1292`. `openGuide` runs `content-renderer.tsx:1164` → `interactive-step.tsx` → `InteractiveActionRequest` → `interactive.hook.ts:411,422` → `navigate-handler.ts:92`. The section hop is real: `interactive-section.tsx:633` spreads `...stepInfo`, and `openGuide` is on `StepInfo` via `toStepInfoExtension`. Nothing is threaded one layer deeper and dropped again — which is what item 1 above should be asserting.
- **`formHint` is not a behavioural redefinition.** `useFormValidation.ts:204` and `guided-handler.ts:976` already rendered it as the validation-failure message, and the block editor already labels it "Validation hint". The one bundled guide that sets it (`loki-grafana-101/content.json:201`) is written as a failure message and has no `validateInput`. The schema string was the drifted half.
- **The schema change is describe-only.** No shape, enum, optionality or default. Published guides are unaffected.
- **Cross-tab does not widen.** `openGuide` is validated as an optional string on the action and every sub-action (`cross-tab.types.ts:266,285`) with a rejection test, and resolves through `findDocPage` against the bundled index and the `isGrafanaDocsUrl` / `isInteractiveLearningUrl` allowlists. The same guide-open was already reachable via `refTarget`'s `?doc=` fallback.
- **The formfill observer gate narrows nothing.** `formTargetElement` has one consumer whose `enabled` predicate already carried the same four conditions at base and head.
- **`data-attributes.contract.test.tsx`'s removed line was an `it.todo`** inside a `describe.skip`, and the two new RTL cases assert the transition it described. Retiring it is correct.
- **Tier discipline holds.** Two added imports, tier 4 → tier 0 and tier 2 → tier 1. Both downward.
- **`Closes #1506`** is present and correct.

## Diff-base note for anyone re-reviewing

The merge base is `b4d2f505`, not the current `main` tip. A two-dot `main`..head diff shows 85 files and ~3,200 deletions — all phantom, from #1848/#1864 moving main forward. Use `git diff b4d2f505..a0c29a95`. The real change is 20 files, +278/-140.

## Ordering against #1867

#1867 (mine, "forward objectives and hints into useStepChecker") also touches `interactive-step.tsx` and `interactive-step.test.tsx`. `interactive-step.tsx` has no textual conflict — disjoint hunks. `interactive-step.test.tsx` will conflict trivially: both branches insert a new top-level `describe` at the same anchor after `InteractiveStep: showMeText label override`. Resolution is "keep both". They are complementary: your new observer gate tests `isCompletedWithObjectives`, which is dead today because `objectives` never reaches `useStepChecker` — exactly what #1867 fixes. **#1633 lands first.** I will absorb the conflict on my side.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1633
Summary: forward interactive step props
Verdict: Request Changes
Results: 3 blockers, 2 non-blocking findings, 0 follow-ups

---

Next step, and I would rather you took the second option: either restore the three items above in one commit on top of `a0c29a95` — no rebase, please — or say the word and **I will push all three to this branch myself**, since two of them are fixes we destroyed and the third is a guard my own instruction told you to drop. Either way I merge behind them, and B1–B4 stay closed.

moxious (changes requested):
We can't review this PR yet: the head commit `2d9b7b80a860` ("Merge upstream main into fix/forward-interactive-step-props") is unsigned. This repo requires every commit to be signed. Please sign it (re-sign via rebase, or add a signed follow-up) and push — we'll pick the review back up once the signed-commits check is green.

moxious:
Thanks for sticking with this one through the signing and rebase saga — the substantive review is done on my side. I re-verified B5, B6, and B7 from my last review directly against the current head and all three are fixed as described, and the objectives-suite regression flagged today is already fixed as well.

One thing is still failing CI: `docs/design/CONCERN_DETAILS.md` is missing its `tenant-settings` row in the "Concern review table" section (it's still present further down in the per-concern detail section, just not in the summary table). That's what's failing `npm run test:review-contract` and the required `CI Gate` check — looks like a casualty of one of the rebases, same root cause as B1–B4 from the earlier round. Could you add that row back? The content is unchanged on `main`, it just needs to be present here too.

Once `CI Gate` is green I'll approve and we can merge. I'll file the `KNOWN_UNFORWARDED` docblock cleanup and the unused `fullScreenFallbackLocation` field as separate follow-up issues rather than holding this up further.

I would just amend the commit and get a teammate to merge this but I probably don't have push perms to your fork

mikemikimike:
Restored the missing tenant-settings row in docs/design/CONCERN_DETAILS.md in commit b83b6b823424d7665686d70949544e4c4f747a7d. npm run test:review-contract passes all 85 tests. The existing 8 PR commits remain GitHub-verified, and the signed-commits check is green. Please re-review the current signed head.

moxious:
@mikemikimike I re-ran CI and it's still blocked on some failing tests, please have a look.

mikemikimike:
Pushed signed commit d4faecb2 to correct the remaining CLI documentation entry for --open-guide. It now matches the schema: internal navigate action only, ignored for external http(s) targets. Local verification: review contract 85/85, focused Jest 7/7, Prettier, and git diff --check pass. Please re-review the current head.

PR #1634, opened 2026-08-17
refactor(tour): extract a reusable BubbleTour from the block-editor tour

Jayclifford345:
## Summary

Lifts the block editor's tour into a reusable `src/components/BubbleTour/`, leaving `BlockEditorTour` as a thin wrapper that supplies its own steps. Along the way it fixes three defects that exist on `main` today.

`BlockEditorTour` was already headless — it `return null`ed and owned only step data, an index, and a keydown listener, while all the bubble UI came from `NavigationManager.highlightWithComment`. This makes that reusability real rather than incidental.

## Why

A second caller wants the same next/previous bubble. Rather than copy the controller, this extracts it — and extracting forced the three bugs below into the open, each of which is reachable on `main` right now.

## What to look at

**`src/interactive-engine/navigation-manager.ts`** is the only shared-engine change, and it has three parts:

1. `'Start creating'` was block-editor copy hardcoded in the engine (`:980`). It becomes `options.nextLabel`, defaulting to `Done`. `BlockEditorTour.test.tsx` pins the editor's own label so its visible copy can't regress.
2. `createCommentBox` now accepts null rects and returns a viewport-centered box **with its footer intact**, exposed as `showCenteredComment`. Previously a missing target fell through to `showNoopComment`, which takes only a string — so the user lost Back/Next and only the keyboard could move them on.
3. In-panel overlays get `data-pathfinder-internal`, derived from `isPathfinderContent(highlightTarget)`.

`showNoopComment` is deliberately **not** touched — its box has different chrome and three contract tests pin its `data-noop` marker.

**The signature stays positional.** `guided-handler.test.ts:145-155` asserts all nine arguments; options are matched with `objectContaining`, so adding a field is safe where an options-object refactor would not be.

**`useHighlightDodge.ts`** — one selector line, and the actual bug fix. See below.

## Bugs fixed

**The floating panel fled its own highlight.** `useHighlightDodge` moves the panel away from `.interactive-highlight-outline` / `.interactive-comment-box`, and the block editor renders inside `FloatingPanel` (`FloatingPanelManager.tsx:376`), whose root carries `data-pathfinder-content="true"`. Pop the editor out, take the tour, and the panel runs from the thing it's pointing at. Both the outline and the bubble are exempted — exempting only the outline lets the panel dodge the bubble, which moves the target, which re-anchors the bubble, which oscillates.

**The keydown listener hijacked typing.** It listened on `document` with no target check and treated `Enter` as "next", so `Enter` or an arrow key inside any editor text field advanced the tour. Now scoped to ignore `input` / `textarea` / `contenteditable`, `Enter` dropped, and `preventDefault` called — that last part matters because `FloatingPanel.tsx:103-110` minimizes on Escape unless the event is already `defaultPrevented`, so one Escape would otherwise both close the tour *and* minimize the panel.

**A missing target stranded the user.** Covered above.

## Test coverage

The tour branch of `createCommentBox` had **no** coverage — no Back/Next, no step badge, no dots, no last-step label. `comment-box.tour-chrome.test.ts` covers it, including a regression guard that guided mode still renders Cancel/Skip and *not* Back/Next, which protects the six other `highlightWithComment` callers (`guided-handler`, `hover-handler`, `button-handler`, `form-fill-handler`, `focus-handler`, `action-replay`).

## How to verify

```bash
npm run check
```

Then in a local Grafana, open the block editor → kebab → **Take tour**:

- Walk it with both mouse and arrow keys. The last button still reads **Start creating**.
- Type in a block's text field mid-tour — `Enter` and arrows no longer advance it.
- Press Escape once — the tour closes and the panel does **not** also minimize.
- Pop the editor out and repeat. The panel must not flee its own highlight.

## Breaking changes

None. `BlockEditorTour`'s props narrow from `{ onClose, steps? }` to `{ onClose }` — the `steps?` override was never passed by any caller, and `BlockEditorModals.tsx:85` is unchanged.

## Reversibility

Fully reversible. The engine change is additive on the options bag; the dodge change is one selector. Reverting restores the previous behaviour including the three bugs.

### DIFF
--- a/src/components/BubbleTour/BubbleTour.tsx
+++ b/src/components/BubbleTour/BubbleTour.tsx
@@ -87,17 +87,26 @@ export function BubbleTour({ steps, onClose, finalStepLabel }: BubbleTourProps) 
         }
 
         if (resolved) {
-          return navigationManager.highlightWithComment(
-            resolved.element,
-            step.content,
-            false,
-            stepInfo,
-            undefined,
-            close,
-            goToNext,
-            onPrevious,
-            options
-          );
+          // highlightWithComment awaits navigation/scroll before it paints. If the tour is
+          // torn down mid-await, `cancelled` is already true by the time it resolves — clear
+          // what it just painted so it doesn't outlive its own component and key listener.
+          return navigationManager
+            .highlightWithComment(
+              resolved.element,
+              step.content,
+              false,
+              stepInfo,
+              undefined,
+              close,
+              goToNext,
+              onPrevious,
+              options
+            )
+            .then(() => {
+              if (cancelled) {
+                navigationManager.clearAllHighlights();
+              }
+            });
         }
 
         navigationManager.showCenteredComment(
@@ -149,8 +158,11 @@ export function BubbleTour({ steps, onClose, finalStepLabel }: BubbleTourProps) 
       }
     };
 
-    document.addEventListener('keydown', handleKeyDown);
-    return () => document.removeEventListener('keydown', handleKeyDown);
+    // Capture phase: FloatingPanel's own Escape listener is bubble-phase on document and
+    // registered first, so on bubble order alone it would see defaultPrevented as false and
+    // minimize before this handler runs. Capture always runs first regardless of mount order.
+    document.addEventListener('keydown', handleKeyDown, { capture: true });
+    return () => document.removeEventListener('keydown', handleKeyDown, { capture: true });
   }, [currentStep, close, goToNext, goToPrevious]);
 
   return null;
--- a/src/components/BubbleTour/BubbleTour.tsx
+++ b/src/components/BubbleTour/BubbleTour.tsx
@@ -1,7 +1,8 @@
-import { useCallback, useEffect, useMemo, useState } from 'react';
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 
 import { NavigationManager } from '../../interactive-engine';
 import { resolveWithRetry } from '../../lib/dom';
+import { logger } from '../../lib/logging';
 import { safeEventHandler } from '../../utils/safe-event-handler.util';
 
 const MISSING_TARGET_NOTE = "This part of the interface isn't on screen right now.";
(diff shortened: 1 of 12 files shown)


src/components/BubbleTour/BubbleTour.tsx
```diff
@@ -0,0 +1,157 @@
...
+      if (e.key === 'ArrowRight') {
+        safeEventHandler(e, { preventDefault: true });
+        goToNext();
+      } else if (e.key === 'ArrowLeft' && currentStep > 0) {
+        safeEventHandler(e, { preventDefault: true });
+        goToPrevious();
+      }
+    };
+
+    document.addEventListener('keydown', handleKeyDown);
```
moxious: **Escape still minimizes the popped-out editor**

You register the tour's bubble-phase `document` listener after the already-mounted `FloatingPanel` listener. When a user presses Escape from an editor field, the panel sees `defaultPrevented === false` and minimizes before your handler closes the tour, so the one-key close still changes panel state.

Register the tour handler in capture phase or centralize Escape arbitration, and add an integration test that mounts the panel before the tour.

src/components/BubbleTour/BubbleTour.tsx
```diff
@@ -0,0 +1,157 @@
...
+    };
+
+    void resolveWithRetry(step.target, 'highlight').then((resolved) => {
+      const paint = () => {
+        if (cancelled) {
+          return;
+        }
+
+        if (resolved) {
+          return navigationManager.highlightWithComment(
```
moxious: **An in-flight paint can survive tour teardown**

Your `cancelled` check runs only before `highlightWithComment`, but that method awaits navigation and smooth scrolling before it appends the overlay. If the user closes the tour during that await, unmount cleanup clears the current DOM and the in-flight call then appends a persistent outline and comment box after their owner and key listener are gone.

Await the paint and clear or abort it when cancellation is observed after it settles, then test unmount after the manager call has started.

moxious (changes requested):
Thanks for extracting the tour controller and pinning the shared comment-box chrome. I found two lifecycle defects that should be fixed before merge.

moxious (approved):



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1634
Summary: extract a reusable BubbleTour from the block-editor tour
Verdict: Approve

PR #1638, opened 2026-08-17
fix(learning-paths): thread packageInfo through Discover More launches

### LINKED ISSUE #1637: Discover More launches never resolve package/milestone context
## Summary

Starting a course from My Learning's **Discover More** section never resolves the course's package/manifest context — no milestone toolbar ("Milestone 1 of 10"), no progress chrome, no next/prev navigation. It renders as a bare, standalone document instead.

## Repro

1. Go to My Learning.
2. Under "Discover More", click "Start" on any package-backed course (confirmed with `alerting-query-language-lj`, `visualization-metrics-lj`, and `infinity-json-lj` — not course-specific).
3. Expected: the course opens with its milestone/journey chrome — progress indicator, milestone counter, next/prev navigation — same as launching the same course from a My Learning card.
4. Actual: renders as plain flowing text with none of that chrome, as if it were a single standalone document. Network trace confirms the milestone `manifest.json` requests never happen at all.

## Isolation

Confirmed via `git stash` isolation directly on `main` (not the in-progress `feat/course-cover-page-foundation` branch) — this reproduces independent of that branch's work.

All `interactive-learning.grafana.net` network requests that do fire succeed (200) — this isn't a fetch failure, the milestone requests just never get made.

## Cause

`MyLearningTab.tsx`'s `handleDiscoverStart` calls `launch(item.contentUrl, item.title, item.id)` with no `packageInfo` argument, unlike the My Learning card path (`openPathCover`), which explicitly resolves and passes `packageInfo` before launching. Without it, `fetchPackageContent` never learns the content is a multi-milestone package at all, so milestone resolution never runs.

## Note for #1621

The in-progress course-landing-page work (#1621) adds a real cover page for package-backed paths on a fresh launch — package paths don't get one from any launch source on `main` today. Once #1621 lands, this same missing `packageInfo` would also prevent Discover More from reaching that cover page, but that's a downstream consequence, not the bug itself — this issue is scoped to what's observable on `main` right now.

## Files to look at

- `src/components/LearningPaths/MyLearningTab.tsx` — `handleDiscoverStart` vs `openPathCover`
- `src/components/docs-panel/utils/prepare-guide-launch.ts` — `packageInfo` auto-derivation
- `src/docs-retrieval/package-info-from-url.ts` — `fetchPackageInfoFromUrl`

Spidy88:
## Summary

Fixes #1637 — Discover More launches never resolve the course's package/manifest context (no milestone toolbar, no progress chrome, no next/prev navigation), rendering as a bare, standalone document instead.

**The actual root cause**: `setPackageResolver()` was only ever called inside `CombinedLearningJourneyPanel`'s constructor, so a launch from My Learning — before any docs-panel/sidebar/full-screen instance has mounted this session — races ahead of the resolver ever being configured. `module.tsx`'s `plugin.init` now wires it at plugin load instead.

(An earlier version of this PR also attributed part of the fix to threading `packageInfo` through Discover More's launch call. That turned out not to be a correctness fix — `prepareGuideLaunch` already backfills `packageInfo` from the URL when the caller passes none, and every live Discover More `contentUrl` already matches the pattern that triggers that backfill. The threading is a worthwhile latency optimization — it saves that redundant sibling-manifest re-fetch — but the resolver wiring above is what actually fixes #1637.)

## What changed

- `src/docs-retrieval/content-fetcher/package-resolver-registry.ts` (new) + `src/module.tsx` — the package resolver singleton, split into a dependency-free leaf module so `plugin.init` can register it without statically pulling `docs-retrieval`'s heavy `content-fetcher` chain into the entry bundle (see "Bundle size" below). `module.tsx` registers a factory that dynamically imports `composite-resolver.ts` directly; `package-content.ts`'s four resolver read sites (all already `async`) await it.
- `src/learning-paths/useDiscoverMore.ts` — `DiscoverMoreItem` now carries the package's inlined `manifest` (already fetched for milestone-count display), parsed through `ManifestJsonObjectSchema.loose().safeParse` (matching `online-cdn-resolver.ts`'s handling of the same shape) before being threaded through.
- `src/components/LearningPaths/MyLearningTab.tsx` — `handleDiscoverStart` builds `packageInfo` from that manifest before calling `launch`, mirroring the existing pattern used above it for My Courses / App Platform guides — a latency win, not the fix (see above).
- `src/components/docs-panel/docs-panel.tsx` — `CombinedLearningJourneyPanel`'s constructor now seeds the resolver from the published `window.__pathfinderPluginConfig` global rather than whichever `pluginConfig` that surface happened to be constructed with. The resolver is one app-wide singleton; sidebar previously built its `pluginConfig` from Grafana's `usePluginContext()` snapshot, which can lag a settings save independently of the published global, so every sidebar (re)mount could silently re-stale it.

## Bundle size

Wiring the resolver via `module.tsx` initially imported the `docs-retrieval` and `package-engine` barrels statically, growing `module.js` 6.2x (115 KB → 696 KB raw — CI's bundle-size bot flagged `+518.96%`) since `docs-retrieval` statically imports the whole `content-fetcher` orchestrator (zod, dompurify, the bundled guide index) and `package-engine`'s composite resolver pulls in every resolver implementation. Total bundle size was unchanged, so this was pure reclassification from lazy to eager, not new functionality.

Fixed by deferring construction behind a factory + dynamic import (see "What changed" above); `module.js` is back to ~116 KB. Guarded against recurrence: `entry-bundle-boundary.test.ts` now denies `module.tsx` importing either barrel, mirroring the existing hooks-barrel check; both new intentional bypasses (the registry leaf module and the dynamic `composite-resolver` import) are recorded in `architecture.test.ts`'s `ALLOWED_BARREL_VIOLATIONS`.

## How to verify

- Unit: `useDiscoverMore.test.ts` and `MyLearningTab.test.tsx` cover the manifest threading; `package-content.test.ts` covers `setPackageResolverFactory` (resolves once available, invoked once across repeated reads, a later `setPackageResolver` overrides a pending factory); `docs-panel.alignment.test.ts` covers the resolver's config-source fix (global wins over constructor `pluginConfig`; falls back when the global is unset).
- `npm run build` — confirm `dist/module.js` is ~116 KB raw with no entrypoint-size warning.
- Live, local dev stack: cleared all Pathfinder localStorage state (fresh session, no prior docs-panel mount), clicked "Start" on a Discover More card as the very first action. Confirmed via network trace that all milestone `manifest.json` requests now resolve — previously none did.
- `npm run check`'s non-Go steps pass clean (7123 tests); architecture/tier/entry-bundle tests pass with the two new bypasses recorded, no unrecorded violations.

## Breaking changes

None. Additive only — `DiscoverMoreItem.manifest` is a new optional field (now typed `ManifestJson`, schema-validated), and the resolver registry is a drop-in replacement for the previous module-level variable with the same public API plus one new factory-registration function.

## Note

Package-backed paths don't render a cover page from any launch source on `main` today — that's the separate, in-progress #1621. This PR is scoped to the underlying missing-context bug, which is independent of covers; #1621 will build on top of this fix once both land.

### DIFF
--- a/src/components/docs-panel/docs-panel.tsx
+++ b/src/components/docs-panel/docs-panel.tsx
@@ -234,7 +234,7 @@ class CombinedLearningJourneyPanel extends SceneObjectBase<CombinedPanelState> i
 
     // Wire the composite PackageResolver into docs-retrieval so that
     // fetchPackageContent() and fetchPackageById() can resolve bundled and
-    // remote packages. This is the Tier 3/4 injection point described in Phase 4g.
+    // remote packages — the Tier 3/4 injection point.
     //
     // The resolver is one app-wide singleton, not one per surface, so it must
     // always be seeded from the one authoritative config source — the global
--- a/src/components/docs-panel/docs-panel.tsx
+++ b/src/components/docs-panel/docs-panel.tsx
@@ -236,15 +236,7 @@ class CombinedLearningJourneyPanel extends SceneObjectBase<CombinedPanelState> i
     // fetchPackageContent() and fetchPackageById() can resolve bundled and
     // remote packages — the Tier 3/4 injection point.
     //
-    // The resolver is one app-wide singleton, not one per surface, so it must
-    // always be seeded from the one authoritative config source — the global
-    // usePathfinderPluginConfig publishes — rather than whichever pluginConfig
-    // this particular surface happened to be constructed with. Sidebar builds
-    // its pluginConfig from Grafana's usePluginContext() snapshot, which can
-    // lag a settings save independently of the published global; using it here
-    // would silently re-stale the resolver on every sidebar (re)mount, undoing
-    // module.tsx's post-refresh re-wire. Falls back to pluginConfig when the
-    // global isn't set (e.g. unit tests constructing this class directly).
+    // Seed from the published global, not this surface's snapshot — the resolver is one app-wide singleton.
     const resolverConfig = (window as unknown as { __pathfinderPluginConfig?: DocsPluginConfig })
       .__pathfinderPluginConfig;
     setPackageResolver(createCompositeResolver(resolverConfig ?? pluginConfig));
--- a/src/components/docs-panel/docs-panel.tsx
+++ b/src/components/docs-panel/docs-panel.tsx
@@ -235,7 +235,19 @@ class CombinedLearningJourneyPanel extends SceneObjectBase<CombinedPanelState> i
     // Wire the composite PackageResolver into docs-retrieval so that
     // fetchPackageContent() and fetchPackageById() can resolve bundled and
     // remote packages. This is the Tier 3/4 injection point described in Phase 4g.
-    setPackageResolver(createCompositeResolver(pluginConfig));
+    //
+    // The resolver is one app-wide singleton, not one per surface, so it must
+    // always be seeded from the one authoritative config source — the global
+    // usePathfinderPluginConfig publishes — rather than whichever pluginConfig
+    // this particular surface happened to be constructed with. Sidebar builds
+    // its pluginConfig from Grafana's usePluginContext() snapshot, which can
+    // lag a settings save independently of the published global; using it here
+    // would silently re-stale the resolver on every sidebar (re)mount, undoing
+    // module.tsx's post-refresh re-wire. Falls back to pluginConfig when the
+    // global isn't set (e.g. unit tests constructing this class directly).
+    const resolverConfig = (window as unknown as { __pathfinderPluginConfig?: DocsPluginConfig })
+      .__pathfinderPluginConfig;
+    setPackageResolver(createCompositeResolver(resolverConfig ?? pluginConfig));
 
     // Note: Tab restoration now happens from React component after storage is initialized
     // to avoid race condition with useUserStorage hook
--- a/src/components/LearningPaths/MyLearningTab.test.tsx
+++ b/src/components/LearningPaths/MyLearningTab.test.tsx
@@ -498,9 +498,9 @@ describe('MyLearningTab launch flow', () => {
     );
   });
 
-  // Regression for #1637: without packageInfo, the loader falls through to
-  // plain fetchContent and the item renders as a bare document — no
-  // milestone toolbar, progress chrome, or next/prev nav.
+  // Without packageInfo, the loader falls through to plain fetchContent and
+  // the item renders as a bare document — no milestone toolbar, progress
+  // chrome, or next/prev nav.
   it('threads the inlined manifest through as packageInfo so the item resolves its milestone context', async () => {
     mockDiscoverItems = [
       {
--- a/src/components/LearningPaths/MyLearningTab.test.tsx
+++ b/src/components/LearningPaths/MyLearningTab.test.tsx
@@ -498,9 +498,6 @@ describe('MyLearningTab launch flow', () => {
     );
   });
 
-  // Without packageInfo, the loader falls through to plain fetchContent and
-  // the item renders as a bare document — no milestone toolbar, progress
-  // chrome, or next/prev nav.
   it('threads the inlined manifest through as packageInfo so the item resolves its milestone context', async () => {
     mockDiscoverItems = [
       {
--- a/src/components/LearningPaths/MyLearningTab.test.tsx
+++ b/src/components/LearningPaths/MyLearningTab.test.tsx
@@ -499,9 +499,9 @@ describe('MyLearningTab launch flow', () => {
   });
 
   // Regression for #1637: without packageInfo, the loader falls through to
-  // plain fetchContent and the item renders as a bare document — no cover,
-  // milestone toolbar, or next/prev nav.
-  it('threads the inlined manifest through as packageInfo so the item lands on its cover', async () => {
+  // plain fetchContent and the item renders as a bare document — no
+  // milestone toolbar, progress chrome, or next/prev nav.
+  it('threads the inlined manifest through as packageInfo so the item resolves its milestone context', async () => {
     mockDiscoverItems = [
       {
         id: 'pkg-1',
--- a/src/components/LearningPaths/MyLearningTab.test.tsx
+++ b/src/components/LearningPaths/MyLearningTab.test.tsx
@@ -67,6 +67,7 @@ let mockDiscoverItems: Array<{
   contentUrl: string;
   milestoneCount?: number;
   description?: string;
+  manifest?: Record<string, unknown>;
 }> = [];
 let mockDiscoverExcludeTitles: Set<string> | undefined;
 
@@ -493,7 +494,36 @@ describe('MyLearningTab launch flow', () => {
     await waitFor(() => expect(prepareMock).toHaveBeenCalledTimes(1));
     expect(prepareMock).toHaveBeenCalledWith(
       'https://cdn.example/pkg-1/content.json',
-      expect.objectContaining({ title: 'Package one' })
+      expect.objectContaining({ title: 'Package one', packageInfo: undefined })
+    );
+  });
+
+  // Regression for #1637: without packageInfo, the loader falls through to
+  // plain fetchContent and the item renders as a bare document — no cover,
+  // milestone toolbar, or next/prev nav.
+  it('threads the inlined manifest through as packageInfo so the item lands on its cover', async () => {
+    mockDiscoverItems = [
+      {
+        id: 'pkg-1',
+        title: 'Package one',
+        contentUrl: 'https://cdn.example/pkg-1/content.json',
+        manifest: { type: 'path', milestones: ['m1', 'm2'] },
+      },
+    ];
+    prepareMock.mockResolvedValue(okResult);
+
+    render(<MyLearningTab onOpenGuide={jest.fn()} />);
+    fireEvent.click(screen.getByTestId(testIds.learningPaths.discoverMoreStart('pkg-1')));
+
+    await waitFor(() => expect(prepareMock).toHaveBeenCalledTimes(1));
+    expect(prepareMock).toHaveBeenCalledWith(
+      'https://cdn.example/pkg-1/content.json',
+      expect.objectContaining({
+        packageInfo: {
+          packageId: 'pkg-1',
+          packageManifest: { type: 'path', milestones: ['m1', 'm2'], id: 'pkg-1' },
+        },
+      })
     );
   });
 });
--- a/src/components/LearningPaths/MyLearningTab.tsx
+++ b/src/components/LearningPaths/MyLearningTab.tsx
@@ -231,10 +231,8 @@ export function MyLearningTab({ onOpenGuide }: MyLearningTabProps) {
         interaction_location: 'my_learning_discover_more',
       });
 
-      // Without packageInfo, the loader falls through to plain fetchContent
-      // and the item renders as a bare document — no milestone toolbar, no
-      // progress chrome, no next/prev nav. Mirrors the packageInfo threading
-      // already done for My Courses / App Platform guides above.
+      // prepareGuideLaunch backfills packageInfo from the URL when absent, but
+      // the manifest is already inlined here — passing it saves that re-fetch.
       const packageInfo: PackageOpenInfo | undefined = item.manifest
         ? { packageId: item.id, packageManifest: { ...item.manifest, id: item.id } }
         : undefined;
--- a/src/components/LearningPaths/MyLearningTab.tsx
+++ b/src/components/LearningPaths/MyLearningTab.tsx
@@ -232,9 +232,9 @@ export function MyLearningTab({ onOpenGuide }: MyLearningTabProps) {
       });
 
       // Without packageInfo, the loader falls through to plain fetchContent
-      // and the item renders as a bare document — no cover, no milestone
-      // toolbar, no next/prev nav (#1637). Mirrors the packageInfo threading
-      // already done for My Courses / App Platform guides above.
+      // and the item renders as a bare document — no milestone toolbar, no
+      // progress chrome, no next/prev nav (#1637). Mirrors the packageInfo
+      // threading already done for My Courses / App Platform guides above.
       const packageInfo: PackageOpenInfo | undefined = item.manifest
         ? { packageId: item.id, packageManifest: { ...item.manifest, id: item.id } }
         : undefined;
--- a/src/components/LearningPaths/MyLearningTab.tsx
+++ b/src/components/LearningPaths/MyLearningTab.tsx
@@ -230,7 +230,16 @@ export function MyLearningTab({ onOpenGuide }: MyLearningTabProps) {
         content_type: AnalyticsContentType.LearningJourney,
         interaction_location: 'my_learning_discover_more',
       });
-      void launch(item.contentUrl, item.title, item.id);
+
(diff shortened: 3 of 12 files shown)


Spidy88:
@cursoragent review

src/components/docs-panel/docs-panel.tsx
```diff
@@ -234,8 +234,20 @@ class CombinedLearningJourneyPanel extends SceneObjectBase<CombinedPanelState> i
...
+    // usePathfinderPluginConfig publishes — rather than whichever pluginConfig
+    // this particular surface happened to be constructed with. Sidebar builds
+    // its pluginConfig from Grafana's usePluginContext() snapshot, which can
+    // lag a settings save independently of the published global; using it here
+    // would silently re-stale the resolver on every sidebar (re)mount, undoing
+    // module.tsx's post-refresh re-wire. Falls back to pluginConfig when the
+    // global isn't set (e.g. unit tests constructing this class directly).
+    const resolverConfig = (window as unknown as { __pathfinderPluginConfig?: DocsPluginConfig })
+      .__pathfinderPluginConfig;
+    setPackageResolver(createCompositeResolver(resolverConfig ?? pluginConfig));
```
Spidy88, line 242: This was the AI recommended approach but it seems odd to take the global config over an explicit constructor config. @Jayclifford345 thoughts?

Spidy88 (commented):
@Jayclifford345 Will you review this PR as its related to recent work you did. Claude and Cursor are a lot more confident than I am about the architectural implications of this solution.

moxious (changes requested):
Thanks for digging into #1637 — the resolver half of this is a genuinely good catch. I confirmed your diagnosis independently: `HomePanel` renders `MyLearningTab` without ever constructing `CombinedLearningJourneyPanel`, so `resolvePackageMilestones` was returning `[]` for want of `_packageResolver` and the milestone chrome never got built. That's the fix, and it should land.

One blocker, and two things I'd fold into the same PR.

### Blocking: the two new entry-point imports grow `module.js` 6.2×

`src/module.tsx:12-13` imports the `./docs-retrieval` and `./package-engine` **barrels** statically into the webpack entry. Measured with `npm run build` on your head vs `main`:

|                   | `dist/module.js`                                               |
| ----------------- | -------------------------------------------------------------- |
| `main` (38c13a66) | 115,086 B — no webpack warning                                 |
| this PR           | **712,596 B** — new `entrypoint size limit` warning at 696 KiB |

CI's own bundle-size bot already posted this on the PR: `module | 696.23 KB (+583.74 KB) | +518.96%`, while total bundle size moves **−0.04%**. So nothing was added — 583 KB was reclassified from lazy chunks into the eager entry chunk, which is pure loss.

What's newly eager, traced through webpack `reasons`: **zod** (~271 KB unminified, via `docs-retrieval` → `lib/user-storage`), **dompurify** (~118 KB), the whole **`bundled-interactives` guide-JSON sync context module** (~251 KB), plus `learning-paths`, `json-guide.schema`, and prismjs + `prism.css`. The zod chain is the exact one the comment six lines above your imports warns about: _"the barrel would pull every hook (and zod, via user-storage) into module.js."_

Worth knowing why nothing stopped it: `src/lib/telemetry/entry-bundle-boundary.test.ts` is the only blocking entry-bundle guard, and it's a two-entry barrel denylist (`/telemetry`, `/hooks`). Its own comment records the incident it was written for — the hooks barrel _"roughly doubled module.js (113 KB -> 224 KB raw)"_. This is 5× that and passes clean, because the guard enumerates known offenders instead of bounding the outcome.

**Deep imports won't fix it** — I tried, 633,822 B, recovering only 79 KB (13%), because `package-content.ts` statically imports the whole `content-fetcher` orchestrator. Two shapes that do work, both measured:

- **A lazy factory** (cleanest): a leaf module holding the singleton with `setPackageResolverFactory(thunk)` and an async `getPackageResolver()`, registered **synchronously** in `plugin.init`, with `/* webpackPrefetch: true */ import('./package-engine/composite-resolver')` inside the thunk. → **116,329 B**, `tsc` clean, no warning. Every `_packageResolver` read site in `package-content.ts` is already inside an `async` function, so the getter drops straight in. (Keep the read inside `fetchPackageContent`'s existing `Promise.all` so the chunk fetch overlaps `fetchContent`.)
- **A plain `void (async () => { await import(...) })()` inside `plugin.init`** → **116,373 B**. I checked the race this raises against `plugin.init`'s "everything must stay synchronous" rule: the auto-open and `?doc=` deep-link paths both run through `FullScreenPanel`/`FloatingPanelManager`, which construct the panel and so install the resolver synchronously anyway. The only path depending on the `plugin.init` wiring is a home-page click, which can't outrun a chunk fetch. So this is adequate — the factory just removes the question entirely, and it also drops the double eager `RepositoryJsonSchema.parse()` this PR currently runs at init.

Either way, could you extend the guard in the same PR? Ideally swap the denylist for an outcome bound — a committed byte ceiling on `dist/module.js`, or `performance: { hints: 'error', maxEntrypointSize: … }` in the production webpack config.

### The `packageInfo` half fixes something that isn't broken

`prepareGuideLaunch` already backfills package context when the caller passes none — `src/components/docs-panel/utils/prepare-guide-launch.ts:86-89`:

```ts
if (!packageInfo && isPackageContentUrl(url)) {
  packageInfo = await fetchPackageInfoFromUrl(url);
}
```

Discover More's `contentUrl` is `https://interactive-learning.grafana.net/packages/<id>/content.json`, which satisfies both `isInteractiveLearningUrl` and `PACKAGE_CONTENT_URL_PATTERN` — I checked the live index and all 71 `type: "path"` entries use single-segment paths, so every one matches. There's a second backfill at `docs-panel.tsx:903-904` in `loadTab` too. So the manifest was already arriving; the bare-document symptom was the resolver alone.

That makes the threading a saved round-trip rather than a bug fix, which matters because the "renders as a bare document — no milestone toolbar, no progress chrome, no next/prev nav" rationale is now recorded in five places (both prod files, the `DiscoverMoreItem.manifest` JSDoc, and as the stated reason for both new tests) and it's the thing a future reader will trust when this breaks again. I'd keep the threading for the latency win but reframe it as "avoids the redundant sibling-manifest fetch `prepareGuideLaunch` would otherwise issue", and correct the PR body so the fix is attributed to the resolver wiring.

One real delta while you're in there: setting `packageInfo` at the call site pre-empts `fetchPackageInfoFromUrl`, which returns a `ManifestJsonObjectSchema`-parsed object — unknown keys stripped, `repository` defaulted to `'interactive-tutorials'`. The threaded raw manifest keeps unknown fields and loses that default, and `repository` is what feeds `guideSource` in `resolveCompletionIdentity` and the public-URL synthesis in `package-content.ts`. It also persists: `docs-panel.tsx:424` writes `packageInfo` into tab storage and `tab-storage-restore.ts:196` reads it back verbatim. A `ManifestJsonObjectSchema.loose().safeParse(entry.manifest)` in `toDiscoverItem` (falling back to `undefined` so the existing path still runs) makes the threaded path no weaker than the one it replaces — and matches what `online-cdn-resolver.ts:83` already does with the same objects.

### The load-bearing half is untested

I reverted both the `module.tsx` wiring and the `docs-panel.tsx` config-source change and ran the suite: **446 suites / 7,117 tests still pass**. The docs-panel tests that construct the panel mock `setPackageResolver` and `createCompositeResolver` as bare `jest.fn()` and never assert arguments, so a typo in the `__pathfinderPluginConfig` property name would fall back to `pluginConfig` silently. The two new tests cover the half that turns out not to be the fix.

The seam from the bundle fix makes this cheap: assert that after registration `resolvePackageMilestones(['m1'])` returns non-empty (it returns `[]` with no resolver installed). Plus ~10 lines in `docs-panel.alignment.test.ts` — set `window.__pathfinderPluginConfig` to a marker, construct with a different `pluginConfig`, assert `createCompositeResolver` got the marker; second case with the global deleted asserts the fallback. That pins both branches and catches the typo case.

### Smaller

- The new comments run 10 and 9 lines. Per AGENTS.md ("if the comment won't fit on one short line, rename or restructure instead") I'd keep one line each — the ordering invariant in `module.tsx`, "seed from the published global, not this surface's snapshot" in `docs-panel.tsx` — and drop the JSDoc body and the two test comments, with the reasoning living here in the PR body instead.
- The `docs-panel.tsx` comment says the resolver "must always be seeded from the one authoritative config source", but nothing enforces that after boot — no `PATHFINDER_CONFIG_UPDATED_EVENT` subscriber. It's unreachable today since every settings-save path calls `window.location.reload()`, so I'd trim the claim rather than build the subscription.

### Follow-ups, not gates

- `setPackageResolver` now has three call sites with no seam modelling "install the resolver for the current config", and its docstring still says "Called once at app startup." Worth giving it one named installer and recording the `package-engine` contract anchor that #1611 also deferred — the bundle fix does most of this anyway.
- `{ packageId: X.id, packageManifest: { ...X.manifest, id: X.id } }` now appears three times (`MyLearningTab.tsx:238`, `:216`, and `CustomGuidesSection.tsx:31`, where it's already named but unexported). One exported helper would do.

### Things I checked and cleared, so you don't chase them

- `manifest.type` is required by `ManifestJsonObjectSchema` and present in every live manifest, so `getPackageRenderType` isn't a failure mode.
- The `id: item.id` override is safe — `build-repository.ts:133` hard-errors when `manifest.id !== entry.id`, so it's a no-op for well-formed repositories.
- No staler config can win: `publishPathfinderPluginConfig` at `module.tsx:149` runs before the wiring at `:162`.
- The `window.__pathfinderPluginConfig` read is fine — #1611 explicitly blessed direct global reads for scene construction, and `FullScreenPanel`, `FloatingPanelManager` and `docsPage` already did it. Your change is net convergence: it removes the sidebar's lagging `usePluginContext()` snapshot as a fourth source.
- Resolver construction does no I/O, and revert is clean — nothing persists.

One caveat on my own analysis: I couldn't fully pin down Grafana core's exact `module.js` load trigger (`plugin.json` has no `preload: true`; extensions target the extension sidebar and command palette), so "every page load" may be too strong. But the cost lands before anyone opens Pathfinder either way, and total bundle size is unchanged, so the regression holds regardless.

Spidy88:
Thanks for the deep dive — every point checked out against the code and I've pushed fixes for all of it.

**Bundle size (blocking)**: went with the lazy-factory approach. Split the resolver singleton out of `package-content.ts` into a new dependency-free leaf module (`package-resolver-registry.ts`); `module.tsx` now registers a factory that dynamically imports `composite-resolver.ts` directly instead of constructing it synchronously via the barrels. `module.js` is back to ~116 KB, confirmed via `npm run build`. Also extended `entry-bundle-boundary.test.ts` with the same denylist pattern already used for the hooks barrel, covering `docs-retrieval` and `package-engine` — same mechanism, not the webpack `performance.hints` route, since I didn't want to touch the vendored build config without understanding how it interacts with this repo's CI.

**packageInfo misattribution**: agreed and corrected — the PR body now attributes the fix to the resolver wiring, and describes the manifest threading as the latency optimization it actually is (avoids the redundant `fetchPackageInfoFromUrl` re-fetch `prepareGuideLaunch` would otherwise issue).

**Schema gap**: fixed — `toDiscoverItem` now runs the manifest through `ManifestJsonObjectSchema.loose().safeParse` before threading it through, same as `online-cdn-resolver.ts:83`. `DiscoverMoreItem.manifest` is now typed `ManifestJson` instead of a raw `Record<string, unknown>`.

**Test coverage**: added — `package-content.test.ts` covers the new `setPackageResolverFactory` (resolves once available, factory invoked once across repeated reads, a later `setPackageResolver` overrides a pending factory); `docs-panel.alignment.test.ts` now asserts `createCompositeResolver`'s actual argument in both branches (global present with a different value than the constructor's `pluginConfig` → global wins; global absent → falls back).

**Comments**: trimmed both to one line each, and dropped the post-boot "must always be seeded" claim — you're right that nothing enforces it once mounted, and it's unreachable today since every settings-save path reloads the page. Also shortened the four comments describing the (now-corrected) "bare document" rationale rather than duplicating it — that context lives in the PR body now.

**Follow-ups**: agreed on scope — not doing the `setPackageResolver` call-site consolidation / `CONCERNS.md` contract anchor or the `packageManifest` object-literal helper in this PR. Happy to file both as follow-up issues if that's useful, or leave it to you/Jay to prioritize.

Commits: `2ffad62e` (bundle size + resolver registry) and `06d84605` (manifest schema fix + attribution corrections).

moxious:
1. Medium — the "lazy" factory isn't lazy; ~231 KB of chunks now download at plugin.init

setPackageResolverFactory does _resolverPromise = factory().catch(...) — it invokes the thunk at registration, not on first read. Built proof in dist/module.js:

function a(e){r=e().catch(...)}                    // registry: calls immediately
(0,f.DY)(()=>Promise.all([n.e(769),n.e(760),n.e(78),n.e(604)])...)   // plugin.init

Main's plugin.init contains no n.e( calls at all. Chunk 760 is already pulled by the existing experiments top-level await, so the true incremental cost is 769+78+604 = 231,252 B raw / ~55 KB gzipped, fetched on every page load where module.js loads — for users who never open Pathfinder. getPackageResolver() is read only from package-content.ts's four fetch paths, so anyone who does open Pathfinder pays nothing extra; the cost lands exclusively on the non-users.

Two riders: the JSDoc claim "lazily constructs the resolver on first read" is false, and it('invokes the factory only once across repeated reads') passes for the wrong reason. webpackPrefetch: true is inert — the chunk is fetched immediately, so the hint can only race the fetch it duplicates.

Fix: store the thunk, invoke it inside getPackageResolver(). If you do, also fold the read into fetchPackageContent's existing Promise.all — package-content.ts:240 currently does const resolver = await getPackageResolver() before the Promise.all, so a genuinely lazy factory would serialize a full chunk round-trip ahead of fetchContent(contentUrl).

Spidy88:
Confirmed and fixed, using the same decompiled-output technique to verify — thanks for the precise proof, it made this fast to track down.

`setPackageResolverFactory` was calling `factory()` immediately at registration (`function a(e){r=e().catch(...)}` in the old build), not on first read. Since `plugin.init` calls it unconditionally, the composite-resolver dynamic import fired on every page load where `module.js` loads — including for users who never open Pathfinder, exactly the cost the lazy design was supposed to eliminate, just moved from the entry bundle to an eager network fetch.

Fixed: `setPackageResolverFactory` now only stores the thunk; `getPackageResolver()` invokes and memoizes it on first actual read. Re-verified in the rebuilt output: `function s(e){r=e,i=void 0}` (store only) and `function c(){if(!i){...i=r().catch(...)}return i}` (invoke lazily, memoize).

Both riders addressed too:
- Dropped `webpackPrefetch: true` on both `import()` call sites — inert while the bug made the import fire immediately anyway, and actively wrong once genuinely lazy (package-content.ts's four read sites need it promptly when they call it, not at idle-time priority).
- `fetchPackageContent`'s baseUrl-resolution branch now awaits `getPackageResolver()` inside the `Promise.all` instead of sequentially before it, so a cold resolver's chunk fetch overlaps `fetchContent(contentUrl)` rather than serializing in front of it.

Also added a test that asserts the factory isn't invoked until the resolver is actually read (`expect(factory).not.toHaveBeenCalled()` immediately after registration) — the existing "invoked once across repeated reads" test passed either way, since it never distinguished eager-at-registration from lazy-at-first-read.

Commit: `288bca49`.

moxious (approved):

PR #1639, opened 2026-08-17
feat(mcp): consolidate authoring tools into enablement suites

dgehring-grafana:
## Summary

Condenses the Pathfinder authoring MCP surface and, more importantly, changes **how** a tool's parameters are defined. Tools no longer restate CLI verbs as hand-written Zod schemas. For CLI-backed tools, the parameter surface is *projected* from Commander at runtime: `pathfinder_help` reads the bound command's own option surface and republishes it as the exact camelCase keys the tool accepts, and the same projection preflights every call.

The intent is that the MCP surface tracks the CLI without an adapter edit. A new CLI option reaches agents through `pathfinder_help` on its own; only a binding that names runner fields needs code changes.

### Two tool contracts, marked in each file

Every tool file now declares `Contract: cli-routed` or `Contract: mcp-native` in its header, because the two are discovered differently.

**`cli-routed`** — takes a kebab-case CLI `operation` plus an opaque `opts` bag. Interface comes from `pathfinder_help`, never from the tool's own schema.

| tool | operations |
| --- | --- |
| `pathfinder_manage_block` | `add-block`, `edit-block`, `remove-block`, `add-step`, `add-choice` |
| `pathfinder_manage_guide` | `set-manifest` |
| `pathfinder_create_package` | `create` |
| `pathfinder_inspect` | `inspect` |
| `pathfinder_get_schema` | `schema` |

**`mcp-native`** — owns an explicit top-level Zod schema; agents do not call `pathfinder_help` for it: `pathfinder_authoring_start`, `pathfinder_help`, `pathfinder_validate`, `pathfinder_read_session`, `pathfinder_read_repository`, `pathfinder_finalize_for_app_platform`.

`add-step` and `add-choice` are operations on `pathfinder_manage_block` rather than dedicated tools. `pathfinder_manage_guide` exists so guide-level metadata has a home that is visibly *not* the per-block tool — agents were reaching for `manage_block` to write the manifest.

### What the projection buys

- **Withheld parameters are rejected, not dropped.** `validateCommandArgs` preflights `opts` against exactly what `pathfinder_help` published, so a misspelled or withheld key fails with `UNSUPPORTED_PARAMETER`. Previously such keys were accepted and silently ignored.
- **The `superRefine` validators are gone.** The CLI runners remain the sole validator of guide *content*; a tool validates only its own *argument shape*.
- **Positionals are projected as named options,** so an agent never assembles a command line. `create --type`, which raw Commander help omits, is republished.
- **Hidden flags stay hidden.** `optBlacklist` is subtractive against the projection: `dir` everywhere, plus `before`/`after`/`position` on `add-block` (with `type` as the subcommand selector) and `orphanChildren` on `remove-block`. MCP authoring is append-only.
- **`hints` are stripped outbound.** They are argv recipes, which do not apply to a caller that never builds a command line.

### Making the contract legible

`pathfinder_authoring_start` gains an `interfaceContract` section that states both halves once — JSON in, using the projected camelCase names; the CLI's own `CommandOutcome` out — so the line between "what we project for convenience" and "what the CLI actually returned" is explicit for agents and for humans reading the code. Per-field `optContext` prose was removed in favor of Commander descriptions, so guidance lives where the projection will pick it up rather than in MCP-only copy that drifts.

`explicitId` is renamed `id`. `pathfinder_help`'s command list is filtered to MCP-bound commands only, so CLI-only verbs (`validate`, `move-block`, `e2e`) no longer appear. Contract anchor recorded in `docs/design/CONCERNS.md` (`mcp-authoring-server` → #1639).

## Reason for change

The MCP started as a near-1:1 mirror of CLI verbs. That inflated the tool catalog, increased tool-pick errors (especially via Assistant), and burned context — but the deeper cost was that every CLI option change needed a matching MCP edit, and the two drifted whenever one was forgotten. Projecting the interface removes the duplicate definition rather than adding a process to keep two copies aligned.

## Alternatives considered

- **Keep the granular per-verb tools.** Lower schema complexity, worse tool-pick and context cost.
- **Single `manage_*` tool with an `operation × resource` matrix.** Tried; the dead cells were worse for agents than folding children in as operations.
- **Tell the agent the translation rule** ("parameters are camelCase, `parent-id` → `parentId`") and let it consume raw CLI help. Rejected: raw help omits `create --type`, leaks flags we withhold, exposes CLI-only verbs, and gives no preflight — an unsupported key would be silently dropped instead of rejected.
- **Full `{operation, params}` passthrough.** This was listed as deferred in the original version of this PR; it is what the PR now does.

## Known gaps

These are understood and recorded, not oversights.

- **The runner seam is still named parameters.** Runners take named arguments, so a projected option is only forwarded if a binding names it. Today the blacklist keeps that surface aligned deliberately — it steers agent behavior — but the structural fix is to have runners accept a single options bag, at which point the alignment stops depending on the blacklist.
- **`--position <n>` types as `string`** in the projection, because the `valueType` heuristic looks for `<number>`. Latent only while the flag is blacklisted; it becomes live if the flag is ever published.
- **`authoring_start`'s prose is unpinned.** Only tool names are asserted, so the routing copy above is exactly the part with no regression test.
- **`valueType` inference is duplicated** between `formatHelpAsJson` and `optionToHelpFlag`. Drift risk, not a bug.
- **`serverInfo.version` is `CURRENT_SCHEMA_VERSION`,** so the pre- and post-change tool catalogs are byte-identical on `initialize`; a rollback can only be confirmed by inspecting `tools/list`.

## Follow-ups

- Convert the runners to a single options bag, deleting the named-parameter seam and `explicitId` with it. Removes the silent-drop class structurally instead of steering around it.
- Fix the `<n>` placeholder (or read `argParser`) before any numeric flag is published.
- Exercise the routing boundary with a real agent authoring pass, ending in a manifest write.
- Collapse the duplicated `valueType` ladder once the serializer is touched upstream.

## Test plan

- [x] Full Jest suite green (447 suites, 7209 tests)
- [x] `command-interface.test.ts` covers the projection and preflight directly — camelCase rekeying, blacklist withholding, subcommand selection, `requiredByType`, and type mismatches — without booting a server
- [x] MCP suite green (23 suites, 263 tests), including assertions that `hints` are absent from outcomes
- [x] `pathfinder_help` command list asserted to contain only MCP-bound commands and to exclude `validate` / `move-block` / `e2e`
- [x] `typecheck` clean; `lint` 0 errors; `prettier` clean
- [x] Governance suite (169) and contract-evolution suite (22) green
- [x] Playwright e2e: 18 passed, 5 skipped (two-tab controller specs gated behind `TWOTAB_CONTROLLER_ENABLED`)
- [x] Smoke-test the condensed surface in Grafana Assistant


dgehring-grafana:
As a follow-up to this, I'm wondering at the possibility of making tool calls mostly/completely CLI driven - my thought is this, the description of the tool denotes the operations it can perform, with a 1:1 mapping of those into CLI call sites.

Then, we can instruct the agent to use the "help" tool to gather the interface for the specific operation call, and leave validation for the CLI layer, passing what the agent gives us through transparently into the CLI invocation.

This would enable the tools to passively maintain parity with the CLI interface should we decide to change it, and would make the MCP server align more closely to the thin passthrough layer the initial RFC envisioned.

There are a few areas of concern with this - for example, this PR explicitly doesn't expose positional arguments for block additions, even through they're supported, because we want to generate guides sequentially. We'd have to find a way to filter certain flags from being exposed to the agent. It also introduces more round trips for context, and might negatively impact the error UX if CLI errors aren't as user friendly. I'm sure there's other things.

moxious (changes requested):
## Review summary

The consolidation itself is well-built. Two-mode input, ETag integrity, session pinning, and every transport cap are provably untouched (`dispatchMutation` is byte-identical to base; `read-input.ts`, `result.ts`, `session-store.ts`, `transports/http.ts`, `utils/etag.ts` have zero diff). The raw-shape → `z.object().superRefine()` switch is sound on both stdio and streamable HTTP. Docs got real attention.

But three defects make the primary authoring path **silently produce wrong guide content**, and two of them are created by this PR's own new agent-facing copy. Requesting changes on those.

**Gate checks (all green on `52525ade`):** `tsc --noEmit` · `eslint src/cli/mcp --max-warnings 0` · prettier · `npm run test:ci -- src/cli/mcp` (22 suites / 232 tests) · `architecture.test.ts` (26 tests, Node closure intact).

Everything below was verified by running it against a real in-memory MCP client on the head commit, not read off the diff.

---

### 1. [blocking] `fields` keys must be camelCase, but this PR tells agents to use the kebab-case flag names — optional fields then vanish silently

`mutation-tools.ts:320`, `:374` and `authoring-start.ts:100` are all newly added and all say _"CLI flag names (e.g. `--content`, `--action`) become keys in `fields`"_ — illustrated only with single-word flags, where it happens to hold. `parseOptionValues` (`schema-options.ts:334-366`) keys off the **camelCase** Zod shape name, and nothing normalizes kebab → camel on the MCP `fields` path.

```
help for add-block/section publishes the flag as:  "auto-collapse"

fields:{title:'S','auto-collapse':true} → status:"ok" → stored {type,id,title,blocks}     ← attribute GONE
fields:{title:'S', autoCollapse:true}   → status:"ok" → stored {…,"autoCollapse":true,…}  ← works
```

Required multi-word fields fail _confusingly_: `fields:{'input-type':'text'}` returns `"inputType: expected string, received undefined"` — the agent is told the key it did not send is wrong. Optional ones fail **silently with `status: ok`**. This reaches `inputType`, `variableName`, `autoCollapse`, `checkboxLabel`, `validationMessage`, `dataCheckQuery`, `stepTimeout`, `showMe`, `doIt`, `authorNote`, and more.

The mechanism predates this PR. What is new is prescribing the wrong transcription in three places, which turns a latent trap into documented guidance.

**Suggested fix:** normalize kebab → camel for `fields` keys at the MCP boundary (a pure widening — camelCase keeps working), or correct the copy to say camelCase and use one multi-word example. Either way, having `parseOptionValues` surface unrecognized keys as an outcome warning would also surface finding 2.

### 2. [blocking] `pathfinder_help` still advertises the five withdrawn capabilities, and they are silently dropped

`ManageBlockInputSchema` is a non-strict `z.object`, so `ifAbsent` / `before` / `after` / `position` / `orphanChildren` are stripped without error — while `pathfinder_help`, the discovery surface this PR points agents at _more_ aggressively than base, still publishes every one of them:

```
help add-block/markdown → optional: ["before","after","position",…]   (+ "if-absent" in text)
help add-block/section  → optional: ["before","after","position",…]
help remove-block       → optional: ["cascade","orphan-children"]
help edit-block         → optional: […,"position","before","after"]
```

Three compounding consequences:

- `{operation:'add-block', position:0, before:'first', …}` → `status:"ok"`, block appended at the end. The agent asked to insert first, got an append, and was told it succeeded. **Guide order silently wrong.**
- `CONTAINER_HAS_CHILDREN` (`mutators.ts:340`) reaches agents saying _"pass `--cascade` … or `--orphan-children`"_. `orphanChildren` is now stripped, so the honest retry returns the **byte-identical error** — a retry loop. The only reachable remedy is `cascade:true`, which deletes the subtree; the session store has no history (`load`/`save`/`delete` only), so "unwrap this section, keep its contents" is now lossy with no undo. `edit-block`'s position error likewise redirects to `move-block`, which has no MCP tool.
- `ifAbsent` exists precisely for this surface — `mutators.ts:97-103` warns that without it a crashed-and-retried add "produces a fresh auto-id instead of the no-op the agent expected — **duplicating content silently**." That is now the only MCP behavior.

The removal is a legitimate, documented product call (`HOSTED-AUTHORING-MCP.md:120`). The defect is that it is invisible at the surfaces agents actually read.

**Suggested fix:** reject the five names in `superRefine` with a message naming the replacement (preferable to `.strict()`, which could trip on client-injected extras); filter them out of `formatHelpAsJson` when served through `pathfinder_help`, or bucket them as `cliOnly`; re-word `CONTAINER_HAS_CHILDREN` and the position redirect in MCP terms on the MCP path. Also worth tightening `authoring-start.ts:79` and `cascade`'s `.describe()` to state that cascading a populated container destroys its subtree.

### 3. [blocking] The new container-id gate is a _broader_ predicate than the CLI rule it mirrors

The CLI condition is "the assembled block has an id from any source" — `add-block.ts:236` is `isContainerBlockType(type) && !block.id`, and `block.id` can come from `flagValues.id`. The MCP gate tests only the top-level arg:

```
{operation:'add-block', type:'section', id:'sec-top'}             → status:"ok"
{operation:'add-block', type:'section', fields:{id:'sec-fields'}} → -32602 add-block with type "section" requires `id`
```

The caller supplied an id and is told it did not. That route worked at base and the CLI still accepts it.

Compounding: the rule now lives in two independent predicates with **no equivalence test**, and the one test that exists pins the strict side — so relaxing the CLI later would drift silently in the unguarded direction.

**Suggested fix:** test `args.id ?? args.fields?.id` and normalize in `manageBlockRunner`, or drop the gate and re-word the CLI error at the envelope. A table-driven test over `BLOCK_TYPES` asserting the two predicates agree would fail today and would catch drift in either direction.

### 4. [blocking] Contract anchor + routing keywords owed in this PR

`contract-evolution-policy.mjs` returned `contract_missing` / advisory / medium. There is no anchor row for `mcp-authoring-server`, and the catalog contract is asserted in eight prose locations with no authority — `HOSTED-AUTHORING-MCP.md:72`, `MCP_SERVER.md:342-344`, `PATHFINDER-AI-AUTHORING.md:46/70` and `AI-AUTHORING-IMPLEMENTATION.md:197/351` all still say the MCP "performs no schema validation of its own" (one calls it _structurally enforced_) while the new `superRefine` makes the MCP a third enforcement site for container-requires-id.

For the record: the scan deliberately declined `contract_branching`, which at `use_ordinal: third_or_later` **would have blocked**, on the grounds that the reconstructed contract is itself incoherent — "the tool list mirrors the CLI command surface" was already violated five times before this PR. History is evidence, not authority.

Per the review skill's §5, a PR that replaces a contract carries the anchor:

- **Add** an `mcp-authoring-server` row to Contract anchors, owner `registerAuthoringTools` (`tools/index.ts`), recording: multi-verb capabilities get one kebab-case `operation` tool; append-only child members stay dedicated; and **the CLI is the sole validator of guide _content_, while a tool may validate its own argument _shape_** — the distinction the PR body draws correctly but leaves nowhere durable. (Note the container-id rule arguably crosses that line.)
- **Fix** `CONCERNS.md:61` `trigger_keywords`, which still name `pathfinder_list_packages` and `pathfinder_get_package`. Keywords pointing at deleted symbols can never fire again, quietly reducing review coverage for this exact subsystem.

---

### 5. [suggestion] A live spec still documents the deleted names — and this PR's own edit links to it

`AI-AUTHORING-IMPLEMENTATION.md` self-describes as "the living index" (line 6), is cited from `repository-tools.ts:13`, and is the target of the sentence this PR rewrote at `HOSTED-AUTHORING-MCP.md:133`. It names the deleted tools at 10 sites (`:181`, `:184`, `:278-279`, `:298`, `:325-327`, `:343`, `:346`). `MCP-AGENT-UX-HARDENING.md:218` and its issue-#4 code pointer are stale too. `docs/design/phases/*.md` are correctly frozen historical records — leave those.

### 6. [suggestion] Test gaps sit exactly on the new semantics

- **`authoring_start`'s `workflow` / `rules` / `discovery` arrays have zero assertions.** This PR's whole purpose is renaming what those arrays advertise. A bad merge could leave the index naming `pathfinder_add_block` while `tools/list` publishes `pathfinder_manage_block`, with all 232 tests green — and then every agent follows a dead index. A ~6-line two-way check against `listTools()` closes it permanently, covering `SERVER_INSTRUCTIONS` and `state-bridge.ts:226` as well.
- **5 of 10 `superRefine` branches unpinned:** `edit-block` missing `id`, `edit-block` empty `fields`, `remove-block` missing `id`, `read_repository get-manifest` missing `id`, and 5 of 6 container types (only `section` is tested — `it.each(CONTAINER_BLOCK_TYPES)` fixes that one).
- **The new boundary tests assert too loosely.** `server.test.ts:288` and `:350` use `/invalid arguments|type/i` and `/invalid arguments|id/i`, which match the SDK's generic prefix; `/id/i` is doubly vacuous ("inval**id**"). They would stay green if the hand-written message vanished. `:309` shows the right shape (``/`id`/`` plus `not.toMatch(/--id/)`).
- No test pins `cascade`'s omitted default. Low risk — the CLI guard is `!options.cascade` on an optional boolean, so it cannot silently invert.

### 7. [suggestion] Observability collapses 3 → 1 with no replacement dimension

`rpcToolName` (`access-log.ts:153`) is the only per-tool dimension and is documented as existing to "break token spend down by tool"; `operation` lives in `params.arguments`, which is never logged. Adds and cascade-deletes are now indistinguishable in the access log — `http.test.ts` even pins the coarser value. `withLoadedSession(…, 'read_session')` similarly loses which operation faulted in the stderr envelope. Consider a bounded `rpcToolOperation?: string` populated only from the known enum values (never free-form params, which reach the log before Zod runs), plus interpolating the operation into the envelope name.

### 8. [question] `destructiveHint` is now `true` for block appends

`writeDestructive('Manage Pathfinder block')` replaces base's `writeAppend` on the append path. Clients that gate on the hint will prompt on every add — a 30-block guide becomes 30 confirmations, which trains users to click through the one `remove-block` that looks identical. `add_step` / `add_choice` keep `writeAppend`, so the annotation surface is now internally inconsistent about what an append costs. This is inherent to the merge rather than a bug — flagging for a conscious sign-off rather than an inherited default.

### 9. [nit] Comment hygiene (QC8)

- **Stale header on a heavily-touched file:** `mutation-tools.ts:9-13` still says the schemas are "intentionally permissive … schema-illegal output is impossible because it is impossible in the CLI" — directly contradicted by the two boundary validators this PR adds. Trim-on-touch applies.
- **Six new comments** restate the code or defend a non-action: the three "Single source of truth for `<tool>` args … (same pattern as manage_block)" blocks (shape 1 + 4 — the const is _named_ `ManageBlockInputSchema`), the two identical "Flat object + superRefine so…" comments sitting directly above `inputSchema: <thatVerySchema>`, and `/** Map a Zod-validated block operation to its existing CLI runner. */` above `manageBlockRunner`. Keepers: the container-id rationale (compress to one line), the `.default(false)` invariant, and the "different data source" naming notes.

### 10. [nit] Smaller items

`edit-block` with `fields:{}` now returns a raw Zod error instead of the CLI's structured `NO_CHANGES` code — and `NO_CHANGES` is still reachable via a _misspelled_ field, so one author mistake yields two envelopes · `; ;` typo at `HOSTED-AUTHORING-MCP.md:105` · the `authoring_start` JSON example in that doc diverges from the live `AUTHORING_CONTEXT` in ~8 of 9 keys (`status`, `authoringContextVersion`, `schemaVersion`, `domain`, `blockTypes`, `tutorial`, `instructions` do not exist; `discovery` is an array in code, an object in the doc) — pre-existing, but the PR edited those lines and disclaimed only `tutorial` · three same-named `describe` blocks per read-tool test file · addressing fields validate trimmed but dispatch untrimmed (benign — the `NOT_FOUND` quotes the padding back).

---

### Claims checked and dropped

Recording these so they do not get re-raised. The SDK **does** validate a `z.object().superRefine()` `inputSchema` on both stdio and streamable HTTP (`safeParseAsync` runs before the handler; unknown top-level keys are stripped; `__proto__` is safe), and the published JSON Schema keeps every property, enum and `required` entry — the schema-shape switch is sound. `cascade`'s `.default(false)` survives the refinement. The three `switch`es are compile-exhaustive (`TS2366`, confirmed by adding a fourth member to each union). The two `CONTAINER_TYPES_TEXT` derivations **cannot** diverge — `BLOCK_TYPES` _is_ `Object.keys(BLOCK_SCHEMA_MAP)`. The missing `notifications/tools/list_changed` is a non-issue: the HTTP transport builds a fresh server per request, so no client can hold a stale catalog. The 135 deleted lines in `artifact-tools.test.ts` were template-only — net `it()` count is unchanged at 150 and no `create_package` coverage was lost. `pkg/` has no MCP surface, so deleting `pathfinder_create_guide_template` orphans nothing. All six `pathfinder_help` command names the new copy cites resolve in `CLI_COMMANDS`. Security: **zero findings at medium or above**; F1–F6 all `not_applicable` (Node-side code); all four transport caps verified live, including a 413 at 1.2 MB.

### Two process notes

**The PR body understates the blast radius.** Gotchas cites clients calling `add` / `list` / `list_blocks` / `pathfinder_repository` — but three of those four never shipped on `main`; they are intermediate names from this branch's own commits. The real break against `main` is **ten shipped tool names deleted with no alias**, plus `explicitId` → `id` and five withdrawn capabilities. Given this deploys as a public `--allow-unauthenticated` Cloud Run endpoint whose own output tells third parties to `claude mcp add` it, that belongs in the body and in a CHANGELOG `### Changed` entry. Relatedly, `serverInfo.version` advertises `CURRENT_SCHEMA_VERSION` (the _guide_ schema, `1.1.0`), so the two catalogs are byte-identical on `initialize` and a rollback cannot be verified from outside except by eyeballing `tools/list`.

**A routing blind spot this PR happens to expose.** `ai-subsystem`'s `trigger_paths` are `.cursor/**` / `AGENTS.md` / `CLAUDE.md` only, so it fired here only because it was dispatched explicitly. A future PR that changes nothing but MCP prompt prose gets `mcp-authoring-server` alone and never gets that lens — even though findings 1 and 2 are both _prose_ defects. Worth a follow-up issue adding `src/cli/mcp/lib/server-instructions.ts`, `agent-routing.ts` and `tools/**` to that row.

---

**Minimum bar to merge:** findings 1 and 2 (both silent-wrong-output on the main authoring path), finding 3 (false rejection of a route the CLI accepts), and the same-PR doc obligations in 4. Findings 5–10 are reasonable follow-ups.

<sub>Reviewed with `/review` (routed concern fan-out + adversarial verification). Findings 1, 2, 3 and the kebab/camel and container-id behaviours were reproduced against a live in-memory MCP client on `52525ade`.</sub>

dgehring-grafana:
## Review findings — resolution

Gates on head: `tsc --noEmit` · `eslint src/cli/mcp --max-warnings 0` · `prettier` · `npm run test:ci -- src/cli/mcp` (23 suites / 263 tests).

Three of the four blocking findings are closed by removing the mechanism rather than patching it — the `fields` → `opts` redesign made two of them unreachable.

**1. kebab vs. camelCase in `fields` — closed.** `fields` became `opts`, and `pathfinder_help` now publishes the real camelCase parameter names, so there is no transcription step left to get wrong. Unrecognized keys are rejected with `UNSUPPORTED_PARAMETER` instead of being silently stripped, which also covers the "surface unrecognized keys" suggestion. All three prescriptive copy sites are gone; the surviving example cites a real flag (`--if-absent` → `ifAbsent`). Pinned in `server.test.ts`: `autoCollapse` present, `auto-collapse` absent.

**2. Withdrawn capabilities still advertised — closed.** `add-block` blacklists `before`/`after`/`position` and `remove-block` blacklists `orphanChildren`, so they are neither published nor accepted. `edit-block` needed no change — its placement flags are already `.hideHelp()` in Commander, so the projection never saw them. `ifAbsent` was **restored** rather than withheld; the review was right that it exists precisely for this surface. `CONTAINER_HAS_CHILDREN` is re-worded on the MCP path to name `cascade: true`, with a test asserting `--orphan-children` no longer appears, and cascade's destructiveness is stated in the option description.

**3. Container-id gate broader than the CLI rule — closed.** The MCP-side `superRefine` is deleted; the CLI predicate is the sole authority, so there is nothing left to drift. Because `id` now travels inside `opts`, the route the CLI accepts is the route the MCP accepts. The only remaining `superRefine` calls are in the two mcp-native read tools, validating their own argument shape — the boundary the new anchor documents.

**4. Contract anchor + routing keywords — done.** Added an `mcp-authoring-server` anchor owned by `registerAuthoringTools`, recording the operation-tool shape, append-only child members, and the CLI-validates-content / tool-validates-argument-shape split. `trigger_keywords` no longer name deleted symbols. The prose was corrected rather than just anchored: `AI-AUTHORING-IMPLEMENTATION.md` now states the MCP performs no *guide-content* validation and preflights argument shape against the `pathfinder_help` interface. `docs/design/phases/*` left frozen as advised.

**5. Live specs naming deleted tools — done.** Swept the living index and `MCP-AGENT-UX-HARDENING.md`. The same defect turned up in nine test names still saying `add_block` / `edit_block` / `remove_block` / `set_manifest`; those now name the kebab-case operations they actually pass.

**6. Test gaps — mostly done.** Added the two-way `authoring_start` ↔ `listTools()` index check. Replaced the vacuous negated `arrayContaining` with per-key `not.toContain` (two sites, not one). Tightened the loose `/invalid arguments|…/` regexes. The five unpinned `superRefine` branches are moot after finding 3. **Not done:** `cascade`'s omitted default is still unpinned (agreed low risk — `!options.cascade` cannot silently invert).

**7. Observability — done, both halves.** Added a bounded `rpcToolOperation` populated only from the published enum, with a test pinning it, and the error envelope now logs `mutation:${operation}` rather than a suite-wide name.

**8. `destructiveHint` — signed off the other way.** `writeDestructive` is removed; appends annotate `destructiveHint: false`, with the rationale recorded in `annotations.ts`. This resolves the internal inconsistency the review flagged against `add_step` / `add_choice`.

**9. Comment hygiene — done.** The stale `mutation-tools.ts` header is accurate again (and true, now that `superRefine` is gone). All three restating comment classes removed. `optContext` deleted from code and docs as an unused feature.

**10. Nits — mostly resolved.** `edit-block` with `opts:{}` now returns a preflight missing-required `id`, and `NO_CHANGES`-via-typo is unreachable behind `UNSUPPORTED_PARAMETER`. Typo fixed; the divergent `authoring_start` JSON example aligned; duplicate `describe` names uniquified. **Left:** untrimmed addressing on dispatch (benign, as noted).

**Process notes.** PR body rewritten with the real blast radius (ten shipped names deleted, `explicitId` → `id`, five withdrawn capabilities) and a `CHANGELOG` `### Changed` entry added, including the `get_schema` default change. `ai-subsystem`'s `load_code` now covers `src/cli/mcp/lib/server-instructions.ts`, `agent-routing.ts` and `tools/**`, so prose-only MCP changes get that lens.

## Known follow-ups — and honest limits of this approach

The projection makes the CLI the source of truth for *content* fields, and that part works: add a field to a block's Zod schema and `pathfinder_help` publishes it, agents can send it, and no MCP code changes. Three things fall short of the set-and-forget goal, and they are worth stating plainly rather than discovering later.

**1. Parameter binding is still explicit, so structural fields do not flow through.** `MUTATION_RUNNERS` receives the validated bag but still hand-lifts fields into named runner arguments — `type`, `parent` → `parentId`, `branch`, `ifAbsent`, `id` → `explicitId` for `add-block`; `id` for `edit-block` / `remove-block`; `cascade`; `parent` → `parentId` for `add-step` / `add-choice`. Consequences:

- Adding or renaming an addressing or structural flag on a CLI command does **not** reach agents automatically. It lands in `flagValues`, and unless a binding is added it is either dropped or arrives as `undefined` at a runner that expected it. The passive-freshness property holds only for schema-projected content fields.
- Three vocabularies coexist for the same value: the published parameter (`id`), the Commander flag (`--id`), and the runner argument (`explicitId`). `explicitId` survives at the runner layer even though nothing on the wire uses that name.
- `STRUCTURAL_SKIP_FIELDS` and `parseOptionValues` exist specifically to keep structural keys *out* of the schema projection, so today the schema cannot describe them even in principle.

**2. Deriving the interface from Commander is a projection, not a contract.** We infer `valueType` from option shape (the `enum` vs `array` precedence had to be pinned in the stability contract because a reorder changed it), synthesize named parameters from positionals, and rely on Commander's `.hideHelp()` to withhold parameters — `edit-block`'s placement flags are excluded only because they are hidden, which makes a *help-presentation* concern load-bearing for *interface* policy. Any Commander-level refactor can move the projection without touching a schema.

**3. Inbound and outbound are not the same contract layer.** Requests are an authored, curated surface: camelCase, blacklisted, preflighted. Responses are the CLI's raw `CommandOutcome`, which speaks flags (`--cascade`) and sometimes names CLI-only verbs (`move-block`). We bridge that with prose in `interfaceContract` plus exactly one hand-written adapter (`adaptContainerHasChildren`), and that adapter is the tell: every CLI message that names a flag or an unreachable command needs either its own adapter or agent-side interpretation. Maintenance is proportional to the number of such messages, which is why review finding 7 is documented as by-design rather than fixed. Agents currently work, but they are reading two vocabularies and trusting instructions to reconcile them.

**Ideal fix: generic opt-bag ingestion in the CLI.** Runners take a single validated bag and the Zod schema becomes the sole authority for *all* parameters, structural ones included. That collapses the binding table to a name → runner map, lets the help interface be generated from the schema rather than parsed out of Commander help (removing the inference fragility and making `--help --format json` a consumer of the same source), and allows outcome messages to be authored once in parameter vocabulary, retiring the adapters and closing the inbound/outbound asymmetry.

**Effort.** Eleven runners exist; five already accept `flagValues`, so the shape is half-established. Mechanically it is: collapse each runner's named args into the bag, move structural and addressing keys into the Zod schemas, update the Commander action handlers, delete the MCP binding table, and absorb the test churn — roughly the eleven command files, the `schema-options` bridge, the tool bindings, and their tests. Most of it is mechanical. The real design work is admitting structural fields into the schemas, since the bridge currently strips them on purpose; that decision should be made deliberately rather than as a side effect of the refactor.

**Current impact if we ship as-is.** Content fields stay fresh automatically. Structural changes need a two-site edit (runner signature + MCP binding) and fail quietly if forgotten. Agents see CLI vocabulary in messages and rely on documented interpretation rules. This is materially better than main — where agents received five capabilities that were silently dropped and were told to transcribe flag names that did not exist as keys — but it is a staging post, not the finished contract.

**Also deferred.** `serverInfo.version` still advertises the guide schema version, so both catalogs are byte-identical on `initialize` and a rollback cannot be confirmed from outside. `pathfinder_help` still resolves CLI commands that have no MCP binding rather than rejecting them with `UNKNOWN_COMMAND`. The `set-manifest` provenance fix rides with the opt-bag refactor, since it changes a runner signature.

moxious (dismissed):
## LGTM

Re-verified all four blocking findings against a live in-memory MCP client on `4cba802e` rather than reading the diff. All four are closed, and three of them by removing the mechanism rather than patching it — which is the better outcome:

1. **kebab/camelCase** — `opts:{'auto-collapse':true}` now returns `UNSUPPORTED_PARAMETER` with an authoritative `exposed` list; `autoCollapse` stores correctly. The silent-drop path is gone.
2. **Withdrawn capabilities** — `before`/`after`/`position`/`orphanChildren` all hard-reject. `ifAbsent` is restored and genuinely idempotent (identical re-add → `ok`, count stays 1), and `DUPLICATE_ID` guards the non-`ifAbsent` retry, which is stronger than I asked for. `CONTAINER_HAS_CHILDREN` names `cascade: true`, states the destructiveness, and gives a keep-the-children recipe.
3. **Container-id gate** — deleted on the MCP side; `id` in `opts` works, missing `id` yields `CONTAINER_REQUIRES_ID` from the CLI. One predicate, nothing left to drift.
4. **Contract anchor** — `CONCERNS.md:127` records the operation/`opts` shape, append-only child members, the content-vs-argument-shape split, the withheld-placement policy, and the honest "unless a binding names runner fields" caveat. Keywords cleaned, prose corrected rather than merely anchored, `phases/*` left frozen.

Also confirmed `session-store.ts`, `transports/http.ts` and `utils/etag.ts` are still zero-diff, so ETag integrity and the transport caps remain untouched.

The follow-up note is the strongest part of the submission. I checked its claims against the code and found none that the code contradicts, including the "not done" admissions. The three stated limits are real and correctly diagnosed, and I agree this should ship as a staging post rather than waiting on the opt-bag refactor — the anchor documents the seam.

### Still to do before merge: the conflict

The only conflict is `docs/design/CONCERNS.md`, on the contract-anchor table. **It has a trap** — I resolved the same table on #1639's sibling (#1662) earlier today. Main gained a sentence in the `completion-records` row via #1660 (the three 404-disarm triggers), so resolving in favour of your side wholesale silently deletes it. Keep main's row content verbatim and add `mcp-authoring-server` on top, then let prettier re-pad the column widths.

### Follow-ups, not merge blockers

Filing these rather than holding the PR; the prose is now correct in each case, which is what drops them below the bar findings 1 and 2 cleared.

- **The silent-drop channel still exists one level up.** The preflight guards inside `opts`; the top level is still a non-strict `z.object` that strips unknowns:
  ```
  opts:{type:'section',id:'s1',title:'T'} + top-level autoCollapse:true, authorNote:'note'
    → status:"ok", stored {type,id,title,blocks}   ← both attributes GONE
  ```
  Same signature as finding 1, triggered by wrong nesting instead of wrong casing. `pathfinder_inspect` with a top-level `block` likewise returns output byte-identical to passing nothing. Rejecting known parameter names seen at the top level (naming `opts`) would close it with the machinery you already have.
- **`pathfinder_help` answers for commands no MCP tool implements.** The gate exists (`finalize`/`launch` → `UNKNOWN_COMMAND`) but is incomplete: `help({command:'validate'})` publishes `package`/`packages`/`files`/`strict` while `pathfinder_validate` accepts only `artifact`/`sessionToken` and ignores them silently — and `validate.ts:6` already says agents shouldn't use help for it. `move-block` and `rename-id` also resolve with no tool behind them, and `move-block` is where the error messages redirect.
- **The projection is fail-open with no exposure golden.** A sweep of all nine registered commands found no filesystem parameter reachable today, but nothing pins those surfaces, so a new Commander flag becomes agent-reachable on a public endpoint with no test failing. A guard asserting no exposed name matches `/dir|path|file|cwd|out/` across every registered command is cheap insurance.
- **Undisclosed in the note:** `sanitizeOutcomeForMcp` strips `hints` from every response. All five CLI hint producers emit pure argv recipes, so it's correct today, but a future non-argv hint would be dropped silently.
- **Nit:** 14 comment/`it()` occurrences of `add_block`/`edit_block`/`remove_block`/`set_manifest` remain in `src/cli/mcp/__tests__/` after your sweep of nine.

Approving on the strength of the four closures. Clear the `CONCERNS.md` conflict and this is good to merge.

moxious (approved):

PR #1642, opened 2026-08-18
fix(block-editor): stop Challenge creating empty markdown in branches

### LINKED ISSUE #1542: BranchBlocksEditor silently creates an empty markdown block when Challenge is picked
Picking **Challenge** in the conditional-branch _inline_ picker silently produces an empty markdown block. No error, no toast.

`ALLOWED_BRANCH_BLOCK_TYPES` (`src/components/block-editor/forms/BranchBlocksEditor.tsx:298`) is `BLOCK_TYPE_ORDER` minus the container types, so it includes `challenge`. But `createDefaultBlock` (`:267-295`) has no `challenge` case and its `default:` returns `{ type: 'markdown', content: '' }`, reached via `buildBlockFromForm`'s `default:` (`:487-488`) from `handleAddBlock` (`:525-531`).

## Reproduce

1. Add a conditional block, edit it.
2. In a branch, open the inline "Block type" combobox and select **Challenge**.
3. Fill the form, click Add.
4. An empty markdown block appears.

## Fix options

Either add a `challenge` case to `createDefaultBlock`, or pass `addableBlockTypes` excluding the types the inline editor cannot build. The second is more robust — `default:`-swallows-the-case is the exact failure class this list keeps reintroducing.

## Note

Pre-existing, not introduced by #1531 — `challenge` was already in the pre-PR hand-written `BLOCK_TYPE_ORDER`, so it was already in `ALLOWED_BRANCH_BLOCK_TYPES`. Natural sibling to #1523: both are hand-maintained per-block-type registries where a missing key is legal and surfaces as wrong runtime behavior rather than a compile error. The separate `BlockPalette` in `ConditionalBranches.tsx:372` routes through the real form and is fine.

Refs #1531, #1523

amarkdotdev:
## Summary
- Restricts the inline branch block picker to types `createDefaultBlock` / the inline form can construct, so Challenge (and other non-inline types) cannot be selected.
- Adds a defensive `challenge` case in `createDefaultBlock` so a missing switch arm cannot silently become empty markdown.
- Adds unit coverage for the creatable-type list and challenge default.

Fixes #1542

## Test plan
- [ ] In a conditional branch inline picker, Challenge is not offered
- [ ] Adding markdown/interactive/image/video/input still works
- [ ] `BranchBlocksEditor` unit tests pass

### DIFF
--- a/src/components/block-editor/forms/BranchBlocksEditor.test.ts
+++ b/src/components/block-editor/forms/BranchBlocksEditor.test.ts
@@ -1,0 +1,21 @@
+import { ALLOWED_BRANCH_BLOCK_TYPES, createDefaultBlock } from './BranchBlocksEditor';
+
+describe('BranchBlocksEditor createDefaultBlock', () => {
+  it('does not offer challenge in the inline add picker', () => {
+    expect(ALLOWED_BRANCH_BLOCK_TYPES).not.toContain('challenge');
+  });
+
+  it('builds a challenge block instead of empty markdown if challenged', () => {
+    expect(createDefaultBlock('challenge')).toEqual({
+      type: 'challenge',
+      title: '',
+      brief: '',
+      successCriteria: '',
+    });
+  });
+
+  it('still defaults unknown types to empty markdown', () => {
+    // html is legacy / palette-excluded and not in the creatable list
+    expect(createDefaultBlock('html' as any)).toEqual({ type: 'markdown', content: '' });
+  });
+});
--- a/src/components/block-editor/forms/BranchBlocksEditor.tsx
+++ b/src/components/block-editor/forms/BranchBlocksEditor.tsx
@@ -264,7 +264,7 @@ const getStyles = (theme: GrafanaTheme2) => ({
 /**
  * Create a default block of a given type
  */
-function createDefaultBlock(type: BlockType): JsonBlock {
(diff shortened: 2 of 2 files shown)


Jayclifford345:
@cursoragent

Jayclifford345:
Hi @amarkdotdev, thank you very much for the submission. It seems there are some typecheck issues to sort out first 
```
src/components/block-editor/forms/BranchBlocksEditor.tsx:40:31 - error TS6133: 'BLOCK_TYPE_ORDER' is declared but its value is never read.

40 import { BLOCK_TYPE_METADATA, BLOCK_TYPE_ORDER, INTERACTIVE_ACTIONS } from '../constants';
```

Then we can review and get this merged

Jayclifford345 (changes requested):
See above message

amarkdotdev:
removed the unused import

moxious (approved):
Thanks for the contribution!  I've verified that Jay's requested changes were addressed. In my follow-up review I found a different issue that isn't in the scope of the PR, which I've filed as a separate issue that shouldn't block this PR's merge. https://github.com/grafana/grafana-pathfinder-app/issues/1655.  This one LGTM

Jayclifford345 (approved):

PR #1645, opened 2026-08-18
fix(coda): stop reporting an absent Coda plugin as a fault

Jayclifford345:
## Summary

On a Grafana without `grafana-coda-app`, every Pathfinder panel render produced:

```
nativeSdkLoader.js:64  GET /api/plugins/grafana-coda-app/settings 404 (Not Found)
logging.ts:158  PluginSettings: getting legacy plugin settings failed  {pluginId: 'grafana-coda-app'}
logging.ts:158  isAppPluginEnabled: failed because of unknown reason   {pluginId: 'grafana-coda-app'}
                Caused by: Error: Unknown Plugin
```

The Coda plugin is **optional** and deliberately not a declared dependency, so its absence is the normal case — not a fault. Two independent causes, both fixed here.

**The probe ran when nobody asked.** `enableCodaTerminal` defaults to `false`, yet `docs-panel.tsx` called `useCodaPluginAvailable()` unconditionally, and `useCodaTerminalGate()` called the availability hook *before* it read `enabled`. So the default configuration probed for a plugin whose feature was switched off. `.cursor/rules/coda.mdc` already documented that `docs-panel.tsx` needs all three gates — the code had drifted from its own contract. `useCodaPluginAvailability(shouldProbe)` now takes the caller's own gate.

**The probe asked the wrong question.** `isAppPluginEnabled` fetches the plugin's *settings*, which is what 404s and what core logs. `@grafana/runtime` also exports `isAppPluginInstalled`, which answers "is it there" from boot data (`config.apps`) with no HTTP request. It runs first, and only a definite "not installed" short-circuits.

## What to look at

- `src/integrations/coda/useCodaAvailability.hook.ts` — the two changes. The `typeof isAppPluginInstalled === 'function'` branch is deliberately **synchronous**: an `await` there would defer `capabilitiesOnce()` past `loadCodaCapabilities()` and cost the pre-13.1 path a second `/capabilities` request (the existing "spends one capabilities request for both answers" test pins this).
- Failure is not absence. A missing helper (core < 13.1), a rejection, or `true` all fall through to `isAppPluginEnabled` — `isAppPluginInstalled` shares the 13.1 floor and does not itself check *enabled*, so semantics are unchanged; only the absent case gets cheaper.
- `src/components/docs-panel/docs-panel.tsx` — one line, restoring the documented three-gate contract. Its only consumer is already `{isDevMode && pluginConfig.enableCodaTerminal && codaAvailable && <TerminalPanel/>}`.

## Why this mattered

Nothing user-facing was broken: core defaults `showErrorAlert: false` so there was no toast, and Pathfinder's Faro is `isolate: true` so none of it reached Pathfinder telemetry. What it did was push up to three `faro.api.pushError` calls **into Grafana core's own JS error tracking** on every page load of every Cloud stack without Coda, plus two `console.error` lines. Core also does not negative-cache the failure — only Pathfinder's own module-level cache kept it to once per page load.

Deliberately unchanged: the learner-facing "the Coda app plugin is not installed or not enabled in this Grafana" message on sandbox blocks, and the `pathfinder_sandbox_unavailable` telemetry event. Both are intended signals, scoped to blocks that actually need a VM.

## Contract evolution

This **extends** the `coda-terminal` contract anchor in `docs/design/CONCERNS.md`. It previously read "Availability is resolved at runtime, never declared as a plugin dependency"; it now also states that resolving it costs nothing when the operator has not enabled the terminal, and never reports an absent plugin as a fault. Updated in this PR, along with `.cursor/rules/coda.mdc` and `docs/developer/CODA.md`.

## How to verify

```bash
npm run check        # passes
```

Six new tests in `src/integrations/coda/useCodaAvailability.hook.test.ts`, including the regression guards that assert the *absence* of the settings fetch — the verdict alone would not catch its return:

- boot data says not installed → `false`, and `isAppPluginEnabled` was never called
- the installed probe rejects → still asks `isAppPluginEnabled`
- a core exporting only `isAppPluginEnabled` → still probes
- `enableCodaTerminal: false` → `'disabled'`, and **neither** runtime probe was called (no test asserted this before, which is what let the drift in)

Manually, on a Grafana without the Coda overlay mounted:

1. Open Pathfinder with the terminal toggle off (the default) — no request to `/api/plugins/grafana-coda-app/*`, none of the four console lines above.
2. Turn on "Enable Coda terminal" in Pathfinder's configuration — still no `/settings` request; `CodaBackendStatus` reports the plugin as absent.
3. Open a guide with a `challenge` block — the learner still sees the plugin-missing explanation.

## Breaking changes

None. `shouldProbe` defaults to `true`, so any caller not passing a gate behaves exactly as before.

## Reversibility

Fully reversible — three source edits, no schema, storage, or API change.


src/integrations/coda/useCodaAvailability.hook.ts
```diff
@@ -75,18 +91,22 @@ export function resetCodaAvailabilityCache(): void {
...
+ * is false, because then it was never asked.
  *
  * Distinct from the boolean below: a caller that renders "Coda is unavailable"
  * needs to know the difference between "not installed" and "not asked yet", or
  * it flashes the failure on every mount.
  */
 export type CodaPluginAvailability = 'checking' | 'available' | 'unavailable';
 
-export function useCodaPluginAvailability(): CodaPluginAvailability {
+export function useCodaPluginAvailability(shouldProbe = true): CodaPluginAvailability {
```
tomglenn: The diagnosis in the PR body is that the gate lived at the call site and the call site drifted from it. `shouldProbe = true` keeps exactly that shape: the next caller who writes `useCodaPluginAvailable()` gets the probe back, and nothing fails — same silent 404, same absent test.

There are only two call sites and both now pass a gate, so a required parameter costs nothing and makes the gate impossible to forget. `useCodaPluginAvailable` is re-exported from `src/integrations/coda/index.ts:20`, so this signature is what future callers see first.

```suggestion
export function useCodaPluginAvailability(shouldProbe: boolean): CodaPluginAvailability {
```

(and the same on `useCodaPluginAvailable` below)

The PR body lists "`shouldProbe` defaults to `true`, so any caller not passing a gate behaves exactly as before" under breaking changes — but there is no external caller to preserve, only the two in this repo.

src/integrations/coda/useCodaAvailability.hook.ts
```diff
@@ -75,18 +91,22 @@ export function resetCodaAvailabilityCache(): void {
...
   const [availability, setAvailability] = useState<CodaPluginAvailability>('checking');
 
   useEffect(() => {
+    if (!shouldProbe) {
```
tomglenn, line 107: Minor: on a `shouldProbe` true → false transition this returns early without resetting `availability`, so the hook keeps reporting the last resolved verdict for a probe the caller has since withdrawn.

Both callers mask it today — `useCodaTerminalGate` returns `'disabled'` before it reads `availability`, and `docs-panel.tsx` ANDs the same two flags into the render — so nothing is wrong now. But the docstring above says `checking` means "it was never asked", and after a true → false flip that is the one thing it does not mean. `setAvailability('checking')` before the early return would make the state match its own documentation.

.cursor/rules/coda.mdc
```diff
@@ -91,14 +91,26 @@ is resolved at runtime:
...
+| `useCodaPluginAvailable()` | `isAppPluginInstalled('grafana-coda-app')` from boot data, then `isAppPluginEnabled` — cached per page load |
 | `isDevMode` | terminal panel only |
 
 `docs-panel.tsx` needs all three; `BlockPalette.tsx` needs the first two. `useCodaTerminalGate()`
 bundles the first two as one verdict (`checking` / `disabled` / `plugin-missing` / `configured`) for
 callers that need to *say* which one is unmet. Never assume the plugin is present — every failure path
 must degrade quietly.
 
+**The first gate is enforced before the probe, not just around its consumer.** `enableCodaTerminal`
+is off by default, so an installation that never turned the terminal on must ask nothing about Coda at
```
tomglenn: "an installation that never turned the terminal on must ask nothing about Coda at all" is stated more strongly than the code enforces.

`useCodaSessionEligibility()` is called unconditionally by `challenge-block.tsx:190`, `terminal-step.tsx:143` and `terminal-connect-step.tsx:124`, and the `loadCodaCapabilities()` inside it is not gated on `enableCodaTerminal`. On a stack where Coda *is* installed, opening a guide with a challenge block still spends a `/capabilities` request with the terminal switched off.

That is silent — no 404, no core error tracking — so it is not the bug this PR fixes, and I would not hold the PR for it. But writing an invariant one layer does not hold is how the `docs-panel.tsx` drift got in. Either gate the eligibility probe on the same verdict, or scope the sentence to the availability probe ("must not probe for the plugin") so the rule stays checkable.

docs/developer/CODA.md
```diff
@@ -111,16 +111,25 @@ Timeouts default to 5 s and are capped at 120 s (the cap accommodates `setupScri
...
 
 The block editor palette needs the latter two. `CodaBackendStatus` on the configuration page reports
 which gate is unmet and links to `/plugins/grafana-coda-app`.
 
-**Minimum Grafana version: 13.1 — for the terminal only.** The third gate calls `isAppPluginEnabled`,
+**The probe is gated on the setting, not merely its consumer.** `useCodaPluginAvailability(shouldProbe)`
+takes the caller's own gate — `useCodaTerminalGate()` passes `enableCodaTerminal`, `docs-panel.tsx`
+passes `isDevMode && enableCodaTerminal` — so a default installation asks nothing about Coda. And the
+question it asks when it does probe is "is the plugin installed", answered from boot data by
+`isAppPluginInstalled` with no request; only then does it ask `isAppPluginEnabled`, whose settings fetch
```
tomglenn: "answered from boot data by `isAppPluginInstalled` with no request" is the flag-off path. In `@grafana/runtime@13.1.3`, `getAppPluginMeta` → `initAppPluginMetas` only reads `config.apps` when `FlagKeys.PluginsUseMTPlugins` is off; with it on it awaits `initPluginMetas()`, which is an HTTP LIST against the plugin-meta API.

Nothing about the fix changes — that request is core's, shared, and never `/api/plugins/grafana-coda-app/*`, and Pathfinder's own `cached` promise holds it to once per page load either way. It is just that "no request" is not unconditionally true, and the same phrasing is repeated in `.cursor/rules/coda.mdc:110`. Something like "from core's app-plugin metadata (boot data, unless `pluginsUseMTPlugins` is on)" would survive that flag flipping on.

Separately, two lines below: the 13.1 sentence now reads "`isAppPluginEnabled` (and `isAppPluginInstalled`, which shares that floor), which core's `@grafana/runtime` did not export before 13.1 and which throws synchronously rather than rejecting when absent." The throws-synchronously half is only true of `isAppPluginEnabled` — `isAppPluginInstalled` is guarded by a `typeof` check, not a catch.

tomglenn (commented):
Correct fix, and the two causes are separated cleanly. The `typeof isAppPluginInstalled === 'function'` branch does stay synchronous, so `capabilitiesOnce()` is still assigned before `loadCodaCapabilities()` on the pre-13.1 path — I traced the mixed-core case too (installed probe present, enabled probe absent) and it still spends one `/capabilities` request, because `capabilitiesOnce` is idempotent. The three regression guards that assert the *absence* of the settings fetch are the right shape for this bug.

**Suggestions:**

- `useCodaAvailability.hook.ts:103,124` — `shouldProbe` defaults to `true`, which preserves the shape that caused this bug: a future caller who writes `useCodaPluginAvailable()` re-introduces the 404 silently. There are two call sites and both now pass a gate, so making the parameter required closes the drift permanently. `useCodaPluginAvailable` is barrel-exported from `src/integrations/coda/index.ts`, so this is the shape callers see.
- `.cursor/rules/coda.mdc:103` — the new invariant says an installation that never enabled the terminal "must ask nothing about Coda at all", but `useCodaSessionEligibility()` is called unconditionally by `challenge-block.tsx:190`, `terminal-step.tsx:143` and `terminal-connect-step.tsx:124`, and `loadCodaCapabilities()` is not gated on `enableCodaTerminal`. On a stack that *has* Coda installed, opening a guide with a challenge block still spends a `/capabilities` request with the terminal off. Silent, so not the bug being fixed — but an invariant the code does not hold is how this drift started.
- `CODA.md:127` / `coda.mdc:110` — "from boot data with no request" holds with `pluginsUseMTPlugins` off; with it on, core's `getAppPluginMeta` falls through to `initPluginMetas()` and does a LIST. Does not weaken the fix (still never `/api/plugins/grafana-coda-app/*`, still cached once per page load), just imprecise.
- `useCodaAvailability.hook.ts:107` — when `shouldProbe` goes true → false the hook keeps its last resolved value instead of returning to `checking`. Both callers mask it today; worth a line if the hook is meant to stand alone.

Verified locally against `@grafana/runtime@13.1.3`: `isAppPluginInstalled` is `Promise<boolean>` over `config.apps`, `pluginConfig.enableCodaTerminal` is `Required<>`-typed so `shouldProbe` can never arrive as `undefined` at the `docs-panel.tsx` call site, and `codaAvailable` really does have exactly the one consumer.

Jayclifford345:
Thanks — all four addressed in 8003031d, and the trace on the mixed-core case (installed probe present, enabled probe absent) matches what I had: `capabilitiesOnce` being idempotent is what makes it safe.

**1. `shouldProbe` defaulting to `true`** — agreed, and this was the sharpest of the four: a default that reproduces the bug-causing shape is not a safe default. `shouldProbe` is now required, and so is the new `shouldLoad` below. Both hooks are barrel-exported, so omission was the realistic drift path.

**2. The invariant the code did not hold** — you're right, and this was the more interesting finding. `useCodaSessionEligibility()` ran unconditionally in all three sandbox blocks, and I had also missed `useCodaOptions`, which read the catalogue with a hardcoded `true` from five block-editor call sites. So on a stack with Coda installed and the terminal off, both opening a guide with a challenge block *and* opening the block editor spent a `/capabilities` request.

Rather than thread a gate through eight call sites, eligibility now takes `shouldLoad` (the three blocks pass `codaGate !== 'disabled'`, which they already had in hand) and `useCodaOptions` reads `useCodaTerminalGate()` itself — one edit covering all five form call sites, `useCodaTemplateOptions` included. The invariant now holds rather than being narrowed to fit.

**3. "from boot data with no request"** — corrected in both `CODA.md` and `coda.mdc`. They now say `isAppPluginInstalled` never touches `/api/plugins/grafana-coda-app/*`, from boot data or from core's own plugin LIST under `pluginsUseMTPlugins`, cached once per page load either way.

**4. `true → false` keeping the last verdict** — fixed, though not the way I first wrote it: resetting state in the effect trips `react-hooks/set-state-in-effect`. Both hooks now *derive* the shut-gate value (`return shouldProbe ? availability : 'checking'`), which avoids the extra render as well as the lint error.

One new test: eligibility with a shut gate reads `checking` and calls neither `getCapabilities` nor the installed probe.

---

**On the `13.2.0` e2e leg** — not this PR, and not flake. `main` fails the identical test on the identical leg (run 32120756741, job 95661558561): `welcome-journey.spec.ts:132` › "Step 0 did not complete after clicking Do it", failing all three retries on both branches. `13.2.0` was not in the matrix when this PR first ran — the image resolver picked up the new release between my two pushes, which is why it looks like it arrived with the change.

Root cause is a Grafana 13.2 mega-menu redesign. `MegaMenu.tsx` at `v13.2.0` gained a pinned-items box (`pinnedHeadingId`, drag-reorder edit mode) and a `DOCK_MENU_BUTTON_ID`, and `13.2.0` adds two selectors that did not exist at `v13.1.3`:

```
NavMenu.sectionToggleButton: { '13.2.0': (navUrl) => `data-testid navigation mega-menu section toggle ${navUrl}` }
NavToolbar.quickAddButton:   { '13.2.0': 'data-testid Quick add button' }
```

`data-testid Nav menu item` itself survives, so `NAV_ITEM_SELECTOR` in `navigation-manager.ts:24` still resolves — but step 0 of the welcome journey is `highlight` on `a[data-testid='data-testid Nav menu item'][href='/']` behind a `navmenu-open` requirement, and our nav handling also depends on `#mega-menu-toggle` and on inferring section-expand buttons structurally. Worth a separate issue against `navigation-manager.ts` rather than folding into this diff; happy to open one.

Jayclifford345:
Correction to my note above, after reproducing this against real containers rather than reading the 13.2 source: I was wrong about `#mega-menu-toggle`. It **does** still exist in 13.2 — it only renders while the menu is undocked, which is unchanged behaviour. My earlier reading was taken while the menu was docked, where its absence is expected.

The actual break is narrower and entirely in guide content: **Grafana 13.2 removed Home from the mega-menu**, so the welcome tour's first step resolved to zero elements. Every other target in that guide (`/dashboards`, `/explore`, `/alerting`, `/connections`, `/admin`, `/admin/plugins`) still resolves to exactly 1 element on 13.2.

Fixed separately in #1646, since it is pre-existing on `main` and unrelated to this diff. The wrinkle was that neither selector spans the range — on 12.3 Home is a nav menu item and the breadcrumb sits outside the menu; on 13.2 the nav item is gone and the breadcrumb sits inside it — so it takes a selector list. Verified on 12.3.10, 13.1.3, and 13.2.0.

This PR's own `13.2.0` leg will stay red until #1646 lands.

tomglenn (approved):

PR #1652, opened 2026-08-19
fix: correct two misleading .describe() strings in guide schema

### LINKED ISSUE #1545: Two guide-schema .describe() strings are wrong and reach authors via CLI help

nivedmahendran:
Fixes two incorrect `.describe()` strings that reach authors via CLI help:

1. **snippetId**: Clarify resolution happens after validation and before render, not at parse time
2. **mode**: Clarify `coda` is the runtime default when the field is omitted, not a schema default; the editor seeds `standard`

Closes #1545

### DIFF
--- a/src/types/json-guide.schema.ts
+++ b/src/types/json-guide.schema.ts
@@ -633,7 +633,7 @@ export const JsonChallengeBlockSchema = z.object({
     .enum(['coda', 'standard'])
     .optional()
     .describe(
-      "Execution model. 'standard' runs against the learner's own Grafana — successCriteria is any Pathfinder requirement (e.g. has-dashboard-named:Foo). 'coda' (default) runs in a Coda VM with a terminal — successCriteria is typically coda-exit-zero:<command>."
+      "Execution model. 'standard' runs against the learner's own Grafana — successCriteria is any Pathfinder requirement (e.g. has-dashboard-named:Foo). 'coda' (runtime default when omitted; the editor seeds 'standard') runs in a Coda VM with a terminal — successCriteria is typically coda-exit-zero:<command>."
     ),
   title: z.string().min(1, 'Challenge title is required').describe('Short title shown above the brief'),
   brief: z.string().min(1, 'Challenge brief is required').describe('Markdown problem statement'),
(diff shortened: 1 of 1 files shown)


moxious:
hi @nivedmahendran , it looks like another merge caused a merge conflict. Can you please fix so we can run it through our acceptance pipeline?

moxious:
This would be a welcome PR but I'm closing it now due to 3 weeks of inactivity.  Please feel free to re-open the PR if you want to fix merge conflicts and continue working on it.

PR #1654, opened 2026-08-19
feat(ai-fix): enable AI auto-heal by default

Jayclifford345:
## Summary

Flips `DEFAULT_ENABLE_AI_AUTO_HEAL` from `false` to `true`. The `enableAiAutoHeal` plugin setting stops being a rollout gate and becomes an admin **kill switch** — the shape the telemetry flags already use.

**The Grafana Assistant prerequisite is untouched.** `useAiFixEnabled()` still ANDs the setting with `useIsAssistantAvailable()`, and `AiFixOrchestrator` re-checks both gates inside its request handler. An instance without the Assistant never reaches the AI write path, whatever the setting says — `useIsAssistantAvailable()` starts at `false` and only flips when the real `isAssistantAvailable()` observable emits. So this changes nothing for tenants without the Assistant; it removes a second opt-in for the tenants that already have it.

## What to look at

- [`src/constants.ts:70`](src/constants.ts:70) — the one-line default flip. Everything else derives from this constant: `getConfigWithDefaults`, the config UI initial state, the reset-to-defaults handler, and the dirty check all read it, so there is no second place to keep in sync.
- [`InteractiveFeatures.tsx:368`](src/components/AppConfig/InteractiveFeatures.tsx:368) — the alert shown while the toggle is on was titled "Write path — opt-in" at `severity="warning"`. With the feature on by default, an admin who never touched the toggle would see a warning about a default. Retitled to "Write path" at `severity="info"`; the body text (in-memory mutation, deterministic "Fix this" unaffected) is unchanged. **Push back if you'd rather keep the warning severity** — it's the only judgement call in the diff.
- Both gate tests mocked `getConfigWithDefaults` as `!!jsonData?.enableAiAutoHeal`, which silently disagreed with the real default once flipped. They now mirror it (`?? true`), and the assertions were regrouped so the assistant gate is the one being exercised with an absent flag.

## Why

The feature has shipped behind a per-tenant admin opt-in since #980. The Assistant gate already scopes it to instances that have opted into Grafana Assistant, so the plugin setting was asking the same tenant to opt in twice — which mostly means nobody sees the feature. Keeping the toggle as a kill switch preserves the escape hatch without the discovery cost.

## How to verify

```bash
npm run check
```

Passes clean (exit 0, 449 suites / 7180 tests). The 24 eslint warnings are the pre-existing `getDataSourceSrv` deprecations, unrelated to this diff.

Manually, on an instance with the Assistant available: a step whose selector can't be matched should now show the sparkle "Fix this ✨" button without visiting plugin config first. Turning the toggle off at `/plugins/grafana-pathfinder-app` should make the button disappear entirely, and a dispatched `pathfinder-ai-fix-request` event should be dropped by the orchestrator.

## Breaking changes

None. Tenants that explicitly saved `enableAiAutoHeal: false` keep it — the default only applies when the field is absent from `jsonData`.

## Reversibility

Fully reversible: revert the constant. No storage, schema, or wire-format change; no persisted state is written by this diff. Per-tenant rollback needs no deploy at all — the admin toggle already does it.

## Docs

`docs/developer/AI_FIX.md` carried the line "No PR in this feature flips the default — activation is always an opt-in admin toggle." That statement is what this PR replaces, so its Enablement section is rewritten to state the kill-switch contract and to name Assistant availability as the load-bearing gate. `CONTEXT_INDEX.md`'s one-line summary follows.


docs/developer/AI_FIX.md
```diff
@@ -21,11 +21,11 @@ The whole flow is **dark by default** — see [Enablement](#enablement).
 The feature is gated by **two conditions that must both hold**, expressed by `useAiFixEnabled()` (`src/integrations/assistant-integration/use-ai-fix-enabled.ts`):
 
 1. **Assistant available** — `useIsAssistantAvailable()` (the Grafana Assistant is reachable in this instance).
-2. **Admin opt-in** — the `enableAiAutoHeal` plugin setting is on.
+2. **Admin has not opted out** — the `enableAiAutoHeal` plugin setting is on.
```
Spidy88: Is this accurate? We're just changing the default value, not the meaning of the value, right?

Spidy88: I would have left this alone. It was clearer before the double negation. The purpose of the value hasn't changed, just the default so this would be fine left as-is.

docs/developer/CONTEXT_INDEX.md
```diff
@@ -26,7 +26,7 @@ Load these files **only when working in the relevant domain**.
...
-| `AI_FIX.md`                   | AI auto-heal ("Fix this") flow for failing interactive steps — event contract, patch schema, confidence gate, `enableAiAutoHeal` opt-in                                                                                                                     | `src/integrations/assistant-integration/ai-fix-*`, `src/components/docs-panel/AiFixOrchestrator.tsx`, `src/components/interactive-tutorial/ai-fix-button.tsx` |
+| `AI_FIX.md`                   | AI auto-heal ("Fix this") flow for failing interactive steps — event contract, patch schema, confidence gate, `enableAiAutoHeal` kill switch                                                                                                                | `src/integrations/assistant-integration/ai-fix-*`, `src/components/docs-panel/AiFixOrchestrator.tsx`, `src/components/interactive-tutorial/ai-fix-button.tsx` |
```
Spidy88: nit: This could be confused as TRUE => KILLED

Spidy88 (dismissed):
LGTM - A few items caught my attention initially but then I realized I was misinterpreting them. I made a note in case there's a better suggestion so others (or AI) don't go the wrong direction with it

moxious (approved):

PR #1660, opened 2026-08-19
fix(completion-records): serve unverifiable write identity as a structural 404

moxious:
## Summary

**This is a recovery PR. It carries no new design.**

Four changes were reviewed, tested and approved on #1433, but never reached `main`: #1433 was squash-merged from an earlier head. The squash of `79238008` landed as `d92385f2`, which left the two post-review commits behind, and the branch was deleted on merge.

This PR reproduces those four changes against current `main`. Everything here was already approved on #1433.

## The four items

**1. The identity gate returns the structural 404 (the approved behaviour).**

`pkg/plugin/completion_records_write.go` folded `identityUnverifiable` into the `default:` arm, so it answered **401**. The write contract defines 401 as *transient* — the client retries it after re-auth — so a stack that can **never** verify identity (no app URL, hence no signing-keys URL resolvable) retried every queued write until the 30-day retention horizon and never disarmed.

It now returns the structural **404** carrying `reasonIdentityUnverifiable`: the session disarms while **retaining** the queued records for a later drain, re-arming on a later app load. This matches the read path, which already treats the status as a standing condition (soft-200 `capability=false`). The distinct reason string keeps the envelope diagnosable apart from `obo-unavailable`.

A four-arm table test pins `verified` / `rejected` / `unverifiable` / `signing-keys-down` so a later edit cannot quietly collapse them again.

**2. The 403 prose correction.**

`main` currently contradicts itself: the handler's header comment says a 403 disarms and **keeps** the record, while the test prose asserts it is terminal and **drops** it. The disarm-and-keep behaviour is correct and unchanged here — the stale test prose is what was wrong. This exact mismatch already caused a reviewer to mis-triage the path as silent data loss once, so it is worth correcting on its own merits.

**3. Cache observability counters.**

In `pkg/plugin/completion_records.go`, the generation-mismatch early return (a refresh fenced off by a concurrent write) served its waiters without incrementing `refreshes` / `refreshFailures` / `staleServes`. The §9 per-namespace vital signs therefore stopped reconciling against `misses` exactly when the write path is busiest — the condition an operator would be diagnosing. Includes the assertion that reads the namespace stats under lock.

**4. Documentation prose.**

The write-path taxonomy now agrees across `docs/design/BACKEND_PROXY_PATTERN.md` (§3 and §11), `.cursor/rules/systemPatterns.mdc`, `docs/design/CONCERNS.md`, and the handler's header contract block: the 404 bullet listed two ways in and needs three. Also in the same prose: `systemPatterns.mdc`'s route inventory was missing the `POST /completion-records` route; `coda.mdc` and `docs/developer/CODA.md` still said "App Platform read proxy", which stopped being true when the write path landed; and the write spec's comment pointed at a `pkg/generated/completionrecord` that does not exist in this repo (the generated spec it tracks is pathfinder-backend's).

## How to verify it is a faithful recovery

The diff is **173 insertions / 36 deletions over 11 files**, matching the union of the two lost commits.

A clean cherry-pick was not assumed, since `main` has advanced. Each file was compared against the squash base `79238008`: **9 of 11 were byte-identical to current `main`**, so those hunks applied unchanged and their resulting blobs are byte-identical to the lost commits. The 2 that differed — `systemPatterns.mdc` and `CONCERNS.md`, the two files #1434 also edited — were reproduced onto main's current text by unique-match replacement, and every line added in them appears verbatim in the lost commits' content. No #1434 work is reverted or reintroduced.

Item 1 carries a negative control: removing the `identityUnverifiable` arm makes the new table test fail with `status = 401, want 404`, confirming the test pins the behaviour rather than passing vacuously.

## Deliberately out of scope

- **The `completion-records` routing-table row in `CONCERNS.md`.** The anchor row is inert because `completion-records` is not a routing-table concern, so `extractConcernPaths` throws. That question was escalated and closed as moot. It is not a coverage hole in the meantime: the always-on reversibility-and-one-way-door concern already triggers on `pkg/**` and `src/completion-records/**`.
- **Anything from #1434's frontend work**, which landed separately and is already on `main`.
- Three cosmetic doc observations surfaced during validation and left for follow-up, to keep the diff equal to the recovered commits: `pkg/plugin/app.go:27` still says "read proxy"; `completion_records_client.go:79` cites a bare `kinds/completionrecord.cue` that does not exist in this repo; and the write-path taxonomy lives as four prose copies that must move in lockstep.

## Checks

`go build ./...`, `go vet ./...`, `golangci-lint run ./...` (0 issues), and `go test -race ./pkg/...` are clean, plus the docs-sensitive `src/validation` suites (architecture, control-bytes, skill-references).

## Reversibility

Low risk to revert. Item 1 changes one response status on a path that is currently unreachable-by-design on any stack that can verify identity; items 2 and 4 are prose only; item 3 adds counters without altering cache control flow.


leslie-heinzen (approved):
Approved.

Review scope: authored PR diff from the main parent to HEAD. No blocking findings remain after the concern reviews and three-skeptic verification passes.

The identity-unverifiable 404 path matches the documented completion-write contract, and the cache-generation fencing preserves the stale fallback without repopulating superseded data.

PR #1661, opened 2026-08-19
feat(cli): stamp manifests with a canonical block count via build-stats

moxious:
## Summary

Two things, plus one authorized third:

1. **`src/lib/guide-stats/`** — the canonical block-count / completion-denominator computation, as a pure schema-driven module at tier 1.
2. **`pathfinder-cli build-stats`** — a thin build-time command that stamps each package's `manifest.json` with the computed stats.
3. The `completion-records` concern routing rows in `docs/design/CONCERNS.md` — authorized consolidation, not scope creep; see [Scope notes carried forward](#scope-notes-carried-forward).

## Why

The 2026-08-19 completion meeting settled how completion is computed, and left the arithmetic scattered: there was no single implementation, and the prior build-time script in the content repo disagreed with the decision in two places. The standing architectural requirement is that the counting algorithm live at a low layer, so the CLI, an upload script, the plugin frontend, and a future Go port all inherit one implementation — with the CLI holding only argument parsing and invocation. Computing the numbers at build time also means no author ever asserts them, so they cannot be wrong.

The rule, as implemented:

- The denominator is a guide's **total block count, excluding containers**. A section holding five blocks contributes five, not six.
- `multistep`, `guided`, `conditional`, and `snippet-ref` each count as **exactly one** block; the traversal never enters them.
- Progress is the **furthest position with clear evidence of completion** — a "Do it" click, or a "Mark as complete" click on a section or the whole guide. Nothing softer counts; scroll position was explicitly rejected as a signal.
- Completion is position-based and **monotonic**: reaching position `n` counts `1..n-1`, so non-interactive preamble is never individually completable and never needs to be.
- Completion is **`n / totalBlockCount` and nothing else**. There is no interactive special case: a "Do it" yields 100% only when it is the guide's final counted block.

The module publishes the total **and** the per-block positions together. That pairing is the point — a bare integer is the failure mode, because a consumer that recomputes positions can drift from the denominator it is measured against.

Nothing consumes the module at runtime yet. This PR establishes the computation and the build-time writer; the frontend and backend consumers are follow-ons.

## What Changed

[...]

PR #1662, opened 2026-08-19
feat(cli): forward unknown top-level manifest keys into repository.json

moxious:
## What & why

The packaging CLI now forwards **any unknown top-level key** in a package's
`manifest.json` through into that package's entry in `repository.json`.

Package manifests want to carry metadata the CLI has no opinion about. Until
now every such field cost a CLI change and a release before it could reach
`repository.json`, because two independent layers dropped it. This makes the
manifest's top level deliberately open instead, so adding a field is a
content-repo change alone.

The alternative considered and rejected was a single namespaced extension field
holding arbitrary content. A flat passthrough was preferred because more fields
are expected, and each one should not cost a code change here.

This is the carriage half of the completion-denominator work. The computation
that produces the first such field — rollup step/block/section counts — is being
built separately and is a distinct concern from forwarding it.

## History

This supersedes **#1393**, now closed. That PR carried `stats` as one more
explicitly-named field under a spike framing gated on an RFC. It was replaced for
two reasons: the single-named-field shape was superseded by this generic
passthrough, and its branch became unrecoverable through the normal validation
path without a force-push that its review history did not allow. This branch is a
clean start off `main`; the content was rebuilt rather than carried over.

## What was blocking it

Two layers dropped unknown keys, and both had to move:

1. **Parse.** `ManifestJsonObjectSchema` was a plain `z.object`, which strips
   unknown keys — so a new field never survived to be copied. Verified against
   the pinned Zod (**4.4.3**) rather than assumed:

   | schema | `parse({ a: 'x', extra: 1 })` |
   | --- | --- |
   | `z.object({ a: z.string() })` | `{"a":"x"}` |
   | `z.looseObject({ a: z.string() })` | `{"a":"x","extra":1}` |

   Known-field validation is identical either way — `{ a: 123 }` still fails
   under `looseObject`.

2. **Copy.** `readPackage()` in `build-repository.ts` copied manifest fields onto
   the entry one at a time by name, which is an allowlist by construction:
   anything not named was never copied.

## Changes

- `ManifestJsonObjectSchema` and `RepositoryEntrySchema` are now
  `z.looseObject`. The entry schema had to move too, or extension keys would be
  stripped again the moment anything re-parses `repository.json`.
- `build-repository` gained `forwardExtensionFields()`, copying every manifest
  key not named explicitly, and reports the keys it forwarded:

  ```
  ℹ️  demo-guide: forwarding 2 extension field(s): stats, estimatedMinutes
  ```

  That line is the reason the open namespace is safe to live with. This change
  removes both checks that used to catch a misspelled manifest field — the
  compile-time excess-property check (see the review note below) and the
  build-time drop — so `startingLocaton` now ships as a plausible-looking
  extension field while the real key never lands. The build log is where that
  becomes findable. It is an info line rather than a warning because the open
  namespace is deliberate, and it goes to **stderr**, so `build-repository`
  piped to `jq` in its stdout mode still emits clean JSON.
- `docs/design/CONCERNS.md` records the contract anchor for
  `guide-schema-and-contracts`, and `package-authoring.md` states that an
  extension key's shape is never checked, so the consumer owns that validation.
- `ManifestJson` and `RepositoryEntry` gained an `[key: string]: unknown` index
  signature so consumers can read extension data.
- `KNOWN_FIELDS._manifest` and its `KnownFieldsMetaKey` entry are **deleted**.
  Opening the manifest top level left that set with no consumer and no purpose
  (`detectUnknownFields` only walks `_guide`, `_step`, `_choice`, and blocks), so
  it is dead code this change itself created. Deleting it was authorized
  in-cycle, deliberately, rather than deferred — leaving it would be net debt
  added by this PR. That authorization is also why a code deletion lands in what
  was otherwise a documentation pass.
- `stats` is **not** special-cased. The named field the previous PR added is
  removed from both schemas and the types; the generic path covers it, and it
  now has no privileged status in the CLI.

## Collision and refusal policy

The refusal list is derived, not hand-maintained:

```
RESERVED_ENTRY_FIELDS = keys(RepositoryEntrySchema.shape)
                        - keys(ManifestJsonObjectSchema.shape)
                        + '__proto__'
```

which today evaluates to **`path`, `title`, `__proto__`**, and stays correct on
its own if either schema gains a field.

- **Fields the build computes** (`path` from the package directory, `title` from
  `content.json`) are refused with a warning, never overwritten. Without this a
  manifest could rewrite the `path` the resolver fetches by.
- **`__proto__`** is refused as defence in depth on an assignment sink fed by
  untrusted file content. Zod 4.4.3 already drops it before the copy sees it, so
  the guard is belt-and-braces rather than the thing currently preventing
  pollution.
- **Known manifest fields are never forwarded generically.** They keep their
  existing hand-written mapping, so fields that deliberately do not appear in an
  entry (`id`, `schemaVersion`, `repository`, `language`) still do not. Nothing
  changes for any field that already existed.
- **Future field names are not reserved.** If a forwarded key is later promoted
  to a named field, the passthrough stops handling it and the explicit mapping
  takes over — the key keeps appearing, now validated. That is the intended
  migration path, and why the refusal list stays at computed fields rather than
  growing into a denylist.

## Nested validation is unchanged

Permissiveness is scoped to the manifest's **top level only**. `content.json`
still parses through `ContentJsonSchema` → `JsonBlockSchema` unchanged, and
nested manifest objects (`author`, `targeting`, `testEnvironment`) keep their own
strict schemas. Tests assert a malformed block is still rejected and an invalid
`type` still fails manifest validation.

## Review notes

Two things reviewers should weigh rather than skim:

- **Index signatures cost excess-property checking.** Adding
  `[key: string]: unknown` to `ManifestJson` and `RepositoryEntry` disables
  TypeScript's excess-property check on object literals of those two types
  repo-wide, not just at the passthrough boundary — a `titel:` typo in a
  `RepositoryEntry` literal no longer errors. This is accepted deliberately: the
  alternative pushes casts onto every consumer that reads extension data.
- **The contract anchor row moved rather than being added alongside.**
  `guide-schema-and-contracts` was a pre-contract candidate whose proposed owner
  was the `managed-by` / `source-package` annotation vocabulary — which this
  change does not establish. A concern cannot both hold a recorded anchor and
  remain a candidate, so the candidate row is removed; the annotation hypothesis
  is carried into the anchor as explicitly uncontracted rather than dropped.

## Deliberately out of scope

Named here so the omissions read as decisions rather than oversights:

- **A "did you mean" edit-distance hint in `validate`** for keys within edit
  distance 1 of a named field. This is the stronger version of the typo signal,
  but it belongs in `validate` rather than `build-repository` and wants its own
  tests.
- **A positive schema for the `stats` payload.** That belongs with the consumer
  that reads it, not with the carriage layer that forwards it unopened.
- **The redundant `.loose()` call sites.** Now that the base schema is loose,
  eight `ManifestJsonObjectSchema.loose()` calls are no-ops and a couple of their
  comments mis-describe the code. Real cleanup, but out of scope here.

This change is the small half of a pair, and its job is to unblock the other
half; it should not grow.

## How to verify

`npm run check` passes. Behavioural tests in
`src/validation/build-repository.test.ts` cover: an unknown key survives; a
nested extension object survives; the round-trip through `RepositoryJsonSchema`
keeps it; an entry gains no stray keys when the manifest has none; `path` and
`title` collisions are refused with a warning while the computed values win; a
known field stays validated; and invalid nested content is still rejected.

Three more cover the forwarded-keys line: that it names what was forwarded, that
it stays silent when a manifest carries no extension fields, and the typo case —
a manifest authoring `startingLocaton` ships that key with the author's value
while `startingLocation` is absent from the entry entirely, because
`preserveAuthoredStartingLocation` strips the value the schema defaulted. A test
in the MCP repository-client suite pins that an entry body carrying its own `id`
cannot override the map key.

End-to-end run over a package tree carrying `stats`, `estimatedMinutes`, a nested
`owner` object, and a hostile `path`:

```json
{
  "demo-guide": {
    "path": "demo-guide/",
    "title": "Demo guide",
    "type": "guide",
    "startingLocation": "/",
    "testEnvironment": { "tier": "cloud" },
    "stats": { "steps": 3, "blocks": 7 },
    "estimatedMinutes": 12,
    "owner": { "team": "enablement", "slack": "#pathfinder" }
  }
}
```
```
WARNING: demo-guide: Ignoring manifest field "path": reserved for a repository
entry field the build computes
```

## Breaking changes

None. No `CURRENT_SCHEMA_VERSION` bump — this widens what is accepted and
carried; it does not change the meaning of any existing field. Repository files
built before this change parse identically after it.

## Reversibility

Straightforwardly revertable. Reverting narrows `repository.json` back to the
named fields; a consumer reading an extension key would need to tolerate it
going missing, which is the same tolerance it already needs for a manifest that
omits the key.

## Follow-up

`interactive-tutorials` CI must bump its pinned Pathfinder CLI SHA to pick up
this change, or extension keys will continue to be stripped during repository
builds there.


src/cli/commands/build-repository.ts
```diff
@@ -26,6 +31,45 @@ interface BuildRepositoryOptions {
...
+
+  for (const key of Object.keys(manifest)) {
+    if (NAMED_MANIFEST_FIELDS.has(key)) {
+      continue;
+    }
+    if (RESERVED_ENTRY_FIELDS.has(key)) {
+      warnings.push(`Ignoring manifest field "${key}": reserved for a repository entry field the build computes`);
+      continue;
+    }
+    entry[key] = manifest[key];
```
Jayclifford345, line 74: **Medium — a forwarded key is silent, and the index signature removed the other check that would have caught a typo.**

Only *refused* keys warn (line 64). A forwarded key produces no output at all, so the content repo's build log says nothing about what it just published.

That matters more than it would have before this PR, because the two checks that used to catch a misspelled manifest field both go away here:

1. **Compile time.** Verified against 1662's base vs head with the same probe:

```
# base
probe.ts(3,51): error TS2561: Object literal may only specify known properties,
  but 'startingLocaton' does not exist in type 'ManifestJson'.
  Did you mean to write 'startingLocation'?
# head
(no output)
```

The review note names this ("a `titel:` typo in a `RepositoryEntry` literal no longer errors") and the disclosure is accurate — my probe reproduces exactly that example.

2. **Build time.** A typo'd key was previously dropped, so it simply never appeared. Now it is forwarded verbatim into `repository.json` and shipped to every user in the CDN index.

Neither alone is serious. Together they mean `startingLocaton: "/foo"` in a manifest now silently becomes an extension field that looks like real metadata, and the guide's actual `startingLocation` silently defaults — with no signal at authoring time, compile time, build time, or publish time. That is a new failure mode, not a pre-existing one.

Cheapest mitigation is one line here: collect forwarded keys and emit them as an info line (`forwarding 2 extension field(s): stats, estimatedMinutes`). The intent is a deliberately open namespace, so a *warning* would be wrong — but visibility in the build output is what makes a typo findable, and it costs nothing. `validate` growing a "did you mean" hint for keys within edit distance 1 of a named field would be the stronger version.

Separately, worth stating in `package-authoring.md`: because forwarding is silent and unvalidated, an extension key's *shape* is never checked. A consumer reading `entry.stats.blockCount` gets whatever the author typed, including a string or a nested object. The doc frames extension fields as metadata the CLI has no opinion about, which is right — it should also say the consumer owns validation.

Jayclifford345 (commented):
Reviewed against `docs/design/CONCERNS.md` at the PR base.

**Classification** `contracts-and-schemas` + `docs-only`.

**Activated concerns** always-on (`security`, `correctness-and-reliability`, `testing-and-verification`, `reversibility-and-one-way-door`, `cross-cutting-architecture`) plus `guide-schema-and-contracts`, `cli-and-e2e-runner`, `mcp-authoring-server`, `ai-subsystem`.

**What I verified rather than read** — 77 suites / 1689 tests in `src/cli` + `src/validation` pass at this head; probed the emitted JSON Schema, `__proto__` handling at both the manifest and the map level, the reserved-set derivation, and the excess-property change against base-vs-head types. Details are in the four inline comments (apologies for the four separate reviews — an anchoring problem on my end, the substance is in the comments).

This is a careful PR and most of what I would normally check is already right:

- **`additionalProperties` lands exactly where the docs say.** Verified on this head: manifest root `{}`, each repository entry `{}`, and nested `author` / `targeting` / `testEnvironment` still `false`. The `E2E_TESTING.md` paragraph describing that is accurate to the byte.
- **The `KNOWN_FIELDS._manifest` deletion is genuinely dead-code removal.** No code referenced it — `detectUnknownFields` walks `_guide`, `_step`, `_choice`, and blocks. Its only two live references were docs (`schema-coupling.mdc:132`, `PATHFINDER-PACKAGE-DESIGN.md:554`) and both are updated here. Deleting it in-cycle was the right call; leaving it would have been debt this PR created.
- **Prototype pollution is closed at both levels**, and the reasoning in the comment is correct rather than merely reassuring.
- **The frontend blast radius is small.** The only non-CLI consumers of `RepositoryJson` are `src/package-engine/resolver.ts` and `src/lib/custom-guide-repository-client.ts`, and neither iterates entry keys nor spreads an entry into a DOM, URL, or analytics payload — so loose entries do not become a new sink.

**One cross-PR thing worth deciding explicitly.** This PR and #1661 are the two halves the descriptions say they are, and they interact in both directions:

1. **#1662 should land first.** Without the `looseObject` here, #1661's stamped `stats` key is *destroyed* by every command that round-trips a manifest through the schema — `set-manifest`, `rename-id`, `add-block`, `add-choice`, `add-step`, `edit-block`, `move-block`, `remove-block`, and the MCP `set_manifest` tool, all via `src/cli/utils/package-io/disk.ts:102`. I verified this both ways: with `z.object` the key is gone after a `readPackage`/`writePackage` cycle; flipping the one line to `z.looseObject` makes it survive. The same fix closes #1661's other schema problem — `pathfinder-cli schema manifest` currently emits `additionalProperties: false`, so the published schema rejects every manifest `build-stats` stamps.
2. **Landing this one makes a sentence in #1661 wrong.** Its `CLI_TOOLS.md` says "Today `repository.json` carries no `stats` field, so the ordering is forward-looking rather than load-bearing." Once this merges, `stats` forwards automatically and the `build-stats` → `build-repository` ordering becomes load-bearing. Whichever lands second should fix that paragraph.

Neither is a defect in this PR — but the merge order is a real decision, and #1662-first is strictly better.

**Reversibility** — agreed with the description's read. Reverting narrows the index back to named fields; a consumer of an extension key needs the same tolerance it already needs for a manifest that omits it. The one asymmetry: the index-signature removal of excess-property checking will have masked typos while it was in effect, so a revert surfaces them all at once as new type errors — noisy, not dangerous.

Two findings I would resolve before merge: the missing contract-anchor row, and forwarded-key visibility. Both inline.

src/cli/commands/build-repository.ts
```diff
@@ -26,6 +31,45 @@ interface BuildRepositoryOptions {
...
+const NAMED_MANIFEST_FIELDS: ReadonlySet<string> = new Set(Object.keys(ManifestJsonObjectSchema.shape));
+
+/**
+ * Repository-entry fields the builder computes itself (`path` from the package
+ * directory, `title` from content.json). A manifest key of the same name is
+ * refused rather than allowed to overwrite the computed value. `__proto__` is
+ * belt-and-braces on an assignment sink fed by file content — zod's loose parse
+ * already drops a JSON `__proto__` own-key before the copy runs.
+ */
+const RESERVED_ENTRY_FIELDS: ReadonlySet<string> = new Set([
```
Jayclifford345, line 47: Confirmed the derivation does what the description claims, on this head:

```
NAMED   : author,category,conflicts,depends,description,id,language,milestones,
          provides,recommends,replaces,repository,schemaVersion,startingLocation,
          suggests,targeting,testEnvironment,type
ENTRY   : author,category,conflicts,depends,description,milestones,path,provides,
          recommends,replaces,startingLocation,suggests,targeting,testEnvironment,
          title,type
RESERVED: path,title + __proto__
```

And it is genuinely self-maintaining in both directions: a new entry-only field enters the reserved set automatically, and a field promoted into the manifest schema leaves it automatically because the explicit mapping then owns it. That is the right shape for this — a hand-maintained denylist would have gone stale on the first promotion.

Also verified the `__proto__` reasoning rather than taking it on trust, against the pinned zod 4.4.3:

```
raw own keys:    [ 'a', '__proto__', 'extra' ]   hasOwnProperty('__proto__'): true
looseObject out: [ 'a', 'extra' ]                hasOwnProperty('__proto__'): false
```

So the guard is belt-and-braces exactly as described. Worth noting it is *not* redundant: `Object.keys(raw)` does include `__proto__`, so if zod's behaviour ever changed, `entry[key] = ...` at line 67 would invoke the setter. Keeping it is right.

The map level is covered too — `RepositoryJsonSchema.safeParse({"__proto__":{...},"good":{...}})` yields keys `['good']` and leaves `Object.prototype` untouched.

Jayclifford345 (commented):
Part of the same review — summary is on the first of these four (an anchoring problem on my end split them; the substance is in the inline comment below).

src/types/package.types.ts
```diff
@@ -142,9 +142,15 @@ export interface PackageMetadataFields {
...
  * Manifest file schema — metadata, dependencies, and targeting.
  * Authored by product, enablement, or recommender teams.
+ *
+ * The index signature carries extension metadata: any top-level key not named
+ * below survives parsing and is forwarded into the package's repository entry.
+ *
  * @coupling Zod schema: ManifestJsonSchema in package.schema.ts
  */
 export interface ManifestJson {
+  [key: string]: unknown;
```
Jayclifford345, line 152: **Medium — no contract anchor row, and the review contract does not allow this one to defer.**

The description is admirably direct that this establishes a contract for `guide-schema-and-contracts` and that the omission is deliberate. But `.cursor/skills/review/SKILL.md` §5 is explicit:

> "when the PR itself establishes or replaces a contract (a typed facade, reducer, schema, or lifecycle owner), require the contract anchor in `docs/design/CONCERNS.md` (Contract anchors) to be added or updated in the same PR — an unrecorded contract silently re-fractures. **Do not accept a follow-up-PR deferral for the anchor row** or the concern's routing paths; only prose documentation may defer."

The stated reason — "the row convention requires PR evidence numbers, which this branch did not have while the docs were being written" — no longer holds: this is #1662, and `#1662` is the value the column wants. `docs/design/PATHFINDER-PACKAGE-DESIGN.md` is the right home for the *design*, but the anchor is what the §3b contract-evolution scan reads, so without the row the next PR touching manifest parsing gets contract *reconstruction* from history instead of conformance against a recorded contract.

This one matters more than average because `guide-schema-and-contracts` is currently a **pre-contract candidate** in that table (evidence `#1504`, proposed owner: the `managed-by`/`source-package` annotation vocabulary). So there is a row to move, and #1661 — landing beside this one — adds a named invariant that reaches straight into manifest field semantics. Two PRs establishing manifest-level contracts in the same window, with the anchor table describing neither, is the exact accretion the section exists to prevent.

The contract itself is already written in the description and reads well as an anchor nearly verbatim: manifest top level deliberately open; extension keys forwarded verbatim and unvalidated; the build refuses the entry fields it computes, from a derived reserved set; nested objects stay strict; promotion to a named field is the migration path and keeps the key appearing.

Jayclifford345 (commented):
Part of the same review — summary is on the first of these four (an anchoring problem on my end split them; the substance is in the inline comment below).

src/cli/mcp/lib/repository-client.ts
```diff
@@ -158,7 +158,7 @@ async function doFetchRepositoryIndex(baseUrl: string, now: number): Promise<Rep
   for (const [id, entry] of Object.entries(rawIndex)) {
     const parsed = RepositoryEntrySchema.safeParse(entry);
     if (parsed.success) {
-      packages.push({ id, ...parsed.data });
+      packages.push({ ...parsed.data, id });
```
Jayclifford345, line 161: Good catch, and it is load-bearing rather than cosmetic: with `RepositoryEntrySchema` now loose, an entry body carrying its own `id` key survives the parse, and under the old `{ id, ...parsed.data }` ordering it would have overwritten the authoritative map key. `repository.json` comes from a CDN, so that is attacker-or-typo-controlled input deciding which package an MCP client thinks it is looking at.

Both branches are covered (the fallback at line 172 pins `path` and `id` after the spread), and the two new tests name the behaviour directly. Nothing to change — flagging it because it is the kind of ordering fix that is easy to lose in a later refactor, and the tests are what will hold it.

Jayclifford345 (commented):
Part of the same review — summary is on the first of these four (an anchoring problem on my end split them; the substance is in the inline comment below).

Jayclifford345 (changes requested):
See above

Jayclifford345 (approved):

Issue #1600, opened 2026-08-12
Block editor and MCP writes drop metadata.annotations, detaching script-managed guides from their provenance


moxious:
## Summary

All three block-editor write paths build `metadata` from scratch as `{ name, namespace, resourceVersion }`. A k8s PUT is a whole-object replace, so every write **erases `metadata.annotations`** set by anything else.

This became load-bearing with #1504, which introduces the repo's first use of annotations for provenance: `scripts/upsert-learning-path.sh` stamps `pathfinderbackend.ext.grafana.app/managed-by` and `.../source-package` on everything it uploads, and refuses to overwrite a guide that lacks its `managed-by` mark. One editor click removes that mark.

## Where

`src/components/block-editor/hooks/useBackendGuides.ts` (main @ `9b9ae4d2`) — all three:

- `saveGuide` — `:144-152`
- `publishGuide` — `:208-212`
- `unpublishGuide` — `:252-256`

```ts
const metadata: any = {
  name: resourceName,
  namespace,
  resourceVersion: currentMetadata.resourceVersion,
};
```

`BackendGuide.metadata` (`:15-21`) doesn't model `annotations` either, so the LIST response never carries them into editor state to begin with.

`src/cli/mcp/tools/finalize.ts:151-153` builds the same annotation-free envelope, so the MCP write path has the identical hazard.

For contrast, `scripts/upsert-guide.sh` at #1504 head does this correctly — it merges `($existingAnn + $ann)` on update specifically so a PUT doesn't drop annotations set by another tool.

## Why it matters

The failure is fail-_closed_, so nothing is destroyed — but the recovery path is worse than the problem:

1. Script uploads a path; every resource carries `managed-by`.
2. An author clicks **Unpublish** on one milestone from the guide library. Annotations gone.
3. The next script run's pre-flight sees that name as `unmanaged` and refuses the **entire package** — not just that milestone.
4. The only unblock the tool offers is `--overwrite`, which is run-global. Re-attaching one self-detached guide therefore means disabling collision protection for every name in the package, including genuinely foreign ones.

So a routine unpublish click bricks the automation for a whole learning path, and the documented workaround is to turn off the guard that exists to prevent an unrecoverable overwrite. The operator has no way to distinguish a self-inflicted detachment from a real foreign guide, and neither does `--overwrite`.

## Suggested fix

Carry existing annotations through all three writes:

```ts
const metadata: any = {
  name: resourceName,
  namespace,
[...]

PR #1663, opened 2026-08-20
fix(block-editor): stop editor saves destroying spec.manifest and provenance

### LINKED ISSUE #1599: Block editor save deletes spec.manifest, silently collapsing a learning path into a flat guide
## Summary

`saveGuide` in the block editor rebuilds `spec` from a literal that has no `manifest` key. A k8s PUT is a whole-object replace, so saving a guide that carries `spec.manifest` **deletes** that field. A learning path silently collapses into a flat guide.

Found while reviewing #1504, which adds `scripts/upsert-learning-path.sh` — the tool that produces `spec.manifest`-bearing resources in the first place. Not caused by that PR; the PR is what makes the bug reachable in practice.

## Where

`src/components/block-editor/hooks/useBackendGuides.ts:161-172` (main @ `9b9ae4d2`):

```ts
const k8sResource = {
  apiVersion: APP_PLATFORM_API_VERSION,
  kind: "InteractiveGuide",
  metadata,
  spec: {
    id: exportable.id,
    title: exportable.title,
    schemaVersion: exportable.schemaVersion || "1.0",
    blocks: exportable.blocks,
    status,
  },
};
```

The root cause is one level up: `JsonGuide` (`src/types/json-guide.types.ts:14-23`) has no `manifest` field, so the editor's model structurally cannot round-trip it. `GuideLibraryModal` builds its `JsonGuide` from `{id, title, schemaVersion, blocks}` on load, and `saveGuide` writes exactly those back.

Note the asymmetry — `publishGuide` and `unpublishGuide` spread `...existing.spec` (`useBackendGuides.ts:222`, `:266`) and **do** preserve `manifest`. Only `saveGuide` drops it.

## Why it matters

`spec.manifest` is what makes a path a path — it carries `type` and the ordered `milestones` list. When it disappears:

- `AppPlatformPackageResolver` falls back to an inferred `{ id, type: 'guide', repository: 'app-platform' }` (`src/package-engine/app-platform-resolver.ts:119`).
- The `logger.warn` at `:111` fires only on a **malformed** manifest, not a missing one — so there is no log line, no UI signal, and no error.
- The path's milestone chrome disappears and its members become unreferenced standalone guides.

Recovery is re-running the upload script from the source package, which discards whatever edit the author just made in the editor. The aggregator has no revision verb, so there is nothing to roll back to.

## Repro

1. Upload a path package with `scripts/upsert-learning-path.sh` (from #1504) so the cover page has `spec.manifest`.
2. In the block editor, open that cover page from the guide library.
3. Make any edit and click Update.
4. `GET` the resource — `spec.manifest` is gone. No warning was shown at any point.

## Suggested direction

Thread `manifest` through as an opaque passthrough rather than modelling it: carry it on the editor's loaded-guide state, and spread it back in `saveGuide` the way `publishGuide` already spreads `...existing.spec`. That avoids teaching `JsonGuide` about a field the editor has no reason to understand.

Worth deciding explicitly whether the editor should be able to edit path cover pages at all, or whether they should be read-only in the library until there is real path-authoring support. That is the design question underneath this bug.

Related: the same three write paths also drop `metadata.annotations` — filed separately.

tomglenn:
Closes #1599. Partly addresses #1600 — see "what this does not cover".

## Summary

An App Platform PUT replaces the whole object, but `useBackendGuides` rebuilt `spec` and `metadata` from literals on every write. Two consequences:

- Saving a learning-path cover page **deleted `spec.manifest`**, silently collapsing the path into a flat guide. The resolver then treats it as `type: "guide"` and the milestones stop rendering, with no warning (#1599).
- Every write **erased `metadata.annotations`**, including the provenance marks `scripts/upsert-learning-path.sh` stamps — so a later run of that script refuses the package until you pass `--overwrite` (#1600).

`preservedSpec` now layers the editor-owned spec fields over the spec last read; `preservedMetadata` carries `resourceVersion`, `annotations` and `labels` through. All three writes — `saveGuide`, `publishGuide`, `unpublishGuide` — go through them.

## What to look at

**Where the manifest comes from.** The passthrough reads the LIST snapshot the hook already holds, not a fresh GET at save time. That is a concurrency decision, not a convenience one: `resourceVersion` comes from the same snapshot, and if the passthrough came from a fresh read then the correct `resourceVersion` to send would be the fresh one too — which silently converts the write from optimistic-concurrency into last-write-wins. Reading both from one snapshot keeps the 409 guarding exactly the object whose unowned fields are being replayed.

Adding `manifest` to `JsonGuide` was rejected: the editor has no reason to understand the field, and modelling it would put it in scope for JSON mode, import/export and the schema.

**Inheritance is scoped to the resource the editor's content came from, and backend tracking is what encodes that.** So tracking is set only after a write lands, never before. Two paths depend on this:

- A **name-collision overwrite** keeps the `resourceVersion` guard but inherits nothing. Inheriting there would render the new guide as the old one's path *and* stamp it with a provenance mark the upload script's ownership check trusts — turning #1600's fail-closed condition into fail-open.
- A **failed overwrite stays untracked**, so the retry re-asks rather than silently becoming an ordinary update with full inheritance.

**The confirm handler owns its own error reporting.** It runs after `orchestrateSave`'s `try/catch` has already returned — the `return new Promise` left that scope — so a rejection previously produced no toast and never settled the awaited promise. It now reports and resolves in a `finally`.

**Absence is confirmed against the server before refusing a write.** `saveGuide` fails closed when the named resource is not in the snapshot, matching `publishGuide`/`unpublishGuide`. But an absent entry does not mean an absent resource: a transient LIST failure resolves to an empty list rather than an error (`UNAVAILABLE_STATUSES` in `fetchBackendGuides`), and in that state `hasManageableBackendGuides` hides the guide-library entry — the only affordance that would refresh it. Trusting the snapshot alone therefore turned one failed GET into a permanently unsaveable guide. It probes the list once and only throws if the server cannot confirm the resource either. The probe calls `fetchBackendGuides` directly rather than `refreshGuides`, because `refreshGuides` commits: it clears `error` up front and stores whatever the LIST resolves to, so using it here would launder a live error into a clean empty catalogue and hide the guide-library entry — reinstating the same dead end in the other order. Probing also lets a genuine LIST failure surface as itself rather than as “the resource does not exist”.

**`schemaVersion` normalisation is deliberate** and left as-is, with a comment saying so. `getGuide()` never emits it, so `preservedSpec` forces the editor's own version rather than replaying a stored one.

## Contract evolution

**Extends** the `block-editor-authoring` anchor in `docs/design/CONCERNS.md` with the partial-writer invariant: which fields each write owns, that inheritance is scoped by tracking, that tracking is set only after a write lands, and that absence is server-confirmed before a refusal. Updated in this PR.

## What this does not cover

`src/cli/mcp/tools/finalize.ts` is untouched, so **#1600 only half closes**. It is not a write path — it emits a handoff descriptor for an external MCP client to write, has no backend access, and never reads an existing resource, so there is nothing to preserve from. Making it annotation-safe means either giving the tool backend access or extending its `clientGuidance` payload to tell clients to merge annotations: a new capability or a contract-payload change, not a field spread.

Also deliberately out of scope: `metadata.ownerReferences` and `finalizers` are still dropped (the metadata passthrough is a named allowlist, so server-managed fields like `uid` and `managedFields` are not echoed back), and the overwrite-confirm modal is now the one remaining path that intentionally discards a manifest without telling the author — #1600's own "worth considering separately" note.

## How to verify

```bash
npm run test:ci    # 459 suites, 7341 tests
npm run typecheck
npm run lint
```

Fourteen new tests in `useBackendGuides.test.ts` plus one in `useBackendSaveFlow.test.ts`. Every assertion is made against the outbound request body, so they pin behaviour rather than restating the implementation. Each was mutation-checked — reverting its corresponding fix kills exactly that test:

| Revert | Test that fails |
|---|---|
| the whole passthrough | the five preservation tests |
| `...existingSpec` moved after the owned fields | the two round-trip tests (they pin field *precedence*, not just presence) |
| the foreign-overwrite gate | `does not inherit a foreign manifest when overwriting a name collision` |
| tracking set before the write | `leaves the guide untracked when the overwrite fails, so the next save re-asks` |
| the confirm-handler try/catch | the failed-overwrite test times out, plus 7 cascading unhandled-rejection failures |
| the fail-closed guard | `refuses to update a resource the server cannot confirm either` |
| the refresh-before-refusing | `re-reads the list before refusing, so a transient list failure does not block the save` |
| the probe back to `refreshGuides` | `probes without committing, so a live error and the guide-library entry both survive a refusal` and `surfaces a failed probe as itself rather than as an absent resource` |
| per-field metadata merge | `falls back to the snapshot resourceVersion…` and `publishGuide keeps annotations and resourceVersion when caller metadata is partial` |

Manually, against a stack with a package-format learning path uploaded by `scripts/upsert-learning-path.sh`:

1. Open the path cover page in the editor, save without editing — the path still renders its milestones, and a re-run of the upload script does not refuse.
2. Start a new guide, title it so it slugs to an existing path's name, publish, confirm the overwrite — the new guide is flat and carries no provenance mark.
3. Same as 2 but with the network offline at the confirm step — an error toast appears, and the next save re-asks rather than overwriting silently.

## Documentation

`docs/developer/CUSTOM_GUIDES.md` and `docs/developer/EXTERNAL_API.md` both documented the destroyed manifest and erased annotations as current behaviour, and `EXTERNAL_API.md` recommended `--overwrite` as the workaround — advice that would tell authors to disable the ownership guard against a problem that no longer occurs. Both are rewritten to the post-fix behaviour, and both now name the overwrite-confirm prompt as the one path that still discards a manifest and its provenance. The collision-detection section of `CUSTOM_GUIDES.md` says so at the point the prompt is described, since that is where an author meets it.

## Breaking changes

None. `saveGuide` and `performBackendSave` gain a trailing optional parameter defaulting to the previous behaviour.

## Reversibility

Fully reversible — four source files and two developer docs, no schema, storage or API change.


Jayclifford345 (commented):
Reviewed the diff at `69a1a985` with `/review`, routed through `docs/design/CONCERNS.md` (`block-editor-authoring` plus the always-on concerns). Verified locally: `tsc --noEmit` clean, `eslint` clean on both changed source files, 52 block-editor suites / 699 tests pass.

### What checks out

- `fetchBackendGuides` returns `response.data.items` unprojected, so `existing.spec.manifest` really is in the snapshot — the passthrough is not writing an empty object back.
- Reading `resourceVersion` and the spec from one snapshot is the right call, and the 409 guards exactly the object whose unowned fields are being replayed.
- `stripAuthorNotes` still wins: `preservedSpec` overrides `blocks` after the `...existingSpec` spread.
- The foreign-overwrite path keeps `resourceVersion` while dropping annotations and the spec passthrough — the fail-closed direction, as described.
- No re-render hazard from adding `guides` to `saveGuide`'s deps: `performSaveDraft` / `handlePostToBackend` are wired only to UI props (`BlockEditor.tsx:813-814`), never to an effect.

### 1. The confirmation re-read can hide the affordance it exists to protect — medium

Bugbot flagged this at `useBackendGuides.ts:204`; I traced it independently and it holds.

`refreshGuides()` is state-committing, not a probe. It calls `setError(null)` up front and then commits whatever the LIST resolves to. Using it as the absence check lets a save launder a previously-visible error state into a clean-but-empty one:

1. Mount-time LIST fails with a non-`UNAVAILABLE_STATUSES` code → `error` set, `guides` stays `[]`. `hasManageableBackendGuides` returns `true` via the `error !== null` branch, so the guide-library entry stays visible.
2. User saves a guide tracked from `localStorage`. `existing` misses, so the new probe calls `refreshGuides()`.
3. The probe clears `error`, then either resolves `[]` from an `UNAVAILABLE_STATUSES` code (which `fetchBackendGuides` returns without throwing) or genuinely returns an empty namespace. Either way the committed state is `guides: []`, `error: null`.
4. `hasManageableBackendGuides` now returns `false` and hides the guide-library entry — the only affordance that would refresh the list — while `saveGuide` throws "try saving again", which can no longer succeed.

The PR body names this exact hazard as the reason for the re-read; the fix reinstates it in the other order.

Smallest fix is to make the probe non-committing, since it is a probe:

```diff
 if (existingResourceName && !existing) {
-  const refreshed = await refreshGuides();
+  const refreshed = await fetchBackendGuides(namespace);
   existing = refreshed.find((g) => g.metadata.name === existingResourceName);
 }
```

That leaves `guides` and `error` untouched, and has a second benefit: `fetchBackendGuides` rethrows on a non-`UNAVAILABLE` status, so a genuine LIST failure surfaces as itself through `orchestrateSave`'s catch instead of being laundered into "the resource does not exist".

### 2. Two developer docs now actively misinform — medium

Neither file is in the diff, but `docs/developer/CUSTOM_GUIDES.md` is a declared `trigger_path` for `block-editor-authoring`, so it is in scope for this concern rather than adjacent to it.

**`docs/developer/CUSTOM_GUIDES.md:126`** documents the bug this PR fixes as current behaviour, and closes with "Re-upload with the script rather than editing a cover page in place."

**`docs/developer/EXTERNAL_API.md:202-211`** goes further. It states the editor's save, publish and unpublish each send `metadata` as `{name, namespace, resourceVersion}` and that **"Nothing else does"** preserve annotations, then recommends `--overwrite`. Once this merges, that paragraph tells authors to disable the ownership guard to work around a problem that no longer occurs.

Suggested replacement for the `CUSTOM_GUIDES.md:126` paragraph — adjust to taste:

> Guides uploaded this way can be loaded and edited in the editor like any other, including a path's **cover page**. Save, publish, and unpublish each layer the editor-owned fields over the spec last read and carry `metadata.annotations` and `metadata.labels` through, so `spec.manifest` survives the write and a later `upsert-learning-path.sh` run still recognises the package as its own. The exception is the overwrite-confirm prompt described above: confirming an overwrite writes a guide that did not come from the stored resource, so it deliberately inherits neither the manifest nor the provenance annotations.

And for `EXTERNAL_API.md:202-211`:

> `upsert-guide.sh` merges its annotations over whatever the resource already carries, so an update through the scripts preserves annotations another tool set. The block editor's save, publish, and unpublish now do the same — each carries `metadata.annotations` and `metadata.labels` through from the resource it last read, so editing a script-uploaded guide no longer detaches it. The exception is a confirmed name-collision overwrite, which intentionally inherits neither the previous resource's manifest nor its annotations; after one of those, the next run refuses the package as foreign and `--overwrite` is the answer.

### 3. The contract anchor range was not advanced — low

The anchor row still reads `#1314 → #1331` while its contract text gained the partial-writer invariant introduced by this PR. Peer rows use the second number for the most recent contract-establishing PR (`completion-records | #1411 → #1433`, `cli-and-e2e-runner | #1372 → #1624`), so as written a future reader cannot find the PR that introduced the invariant they are reading. Suggest `#1314 → #1663`.

### 4. The overwrite path's silent discard is not documented where authors look — low

The PR body correctly identifies the overwrite-confirm modal as the one remaining path that discards a manifest without telling the author. `CUSTOM_GUIDES.md:65` is the paragraph describing that exact modal and says nothing about it — worth a sentence while you are in the file for finding 2. (The suggested replacement above already folds this in.)

### Tech debt — non-blocking

- `replacesForeignResource` is a fifth positional boolean. Call sites read `saveGuide(g, 'existing-guide', {...}, 'draft', true)`, where the most consequential argument is the least legible one. An options object for the trailing three would fix it.
- The PR introduces `BackendGuideMetadata` and `BackendGuideSpec` but leaves the public signatures on `any` (`existingMetadata?: any`, `currentMetadata: any`). The types it needed now exist; the boundary is where they would earn the most.

### Candidates checked and dropped

Listing these so the clean result reads as checked rather than skipped: re-render loop from the new `guides` dependency (refuted — no effect depends on the chain); redundant LIST from a stale closure (refuted — the closure is from the render that found the entry); stale snapshot replaying an old manifest over a newer one (refuted — `resourceVersion` comes from the same snapshot, so it 409s); the fail-closed guard bricking saves when the aggregation toggle is off (refuted — the write would fail at the backend anyway, and this fails faster with a clearer message).

Nice piece of work overall — findings 1 and 2 are the two I would hold for.

<sub>Reviewed with Claude Code using the repo's `/review` skill. Serial rather than the skill's parallel fan-out, so confidence on the *absence* of further findings is lower than a full run; each finding above is verified against head state.</sub>

Jayclifford345 (changes requested):
See above

tomglenn:
Thanks — findings 1–4 are in bc4aae49.

**1. The re-read hiding the affordance.** Confirmed and taken as suggested. `refreshGuides` commits, so using it as the absence check reinstated the dead end in the other order — exactly as you and Bugbot traced. The probe now calls `fetchBackendGuides(namespace)` directly, which touches neither `guides` nor `error`, and the second benefit you called out lands too: a non-`UNAVAILABLE` status now surfaces as itself through `orchestrateSave`'s catch instead of being laundered into "the resource does not exist".

Two tests pin it, both mutation-checked against reverting to `refreshGuides`:

- `probes without committing, so a live error and the guide-library entry both survive a refusal` — walks your exact four-step sequence and asserts `error` and `hasManageableBackendGuides` are unchanged after the refusal.
- `surfaces a failed probe as itself rather than as an absent resource`.

**2. The two misinforming docs.** Both rewritten, close to your suggested text. `EXTERNAL_API.md` was the worse of the two — telling authors to disable the ownership guard against a problem that no longer occurs — and its "**Nothing else does**" claim is now the opposite. Agreed on scope: `CUSTOM_GUIDES.md` is a declared `trigger_path` for this concern, so it was in scope rather than adjacent.

**3. The anchor range.** Advanced to `#1314 → #1663`. Also folded the non-committing probe into the contract text, since it is now part of what the invariant says.

**4. The overwrite path's silent discard.** Added a paragraph at the collision-detection section, where the prompt is actually described, rather than only in the learning-path section further down. Both places now name it as the one remaining path that discards a manifest without telling the author.

**Tech debt.** Took the second one — `existingMetadata`/`currentMetadata` are now `Partial<BackendGuideMetadata> | null` on all three signatures; the file no longer has an `any` outside `BackendGuideSpec['blocks']`. `useBackendSaveFlow`'s minimal duck-typed interface is left alone deliberately, since it exists to avoid importing from `useBackendGuides`.

Left the fifth positional boolean as is. You're right about the legibility, but converting the trailing three to an options object churns every call site in a PR that otherwise doesn't touch them, and the argument that matters is the one the two comments above it are about. Happy to do it as a follow-up if you'd rather it not linger.

Verified: 52 block-editor suites / 701 tests pass, `tsc --noEmit` clean, lint 0 errors.

Jayclifford345 (approved):

PR #1664, opened 2026-08-20
feat(completion-records): record completions for standalone private guides

### LINKED ISSUE #1423: Wire up completion tracking for private (App Platform) guides after completion-records rework lands
## Summary

Completion tracking does **not** record correctly for private (App Platform) guides — the custom guide packages surfaced by the `/custom-guide-repository` proxy. This is a **known, accepted gap**: the completion-tracking rework (the new `CompletionRecord` CRD in `pathfinder-backend` plus the `/completion-records/*` proxy in our Go backend) has only partially landed, so we deliberately did not build frontend keying against a moving target. Track here and wire it up once that rework is fully in.

Rendering, traversal, and locked-milestone display are unaffected — this is purely about completion **recording / keying**.

## Current behavior (broken)

- Per-member milestone completions key to `(bundled, "backend-guide:<id>")` instead of `(app-platform, <id>)`.
- Path-level completion fails closed and is never recorded.
- Standalone / orphan private guides record no completion at all (`onGuideComplete` only handles `bundled:` URLs and the learning-journey render type).
- As a result, **My Learning progress for App Platform paths is stuck at 0%** and paths never mark complete.

## Root cause

- The `/custom-guide-repository` proxy (and the frontend `CustomGuideManifest` type) don't carry `manifest.id` — only the top-level `entry.id`. So the manifest that reaches the completion recorder has no id.
- The launch bridge (`packageInfoForPath`) passes that id-less path manifest, and the My Learning adapter builds `LearningPath.guides` from bare ids that never match the recorded `completedGuides`.
- The recorder keys completions on `(manifest.repository, manifest.id)` but never receives a well-formed id, and the per-milestone package id is not threaded through `markMilestoneDone`.

## Fix outline (do **after** the completion-records rework fully lands)

- Carry `id` on the manifest: add it to the backend `customGuideManifest` struct + the frontend `CustomGuideManifest` type, or inject `entry.id` into the manifest object at the launch bridge / My Learning adapter.
- Thread a per-milestone package id into `markMilestoneDone` so per-member completions key on the member's own id, not the parent path's.
- Handle orphan / standalone private-guide completion (not just `bundled:` + learning-journey).
- Verify: completions land as `(app-platform, <manifest.id>)` and My Learning progress advances for an App Platform path (per the Custom Guide Packages RFC §6.10 completion contract).

## Blocked on

The completion-records rework fully landing (the `CompletionRecord` CRD in `pathfinder-backend`, the `/completion-records/*` proxy, and the completion-recorder boundary).

## Relevant files

- `src/components/docs-panel/CustomGuidesSection.tsx` (`packageInfoForPath`)
- `src/docs-retrieval/learning-journey-helpers.ts` (`markMilestoneDone`, completion-identity resolution)
- `src/learning-paths/app-platform-paths.ts`
- `src/lib/custom-guide-repository-client.ts` (+ the `custom_guide_repository` Go proxy)
- `src/components/docs-panel/components/DocsPanelContentArea.tsx` (`onGuideComplete`)

tomglenn:
Closes #1423.

## Summary

Completion tracking did not record for private (App Platform) guides — the custom guide packages served by the `/custom-guide-repository` proxy. The issue was written before the completion-records rework landed, and most of it is now fixed. This closes the one case that was not.

## What was already fixed, verified against `main`

Three of the issue's four symptoms closed with #1604 / #1433 / #1434. Each checked against the code rather than the issue text:

- **Milestone facts keyed on the scheme.** `getMilestoneSlug` (`src/lib/learning-journey-url.ts`) strips `backend-guide:`, so `markMilestoneDone` receives the bare member id and keys on `context.repository` plus that slug.
- **Path-level completion failed closed.** `stableJourneyId` resolves from `manifestGuideId(packageManifest)`, and both launch bridges (`CustomGuidesSection`, `MyLearningTab`) inject the entry id.
- **My Learning stuck at 0%.** `calculatePathProgress` matches `completedGuides` against `LearningPath.guides`, which `app-platform-paths.ts` fills with the same bare ids.

## What was still broken

Standalone/orphan private guides recorded nothing. Three facts compose into a silent loss:

1. Orphan and flat-list entries open via `openCustomGuide(guide, title)` with **no `packageInfo`**.
2. With no `packageInfo`, `docs-tab-loader` falls through to plain `fetchContent`, so `buildBackendGuideContent` is the producer — and it emitted `metadata: { title }` only.
3. `recordStandaloneGuideCompletion` fails closed on a missing manifest id **by design** (`repository-identity-authority`: never key a durable record on a loader URL).

The same loss applied to `?doc=api:<id>` share links and auto-dock tab restore, which also carry no `packageInfo`.

## What to look at

**The identity is stamped at the loader, not the recorder.** That fixes every entry point at once rather than patching one call site, and it keeps the invariant intact: identity still comes off a manifest built from resource fields (`spec.id ?? metadata.name`) and from nothing else — a resource carrying neither supplies no id, and the recorder fails closed rather than falling back to the URL segment.

**The precedence split is the load-bearing detail.** `id` and `repository` are forced over any persisted `spec.manifest` value, so a stray or stale field cannot outrank the resource actually being loaded, and everything served here is App Platform-sourced whatever the manifest claims. `type` is carried through **unforced**, so a private path cover is not recorded as a standalone guide. That is a gate, not a handoff — the journey trigger cannot fire on this path either (see known gaps below).

**Package-backed launches are unchanged.** `fetchPackageContent` spreads a caller-supplied manifest over the loader's, in source order, so the spread cannot invert. Where a caller supplies nothing (`fetchPackageById` with no manifest), the loader manifest now survives — an improvement, and `repository` resolves to `app-platform` either way.

**No new user-identifying data reaches the durable record.** `recordGuideCompletion` takes an explicit field list and never sends `packageManifest`. The only newly-durable field is `guideTitle` — the private guide's own title, which already flowed for catalogue-launched paths, and which lands in the caller's own namespace-scoped record keyed on their own `sub`.

## Contract evolution

**Follows** the `completion-records` anchor rather than extending or replacing it. The `repository-identity-authority` named invariant gains one clause: a launch that carries no resolved package is not exempt — its loader synthesizes the manifest, so identity still comes off a manifest and never off the scheme. Updated in this PR.

## Assessed and rejected

- **The issue's suggested backend change** (adding `id` to the Go `customGuideManifest` struct) is not needed. `entry.ID` already comes from `spec.id`, and all three catalogue consumers inject it.
- **#1567** (journey percentage never recorded as 100 for `backend-guide:` paths) is separable and left alone. It concerns the *local* `journeyCompletionStorage` percentage; this concerns the *durable* record key, and nothing on this path reads that percentage. #1567 is also `needs-decision` and opens with a product judgment call about what a `backend-guide:` percentage should mean — not something to settle inside a completion-keying fix.

## How to verify

```bash
npm run test:ci
npm run typecheck
npm run lint
```

Six new tests. They fail without the change:

```
● stamps the App Platform completion identity so a launch carrying no packageInfo still records
● supplies no id when the resource carries none, so the recorder fails closed on the loader URL
● carries a persisted manifest through, so a path cover is not mislabelled a guide
● never lets a persisted id or repository outrank the resource being loaded
● records the completion keyed on the App Platform guide id
● emits no guide fact for a path cover, so a path is never recorded as a standalone guide
```

Manually, on a stack with a published custom guide that is not a member of any path:

1. Open it from **Custom guides** in the sidebar (not via a path), complete it — a durable completion record appears keyed on the guide id with `app-platform` provenance.
2. Same via a `?doc=api:<id>` share link, and via auto-dock tab restore.
3. Open a private *path* and complete a milestone — unchanged; the journey trigger still owns that fact and it is not double-counted.

## Known adjacent gaps, deliberately untouched

- **Milestone-slug collision across journeys** — the durable key is the bare final segment, unqualified by owning journey, so two paths under one source sharing a member id conflate in the warehouse. Already documented as an accepted MVP gap in `learning-journey-helpers.ts`; slightly more reachable for App Platform now that bare ids are the slugs, but an existing tracked RFC-reconciliation item.
- **`shapeEntry` does not force `id` onto the catalogue manifest** the way it forces `repository`. Three consumers each inject it, so nothing is broken, but a fourth forgetting it would silently lose completion identity.
- **A path cover opened straight off this scheme records nothing at all.** The unforced `type` keeps a path out of the standalone-guide fact, but the journey trigger it would hand off to cannot fire here either: `buildBackendGuideContent` sets no `learningJourney` metadata and types the content `interactive`, so `willMarkMilestone` is always `false` and `useLastMilestoneAutoComplete` bails on the same check. Not a regression — the outcome was identical before this PR — but the handoff is a gap, not a route. Wiring the journey trigger for scheme-loaded path covers is separate work; #1663 makes it more reachable by letting manifests survive editor saves.
- **`packageInfoForPath` omits `repository`** — harmless, resolves to `app-platform` downstream, but an asymmetry with `MyLearningTab`.

## Breaking changes

None.

## Reversibility

Fully reversible — two source files plus tests, no schema, storage or API change.


Jayclifford345 (commented):
Reviewed the diff at `44e6c35e` with `/review`, routed through `docs/design/CONCERNS.md` (`completion-records`, `docs-retrieval-and-rendering`, plus the always-on concerns). Verified locally: `tsc --noEmit` clean, and 90 suites / 1044 tests pass across `src/docs-retrieval`, `src/completion-records` and `src/components/docs-panel`.

### What checks out

All three load-bearing claims in the description hold against head state:

- **Spread ordering.** `package-content.ts:308` confirms a caller-supplied manifest replaces the loader's in source order; the spread cannot invert.
- **No new user-identifying data reaches the durable record.** `resolveCompletionIdentity` extracts only `id` and `repository`, and `recordGuideCompletion` takes an explicit field list, so the author-controlled `spec.manifest` blob never leaves the client.
- **No storage-shape change.** `PersistedTabData` carries only `packageInfo`, never `content.metadata`, so the synthesized manifest is never persisted. The "fully reversible" claim is accurate.

Also checked clean: `useLastMilestoneAutoComplete` and the docs-panel journey branch are both gated on `type === 'learning-journey'` / `metadata.learningJourney`, neither of which this loader sets, so stamping the manifest changes nothing there.

### 1. The `type` passthrough hands path-typed resources to a trigger that cannot fire on this path — low

`buildLoaderManifest` carries `type` through unforced so a path cover routes to the journey trigger instead of emitting a guide fact. Right intent, but on *this* code path the journey trigger never runs: `buildBackendGuideContent` sets no `learningJourney` metadata and types the content `interactive`, so in `recordGuideCompletionForSurface` both `slug` and `journeyBase` are empty and `willMarkMilestone` is always `false` — and `useLastMilestoneAutoComplete` bails on the same type check. So when `spec.manifest.type === 'path'`, `recordStandaloneGuideCompletion` early-returns and nothing else records anything.

Not a regression — the outcome was identical before this PR — but the test named `leaves a path cover to the journey trigger instead of emitting a guide fact`, and the comment above `buildLoaderManifest`, both assert a handoff that does not exist here. That will read to the next maintainer as covered. And #1663 makes the case materially more reachable, since manifests will start surviving editor saves rather than being destroyed.

No code change needed — suggest renaming the test to what it actually verifies (that no guide fact is emitted) and adding the gap to the "known adjacent gaps" list, which is otherwise an unusually good version of that section.

### 2. `guide.id`'s third fallback is URL-derived — nit

`guideResource.spec.id || guideResource.metadata?.name || resourceName`, where `resourceName` comes off the `backend-guide:` URL. Unreachable in practice — a k8s GET by name always echoes `metadata.name`, and it equals the URL segment by construction — but the new test `keys on the resource name when spec.id and metadata.name are absent` pins that fallback as the durable key, while the invariant it cites says identity never comes off a loader URL, and `completion-identity.ts`'s header comment says the scheme "leaves the completion path entirely". Either worth a clause in the amended invariant, or tighten the fallback to two tiers so the docs and the code agree.

### 3. Comment length — nit, judgment call

The 20-line block comment on `buildLoaderManifest` is well past AGENTS.md's "if the comment won't fit on one short line, rename or restructure instead" bar. The precedence rule it encodes is a genuine hidden invariant so it plausibly earns its place, but it is worth a second look given the repo default is zero comments.

### Interaction with #1663

The two PRs edit different regions of `CONCERNS.md` (this one the named-invariants table, #1663 the routing row and anchors table), so they will not conflict textually. The one real interaction is finding 1 above: #1663 makes preserved manifests the norm, which enlarges the set of resources reaching this PR's unforced-`type` branch.

### Instrumentation

This PR *is* the instrumentation fix — it closes a silent-loss path rather than adding one. No gaps.

The "what was already fixed, verified against `main`" section is genuinely useful and each of the three claims checks out. Findings above are all low or nit; nothing here blocks.

<sub>Reviewed with Claude Code using the repo's `/review` skill. Serial rather than the skill's parallel fan-out, so confidence on the *absence* of further findings is lower than a full run; each finding above is verified against head state.</sub>

Jayclifford345 (dismissed):
LGTM -> one nit but nothing major

tomglenn:
Thanks — all three addressed in 076867ef.

**1. The `type` passthrough.** You're right that the handoff doesn't exist on this path, and the test name and comment both claimed it did. The passthrough itself stays — it's still what keeps a path cover out of the standalone-guide fact — but it's a gate, not a route, and everything describing it now says so:

- test renamed to `emits no guide fact for a path cover, so a path is never recorded as a standalone guide`
- the `buildLoaderManifest` comment now says "carried through unforced so a path cover is not recorded as a standalone guide", with no mention of a trigger taking over
- the PR body's "what to look at" line corrected, and the gap added to the known-gaps list with your reasoning: no `learningJourney` metadata, content typed `interactive`, so `willMarkMilestone` is always false and `useLastMilestoneAutoComplete` bails. Also noted that #1663 enlarges the set of resources that reach it.

**2. The URL-derived fallback.** Tightened the code rather than documenting the exception — the docs were right and the code was wrong. The manifest id is now `spec.id || metadata.name` and nothing else; a resource carrying neither supplies no id, so `manifestGuideId` returns undefined and the recorder fails closed, which is what the invariant says should happen. The content payload's `guide.id` keeps its three-tier fallback, since that one feeds schema validation rather than the durable key. The test is inverted to match: `supplies no id when the resource carries none, so the recorder fails closed on the loader URL`.

**3. Comment length.** Trimmed 20 lines to 7. Kept the precedence rule and the fail-closed clause, dropped the restated call-site inventory — the call sites are in the test names.

Verified after the rebase onto the merge commit: 35 suites / 500 tests pass across `src/docs-retrieval` and `src/completion-records`, `tsc --noEmit` clean, lint clean.

Jayclifford345 (approved):

Issue #124, opened 2025-08-27
Release Req: Enablement Video


Jayclifford345:
The support and GTM teams are aware of the feature based on a short [enablement video](https://docs.google.com/document/d/1nCiG62FxJ9J_qLTnlzNopSsT-VEAlxCUCsrAEpWL68U/edit), release note, or internal announcement on Slack or by email.

PR #1668, opened 2026-08-20
feat(coda): let a guide give its sandbox a gcx credential

Jayclifford345:
## What

Every Coda sandbox image ships the [`gcx`](https://github.com/grafana/gcx) CLI, and `grafana-coda-app` 1.3.0 added `POST /v1/sessions/{id}/credential` to give it a credential. Nothing in Pathfinder could call it, so any gcx command a guide taught would fail unauthenticated.

A `terminal-connect` block now takes `gcx: true`:

```json
{
  "type": "terminal-connect",
  "content": "Connect a sandbox and set up `gcx`:",
  "buttonText": "Connect and set up gcx",
  "vmTemplate": "vm-aws",
  "gcx": true
}
```

## Contract

This **extends** the established `coda-terminal` contract; it does not replace it. One additive optional field on `JsonTerminalConnectBlock`. Absent — which is every existing guide — the block behaves exactly as before, and the field is omitted rather than defaulted to `false`, so nothing existing changes shape. The contract anchor in `docs/design/CONCERNS.md` is updated in this PR.

The additive promise is load-bearing enough to be enforced rather than asserted: one credential store serves both surfaces, so "a block without `gcx` behaves as it did before the field existed" is a property of the code, not of the author's intent. See **One credential per session** below.

## Why minting happens in the browser

It cannot move into the plugin. Grafana refuses to create a service account whose role exceeds the caller's own, and a plugin's managed service account is created with no basic role at all — `plugin.json` `iam` grants fine-grained permissions, never a role. So the backend can mint nothing usable, by design, and minting with the user's own session satisfies Grafana's guard by construction instead of re-implementing it. Whatever comes back is capped at what the user already had.

**That consequence shapes the whole UI.** `serviceaccounts:create` is an Admin permission by default while sandbox sessions are open to Editors, so most people who can open a terminal cannot mint — and Grafana answers 403 on the service-account *search*, not only the create. `mint_forbidden` is a branch to a token field, not an error. A mint-only surface would reach a fraction of learners, so the paste path is the primary one, not a fallback. The mint button is offered regardless and only de-emphasised when the role check says it is unlikely: `canMintGrafanaToken` is a hint, and RBAC can grant `serviceaccounts:create` without the Admin basic role.

**The role cap holds only for a newly created account, so the account name is part of the guard.** `src/integrations/coda/gcx-service-account.ts` names the account itself — `coda-gcx-u<userId>` — rather than taking the client's `coda-gcx-<login>` default. That default lowercases the login and rewrites every non-`[a-z0-9-]` run to `-`, so `a.b` and `a-b` land on one name and reuse is by exact match: a collision would hand one person a token minted against another person's account, at that account's role. Ids do not collide. Reuse also outlives a role change — Grafana caps the role at creation and grants the *creator* write on the account, so an Admin who mints and is later demoted keeps a route to an Admin token — and `assertServiceAccountIsMintable` refuses a reused account that outranks the caller today, under its own `service_account_outranks_caller` code — comparing against the role Grafana holds *now*, read from `/api/user/orgs`, because `config.bootData` carries the role the page was loaded with and a demotion mid-session is the case the guard exists for.

**A lookup that cannot answer holds the mint back rather than waving it through.** The client reuses an exact name match without checking its role, so an account this code could not read is an *unknown* role and not an absent one; passing the name on regardless hands the reuse straight back to the client. Grafana's 403 on `serviceaccounts:read` is the ordinary route to the paste field below Admin, so it is reported as `mint_forbidden`; every other failure is the retryable `service_account_check_unavailable`, which returns to `idle` with the mint still on offer instead of sending a transient 503 to the paste field for the rest of the session. The preflight runs before the token name is claimed, so neither refusal burns one, and a pasted token skips it entirely — there is no account to reason about.

## Five behaviours worth reviewing, each pinned by a test

- **The step completes on the credential, not the connection.** Ticking it off at `connected` would report success while every command it exists to enable still failed unauthenticated.

- **A credential belongs to one step and one session.** One store serves both surfaces, so an install made anywhere reaches every mounted `terminal-connect` step — which would complete a step that never asked for gcx. Two guards: the step registers `onReady` only when `gcx` is set and guards its gcx render on the same flag, and `useGcxCredential(onReady, sessionId)` answers only for the session the caller names, reporting `idle` for any other. It fails closed — a caller that names no session is notified of nothing. An ordinary connect step still completes on its **Continue** button, exactly as before.

- **Do Section stops at a gcx step, through `pausesSectionRun`.** Not through the step's `executeStep` handle: `terminal-connect` is `refTarget: 'none'`, so no ref is attached and that handle has no runner-side caller. The runner routes the step through `executeInteractiveAction`, which has no `terminal-connect` case and therefore takes its `default:` branch and *reports success* — the forge `.cursor/rules/tracked-step-types.mdc` warns about. `TERMINAL_CONNECT_STEP_SCHEMA` now sets `pausesSectionRun: props.gcx === true`, which stops the run cleanly. Guides without `gcx` are unaffected.

- **One credential per session, and a late settlement cannot publish over its replacement.** The state lives in `gcx-credential-store.ts`, not in either component: two independent copies would let the toolbar offer a mint the step has already made. `TerminalProvider` invalidates the store when the session id changes, because a reconnect provisions a fresh VM holding no credential; a `null` id is only a disconnect and does not discard it. Provisioning takes long enough for that reconnect to happen mid-flight, so every reset and run start bumps a generation counter and a run publishes only while it still holds the current generation *and* the snapshot still names its session — otherwise A finishing after the terminal moved to B would report ready, and complete B's step, for a VM that never received a credential. Tokens also expire inside a long session, so the modal keeps a **Set up again** control; Grafana rejects a duplicate token name even after the first has expired, so each mint for a session gets a name of its own (`coda-<sessionId>`, then `coda-<sessionId>-2`). That also covers a mint whose delivery failed and discarded the only token value.

- **A refusal never dead-ends a guide, and a 404 means an old plugin.** The terminal is connected either way, so the step offers "Continue without gcx" and later `is-terminal-active` steps still run. There is no capability flag for the credential route — `features` is still `["exec.readyFile"]` — and the call only ever follows a session we just connected, so `session_not_found` is reported as a Coda plugin predating 1.3.0.

## Only a minted token's role and lifetime are ours

A mint asks for a short-lived token that expires on its own, inside the VM's lifetime, and Grafana's guard on *creating* the account caps it at the learner's own role. Neither claim survives the paste path: a **pasted** token is forwarded to the backend unchanged, so it keeps its own service account's role — a pasted Admin token is an Admin credential readable at root — and Pathfinder cannot shorten it or revoke it, holding no `serviceaccounts:delete`, so one created without an expiry stays valid long after the sandbox is gone.

So the form makes each claim only where it holds: the role cap and the expiry sit on the mint button, and the field carries "keeps its own service account's role, so give it the least privilege that works — and an expiry when you create it". Enforcing either is not available — the users the paste field exists for are precisely the ones with no service-account permissions, and the token is opaque to us before it reaches the backend.

## The block-editor form, and a limitation recorded not hidden

The form carries the field because it must: it rebuilds the block from scratch on submit, so a field it does not know is dropped the moment an author opens an existing block for editing.

**`gcx` does not survive a write to the backend yet.** `kinds/interactiveguide.cue` in `grafana-pathfinder-backend` declares `vmTemplate`, `vmApp` and `vmScenario` but not `gcx`, and the API server prunes an undeclared field with a 200 and no message. The prune belongs to the CRD, not to the upload script, so it is every path through that resource — `scripts/upsert-guide.sh` **and** a block-editor save or publish. Bundled packages and locally loaded JSON are unaffected, so a guide that needs `gcx` is servable today, just not from a Cloud stack.

It is registered in `src/validation/upsert-script-crd-fields.test.ts` as a documented prune alongside the `dataCheck*` family, called out for authors in `json-guide-format.md`, and — so it is not a discover-on-reload failure — raised as a warning in the form itself the moment the box is ticked. The fix is a `#Block` field in that repo's CUE.

## Requirements re-check on connect

`is-terminal-active` and `coda-exit-zero:` read live module state, so they can see a connection — but nothing tells them to look again when one arrives, and provisioning a VM takes about a minute, so every terminal step has long since answered "not connected". `TerminalProvider` dispatches `TERMINAL_STATUS_CHANGED_EVENT` (in `lib/event-names.ts`, with every other `pathfinder-*` window event) and `step-checker.hook.ts` rechecks on it. Only steps whose requirements or objectives actually name the terminal subscribe: every step holds one of these listeners and a reconnect is three status changes, so an ungated listener would run a full requirements check for every blocked step in the guide, three times over.

## Instrumentation

`recordGcxCredentialDegradation` emits the rung that stopped an install (`mint-forbidden`, `account-outranks-caller`, `account-check-unavailable`, `plugin-too-old`, `refused`), and `gcx_credential_installed` / `gcx_setup_skipped` count the outcomes. The whole shape of this surface rests on how often `mint_forbidden` comes back, so that rate cannot be left unmeasurable; `account-outranks-caller` is separate because it is the one rung an operator can clear — there is an account to delete — while `mint-forbidden` needs a role change. `account-check-unavailable` is separate again because it is the rung that is nobody's decision — a rate that climbs means the preflight is failing closed on people who should have been allowed through. No token, session id, or backend error text is attached.

## One fix riding along

Independent of gcx and separately reviewable:

- **The terminal no longer leaks an OSC 11 reply as text.** xterm answers OSC 10/11/12 through `onData`, `onData` goes straight to the PTY, and over a relay the reply lands after the asking program has exited and restored termios ECHO — so the shell echoes `11;rgb:1e1e/1e1e/1e1e` as text. `TerminalPanel` registers suppressing handlers after `open()`; they are disposed with the terminal.

## Client dependency

Moves `@grafana/coda-client` from 1.1.1 to **1.2.2** for `provisionGcxCredential`, `canMintGrafanaToken` and the three new error codes.

`1.2.2` rather than `1.2.1` because 1.2.1 was published without its `dist/` — three files, no entry point, unusable. `grafana/grafana-coda-app#120` fixed the publish workflow and cut 1.2.2; both are now merged and released, and the published tarball was checked by hand rather than by trusting the release job: 27 files, both entry points, all four gcx exports present.

`npm ci` from the committed lockfile succeeds under this repo's `.npmrc`. Worth knowing for the next bump: `min-release-age=3` filters at *resolution* time, so `npm install` refuses a same-day version (needs `--min-release-age=0`) while `npm ci` is unaffected.

Note that the client's own `gcxServiceAccountName(login)` default still has the collision described above; this PR routes around it by naming the account itself. The upstream fix belongs in `grafana/grafana-coda-app` and is tracked there as [#124](https://github.com/grafana/grafana-coda-app/issues/124), which also covers the non-retryable default token name.

## E2E selector contract

The thirteen new `data-testid` values across both id spaces — step-scoped and toolbar — are documented in `docs/developer/E2E_TESTING_CONTRACT.md` with the condition each renders under and the lifecycle a runner has to honour (the form detaches entirely while provisioning; a step without `gcx` renders none of them; a reconnect brings the form back). `src/integrations/coda/GcxSetupPanel.contract.test.tsx` pins the values, the source references and the visibility states.

## Testing

`npm run check` clean — typecheck, lint, prettier, `lint:go`, `test:go`, coverage thresholds, `test:review-contract` and `test:scripts`.

Coverage for the behaviours above: 19 cases on `terminal-connect-step` (happy path, `mint_forbidden` → paste → install, an unlikely-to-mint user, old plugin, other refusals, continue-without and its analytics, both `executeStep` paths, and three mixed-step cases proving an install elsewhere leaves an ordinary step and an other-session gcx step alone), 26 on the credential store and hook including cross-surface sharing, session scoping, a reconnect during provisioning, retry token names, concurrent-run refusal and the four preflight outcomes, 16 on the service-account guard, including the demoted Admin, an unreadable account and an unreadable caller role, 16 + 7 on the form and its selector contract, 6 on the block-editor form, 3 on `TERMINAL_CONNECT_STEP_SCHEMA`'s pause behaviour, 4 on parser passthrough including rejection of a non-boolean `gcx`, and 6 on `coda-api`.

The plugin-settings read-back fix that used to ride along here has moved to **#1723**. It is unrelated to gcx, and the rewrite had dropped the raw `enableAiAutoHeal` pass-through — restored there, with component tests on both write paths.

A hands-on guide of real gcx commands was authored alongside this and validates in `--strict` mode; it is not part of this PR by request.


package.json
```diff
@@ -121,7 +121,7 @@
     "@dnd-kit/utilities": "^3.2.2",
     "@emotion/css": "^11.13.5",
     "@grafana/assistant": "^0.1.25",
-    "@grafana/coda-client": "npm:@grafana-coda/sandbox-client@^1.1.1",
+    "@grafana/coda-client": "npm:@grafana-coda/sandbox-client@^1.2.2",
```
moxious, line 124: **High: Reused service accounts can exceed the caller's role**

The 1.2.2 client maps distinct logins such as `a.b` and `a-b` to the same service-account name, then reuses an exact match without checking its owner or role. An Admin who mints once and is later demoted can reuse that Admin account and mint an Admin token because Grafana grants the creator account-scoped write permission and does not reapply the caller-role cap during token creation.

Key accounts by a collision-free user identifier and reject or reconcile any existing account whose owner or role differs before enabling minting.

src/components/block-editor/forms/TerminalConnectBlockForm.tsx
```diff
@@ -134,6 +136,13 @@ export function TerminalConnectBlockForm({
         </Field>
       )}
 
+      <Checkbox
```
moxious, line 139: **High: Publishing from the editor silently strips `gcx`**

You expose this checkbox in the editor, but editor saves send `guide.blocks` through the `InteractiveGuide` CRD, whose `#Block` does not declare `gcx`. Selecting it and saving or publishing succeeds while the API server prunes the field, so reloading the guide reconnects without a credential; the current warning incorrectly limits this to script uploads.

Land the backend CUE field before exposing this option, then cover an editor save/read round trip.

src/components/interactive-tutorial/terminal-connect-step.tsx
```diff
@@ -150,15 +159,51 @@ export const TerminalConnectStep = forwardRef<
...
-    const handleConnect = useCallback(() => {
+    const {
+      state: gcxState,
+      error: gcxError,
+      credential: gcxCredential,
+      offerMint,
+      mintLikely,
+      isPending: gcxCredentialPending,
+      run: runGcxCredential,
+    } = useGcxCredential(markComplete);
```
moxious: **High: One credential completes unrelated terminal steps**

You register `markComplete` with the shared gcx store for every `terminal-connect`, even when `gcx` is omitted, so an install from the toolbar or another step completes every mounted connect step. In a guide with an ordinary connect step or a later step targeting another VM, the credential silently completes that step without its action, breaking the additive contract and sequential tracking.

Register completion only for a gcx-enabled, eligible step awaiting the matching session, and add a mixed-step regression test.

src/integrations/coda/gcx-credential-store.ts
```diff
@@ -0,0 +1,119 @@
...
+    publish({ ...IDLE, state: 'failed', error: 'The sandbox is not connected, so gcx cannot be set up yet.' });
+    return;
+  }
+  if (snapshot.state === 'provisioning') {
+    return;
+  }
+
+  publish({ sessionId, state: 'provisioning', credential: null, error: null });
+  try {
+    const written = await provisionGcx(sessionId, token ? { token } : {});
```
moxious: **High: An old install can mark the replacement session ready**

If session A is provisioning while the terminal reconnects to B, invalidation resets the snapshot but A's eventual success or failure publishes unconditionally over B. The UI can then show ready and complete B's step even though B never received a credential, or A can clobber B's own run.

Guard every settlement with a per-run generation and session check, and test a reconnect during provisioning.

src/integrations/coda/GcxSetupPanel.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+  return (
+    <div className={styles.panel}>
+      {error && (
+        <span className={styles.error} data-testid={testIds.error}>
+          {error}
+        </span>
+      )}
+
+      <span className={styles.warning}>
```
moxious, line 121: **High: Pasted tokens do not expire with the VM**

You tell every learner that the token expires with the VM, but the primary paste path forwards the supplied token unchanged and cannot control its lifetime. A non-expiring or long-lived service-account token remains valid after the VM and session end while the UI says its exposure is bounded.

State that a pasted token keeps its original lifetime and require a short-lived token, or enforce and revoke one.

src/integrations/coda/TerminalPanel.tsx
```diff
@@ -603,6 +631,40 @@ export function TerminalPanel({ onClose }: TerminalPanelProps) {
...
+        {gcxCredential.credential ? (
+          <>
+            <GcxReadyLine credential={gcxCredential.credential} testId={testIds.codaTerminal.gcxReady} />
+            {/* Tokens expire well inside a long session, and the VM cannot be
+                re-provisioned from here, so the form has to stay reachable. */}
+            <Button
+              size="sm"
+              variant="secondary"
+              fill="text"
+              onClick={gcxCredential.reset}
```
moxious, line 645: **Medium: “Set up again” cannot mint again**

After expiry, this action resets the store and mints another token named `coda-<sessionId>`, but Grafana rejects that duplicate name even when the old token has expired. A transient Coda delivery failure after a successful mint reaches the same dead end because the only token value was discarded.

Use unique retry names, delete the prior token, or retain it for delivery retries, and test renewal plus post-mint delivery failure.

src/components/AppConfig/ConfigurationForm.tsx
```diff
@@ -79,15 +81,17 @@ const ConfigurationForm = ({ plugin }: ConfigurationFormProps) => {
   useEffect(() => {
```
moxious, line 79: **Medium: Settings still use a mount-time snapshot**

You fetch plugin settings once when the form mounts, then both write paths later reuse `resolvedEnabled`, `resolvedPinned`, and `resolvedJsonData`. If another tab changes those values after this page opens, saving here restores the old values, which is the stale-snapshot failure this PR says it fixes.

Fetch and merge immediately before each whole-object settings write, or add concurrency control.

src/constants/testIds.ts
```diff
@@ -170,6 +170,12 @@ export const testIds = {
     terminalConnectStep: (stepId: string) => `interactive-terminal-connect-${stepId}`,
     terminalSkipButton: (stepId: string) => `interactive-terminal-skip-${stepId}`,
     terminalCopyButton: (stepId: string) => `interactive-terminal-copy-${stepId}`,
+    gcxMintButton: (stepId: string) => `interactive-gcx-mint-${stepId}`,
```
moxious, line 175: **Medium: Document the new gcx test selectors**

These thirteen new `data-testid` values add a stable DOM-testing surface, but `docs/developer/E2E_TESTING_CONTRACT.md` does not define their names or visibility states. That leaves future E2E consumers without the repository's required selector contract and lets a UI refactor break the flow silently.

Add a gcx selector section and a contract test for the values and lifecycle.

moxious (changes requested):
Thanks for the careful write-up. I found credential-safety and correctness issues that need another pass. CI is green (37 passed, 0 failed), but GitHub also reports this branch as dirty and non-mergeable; please rebase it and I will re-look after these are addressed.

moxious:
Two coordination notes on top of the review, both of which are about *where* work lands rather than about the code in this diff.

**The service-account role-cap issue is not really yours.** The first high finding traces to `@grafana/coda-client` 1.2.2, which keys accounts by normalized login and reuses an exact match without checking the account's owner or role. This PR only surfaces it by being the first caller to mint. Fixing it properly means the client stops reusing an account whose owner or role differs, so it probably wants raising with the coda-client owners rather than being solved here. Worth deciding whether this PR blocks on that or takes a local guard and lets the client fix follow.

**The `gcx` field has to exist in the backend before this can be complete.** The `InteractiveGuide` CRD's `#Block` does not declare `gcx`, so any guide saved from the editor with the checkbox set loses it silently on write. That makes the CUE change a hard predecessor, not a follow-up — no amount of front-end work here closes it. Sequencing that first also gives the editor round-trip something real to test against.

Neither is a criticism of the approach; the additive block-field shape looks right. It's the ordering that needs settling.

src/components/interactive-tutorial/terminal-connect-step.tsx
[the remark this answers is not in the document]
Jayclifford345: Fixed in f8b2c56d — you and the review below found the same bug.

`onReady` is now registered only when the step actually has `gcx`, and `useGcxCredential(onReady, sessionId)` answers only for the session the caller names, so a toolbar install reaches neither an ordinary connect step nor a gcx step waiting on a different VM. The step's gcx render is guarded on the same flag, so it does not borrow the ready line either.

Three cases in `terminal-connect-step.test.tsx` cover it: ordinary step untouched, same-session gcx step completed, gcx step on another session untouched.

package.json
Jayclifford345: Fixed in f8b2c56d and a1c0b05e, in this repo rather than by waiting on a client release.

**Keyed by a collision-free identifier.** New `src/integrations/coda/gcx-service-account.ts` names the account `coda-gcx-u<userId>` from the numeric Grafana user id and passes it as `serviceAccountName`, so the client's login normalisation — lowercase, `[^a-z0-9-]+` → `-`, truncate to 40 — is never reached. `a.b` and `a-b` no longer meet. Ids do not collide, and an anonymous visitor throws rather than deriving a shared name.

**Reconciled before minting.** `assertServiceAccountIsMintable` searches for that exact name and refuses when the account it resolves to holds a higher role than the caller holds *today* — the demoted-Admin case, which keying alone does not close, because Grafana caps the role at creation and grants the creator write on the account. A failed search is deliberately not a refusal: below Admin it is Grafana's own 403 on `serviceaccounts:read`, the ordinary route to the paste field, and the client's own search hits it a moment later.

a1c0b05e gives that refusal its own code, `service_account_outranks_caller`, rather than folding it into `mint_forbidden`. Both branch to the paste field, but only this one names something an operator can delete, and the generic sentence was swallowing the account name. It has its own `account-outranks-caller` telemetry rung for the same reason: it is the rung an operator can clear.

Nine cases in `gcx-service-account.test.ts`, including the `a.b`/`a-b` pair, the demoted Admin, a partial name match, and the 403-on-search path.

What this does **not** fix is the client's own default for any other consumer — `gcxServiceAccountName(login)` in `@grafana/coda-client` still collides. Worth an upstream issue on `grafana/grafana-coda-app`; happy to file it if you agree with the read.

src/components/block-editor/forms/TerminalConnectBlockForm.tsx
Jayclifford345: The fact is right and the note was understated — thank you. The prune belongs to the CRD, not to the upload script, so it applies to every write through the resource, a block-editor save or publish included. Corrected in f8b2c56d across `docs/developer/CODA.md`, `json-guide-format.md` and the comment in `upsert-script-crd-fields.test.ts`.

I have **not** taken the remedy you asked for, so leaving this open for your call. Two reasons, and I would rather you overrule me than have me quietly do it:

**Landing the CUE first is out of this repo,** and it makes this PR wait on a `grafana-pathfinder-backend` release. If you want that ordering, say so and I will file the `#Block` field there and mark this blocked on it.

**Removing the checkbox makes the editor worse, not safer.** The form rebuilds the block from scratch on submit, so a field it does not know is dropped the moment an author opens an existing block — including a `gcx` block that arrived from a bundled package or a local file, where the flag *does* work. Today the editor round-trips it; without the checkbox the editor becomes a way to destroy `gcx` locally as well as on the backend. There is also shipped precedent: the whole `dataCheck*` family from #1612 is exposed in `InputBlockForm` under exactly this prune, and is registered in the same `PRUNED_BY_CRD` list.

What I did instead is stop it from being a discover-on-reload failure: ticking the box now raises a warning in the form itself — "gcx is not stored on the backend yet … saving or publishing this guide drops it without an error and the step reloads without it. Serve the guide from a bundled package or a local file until the backend declares `gcx`." Covered by "warns that a backend write drops the field, only once it is ticked" in `TerminalConnectBlockForm.test.tsx`.

On the round-trip test you asked for: an editor save/read round trip through the CRD cannot be tested from this repo — the prune happens in the API server, and mocking it would only assert our own guess about the CRD. `upsert-script-crd-fields.test.ts` is the seam we do have, and it now says the prune covers every write rather than just the script.

src/components/interactive-tutorial/terminal-connect-step.tsx
Jayclifford345: Fixed in f8b2c56d.

Two guards, because the leak had two halves:

- **The flag.** `useGcxCredential(gcx ? markComplete : undefined, ...)`, and the step's `GcxReadyLine` and `provisioning` step-state are guarded on `gcx` too — so an ordinary connect step neither completes on someone else's credential nor renders one. It still completes the way it always did, on its **Continue** button.
- **The session.** The hook now takes the session it is asking about and reports `idle` for any other, so a later step targeting another VM is not completed by the previous VM's credential. It fails closed: a caller that names no session is notified of nothing.

Mixed-step regression tests as you asked, in `terminal-connect-step.test.tsx` — "leaves an ordinary connect step alone", "completes a gcx step waiting on the same session", "leaves a gcx step targeting another VM alone" — plus "answers idle for a session other than the one installed into" and "does not complete a caller waiting on another session" in `useGcxCredential.hook.test.ts`.

src/integrations/coda/gcx-credential-store.ts
Jayclifford345: Fixed in f8b2c56d.

Every reset and every run start bumps a module-level generation counter, and `settle()` publishes only while the run still holds the current generation **and** the snapshot still names its session. So A finishing after the terminal moved to B is discarded — no ready state, no completion, no telemetry rung for a session nobody is looking at.

Test: "does not let an install that settles late publish over its replacement" in `useGcxCredential.hook.test.ts` — it starts a run for `s_abc`, invalidates to `s_def` mid-flight, then settles the original and asserts the store is still `idle` and `onReady` never fired.

src/integrations/coda/GcxSetupPanel.tsx
Jayclifford345: You are right, and the copy said something the paste path cannot deliver. Fixed in f8b2c56d.

The exposure sentence and the lifetime claim are now separate, and the lifetime claim is attached to the path it is true of:

- next to the mint button: "Mints a short-lived token that expires on its own";
- next to the paste field, always: "Give a pasted token an expiry when you create it. Pathfinder cannot shorten or revoke one, so a token with no expiry stays valid long after this sandbox is gone."

I did not take the enforce-and-revoke option. Revoking needs `serviceaccounts:delete` on the account, and the whole reason the paste path exists is that most people who can open a sandbox hold no service-account permissions at all — so the enforcement would fail for exactly the users the field is for. Rejecting a long-lived token is not available either: the token is opaque to us, and the plugin has no route to read its expiry before handing it to the backend. Stating it accurately and asking is what is actually within reach here.

The warning has its own test id (`interactive-gcx-token-lifetime-${stepId}` / `coda-terminal-gcx-token-lifetime`) so it cannot be dropped silently, and `GcxSetupPanel.test.tsx` asserts the minted claim is absent once minting has been refused while the pasted warning stays up.

src/integrations/coda/TerminalPanel.tsx
Jayclifford345: Fixed in f8b2c56d — unique retry names, which covers both cases you name.

The store keeps a per-session attempt count and hands each mint a name of its own: `coda-<sessionId>`, then `coda-<sessionId>-2`, and so on. The count resets when the session changes, since a new session's names are free anyway. It is passed explicitly as `tokenName` rather than relying on the client's default.

That is deliberately the "unique names" option and not "delete the prior token" — the browser would need `serviceaccounts:write` on the account to delete a token, and the surface is built for people who cannot even mint. Retaining the token value for a delivery retry was the other option and I did not take it: it means holding a live credential in module state across an error path, which is worse than asking Grafana for a second short-lived one.

Tests: "asks for a fresh token name on a repeat mint for the same session" (the failed-delivery path, then a retry) and "starts token names over for a new session".

src/components/AppConfig/ConfigurationForm.tsx
Jayclifford345: Fair — that was the same bug one level up. Fixed in f8b2c56d.

Both write paths now `await fetchPluginSettings(...)` immediately before the write and take `enabled`, `pinned` and the base `jsonData` from that read. The mount-time `resolvedEnabled` / `resolvedPinned` state is deleted rather than left as a trap; `resolvedJsonData` stays only as the seed for what the form renders. This is the shape `dev-mode.ts` already used.

New `ConfigurationForm.test.tsx` covers it: a second tab flipping `pinned` and adding a provisioned field between mount and save, the assistant dev-mode toggle taking the same path, and a failed read-back writing nothing at all rather than falling back to the stale values.

src/constants/testIds.ts
Jayclifford345: Added in f8b2c56d.

`docs/developer/E2E_TESTING_CONTRACT.md` gains a **gcx credential setup** section: a table of all the ids in both id spaces — step-scoped and toolbar — with the condition each renders under, plus the lifecycle notes a runner actually needs:

- the whole form detaches while provisioning, so poll for the ready line or the error, never for the control you just clicked;
- a step without `gcx` renders none of the step-scoped ids even while another surface installs into the same session, and still completes on `interactive-terminal-skip-${stepId}`;
- a credential belongs to one session, so a reconnect detaches the ready line and brings the form back;
- "Continue without gcx" still reaches `data-test-step-state="completed"`.

`src/integrations/coda/GcxSetupPanel.contract.test.tsx` pins the constant values, the source references, and the visibility states, and is listed under Contract Tests alongside the other four.

package.json
Jayclifford345: Filed upstream: grafana/grafana-coda-app#124 (private repo, `security` label).

It covers both halves of the same code path — the lossy name in `gcxServiceAccountName`, and `findOrCreateServiceAccount` reusing an exact match without an owner or role check — with the demoted-Admin case that keying alone does not close, and an honest note on reachability: below Admin the search 403s first, so a default stack is not exposed by the collision. What makes it worth fixing is that the module's own guidance steers operators toward granting `serviceaccounts:*` to Editors, which is exactly the configuration that opens it. The role-drift half needs no RBAC change at all.

I offered to port this PR's `gcx-service-account.ts` up as a PR there rather than have them write it twice.

Leaving this thread open for you — nothing here is blocked on the upstream fix, but you may want to see the read before resolving.

moxious (changes requested):
## Merge contract

1. [blocking] **gcx-service-account-role-guard-fails-open — The elevated-account safety check fails open when its lookup fails** (high · security · irreversible without cleanup)
   The guard says that checking the current account role is the only barrier against a demoted creator reusing an Admin service account, but src/integrations/coda/gcx-service-account.ts:60-74 uses a boot-time caller role and treats every search failure as success. The store then passes that same reusable name to @grafana/coda-client at src/integrations/coda/gcx-credential-store.ts:114-117 and 153-156; the client performs another lookup and reuses an exact match without checking its role. A stale role or a transient first failure can therefore mint and install a token that outranks the caller.
   Required: Fail closed unless an authoritative check proves the account is absent or no stronger than the caller. Map expected authorization denial to mint-forbidden, surface transient lookup failures as retryable, and add a store-level test asserting provisionGcx is not called after a failed preflight.

2. [blocking] **configuration-save-materializes-ai-auto-heal-default — An unrelated configuration save freezes the AI auto-heal default** (medium · correctness-and-reliability · irreversible without cleanup)
   Both writes at src/components/AppConfig/ConfigurationForm.tsx:168-176 and 237-251 apply getConfigWithDefaults without restoring raw enableAiAutoHeal. The repository's own demonstration at src/components/AppConfig/settings-preservation.test.ts:161-190 shows that this persists today's default for a field the form does not own; base restored the raw value in both paths. Existing stacks can therefore stop following future defaults after an unrelated settings save.
   Required: Restore enableAiAutoHeal: current.jsonData?.enableAiAutoHeal after applying defaults in both write paths, and add component tests proving omitted stays omitted and explicit false stays false.

3. [blocking] **gcx-pasted-token-role-warning-is-false — The credential warning misstates the role of a pasted token** (medium · security · partially reversible)
   src/integrations/coda/GcxSetupPanel.tsx:121-124 says every token is scoped to the learner's Grafana role, but the paste path at lines 135-163 accepts an opaque service-account token and src/integrations/coda/gcx-credential-store.ts:153-156 forwards it unchanged. A pasted Admin token stays Admin, so the warning understates the privilege of a root-readable credential that Pathfinder cannot revoke.
   Required: Limit the role-capping statement to minted tokens, explicitly say that pasted tokens retain their service account's role and should be least-privilege, and assert that distinction in the panel test.

## Follow-ups

1. [follow_up] **terminal-recheck-drops-enabled-objective-event — A queued terminal event is discarded for enabled steps with objectives** (medium · requirements-manager)
   The normal trigger at src/requirements-manager/step-checker.hook.ts:1022-1038 rechecks blocked steps and enabled steps with objectives. An event during a check is queued at lines 1082-1095, but the drain at lines 1119-1128 clears it and rechecks only disabled steps. An enabled terminal-dependent objective can remain stale after its only connected event; the tests cover each half separately, not the combined race.
   Follow-up: Use the same eligibility predicate in the drain as the normal trigger, and add a race test combining an enabled terminal-dependent objective with an event during the in-flight check.

2. [follow_up] **terminal-connect-editor-drops-unmanaged-fields — Editing a terminal-connect block drops its stable ID and author note** (medium · guide-schema-and-contracts · irreversible without cleanup)
   JsonTerminalConnectBlock declares id and authorNote at src/types/json-guide.types.ts:641-656, but src/components/block-editor/forms/TerminalConnectBlockForm.tsx:54-70 rebuilds the block without either and the editor replaces rather than merges it. Saving an existing block silently loses addressing and private author metadata. The omission is also present in the base revision, so it does not block this PR.
   Follow-up: Open a follow-up to preserve initial.id and initial.authorNote in the form and add an edit round-trip regression test.

3. [follow_up] **terminal-listener-misses-objective-prop-changes — Terminal-listener gating does not update when objectives change** (low · requirements-manager)
   src/requirements-manager/step-checker.hook.ts:1097-1105 gates subscription using objectivesRef.current, but the effect dependency list at line 1117 omits objectives. A mounted checker whose objective changes to or from a terminal-dependent clause retains the old subscription decision.
   Follow-up: Open a follow-up to derive a stable readsTerminalStatus value from current requirements and objectives, depend on it, and add rerender coverage for both transitions.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1668
Summary: let a guide give its sandbox a gcx credential
Verdict: Request Changes
Results: 3 blockers, 0 non-blocking findings, 3 follow-ups

Jayclifford345:
All three blockers addressed in `c0eca3bf` and `a694c796`, with the third one leaving this PR entirely. Taking them in your order.

---

### 1. `gcx-service-account-role-guard-fails-open` — fixed

You are right that the guard was the only barrier and that it was not one. Both halves you named are gone.

**The lookup no longer passes on a non-answer.** `catch { return; }` is deleted. An account this code could not read is an *unknown* role rather than an absent one, and the client reuses an exact name match a moment later, so a silent pass handed the reuse straight back to it. The two failure modes are now separated, as you asked:

- Grafana's `401`/`403` on `serviceaccounts:read` — the ordinary route below Admin — is reported as `mint_forbidden`, so it branches to the paste field exactly as the search 403 always did.
- Every other failure is a new retryable code, `service_account_check_unavailable`, which returns the store to `idle` with the mint **still on offer** rather than to `needs-token`. A transient 503 should not remove the mint path for the rest of the session, and nothing about it says the mint is disallowed. It carries its own telemetry rung, `account-check-unavailable`, because it is the rung that is nobody's decision: a rate that climbs means the preflight is failing closed on people who should have been let through.

**The comparison no longer uses a boot-time role.** `assertServiceAccountIsMintable` reads the caller's current role from `/api/user/orgs`, matched to `config.bootData.user.orgId`, instead of `config.bootData.user.orgRole`. The demoted-Admin case this guard exists for happens mid-session, which is precisely when the boot snapshot is on the wrong side of the comparison. An unreadable caller role, an org missing from the answer, and an account whose own role is unreadable all hold the mint back rather than allowing it.

The search runs first and the role read only happens when there is an account to compare against, so the common case is still one request.

**`provisionGcx` is not called after a failed preflight** — three store-level tests as you asked, in `useGcxCredential.hook.test.ts`: "never reaches the mint when the account cannot be read", "holds a mint back as retryable when the preflight cannot answer, keeping it on offer", and "burns no token name on a refused preflight, so the retry asks for the first one". A fourth, "runs no preflight at all for a pasted token", pins the other side of it.

`gcx-service-account.test.ts` is 16 cases now, including the demoted Admin against a stale boot role, an account in a second org, the 403-on-search, a transient 503, an unreadable caller role, and an account with no readable role. The per-route mock means a case can fail the search without failing the role read.

---

### 2. `configuration-save-materializes-ai-auto-heal-default` — fixed, in **#1723**

You are right on the fact, and this one should not have been in this PR at all. The plugin-settings read-back had nothing to do with gcx — it rode along, and reviewing it here is what let a regression into it. `ConfigurationForm.tsx`, `dev-mode.ts`, `utils.plugin.ts` and both of their test files are reverted out of this branch and moved to **#1723**, where `enableAiAutoHeal: current.jsonData?.enableAiAutoHeal` is restored after `getConfigWithDefaults` on both write paths.

One thing worth recording about why it slipped: `settings-preservation.test.ts`, which you cited, never renders or imports `ConfigurationForm`. Every case builds its own `buggyNewJsonData` / `fixedNewJsonData` literal and asserts on that, so it documents the shape of the spread without checking the component still follows it — it stayed green through the regression. #1723 adds the component tests you asked for, reading `updatePluginSettings.mock.calls`: "leaves a never-set enableAiAutoHeal unset on both write paths" and "keeps an explicit enableAiAutoHeal choice on both write paths". I checked that the first genuinely fails without the pass-through. The second passes either way, because `getConfigWithDefaults` resolves a stored `false` to `false` — it is a guard against over-correcting to `undefined`, not a regression test, and I would rather say so than imply it caught something.

---

### 3. `gcx-pasted-token-role-warning-is-false` — fixed

The sentence was simply false for the primary path, and I had already split the *lifetime* claim in the last round without noticing the *role* claim sitting next to it in the same breath.

The blanket "It is scoped to your own Grafana role" is deleted. What is left says only what each path delivers:

- the always-shown warning keeps just the fact that holds either way — "The token is readable inside the VM — you have a root shell on the same box";
- next to the mint button: "Mints a short-lived token that expires on its own, capped at your own Grafana role" — where the cap is real, being Grafana's guard on *creating* the account;
- next to the field, always: "A pasted token keeps its own service account's role, so give it the least privilege that works — and an expiry when you create it. Pathfinder cannot narrow, shorten or revoke one …".

Two tests in `GcxSetupPanel.test.tsx` assert the distinction: "claims the role cap only for the token it mints" (and that the readable-in-VM warning makes no role claim at all), and "makes no role claim at all once minting has been refused" — the state where the paste path is the only one left. `CODA.md` no longer says a token is "capped at their own role" without qualifying which one.

---

### Follow-ups

Left alone, as you classified them — including `terminal-connect-editor-drops-unmanaged-fields`, which Bugbot flagged again and which is present in the base revision. Happy to file the three as issues if you would rather they were tracked outside the review.

### State

`npm run check` clean on both branches — typecheck, lint, prettier, `lint:go`, `test:go`, coverage thresholds, `test:review-contract`, `test:scripts` — and the full suite green each side. This branch is 39 changed files now; no `AppConfig`, `utils.plugin` or `dev-mode` left in the diff. The PR description is updated to match, and the E2E contract's mint-button condition with it, since a held-back mint re-attaches its own control and "no mint has been tried" was no longer the condition it renders under.

src/components/AppConfig/ConfigurationForm.tsx
```diff
@@ -162,16 +163,14 @@ const ConfigurationForm = ({ plugin }: ConfigurationFormProps) => {
...
-        enabled,
-        pinned,
+        enabled: current.enabled,
+        pinned: current.pinned,
         jsonData: {
-          ...(resolvedJsonData || {}),
-          ...getConfigWithDefaults(resolvedJsonData || {}),
-          // Pass through raw, not the resolved default — this tab doesn't own the
-          // field, and materializing it here would freeze out a future default change.
-          enableAiAutoHeal: resolvedJsonData?.enableAiAutoHeal,
```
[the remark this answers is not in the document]
Jayclifford345: This file has left the PR. The plugin-settings read-back was riding along here and is unrelated to gcx, so it moved to #1723 — where the raw `enableAiAutoHeal` pass-through you flagged is restored after `getConfigWithDefaults` on both write paths, with component tests reading `updatePluginSettings.mock.calls` on each.

Correct catch: the human review found the same regression independently. Worth noting for anyone reading the file that `settings-preservation.test.ts` does not exercise the component — it asserts hand-written jsonData literals — which is why nothing failed.

src/components/block-editor/forms/TerminalConnectBlockForm.tsx
```diff
@@ -61,11 +62,12 @@ export function TerminalConnectBlockForm({
         ...(vmTemplate.trim() && { vmTemplate: vmTemplate.trim() }),
         ...(vmApp.trim() && { vmApp: vmApp.trim() }),
         ...(vmScenario.trim() && { vmScenario: vmScenario.trim() }),
+        ...(gcx && { gcx: true }),
```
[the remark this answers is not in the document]
Jayclifford345: Real, and deliberately not fixed here. `handleSubmit` does rebuild the block from form-owned fields only, so `id` and `authorNote` are dropped on an edit save — but it did that before this PR too: the omission is in the base revision, and this PR only adds `gcx` to the set of fields the form does carry.

The human review reached the same conclusion and classified it as a follow-up (`terminal-connect-editor-drops-unmanaged-fields`) rather than a blocker for that reason. Fixing it properly means preserving `initial.id` and `initial.authorNote` plus an edit round-trip regression test, which is its own change against its own test — not something to graft onto a credential PR.

moxious (approved):

## Follow-ups

1. [follow_up] **gcx-install-completes-every-comounted-gcx-step — One gcx install completes every co-mounted gcx step** (medium · coda-terminal)
   terminal-connect-step.tsx:170 scopes the hook on the live session (useGcxCredential(gcx ? markComplete : undefined, terminalCtx?.sessionId)), and useGcxCredential.hook.ts:58-76 fires onReady for every subscriber whose sessionId matches. TerminalContext holds one live session id, so every mounted gcx step names the same session and all of them complete on the first install. Reproduced with two co-mounted gcx steps: both onComplete handlers fired. The existing 'another VM' case at terminal-connect-step.test.tsx:382 sets the global mockSessionId to s_other, a state two co-mounted steps cannot occupy. No harm at this head — no guide uses gcx yet and the flag defaults off — so it does not block, but a two-gcx-step guide would silently skip its second VM.
   Follow-up: Scope onReady to the step that requested the run rather than to the live session alone, and replace the global-mockSessionId test with two co-mounted gcx steps.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1668
Summary: let a guide give its sandbox a gcx credential
Verdict: Approve with Minor

moxious:
Follow-up from the review above is filed as https://github.com/grafana/grafana-pathfinder-app/issues/1772 — the co-mounted gcx step completion. Nothing blocking left on my side; merge when you're ready.

PR #1672, opened 2026-08-21
fix(e2e): extend pool lease acquisition timeout

leslie-heinzen:
## Summary

Managed pool lease acquisition now uses a 90-second timeout. Retirement keeps its 15-second timeout.

After a structured `no_capacity` response, each retry uses only the remaining maximum-wait budget. Transport timeouts remain non-retryable because the request can create a lease with a runner token that the client cannot replay.

## Why

This problem surfaced after the pool manager was retargeted to the Grafana dev environment. That environment applies a much stricter rate limit to non-GET lifecycle requests used for stack leasing, credential management, retirement, and replacement.

The pool manager now paces those requests to stay within the dev API limit. Compliant pacing can make synchronous lease requests wait for quota, but the runner still aborted every pool-manager request after 15 seconds.

During the contained cycle, the runner abandoned 15 lease requests that the pool manager later completed successfully in 15–53 seconds. Each response contained a one-time runner token that the pool manager does not persist. The abandoned requests therefore consumed stacks that the runner could not use.

Lease acquisition now permits 90 seconds, which covers the observed 53-second maximum with operating margin. Acquisition timeouts remain non-retryable because the first request can still issue a lease. Retrying can consume another stack without recovering the first token.

Retirement keeps its separate 15-second deadline because the runner does not need to wait for physical cleanup. Asynchronous retirement is handled in [pathfinder-e2e-platform #194](https://github.com/grafana/pathfinder-e2e-platform/pull/194).

## How to verify

1. Return `no_capacity` from the first lease request, then allow a retry. Confirm that the retry signal uses the remaining wait budget.
2. Let the retry signal expire. Confirm that the client reports the timeout and does not send another lease request.
3. Retire a lease through a delayed endpoint. Confirm that the retirement request uses a 15-second timeout.

## Breaking changes

None. This change is fully reversible.

Co-Authored-By: Warp <agent@warp.dev>

### DIFF
--- a/src/cli/e2e/cloud-stack-pool-manager.test.ts
+++ b/src/cli/e2e/cloud-stack-pool-manager.test.ts
@@ -71,11 +71,60 @@ describe('createCloudStackPoolManagerConfig', () => {
 
   it('exhausts retry budget correctly when the injected clock advances with each wait', async () => {
     let now = 0;
+    const waits: number[] = [];
     const fetchImpl = jest
       .fn()
       .mockResolvedValue(
         jsonResponse({ error: { code: 'no_capacity', message: 'capacity is temporarily exhausted' } }, 503)
       );
+    const manager = new CloudStackPoolManager(
+      { ...CONFIG, maxWaitSeconds: 1 },
+      false,
+      fetchImpl as unknown as typeof fetch,
+      async (ms) => {
+        waits.push(ms);
+        now += ms;
+      },
+      () => 0,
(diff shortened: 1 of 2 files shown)


Spidy88 (approved):
LGTM. Seems pretty straight forward. Without knowing more, I'd only caution about bumping timeouts without identifying whether the root cause is unavoidable or truly a performance problem :+1:

PR #1674, opened 2026-08-21
fix(deps): update grafana to 13.2.0 and migrate to React 19

Jayclifford345:
Replaces #1671, which could not land: every job on it died at `npm ci` with `ERESOLVE`, because `@grafana/{data,ui,runtime,i18n}` 13.2.0 declare a hard `react@">=19"` peer while this repo pinned `react@18.3.1`. Renovate's own `renovate/artifacts` check failed for the same reason, so the lockfile was never regenerated. The version bump and the React 19 migration have to land in one PR.

## Summary

- **`react` / `react-dom` 18.3.1 → 19.2.8**, matching what Grafana 13.2.0 itself builds against; `@types/react` and `@types/react-dom` to `^19`.
- **`@grafana/{data,e2e-selectors,i18n,runtime,schema,ui}` 13.1.3 → 13.2.0** and **`@grafana/faro-instrumentation-replay` `~2.9.0` → `~2.10.0`** — the original renovate bump, unchanged.
- **Override `@grafana/scenes`' `react` / `react-dom` peers to the root spec.** Scenes still caps `react@^18.0.0` as of 8.16.1 (there is no stable 9.x), but Grafana core 13.2 ships scenes on React 19, so the cap is stale rather than a real constraint. `react` is a webpack external here, so the override is install-time only and cannot change what scenes resolves at runtime.
- **Type fixes for `@types/react` 19** — mechanical, two shapes:
  - `RefObject<T>` prop and parameter types widened to `RefObject<T | null>` (28 declarations across 20 files). `useRef<T>(null)` now yields `RefObject<T | null>`, so the consumer signatures had to widen to match.
  - The five bare `useRef<T>()` calls given an explicit `undefined` initial value. React 19 types require an argument; the resulting `useRef<T | undefined>(undefined)` form already matches ~10 existing call sites in this codebase.
- **Jest ESM transform** — `@grafana/data` 13.2.0 reaches `@react-hookz/web` (and its `@ver0/deep-equal` dependency) from `hooks/useObservable`. Both are ESM-only with no CJS build, which took out **138 test suites** until they were added to `transformIgnorePatterns` alongside the existing `@grafana/coda-client` entry.

## What to look at

**`src/requirements-manager/step-checker.hook.test.ts`** is the only behavioural change, and it is a test-side fix. The objectives-check timer regression test asserted `jest.getTimerCount()` had not grown across `checkStep()`. Under React 19 that proxy is no longer valid: `act()` leaves a scheduler timer of its own pending on every flush, so the count grows even when nothing leaked. I confirmed this with a control — an **empty** `act(async () => {})` with no `checkStep()` call at all already moves the count from 2 to 3 — and by tracing set/clear, which showed the 3s timer being armed and cleared correctly.

The test now tracks that 3000ms timer by id and asserts it was cleared. It is a strictly tighter guard than the count diff: I re-introduced the leak by deleting the `.finally(clearTimeout)` in `step-checker.hook.ts` and the new assertion caught it, then restored the hook (no production change in this PR).

**`package.json` overrides** is the one judgement call — see the scenes bullet above for why the stale peer is safe to override rather than block on.

## How to verify

Locally green:

- `npm run check` (typecheck + lint + prettier + docs:sync-terms + lint:go + test:go + test:coverage + test:review-contract + test:scripts) — exit 0
- `npm run build`, `npm run build:cli`, `npx jest src/cli` (870 tests)
- `npm ci` from the regenerated lockfile, reproducing react 19.2.8 / @grafana/ui 13.2.0 with a single react copy in the tree
- `npx @grafana/react-detect@latest` — "No React 19 breaking changes detected"
- Full suite: 465/465 suites, 7440 tests passing

`npm audit` reports 8 pre-existing findings (3 moderate / 5 high). I checked every one against `HEAD`: `brace-expansion`, `dompurify`, `fast-uri`, `hono`, `js-yaml`, `nanoid`, `react-router`, `react-router-dom` are all at identical versions before and after, so this PR introduces none of them.

**Not verified locally:** Docker was not running on this machine, so the e2e matrix, the CLI image smoke test, and the e2e-runner image build were not exercised. All three failed on #1671 only at `npm ci`, which is fixed here. The **`e2e test grafana@dev-preview-react19`** job is the signal that matters most on this PR — it is the only one that runs the plugin against a React 19 host.

## CI warnings

**Lint is now at zero warnings on this branch, against 24 on `main`.**

Two commits, in two passes.

**Pass 1 — warnings this bump introduced (25).** `@types/react` 19 deprecates two things the codebase used:

- **`FormEvent` → `SubmitEvent`** (22 sites). The deprecation message is literally "FormEvent doesn't actually exist"; `@types/react` 19 adds a real `SubmitEvent` and types `onSubmit` as `SubmitEventHandler`. All 22 sites are the sole `<form onSubmit>` handler in their file, so the rename is direct.
- **`MutableRefObject` → `RefObject`** (3 sites), now that `RefObject` is itself mutable. All three come from a `useRef` with an initial value, so they were already `RefObject` in practice.

**Pass 2 — the 5 warnings that predated the branch.**

The interesting one is `_recordAutoLaunchSource` (3 sites). Its deprecation note carved out an exception for "`loadDocsTabContent` callers that don't go through `openDocsPage`" — so suppressing those three would have been defensible. But that exception only existed because `loadTab` had no way to carry a launch source. `loadTab` already took an options object, so it now accepts `source` and records it exactly as `openDocsPage` does. The three callers (restored-tab init, dev-mode reload, `useContentReset`'s reset path) pass `{ source }` instead of stashing it out-of-band, and the two internal bridges inside the panel assign the private carrier directly — which retires both existing `eslint-disable @typescript-eslint/no-deprecated` comments as well. The public method stays for callers that genuinely cannot carry a source across a callback boundary owned elsewhere, with its docs narrowed to that case.

Worth reviewing as a behaviour change rather than a lint fix. It is covered behaviourally, not just by mock shape: deleting the new `loadTab` bridge makes `docs-panel.alignment.test.ts` fail with the restored-tab case classified as `launchSource: "unknown"`, raising the spurious alignment prompt the tagging exists to prevent.

The other two:

- **`ImportClause.isTypeOnly` → `phaseModifier === ts.SyntaxKind.TypeKeyword`** (deprecated in TypeScript 6, which is what this repo resolves). This one needed care: `phaseModifier` is *also* set for `import defer`, which still evaluates the module at runtime. A naive `phaseModifier !== undefined` check would classify a deferred import as erased and let a Node-unsafe import skip `architecture.test.ts`'s reachability walk. I verified the comparison against `import type X`, `import type { A }`, inline `{ type A }`, mixed `{ type A, B }`, plain imports, and `import defer * as N` — all seven classify correctly.
- **`ConfirmModal`'s `confirmButtonVariant` → `confirmVariant`.** Same type, and the component resolves `confirmButtonVariant ?? confirmVariant`, so it is behaviour-preserving.

**Deliberately out of scope:**

- `Test` job: `No files were found with the provided path: coverage/`. Pre-existing on `main` and unrelated to this bump — the job runs the intentionally coverage-free `npm run test:ci` and then unconditionally uploads `coverage/`. Nothing consumes the artifact. Being fixed separately, since it needs a decision between deleting the step and making CI enforce coverage.
- `Node.js 20 is deprecated` on ~11 jobs — pinned third-party action digests, which renovate owns.

One note: `npm run test:coverage` occasionally prints Jest's `A worker process has failed to exit gracefully`. It is not deterministic — back-to-back runs on an identical tree gave 1 then 0 — so it is teardown timing, not a leak from this branch. It does not fail the run.

## Breaking changes

None for guide authors or the plugin API. `grafanaDependency` stays at `>=12.3.0-0`: `react`, `react-dom`, and the `@grafana/*` runtime packages are all webpack externals, and `src/lib/create-root-compat.ts` already resolves `react-dom/client` at runtime with a `ReactDOM.render` fallback, so React 18 hosts (Grafana 12.3 through 13.1 in the e2e matrix) keep working.

## Reversibility

Fully reversible — revert the commit. No migrations, no persisted state, no flag to retire.


moxious (approved):
**Approve with minor.** No blocking concerns. I ran the routed review (5 always-on concerns + the standalone security lens + 11 routed subsystem concerns, plus the contract-evolution scan on `docs-retrieval-and-rendering`), and five medium-severity candidates all died under adversarial verification because a load-bearing premise turned out to be false. The PR body's account holds up on every claim I spot-checked. Details below, then a handful of low-severity ride-alongs.

## What I verified independently

- **`confirmVariant`** exists on `@grafana/ui`'s `ConfirmModal` back to 12.3.0 — I read the published `.d.ts` for 12.3.0, 12.4.9, 13.1.3 and 13.2.0. `confirmButtonVariant` has been `@deprecated` since 13.1.3. Safe across the whole declared host range, so this one is not a 13.2-only prop.
- **`React.SubmitEvent<T>`** is real in @types/react 19 (`extends SyntheticEvent<T, NativeSubmitEvent>`) and `onSubmit` is typed `SubmitEventHandler<T>`. All 22 renamed handlers are wired to exactly one `<form onSubmit>` each and to nothing else — no `onClick`, no imperative call with a synthesized event. `SectionBlockForm`'s separate zero-arg `handleSubmitAndRecord` on an `onClick` is a different function, so it is a non-issue.
- **The `overrides` block is load-bearing and correct.** Deleting just those two keys and re-resolving reproduces the exact `ERESOLVE` that killed #1671. Re-resolving with root `react` at 19.2.7 tracks it, so `$react` needs no maintenance on future bumps. Scenes' `react@^18` cap persists even in `9.0.0-canary`, so there is no exit horizon worth filing. Renovate will not fight it — `$react` is not a parseable version, so the npm manager skips it.
- **Supply chain:** the `npm audit` advisory set is byte-identical base vs head — same 8 advisories, same packages, same severities, empty set difference in both directions. Your "introduces none of them" claim is exact, not just equal in count. No non-registry resolved hosts, no missing integrity, no new install scripts, single `react` copy at 19.2.8.
- **React 19 runtime hazards:** exactly one ref callback exists repo-wide (`AuthorNoteModal.tsx:84`) and it has a block body, so the ref-cleanup-return error is unreachable. No removed API (`findDOMNode`, `unmountComponentAtNode`, `react-dom/test-utils`, function-component `defaultProps`) and no React-19-*only* API anywhere in `src`, which is what keeps the bundle runnable on React 18 hosts alongside `create-root-compat.ts`.
- **The step-checker rewrite is tighter, not weaker** — I reproduced your mutation test: deleting `.finally(clearTimeout)` fails it on the `toContain` assertion. The `delay === 3000` match is also not a flake risk; I instrumented the observation window and captured exactly `[3000]`, because the same-delay `requirements.heartbeat` effect early-returns when the test passes no `requirements`.
- **The launch-source rework is behaviour-preserving.** I built the "stale source leaks past the guide arm" repro and ran it at both base and head — identical output, because `_recordAutoLaunchSource` was always a bare setter and `loadTab` assigns the field as its first synchronous statement. Worth saying out loud: at head the deprecated setter has **zero** production callers where base had five out-of-band writers, so this converges the carrier rather than branching it. The contract-evolution scan initially graded that `contract_missing`; all three skeptics correctly reduced it to `coherent_extension`, and no contract anchor is owed since `OpenDocsOptions.source` predates the PR.
- **Bundle:** +1,533 B on `module.js` (+1.31%), matching the `compare` job to within 4 B. Exactly one new package reaches `dist/` (`@grafana/rrweb-types`, in a lazy chunk). Code-splitting parity is exact: 30 `await import`, 15 `lazy(` on both sides.
- **Architecture ratchets** produce byte-identical verdicts base vs head, and `src/validation` is 840/840.

## Low-severity, none blocking

1. `[suggestion]` **`.cursor/rules/techContext.mdc:44` still says `react` + `react-dom` (18.3.1).** This is the one version entry the PR makes false, and it is now the only React-18 assertion in the docs tree outside CHANGELOG history. Since you are already editing line 14 of that file, worth fixing line 42 to `(13.2.0 / 8.13.4)` and the `typescript` entries on lines 14 and 55 to `6.0.3` in the same pass — the TS 6 API is exactly what the `phaseModifier` change depends on, so an agent reading `5.9.3` there could "fix" that change back.
2. `[suggestion]` **The `phaseModifier === TypeKeyword` guard has no test.** I mutated it to `phaseModifier !== undefined` — the precise regression the new comment warns about — and `import-graph.test.ts` + `architecture.test.ts` stayed green, 156/156. Type-only edges are exempted from the Node-reachability closure and the cycle ratchet, so a misclassification makes those ratchets pass *vacuously* rather than fail. One assertion closes it: `expect(recordFor("import defer * as m from './mod';", './mod').typeOnly).toBe(false)`. (For the record, the change itself is an exact no-op — TS 6's `createImportClause2` sets `isTypeOnly = phaseModifier === TypeKeyword` — so the comment is justifying the `=== TypeKeyword` form over a truthiness test, which is the right thing to explain.)
3. `[nit]` **`src/interactive-engine/use-sequential-step-state.hook.ts:6` still says "React 18 hook".** You fixed that exact phrase in `docs/developer/engines/interactive-engine.md:20`; the source JSDoc it describes still carries it, so doc and code now disagree. The next line already states the mechanism, so the version reference can just go.
4. `[nit]` **`.cursor/rules/react-antipatterns/hooks-and-effects.mdc` R15** uses zero-argument `useRef()` in both its Don't and Do examples (lines 236, 253, 254). That no longer compiles under @types/react 19 — the same break you fixed at 5 real call sites — and R15 is a rule agents copy verbatim.
5. `[suggestion]` **Document the scenes override where agents will look.** It is a load-bearing install-time constraint that nothing records, so the next person to tidy `overrides` will delete it and rediscover the ERESOLVE. A bullet under `techContext.mdc` §Technical Constraints would do: *"Single React copy: `@grafana/scenes` still declares a `react@^18` peer, so `package.json` pins `overrides["@grafana/scenes"].react`/`.react-dom` to `$react`. Both entries are load-bearing — dropping either breaks `npm install`."*
6. `[suggestion]` **`_recordAutoLaunchSource` is now dead on the public interface.** Zero production callers at head, still declared on `DocsPanelModelOperations`, and its retained-for rationale ("`ContextPanel`'s recommender callbacks") is contradicted at `docs-panel.tsx:217,220`, which pass `{ source: 'recommender' }` through the options bag. Either delete it and repoint `docs-panel.alignment.test.ts:720-728` at `openDocsPage(url, title, { source })`, or restate a rationale head state supports. Fine as a follow-up.
7. `[nit]` **`useContentReset.test.ts:118-127` is now a byte-identical duplicate** of the assertion 15 lines above it — the ordering property it used to guard became unrepresentable once the source rides in the arguments, which is the right call. Delete it, or repoint it at something still uncovered.

## Two things worth knowing, not asks

- `[question]` **Commit 3 contains nothing this bump forced.** All three of its edits target deprecations that predate the PR: TS 6's `isTypeOnly` (`typescript` was already `^6.0.3` on main), `confirmButtonVariant` (deprecated as of 13.1.3, the *base* version), and the repo's own `_recordAutoLaunchSource`. `npm run lint` carries no `--max-warnings`, so none of them was failing. It is also where the one public signature change lives — `loadTab` gaining `options.source`. Happy either way, but given the repo's one-ratchet-per-PR preference, either name the launch-source change in the PR body as an extension of the existing contract, or split commit 3 out. It is cleanly separable.
- **`@grafana/i18n` is aliased into the bundle**, so its transitive `react-i18next` 17 (peer `i18next >= 26.2.0`) now sits a major above the `i18next` that hosts below 13.2.0 supply, since `i18next` itself is a host-provided external. Unexercised today — no `<Trans>` and no multi-namespace `useTranslation([...])` in `src` — and the green 12.3.11/12.4.9 jobs boot `initPluginTranslations` on the real host, so nothing to do here. Flagging it as a tripwire for the next `@grafana/i18n` bump.

Also noted, unchanged by this PR: `playwright-tests` is absent from `ci-gate`'s `needs:`, so the e2e matrix — the only React-19-host signal — is advisory and cannot block merge. Moot with all seven green, but it means the matrix protects us only as long as someone reads it.

The `dev-preview-react19` job plus 12.3.11 and 12.4.9 all passing is the evidence that matters here, and it is in hand. Nice work on the warning-baseline measurement in particular — measuring against a pristine `main` worktree instead of trusting GitHub's capped annotation list is what made that 24 → 5 number trustworthy.

tomglenn (changes requested):
The migration itself looks right, and the framing in the description holds up. I checked out the head, installed, and ran the gates: `typecheck`, `lint` (0 warnings), `prettier-test`, `docs:sync-terms:check`, `test:ci` (465 suites / 7441 tests), `test:coverage`, `test:scripts`, `test:review-contract`, `go build ./...`, `test:go`, and `build` all pass. I also confirmed the parts that types and CI can't cover:

- `react` / `react-dom` are genuinely external — `dist/module.js` emits them in the AMD `define([...])` list, and grepping `dist/*.js` for React internals markers (`ReactCurrentOwner`, `__SECRET_INTERNALS`) returns nothing. So the `@grafana/scenes` peer override is install-time only, as you argued.
- `@grafana/{data,ui,runtime}` 13.2.0 do declare `react: ">=19"` / `react-dom: ">=19"` (and `i18n` declares `react: ">=19"`), so with `react` pinned to exactly `18.3.1` the `ERESOLVE` was unavoidable. The bump and the migration really can't be split.
- `confirmVariant` already exists in `@grafana/ui@12.3.11`, below the `>=12.3.0-0` floor, so the `ConfirmDeleteButton` rename is safe across the whole supported host range. I checked this one specifically because `@grafana/ui` is external, so local types prove nothing about an older host.
- No React 19-only APIs anywhere in `src/`, which matters more than the usual direction of a migration check: those would be `undefined` on the React 18 hosts still in the matrix.
- The ref callbacks that matter all return `undefined` — including `makeRefCallback` in `interactive-section.tsx:1197` — so React 19 keeps the legacy null-on-unmount path and step refs are still deleted.

Three requested changes below. None of them is a production defect, and none changes runtime behaviour. Two are test/comment accuracy and one is merge mechanics.

---

## 1. The symmetry tripwire's `ref` invariant is no longer true under React 19

**`src/components/interactive-tutorial/interactive-section.symmetry.tripwire.test.tsx:21-24` and `:173-176`**

Both comments state that React strips `ref` from `props`:

```
:21  Note on `ref` and `key`: React strips these from `props` before
:22  passing them to the child component, ...

:173 `ref`, `key`, and `isCompleted` are intentionally excluded — React
:174 handles `ref`/`key` specially and they do not appear in the captured
:175 `props`.
```

Under React 19 that holds for `forwardRef` components but not for plain function components, and `makeCapture` (`:37`) returns a plain `React.FC<any>`. I instrumented `makeCapture` on this branch to check:

```
PROBE_REF_IN_PROPS InteractiveStep      …,disabled,resetTrigger,ref
PROBE_REF_IN_PROPS InteractiveMultiStep …,disabled,resetTrigger,ref
PROBE_REF_IN_PROPS InteractiveGuided    …,disabled,resetTrigger,ref
PROBE_REF_IN_PROPS CodeBlockStep        …,disabled,resetTrigger,ref
```

And confirmed the contrast directly on React 19.2.8 — plain FC gets `["a","ref"]`, `forwardRef` gets `["a"]` plus the ref as its second argument.

**Why it matters:** the mocks now differ from the real components by exactly `ref`. The production step components are `forwardRef`, so they still receive `ref` as the second argument and never in `props`. That makes the file's claim that "the forwardable-prop surface is fully observable" weaker than it reads, which is a shame for a test built specifically as a tripwire. `expectSurface` (`:212`) only asserts inclusion (`key in props`), never exclusion, so nothing fails and CI can't flag the drift.

**Suggested fix:** update both comments to say `key` is stripped while `ref` is a normal prop on function components under React 19, and note that the real components are `forwardRef` so they don't see it in `props`. If you want the tripwire to actually hold the line, adding `expect('ref' in props).toBe(false)` against a `forwardRef` mock would pin the distinction that matters.

**Also at `:39-45`:** the `props === undefined` guard is dead code now. Its comment says it exists for React 18's dev-mode `validateFunctionComponentInDev` probe; React 19 has no such probe, and no warning is emitted. Worth deleting along with its comment so it doesn't read as a live concern.

## 2. `import defer` isn't pinned by a test, and the naive form passes everything

**`src/validation/import-graph.ts:271-273`**

To be clear about what is and isn't covered here: the `phaseModifier === ts.SyntaxKind.TypeKeyword` change *is* covered. I disabled the branch and three tests failed, including `scanNodeEnvReachability › exempts type-only imports`:

```
● extractImportRecords › detects `import type` clauses as typeOnly
● extractImportRecords › merges occurrences: a type-only import does not make a dynamic value edge eager
● scanNodeEnvReachability › on synthetic fixtures › exempts type-only imports
```

I also verified your classification independently against TypeScript 6.0.3 across eight import forms — `import type X`, `import type { A }`, `{ type A }`, `{ type A, B }`, plain named, plain default, `import defer * as N`, and `import * as N`. It's exactly equivalent to the old `isTypeOnly` on all eight, and `import defer` correctly comes back `phaseModifier=DeferKeyword` → not type-only. The reasoning in the description is right.

**What's missing is the guard on the mistake you identified.** I replaced the comparison with the naive `clause.phaseModifier !== undefined` and ran both `import-graph.test.ts` and `architecture.test.ts`:

```
Test Suites: 2 passed, 2 total
Tests:       156 passed, 156 total
```

All 156 pass. So the one distinction the comment at `:271-272` exists to preserve is the one thing no test holds.

**Why it matters:** if someone later simplifies this to `!== undefined` — which reads as equivalent and is the obvious tidy-up — a deferred import gets classified as erased, and a Node-unsafe import can then skip `architecture.test.ts`'s reachability walk silently. Low likelihood, but the failure mode is a quiet hole in that guard rather than a visible break.

**Suggested fix:** one case alongside the existing ones near `import-graph.test.ts:714`:

```ts
it('treats `import defer` as a value import', () => {
  expect(recordFor(`import defer * as N from './m';`, './m').typeOnly).toBe(false);
});
```

That pins the comment's claim and makes the naive refactor fail loudly.

## 3. Commit `b48d7d93` isn't part of the bump, which affects the stated reversibility

The three commits split cleanly by cause:

| Commit | Caused by the bump? |
|---|---|
| `041fd775` — bump + React 19 type migration | yes |
| `606bd460` — `FormEvent`→`SubmitEvent`, `MutableRefObject`→`RefObject` | yes, new `@types/react` 19 deprecations |
| `b48d7d93` — `ConfirmDeleteButton` variant, `loadTab({ source })` refactor, `import-graph` TS 6 fix | no, pre-existing warnings |

All three of `b48d7d93`'s changes look correct to me, and the `loadTab` one is properly covered — I deleted the new bridge in `docs-panel.tsx:526-528` and `docs-panel.alignment.test.ts` failed with `launchSource: "unknown"` and the spurious `pendingAlignment`, exactly as you described. I also confirmed `_recordAutoLaunchSource` is a plain setter (`docs-panel.tsx:181-183`), so the direct field assignments are equivalent, and the `if (options?.source)` guard stops a sourceless caller clobbering a stashed value.

**Why it matters:** the description says "Fully reversible — revert the commit", but if this is squash-merged, reverting the React 19 bump also reverts the `import-graph` TypeScript 6 fix and the launch-source refactor — two changes with independent value and no relation to React.

**Suggested fix:** merge preserving the three commits rather than squashing, so each reverts on its own. If you'd rather squash, splitting `b48d7d93` into its own PR gets the same outcome — that's a maintainer call on how much a round trip is worth, not something I'd push either way.

---

One thing worth flagging that isn't a requested change: the e2e matrix is reassuring but thinner than 36 green checks suggests for this particular bump. It's 22 tests across 6 specs, and the interactive tutorial's "Show me" / "Do it" automation — the surface most sensitive to React ref and effect semantics — is covered by one test (`tests/welcome-journey.spec.ts:132`, first two steps). Not a regression from this PR, and `skip-grafana-react-19-preview-image: false` means the React 19 host was already in the matrix before it, so this closes a gap rather than opening one. Just worth knowing what the green is actually load-bearing for.

PR #1679, opened 2026-08-21
fix(block-editor): compile-check branch block builders

### LINKED ISSUE #1655: Make the branch add-picker's block-type list compile-checked against createDefaultBlock
#1642 fixed #1542 by replacing the derived `ALLOWED_BRANCH_BLOCK_TYPES` with a hand-written literal list. That is a strict improvement — the six broken types are gone, and a newly added block type now fails closed (absent from the picker) instead of failing open (silent empty markdown). But the invariant the list depends on is still enforced by nothing.

## The invariant

Every type offered in `BranchBlocksEditor`'s add picker must have a non-`default` arm in `createDefaultBlock`. If it doesn't, `handleAddBlock` → `buildBlockFromForm` → `createDefaultBlock` falls through to `{ type: 'markdown', content: '' }` and the author silently gets an empty markdown block. That is exactly #1542.

Today nothing checks it. Delete `case 'guided':` from `createDefaultBlock` (`BranchBlocksEditor.tsx:290-291`) and `'guided'` stays in the list at `:311`, still typechecks, still lints, still passes tests — and #1542 is live again.

## There are two unguarded entry points, not one

`addableBlockTypes` is a **prop** (`BranchBlocksEditor.tsx:362`), so the default list is not the only source:

- `BranchBlocksEditor.tsx:372` — defaults to `ALLOWED_BRANCH_BLOCK_TYPES`
- `CollapsibleBlockForm.tsx:90` — passes its own `PRESENTATIONAL_ADDABLE_TYPES` (`CollapsibleBlockForm.tsx:20`)

`PRESENTATIONAL_ADDABLE_TYPES` is `['markdown', 'image', 'video']` — all three have `createDefaultBlock` arms, so **there is no live bug here today**. But it is a second hand-maintained list feeding the same picker with no guard, and the prop's type (`BlockType[]`) permits any caller to pass a type that has no builder.

So the check belongs on the prop. That covers both call sites and every future one at once.

## Fix

Mirror the idiom already in this subsystem — `_palettePartitionIsTotal` in `src/components/block-editor/constants.ts:196-206` — which does exactly this for the palette.

Make the set of buildable types a type, then constrain both the lists and the prop to it:

```ts
const BLOCK_DEFAULT_BUILDERS = {
  markdown: () => ({ type: 'markdown', content: '' }),
  interactive: () => ({ type: 'interactive', action: 'highlight', reftarget: '', content: '' }),
  image: () => ({ type: 'image', src: '' }),
  video: () => ({ type: 'video', src: '' }),
  input: () => ({ type: 'input', prompt: '', inputType: 'text', variableName: '' }),
  quiz: () => ({ /* … */ }),
  multistep: () => ({ type: 'multistep', content: '', steps: [] }),
  guided: () => ({ type: 'guided', content: '', steps: [] }),
} as const satisfies Partial<Record<BlockType, () => JsonBlock>>;

/** Block types the inline branch editor can construct. */
export type DefaultableBlockType = keyof typeof BLOCK_DEFAULT_BUILDERS;
```

Then:

- `ALLOWED_BRANCH_BLOCK_TYPES: readonly DefaultableBlockType[]`
- `PRESENTATIONAL_ADDABLE_TYPES: readonly DefaultableBlockType[]`
- `addableBlockTypes?: readonly DefaultableBlockType[]`

Now removing a builder is a compile error at every list that names it, and no caller can pass a type the editor cannot build.

Keeping the `switch` and adding a standalone type-level assertion is equally acceptable if it reads better — the requirement is that the drift is caught at compile time, not the specific shape.

The sketch above was checked against this repo's types: it compiles as written, and deleting the `guided` builder produces `TS2322: Type '"guided"' is not assignable to type '"input" | "video" | … | "quiz"'` on the list that still names it.

## Cleanups this forces

These all live in the same ~20 lines and should land in the same PR rather than separately:

1. **`BranchBlocksEditor.tsx:299-302`** — the comment is wrong today. It says types needing dedicated forms `(challenge, quiz, terminal, …) are also excluded`, but `quiz` is in the list at `:309`, as are `multistep` and `guided`. The real rule is "types `createDefaultBlock` can build", which has nothing to do with inline forms. Once the type expresses the rule, most of this comment can go.
2. **`BranchBlocksEditor.tsx:303-314`** — `ALLOWED_BRANCH_BLOCK_TYPES` is a pure alias of `BRANCH_INLINE_CREATABLE_TYPES`; the private name has no other consumer. Collapse to one const.
3. **`BranchBlocksEditor.test.ts:18-20`** — `expect(createDefaultBlock('html' as any)).toEqual({ type: 'markdown', content: '' })` pins the silent fallback as intended behaviour, which is the root cause #1542 identified. Once the picker can only offer buildable types, this test contradicts the design; delete it. (The `as any` is also unnecessary — `html` is a valid `BlockType`.)
4. **`BranchBlocksEditor.test.ts`** — the surviving assertion only checks `challenge` is absent. Five other types were removed in #1642 (`code-block`, `terminal`, `terminal-connect`, `grot-guide`, `snippet-ref`) and nothing pins them. A single `toEqual` on the whole array covers the contract; with the type-level guard in place it becomes a belt-and-braces check rather than the only one.

## Out of scope

Narrowing `createDefaultBlock`'s own parameter to `DefaultableBlockType` and replacing its `default:` arm with a `never` exhaustiveness check. It is reachable only via `buildBlockFromForm`'s `default:` (`:501-502`), which is typed `BlockType`, so that narrowing pulls in a second signature change. Worth doing, but it is a separate step — do not let it block this one.

## Acceptance

- Deleting any arm from `createDefaultBlock` / `BLOCK_DEFAULT_BUILDERS` produces a **typecheck error** at the list(s) that still name that type. Verify by hand: delete `guided`, confirm `npm run typecheck` fails, restore.
- Passing a non-buildable type to `addableBlockTypes` is a compile error.
- No runtime behaviour change: the picker offers the same eight types in branches and the same three in collapsibles.
- `npm run check` passes.

## Notes

Same failure class as #1523 and the wider type-design hardening under #603: a hand-maintained per-block-type registry where a missing key is legal TypeScript and surfaces as wrong runtime behaviour. `constants.ts:196-206` is the in-repo exemplar of the fix.

Refs #1642, #1542, #1523, #603

Line numbers are as of the merge of #1642.

mojidev7:
## Summary

- derive the branch editor's constructible block type from its default-builder registry
- constrain both add-picker lists and the public `addableBlockTypes` prop to constructible types
- remove the permissive unknown-type fallback that silently produced empty markdown
- add compile-time contracts and a runtime whole-picker regression test

Fixes #1655

## Root cause

The picker lists, public prop, and `createDefaultBlock` all accepted the broad `BlockType` union independently. TypeScript therefore could not detect when a picker exposed a type without an intentional builder, allowing the silent empty-markdown behavior from #1542 to return.

## Verification

- `npm run typecheck`
- `npx jest src/components/block-editor/forms/BranchBlocksEditor.test.ts --coverage=false --runInBand`
- `git diff --check origin/main...HEAD`

The focused suite passes both regression tests. A broader local `npm run test:ci` run passed 7,325 tests across 457 suites; the only unrelated failure was the `node:test` builtin assertion under local Node 22, while this repository requires Node 24+. The full `npm run check` also reached the Go lint stage but could not run it because `mage` is not installed locally.

## Risk

Low. The change is limited to the branch block editor's type contract and preserves the existing picker entries and generated block shapes.

### DIFF
--- a/src/components/block-editor/forms/BranchBlocksEditor.test.ts
+++ b/src/components/block-editor/forms/BranchBlocksEditor.test.ts
@@ -1,8 +1,4 @@
-import {
-  ALLOWED_BRANCH_BLOCK_TYPES,
-  createDefaultBlock,
-  type BranchBlocksEditorProps,
-} from './BranchBlocksEditor';
+import { ALLOWED_BRANCH_BLOCK_TYPES, createDefaultBlock, type BranchBlocksEditorProps } from './BranchBlocksEditor';
 
 const validProps: BranchBlocksEditorProps = {
   label: 'Branch',
@@ -20,22 +16,28 @@ const invalidProps: BranchBlocksEditorProps = {
 
 void invalidProps;
 
+if (false) {
+  // @ts-expect-error collapsible has no default builder
+  createDefaultBlock('collapsible');
+}
+
 describe('BranchBlocksEditor createDefaultBlock', () => {
-  it('does not offer challenge in the inline add picker', () => {
-    expect(ALLOWED_BRANCH_BLOCK_TYPES).not.toContain('challenge');
+  it('offers exactly the block types the inline picker can build', () => {
+    expect(ALLOWED_BRANCH_BLOCK_TYPES).toEqual([
+      'markdown',
+      'interactive',
+      'image',
+      'video',
+      'input',
+      'quiz',
+      'multistep',
+      'guided',
+    ]);
   });
 
-  it('builds a challenge block instead of empty markdown if challenged', () => {
-    expect(createDefaultBlock('challenge')).toEqual({
-      type: 'challenge',
-      title: '',
-      brief: '',
-      successCriteria: '',
-    });
-  });
-
-  it('still defaults unknown types to empty markdown', () => {
-    // html is legacy / palette-excluded and not in the creatable list
-    expect(createDefaultBlock('html' as any)).toEqual({ type: 'markdown', content: '' });
+  it('constructs every offered block without changing its type', () => {
+    for (const type of ALLOWED_BRANCH_BLOCK_TYPES) {
+      expect(createDefaultBlock(type).type).toBe(type);
+    }
   });
 });
--- a/src/components/block-editor/forms/BranchBlocksEditor.test.ts
+++ b/src/components/block-editor/forms/BranchBlocksEditor.test.ts
@@ -1,4 +1,24 @@
-import { ALLOWED_BRANCH_BLOCK_TYPES, createDefaultBlock } from './BranchBlocksEditor';
+import {
+  ALLOWED_BRANCH_BLOCK_TYPES,
+  createDefaultBlock,
+  type BranchBlocksEditorProps,
+} from './BranchBlocksEditor';
+
+const validProps: BranchBlocksEditorProps = {
+  label: 'Branch',
+  variant: 'success',
+  blocks: [],
+  onChange: () => undefined,
+  addableBlockTypes: ['markdown', 'guided'],
+};
+
+const invalidProps: BranchBlocksEditorProps = {
+  ...validProps,
+  // @ts-expect-error challenge has a dedicated editor and cannot be built by the branch add picker
+  addableBlockTypes: ['challenge'],
+};
+
+void invalidProps;
 
 describe('BranchBlocksEditor createDefaultBlock', () => {
   it('does not offer challenge in the inline add picker', () => {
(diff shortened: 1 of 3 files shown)


moxious:
hi @mojidev7 , can you please have a look at this PR and resolve conflicts?  This PR came in at a time when we were already working on something similar in the blast radius; can review once it's deconflicted.

moxious:
hi @mojidev7 -- first, thank you for this contribution, and we would like to get it in, but as we haven't seen activity on this PR I'm going to close it for now.  I'm still willing to review it and provide feedback, we do appreciate contributions to this repo.  If you're available to keep working on it, you can feel free to push additional commits and re-open.

Issue #1681, opened 2026-08-21
A guide opened from inside a learning path never gets its alignment prompt: the complete manifest is shadowed by the stripped catalogue one


tomglenn:
## What a user sees

Open a guide from **inside a learning path** — click a member in a path card under
Custom guides — and it never shows the "this guide starts on page X, go there first"
prompt, even when that guide declares a starting location.

Open the **same guide** standalone from the Custom guides list and the prompt appears
correctly.

So the alignment prompt is present or absent depending on which card you clicked, not
on anything about the guide.

## Why

The member's starting location **arrives intact and is then shadowed.** It is not
missing.

1. `openMember` opens `member.url` (`src/components/docs-panel/CustomGuidesSection.tsx:116`).
   That URL is `backend-guide:<memberId>` (`src/package-engine/app-platform-resolver.ts:224`),
   so the content comes through the `backend-guide:` loader.
2. That loader spreads `spec.manifest` through unchanged
   (`src/docs-retrieval/content-fetcher/backend-guide.ts`), so
   `content.metadata.packageManifest` **does** carry
   `additionalFields.startingLocation`.
3. But `openMember` also passes `packageInfo`, and the reader prefers it:

   ```ts
   // src/components/docs-panel/docs-panel.tsx:925
   const startingLocation = resolveStartingLocation(
     url,
     packageInfo?.packageManifest ?? fetchedContent.metadata.packageManifest
   );
   ```

   `packageInfo.packageManifest` is built from the catalogue-proxy entry, whose
   `customGuideManifest` (`pkg/plugin/custom_guide_repository_client.go:48`) declares
   no `additionalFields` — so `encoding/json` drops the key at the wire boundary.

The result is a manifest that is present but stripped, winning over one that is
present and complete. `resolveStartingLocation` never sees the real value.

## Why promoting `startingLocation` to a typed CUE field does not fix this

There is separate work queued to promote `startingLocation` out of
`additionalFields` into a real field on `#Manifest` in `grafana-pathfinder-backend`.
That fixes **path cover pages**, whose value genuinely is dropped at the wire.

It does **not** fix members. A member's problem is precedence, not transport: the
complete manifest is already in hand and is being passed over. Promoting the field
changes which manifest carries it, not which manifest is read first.

Worth stating plainly because the two cases look identical from the outside — both
"a path-related guide gets no prompt" — and it would be easy to close this by
promoting the CUE field and assuming both were covered.

## Deliberately not done

[...]

Issue #1678, opened 2026-08-21
Editor saves stamp schemaVersion 1.0 while the editor emits 1.1.0


tomglenn:
## Summary

Every guide the block editor saves to App Platform is stamped `spec.schemaVersion: "1.0"`, while the
editor actually emits schema `1.1.0`.

## Evidence

`getGuide()` returns no `schemaVersion` (`src/components/block-editor/hooks/useBlockEditor.ts:1144-1153`):

```ts
const getGuide = useCallback((): JsonGuide => {
  return {
    id: state.guide.id,
    title: state.guide.title,
    blocks: state.blocks.map((b) => b.block),
  };
}, [state.guide, state.blocks]);
```

so `preservedSpec` always takes its fallback (`hooks/useBackendGuides.ts:79`):

```ts
// Normalised deliberately: the editor writes the schema version it emits, not the stored one.
schemaVersion: guide.schemaVersion || '1.0',
```

But `CURRENT_SCHEMA_VERSION` is `'1.1.0'` (`src/types/json-guide.schema.ts:1103`).

Confirmed on the wire by driving the real `saveGuide` in a spike and dumping the request body:

```json
"spec": {
  "id": "my-private-guide",
  "title": "My private guide",
  "schemaVersion": "1.0",
  ...
}
```

The comment on that line is correct about intent — the editor *should* write the version it emits —
it just names the wrong constant.

## Impact

Low today: nothing on the App Platform read path branches on `spec.schemaVersion`. It matters for the
CLI's cross-file drift check if a guide is ever round-tripped out to a package directory, and it makes
every editor-authored guide misreport which schema produced it, which is exactly the question a
`schemaVersion` exists to answer.

## Suggested fix

Import `CURRENT_SCHEMA_VERSION` and use it as the fallback. Small, but it changes what **every**
future save writes, so it deserves its own commit rather than riding along in an unrelated PR.

Existing guides keep their stored `"1.0"` until re-saved, which is correct — the value should describe
the schema that produced the blocks on the resource.

## Provenance

Found while designing block-editor manifest support (scout `pf-editor-manifest-s1`).

moxious:
Closed by merge #1682.

Issue #1666, opened 2026-08-20
guide-stats: declare manifest.stats positively as a tier-0 Zod schema


moxious:
`manifest.json` gained a persisted `stats` field, written by `pathfinder-cli build-stats`. Its shape exists only as a writer-side TypeScript interface (`GuideStatsSummary` in `src/lib/guide-stats/summary.ts`):

- there is no runtime validation of a `stats` object a consumer reads back, and
- there is no schema a cross-language port (the planned Go consumer) can generate from.

The fix is a positive declaration:

- `GuideStatsSummarySchema` in `src/types/` (tier 0), with `stats: GuideStatsSummarySchema.optional()` on the manifest schema.
- `GuideStatsSummary` derived from it via `z.infer`, so the writer and the schema cannot drift.

Tier 0 is the right home: `package.schema.ts` importing from `lib/` would be an upward tier import, so the shared definition has to land below both.

Raised in review of #1661 and deferred to the first consumer PR, which is where the read side gets exercised.

moxious:
Two documentation-drift dimensions surfaced while landing #1661, both of which this issue would close:

- The `stats` field list in `docs/developer/CLI_TOOLS.md` (`version`, `blockCount`, `sectionCount`, `completableBlockCount`, `finalCompletablePosition`) is hand-copied from the `GuideStatsSummary` interface, because `stats` has no schema to generate from. Nothing drift-checks the doc against the type.
- #1661 had to rename two of those keys (`interactiveBlockCount` → `completableBlockCount`, `finalInteractivePosition` → `finalCompletablePosition`) and update every prose copy by hand. A generated field table would have made that mechanical.

Worth regenerating or drift-checking the doc rows from the schema when it lands.

PR #1682, opened 2026-08-21
feat(block-editor): stamp a derived manifest on editor-authored guides

tomglenn:
## What changed

The block editor now stamps a derived `spec.manifest` on every save, so a guide it authors is a complete package rather than bare content.

The manifest is **merged over** the one last read, never replaced — that is what keeps a learning path cover page's `type` and `milestones` when its page is edited (#1599, #1663). The editor owns only fields it can compute from the blocks:

| field | rule |
| --- | --- |
| `type` | `guide` when nothing was inherited; an inherited value is never overwritten |
| `repository` | `app-platform`, and only when minting a fresh manifest |
| `additionalFields.stats` | the block-count stamp from `src/lib/guide-stats`, for a plain `guide` only — a path's stats are a rollup over its milestones, which a cover page cannot see |
| `additionalFields.startingLocation` | the first absolute `on-page:` requirement in document order, stopping at the first navigation. Written only when provenance shows the editor authored the inherited value, so a title-only edit cannot erase an externally authored one |

Both derived values sit under `additionalFields` because the `InteractiveGuide` CRD's `#Manifest` declares eight keys and prunes anything else — the write returns 201 with a `Warning:` header the frontend never surfaces, so a top-level `stats` would report success and vanish. `resolveStartingLocation` reads both locations.

Picked up along the way:

- `manifest.stats` is now a tier-0 Zod schema with `GuideStatsSummary` inferred from it, so the writer and the validator cannot drift (#1666).
- `preservedSpec` stopped hard-coding `schemaVersion: '1.0'` (#1678).
- A manifest `startingLocation` reaches `locationService.push`, so it now goes through the same same-origin and denied-route check as an authored `navigate` action. Those steps moved into `validateInternalNavigationPath`, and `navigate-handler` calls it instead of keeping its own copy.
- `build-repository` carries `stats` onto the repository entry by name. Declaring `stats` on the manifest schema took it out of the unknown-key forwarding that used to carry it (#1662), so without this the stamp silently stopped reaching `repository.json`.
- `validate-package` warns when a manifest carries a `stats` stamp that does not match the stamp shape. The schema reads a malformed stamp as absent so `build-stats` can still repair it; the warning is what stops that leniency from also hiding the problem from the author.

## What to look at

- **`derive-manifest.ts`** — the derivation itself. Specifically the two narrowings on the `on-page:` walk (a `formfill` declaration counts only on the first executable block; the walk stops at the first navigation, including one inside a conditional branch) and the provenance rule that decides who owns `startingLocation`.
- **`app-platform-resolver.ts`** — it drops a schema-defaulted `startingLocation` when the raw `spec.manifest` declared none, so a synthetic `/` cannot shadow the real value under `additionalFields`.
- **`build-repository.test.ts`** — where the #1662 extension-forwarding tests now live, alongside the new `stats` carry.

## Scope

Path members and path covers still get no alignment prompt. The Go catalogue struct `customGuideManifest` declares no `additionalFields`, so the complete manifest is shadowed by the stripped catalogue one; that is tracked separately in #1681 and needs a backend change. This branch touches no Go — `git diff origin/main...HEAD -- pkg/` is empty.

Also not here: a Package settings modal for description and category, private path authoring, and `manifest.json` on the public GitHub-PR path (#1693). `manifest.author` is not auto-stamped.

## Verification

`npm run typecheck`, `npx eslint src`, `npx prettier --check`, and `npm run test:ci` (476 suites, 7679 tests) all pass locally on Node 24.


src/components/docs-panel/docs-panel.tsx
```diff
@@ -920,7 +920,10 @@ class CombinedLearningJourneyPanel extends SceneObjectBase<CombinedPanelState> i
 
         // Implied 0th step: decide whether to prompt the user to navigate to
         // the guide's declared starting location before step 1 begins.
-        const startingLocation = resolveStartingLocation(url, packageInfo?.packageManifest);
+        const startingLocation = resolveStartingLocation(
+          url,
+          packageInfo?.packageManifest ?? fetchedContent.metadata.packageManifest
```
moxious: **[P1] Keep the member manifest separate from path context.** The new starting-location behavior works for standalone guides but not path members. `packageManifest` can already have overwritten the fetched member's full manifest with the parent path's slim manifest, and this preference keeps that parent manifest authoritative. Centralize App Platform manifest shaping and carry separate current-guide and journey/path manifests so alignment always reads the member manifest.

src/components/block-editor/utils/derive-manifest.ts
```diff
@@ -0,0 +1,248 @@
...
+  }
+
+  // Written unconditionally, so removing the requirement removes the prompt. A
+  // skipped write would leave a stale value outliving the content that justified
+  // it, with no in-product way to clear it.
+  const startingLocation = deriveStartingLocation(guide.blocks);
+  if (startingLocation !== undefined) {
+    additionalFields.startingLocation = startingLocation;
+  } else {
+    delete additionalFields.startingLocation;
```
moxious: **[P1] Preserve externally authored `startingLocation`.** Uploaded manifests store an untyped `startingLocation` under `additionalFields`, but any editor save deletes it when the current blocks produce no replacement. A title or markdown-only edit can therefore permanently erase independently authored package metadata. Preserve inherited values unless provenance proves they were editor-derived, or expose an explicit manifest control for removal.

src/recovery/starting-location.ts
```diff
@@ -31,13 +36,30 @@ export function resolveStartingLocation(url: string, packageManifest?: Record<st
...
   return null;
 }
 
+function readAdditionalStartingLocation(packageManifest?: Record<string, unknown>): string | null {
+  const additional = packageManifest?.additionalFields;
+  if (!additional || typeof additional !== 'object' || Array.isArray(additional)) {
+    return null;
+  }
+  const value = (additional as Record<string, unknown>).startingLocation;
+  return typeof value === 'string' && value.length > 0 ? value : null;
```
moxious, line 93: **[P2] Validate `startingLocation` before navigation.** This newly reachable App Platform value is accepted as any non-empty string and later passed directly to `locationService.push`. That bypasses the protocol-relative and role-sensitive route checks used by authored navigate actions. Require a same-origin, single-leading-slash internal path and apply the established redirect-path validation before storing or pushing it.

src/components/block-editor/utils/derive-manifest.ts
```diff
@@ -0,0 +1,248 @@
...
+  }
+  return undefined;
+}
+
+/**
+ * The absolute `on-page:` this carrier declares, ignoring a `formfill`'s — see
+ * narrowing 2 above.
+ */
+function declaredOnPage(carrier: RequirementCarrier): string | undefined {
+  if (actionOf(carrier) === FORMFILL_ACTION) {
```
moxious: **[P2] Honor the first formfill entry declaration.** The editor suggests `on-page:` for a first-step formfill, and the authoring lint accepts it as satisfying the entry-location contract, but derivation discards every formfill declaration. Such a guide gets no alignment prompt and can immediately fail its first requirement. Use the lint's first-executable traversal semantics and accept the first formfill declaration.

moxious (changes requested):
Request changes: two blocking manifest-ownership/data-loss issues and two medium correctness/security issues. Focused verification passed (9 suites, 302 tests), lint passed, and CI is green.

moxious (dismissed):
No showstoppers; risks worth flagging
 
1. The PR body is now stale and actively misleading — fix before merge. Its "ALL OF THE FOLLOWING ARE CAPTAIN-DECIDED. Do not re-flag them as defects" list asserts three rules the head contradicts: "a formfill's on-page: is ignored entirely" (now honoured in first position), "ignoring … conditional branches" (now walked, both to STOP and to harvest), and "written unconditionally, set or cleared" (now provenance-conditional). The Risk Assessment also still claims two unresolved issues that are both fixed. A do-not-re-flag list describing superseded rules is a trap for the next reviewer.
2. Conditional-branch harvest can stamp a branch-only page (minor, deliberate). scanBlocks walks whenTrue before whenFalse, so a guide whose branches declare different entry pages stamps the true-branch page; sawExecutable also leaks across mutually exclusive branches, so an executable in whenTrue consumes first-executable standing for whenFalse's formfill. Pinned by a test, prompt is advisory and dismissible — not worth blocking, but it is a guess the previous "ignore conditionals" rule didn't make.
3. stats: …optional().catch(undefined) also silences validation feedback. A partial or hand-edited stamp now reads as absent everywhere, including validate-package.ts — so the CLI will never tell an author their stamp is malformed. Intended (build-stats repairs it) and documented, but the author-facing consequence isn't stated anywhere.
4. Scope widened late. 5020b935 (stats stopped reaching repository.json once declared on the manifest schema) is a genuine latent break this PR's own schema change would have introduced, so it belongs here — but it pulls the CLI and RepositoryEntrySchema into a block-editor PR. Worth a glance at the build-repository tests, which is where the previous #1662 tests were re-homed.

moxious (approved):

PR #1685, opened 2026-08-23
feat(docs-panel): consolidate the learning-journey milestone toolbar

Spidy88:
## Summary

Replaces the learning-journey milestone toolbar's four stacked control rows (nav arrows, milestone counter, an Open/Pop out/Full screen/Reset guide button row, and a continuous progress bar) with one consolidated band: a left-aligned title and "Milestone X of Y" subtitle between compact nav arrows, a kebab menu absorbing all four actions, and a segmented per-milestone progress bar. This is the second of three planned guide-visual redesign PRs (the [Callout block](https://github.com/grafana/grafana-pathfinder-app/pull/1684) shipped first; the Knowledge-check/quiz restyle is a separate, later PR).

## Screenshots

sidebar-toolbar

Sidebar: left-aligned title + "Milestone X of Y" subtitle, compact nav arrows, segmented progress bar, kebab trigger.

sidebar-kebab-menu

Sidebar kebab open: Open, Reset guide, Pop out, Full screen.

fullscreen-kebab-menu

Full-screen surface: same toolbar, kebab now shows Dock instead of Pop out and hides Full screen (already fullscreen).

## What to look at

- `src/components/docs-panel/components/LearningJourneyMilestoneToolbar.tsx` — the kebab is built with `usePanelModeControls()` directly (the same pattern as the block editor's `HeaderKebab`) rather than a consumer-injected slot. That's a deliberate behavior change, not just a restyle: the fullscreen and floating surfaces now get a working Pop out/Dock item from this toolbar, which they didn't before (only the sidebar wired that up via `trailingActions`).
- The nav arrows are an icon-only `Button` (`variant="primary"`, `size="md"`), not `IconButton` — `IconButton`'s own doc comment says it "looks just like an icon but behaves like a button" and it has no background-color option, which review feedback flagged as the real readability problem (not the size). `Button` always renders a solid background/border per variant, no custom CSS needed. The kebab "More actions" button matches at `size="md"` too; a leftover padding override that was fighting that size prop got removed along with it (net reduction in custom CSS). The old hand-rolled solid-pill style (`navButton` in `src/styles/docs-panel.styles.ts`) is untouched — it's still used by the loading-skeleton block in `DocsPanelContentArea.tsx` (shown briefly while a journey tab's content is still fetching), a separate code path from this toolbar.
- `src/styles/docs-panel.styles.ts` — the new segmented bar uses distinct `progressSegments`/`progressSegment` keys rather than reusing `progressBar`/`progressFill`, specifically to avoid colliding with that same loading-skeleton block, which still renders the old continuous fill bar.
- `src/components/full-screen/full-screen.styles.ts` and `src/components/floating-panel/floating-panel.styles.ts` — each had a `secondaryActionButton` style that's now orphaned (it existed only to style the old Open/Reset button row via the removed `actionButtonClassName` prop) and was deleted. The sidebar's copy in `src/styles/docs-panel.styles.ts` stays — it's still used by the separate, untouched plain-docs header.
- `src/components/docs-panel/docs-panel.contract.test.tsx` — `popOutButton`/`fullScreenButton` test ids are now intentionally rendered by two files (the pre-existing `PanelModeActionButtons.tsx` for the plain-docs header, and this toolbar's kebab for journeys). The contract test's exhaustiveness check only tracks one canonical owner per id, so the mapping still points at `PanelModeActionButtons.tsx`; a comment explains the dual usage.

## Why

The current toolbar says the same thing in two visual languages at once — a "Milestone X of 8" counter and a continuous progress bar both signal position, and they're separated by an unrelated action row. Design supplied a mock consolidating this into title + segmented bar + kebab, mirroring how the block editor already collapsed a similar sprawl of toolbar buttons into a kebab.

We looked at also unifying this with the separate, simpler button row shown for plain (non-journey) docs/interactive tabs, using the same mock as inspiration there too. That would have required a new live completion-percentage hook — nothing today wires a reactive "N% complete" value into that header — and turned this into a much larger change touching code that isn't currently shared between the two surfaces. We deliberately scoped that out: this PR touches the journey toolbar only.

## How to verify

1. Open a learning-journey tab in the sidebar. Confirm the single band renders: left-aligned title + "Milestone X of Y" subtitle, small nav arrows, a segmented bar (one block per milestone), and a kebab (⋮).
2. Open the kebab and confirm it shows Open (when an external URL resolves), Reset guide (when there's interactive progress), a divider, then Pop out, then Full screen.
3. Click Pop out — the guide moves to a floating panel. In the floating panel, confirm the kebab is hidden (compact mode) but the floating panel's own header still has its own dock control.
4. Dock back to the sidebar, then use the kebab's Full screen — confirm the same consolidated toolbar renders in full screen, with the kebab now showing Dock (no Full screen item, since already fullscreen). Use Dock to return to the sidebar.
5. Click the forward/back arrows a few times and confirm the segmented bar updates correctly (done segments green, current segment highlighted, upcoming segments dim) and the subtitle count matches.

## Breaking changes

None. This change is additive and fully reversible — no storage, telemetry, or backend contract changes. Existing `docsPanel.popOutButton`/`fullScreenButton` test id values are unchanged, only which component renders them.

## Out of scope

- Unifying the journey toolbar with the plain-docs/interactive header — investigated, deliberately deferred (see Why).
- The Knowledge-check (quiz block) restyle — the third item in this redesign series, tracked as its own separate PR.


Jayclifford345:
@cursoragent review

Jayclifford345 (changes requested):
Howdy, @Spidy88, really like the new slick design however, we did have feedback in the past from writers that they felt the buttons were too small and should be big primary buttons. I think we could perhaps find a middle ground as i like the new design but perhaps we make the arrow keys full length and easier to identify. We just need to make sure we keep things easy to use for our accessibility users :)

Spidy88:
@Jayclifford345 addressed in 699b47d9 — swapped the prev/next arrows from `IconButton` to icon-only `Button` (`variant="primary"`, `size="md"`). `IconButton`'s own doc comment says it "looks just like an icon but behaves like a button" and it has no background-color option at all, which was the real problem — not the size. `Button` always renders a solid background/border per variant, no custom CSS needed.

Tried a few combos live before landing here: `primary`/`lg` looked oversized next to the title and kebab menu, and `secondary`/`md` was too subtle again (same complaint as before). `primary`/`md` matched the color that was already there and fit the row best.

Also bumped the kebab "More actions" button to `size="md"` to match, which turned up a leftover CSS padding override that was fighting the size prop — removed it, so this is a net reduction in custom CSS.

Jayclifford345 (approved):

PR #1686, opened 2026-08-23
fix(block-editor): stamp saves with the schema version the editor emits

### LINKED ISSUE #1678: Editor saves stamp schemaVersion 1.0 while the editor emits 1.1.0

devtejasx:
## Summary

Every guide the block editor saves to App Platform was stamped `spec.schemaVersion: "1.0"` while the editor actually emits `1.1.0`. `preservedSpec` fell back to a hardcoded `'1.0'`, and since `getGuide()` never returns a `schemaVersion`, that fallback was taken on **every** editor save — so each editor-authored guide misreported which schema produced its blocks, which is the one question `schemaVersion` exists to answer.

The comment on that line already stated the correct intent — "the editor writes the schema version it emits, not the stored one" — it just named the wrong constant. Pointing the fallback at `CURRENT_SCHEMA_VERSION` makes the two impossible to drift apart again.

Guides already stored as `"1.0"` keep that value until re-saved, which is correct: the field describes the schema that produced the blocks on the resource.

## How to verify

1. Open the block editor, create a new guide with a single markdown block, and save it as a draft.
2. Read the resource back — `GET /apis/pathfinderbackend.ext.grafana.app/v1alpha1/namespaces/<ns>/interactiveguides/<id>` — and confirm `spec.schemaVersion` is now `1.1.0` rather than `1.0`.
3. Load an existing guide that was stored with `"1.0"` and save it without editing; confirm it is re-stamped to the current version rather than pinned to the stale literal.

Two regression tests are added alongside: one asserting the saved version tracks `CURRENT_SCHEMA_VERSION` (and explicitly is *not* `'1.0'`, so a re-hardcode fails loudly), and one confirming a caller-supplied `schemaVersion` is still respected rather than overwritten.

## Breaking changes

None. This changes what future saves write; stored resources are untouched until they are next saved. Nothing on the App Platform read path branches on `spec.schemaVersion` today.

Fixes #1678

### DIFF
--- a/src/components/block-editor/hooks/useBackendGuides.test.ts
+++ b/src/components/block-editor/hooks/useBackendGuides.test.ts
@@ -286,6 +286,19 @@ describe('saveGuide — round-trip of fields the editor does not own', () => {
     expect(request.data.metadata).toEqual({ name: 'fresh', namespace: 'stacks-1' });
   });
 
+  it('stamps the schema version the editor emits, not a stale literal', async () => {
+    const result = await renderLoaded([]);
+
+    // getGuide() never returns a schemaVersion, so every editor save takes the fallback.
+    const guide = { id: 'fresh', title: 'Fresh guide', blocks: [] };
+    await act(async () => {
+      await result.current.saveGuide(guide as JsonGuide);
+    });
+
+    expect(lastRequest().data.spec.schemaVersion).toBe(CURRENT_SCHEMA_VERSION);
+    expect(lastRequest().data.spec.schemaVersion).not.toBe('1.0');
+  });
+
   it('does not borrow a manifest from an unrelated resource in the loaded list', async () => {
     const resource = pathCoverPage();
     const result = await renderLoaded([resource]);


Jayclifford345 (dismissed):

moxious:
Thanks for the contribution!

moxious (approved):

moxious (changes requested):
Oh unfortunately @devtejasx this can't be merged yet because the commits aren't signed, this is a policy in our github org. It looks like you need to revise the commits again to make sure ALL are signed, otherwise we're blocked by CI and unable to merge.

devtejasx:
@moxious That's sorted now — thanks for the clear pointer.

The commits did carry signatures when you looked; GitHub was reporting them `Unverified` because the signing key wasn't registered on my account, so it had nothing to check them against. That's now fixed, and since GitHub re-verifies retroactively, the existing commit flipped to Verified without a re-push:

- `9d282069` — **Verified**

The diff is byte-identical to what you approved — no force-push, no content change since your review. Could you re-run the review when you have a moment?

(Same applies to #1687, which is signed and verified too.)

moxious (changes requested):
Thanks for this work, and sorry for the confusion. The delay here is our fault, not yours.

## What happened

Your fix is correct. Another pull request landed the same change before this one merged.

PR #1682 changed `preservedSpec` to use `CURRENT_SCHEMA_VERSION`. It merged on 2026-08-25. That PR had a wider scope, so the schema-version fix arrived inside it. Nobody linked that PR to issue #1678, so the issue stayed open and your PR stayed in review.

We closed #1678 today. Main now has the exact line your PR adds:

```ts
schemaVersion: guide.schemaVersion || CURRENT_SCHEMA_VERSION,
```

This leaves your PR in an awkward state. Three of its four parts no longer add anything:

- The source change in `useBackendGuides.ts` is identical to main.
- The `CURRENT_SCHEMA_VERSION` import in the test file is already on main.
- The literal in the existing `toEqual` block is already `CURRENT_SCHEMA_VERSION` on main. PR #1682 also added a `manifest` field to that same block, so this part now conflicts.

To be clear about the cause: there is no disagreement between the two changes. They do the same thing. The conflict is textual only.

## What is still valuable

Your two new tests do not exist on main. We ran them against main. Both pass.

The first test is the one we want:

```js
expect(lastRequest().data.spec.schemaVersion).not.toBe('1.0');
```

Nothing on main guards against a return to a hardcoded version literal. That is the exact bug in #1678. Your test closes that gap, and we want to keep it.

## Suggested path

1. Rebase the branch on the current main.
2. Remove the change to `src/components/block-editor/hooks/useBackendGuides.ts`.
3. Remove the import line and the `toEqual` edit from the test file.
4. Keep the test named `stamps the schema version the editor emits, not a stale literal`.
5. Push the branch.

We will then dismiss the old review and approve. The result is a test-only pull request of about 12 lines. Main already imports `CURRENT_SCHEMA_VERSION` in that test file, so the test needs no import of its own.

One note on your second test, `keeps a schema version the caller supplied`. It pins a path that the editor cannot reach, because `getGuide()` never sets `schemaVersion`. Your own comment in the first test states this. You can drop the second test. If you prefer, keep it. It does no harm.

If you prefer not to spend more time on this, tell us. We will cherry-pick your test and credit you as co-author.

devtejasx:
@moxious Thanks for tracking down what actually happened — that saved me a lot of guessing.

I verified all three of your points against main before touching anything:

- `preservedSpec` on main is already `guide.schemaVersion || CURRENT_SCHEMA_VERSION` (`useBackendGuides.ts:91`)
- the test file already imports `CURRENT_SCHEMA_VERSION` (line 26)
- the `toEqual` block already uses it, and #1682 added the `manifest` field alongside it — that was the conflict

So I followed your steps exactly. The branch is rebased on current main and now adds one test and nothing else:

```
src/components/block-editor/hooks/useBackendGuides.test.ts | 13 +++++++++++++
1 file changed, 13 insertions(+)
```

I dropped the second test as you suggested — you were right that `getGuide()` never sets `schemaVersion`, so it pins a path the editor can't reach.

Two checks on the remaining test:
- full suite passes, 19/19
- reverting `preservedSpec` back to `|| '1.0'` makes it fail, so it does guard the regression rather than just restating current behaviour

`signed-commits` is green now too. Ready whenever you want to dismiss the old review.

moxious (approved):

PR #1687, opened 2026-08-23
fix(mcp): emit spec.manifest from the App Platform handoff descriptor

### LINKED ISSUE #1675: MCP finalize.ts drops spec.manifest from its App Platform write descriptor, publishing paths as flat guides
## Summary

`pathfinder_finalize_for_app_platform` builds its App Platform write descriptor as
`spec: { ...content, status }` and **never folds in `artifact.manifest`**, even though the artifact
it just validated has a real manifest in hand.

An agent following the `grafanaAppPlatform` branch of `clientGuidance` therefore writes a
**manifest-less** `InteractiveGuide` — so a package authored as a `path` publishes as a flat guide
with no `milestones`, no `description`, and no stats stamp.

## Evidence

`src/cli/mcp/tools/finalize.ts:148-158`:

```ts
resource: {
  apiVersion: APP_PLATFORM_API_VERSION,
  kind: APP_PLATFORM_KIND,
  metadata: { name: id },
  spec: {
    ...content,
    status,
  },
},
```

The manifest is returned separately at `:248-250` for the `localExport` branch only.

The snapshot test proves it — `src/cli/mcp/__tests__/finalize.test.ts:230-249` shows
`resource.spec` = `{blocks, id, schemaVersion, status, title, type}` while `artifact.manifest`
sits at `:130-135` with `{id, repository, schemaVersion, type}`.

## Second, smaller defect in the same payload

That snapshot also shows `spec.type: "guide"` leaking from `content` into the spec. `spec` has no
`type` field in the CRD, so the API server **prunes it with a 200 and a `Warning:` header** — verified
against the real generated CRD in a throwaway k3d cluster. Harmless today, but it is noise the
descriptor should not emit.

## Why this was missed

#1663's commit message states the omission was deliberate:

> The MCP `finalize.ts` handoff is deliberately untouched — it emits a descriptor for an external
> client to write and reads no existing resource, so it has nothing to preserve.

That reasoning covers **preservation** and is correct. It does not cover **emission**: there is a
manifest to emit, and it is being dropped.

## Suggested fix

Add `...(manifest ? { manifest: projectForCrd(manifest) } : {})` to `resource.spec`, where
`projectForCrd` mirrors `build_manifest` in `scripts/upsert-learning-path.sh:197-224` — the eight
CRD-typed keys verbatim, `depends` widened from bare ids to CNF singleton clauses, and every
remaining key swept into `additionalFields`. Drop the stray `spec.type`. Update the snapshot.

Related: #1600 covers the `metadata.annotations` half of the same MCP write path.

## Provenance

Found while designing block-editor manifest support (scout `pf-editor-manifest-s1`). Not blocked on
that work.

devtejasx:
## Summary

`pathfinder_finalize_for_app_platform` built its App Platform write descriptor as `spec: { ...content, status }` and never folded in `artifact.manifest`, even though it had just validated an artifact carrying one. An agent following the `grafanaAppPlatform` branch of `clientGuidance` therefore wrote a manifest-less `InteractiveGuide` — so a package authored as a `path` published as a **flat guide**: no `milestones`, no `description`, no stats stamp.

This also drops the stray `spec.type` from the descriptor. `type` describes the package rather than the guide content, and the CRD declares no `spec.type`, so sending it earned a prune-with-`Warning` on every write. The projected manifest carries it instead.

## What to look at

- **`src/cli/mcp/lib/crd-manifest.ts`** — the new `projectManifestForCrd`. It deliberately mirrors `build_manifest` in `scripts/upsert-learning-path.sh:197-224`, the only other writer of `spec.manifest`, so both entry points put the same bytes on the wire. The behaviours worth tracing are the ones that differ from a straight copy: `depends` widened from bare package IDs to CNF singleton clauses, `milestones` emitted only for the meta types (`path`/`journey`), author sub-keys beyond `name`/`team` swept into `additionalFields` rather than dropped between the typed field and the escape hatch, and `id` deliberately *not* emitted because `metadata.name` already carries it.
- **`src/cli/mcp/tools/finalize.ts`** — the descriptor now spreads `specContent` (content minus `type`) and conditionally folds in the projected manifest. The `content as unknown as Record<string, unknown>` cast looks loose but is load-bearing: `type` is not declared on `ContentJson`, yet clients send it alongside the typed fields, so it has to be stripped through a widened view.
- **`docs/design/APP-PLATFORM-PUBLISH-HANDOFF.md`** — the doc described the manifest drop as a deliberate MVP limitation and left "what is the smallest CRD change that round-trips manifest data?" as open question 1. The CRD answered that with `spec.manifest`, so the doc was contradicting shipped behaviour on both counts.

## Why

#1663's commit message states the omission here was deliberate: the MCP handoff "emits a descriptor for an external client to write and reads no existing resource, so it has nothing to preserve." That reasoning covers **preservation** and is correct. It does not cover **emission** — there was a manifest in hand and it was being dropped on the floor.

## How to verify

1. Author a `path` package with two or more milestones and drive it to `pathfinder_finalize_for_app_platform`.
2. Inspect `resource.spec` in the returned handoff: `manifest` should be present, with `type: "path"`, the `milestones` array intact, and `depends` rendered as CNF clauses (`["grafana-basics"]`, not `"grafana-basics"`).
3. Confirm `spec.type` is absent.
4. Write the descriptor to a stack with the CRD installed and confirm the API server returns no `Warning:` header for a pruned field.
5. Finalize a plain `guide` artifact and confirm the manifest still projects (as `type` + `repository`) rather than being skipped.

Snapshot coverage is updated in `src/cli/mcp/__tests__/finalize.test.ts`, plus a new case asserting a `path` artifact publishes with its milestones. The projection has its own unit tests in `src/cli/mcp/lib/__tests__/crd-manifest.test.ts`.

## Breaking changes

The handoff payload shape changes, and that payload is a contract Assistant and other MCP clients read verbatim: `resource.spec` gains `manifest` and loses `type`. Any client that was reading `resource.spec.type` needs to read `resource.spec.manifest.type` instead. The inline snapshot exists precisely to make this drift loud, and it is updated here.

## Reversibility

Reverting restores the previous descriptor shape cleanly — the change is confined to what the handoff emits and the MCP performs no writes itself. Resources already written with `spec.manifest` by this path stay valid, since the CRD types the field and the block editor's `preservedSpec` (#1663) carries it through subsequent saves rather than stripping it.

## Out of scope

- `metadata.annotations` on the same MCP write path — tracked separately in #1600.

Fixes #1675


Jayclifford345:
@cursoragent review

dgehring-grafana:
Hey @devtejasx! Thanks so much for your contribution, and good catch :) an automated review found nothing notable (which makes sense given the concise nature of the changes) but did detect some doc drift, full automated review below:

Routed review of [PR #1687](https://github.com/grafana/grafana-pathfinder-app/pull/1687) (`fix(mcp): emit spec.manifest from the App Platform handoff descriptor`).

**Classification:** `mixed` (MCP runtime handoff + public descriptor contract + design docs)  
**Coverage confidence:** high

| Field | Result |
| --- | --- |
| Activated concerns | Always-on: `security`, `correctness-and-reliability`, `testing-and-verification`, `reversibility-and-one-way-door`, `cross-cutting-architecture`. Conditional: `mcp-authoring-server` (paths + `pathfinder_finalize_for_app_platform`), `ai-subsystem` (`src/cli/mcp/tools/**` + agent-consumed payload), `guide-schema-and-contracts` (`manifest` / CRD projection). |
| Activation reason | Four files under `src/cli/mcp/**`, plus the handoff design doc. Semantic hits: finalize descriptor, `spec.manifest`, strip `spec.type`. |
| Risk signals | Public MCP outcome shape changes; clients write App Platform resources from that object. |
| Likely one-way doors | Cluster objects written after merge keep `spec.manifest` (already true for `upsert-learning-path.sh`; new for this MCP path). Git revert restores the descriptor only. |
| Reviewers run | Always-on + the three conditionals. No standalone `secure` skill (no auth/tokens/transport/lockfile/CI). |
| Contract evolution | `mcp-authoring-server` gate **not** triggered (`prior_semantic_pr_count=1`); **discretionary** second writer of `spec.manifest`. `guide-schema-and-contracts` and `ai-subsystem` gates **triggered**. Policy: `coherent_extension` / `follows_contract` / `coherent_extension` — no contract findings. GitHub `baseRefOid` is not an ancestor of the PR head; gates used merge-base `f73eb50b` → head `26897794`. |

---

### Findings (severity, then confidence)

**1. `ai-subsystem` — sibling design docs still teach “manifest stripped at publish”**  
- **Severity:** medium · **Confidence:** high · **Reversibility:** reversible  
- **Problem:** The tool result and `APP-PLATFORM-PUBLISH-HANDOFF.md` now project `spec.manifest` and drop `spec.type`. `clientGuidance` still says write `resource` as returned — that part is correct. Three long-lived authoring docs still describe the old MVP strip and link the removed `#fields-dropped-at-publish-mvp` heading:
  - `docs/design/PATHFINDER-AI-AUTHORING.md` (index table + principle 7)
  - `docs/design/HOSTED-AUTHORING-MCP.md`
  - `docs/design/AUTHORING-SESSION-ARTIFACTS.md`
- **Why it matters:** Agents that load those docs can treat a path publishing as a flat guide as intended, or rebuild `spec` from content+status instead of forwarding `resource`.
- **Suggested action:** Update those three docs to the projection (`spec.manifest`, no `spec.type`, `additionalFields` escape hatch) and retarget links to `#manifest-fields-at-publish`. Do not change `clientGuidance` unless a step still tells agents to reconstruct spec from content.

---

Activated concerns with **no remaining findings:** `mcp-authoring-server`, `guide-schema-and-contracts`, `security`, `correctness-and-reliability`, `testing-and-verification`, `reversibility-and-one-way-door`.

The implementation matches the stated intent: `projectManifestForCrd` tracks `build_manifest` in `scripts/upsert-learning-path.sh`; finalize snapshots and `crd-manifest.test.ts` pin milestones/CNF `depends` and the absence of `spec.type`. `#1662` (manifest top-level openness) and `#1663` (`preservedSpec` keeps `spec.manifest`) are compatible. `#1639` tool input catalog is untouched.

---

### Cross-cutting

No extra architecture finding. Two writers of `spec.manifest` (jq vs TypeScript) are an explicit parity pair with tests, not a second implicit contract. Residual risk is future drift if one writer changes without the other — already the stated design.

Does not establish a new typed facade that would require a `CONCERNS.md` contract-anchor row in this PR.

---

### Verification dropped (≥2/3 skeptics)

| ID | Why dropped |
| --- | --- |
| CR-1687-1 (non-meta `milestones` not swept into `additionalFields`) | Same META_TYPES / typed-key behavior as the shell writer; Rule 2 rejects that combo before finalize projects. |
| CR-1687-2 (only `content.type` stripped from spec) | Pre-existing content spread; this PR only removes the known CRD prune field. |
| REV-1687-1 (revert cannot unwind cluster `spec.manifest`) | Intended persist; already disclosed in the PR; CRD and upsert path already store this field. |

---

### Tech debt

Tech debt: no high-confidence patterns found in the changed files.

### Doc-drift updates recommended

`prevent-doc-drift` (review mode): no new `src/<dir>/`, npm scripts, flags, or plugin routes — **no `AGENTS.md` / rules edits required.** That skill cannot edit `docs/design/`; the three stale authoring docs above are the drift that matters.

### Instrumentation

Skipped: no new plugin user-facing action, fallback ladder, or panel. MCP descriptor-only.

---

**Bottom line:** Merge-ready on the code/contract path that fixes #1675. The blocking issue for agent behavior is keeping the three sibling design docs in the same “strip the manifest” story this PR just retired.

devtejasx:
Thanks @dgehring-grafana — the doc-drift finding was right, and it's fixed in `36659d20`.

I found the stale claim in five places rather than three, so I updated all of them:

- `PATHFINDER-AI-AUTHORING.md` — the component-index row, principle 7, and the open question (which this branch actually resolves, so it's now struck through the same way `APP-PLATFORM-PUBLISH-HANDOFF.md` records it)
- `HOSTED-AUTHORING-MCP.md` — the `pathfinder_set_manifest` paragraph
- `AUTHORING-SESSION-ARTIFACTS.md` — the artifact-shape paragraph and the finalization-output bullet

All of them now describe the projection rather than a strip: manifest goes to `spec.manifest`, CRD-typed keys verbatim, the remainder swept into `additionalFields`, and `spec.type` not emitted. Every `#fields-dropped-at-publish-mvp` link is retargeted to `#manifest-fields-at-publish`; `grep` for the old anchor across `docs/` now returns nothing.

I left `clientGuidance` alone — as your review noted, it already tells agents to write `resource` as returned, and no step reconstructs `spec` from content.

dgehring-grafana (approved):
LGTM, thanks again!

Issue #85, opened 2025-07-30
Choose and implement a license together with internal GOSSIP office


moxious:
https://raintank-corp.slack.com/archives/C08VB4WC64R/p1753878432384199

PR #1691, opened 2026-08-24
feat(settings): move tenant settings to a PathfinderSettings App Platform kind

Jayclifford345:
## What changes

Move org-owned Pathfinder settings from Grafana's provisioning-shared plugin settings document to the `PathfinderSettings` App Platform singleton. Keep the legacy `jsonData` store for OSS, self-managed and local development without the settings kind.

This extends the existing settings-preservation contract from #1514 and #1654: one shared reader, one writer, explicit org overrides only, and provisioned fields preserved. The contract anchor in `docs/design/CONCERN_DETAILS.md` is updated in this PR.

## Settings ownership

| Owner | Settings | Storage |
| --- | --- | --- |
| User | Developer opt-in | Browser-local key scoped by Grafana subpath, organization ID and user ID. It does not sync across devices. |
| Org | Explicit content-source, terms, interactive-feature, timeout, collaboration and developer-gate overrides | Sparse `PathfinderSettings` resource, named `default` in the stack namespace; viewer read and admin write. Legacy `jsonData` fallback where the kind is absent. |
| System | Stack identity and credentials, defaults, rollout decisions | SSS provisions `stackId`/`secureJsonData.accessToken`; runtime code owns defaults; OpenFeature owns rollout decisions. Neither evaluated flags nor untouched defaults are written as org overrides. |

The companion backend PR is grafana/grafana-pathfinder-backend#85. Its tenant fields are optional and do not carry schema defaults. Existing stored values remain explicit overrides; this change does not guess whether an existing value originally came from a default.

## Read and write behavior

- All config tabs and runtime consumers use the shared resolved config, including Coda, AI recovery, auto-completion, two-tab controls, recommendations and workshop configuration.
- Startup controller/executor, kiosk and config auto-open mounts wait for the first successful authoritative settings read, including recovery after a failure. Navigation listeners still register synchronously, and controller pairing data is preserved before URL cleanup.
- Plugin metadata does not establish readiness. Failed reads release the single-flight cache; the sidebar exposes a retry action. It constructs its scene only after readiness and preserves the scene and open content across subsequent config updates.
- Form drafts overlay only edited fields on the latest config, and submit only those edits. A slow initial read cannot cause untouched sibling fields to be overwritten.
- Saves read authoritatively first. Existing resources require `resourceVersion`. Transient update failures (404/405/500/501/502/503/504) retry twice after 250 ms and 750 ms using the same target, spec and resource version. Conflicts, permission/validation errors and exhausted retries surface without switching stores or recreating the object. First save creates with POST, so concurrent creation conflicts safely.
- GET/create 404, 405 and 501 permit the legacy fallback. Read errors, malformed snapshots and create-time 503 outages stop the save.
- All three settings forms retain their drafts, display a visible error, and allow another save after failure. A failed existing-resource update never writes legacy settings: the resource would still win on the next successful read, silently hiding that fallback edit.
- The OSS fallback preserves unknown provisioning fields and authoritative `enabled`/`pinned`, without submitting credentials. Existing OBO bridge handlers and identity exchange are unchanged.
- The legacy `devModeUserIds` entry migrates only for the signed-in user with no recorded choice. The old unscoped browser key is ignored because its owner cannot be determined. An explicit opt-out survives reloads.

## Validation

- Latest frontend run after merging main: 575 suites / 9,524 tests pass (18 skipped, 5 todo).
- Typecheck, lint, formatting and production build pass. The build retains an asset-size warning.
- Regression coverage verifies stable sidebar scene identity and open content, authoritative readiness, failed-read retry and startup recovery, sparse saves, CAS conflicts, update recovery and subsequent read-back, draft retention, and user/org isolation.
- Plugin Go lint/tests/build, review-contract tests and shell checks passed in the preceding validation round; this follow-up changes frontend readiness and its tests/docs.
- The frontend tenant key list matches all 23 documented backend schema properties (22 overrides plus schemaVersion). Backend #85 passed Go tests, build, lint, repeat generation with no drift, and CI.
- Frontend CI has restarted on the conflict-free follow-up head.

## Deployment and rollback

Deploy backend #85 before this frontend. Live Cloud validation is still required against a stack serving the kind: viewer read, admin create/update, conflict handling, persisted values after restart, and OSS fallback. This work did not deploy infrastructure or verify live Cloud behavior.

A frontend rollback reads the old `jsonData` values; resource-only edits would need to be copied back or re-entered. Reintroducing a defaulted backend schema can materialize defaults again. Treat these as rollback limitations, not automatic data migration.

The dependency audit found the pre-existing Tiptap advisory GHSA-j95f-988m-3j2f. Dependencies are unchanged here; track that upgrade separately.


tomglenn (changes requested):
Reviewed at `4675d541`. The store migration is the right move and the projection through `TENANT_SETTING_KEYS` is a genuinely good structural fix — `configToSpec` cannot leak provisioned or per-user fields into the tenant resource, and `settings-preservation.test.ts` pins both historical regressions directly, including the case where a stack really does have `enabled`/`pinned` off. That's the version of that test most people forget to write.

Requesting changes on three things that a green CI can't see, plus a set of smaller in-scope fixes.

## Blocking

**1. The write side was migrated, the read side wasn't — on Cloud, settings look like they don't save.**

`ConfigurationForm.tsx:85` refreshes from `fetchPluginJsonData(...)`; `InteractiveFeatures.tsx:44` and `TermsAndConditions.tsx:27` seed from `plugin.meta.jsonData`. Nothing under `src/components/AppConfig/` reads the resolved config. On a stack where `savePathfinderSettings` succeeds, `jsonData` is never written (`save-settings.ts:56-58`), so:

1. Admin sets `tutorialUrl`, saves, page reloads.
2. The form re-seeds from `jsonData`, which doesn't have it, and `buildStateFromJsonData` falls back to `DEFAULT_TUTORIAL_URL` — the field is blank again.
3. The next save from that tab sends those defaults as its `changes`, overwriting the stored value. Set `guidedStepTimeout` to 45000, later toggle `enableAutoDetection` on the same tab, and the timeout goes back to 30000.

`save-settings.ts:15-17` says a stale form snapshot "can only carry the fields that form actually owns" — that holds for cross-tab collisions and is exactly the problem for same-tab ones, because on Cloud the fields the form owns are the stale ones. `ConfigurationForm.tsx:110` and `:115` read `resolvedJsonData` too, so the page also shows pre-migration dev-mode state.

Seeding and refreshing all three tabs from `refreshPathfinderPluginConfig()` / `usePathfinderPluginConfig()` fixes it; that hook already merges resource over `jsonData` over defaults, and the `isDraftEdited` guard carries over unchanged.

**2. The dev-mode toggle no longer turns dev mode on.**

Before this change, `enableDevMode` wrote both halves — `jsonData.devMode = true` and the user id. Now it writes only `writeDevModeOptIn(true)` to `localStorage` (`dev-mode.ts:63-70`), while `isDevModeEnabled` still requires the tenant gate first and `DEFAULT_DEV_MODE` is `false`. Grepping for a writer of `devMode` outside tests turns up type declarations, `TENANT_SETTING_KEYS` membership, defaults and the rename — no writer. The only one left in the repo is the raw POST in `tests/block-editor.spec.ts`.

So on any stack that isn't already `devMode: true`, an admin opens the config page with `?dev=true`, flips the switch, sees "Saving to server and reloading…", and the switch comes back off with no error. The body says an admin "keeps the instance-level veto" and "re-enabling is one click" — right now the veto can't be lifted from the UI, and that removal isn't mentioned in the body.

Either add an admin-only tenant `devMode` field routed through `saveTenantSettings({ changes: { devMode: true } })`, or have `onChangeDevMode` write both halves for an admin. Worth a test asserting the toggle lands in a state where `isDevModeEnabled` is true.

**3. On a stack that has the plugin but not the kind, every save fails with no fallback.**

`isSettingsApiAvailable()` (`pathfinder-settings-api.ts:77-79`) checks the group aggregation toggle and the namespace. That toggle is shared with InteractiveGuide and is already on in Cloud, so it doesn't tell you whether `pathfindersettings` is served. In that window `savePathfinderSettings` does `PUT` → 404 → `POST` to the collection → 404 → and the error propagates out of the unguarded create at `:178-190`. `saveTenantSettings` doesn't catch it, so the `jsonData` write at `save-settings.ts:66-70` is never reached.

The read path degrades correctly (404 is in `UNAVAILABLE_STATUSES`); the write path doesn't, which contradicts "callers get `null` rather than a throw when the API is not there" and "the legacy `jsonData` path remains for OSS, self-managed, and local dev" — both true of reads only.

Treat an `UNAVAILABLE_STATUSES` code on the create as "not served here" and return `false` so the caller falls back. Keep 403 throwing, so a permission failure surfaces instead of silently writing to the wrong store.

Since the two halves of this pair deploy from separate repos on separate cadences, it's worth stating the required order in the body — grafana-pathfinder-backend#85 merged and applied first, and a minimum backend version — because that ordering is currently implicit.

## Also in scope

- **`devModeUserIds` opt-out can't stick.** `withPerUserSettings` recomputes `resolveDevModeOptIn() || hasLegacyDevModeOptIn(jsonData)` on every publish, and `readDevModeOptIn` can't tell "never set" from "explicitly false" (`dev-mode-opt-in.ts:38-44`). Nothing clears `devModeUserIds`, so a legacy `true` wins forever and `disableDevMode()` is undone on the next publish. The comment at `dev-mode.ts:181-182` calls this "used once, at bootstrap" — it re-applies every time. A tri-state read plus a one-shot write of `true` on first boot makes it a real migration.
- **The "Breaking changes" note says the opposite of the code.** It states `devModeUserIds` is no longer read and no migration is attempted; `hasLegacyDevModeOptIn` reads it and `withPerUserSettings` carries the opt-in forward. Worth rewriting to describe the carry-forward that's actually there.
- **The network client has no test.** `pathfinder-settings-api.test.ts` covers the mappers and `isSettingsApiAvailable`; `settings-preservation.test.ts` mocks the module. Nothing exercises the read ladder, the 404-on-PUT → POST path, the create-also-fails case, or the unavailable-write path. That's the riskiest code here and the body already notes a real Cloud stack is unverified, so unit coverage of the ladders is the only cheap net before merge — and it's what would have caught (3).
- **The PUT is an unconditioned full replace.** `configToSpec` can't carry `schemaVersion` (it iterates `TENANT_SETTING_KEYS`, and `:83-85` discards it), and #85's CRD marks all 23 fields required with defaults — so the write is accepted and defaulting resets `schemaVersion` to `1`. A resource at `schemaVersion: 2` is downgraded by the next save from this client, and any future field this client doesn't know goes the same way. Separately, `resourceVersion` is declared at `:48` and never sent, so two admins still last-write-wins; the read that would supply it is already being made. Sending `schemaVersion` from a client constant and threading `resourceVersion` through covers both, or a merge-patch for updates sidesteps the first entirely.
- **`docs/developer/DEV_MODE.md` still documents the removed model** — jsonData-scoped `devModeUserIds` as the storage and the gate (`:7`, `:10`, `:73`, `:86`). Four other docs were updated; this is the one whose whole subject changed. The new `components/AppConfig/README.md` table is a good template.
- **`peerjsPort` is unbounded client-side** (`ConfigurationForm.tsx:209-213`, `:500-507`) while the kind enforces `1..65535`. `0` or `70000` is accepted, persisted on OSS, and 422s on Cloud. The knock-on is worse than the direct case: `saveTenantSettings` builds `next` from the whole resolved config, so one out-of-range legacy value blocks the first migrating save from *any* tab until someone hand-edits `jsonData`. The two timeout fields are fine — their form ranges sit inside the kind's. Clamping the whole payload to the kind's bounds before writing would close the class.
- **Three comments say "Grafana per-user storage"** (`constants.ts:115`, `:202`, `dev-mode.ts:10-11`) where the implementation is `localStorage` — and `dev-mode-opt-in.ts:12-27` explains at length why `lib/user-storage` was deliberately not used. The body is accurate; these contradict it and the module they point at. (Same wording is in #85's CUE.)
- **No contract anchor or routing row.** This establishes a settings-ownership contract, a single-writer facade, and a persisted schema shared with another repo, and `docs/design/CONCERNS.md` has no anchor row and no routing row for `src/components/AppConfig/**` or `src/constants.ts`. Per the review skill that isn't deferrable to a follow-up, and `CONCERNS.md:118` asks for the routing row when the directory is created rather than when the contract is finally written.

## Scope

There are four separable concerns here: the `DocsPluginConfig` → `PathfinderPluginConfig` rename, the settings-store migration, the dev-mode redesign, and the e2e helper rewrite that follows from it. Findings 2, the `devModeUserIds` trap, the breaking-changes note and the comment drift all sit in the dev-mode half, which shares only the observation that `jsonData` is a poor store and needs its own tenant-gate UI, migration semantics and tests. Landing it inside the store migration is a plausible reason a regression as visible as "the toggle does nothing" got this far with CI green.

A split worth considering: rename + store migration + the CONCERNS anchor + the config tabs reading resolved config as one PR, and the dev-mode redesign with its tenant field, one-shot migration, rewritten `DEV_MODE.md` and corrected breaking-changes note as the other. Not strictly required if everything above lands together.

## Smaller things, take or leave

- `void schemaVersion;` at `:84` reads better as `schemaVersion: _schemaVersion` in the destructure — moot if `schemaVersion` becomes load-bearing.
- `_currentUserId` / `currentUserId` (`dev-mode.ts:45`, `:142`) are kept only so call sites compile, and `tab-gates.ts:29` still passes a user id into an ignored parameter.
- `400` in `UNAVAILABLE_STATUSES` (`:57`) is inherited from `fetchBackendGuides`; on a single-object GET a 400 is more likely a malformed request than an absent API, so it hides a real bug behind a silent fallback.
- `save-settings.ts:44-51` and `usePathfinderPluginConfig.ts:79-87` implement the same resolve-and-merge independently and have to agree; a shared helper would make (1)'s fix land in one place.
- `TELEMETRY.md:41` asks for a countable facade op on a degradation ladder. This adds one — resource → `jsonData` → defaults — with only a `logger.warn`, and nothing on the unavailable branches. A single low-cardinality op on the read path would answer "how many stacks are silently on the legacy store", which is the question that would have surfaced (3) in production.

## On the existing bot review

Its "config forms ignore App Platform" and "dev mode toggle never sets gate" both hold — same conclusions reached independently here. Its "settings PUT omits resourceVersion → later saves are rejected" is right about the omission but I can't support the rejection claim: an unconditioned update is legal Kubernetes semantics. The concurrency gap and the `schemaVersion` reset stand either way.

## Do the two known regressions stay foreclosed?

Wiped provisioned `stackId` (#1514): yes, structurally — `configToSpec` projects through `TENANT_SETTING_KEYS`, and the fallback path spreads `legacyJsonData` first with `getConfigWithDefaults` deliberately not emitting `stackId`. Silently unpinned plugin (`aa1c2efd`): yes — `fetchPluginSettings` reads `enabled`/`pinned` in the same operation, no tab echoes the stale `plugin.meta` snapshot, and on the App Platform path the plugin-settings record isn't touched at all. The class itself — two writers on one document — is fixed across tabs, but reintroduced within a tab by (1) and in a narrower form by the unconditioned replace. Both are fixable here.

Jayclifford345:
@tomglenn thanks — this was a genuinely useful review. All three blocking items and all eight "also in scope" items are addressed; details below, including one place where I disagree and one regression I introduced fixing your telemetry point.

Now at `bdd1f4cf`, CI green.

## Blocking

**1. Read side wasn't migrated.** Correct, and worse than the three tabs — `context-panel.tsx:1085` seeded `configWithDefaults` from `pluginContext.meta.jsonData` too, so on Cloud the recommender-terms banner would keep showing after an admin accepted.

All four now read through the resolved config. The three config tabs go through a new `useSeededDraft`, which owns the seed/re-seed/`isDraftEdited` pattern once instead of three times. `InteractiveFeatures`' `hasChanges` compared the draft against `jsonData` field-by-field; it now compares against `buildStateFromConfig(resolvedConfig)`, so the same fix covers it.

**2. Dev-mode toggle.** Confirmed — and it was worse than "the veto can't be lifted": `onChangeDevMode` also had an `if (!currentUserId) { alert(...); return; }` guard that made the toggle a no-op whenever boot data had no user id, for a decision that is no longer keyed by user id at all. Gone.

I took both of your options rather than either. The **Dev mode** switch lifts the tenant gate when turning on (never when turning off — revoking stack-wide on one person's opt-out is the org-wide-write mistake again), and a separate admin **Dev mode for this stack** switch is the veto. Without the second one the gate is write-once-true, which isn't a veto. `dev-mode-toggle.test.tsx` asserts the toggle lands in a state where `isDevModeEnabled` is true, plus the three negative cases.

**3. Create path throws with no fallback.** Fixed as you described: `UNAVAILABLE_STATUSES` on the create returns `false`, 403 still throws. Deploy order is now an explicit section in the body — grafana-pathfinder-backend#85 first, and what happens if it isn't.

## Also in scope

- **`devModeUserIds` opt-out can't stick** — right, and the fix needed the tri-state you called for. `readDevModeOptIn` returns `boolean | undefined`; `undefined` alone consults the allow-list, and adopting it calls `adoptLegacyDevModeOptIn()` so it's a one-shot. Two tests: opt-out beats a legacy entry, and the adopt happens exactly once.
- **Breaking-changes note** — rewritten to describe the carry-forward that's actually there.
- **No network test** — `pathfinder-settings-network.test.ts`, 24 cases: the read's degrade-vs-escalate split, `PUT`→404→`POST`, create-also-fails, the CAS token, the unknown-field carry-forward, the clamp. You were right that this is what would have caught (3).
- **Unconditioned full replace** — `resourceVersion` threaded through, and the write is now layered over the read spec rather than built from scratch. That handles `schemaVersion` as a side effect and, more importantly, any future field this client doesn't model. Sending a client constant alone would still have downgraded a `schemaVersion: 2`.
- **`DEV_MODE.md`** — rewritten around the two-gate model.
- **`peerjsPort` unbounded** — `TENANT_SETTING_BOUNDS` mirrors the kind, `clampToKindBounds` applies it to the whole payload before every write, and the field validates in the form. I pulled the ranges from #85's CUE rather than guessing: `requirementsCheckTimeout` 100–60000, `guidedStepTimeout` 1000–600000, `peerjsPort` 1–65535. Confirms your read that the two timeout form ranges sit inside the kind's.
- **"Grafana per-user storage" comments** — fixed in all three. The same wording is still in #85's CUE; that needs a separate PR in the backend repo.
- **CONCERNS.md** — `tenant-settings` routing row, review row and contract anchor added, plus two `Key dependency edges` rows in `systemPatterns.mdc`.

## Smaller things

All taken. `void schemaVersion` → destructure rename; `_currentUserId`/`currentUserId` removed from `isDevModeEnabled`/`isAssistantDevModeEnabled` and all four call sites including `tab-gates.ts`; `400` dropped from `UNAVAILABLE_STATUSES` with the reasoning in the comment; the duplicated resolve-and-merge is now `resolveTenantSettings`, shared by the config layer and the writer.

## Where I disagree

You're right that an unconditioned update is legal Kubernetes semantics and the bot's "later saves are rejected" claim doesn't hold — I read #85's generated client and there's nothing requiring a version on update. Worth adding that the bot's proposed fix would have been actively wrong on the create: a `POST` carrying a `resourceVersion` *is* rejected. The `resourceVersion` is now threaded on the update only, for your concurrency reason rather than the bot's.

## One regression, caught after the first push

Your TELEMETRY.md point was correct, but my first cut imported `lib/telemetry/facade` statically into `pathfinder-settings-api.ts` — which the config bootstrap reaches, and `module.tsx` runs on the critical path. That dragged the facade and Faro into `module.js`: **+14,076 bytes on first paint**, measured at `4675d541` vs `4f4911ad` on the same base. Exactly the hazard `lib/dev-mode-opt-in.ts` documents for `lib/user-storage`, hit from the other side.

`bdd1f4cf` moves it to a dynamic import at the emission site: 349 bytes instead of 14KB, and the counter still exists. Bundle bot went from +20.99 KB (+17.75%) to **+7.29 KB (+6.16%)**, most of which is the settings client itself.

## Scope

Not split — everything landed together, plus a merge of main to clear conflicts. Your reasoning for splitting is sound and I don't think the outcome contradicts it; I'd rather not unpick a coherent diff now that the dev-mode half has its tenant-gate UI, its one-shot migration, its rewritten `DEV_MODE.md` and its tests. Happy to split if you'd still prefer it before merge.

Still unverified: a real Cloud stack with the `pathfindersettings` kind deployed. That's called out in the body and I don't think unit coverage substitutes for it.

src/module.tsx
```diff
@@ -156,7 +156,7 @@ const plugin = new AppPlugin<{}>()
   });
 
 // Override init() to handle auto-open when plugin loads
-plugin.init = function (meta: AppPluginMeta<DocsPluginConfig>) {
+plugin.init = function (meta: AppPluginMeta<PathfinderPluginConfig>) {
```
moxious, line 159: The snapshot at line 180 still controls controller/executor/kiosk mounting, but the refresh below updates only interception and the package resolver. With resource-only settings saves, those gates retain metadata/default values. `AiFixOrchestrator.tsx:47`, `components/DocsPanelContentArea.tsx:122`, and `useCodaAvailability.hook.ts:147` also read metadata; a focused test confirms the AI handler ignores a resolved opt-out. Please migrate the remaining runtime readers and bootstrap gates to authoritative resolved configuration.

src/utils/pathfinder-settings-api.ts
```diff
@@ -0,0 +1,298 @@
...
+      recordSettingsStoreResolved(status === 404 ? 'not-created' : 'kind-not-served');
+      return null;
+    }
+    if (status === 403) {
+      recordSettingsStoreResolved('forbidden');
+    } else {
+      recordSettingsStoreResolved('read-error');
+      logger.warn('Failed to read Pathfinder settings resource', { error: err });
+    }
+    return null;
```
moxious, line 217: Returning null for a failed GET also tells the writer that no resource snapshot exists. `saveTenantSettings` then builds a full config from jsonData/defaults and sends a PUT without resourceVersion or the existing spec. The request sequence reproduces with GET 400/403/500. Please distinguish absent resources from read failures and abort the save when the authoritative read fails, before any replacement can be attempted.

src/components/AppConfig/use-seeded-draft.ts
```diff
@@ -0,0 +1,55 @@
...
+export function useSeededDraft<S extends object>(build: (config: ResolvedPathfinderConfig) => S): SeededDraft<S> {
+  const { config, isResolved } = usePathfinderPluginConfig();
+  const isEdited = useRef(false);
+  const [draft, setDraft] = useState<S>(() => build(config));
+
+  // Re-seed when the authoritative read lands, but never over an edit the admin
+  // has already made — losing someone's typing to a background read is worse
+  // than showing them a value one read behind.
+  useEffect(() => {
+    if (!isEdited.current) {
```
moxious, line 44: The first edit freezes every draft field, including untouched fields whose authoritative values have not arrived. Edit tutorialUrl before the resource response, let it return enableLiveSessions:true, then submit: ConfigurationForm still sends enableLiveSessions:false. The delayed-response test fails while the response-before-edit control passes. Please preserve edits per field while refreshing untouched fields, or wait for authoritative readiness before allowing edits/saves; the existing isResolved flag also marks metadata as resolved.

src/utils/pathfinder-settings-api.ts
```diff
@@ -0,0 +1,298 @@
...
+      getBackendSrv().fetch({
+        url: itemUrl(namespace),
+        method: 'PUT',
+        data: requestBody(spec, base?.resourceVersion),
+        showErrorAlert: false,
+      })
+    );
+    return true;
+  } catch (err) {
+    if (statusOf(err) !== 404) {
```
moxious, line 272: The original PUT404→POST404 fallback is fixed, but an initial PUT405/501/503 still throws while the read and create paths classify the same responses as unavailable. That prevents the legacy save and contradicts the new contract anchor. Please handle confirmed kind absence consistently on the initial update, keep transient read failures distinct, and cover the direct-PUT branches.

src/components/AppConfig/use-seeded-draft.ts
```diff
@@ -0,0 +1,55 @@
+/**
```
moxious, line 1: Please trim this historical preamble to the short invariant a reader needs. Also remove the obsolete plugin.meta seed wording in InteractiveFeatures.tsx:140-142 while touching the draft code; the shared resolved config now owns that value.

moxious (dismissed):
Blockers:

1. [blocking] **tenant-runtime-resolved-readers — Complete the runtime migration to resolved settings** (high · tenant-settings)
   `src/module.tsx:180` still captures `jsonData` for controller/executor/kiosk mounting (`:220`, `:247`, `:286`); the authoritative refresh at `:194` updates only interception and the package resolver. `AiFixOrchestrator.tsx:47`, `components/DocsPanelContentArea.tsx:122`, and `useCodaAvailability.hook.ts:147` also read raw metadata. Once `save-settings.ts:52` saves only the resource, these readers retain old/default values: enabled controller/kiosk/Coda settings are ignored, and an AI auto-heal opt-out is ignored because its default is true. A focused render test confirms generation still runs with resolved `enableAiAutoHeal: false` and empty `jsonData`.
   Required: Use the resolved config for every remaining tenant-setting consumer and make bootstrap mounts honor its authoritative result. Add disagreement tests where the resource and metadata have opposite gate values.

2. [blocking] **tenant-write-requires-successful-read — Abort the write when the authoritative settings read fails** (high · tenant-settings · irreversible without cleanup)
   `src/utils/pathfinder-settings-api.ts:205-217` turns GET failures, including 400/403/500, into the same `null` as an absent resource. `resolve-tenant-settings.ts:43` then supplies legacy values, `save-settings.ts:47-52` adds defaults, and `pathfinder-settings-api.ts:259-266` sends a full PUT without `resourceVersion` or the existing spec. The focused request test reproduces this sequence. If the backend accepts that unconditioned update, saved settings, newer fields, and `schemaVersion` can be overwritten; if it requires a version, the save fails later for the wrong reason. The successful-read CAS test does not cover this path.
   Required: Distinguish absent/unserved resources from failed reads. Abort saves when the authoritative read fails; update existing objects only with their snapshot/version, and create missing objects without a blind PUT. Add failed-read-before-write and concurrent-first-create tests.

3. [blocking] **tenant-settings-draft-refresh-overwrite — Preserve untouched draft fields when settings arrive after an edit** (high · tenant-settings · partially reversible)
   `src/components/AppConfig/use-seeded-draft.ts:43-51` stops all reseeding after the first edit, while `ConfigurationForm.tsx:230-240` submits every field owned by the tab. Edit `tutorialUrl` while the resource GET is pending, then let it return `enableLiveSessions: true`: saving still submits the initial `enableLiveSessions: false`. The delayed-response render test fails on that value; the response-before-edit control passes. Reading again inside `saveTenantSettings` cannot help because the stale values are supplied as explicit changes. `usePathfinderPluginConfig.ts:132` also marks metadata as resolved, so the existing `isResolved` flag is not authoritative-read readiness.
   Required: Merge incoming values into untouched draft fields while preserving edited fields, or prevent edits/saves until the authoritative read completes. Keep the delayed-response render-and-save regression test.

4. [blocking] **tenant-settings-unavailable-put-fallback — Make the initial PUT follow the documented availability contract** (high · tenant-settings)
   The original PUT-404 → POST-404 failure is fixed. The remaining branch is `src/utils/pathfinder-settings-api.ts:272`: an initial PUT returning 405/501/503 still throws, while the read and create paths classify those statuses as unavailable. `save-settings.ts:62` is therefore unreachable for that response. This contradicts the new tenant-settings anchor in `docs/design/CONCERN_DETAILS.md:71`; all three direct-PUT regression assertions fail while the existing create-fallback cases pass.
   Required: Handle confirmed unavailable-kind responses consistently on the initial update while retaining the 404 create path and surfacing permission/conflict/validation failures. Keep transient read failures distinct from confirmed absence, and align the documented status policy with that distinction. Add direct-PUT coverage.

Follow-ups:

1. [follow_up] **S-TIPTAP-GHSA-j95f-988m-3j2f — Track the existing Tiptap advisory separately** (high · security)
   `package-lock.json:5535` contains `@tiptap/core` 3.29.1, which the production audit matches to the high-severity [Markdown parsing advisory](https://github.com/advisories/GHSA-j95f-988m-3j2f), fixed in 3.30.5. The same dependency is present at base and this PR does not change the manifests. Reachability of the advisory’s specific token helpers through Pathfinder was not established.
   Follow-up: File separate dependency-remediation work through the repository’s security process and update the Tiptap family to compatible patched versions. This is not a requirement for the settings migration.

2. [follow_up] **tenant-settings-stale-draft-comment — Trim the touched draft comments to the current contract** (medium · testing-and-verification)
   `src/components/AppConfig/InteractiveFeatures.tsx:140-142` still describes the draft as seeded from `plugin.meta`, and the new `use-seeded-draft.ts:1-18` adds a lengthy migration-history preamble around a small hook. These comments obscure the current resolved-config ownership and miss the repository’s trim-on-touch rule.
   Follow-up: Remove the obsolete seed-source wording and historical preamble while updating these functions; keep a short comment only for the non-obvious draft invariant.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1691
Summary: move tenant settings to a PathfinderSettings App Platform kind
Verdict: Request Changes
Results: 4 blockers, 0 non-blocking findings, 2 follow-ups

Jayclifford345:
Fixes pushed in `f0495213` and `07f9d452`, including the merge conflicts with main.

- **Runtime readers and startup:** migrated the remaining settings consumers to the shared resolver. Coda, AI recovery, guide behavior, recommendations, workshop configuration and two-tab controls now read App Platform overrides. Controller/executor, kiosk and config auto-open mounting wait for the authoritative read; navigation listeners still register synchronously.
- **Safe writes:** failed reads now abort saves. Existing resources require the read `resourceVersion` and preserve unknown fields; absent resources use POST, so concurrent creation conflicts safely. Only confirmed API absence permits the legacy fallback. Permission failures, malformed responses and 503 outages cannot silently switch stores.
- **Draft preservation:** incoming settings refresh untouched fields while preserving edits, and forms submit only edited fields. An early edit no longer overwrites stale sibling values.
- **Settings ownership:** user opt-in is scoped by user, org and Grafana subpath. Org settings are sparse explicit overrides. SSS provisioning, runtime defaults and OpenFeature decisions remain system-owned; unrelated saves no longer materialize defaults. The companion schema change is grafana/grafana-pathfinder-backend#85.
- **OSS compatibility:** the legacy save preserves provisioning fields and authoritative `enabled`/`pinned` without submitting credentials. The existing OBO bridge remains unchanged. Updated the ownership documentation and removed stale draft comments.

Validation: 554 frontend suites passed across the full run and an isolated retry of one SIGSEGV-terminated worker (9,080 passing tests). The final startup follow-up additionally passed 38 focused tests covering behavior, module wiring and architecture. Local typecheck, lint, production build and plugin Go tests pass. Latest-head CI is still running; no failures reported at posting time.

Live Cloud validation remains required against a stack serving the new kind, with backend #85 deployed first. Resource-only edits are not automatically copied back to `jsonData` on rollback. The pre-existing Tiptap advisory remains separate follow-up work.

tomglenn (changes requested):
Thanks for the work here — most of what I flagged in my August review (stale settings readers, the admin dev-mode switch not actually enabling dev mode, no fallback on first-time saves) looks fixed in the latest commit. The core idea of moving settings into their own storage record, with a fallback to the old plugin-config location for installs without the new backend, is a solid design.

One blocker remains, and it's the same shape as the issue I raised before, just on the other side of the save path.

## The blocker: update saves have no fallback

Once an org already has a settings record (i.e., this isn't the first save), updating it is not resilient to the backend being briefly unavailable. If the backend returns an "unavailable"-style status while writing an update, the save throws and fails outright — it does **not** fall back to the old storage location the way a first-time save does. That means the most common real case — a second or later save — is the least resilient one, while the rare first-time save is protected.

Concretely:

- `src/utils/pathfinder-settings-api.ts`, in `savePathfinderSettings`: the fallback branch is gated by `!base`, so it only applies when creating a new record. When `base` is non-null (updating an existing record), the same statuses fall through to `throw err` instead of returning `false`.
- `src/components/AppConfig/save-settings.ts`: `saveTenantSettings` has no try/catch around that call, so its jsonData fallback logic (~lines 21-26) is unreachable when the update throws.
- `src/utils/pathfinder-settings-network.test.ts` (~lines 68-76): this is tested and locked in, not just an oversight. The parameterized test over `[403, 404, 405, 409, 422, 500, 501, 503]` asserts the update call `rejects.toMatchObject({ status })` for all of them — including 405/501/503, which are exactly the statuses that should trigger fallback instead.

I saw in the PR description that this is intentional — the stated design is that update failures should "surface without switching stores." I don't think that's the right call for a production settings page: it means a routine second save has no recovery path when the backend hiccups, while the initial save does. That inconsistency is going to show up as "my settings change just failed for no reason" for real users.

## What I need to see to approve

1. Make the update-failure path fall back to the old storage location for the same "unavailable" statuses, the same way the create path already does.
2. Fix (or replace) the test above so it asserts the fallback behavior instead of asserting the throw.

## Optional, non-blocking

`docs/developer/DEV_MODE.md` may still describe the old single dev-mode switch rather than the new per-org/per-user pair introduced here. Worth a quick check, but not something I need before approving.

src/module.tsx
Jayclifford345: Addressed in 07f9d452 and retained after the main merge in e38eaf64. Controller/executor, kiosk and config auto-open initialization now wait for refreshPathfinderPluginConfig through initializeConfiguredSurfaces. AI recovery, DocsPanelContentArea and Coda availability read the shared resolved configuration. Behavioral and module-wiring tests cover the authoritative startup gate; the full test run includes these tests.

Jayclifford345:
Pushed e38eaf64 with the current-main conflict resolution and save recovery fixes. GitHub now reports the PR as conflict-free.

For the latest update-failure review: existing-resource writes now retry transient HTTP failures (404/405/500/501/502/503/504) twice, after 250 ms and 750 ms, using the same target, spec and resourceVersion. A recovered save remains authoritative on the next read. A 409, authorization/validation failure, or exhausted retry stops the save. All three settings forms display an error, retain edits and permit another attempt instead of silently failing or throwing an unhandled rejection.

I used retries rather than the requested write-through to legacy jsonData. The existing resource wins in resolveTenantSettings when the backend recovers, so such a fallback would report success and then silently hide the user's edit. It also reintroduces the provisioning collision this migration removes. OSS/absent-kind GET/create fallback remains intact. Please review this recovery approach as the alternative to switching stores.

Validation: 570 suites / 9,436 tests pass in a complete two-worker coverage run, including recovery, conflict protection, read-back and form error tests. The initial high-concurrency run had a worker SIGSEGV; its suite passed alone and the full rerun passed. Typecheck, lint, formatting, terms sync, Go lint/tests/build, production build, review-contract and shell checks all pass. The frontend keys match all 23 documented backend schema fields. Frontend CI is still running; backend #85 is green. Live Cloud persistence validation remains outstanding.

Also replied to and resolved the earlier authoritative-bootstrap/runtime-reader thread, whose fixes are retained in this head.

src/components/App/ContextPanel.tsx
```diff
@@ -28,22 +26,20 @@ export default function MemoizedContextPanel() {
...
 }
 
-function SidebarContent({ pluginJsonData }: { pluginJsonData: Record<string, unknown> | undefined }) {
-  const panel = useMemo(() => {
-    const config = getConfigWithDefaults(pluginJsonData || {});
-    return new CombinedLearningJourneyPanel(config);
-  }, [pluginJsonData]);
+function SidebarContent() {
+  const { config } = usePathfinderPluginConfig();
+  const panel = useMemo(() => new CombinedLearningJourneyPanel(config), [config]);
```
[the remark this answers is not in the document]
Jayclifford345: Fixed in 1a37be0c. The sidebar waits for authoritative settings before constructing its scene, then owns one scene instance through useState. Later config updates are adopted by the existing syncPluginConfig path. A regression test edits open content, publishes a different config object, and verifies the scene is not reconstructed and the content survives.

src/hooks/usePathfinderPluginConfig.ts
```diff
@@ -58,13 +97,14 @@ export function publishPathfinderPluginConfig(jsonData: DocsPluginConfig): Resol
...
+ * Single-flight read of the saved settings across every store. Grafana's
+ * `meta.jsonData` snapshot can lag a save, so this is the authoritative value —
+ * but nothing waits on it: failure leaves whatever was already published in
+ * place.
  */
 export function refreshPathfinderPluginConfig(): Promise<ResolvedPathfinderConfig | undefined> {
-  refreshInFlight ??= fetchPluginJsonData(pluginJson.id)
-    .then((jsonData) => publishPathfinderPluginConfig(jsonData))
+  refreshInFlight ??= resolvePathfinderSettings()
+    .then((resolved) => publishPathfinderPluginConfig(resolved))
```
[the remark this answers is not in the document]
Jayclifford345: Fixed in 1a37be0c. Failed reads release the cached promise, so later mounts and the new sidebar Try again button can start a fresh single-flight read. Bootstrap now waits for the first successful authoritative publication across failures instead of finishing with undefined. Tests cover a failed read followed by concurrent retry callers and startup mounting exactly once after recovery.

src/hooks/usePathfinderPluginConfig.ts
```diff
@@ -85,7 +125,10 @@ export function usePathfinderPluginConfig(): PathfinderPluginConfigState {
   const pluginContext = usePluginContext();
   const pluginMeta = pluginContext?.meta;
   const contextState = useMemo<PathfinderPluginConfigState | undefined>(
-    () => (pluginMeta ? { config: getConfigWithDefaults(pluginMeta.jsonData || {}), isResolved: true } : undefined),
+    () =>
+      pluginMeta
+        ? { config: getConfigWithDefaults(withPerUserSettings(pluginMeta.jsonData || {})), isResolved: true }
+        : undefined,
```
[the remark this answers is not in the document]
Jayclifford345: Fixed in 1a37be0c. Plugin init no longer publishes meta.jsonData, and the hook no longer treats plugin-context metadata as authoritative. The readiness global is published only after the shared resolver succeeds (including the OSS fallback). The sidebar waits before constructing a scene; metadata and failed reads leave isResolved false. Updated the readiness contract and its regression tests.

Jayclifford345:
Confirmed and fixed all three new Cursor findings in 1a37be0c, followed by a main merge/conflict resolution in 5fba5dde:

- Sidebar hydration no longer recreates the scene. It waits for authoritative settings on first mount, then preserves the same scene and open content while applying later config updates in place.
- Failed reads no longer poison the cached promise. Later mounts and the sidebar's Try again action can retry; bootstrap stays pending until a successful publication so controller/executor/kiosk/auto-open initialization can recover without a reload.
- Plugin metadata no longer marks settings resolved. Only a successful shared-resolver read, including the OSS fallback, publishes readiness. Updated the readiness contract and its tests.

Replied to and resolved all three Cursor threads. The final merged branch passes 575 suites / 9,524 tests, typecheck, lint, formatting and production build. The build retains its existing asset-size warning. CI has restarted for the new head; live Cloud persistence validation remains outstanding. No new Cursor findings were present on backend #85.

tomglenn (approved):
# PR #1691 Re-Review: Settings Migration to App Platform

## Summary

This is a re-review of Jay's (@Jayclifford345) PR that migrates org-owned Pathfinder settings from Grafana's plugin `jsonData` to the App Platform `PathfinderSettings` resource. **All five prior blocking issues from @tomglenn and @moxious have been comprehensively addressed.** The implementation now follows the specified contract: authoritative read before write, CAS with `resourceVersion`, proper fallback strategy, and runtime resolved readers throughout.

---

## Prior Feedback Resolution

### From @tomglenn's Review (3 blockers) — All Resolved ✅

| Issue | Status | Evidence |
|-------|--------|----------|
| **Read side wasn't migrated** — Config tabs seeded from `jsonData`, showed stale values on Cloud | ✅ Resolved | Config tabs now seed from `useSeededDraft()` → `usePathfinderPluginConfig()` → `resolveTenantSettings()`. See `use-seeded-draft.ts:15`, `ConfigurationForm.tsx:68`. Test at `dev-mode-toggle.test.tsx:133-137` confirms form renders resolved values, not stale `jsonData`. |
| **Dev-mode toggle no longer turns dev mode on** — No writer for tenant `devMode` field | ✅ Resolved | Toggle now calls `saveTenantSettings({ changes: { devMode: true } })` to lift tenant gate. See `ConfigurationForm.tsx:102-106`, `save-settings.ts:11-17`. Test at `dev-mode-toggle.test.tsx:79-90` validates both gates are lifted. |
| **Create path throws with no fallback** — 404 on create propagated instead of falling back | ✅ Resolved | `UNAVAILABLE_STATUSES` (404/405/501) cause `savePathfinderSettings()` to return `false`, triggering `jsonData` fallback. See `pathfinder-settings-api.ts:42,185-190`, `save-settings.ts:18-26`. Test at `pathfinder-settings-network.test.ts:97-101` confirms fallback behavior. |

### From @moxious's Review (2 blockers) — All Resolved ✅

| Issue | Status | Evidence |
|-------|--------|----------|
| **tenant-runtime-resolved-readers** — Controller/executor/kiosk read raw `jsonData` | ✅ Resolved | All runtime surfaces receive resolved config from `initializeConfiguredSurfaces()`. See `configured-bootstrap.ts:23-42`, `module.tsx:202-269`. Global `window.__pathfinderPluginConfig` is published from resolved settings at `usePathfinderPluginConfig.ts:81-92`. |
| **tenant-write-requires-successful-read** — GET failures became `null`; blind PUT could overwrite | ✅ Resolved | **Absent vs failed distinction**: 404/405/501 return `null`; 403/500/etc throw (`pathfinder-settings-api.ts:131-142`). **Abort on failure**: `resolveTenantSettings()` uses `Promise.all()`, propagating throws (`resolve-tenant-settings.ts:11-20`). **CAS enforcement**: `resourceVersion` required for updates (`pathfinder-settings-api.ts:165-167`), included in PUT body (`pathfinder-settings-api.ts:146-151`). **Spec preservation**: Merges `base?.spec` with new values (`pathfinder-settings-api.ts:169`). Tests at `settings-preservation.test.ts:93-113` and `pathfinder-settings-network.test.ts:44-56` validate the contract. |

---

## New Observations (Non-Blocking)

### 1. Bootstrap Timing for Fallback Paths
Some runtime surfaces (`pages/docsPage.ts`, `FloatingPanelManager`, `FullScreenPanel`) defensively fall back to defaults if the global resolved config isn't available yet. This is proper defensive coding, not a bug—these paths handle the edge case where a deep link loads before `waitForPathfinderPluginConfig()` completes. No action needed.

### 2. Legacy `jsonData` Fallback Lacks CAS
The App Platform path correctly uses `resourceVersion` for CAS, but the legacy `jsonData` fallback (for OSS/self-managed stacks) does read-modify-write without optimistic locking. Concurrent admin writes in legacy mode could race. This is an acceptable trade-off since:
- Legacy mode is a fallback for environments without App Platform
- Such environments typically have fewer concurrent admins
- Adding CAS to the Grafana plugin settings API is out of scope

### 3. Permission Checks Delegated to Grafana
The frontend `saveTenantSettings()` has no explicit permission check—it relies on Grafana's backend to enforce admin permissions on plugin settings routes and the App Platform API. This follows the standard Grafana plugin pattern where the app registers a config page via `.addConfigPage()` and Grafana restricts access based on role.

---

## Test Coverage

Excellent regression coverage has been added:

| Test File | Lines | Coverage |
|-----------|-------|----------|
| `settings-preservation.test.ts` | ~250 | Provisioned fields survive saves, CAS semantics, concurrent write handling, per-user state isolation |
| `pathfinder-settings-network.test.ts` | ~200 | Read/write ladders, fallback paths, read failure abort behavior, retry logic, unknown field preservation |
| `dev-mode-toggle.test.tsx` | ~80 | Both gates (tenant + user) lifted correctly, form seeds from resolved config |
| `usePathfinderPluginConfig.test.ts` | ~300 | Resolution cascade (App Platform > jsonData > defaults), settings update broadcasts |

---

## Documentation

- **CONCERN_DETAILS.md**: New anchor for #1691 documenting invariants, storage strategy, field ownership
- **DEV_MODE.md**: Updated for two-gate model (tenant gate + user opt-in)
- **AppConfig README**: Comprehensive rewrite documenting three-way settings split (tenant/user/system)

---

**Verdict: Approve**

All five prior blocking issues have been comprehensively addressed with specific file:line evidence and validated by thorough test coverage. The implementation correctly follows the specified contract for authoritative reads, CAS writes, fallback behavior, and runtime resolution. Nice work addressing the feedback, Jay!

PR #1692, opened 2026-08-24
fix(recovery): let a complete manifest supply the starting location

### LINKED ISSUE #1681: A guide opened from inside a learning path never gets its alignment prompt: the complete manifest is shadowed by the stripped catalogue one

devtejasx:
Fixes #1681.

## What

A guide opened from inside a learning path never got its alignment prompt; the
same guide opened standalone did.

As the issue sets out, the member's starting location is not missing — it is
shadowed. `openMember` passes the path's `packageInfo`, whose manifest comes
from the catalogue proxy, and `customGuideManifest`
(`pkg/plugin/custom_guide_repository_client.go`) declares no starting location,
so the key is dropped at the wire boundary. The `backend-guide:` loader spreads
`spec.manifest` through unchanged, so `content.metadata.packageManifest` does
carry it — the reader just never looked there.

## Approach

The issue offers two options. This takes a third, narrower one.

`resolveStartingLocation` now accepts the manifests in priority order and
returns the first declared value. The call site passes `packageInfo` first, so
it remains authoritative wherever it declares a value; the only behaviour that
changes is where the lookup previously resolved to `null`.

That was chosen over the two listed options because:

- **Not the precedence change.** Launch-path precedence is untouched — nothing
  that reads `packageInfo` elsewhere is affected, and no case that previously
  resolved to a value resolves differently. The concern about a load-bearing
  reorder does not apply to a fallback.
- **Not `packageInfoForMember`.** `Milestone` (`src/types/content.types.ts:92`)
  has no manifest field, so that route means widening the milestone shape and
  threading a manifest through `resolvePackageMilestones` — a larger change to
  the resolver contract for the same outcome.

Path covers are out of scope, per the issue.

## Testing

- `src/recovery` — 89 tests pass, including five new cases covering ordered
  lookup, wire-pruned earlier manifests, `undefined` entries, empty-string
  values, and the bundled-index fallback still being reached.
- `src/components/docs-panel` — 544 pass, 6 skipped.
- `tsc --noEmit` clean; eslint and prettier clean on the changed files.

### DIFF
--- a/src/components/docs-panel/docs-panel.tsx
+++ b/src/components/docs-panel/docs-panel.tsx
@@ -924,9 +924,17 @@ class CombinedLearningJourneyPanel extends SceneObjectBase<CombinedPanelState> i
 
         // Implied 0th step: decide whether to prompt the user to navigate to
         // the guide's declared starting location before step 1 begins.
+        // Two manifests can describe this launch and they are not equally complete.
+        // `packageInfo` comes from the catalogue proxy, whose Go `customGuideManifest`
+        // declares no starting location, so the key is dropped at the wire boundary;
+        // the loader's manifest on the fetched content carries it intact. Passing both
+        // keeps `packageInfo` authoritative wherever it actually declares a value and
+        // only falls back where it previously resolved to null — so a guide opened
+        // from inside a learning path gets the same prompt as the same guide opened
+        // standalone.
         const startingLocation = resolveStartingLocation(
           url,
-          packageInfo?.packageManifest ?? fetchedContent.metadata.packageManifest,
+          [packageInfo?.packageManifest, fetchedContent.metadata.packageManifest],
           { isAdmin: currentUserIsAdmin() }
         );
         const currentPath = locationService.getLocation().pathname;
--- a/src/components/docs-panel/docs-panel.tsx
+++ b/src/components/docs-panel/docs-panel.tsx
@@ -67,7 +67,7 @@ import { journeyContentHtml, docsContentHtml } from '../../styles/content-html.s
 import { getInteractiveStyles } from '../../styles/interactive.styles';
 import { getPrismStyles } from '../../styles/prism.styles';
 import { config, getAppEvents, locationService } from '@grafana/runtime';
-import { evaluateAlignment, resolveStartingLocation, type LaunchSource } from '../../recovery';
+import { coerceLaunchSource, evaluateAlignment, resolveStartingLocation, type LaunchSource } from '../../recovery';
 import { currentUserIsAdmin } from '../../utils/current-user-role';
 import { SessionProvider, useSession, ActionReplaySystem, ActionCaptureSystem } from '../../integrations/workshop';
 import { panelModeManager } from '../../global-state/panel-mode';
@@ -218,7 +218,11 @@ class CombinedLearningJourneyPanel extends SceneObjectBase<CombinedPanelState> i
         return this.openLearningJourney(url, title, { source: 'recommender' });
       },
       (url: string, title: string, packageInfo?: PackageOpenInfo) => {
-        return this.openDocsPage(url, title, { source: 'recommender', packageInfo });
+        // Sections other than the recommender declare their own source.
+        return this.openDocsPage(url, title, {
+          source: coerceLaunchSource(packageInfo?.launchSource) ?? 'recommender',
+          packageInfo,
+        });
       },
       () => this.openEditorTab()
     );
--- a/src/components/docs-panel/CustomGuidesSection.test.tsx
+++ b/src/components/docs-panel/CustomGuidesSection.test.tsx
@@ -96,6 +96,7 @@ describe('CustomGuidesSection — path cards (launch bridge)', () => {
       // can recover the cover baseUrl.
       packageManifest: { ...pathGuide.manifest, id: 'fe-alerting-path' },
       resolvedMilestones: undefined,
+      launchSource: 'custom_guide',
     });
   });
 
@@ -141,7 +142,7 @@ describe('CustomGuidesSection — path cards (launch bridge)', () => {
     expect(openDocsPage).toHaveBeenCalledWith(
       'backend-guide:fe-alerting-01',
       'Alerting enablement',
-      expect.objectContaining({ packageId: 'fe-alerting-path' })
+      expect.objectContaining({ packageId: 'fe-alerting-path', launchSource: 'custom_guide' })
     );
 
     // Locked member is rendered but disabled and not clickable.
--- a/src/components/docs-panel/CustomGuidesSection.tsx
+++ b/src/components/docs-panel/CustomGuidesSection.tsx
@@ -36,6 +36,9 @@ function packageInfoForPath(path: PublishedGuide, resolvedMilestones?: Milestone
     // `packageManifest.id` — thread it through, or "back to cover" breaks.
     packageManifest: { ...(path.manifest as unknown as Record<string, unknown>), id: path.id },
     resolvedMilestones,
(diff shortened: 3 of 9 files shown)


src/components/docs-panel/docs-panel.tsx
```diff
@@ -920,7 +920,19 @@ class CombinedLearningJourneyPanel extends SceneObjectBase<CombinedPanelState> i
...
-        const startingLocation = resolveStartingLocation(url, packageInfo?.packageManifest);
+        // Two manifests can describe this launch and they are not equally complete.
+        // `packageInfo` comes from the catalogue proxy, whose Go `customGuideManifest`
+        // declares no starting location, so the key is dropped at the wire boundary;
+        // the loader's manifest on the fetched content carries it intact. Passing both
+        // keeps `packageInfo` authoritative wherever it actually declares a value and
+        // only falls back where it previously resolved to null — so a guide opened
+        // from inside a learning path gets the same prompt as the same guide opened
+        // standalone.
+        const startingLocation = resolveStartingLocation(
```
moxious, line 939: **The production launch still cannot reach the fallback manifest**

When a user clicks a member under Custom guides, `fetchPackageContent` replaces the backend loader's `metadata.packageManifest` with the parent path's slim catalogue manifest, so the two arguments you pass here are the same stripped manifest; App Platform also stores the value under `additionalFields.startingLocation`, while the resolver reads only the top level. That launch is then tagged `recommender`, which `evaluateAlignment` deliberately suppresses, so the reported member still never gets a prompt even after this fallback.

Preserve and normalize the member's backend manifest, pass `source: 'custom_guide'` for this surface, and add a regression test through `CustomGuidesSection` → `fetchPackageContent` → the docs panel.

moxious (changes requested):
Thanks for the careful wire-boundary diagnosis and for choosing a narrowly scoped precedence fallback; the priority and empty-value unit cases are useful. I found one production-path blocker before this can fix the reported flow.

devtejasx:
You were right on all three counts, and I confirmed each in the tree before changing anything. The first revision was necessary but not sufficient — thanks for catching that it could never reach the reported flow.

- `package-content.ts` attached the caller's manifest over the loader's, so both arguments I was passing were the same slim projection. It now merges: the caller still wins field by field, but no longer erases what the loader resolved.
- `additionalFields` does not appear anywhere in `src/`, so the nested App Platform shape was simply never read. `resolveStartingLocation` now accepts both shapes.
- The launch was tagged `recommender` via the `ContextPanel` callback in `docs-panel.tsx`, which is aligned-by-construction. `PackageOpenInfo` now carries the launching surface and the custom guides section declares `custom_guide` — already in `NEEDS_ALIGNMENT_CHECK_SOURCES`, so no classifier change was needed.

One deviation worth flagging: I typed `PackageOpenInfo.launchSource` as `string` rather than `LaunchSource`, and narrow it with `coerceLaunchSource` at the launch boundary. Importing the union into `types/` trips the Tier 0 rule in `architecture.test.ts`. Moving `LaunchSource` down to Tier 0 looked like the architecturally correct answer but a wider change than this fix should carry — happy to do it as a follow-up if you'd prefer the stronger type here.

Tests: a new `package-content.member-manifest.test.ts` drives `backend-guide:` content through `fetchPackageContent` with a slim path manifest and asserts the resolver finds the starting location in both shapes, and still yields null when nothing declares one. `CustomGuidesSection` asserts the source it emits, and the evaluator asserts a `custom_guide` launch prompts. 94 suites / 1101 tests pass across `recovery`, `docs-retrieval`, `docs-panel` and `types`; typecheck, eslint and prettier clean.

Note the 4 failures in `import-graph.test.ts` are pre-existing on this machine (path-separator handling on Windows) — they reproduce unchanged on a pristine checkout.

devtejasx:
@moxious Rebased onto current main — the branch had gone into conflict since the last push.

The conflict was confined to `src/recovery/starting-location.test.ts`, and it was purely additive on both sides: main gained the real-schema regression test for a manifest that never authors `startingLocation`, and this branch adds the `with more than one manifest` describe block. Both are kept, nothing was dropped.

State after the rebase:

- 0 commits behind main, both commits signed and verified
- `npx jest src/recovery src/docs-retrieval/content-fetcher/package-content.member-manifest.test.ts src/components/docs-panel/CustomGuidesSection.test.tsx` — 124/124 pass
- `tsc --noEmit` clean, prettier clean on all 9 changed files

The three blockers you raised were addressed in the previous push and are unchanged by the rebase: the member's backend manifest is preserved and normalized rather than overwritten by `fetchPackageContent`, this surface passes `source: 'custom_guide'` so `evaluateAlignment` no longer suppresses it, and there's a regression test through `CustomGuidesSection` → `fetchPackageContent` → the docs panel.

Ready for another look when you have time.

moxious (approved):
Thank you for the contribution.  This will be merged within a few hours once we sort out some deploys.

PR #1694, opened 2026-08-24
chore: prep v2.17.0 release

Jayclifford345:
## Summary

Version bump, CHANGELOG entry, and the `release.yml` Node pin for v2.17.0, prepared on a frozen release branch.

**This PR is the review gate for the release.** `v2.17.0` is deliberately **not tagged yet** — the tag will be cut from this branch's head once this PR is approved, so the release notes are reviewed before the tag exists rather than after.

## Why a release branch

Previously the prep commit went to `main` and the tag followed `main`'s tip. Because `main` is protected, the prep PR is squash-merged onto whatever `main`'s tip is at merge time — so any PR merged during review was folded into the release commit: present in the tag, absent from the CHANGELOG, never covered by the `npm run check` run that validated the release.

`release/v2.17.0` branches from `2d3cc34a` (`main`'s tip at prep time). The tag will point at this branch, so nothing merged to `main` after `2d3cc34a` can reach `v2.17.0`.

Because the repo allows only squash and rebase merges, this PR returns to `main` as a new SHA, so the tagged commit will not be an ancestor of `main`. That is intentional — `actions/checkout` on the tag ref and the GitHub release both work, and `git tag --sort=-v:refname` is ancestry-independent.

## Why the CI fix is in this PR and not its own

`grafana/plugin-actions/build-plugin` defaults `node-version` to `20`; this repo requires Node >=24 (`.nvmrc` `24.19`, `engines.node`). On Node 20, `webpack-cli` cannot load `webpack.config.ts` — ESM rejects the `.ts` extension, and the ts-node CJS path dies in `undici` on `webidl.util.markAsUncloneable`, a Node 22+ API. Every tag-triggered release run has failed this way since December 2025, v2.15.0 included ([example run](https://github.com/grafana/grafana-pathfinder-app/actions/runs/32740470769)).

A tag-triggered run uses the workflow file **at the tag ref**. So the fix has to be inside the commit that gets tagged, or v2.17.0's own release run fails exactly like its predecessors — landing it separately on `main` would not help this release. It was originally opened as #1695, now closed in favour of this PR.

## Contents

| Commit | |
| --- | --- |
| `8715e074` | version bump + CHANGELOG entry |
| `df8a69a7` | `release.yml` pinned to Node 24 |
| `2fd749e6` | CHANGELOG bullet for the above |

- `package.json` + both `package-lock.json` version fields → `2.17.0`
- CHANGELOG: the pre-existing `## Unreleased` section folded into `## 2.17.0`, covering the 23 commits in `v2.16.0..2d3cc34a` — 5 Added, 1 Changed (Grafana 13.2 + React 19), 2 Security, 7 Fixed, 3 Chore. Renovate bumps dropped unless carrying a security advisory.

## What to look at

**The CHANGELOG prose is the substance of this review** — it is what ships as the release notes. Worth checking the two Security entries in particular (#1604 JWKS verification, #1680 CVE-2026-53668) and the Changed entry for the React 19 / Grafana 13.2 migration (#1674).

In `release.yml`, the commented-out signing lines were previously under a commented-out `#with:`. Now that `with:` is real, they are indented one level deeper so uncommenting `#policy_token:` puts it inside the `with:` block rather than alongside it. Their text is otherwise unchanged.

## How to verify

- `git log --format='%s' v2.16.0..2d3cc34a` is the exact set of PRs the CHANGELOG describes
- `npm run check` and `npm run build` passed on `8715e074`; `dist/plugin.json` stamped `2.17.0`
- The workflow input parses: `python3 -c "import yaml;print(yaml.safe_load(open('.github/workflows/release.yml'))['jobs']['release']['steps'][1]['with'])"` → `{'node-version': '24'}`
- A dev deploy from this branch succeeded: [run 32740919175](https://github.com/grafana/grafana-pathfinder-app/actions/runs/32740919175)

The Node pin can only be confirmed end to end by this release's own tag push, for the reason given above.

## After approval

1. Tag `v2.17.0` at this branch's head, push the tag
2. Merge this PR
3. Dispatch `publish.yml` against `release/v2.17.0`: ops, then prod — prod is what mints the GitHub release

## Breaking changes

None.

## Reversibility

Fully reversible — no tag exists yet, and the workflow change reverts as a single commit.

## Follow-up

`/release-prep` and `/changelog` still document the old main-tip flow, including a `git push origin main` step that cannot work against branch protection. Those skill updates land in a separate PR.


tomglenn:
This release is blocked pending investigation into an issue introduced in https://github.com/grafana/grafana-pathfinder-app/pull/1604.

moxious (approved):

Issue #1402, opened 2026-07-22
Runtime smoke: verify /custom-guide-repository credential path on a dev stack


moxious:
## What

Run the runtime smoke procedure from #1400's test plan (per `docs/design/BACKEND_PROXY_PATTERN.md` §3/§10), on a dev stack with `idForwarding` and the aggregation toggle (`aggregation.pathfinderbackend-ext-grafana-com.enabled`) on:

1. Create an `InteractiveGuide` upstream (`pathfinderbackend.ext.grafana.com/v1alpha1`, `interactiveguides`).
2. Hit `GET /custom-guide-repository`.
3. Confirm a slim shaped catalogue comes back (`{ capability, guides, asOf }`, no `blocks` in any entry).
4. Confirm the first-request credential diagnostics log (`app platform proxy: first upstream LIST`) shows the upstream LIST authenticating (status 200, `idTokenPresent=true`).
5. If the smoke proves `X-Grafana-Id` alone suffices, decide whether to narrow `forwardIdentityHeaders` and record the result in the pattern doc.

## Why this is an issue and not just a test-plan checkbox

Mocked tests cannot prove the live credential path. The outbound header shape (`Authorization: Bearer <id-token>` + `X-Grafana-Id`) is runtime-verified only from the sibling guides-import proxy, not from these routes. This smoke is the **gate before frontend work binds to the envelope** — and because #1398 (completion records) and #1400 (custom guide catalogue) terminate at the same shared LIST client and identity helpers, one smoke run validates both proxies.

Merged-PR checkboxes get lost; this issue keeps the gate visible while frontend work starts.

Refs #1400, #1398, #1401.

moxious:
## 🧭 Implementation readiness

**Stakes:** Nothing changes for the user. This is internal work. A live run is the only proof that the one outbound credential authenticates against a real stack.

**Sanity check:** The route, the identity gate, the outbound credential and the diagnostics log all exist at this base, so the check has a real target. Three of the five listed steps still name coordinates that do not match the code.

**Severity:** 🟠 Medium — Running the check changes nothing a user sees, so this is verification and not repair. It is not low, because the thing under test is the only credential path for both App Platform proxies and a failure there is served as a soft success.

**Code basis:**
- `pkg/plugin/resources.go:16` — Where the route under test is registered.
[...]

PR #1700, opened 2026-08-25
fix(auth): verify caller ID tokens against auth-api's JWKS

tomglenn:
## What this fixes

Every Grafana Cloud stack currently reports `identity-unavailable` on the App Platform proxy routes, so custom guides and completion records are dark.

#1604 verifies the caller's `X-Grafana-Id` token against `{appURL}/api/signing-keys/keys`. That is the right endpoint for self-hosted Grafana, which signs its own ID tokens. A Cloud stack does not sign them — auth-api does — so the stack serves `{"keys":null}`, no `kid` ever resolves, and every caller is refused.

Production today runs v2.16.0, whose gate never consults the signing keys, so it is unaffected. **v2.17.0 as currently cut is not.** That release branch predates this fix and still carries the stack-only gate, so publishing it as it stands would take every Cloud stack dark on the guide catalogue and completion records. The release is being held and cut again from a `main` that carries this fix, rather than cherry-picked onto the existing branch — so the CHANGELOG entry that describes the stack-only mechanism and its retryable 503 is rewritten there rather than amended.

## What changed

The verifier resolves a token's `kid` against two JWKS endpoints. **Both are full ES256 signature verification.** Neither source is a relaxed check and there is no unverified path.

**The stack's own endpoint is asked first, then auth-api.** The stack is the authority for its own callers. Asking an unauthenticated in-cluster address ahead of it would let whatever answers that name decide who the caller is — and on a self-hosted cluster that happens to run a service called `api-lb` in a namespace called `auth`, that name resolves. The Cloud fix still works because the chain falls through: a Cloud stack's `{"keys":null}` is an *answer* that does not resolve the `kid`, so auth-api is still reached and verifies. `TestNewIDTokenVerifier_CloudFallsThroughToAuthAPIAfterTheStack` pins the order itself, not just the verdict, by recording which endpoint is asked first.

auth-api's key set is **cell-wide**, so a valid signature alone does not identify which stack a token was issued for. `Verify` therefore also binds the token's `namespace` claim to the server-derived plugin-context namespace — never a request header, never the token's own claim. Two properties now hold together:

- the token was issued **for this stack** — the namespace binding, and
- the caller presenting it **is** that user according to the issuing authority — the ES256 signature.

A token carrying no `namespace` claim is rejected rather than passed. The comparison is exact string equality, so a `*` (all-namespaces) claim is refused; authlib's `types.NamespaceMatches` would accept it, and these are browser-facing per-user routes.

Identity failure is also no longer a 503. A key lookup that comes back empty-handed serves the in-band capability envelope on reads and the structural 404 that disarms on writes, because none of these conditions recovers on retry — a retryable status would leave every queued write retrying to the 30-day horizon without ever disarming. The two causes stay distinguishable: a source that **answered** without the key keeps `identity-unavailable`; **no source reachable at all** gets a new `signing-keys-unreachable` token.

## What to look at

- `pkg/plugin/auth/id_token.go` — the source order and its rationale, and how exhausting the chain is classified. Per-source causes are folded into the error message rather than wrapped, deliberately: wrapping would let an unreachable source flip a rejection into an outage.
- The ordering trade-off, stated in `BACKEND_PROXY_PATTERN.md` §3 rather than left implied: on a colliding `kid`, a Cloud stack that published a stale or wrong key would now outrank auth-api, which is the actual issuer there. Cloud stacks publish `{"keys":null}` today, so it does not arise in practice.
- `pkg/plugin/app_platform_identity.go` — the namespace binding is enforced inside `Verify` rather than returned for each caller to compare, so a new call site cannot silently skip it.
- `.golangci.yaml` — the `exhaustive` linter, enabled in opt-in mode with `//exhaustive:enforce` on the two identity switches. This PR collapsed three review-visible switch sites into two, so without it a fifth `identityStatus` would compile clean and take the write path's retried 401. The `default:` arms stay, because Go does not require a switch to cover an enum and removing them would trade a silent 401 for a silent fall-through.
- The redirect guard on the JWKS fetch now compares scheme and host:port rather than hostname alone, and re-imposes the ten-hop cap that replacing net/http's `CheckRedirect` had removed.

## Trade-offs to weigh

- On a Cloud stack the local endpoint always answers `{"keys":null}`, and that counts as an answer — so `signing-keys-unreachable` does not fire there, and a dead or misaddressed auth-api reports `identity-unavailable`, the same envelope a forged token produces. The logs do separate them: the key-lookup lines carry a per-source `<name>: <cause>` field naming which endpoint failed and why.
- The catalogue client caches a soft `available: false` for its 30s TTL, whereas a thrown 503 was never cached. A signing-keys outage therefore stays visible for up to 30s after recovery, where previously the next panel open re-probed. The envelope is the right shape because the condition is standing, not because it is cheaper.
- Each source gets a slice of the fetch budget rather than sharing one deadline, so a stalling first source cannot starve the one behind it. The cost is that a first source which would have answered inside the whole budget is cut off at its share; with nothing behind it, that reports as unreachable.

## How to verify

`go test ./pkg/...` covers the source-chain matrix: the Cloud fall-through, self-hosted with auth-api absent or answering empty, both key sets empty, no source reachable, wrong-key rejection, sibling-stack rejection, missing-namespace rejection, and the `*` refusal.

Reverting the seam to the pre-fix wiring reproduces the reported symptom exactly — `200`, `available: false`, `identity-unavailable` — and fails the Cloud regression test.

Two things worth knowing when reading the check results:

- **The browser test matrix is not coverage for this code.** `tests/` names none of these routes and the compose file enables no aggregation toggle, so no Playwright row on any Grafana version has ever exercised this path. A green matrix is not evidence for this change. #1402 is the gate for exactly this and is still open; it should run on a Cloud stack before v2.17.0 ships.
- **The equality the binding depends on holds by construction, not by luck.** On Grafana 13.0 and later, `PluginContext.Namespace` comes from `user.GetNamespace()` and the token's `namespace` claim comes from the same `request.GetNamespaceMapper(cfg)`, on the same org id, in the same process. Verified in `base_plugincontext.go` at v12.3.11, v12.4.9, v13.0.7 and v13.2.0. Grafana 12.3 and 12.4 do not set the field at all, so the binding returns `identity-unverifiable` there — those routes were already unavailable on those versions, so nothing regresses.

## Breaking changes

None on the wire for a stack that was working. `signing-keys-unreachable` is a new capability reason token; the front end gates on `available` and ignores the string.

## Reversibility

Partially reversible. A plain revert restores single-source verification and re-breaks Cloud stacks, which is why the tests pin the Cloud path. No persisted data shape changes and no irreversible external effects.


Jayclifford345:
@cursoragent review

moxious (dismissed):
## Request changes

The core fix is correct and the two-source chain is well built. I ran 10 concern
reviewers and 15 skeptic passes. CI is green on all 30 checks. Two findings did
not survive verification, and I removed them (see the end).

Four items need a change in this PR. One item needs a decision outside it.

---

### Must fix

**1. `.cursor/rules/systemPatterns.mdc:204` inverts the security property.**

The row says the verifier binds the `namespace` claim to the "caller-supplied"
expected namespace. That value is server-derived. `Verify` says "never a request
header and never the token's own claim", and §3 says the same twice.

The word "caller" means the HTTP requester everywhere else in this table. An
agent that reads this row can wire `expectedNamespace` from a request header.
The cell-wide key set then accepts a sibling stack's token. No test catches this,
because no test pins where the argument comes from — only its value.

Fix: write "server-derived expected namespace — the namespace from the plugin
context, never a request header, never the claim in the token itself."

**2. `pkg/plugin/app_platform_identity.go:59-61` states the opposite of the truth.**

The comment says a 503 "would darken every proxy route at once for the whole
client cache TTL". Your own comment at `pkg/plugin/custom_guide_repository.go:119-121`
says the client "never caches a thrown 503". I verified this. `cache.set` sits
inside the `.then` block, so a thrown 503 never reaches it.

The new envelope stays for the full 30-second TTL. The 503 did not. So the
comment argues for the opposite of the change it justifies.

Fix: keep the "far more likely wrong than briefly down" sentence. Remove the
cache claim.

**3. The `identityStatus` enum has no exhaustive check.**

This PR removed the three named switches on the read paths. `capabilityReason()`
and the write switch both end in `default:`. The write default is the 401 that
this file defines as transient.

`.golangci.yaml` holds only the golangci-lint v2 defaults, so the `exhaustive`
linter is off. A fifth status will compile clean and take the retried 401.
`pkg/plugin/completion_records_write.go:163-174` says that must never happen.
Before this PR a new status produced three review-visible sites. Now it produces
none.

Fix: list every status explicitly in both switches. Remove the `default:` arms.

**4. Four documentation statements are now wrong.**

- `docs/design/BACKEND_PROXY_PATTERN.md:332-333` lists "no app URL, no namespace"
  as live custom-guide reasons. The identity gate shadows both. Your new block at
  `src/lib/custom-guide-repository-client.ts:57-60` says so.
- `pkg/plugin/app_platform_identity.go:29-33` claims the envelope alone is
  diagnosable. §3 disclaims this for Cloud, in this same PR.
- `docs/design/BACKEND_PROXY_PATTERN.md:237` says "Binding the token to its
  presenter is tracked separately". No such issue exists. File it, or write
  "accepted residual".
- `docs/design/BACKEND_PROXY_PATTERN.md:245` says verification "avoids
  per-request network calls". That is not true for source 0. See item 6.

---

### Decide before release

**5. v2.17.0 carries #1604 without this PR.**

```
git merge-base --is-ancestor db075e5b origin/release/v2.17.0  → YES
git merge-base --is-ancestor 79ec7b9c origin/release/v2.17.0  → NO
```

PR #1694 is open and frozen at `2d3cc34a`, so this PR cannot reach that tag. The
CHANGELOG on that branch ships the stack-only mechanism as a Security highlight,
and it names the "retryable 503 with Retry-After" that you replace.

If #1694 is tagged and published as it stands, every Cloud stack reports
`available: false` with `identity-unavailable` on `/custom-guide-repository` and
`/completion-records/*`. That is the outage this PR corrects.

Choose one:

1. Cherry-pick this PR onto `release/v2.17.0`. Then rewrite the #1604 CHANGELOG
   entry and re-run `npm run check` there.
2. Hold #1694. Cut v2.17.0 again from a main that holds this PR.

Add a qualifier to the PR body either way. "Production is unaffected" is true of
v2.17.0 and misleading about the release in flight.

---

### Worth doing — these do not block merge

**6. `chainedKeyRetriever` re-dials auth-api on every `Verify`.**

authlib caches an answered-but-unknown `kid`, never a fetch error, so source 0
never populates its cache off Cloud. Two independent tests measured 10 auth-api
fetches against 1 stack fetch over 10 verifications.

The reachable cost is small. The two completion-records read routes have no
frontend caller. `fetchCustomGuideRepository` returns early when the aggregation
toggle is off, so a self-hosted stack never calls it. `POST /completion-records`
disarms for the session on its first 404. So a self-hosted stack pays about one
dead dial per browser session.

Fix: add a short negative cache on a fetch failure in `chainedKeyRetriever`. That
also makes §3:245 true.

**7. `CheckRedirect` compares only `Hostname()`.**

It accepts an https-to-http downgrade and a port change on the same host — the
outcome the comment says it prevents. It also replaces `defaultCheckRedirect`, so
the 10-redirect cap in net/http is gone.

Fix: compare `Scheme` and `Host`. Then add a hop cap.

**8. The contract anchor column still reads `#1411 → #1433`.**

This PR replaces part of the taxonomy that #1433 established.

Fix: bump the column to `#1411 → #1700`. Add an anchor row for the identity seam,
owned by `pkg/plugin/auth/id_token.go` and `pkg/plugin/app_platform_identity.go`.

**9. Three test gaps.**

- A `namespace` claim of `"*"` is asserted nowhere. That is the divergence from
  `types.NamespaceMatches` that the PR body names as deliberate.
- A slow-but-working first source is untested. Both stall tests stall forever, so
  neither covers a source that answers inside the full budget but not its slice.
- `TestVerifyIDToken_KeySetFetchedOnceWithinMaxAge` counts only the source that
  succeeds. That is why item 6 stayed invisible.

**10. Commit `3bda4d2d` belongs in its own PR.**

It adds the "Maintaining this file" section to `AGENTS.md`, and you state that it
is unrelated to the auth fix. It is a standing instruction to every future agent,
and no auth reviewer is scoped to judge it.

It also needs an exception. "Do not repeat what the codebase already shows"
conflicts with the tier list that `src/validation/architecture.test.ts` enforces,
and with the headings that `prevent-doc-drift` names by string.

---

### The e2e matrix is not coverage for this code

`tests/` names none of these routes, and `docker-compose.yaml` enables no
aggregation toggle. So no Playwright row on any Grafana version has ever run this
path. The green matrix is not evidence for this PR. Issue #1402 is the gate for
exactly this, and it is still open. Run it on a Cloud stack before v2.17.0 ships.

One good result. On Grafana 13.0 and later, `PluginContext.Namespace` comes from
`user.GetNamespace()`. The `namespace` claim in the ID token comes from the same
`request.GetNamespaceMapper(cfg)`, on the same org id, in the same process. So the
equality your gate depends on holds by construction, not by coincidence.

I verified `base_plugincontext.go` at v12.3.11, v12.4.9, v13.0.7 and v13.2.0.
Grafana 12.3 and 12.4 do not set the field at all, so the binding returns
`identity-unverifiable` there. Those routes were already unavailable on those
versions, so nothing regresses.

---

### Two findings I removed

- **Self-hosted forged-JWKS impersonation.** Three skeptics refuted it. Every
  route re-gates on `a.oboExchanger == nil` after the identity gate, and a
  self-hosted stack never holds a CAP token. The same host already carried that
  CAP token in cleartext at base. That is stronger than one forged `sub`.
- **The Cloud 401 retry loop.** Two of three skeptics refuted it. Your PR body
  declares the trade-off and the write-path cost, and a test pins it. On main
  today every Cloud write already sits in that loop, so this PR improves it.

---

### One more piece of evidence

I read `deployment_tools`. Both `ksonnet/lib/hosted-grafana/config.libsonnet:41`
and `ksonnet/environments/grafana-authz-service/config.jsonnet:66` use
`http://api-lb.auth.svc.cluster.local./v1/keys`, with the trailing dot. Your
constant matches production configuration exactly.

All 48 `prod-*` hosted-grafana clusters run auth-api, and
`ksonnet/environments/hosted-grafana/prod/iptables.libsonnet:208` adds the egress
rule only for clusters that run it. So reachability holds in production today. It
depends on a rule this repository cannot see. If that rule ever lapses, the
symptom is a silent `identity-unavailable` with no `signing-keys-unreachable`
token, because the stack still answers. That is the concrete cost of decision 3.

Jayclifford345 (approved):

PR #1713, opened 2026-08-26
fix(interactives): reset mounted sections with guide

### LINKED ISSUE #887: Reset Guide button doesn't visually reset sections; partial storage clear

rajvveer:
## Summary

Resetting a guide clears its mounted sections without a remount. The existing reset event now has explicit `section | content | path | global` scopes: section resets leave siblings alone, path resets target the deduplicated journey and milestone keys, and global resets reach every mounted live guide. This extends the reset contract recorded in the `interactive-engine` registry row; reset initiators retain persistence ownership.

An active section keeps its execution lock, blocking overlay, and action-monitor/scroll ownership until the pending operation settles. Reset cancels remaining work and discards late completion, including nested multi-step and guided callbacks. Guided resets cancel active listeners and pending setup/retries; stale local and controller callbacks cannot restore progress. User Cancel preserves credit for an action that already succeeded, so Resume starts at the next step. Startup requirement checks also stop before launching navigation fixes after cancellation.

Preview progress remains driven by step and section progress events. Dependent `section-completed:` requirements recheck after the persisted done bit is cleared; stale incomplete notifications are dropped on recompletion or unmount.

## Validation

- 133 tests passed across 9 guided-handler, guided, multi-step and section suites; 4 existing runner todos remain.
- 76 review-contract tests passed.
- TypeScript, ESLint and Prettier checks passed.
- Production build passed with asset-size warnings.
- Governance: 184 tests passed; 4 existing Windows path-separator assertions failed in `import-graph.test.ts`. Both the test and implementation are identical to `main`.
- Merged current `main` and migrated its shared `resetGuideProgress` helper to the scoped dispatcher.

Closes #887


src/components/interactive-tutorial/interactive-section.tsx
```diff
@@ -445,6 +445,27 @@ export function InteractiveSection({
...
+    }
+
+    const contentKey = getContentKey();
+    const handleProgressCleared = (event: Event) => {
+      const clearedContentKey = (event as CustomEvent<{ contentKey?: string }>).detail?.contentKey;
+      if (clearedContentKey !== contentKey) {
+        return;
+      }
+
+      dispatch({ type: 'CLEAR_ACK' });
```
moxious: [blocking] This listener cannot distinguish a guide reset from the existing section-level reset because both carry `{ contentKey }`. With one listener per mounted section, resetting one section clears acknowledgement, collapse, and done state for every sibling sharing that key; path/global resets using `contentKey: '*'` are ignored. Introduce a discriminated section/content/path/global scope, keep persistence mutation with one owner, and add sibling-isolation plus wildcard contract tests.

moxious (changes requested):
Thanks for fixing mounted-section reset synchronization — keeping guide progress visually consistent without a remount is useful.

Request changes: reset events need an explicit scope.

The new consumer turns `interactive-progress-cleared` into a reset command, but existing producers use the same `{ contentKey }` payload for section- and guide-wide resets and use `'*'` for path/global resets. This over-clears sibling section state while under-resetting wildcard targets. The contract-evolution check classified this as blocking contract branching; three independent refutation passes confirmed it.

Tech debt: A2 — `InteractiveSection` remains a god component, and this change adds another event and persistence responsibility. After fixing the event contract, consider extracting the listener lifecycle into a focused hook. This is non-blocking.

Doc drift: no updates required.

Instrumentation: skipped because this is a bug fix of existing behavior.

CI attribution: the only failed job is `publish-report`, caused by fork comment permissions; it is infrastructure, not contributor-caused.

Once the reset scope is explicit and the sibling/wildcard regression tests are added, push an update and we’ll pick this up again.

src/components/interactive-tutorial/interactive-section.tsx
rajvveer: Addressed in 7fc81e2. `interactive-progress-cleared` now has a discriminated `section | content | path | global` payload owned by `lib/event-names.ts`. Section consumers match both `contentKey` and `sectionId`; path resets carry the exact cleared content keys; global resets no longer overload `*`. Reset initiators retain persistence ownership, while mounted listeners update ephemeral UI only. Added sibling-isolation, path-target, global-target, and preview-sibling regressions. Focused 73 tests, typecheck, ESLint, Prettier, the contract gate, and the production build pass.

rajvveer: Follow-up pushed in 8d33b9d after rebasing onto current main. The listener now resolves `getContentKey()` when the event fires. A single `resetLiveSectionState` callback cancels an active run, clears execution and blocking state, bumps child resets, and resets requirement-manager state; cancellation guards prevent an in-flight action from writing progress back after the clear. Section transitions back to incomplete now emit `kind: 'section', completed: false`, removing preview ghost keys. Added regressions for reused guide keys, mounted child reset, in-flight cancellation, single reset dispatch, and step redo. Focused 76 tests, typecheck, ESLint, Prettier, contract tests, and production build pass.

src/components/block-editor/hooks/useGuidePreviewProgress.ts
```diff
@@ -48,17 +57,29 @@ export function useGuidePreviewProgress(progressKey: string): GuidePreviewProgre
...
+          if (detail.kind === 'section') {
+            const prefix = `step:${detail.sectionId}:`;
+            for (const activeKey of activeProgress.current) {
+              if (activeKey.startsWith(prefix)) {
+                activeProgress.current.delete(activeKey);
+              }
+            }
+          }
+        }
+        setHasProgress(activeProgress.current.size > 0);
```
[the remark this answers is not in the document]
rajvveer: Fixed in 3e1a801. A section-incomplete event now removes only the section marker; individual step events remain the authority for step markers, so surviving completed steps keep the preview Reset control visible. Added a two-step regression that redoes one completed step, verifies progress remains, then clears the final step and verifies it drops.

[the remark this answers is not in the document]
rajvveer: Fixed in 3e1a801. Reaching zero through Redo now emits the existing guide progress event with `hasProgress: false`; it no longer emits `interactive-progress-cleared`. Content/path/global reset commands are reserved for explicit reset actions, so mounted sibling sections are untouched. The regression asserts the zero guide event and proves that no reset command is dispatched.

moxious (changes requested):
Thanks for reworking this into a discriminated reset contract — declaring the four scopes in `src/lib/event-names.ts` and routing every producer through a single dispatcher is exactly what the previous round asked for, and it removes the `'*'` sentinel that no consumer ever decoded. The contract-evolution scan re-ran against this head and came back `coherent_extension`, so that blocker is cleared.

Two things need fixing before merge, both in the section runner rather than the event contract: the mid-run reset releases `isRunning` while the async loop is still in flight, and the new post-`executeStep` cancel check cannot tell a reset from the user's own Cancel button. Details and a suggested fix for each are below. Push an update when they are addressed and I will pick this up again; the follow-ups and suggestions do not block merge.

## Merge contract

1. [blocking] **run-latch-released-while-loop-still-in-flight — A reset arriving mid-run releases the section's only mutual-exclusion latch, allowing two concurrent runs** (high · interactive-engine)
   resetLiveSectionState (interactive-section.tsx:1094-1101) synchronously sets isCancelledRef, clears isRunning, and calls stopSectionBlocking / endProgrammaticScroll / ActionMonitor.forceEnable while the async run loop is still awaiting an action. isRunning is handleDoSection's only mutual-exclusion guard (:678) and there is no generation token. On the floating-panel and full-screen surfaces the section is not unmounted by the reset's tab reload (FloatingPanelContent.tsx early-returns only on !content and keys ContentRenderer on content.url, which an internal_reload does not change), so the Do/Resume button is live for the whole reload window. A second click re-enters handleDoSection and sets isCancelledRef.current = false (:701), un-cancelling the first loop; two loops then run concurrently against the live Grafana UI, and the first loop's finally (:1015-1025) tears down the second run's blocking overlay and ActionMonitor suppression because stopSectionBlocking ignores its sectionId and clears a singleton flag. At base setIsRunning(false) was never called from outside the loop while it was in flight.
   Required: When isRunning, have resetLiveSectionState set only isCancelledRef.current = true and let the loop's existing finally perform setIsRunning(false), stopSectionBlocking, endProgrammaticScroll and forceEnable. If the reset must be visible before the loop unwinds, gate handleDoSection on a separate in-flight ref set at run entry and cleared in the finally, rather than on the React isRunning state.

2. [blocking] **post-execute-cancel-break-changes-user-cancel-semantics — The new post-executeStep cancel check also fires on the user's Cancel button, dropping credit for an action that already ran** (medium · interactive-engine)
   The new check at interactive-section.tsx:955-961 sits between `const success = await executeStep(stepInfo)` and the success branch, so markStepCompleted and handleStepComplete are skipped for a step whose action already took effect in Grafana. isCancelledRef is written by exactly two places: the new resetLiveSectionState and handleSectionCancel (:246-249), which is wired to the Cancel button rendered for the whole run and not covered by the blocking overlay. The loop cannot tell the two causes apart. Because currentStepIndex is derived from completedSteps and drives the Resume affordance, the cancelled step stays under the cursor and Resume re-runs show+do for an already-applied action, a real double-apply for button targetActions. The bulk markStepsCompleted at :1006-1012 is gated on !isCancelledRef so it does not restore the credit. At base a cancel landing there still credited the step.
   Required: Distinguish the two causes, for example a cancelCauseRef of 'user' | 'reset' set by handleSectionCancel and resetLiveSectionState respectively, and take the drop-completion plus setResetTrigger exit only for 'reset'.

## Follow-ups

1. [follow_up] **contract-anchor-must-be-a-registry-row — The new reset contract is recorded where the anchor extractor cannot read it** (medium · cross-cutting-architecture)
   The addition to docs/design/CONCERN_DETAILS.md is a level-4 heading plus prose placed after the contract-anchor table, not a row in it. concern-context.mjs:40 reads anchors only through findTable(detailMarkdown, ['concern','anchor','contract']) and matches rows by concern id at :53, and findTable stops at the first line that does not start with a pipe. Running the unmodified extractor over the head blobs returns the same 12 anchor rows as base, with no interactive-progress-cleared among them, and validateConcernRegistry reports no new error, so no test catches it. A later scan therefore routes to interactive-engine and reports a recorded anchor whose prose says nothing about this contract, which is false assurance rather than a missing record. A row keyed interactive-progress-cleared would fail 'Unknown concern reference' and a second interactive-engine row would fail 'Duplicate contract anchor'.
   Follow-up: Extend the existing interactive-engine anchor row in place, bumping its evidence to #1616 -> #1713 and appending the new prose after a <br>, which is the house style main already uses on that row.

2. [follow_up] **global-scope-widens-reset-blast-radius — global and path resets now reach every listener where the '*' sentinel reached only the context engine** (medium · cross-cutting-architecture)
   At base every listener compared detail?.contentKey === progressKey, so the '*' payload dispatched by MyLearningTab and the path reset matched nothing except the payload-blind context.hook refresh. At head isProgressClearForContent returns true for scope 'global' (event-names.ts:30-31) and isProgressClearForSection delegates to it, so a global clear now matches every mounted section regardless of content key and invokes resetLiveSectionState, which aborts a run in progress. The same widening applies to scope 'path' for keys in contentKeys. This is plausibly the intended fix, but it is a real change in blast radius that the PR describes as a scoped-payload refactor, and no test covers a global clear arriving while a section is running.
   Follow-up: State the widening explicitly in the PR body and the contract anchor, and add a test that a global clear arriving while a section isRunning leaves the UI coherent: blocking released, ActionMonitor re-enabled, no orphaned programmatic scroll.

3. [follow_up] **section-scoped-reset-no-longer-reaches-content-listeners — The comment citing #842 Bug 2 now describes wiring that no longer exists** (low · cross-cutting-architecture)
   isProgressClearForContent (event-names.ts:21-34) returns false for scope 'section', so the section-scoped dispatch at interactive-section.tsx:1173 reaches none of the content-level listeners. The comment immediately above it (:1163-1170) still states that the dispatch exists so useGuidePreviewProgress sees the section-level reset path and keeps the block-editor Reset guide button correct. That is no longer true; the guard has relocated to the per-step and per-section progress events, which nothing records.
   Follow-up: Rewrite the comment to state the actual mechanism, namely that a section-scoped event is consumed by the section's own listener and that preview reset-button visibility is now maintained by the per-step and per-section progress events.

## Suggestions

1. [suggestion] **path-scope-content-keys-must-list-everything-cleared — clearedContentKeys omits the journey keys the same reset clears** (low · correctness-and-reliability)
   In the url branch of resetPath, journeyKeys is built as [path.url, ...] and cleared via journeyCompletionStorage.clearMany(journeyKeys), but clearedContentKeys is then assigned milestoneKeys only (learning-paths.hook.ts:430-438). The path-scoped event therefore omits path.url and any journey-completion-only key, while the new anchor states that path carries the exact content keys cleared by that reset. A tab open on the journey landing URL will not have its progress flag cleared.
   Suggested: Include journeyKeys in clearedContentKeys, deduplicated, or soften the anchor wording to the content keys this reset targeted.

2. [suggestion] **section-incomplete-event-not-consumed-for-dependent-reblocking — The new completed:false section event is inert, so section-completed dependents are not re-blocked** (low · requirements-manager)
   The effect at interactive-section.tsx:472-486 emits dispatchProgress({ kind: 'section', sectionId, completed: false }) and its comment states the intent that section-completed: checks on dependent steps re-block until the user re-completes the section. The only section-completed-aware subscriber gates on detail.completed (step-checker.hook.ts:967-969), so the false value never triggers checkStep. The addition is otherwise coherent: hasEmittedGuideCompletionRef starts false and the dispatch is guarded on wasCompleted, so it cannot fire on mount, and no subscriber mis-handles it.
   Suggested: Extend the step-checker subscriber to also recheck on kind === 'section' && !detail.completed when the step's requirements include section-completed:, with a test.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1713
Summary: reset mounted sections with guide
Verdict: Request Changes
Results: 2 blockers, 2 non-blocking findings, 3 follow-ups

moxious:
We appreciate the contribution, but the branch is moving fast and this has picked up a number of merge conflicts; I'm assuming that this PR has been abandoned, we're open to further contributions though.

PR #1717, opened 2026-08-27
fix(ci): pin release workflow Go toolchain, cut release-prep over to release branches

Jayclifford345:
## Summary

- `release.yml` had no Go setup step, so it relied on whatever Go the `ubuntu-latest` runner image ships with. `go.mod` requires `go 1.26.5`; the runner had `1.25.14`, and `GOTOOLCHAIN=local` (set by the mage-action) blocks it from fetching the right version automatically. Result: `mage` fails with `go.mod requires go >= 1.26.5 (running go 1.25.14; GOTOOLCHAIN=local)`, and the v2.17.0 tag push failed exactly this way — as did an earlier v2.17.0 attempt on 2026-08-24.
- Adds the same `actions/setup-go@v7` step (with `go-version-file: go.mod`) that every job in `ci.yml` already runs, before the `grafana/plugin-actions/build-plugin` step.
- Updates the `release-prep` skill (`.cursor/skills/release-prep/SKILL.md` + its `.claude/` pointer stub) and `docs/developer/RELEASE_PROCESS.md` to cut a `release/v<version>` branch off `origin/main` at the *start* of prep, rather than working on `main` directly. Anything merged to `main` between drafting the changelog and tagging used to land in the release with no changelog entry (and `git push origin main` never actually worked anyway, since `main` is protected). Freezing scope on a dedicated branch the moment prep starts closes that race — everything merged to `main` afterward just rolls into the next release.
- Also documents that `publish.yml`'s `branch` workflow-dispatch input should point at the `release/v<version>` branch, not `main` — that input re-resolves live at dispatch time, so pointing it at `main` would silently reintroduce the same race for the Cloud publish step even though the GitHub tag itself (which pins an exact SHA) was already race-free.

## What to look at

- `.github/workflows/release.yml` — the added `setup-go` step
- `.cursor/skills/release-prep/SKILL.md` — restructured Phase 0 (cut the branch first), updated hard constraints, Phase 5 hand-off commands
- `docs/developer/RELEASE_PROCESS.md` — official release steps + dev/prod deployment steps now reference the release branch

## Why

Discovered live while cutting v2.17.0: the changelog PR (#1694) merged, another PR (#1711) landed on `main` in the gap before the release branch existed, and then the tag push itself failed on the Go-version gap. All three of those were real, so this PR fixes the CI gap and updates the documented process so the next release doesn't hit the same two problems.

## How to verify

```bash
npm run check
npm run build
npx --yes npm@12.0.2 run test:ci -- src/validation/skill-references.test.ts src/validation/control-bytes.test.ts --coverage=false
```

Both pass on this branch. The real proof is the next tag push actually building — this PR doesn't retest `release.yml` itself since it only fires on tag push.

## Breaking changes

None — CI/process only.

### DIFF
--- a/.claude/skills/release-prep/SKILL.md
+++ b/.claude/skills/release-prep/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: release-prep
-description: Orchestrate the pre-release flow for grafana-pathfinder-app — bump the version in package.json, draft a CHANGELOG entry (via the `changelog` skill), run `npm run check` and `npm run build`, then print the exact `git tag` command for the user to execute. Never creates or pushes the tag itself; tagging is a one-way door the user owns.
+description: Orchestrate the pre-release flow for grafana-pathfinder-app — cut a release branch off origin/main to freeze scope, bump the version in package.json, draft a CHANGELOG entry (via the `changelog` skill), run `npm run check` and `npm run build`, then print the exact `git push`/`git tag` commands for the user to execute. Never pushes the branch or the tag itself; both are one-way doors the user owns.
 ---
 
 Read `.cursor/skills/release-prep/SKILL.md` and follow it exactly.
--- a/.cursor/skills/release-prep/SKILL.md
+++ b/.cursor/skills/release-prep/SKILL.md
@@ -1,31 +1,33 @@
 ---
 name: release-prep
-description: Orchestrate the pre-release flow for grafana-pathfinder-app — bump the version in package.json, draft a CHANGELOG entry (via the `changelog` skill), run `npm run check` and `npm run build`, then print the exact `git tag` command for the user to execute. Never creates or pushes the tag itself; tagging is a one-way door the user owns.
+description: Orchestrate the pre-release flow for grafana-pathfinder-app — cut a release branch off origin/main to freeze scope, bump the version in package.json, draft a CHANGELOG entry (via the `changelog` skill), run `npm run check` and `npm run build`, then print the exact `git push`/`git tag` commands for the user to execute. Never pushes the branch or the tag itself; both are one-way doors the user owns.
 ---
 
 # Release prep
 
-Pre-release orchestrator. Handles the safe, reversible parts of cutting a release (version bump, changelog, validation) and stops short of the irreversible step (tag push) so the user reviews everything before the GitHub release workflow fires.
+Pre-release orchestrator. Handles the safe, reversible parts of cutting a release (branch, version bump, changelog, validation) and stops short of the irreversible steps (branch push, tag push) so the user reviews everything before the GitHub release workflow fires.
 
 This skill pairs with `.cursor/skills/changelog/SKILL.md`, which drafts the actual CHANGELOG entry. `release-prep` is the wrapper that runs `changelog` plus the surrounding validation.
+
+## Why a release branch instead of prepping on main
+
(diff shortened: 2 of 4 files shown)


moxious (changes requested):
## Merge contract

1. [blocking] **PR1717-001 — Resolve the version before target-dependent release steps** (high · ai-subsystem)
   .cursor/skills/release-prep/SKILL.md:47-61 checks `## <version>` and creates `release/v<version>` before Phase 1 resolves or validates the target at lines 71-97, so the advertised no-argument flow cannot execute as written and invalid input can mutate local refs before rejection.
   Required: Capture the frozen main SHA and authoritative tags first, then resolve and validate the explicit or inferred version before the version-specific CHANGELOG check or branch creation.

2. [blocking] **PR1717-002 — Bind changelog collection to the frozen release SHA** (high · reversibility-and-one-way-door · irreversible without cleanup)
   .cursor/skills/release-prep/SKILL.md:40-59 freezes origin/main, but its delegated changelog query at .cursor/skills/changelog/SKILL.md:47-53 reads live main with only a lower date bound, so a post-cut merge can appear in notes without its code and then be omitted from the next release notes.
   Required: Pass the captured SHA through version suggestion and changelog drafting, and derive the included PRs from an immutable previous-release-to-frozen-SHA scope, explicitly accounting for previous release tags that are not ancestors of main.

3. [blocking] **PR1717-003 — Fetch authoritative tags before deriving the release boundary** (high · build-and-ci · partially reversible)
   .cursor/skills/release-prep/SKILL.md:43 fetches only origin/main and then treats local tags as authoritative at lines 63-79, but tags on the new off-main release branches are not guaranteed to follow that fetch, so a stale clone can choose the wrong previous release or miss an existing target tag.
   Required: Fetch or query the remote tag namespace explicitly before every tag-derived check, fail closed on local/remote conflicts, and make changelog scoping consume the same authoritative boundary.

## Suggestions

1. [suggestion] **PR1717-004 — Keep Cloud publish pinned to the release commit** (high · cross-cutting-architecture · irreversible without cleanup)
   .cursor/skills/release-prep/SKILL.md:191-200 replaces mutable main with the still-mutable `release/v<version>` ref, while .github/workflows/publish.yml:41-43 forwards that branch without verifying it equals the version tag, so a later branch update could make Cloud receive a different commit from the GitHub release.
   Suggested: Publish from the immutable tag or SHA when supported; otherwise fail closed unless the remote release branch and version tag resolve to the same commit before ops or prod dispatch.

2. [suggestion] **PR1717-005 — Exercise the changed release contract before the next tag** (medium · testing-and-verification)
   The passing checks do not execute release.yml or release-prep semantics, and skill-references.test.ts checks only reference/frontmatter parity, so the changed phase ordering, tag discovery, and frozen-scope behavior currently rely on manual review.
   Suggested: Add non-publishing focused verification for explicit and inferred versions, CHANGELOG cutoff behavior, authoritative tag discovery, and branch/tag publish identity.

3. [suggestion] **PR1717-006 — Guard the release Go pin against go.mod drift** (medium · build-and-ci · partially reversible)
   .github/workflows/release.yml:26 duplicates go.mod:3 as a hard-coded `1.26.5` while PR CI derives the toolchain from go.mod and no validation couples the release-only literal to that source of truth.
   Suggested: Derive the build-plugin input from go.mod or add a repository check that fails whenever release.yml and go.mod diverge.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1717
Purpose: pin release workflow Go toolchain, cut release-prep over to release branches
Verdict: Request Changes
3 blocking, 3 suggestions, 0 nits

PR #1719, opened 2026-08-27
fix(custom-guides): carry manifest stats through the catalogue proxy

tomglenn:
Pathfinder stamps a guide's block statistics onto its manifest at build time, and those counts are the denominator for completion percentages. For guides stored in the platform, the catalogue reaches the browser through this repo's proxy, which decodes each manifest into a hand-written struct. That struct had no field for `stats`, so JSON decoding dropped the counts before the catalogue was ever shaped — nothing downstream could read them.

This adds a typed `Stats` field to `customGuideManifest`, with the five members the stamp carries. `version` is the version of the counting rules, not a content version.

## What to look at

`customGuideStats.UnmarshalJSON` in `pkg/plugin/custom_guide_repository_client.go`. A stamp counts as usable only when the stored object supplies all five members; anything else reads as absent. Zero is a legitimate count, so a missing member cannot be detected by inspecting decoded values — a partial stamp would otherwise be zero-filled into a plausible-looking denominator that passes the reader's own validation. It reports no error, so an unusable stamp degrades to an absent one rather than failing the spec around it.

The decode loop in `ListPage` is the other half: a spec this hand-written mirror disagrees with neither fails the page nor drops the guide. Failing the page would answer 503 for the whole namespace indefinitely, since a malformed stored guide never self-heals. Per-page warnings are capped.

## Two things a reviewer should know

Platform-stored guides show no numbers yet. The CRD's `#Manifest` declares no `stats` and prunes a top-level write silently, so both writers stamp `additionalFields.stats`, which this proxy does not carry. The counts appear once the typed field lands in `grafana-pathfinder-backend`; that work is in flight and the two are coupled.

A guide whose stats cannot be read, or are incomplete, still appears in the catalogue — just without numbers.

## Scope

The struct, its tests, and the wire contract that mirrors it: the Go struct-tag golden and the TypeScript schema that `src/validation/backend-api-contract.test.ts` checks it against. Application code under `src/` is untouched.


moxious (changes requested):
## Merge contract

1. [blocking] **GO-PARTIAL-COMPOSITE-DECODE — Partial decoding launders malformed composite fields** (high · go-backend)
   ListPage retains an entry after any json.Unmarshal error, but encoding/json can leave slices and nested structs partially populated, so malformed upstream manifests can become valid-looking catalogue data instead of degrading at the affected field boundary.
   Required: Decode fallible composite fields independently through json.RawMessage and retain each field only after it decodes successfully; add regression cases for malformed milestones and author values.

2. [blocking] **GSC-STATS-NONNEGATIVE — The proxy branches the canonical stats contract** (high · guide-schema-and-contracts)
   The Go completeness check accepts negative counts and CustomGuideStatsWireSchema declares them valid with unconstrained z.int(), contradicting the canonical GuideStatsSummarySchema non-negative invariant and allowing an invalid completion denominator onto the wire.
   Required: Treat any negative stats member as an absent stamp, align the wire schema with GuideStatsSummarySchema, and add a negative-value case for each member.

## Suggestions

1. [suggestion] **GO-DECODE-WARN-BUDGET — Decode-warning suppression resets on every page** (medium · go-backend)
   The warning counter is local to ListPage, so a multi-page namespace-wide schema drift receives a fresh detail budget and suppression summary on every page, weakening the intended log-flood bound across one catalogue drain.
   Suggested: Share one warning budget across the full paginated drain, emit one final suppression summary, and cover the multi-page case with a captured logger.

2. [suggestion] **CR-QC8-NARRATING-TESTS — QC8.1 test comments narrate the assertions** (medium · correctness-and-reliability)
   Several new comments in custom_guide_repository_test.go restate the following test name and assertions, creating duplicate prose that must drift with the executable contract.
   Suggested: Remove the narrating comments and retain only the short explanation of counterintuitive invariants such as why missing counts cannot be decoded as zero.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1719
Purpose: carry manifest stats through the catalogue proxy
Verdict: Request Changes
2 blocking, 2 suggestions, 0 nits

moxious (approved):

## Follow-ups

1. [follow_up] **TEST-GO127-RAW-MESSAGE-GOLDEN — The reflected contract golden fails under Go 1.27** (medium · testing-and-verification)
   The documented `npm run test:go` command fails under Go 1.27 because `json.RawMessage` reflects as `jsontext.Value`, changing the generated tag golden. The same failure reproduces at the prior reviewed head, while the repository's Go 1.26.5 toolchain passes, so this is pre-existing rather than induced by this PR.
   Follow-up: Normalize `json.RawMessage` and `jsontext.Value` to the same wire descriptor in the reflected contract inventory, and add a toolchain-stability regression test.

## Suggestions

1. [suggestion] **GO-DECODE-WARN-BUDGET — The suppression summary disappears when the drain stops at its entry budget** (low · go-backend)
   `ListPage` emits the summary only for an empty continue token (custom_guide_repository_client.go:297-301), but `drainCustomGuides` also terminates at its aggregate entry budget with a non-empty token (custom_guide_repository.go:215-228). A probe reproduced ten detail logs and zero summary logs on that path.
   Suggested: Finalize the warning budget on every drain terminal path, including aggregate-budget truncation, and cover that path through `drainCustomGuides`.

2. [suggestion] **TESTING-COMPOSITE-SIBLING-SURVIVAL — The field-local degradation test does not prove valid sibling composites survive** (low · testing-and-verification)
   The malformed-composite cases provide one valid sibling composite (custom_guide_repository_test.go:764-780), but the assertions at lines 808-818 check only scalars and absence of the malformed value. A decoder that drops every composite after one failure would still pass.
   Suggested: Assert that author survives the malformed-milestones case and milestones survive the malformed-author case.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1719
Summary: carry manifest stats through the catalogue proxy
Verdict: Approve with Minor

PR #1723, opened 2026-08-28
fix: read plugin settings back immediately before every write

Jayclifford345:
## What

`POST /api/plugins/:id/settings` is a **whole-object write**. A body without `pinned` unpins the plugin; one without `enabled` disables it; one without a provisioned `jsonData` field such as `stackId` drops it.

Two write paths were getting that wrong:

- `ConfigurationForm` sent `plugin.meta.enabled` and `plugin.meta.pinned` — a mount-time snapshot. Another tab saving after this page opened made it a stale snapshot that silently restored the old values.
- `dev-mode.ts` sent no `pinned` at all, so toggling dev mode unpinned the plugin.

Both now `await fetchPluginSettings(...)` immediately before the write and take `enabled`, `pinned` and the base `jsonData` from that read. The stale `resolvedEnabled` / `resolvedPinned` state is deleted rather than left as a trap, and a failed read-back writes nothing at all.

`fetchPluginJsonData` becomes a thin wrapper over the new `fetchPluginSettings`, so `usePathfinderPluginConfig` is unchanged.

## Why this is its own PR

It rode along on [#1668](https://github.com/grafana/grafana-pathfinder-app/pull/1668) and has nothing to do with that PR's subject. Split out on review so it can be judged on its own — and because the read-back rewrite there had dropped a field it should not have (below).

## The field this tab does not own

`enableAiAutoHeal` is passed through **raw**, not resolved:

```ts
...getConfigWithDefaults(current.jsonData),
enableAiAutoHeal: current.jsonData?.enableAiAutoHeal,
```

Spreading `getConfigWithDefaults` alone bakes today's `DEFAULT_ENABLE_AI_AUTO_HEAL` into `jsonData` for a field this tab has no opinion about, and the value can then never again follow a change to that default for the stack. `src/components/AppConfig/settings-preservation.test.ts` already demonstrated the shape; it does not exercise the component, which is why the regression could pass unnoticed.

## How to verify

Two `data-testid`-driven cases in the new `ConfigurationForm.test.tsx` prove the raw pass-through on **both** write paths:

- `leaves a never-set enableAiAutoHeal unset on both write paths` — fails without the pass-through.
- `keeps an explicit enableAiAutoHeal choice on both write paths` — `false` is a decision, not an absent value, and must survive as faithfully as omission does.

Plus the read-back itself: another tab flipping `pinned` and adding a provisioned field between mount and save, the assistant dev-mode toggle taking the same path, and a failed read-back writing nothing. `utils.plugin.test.ts` covers `fetchPluginSettings`' defaults — a missing `enabled` reads as `true` (Grafana omits it when enabled), a missing `pinned` as `false`, and the plugin id is URL-encoded.

`npm run check` clean.

## Reversibility

Fully reversible — three files, no schema, no stored state, no flag.

### DIFF
--- a/src/components/AppConfig/ConfigurationForm.test.tsx
+++ b/src/components/AppConfig/ConfigurationForm.test.tsx
@@ -1,0 +1,147 @@
+/**
+ * `POST /api/plugins/:id/settings` replaces the whole object, so anything the
+ * form does not send is cleared: a body without `pinned` unpins the plugin, one
+ * without `enabled` disables it, and one without a provisioned `jsonData` field
+ * such as `stackId` drops it. The mount-time read is a seed for the form and
+ * goes stale the moment another tab saves, so what matters here is that every
+ * write re-reads first.
+ */
+
+import React from 'react';
+import { render, screen, waitFor, fireEvent } from '@testing-library/react';
+import type { AppPluginMeta, PluginConfigPageProps } from '@grafana/data';
+
+import ConfigurationForm from './ConfigurationForm';
+import { fetchPluginSettings, updatePluginSettings } from '../../utils/utils.plugin';
+import { testIds } from '../../constants/testIds';
+import type { DocsPluginConfig } from '../../constants';
+
+jest.mock('../../utils/utils.plugin', () => ({
+  ...jest.requireActual('../../utils/utils.plugin'),
+  fetchPluginSettings: jest.fn(),
+  updatePluginSettings: jest.fn(),
+}));
+
+jest.mock('./CodaBackendStatus', () => ({ CodaBackendStatus: () => null }));
+
+jest.mock('@grafana/runtime', () => ({
+  ...jest.requireActual('@grafana/runtime'),
+  config: { ...jest.requireActual('@grafana/runtime').config, bootData: { user: { id: 7 } } },
+}));
+
+/** Dev mode for user 7 is what renders the assistant toggle, the other write path. */
+const DEV_MODE_SETTINGS = {
+  jsonData: { stackId: '123', devMode: true, devModeUserIds: [7] } as DocsPluginConfig,
+  enabled: true,
+  pinned: true,
+};
+
+const mockedFetch = fetchPluginSettings as jest.MockedFunction<typeof fetchPluginSettings>;
+const mockedUpdate = updatePluginSettings as jest.MockedFunction<typeof updatePluginSettings>;
+
+const MOUNT_SETTINGS = { jsonData: { stackId: '123' } as DocsPluginConfig, enabled: true, pinned: true };
+
+function renderForm() {
+  const plugin = {
+    meta: { id: 'grafana-grafanadocsplugin-app', enabled: false, pinned: false, jsonData: {} },
+  } as unknown as PluginConfigPageProps<AppPluginMeta<DocsPluginConfig>>['plugin'];
+
+  return render(<ConfigurationForm plugin={plugin} query={{}} />);
+}
+
+beforeEach(() => {
+  jest.clearAllMocks();
+  mockedUpdate.mockResolvedValue({});
+});
+
+describe('ConfigurationForm settings writes', () => {
+  it('re-reads enabled, pinned and jsonData immediately before saving', async () => {
+    mockedFetch.mockResolvedValueOnce(MOUNT_SETTINGS);
+    renderForm();
+    await waitFor(() => expect(mockedFetch).toHaveBeenCalledTimes(1));
+
+    // Another tab saves after this page opened: pinned goes false, and a
+    // provisioned field appears. The form owns neither.
+    mockedFetch.mockResolvedValueOnce({
+      jsonData: { stackId: '123', slug: 'moved' } as DocsPluginConfig,
+      enabled: true,
+      pinned: false,
+    });
+
+    fireEvent.submit(screen.getByTestId(testIds.appConfig.form));
+
+    await waitFor(() => expect(mockedUpdate).toHaveBeenCalled());
+    expect(mockedFetch).toHaveBeenCalledTimes(2);
+    const [, body] = mockedUpdate.mock.calls[0]!;
+    expect(body.enabled).toBe(true);
+    expect(body.pinned).toBe(false);
+    expect(body.jsonData).toMatchObject({ stackId: '123', slug: 'moved' });
+  });
+
+  it('re-reads before the assistant dev-mode toggle writes too', async () => {
+    mockedFetch.mockResolvedValueOnce(DEV_MODE_SETTINGS);
+    renderForm();
+    const toggle = await screen.findByTestId(testIds.appConfig.assistantDevModeToggle);
+
+    mockedFetch.mockResolvedValueOnce({ ...DEV_MODE_SETTINGS, pinned: false });
+    fireEvent.click(toggle);
+
+    await waitFor(() => expect(mockedUpdate).toHaveBeenCalled());
+    const [, body] = mockedUpdate.mock.calls[0]!;
+    expect(body.pinned).toBe(false);
+    expect(body.jsonData).toMatchObject({ stackId: '123', enableAssistantDevMode: true });
+  });
+
+  it('leaves a never-set enableAiAutoHeal unset on both write paths', async () => {
+    // This tab does not own the field. Spreading `getConfigWithDefaults` alone
+    // would bake today's default into jsonData, and the field could then never
+    // again follow a change to DEFAULT_ENABLE_AI_AUTO_HEAL for this stack.
+    mockedFetch.mockResolvedValue(DEV_MODE_SETTINGS);
+    renderForm();
+    const toggle = await screen.findByTestId(testIds.appConfig.assistantDevModeToggle);
+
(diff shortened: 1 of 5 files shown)


Jayclifford345:
Split out of #1668 on review. It was riding along there, is unrelated to that PR's subject, and the read-back rewrite had dropped the raw `enableAiAutoHeal` pass-through in both write paths — restored here with component tests on each. @moxious added, since you raised the finding (`configuration-save-materializes-ai-auto-heal-default`).

moxious (changes requested):
## Merge contract

1. [blocking] **merge-breaks-coda-flag-mock — Merged with current main, the PR breaks ConfigurationForm.coda-flag.test.tsx (6 failures) and nothing forces CI to re-run before merge** (high · testing-and-verification)
   The branch is 54 commits behind main and its green checks date from 2026-08-28. `ConfigurationForm.coda-flag.test.tsx` landed on main after this branch forked, and it mocks `../../utils/utils.plugin` with only `{ fetchPluginJsonData, updatePluginSettings }` and no `jest.requireActual` spread. Once this PR makes `ConfigurationForm.tsx` import `fetchPluginSettings`, that mock leaves it `undefined`. Merging current main into the branch auto-merges cleanly with no textual conflict, then fails 6/6 in that suite with `TypeError: (0 , _utilsplugin.fetchPluginSettings) is not a function` at `ConfigurationForm.tsx:84`. The rest of the tree is fine: 1 failed suite, 531 passed. Nothing forces a re-check first - `ci.yml` triggers only on `push` to main and default `pull_request` types, so a base advance re-runs nothing, and the only required ruleset (`main-CI Gate`) sets `strict_required_status_checks_policy: false` with no merge queue. The stale green CI Gate satisfies the rule, so this can merge as-is and redden main.
   Required: Merge origin/main into the branch and update ConfigurationForm.coda-flag.test.tsx to mock fetchPluginSettings (resolving a { jsonData, enabled, pinned } snapshot), or spread jest.requireActual as the new ConfigurationForm.test.tsx does; re-run the suite before merge.

## Follow-ups

1. [follow_up] **sibling-write-paths-keep-the-fixed-bug — Two of the four modules that write plugin settings still send a mount-time snapshot, so the read-back invariant holds for only part of the surface** (medium · cross-cutting-architecture)
   `InteractiveFeatures.tsx:38` and `TermsAndConditions.tsx:20` still destructure `{ enabled, pinned, jsonData }` from `plugin.meta` and POST them with no read-back - the exact shape this PR removes from `ConfigurationForm`. `plugin.meta` is the shared `AppPlugin` singleton registered in `module.tsx:141-156`; it is not refetched per config tab, and `usePathfinderPluginConfig` only writes `window.__pathfinderPluginConfig`, which none of these three components read. Each sibling save reloads the page ~100ms later, so the same-page-load "save tab A then tab B" sequence is forestalled; the residual window is the cross-tab, second-admin, or provisioning drift this PR judged worth closing for `ConfigurationForm`. `updatePluginSettings` carries no doc comment at head (this PR deleted the one line that was there) and its `Partial<PluginMeta>` signature reads as a partial update, so the new invariant lives only in `ConfigurationForm.tsx:235-239` where the other callers never see it. Pre-existing, so not a blocker for this PR.
   Follow-up: Either extend the read-back to InteractiveFeatures.tsx and TermsAndConditions.tsx, or move the read-back-before-write invariant onto updatePluginSettings as a doc comment - ideally a helper that takes the fresh snapshot - so the remaining callers inherit it.

2. [follow_up] **assistant-toggle-test-is-vacuous — The assistant dev-mode toggle test passes against fully unfixed code, and no test discriminates the pinned fix on either write path** (medium · testing-and-verification)
   The new test named "re-reads before the assistant dev-mode toggle writes too" passes against fully unfixed code. `renderForm()` sets `plugin.meta` to `{ enabled: false, pinned: false, jsonData: {} }`, and both write-path tests assert `body.pinned === false`, which the old `plugin.meta` behaviour supplies for free; the toggle test asserts no `enabled` and no fetch call count, and its `jsonData` assertion matches what the mount read already supplied. Mutation testing in an isolated worktree: reverting `onChangeAssistantDevMode` to its real base body leaves 5/5 green; changing `pinned: current.pinned` to `pinned: plugin.meta.pinned` on one path or both leaves 5/5 green; the control mutation on `enabled` correctly fails exactly 1 test. So `pinned` - the headline behaviour - is discriminated by nothing, and the toggle path is covered by nothing. The submit path is genuinely covered, via `enabled` and the read-back-failure test.
   Follow-up: Give the re-read fixture values that differ from both plugin.meta and the mount snapshot, assert body.enabled and the fetch call count on the toggle path, and add a jsonData field present only in the re-read (the submit test's slug: 'moved' trick) so the toggle path's jsonData source is pinned down too.

3. [follow_up] **dev-mode-hardcodes-enabled-true — dev-mode writes keep a hardcoded enabled: true while reading pinned back from the same fresh snapshot** (low · correctness-and-reliability)
   All three writes in `dev-mode.ts` (:79-87, :112-120, :136-144) keep a literal `enabled: true`, with `current.enabled` sitting unused on the object the PR just fetched, and the `pinned: current.pinned` line was added directly above it. The `fetchCurrentSettings` doc comment this PR rewrote claims the fetch exists "to preserve enabled/pinned", which is not what the call sites do. `toggleDevMode` is reachable from the plugin config page, which Grafana serves for a disabled app plugin too, so this can silently re-enable a disabled plugin. Pre-existing behaviour, but the surrounding change now makes it read as an oversight.
   Follow-up: Use enabled: current.enabled at all three sites, or, if enabling is intentional, say so in one line next to the literal and correct the fetchCurrentSettings doc comment.

4. [follow_up] **dev-mode-userids-still-stale — The dev-mode read-back protects every jsonData field except the one these functions actually own** (low · correctness-and-reliability)
   `dev-mode.ts:75` computes `updatedUserIds` from the caller-supplied `currentUserIds` argument, before the read-back on :77. Then :82-86 spreads the fresh `current.jsonData` - including its `devModeUserIds` - and immediately overwrites it with the stale-derived value; same shape at :115-119. The argument originates at `ConfigurationForm.tsx:106` as `resolvedJsonData?.devModeUserIds ?? []`, the mount-time seed. So the read-back now protects every jsonData field except the one these functions exist to manage: a dev user added by another admin after this form mounted is dropped from the allow-list by the next toggle.
   Follow-up: Derive the base list inside enableDevMode/disableDevModeForUser from current.jsonData.devModeUserIds after the read-back, keeping the parameter only as a fallback.

## Suggestions

1. [suggestion] **dev-mode-writes-have-no-tests — The three changed writes in dev-mode.ts have no test file at all** (medium · testing-and-verification)
   Three of the five fixed writes live in `dev-mode.ts`, and there is no `dev-mode.test.ts` anywhere in `src/utils/`. The only references outside the module are `jest.fn()` mocks in `src/security/url-validator.test.ts:21-23` and the dynamic import in `SelectorDebugPanel.tsx:43-44`, neither of which exercises the payload. The PR adds tests for the other two changed files.
   Suggested: Add src/utils/dev-mode.test.ts mocking ./utils.plugin and asserting, for each of the three exported writes, that fetchPluginSettings is called before updatePluginSettings, that pinned in the body comes from the read-back, and that unrelated jsonData keys survive.

2. [suggestion] **silent-save-abort-on-read-failure — A failed read-back makes Save a silent no-op with no user-visible feedback** (low · correctness-and-reliability)
   The read-back at `ConfigurationForm.tsx:240` is a new failure point inside `onSubmit` that did not exist at base. The catch at :272-275 does `logger.error` and `setIsSaving(false)` only - no Alert, no toast, no error state - so the missing page reload is the admin's only cue that nothing was saved. Writing nothing on a failed read is the right call; the two sibling handlers in the same file (:150-158, :186-192) surface their failures via `alert()`, so the file is now internally inconsistent.
   Suggested: Surface the failure - an Alert in the form, or the same alert() the sibling handlers use - distinguishing 'could not read current settings, nothing was saved' from a write failure.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1723
Summary: read plugin settings back immediately before every write
Verdict: Request Changes
Results: 1 blocker, 2 non-blocking findings, 4 follow-ups

Jayclifford345:
Closing as superseded by #1691, merged September 16, rather than reviving the old settings write paths.

Verified against current main (`ce1b0018`):
- `fetchPluginSettings` already reads fresh `jsonData`, `enabled`, and `pinned` with the same defaults and URL encoding introduced here.
- ConfigurationForm, InteractiveFeatures, and TermsAndConditions now use the shared `saveTenantSettings` boundary. It resolves current settings on every save and preserves the fresh plugin flags and unrelated/provisioned fields on the legacy fallback.
- Unset `enableAiAutoHeal` remains unset; explicit choices survive unrelated saves.
- Dev-mode opt-in now uses browser-local storage, removing the old plugin-settings writes, hardcoded enablement, and stale shared user-list mutation discussed in this review.
- The configuration form surfaces save failures, and the Coda flag suite now mocks the shared save boundary rather than the obsolete fetch API.

Validation on main: settings-preservation, ConfigurationForm.coda-flag, dev-mode-toggle, and dev-mode-opt-in suites all passed (4 suites, 32 tests). The implementation intent is already present through #1691; merging this older branch would add no needed behavior. No branch deletion or code changes made for this closure.

PR #1727, opened 2026-08-28
docs(design): record the completion model's design rationale

tomglenn:
Adds `docs/design/COMPLETION-MODEL.md`, the leave-behind rationale for how Pathfinder measures completion.

This is a **record of a decision already taken**, not a proposal. Reviewing it means checking that it describes the model faithfully and points at the right code, not re-opening the choices.

What it records:

- The formulas: guide progress as completed steps over total block count, path progress as the mean of its milestones' percentages, and why that means no path-wide total ever has to be pre-calculated.
- The unconditional "mark complete" button, and the library measurements that justify it — 308 of 598 published guides have no completable block at all, and only 9 can reach 100% by doing things today.
- The alternative that was considered and rejected, in full technical substance, because it is the most likely reason this design changes later.
- Four falsifiable bets, each with its evidence source and baseline.
- The open questions, still open, including the nudge's scope and the deferred UX review.
- How this relates to `grafana/pathfinder-rfcs#14`: which of its four options this model resolves, and why Option A's declared step counts are no longer required. That PR is untouched.

Also adds the doc to the `completion-records` concern's `load_docs` and to `CONTEXT_INDEX.md`, so a future change to the completion arithmetic surfaces the rationale. No product code.


moxious:
Good doc. This is my concept review on the PLG and enablement tension. 

## Positions taken

**The honor system is right.** Self-reported completion is sufficient. The alternative a knowledge gate that forces the reader to prove something is futile since AI is built into every tool now. A knowledge gate only makes the reader paste the challenge into another window. The content must add value on its own terms. It must not gate progress or force readers through hoops.  @Spidy88 has some great ideas around "Mastery proofs" we should consider over time, but those are more medium term and won't get us out of having to measure completion.

**Decision 7 is really important.** The `markCompleteClicked` event makes the honor system observable. It is the only instrument that separates "completed by doing things" from "completed by clicking a button." It must ship before or with the button, not after.  I think it will end up being an important data signal for @mariadubinets .

**Cross-path guide sharing: yes, get credit.** If a reader sets up Alloy in path A, path B must not ask them to set it up again. Completion is per guide, not per path context.  If we modularize content, and you complete modules, then you're partially done with other stuff. That's good (and accurate imo)

**Content versioning: guide updates must not affect completion.** A reader who completed a guide at version N stays completed at version N+1. Bake this into the assumptions. The stamped percentage uses the denominator at completion time. The system does not restate it.  **This means that completion % in rare situations where the guide is changing quickly can be slightly wrong** but I think this is an unusual case and we're gold-plating if we try to solve for that right off of the bat.

**Path membership changes: the old percentage persists.** Scenario, A reader at 50% on a 4-milestone path where the author adds a 5th milestone keeps 50% until they revisit. This makes the calculation slightly wrong. Known issue, not MVP scope. See last item.

**Button placement: fine as is.** A minor UX follow-up can look at discoverability. Not critical path.  We already had several conversations that amounted to "we're going to need another UX look later when we can get it" (I want it)

## Questions the doc must address

**The transition from old to new.** `calculatePathProgress` in `learning-paths.hook.ts` is binary per milestone. The new model is fractional. The doc does not describe how to get from A to B. Do both systems run at the same time? Does `milestoneCompletionStorage` carry forward? Is there a flag or a hard cut? "The series breaks" is a data-analysis instruction, not an implementation plan.

**Adopt dual-write semantics**.  We already have a pattern in the code for completion records themselves; you want the front-end to record an event and have it sent through both channels (rudderstack & app platform).  This should be a constraint on the design, so we don't spam single track inconsistent writes.

**Reset semantics.** Analytics events are immutable (a thing happened, it's append-only). App Platform records can be modified or deleted. If a reader resets guide 5 of a completed 5-guide path, is the path now incomplete? Position: probably not. A completed path stays completed. But the doc must state this. The existing `resetPath` clears five storage layers. The new model adds a sixth (durable completion records). The doc must specify which sources change on reset.  As a consequence, we might consider **removing the reset guide feature** as it adds a lot of complexity. We need to look at whether we need/want that feature, and if so, how it should work.

**The read path - who uses this stuff?** Two consumers exist: analytics events for internal analytics, and App Platform completion reads for the app. The doc should say this so downstream implementers do not try to build app features on but always go back to the app platform endpoints as a settled right way to get the info.

**What triggers downstream systems.** Badges, durable records, and the "continue learning" CTA all trigger on 100%. The doc does not specify the trigger. Is it the evidence kind, the `GuideProgress.complete` flag, the durable record write, or the percentage field? These are different code paths. They can disagree.

## Assumed risks

These are accepted and tracked. They do not block the PR.

**The percentage becomes a de facto contract.** Once dashboards and reports read the durable record percentage, the formula is frozen. A future change (weighted milestones, softer evidence) makes every historical record a different unit. You then either restate history or carry two series. Both are expensive after 6 months of data.

**Bet 4 (authors will instrument their content) will probably not land in 6 months.** Authors have roadmaps. A completable block in a prose guide is a design problem, not a markup problem. The 51.5% will stay near where it is.

**The nudge will become furniture.** By month 3, frequent nudges get dismissed on reflex. Infrequent nudges get forgotten. Bet 2 will probably be wrong. Plan to remove the nudge rather than tune it.

**The mark-complete button is a single point of failure.** It is the only way most guides reach 100%. A bug in the button silently breaks badges, durable records, path progress, and the CTA. Decision 7 is the early warning, but someone must watch it.

**Mixed-content paths will produce counterintuitive progress.** A 4-milestone path where 2 interactive milestones show 60% and 80% and 2 prose milestones show 0% reports 35%. This is correct and feels wrong. Expect confusion from readers who feel further along than the number says.

## Follow-on items

- **Flag prose-only guides in the stamped metadata.** The bimodal distribution (interactive guides produce smooth curves, prose guides produce 0% or 100%) makes aggregate metrics useless. The fix is to split them into two distributions. `completableBlockCount === 0` already serves this purpose in the stamp. Surface it to the analytics layer.
- **UX follow-up on button discoverability.** Make sure that the reader knows the button exists, especially on long guides. Not critical path.
- **Specify reset behavior end to end.** State which storage systems (local, App Platform, RudderStack) change on reset. State what a completed-then-reset path looks like to a manager who reads App Platform.

Jayclifford345 (approved):

PR #1730, opened 2026-08-28
test: guard CLI Docker compile closure

### LINKED ISSUE #1603: Dockerfile.cli COPY list can silently miss the CLI's compile closure
## What happened

`Dockerfile.cli` copies a hand-picked subset of `src/` into the build stage:

```dockerfile
COPY src/cli ./src/cli
COPY src/types ./src/types
COPY src/validation ./src/validation
```

When a file inside that subset gains an import reaching outside it, the CLI image's `tsc -p tsconfig.cli.json` fails with a bare `TS2307`, while **every local check passes** — `npm run check`, and even `npm run build:cli`, because the file is simply present outside Docker. The failure only appears minutes into the `CLI publish` workflow's Docker build.

This bit us on #1597: `src/types/json-guide.types.ts` gained

```ts
import type { SupportedDatasourceType } from '../constants/datasource-types';
```

and `src/constants` was not in the COPY list. Fixed for that instance by adding the directory, but nothing stops the next one.

Note the import was **type-only** — erased at runtime, but still a compile input. `scanNodeEnvReachability` in `src/validation/import-graph.ts` deliberately skips type-only edges, so it cannot catch this class.

## Suggested fix

A validation test that walks the compile closure from `tsconfig.cli.json`'s `include` (following type-only edges) and fails when it reaches a `src/` directory with no matching `COPY` line.

I wrote one and it works — with the COPY line removed it reports:

```
Dockerfile.cli does not copy every src/ directory the CLI compile needs, so
`npm run build:cli` will fail inside the image with TS2307 even though it
passes locally:
  src/constants — reached via src/types/json-guide.types.ts → src/constants/datasource-types.ts
```

**It is not shippable as written.** It causes an intermittent jest worker `SIGSEGV` in roughly half of full `npm run test:ci` runs — confirmed by three consecutive clean runs with the file removed. Unproven cause; the likely candidate is that it re-walks the closure per test case, parsing a few hundred files through the TypeScript compiler in a worker alongside 400+ other suites. Memoizing the closure, or replacing the full walk with a much cheaper check (e.g. only the handful of `src/types/*.ts` files the CLI tsconfig names explicitly), is probably enough.

The draft is parked locally rather than committed. Anyone picking this up, ping me and I'll paste it — or just rewrite it, it's ~150 lines and the cheap version is likely shorter.

## Alternatives

- Have `Dockerfile.cli` copy all of `src/` and let `tsconfig.cli.json` scope the compile. Loses the tight build context and the npm-cache-invalidation property the current COPY list is there to preserve.
- Keep `src/types` free of cross-directory imports by convention. No enforcement, so it degrades the same silent way.

loglapa:
## Summary

- derive the CLI's TypeScript compile closure from `tsconfig.cli.json`, including type-only imports
- guard every required source against the builder-stage `COPY` destinations and `.dockerignore`
- require both CLI image workflows' `push` and `pull_request` path filters to cover the same closure
- report the shortest discovered import chain when a required source is missing

The repository-level closure is memoized once for the suite. The helper itself remains test-only and is excluded from the emitted CLI build because it uses the development-only TypeScript compiler API.

## Why

The CLI Docker build currently maintains its source inputs in several places. A new transitive import can therefore pass the normal TypeScript build while being absent from the Docker context or a publish workflow trigger, surfacing later as a Docker-only `TS2307` failure.

The new invariant derives the source of truth from the compiler rather than duplicating another source list. Synthetic tests cover type-only chains, empty roots, wrong `COPY` destinations, ordered `.dockerignore` negations, root-anchored ignore patterns, and both workflow events.

## Verification

- focused contract tests: 2 suites, 11 tests passed
- `tsc --noEmit`
- `tsc -p tsconfig.cli.json`
- ESLint and Prettier on all changed files
- full Jest suite twice: 503/503 suites, 8089 passed each run
- full Jest coverage gate: 503/503 suites, 8089 passed; new helper at 91.26% statements / 90.95% lines
- `git diff --check`

Closes #1603

### DIFF
--- a/src/validation/cli-build-contract.test.ts
+++ b/src/validation/cli-build-contract.test.ts
@@ -1,0 +1,31 @@
+/** @jest-environment node */
+
+import * as fs from 'fs';
+import * as path from 'path';
+
+import { getTypeScriptCompileClosure, validateCliBuildContract } from './cli-build-contract';
+import { REPO_ROOT } from './import-graph';
+
+const readRepositoryFile = (relativePath: string): string =>
+  fs.readFileSync(path.join(REPO_ROOT, relativePath), 'utf-8');
+
+const compileClosure = getTypeScriptCompileClosure(path.join(REPO_ROOT, 'tsconfig.cli.json'));
+
+describe('CLI compile closure build contract', () => {
+  it('keeps every required source file in Docker and publish workflow inputs', () => {
+    const errors = validateCliBuildContract(compileClosure, {
+      dockerfile: readRepositoryFile('Dockerfile.cli'),
+      dockerignore: readRepositoryFile('.dockerignore'),
+      workflows: {
+        '.github/workflows/cli-publish.yml': readRepositoryFile('.github/workflows/cli-publish.yml'),
+        '.github/workflows/e2e-runner-publish.yml': readRepositoryFile('.github/workflows/e2e-runner-publish.yml'),
+      },
+    });
+
+    if (errors.length > 0) {
+      throw new Error(
+        `CLI compile closure build contract failed:\n\n${errors.map((error) => `  - ${error}`).join('\n')}`
+      );
+    }
+  });
+});
--- a/src/validation/cli-build-contract.ts
+++ b/src/validation/cli-build-contract.ts
@@ -1,0 +1,348 @@
+import * as fs from 'fs';
+import * as path from 'path';
+import * as ts from 'typescript';
+
+import { extractImportRecords, toPosixPath } from './import-graph';
+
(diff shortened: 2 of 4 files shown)


moxious (approved):
Thanks for adding the CLI compile-closure contract — it should catch missing Docker and workflow source coverage before those mistakes become Docker-only failures.

PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1730
Summary: guard CLI Docker compile closure
Verdict: Approve

The next step is for a maintainer to merge this PR.

PR #1731, opened 2026-08-28
fix(cli): classify every guide block side effect

### LINKED ISSUE #1706: classifyBlock has no totality guard, so an unclassified block type silently skips a cloud e2e chain
## What is wrong

`classifyBlock` in `src/cli/e2e/side-effects.ts:158-203` is a chain of `isXBlock()` checks that ends in a catch-all:

```ts
return reason('unknown', path, 'Unknown block type');
```

The chain covers 18 block types. It misses `callout` and `collapsible`, and it will miss `divider` when #1702 merges. Nothing detects the gap: TypeScript cannot check an if-chain for completeness, and `side-effects.test.ts` does not enumerate the block types.

I confirmed there is no guard. Removing the `isVideoBlock(block)` arm produces one unused-import warning and no failed test.

## Why it matters

`unknown` scores 2 in `LEVEL_SCORE`, above `possibly_mutating`. `isUnsafeSideEffectLevel` returns true for any level that is not `readonly`. So one unclassified block makes its whole guide unsafe.

When a cloud-tier chain holds an unsafe guide and no stack pool is configured, `src/cli/commands/e2e.ts:689-694` skips the **entire chain** with the status `skipped_unsafe_shared_stack`. That status carries `ExitCode.SUCCESS`, so the run stays green while the tests do not run.

A `callout` is an author label plus markdown text. It cannot mutate anything. `collapsible` is worse than a plain miss: it holds child blocks, but the chain returns `unknown` instead of recursing into them, which `section`, `conditional`, and `assistant` all do.

## The fix

Two parts.

**1.** Add the missing types to the readonly chain at `src/cli/e2e/side-effects.ts:159-167`:

- `isCalloutBlock` — a title and markdown content, no actions.
- `isDividerBlock` — no fields at all. This guard arrives with #1702, so rebase on it first.

Give `collapsible` its own arm that recurses, next to the `isSectionBlock` arm:

```ts
if (isCollapsibleBlock(block)) {
  return classifyBlocks(block.blocks, `${path}.blocks`);
}
```

**2.** Add a totality test to `src/cli/e2e/side-effects.test.ts`. Classify a one-block guide for every member of `VALID_BLOCK_TYPES` and assert that none returns the message `Unknown block type`. Types that are unclassifiable by design — `terminal`, `terminal-connect`, `challenge`, `snippet-ref` — already return their own messages, so they pass without an exclusion list.

`src/cli/__tests__/block-registry.test.ts:38` is the pattern to copy. It is why `BLOCK_SCHEMA_MAP` cannot be forgotten, and `classifyBlock` has no equivalent.

## How to test your change

1. Run `npx jest src/cli/e2e --coverage=false`. It must pass.
2. Remove the `isCalloutBlock` arm you added.
3. Run the test again. The totality test must now fail and name `callout`.
4. Restore the arm.

## One thing to know

Keep the catch-all `return reason('unknown', ...)`. Guide JSON reaches this code from outside, so the runtime needs a safe default. The test is what makes a forgotten type loud.

## Context

Found during review of #1702. That PR does not touch this file, and the gap predates it — `callout` and `collapsible` fall through on `main` today.

loglapa:
## Summary

- classify `callout` blocks as readonly
- recurse through `collapsible.blocks` instead of falling through to the unknown classifier
- add a `VALID_BLOCK_TYPES`-driven totality guard while retaining the safe runtime fallback for external JSON

## Why

An unclassified block becomes `unknown`, which marks the guide unsafe. On a shared cloud stack without a stack pool, that can skip the whole E2E chain with a success-coded result. `callout` and `collapsible` were both missing from the classifier even though they are valid guide block types.

The registry-driven test makes future omissions fail with the missing discriminator names. A separate nested fallback assertion verifies that recursive collapsibles preserve the precise child path for malformed or future external input.

## Verification

- `jest src/cli/e2e --coverage=false --runInBand`: 20 suites, 297 tests passed
- mutation check: removing the `callout` arm makes the totality test fail with `callout`
- full Jest suite: 501/501 suites, 8081 passed
- `tsc --noEmit`
- ESLint and Prettier on changed files
- `git diff --check`

Closes #1706

### DIFF
--- a/src/cli/e2e/side-effects.test.ts
+++ b/src/cli/e2e/side-effects.test.ts
@@ -1,5 +1,6 @@
 import { classifyGuideSideEffects, classifyGuideSideEffectsFromString } from './side-effects';
-import type { JsonGuide } from '../../types/json-guide.types';
+import { VALID_BLOCK_TYPES } from '../../types/json-guide.schema';
+import type { JsonBlock, JsonGuide } from '../../types/json-guide.types';
 
 function guide(blocks: JsonGuide['blocks']): JsonGuide {
   return { id: 'g', title: 'Guide', blocks };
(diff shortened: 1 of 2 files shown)


moxious (approved):
Thanks for tightening the guide side-effect classifier and adding a schema-driven totality guard.

PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1731
Summary: classify every guide block side effect
Verdict: Approve

This is ready to merge once the repository’s normal branch protections are satisfied.

PR #1732, opened 2026-08-28
test(custom-guides): verify valid sibling composites survive malformed fields

### LINKED ISSUE #1726: test(custom-guides): malformed-composite tests do not prove sibling composites survive
The malformed-composite tests in `pkg/plugin/custom_guide_repository_test.go` do not prove the property they exist to protect.

Each case supplies one valid sibling composite (`:764-780`), but the assertions (`:808-818`) check only the scalar fields and the absence of the malformed value. A decoder that discarded *every* composite after a single failure would pass these tests unchanged — which is precisely the whole-object laundering that the field-local decoding in #1719 was written to prevent.

Suggested fix: assert that `author` survives the malformed-`milestones` case, and that `milestones` survives the malformed-`author` case. Two assertions, and they turn the test from evidence of "the bad field is gone" into evidence of "only the bad field is gone".

Priyanshu-1729:
## What changed

Updated `TestCustomGuideHTTPClient_MalformedCompositeFieldDropsOnlyThatField`
to verify that:

- a valid `author` survives when `milestones` is malformed
- valid `milestones` survive when `author` is malformed

## Why

The existing assertions proved that the malformed composite was removed, but
did not prove that the valid sibling composite was retained.

These assertions verify that decoding drops only the malformed field rather
than all composite fields.

## Testing

- `go test ./pkg/plugin -run TestCustomGuideHTTPClient_MalformedCompositeFieldDropsOnlyThatField -count=1`
- `npm run lint:go`
- `npm run test:go`

Closes #1726

### DIFF
--- a/pkg/plugin/custom_guide_repository_test.go
+++ b/pkg/plugin/custom_guide_repository_test.go
@@ -758,8 +758,11 @@ func TestCustomGuideHTTPClient_NegativeStatsMemberReadsAsAbsent(t *testing.T) {
 // actually stored.
 func TestCustomGuideHTTPClient_MalformedCompositeFieldDropsOnlyThatField(t *testing.T) {
 	cases := map[string]struct {
-		manifest map[string]any
-		leaked   string
+		manifest       map[string]any
+		leaked         string
+		wantAuthorName string
+		wantAuthorTeam string
+		wantMilestone  string
 	}{
 		"malformed milestones": {
 			manifest: map[string]any{
@@ -768,7 +771,9 @@ func TestCustomGuideHTTPClient_MalformedCompositeFieldDropsOnlyThatField(t *test
 				"milestones": []any{"fe-alerting-01", 7},
 				"author":     map[string]any{"name": "Field engineering", "team": "field-eng"},
 			},
-			leaked: "fe-alerting-01",
+			leaked:         "fe-alerting-01",
+			wantAuthorName: "Field engineering",
+			wantAuthorTeam: "field-eng",
 		},
 		"malformed author": {
 			manifest: map[string]any{
@@ -777,7 +782,8 @@ func TestCustomGuideHTTPClient_MalformedCompositeFieldDropsOnlyThatField(t *test
 				"milestones": []any{"fe-alerting-01"},
 				"author":     map[string]any{"name": "Field engineering", "team": []string{"field-eng"}},
 			},
-			leaked: "Field engineering",
+			leaked:        "Field engineering",
+			wantMilestone: "fe-alerting-01",
 		},
 	}
 	for name, tc := range cases {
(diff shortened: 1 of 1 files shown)


moxious:
Thanks for adding tests that verify valid sibling composites survive malformed fields — this strengthens custom guide parsing coverage.

One thing to clear before we review: the Contributor License Agreement is not signed yet. Please follow the contribution requirements in the pinned [contributor guide](https://github.com/grafana/grafana-pathfinder-app/blob/243648de83872ab377c6fa8e3e9ee900e7489313/CONTRIBUTING.md).

moxious:
Thanks for adding malformed-field test coverage for custom-guide sibling composites — it helps protect valid composite content from being dropped during parsing.

One thing to clear before we review: the Contributor License Agreement is not signed yet. Please follow the [contributor guide](https://github.com/grafana/grafana-pathfinder-app/blob/f5549ef64a8e048eed0280c86c067cd90831f731/CONTRIBUTING.md) to complete it.

moxious (approved):
Thanks for closing the assertion gap in `TestCustomGuideHTTPClient_MalformedCompositeFieldDropsOnlyThatField` — pinning that the *valid* sibling composite survives is exactly what the existing test was missing, and it makes the per-field decode contract in `customGuideManifest.UnmarshalJSON` actually enforceable.

Next step: this is approved with one optional suggestion below. Nothing blocks merge; take the suggestion now if you'd like to, or a maintainer can land it as-is.

---

## Suggestions

1. [suggestion] **composite-drop-absence-unasserted — Malformed composite is asserted absent from the wire, never absent as a value** (low · testing-and-verification)
   The new assertions are opt-in per case, so each case pins only one direction of the contract. In "malformed author", wantAuthorName is empty, the `if tc.wantAuthorName != ""` block is skipped, and nothing asserts got.Manifest.Author is nil. The only remaining guard is strings.Contains(raw, "Field engineering"). Author is a *customGuideManifestAuthor tagged `json:"author,omitempty"`, and omitempty on a non-nil pointer still marshals, so a regression that kept a zero-valued husk would serialise as "author":{} and pass both the leak check and the skipped assertion. decodeManifestField leaves Author nil at head, but the test does not pin that.
   Suggested: Assert the complementary field directly per case - a wantAuthorAbsent / wantMilestonesAbsent flag, or replace the optional scalar expectations with one expected customGuideManifest compared whole - so both halves of the drop-only-that-field contract fail loudly.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1732
Summary: verify valid sibling composites survive malformed fields
Verdict: Approve with Minor

Priyanshu-1729:
Thanks for the review and for pointing out that assertion gap i appreciate the feedback and the approval 👍

PR #1733, opened 2026-08-28
fix(review): read concern routing from head

### LINKED ISSUE #1667: contract-evolution gate reads CONCERNS.md at base but routes from head, so any PR adding a routing row fails its own gate
`contract-evolution-gate.mjs` reads `docs/design/CONCERNS.md` at the **base** SHA, while `.cursor/skills/review/SKILL.md` §3 routes concerns from the **head** table.

The two disagree for any PR that adds a routing row: the new concern is activated from the head table, then the gate looks it up in the base table, does not find it, and exits 2 with

```
Concern <id> is not present in the routing table
```

So a PR that adds a routing row fails its own gate, no matter how well-formed the row is. Simulating the same run with the row already present in base resolves cleanly (`triggered: true`, prior semantic PRs found), which confirms the row is not the problem.

Pre-existing for every new routing row, but #1661 is the first PR to add one since the gate landed, so it is the first time it surfaces.

Fix is presumably to read the routing table from head (matching how the skill routes), or to union base and head so a newly added row resolves.

Found while reviewing #1661. Deliberately not worked around there.

Nikshay1:
## What does this PR do?

The contract-evolution gate was reading `docs/design/CONCERNS.md` from the base commit, even though concern routing is based on the PR head.

This changes the routing lookup to read from `head`, while keeping the history query anchored at `base`.

I also added a regression test for a concern routing row introduced by the PR. The test makes sure the new routing is picked up without counting the current PR commit as prior history.

## Linked issue(s)

Closes #1667

## Checklist

- [x] This PR addresses a **single concern**.
- [x] All commits are signed.
- [x] I’ve added tests for the change.
- [x] No user-facing text or docs were changed.

### DIFF
--- a/.cursor/skills/review/scripts/contract-evolution-gate.mjs
+++ b/.cursor/skills/review/scripts/contract-evolution-gate.mjs
@@ -151,7 +151,7 @@ export function computeGate({ base, head, concern, windowDays = DEFAULT_WINDOW_D
   const inStackOutput = git(['rev-list', head, `^${base}`], cwd);
   const inStackShas = inStackOutput ? inStackOutput.split('\n').filter(Boolean) : [];
 
-  const concerns = git(['show', `${base}:docs/design/CONCERNS.md`], cwd);
+  const concerns = git(['show', `${head}:docs/design/CONCERNS.md`], cwd);
   const paths = extractConcernPaths(concerns, concern);
   const baseTimestamp = Number(git(['show', '-s', '--format=%ct', base], cwd));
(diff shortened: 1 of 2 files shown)


moxious (approved):
Thanks for fixing concern routing so new PR-head entries are visible without counting the current branch as prior history.

PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1733
Summary: read concern routing from head
Verdict: Approve

All contributor-controlled checks passed. A maintainer can merge this once the fork-comment infrastructure failure is waived or resolved.

PR #1734, opened 2026-08-28
Enforce exhaustive TypeScript switches

### LINKED ISSUE #1524: Enable switch-exhaustiveness-check and extract a shared assertNever (44 of 49 switches unguarded)
## Summary

Enable `@typescript-eslint/switch-exhaustiveness-check` and extract a shared `assertNever` helper. This is the cheapest high-leverage item from the type-design investigation (epic #603): the rule is **config-only** — type-aware linting is already wired at `.config/eslint.config.mjs:15-19` and typescript-eslint 8.65.0 is already installed.

## Current state

- **49 production switches. 5 are exhaustiveness-checked (~10%).**
- No shared `assertNever` exists. The pattern is open-coded **three different ways**:
  - `src/components/docs-panel/utils/requires-grafana-ui.ts:100` — `const unhandled: never = block;`
  - `src/requirements-manager/step-state.ts:176`, `src/components/floating-panel/dodge-session.ts:83`, `src/docs-retrieval/content-fetcher/bundled.ts:184` — `const _exhaustive: never = …;`
  - `src/components/interactive-tutorial/section-state.ts:110` — `const _never: never = action;`

So the technique is understood and endorsed here — it just isn't extracted or enforced.

## What it would catch

The highest-value instance is `src/docs-retrieval/json-parser.ts:1071`, `extractDefaultValueFromBlock(block: JsonBlock)` — it handles **13 of 19** block types and ends in `default: return ''`. Missing: `assistant`, `collapsible`, `conditional`, `terminal-connect`, `challenge`, `code-block`, `snippet-ref`. A new block type silently yields empty assistant-customization content.

Also `json-parser.ts:277` (the main block→element dispatch — all 19 handled, but the `default:` means a *removed* case degrades to a runtime warning rather than a compile error).

Note the specific failure mode this addresses: `noImplicitReturns` plus an explicit return annotation already gives poor-man's exhaustiveness for switches with **no** `default`. The gap is switches that **do** have a `default`, where a new union member is silently swallowed.

## Proposed change

```ts
// src/lib/assert-never.ts
export function assertNever(x: never, ctx?: string): never {
  throw new Error(`Unhandled variant${ctx ? ` in ${ctx}` : ''}: ${JSON.stringify(x)}`);
}
```

Rule config with an agent-facing message, matching the Phase 5/6 precedent already in `eslint.config.mjs`:

```js
'@typescript-eslint/switch-exhaustiveness-check': ['error', {
  allowDefaultCaseForExhaustiveSwitch: false,
  considerDefaultExhaustiveForUnions: false,
}],
```

The remediation message should name the fix explicitly — *"add a case, or `default: assertNever(x)` from `src/lib/assert-never.ts`"* — since per epic #603 the error text is the teaching surface for agents.

## Sequencing note

Two switches will resist until a prerequisite lands: `src/interactive-engine/interactive.hook.ts:419` and `src/interactive-engine/auto-completion/action-matcher.ts:315` switch on `targetAction`, which is typed `string` at those sites. Exhaustiveness is unavailable there until the action-verb union is unified — tracked separately. Either land that first, or scope the rule by directory initially.

## Suggested PR shape

Per the ratchet model: land the helper plus the rule, and fix exactly the switches the rule flags. Nothing else.

## Related

Epic #603 — the constraint hierarchy in that epic ranks lint-with-remediation-message above prose and below compiler errors; this is a case where the lint rule is strictly cheaper than the equivalent type change.

loglapa:
## What does this PR do?

Enables `@typescript-eslint/switch-exhaustiveness-check` for production TypeScript with the options agreed in #1524. It explicitly classifies every existing union member, replaces open-coded `never` anchors with a small non-throwing `assertExhaustive` helper, and preserves existing fail-safe fallback behavior.

The lint configuration is pinned by a governance test and the convention is documented in `AGENTS.md`, so future union additions fail with ESLint's standard missing-case diagnostic.

## Linked issue(s)

Closes #1524

## Verification

- production lint with a cold cache
- negative lint demonstration by temporarily removing the `default` quiz case, which reported `Cases not matched: "default"`; the case was restored before commit
- TypeScript typecheck and Prettier checks
- focused tests: 15 suites, 337 tests
- governance tests: 4 suites, 170 tests
- frontend suite: 500 suites, 8,057 tests
- repository pre-merge gate stages; the final `mage lint` step was also run directly with the pinned Mage 1.17.2 and reported `0 issues`

## Checklist

- [x] The `npm run check` pre-merge stages pass locally; the environment lacked the `mage` executable, so its final pinned Go-lint target was run separately and passed.


moxious (commented):
Thanks for enforcing exhaustive TypeScript switches.

CI is failing on this one: `Build CLI, build image, smoke test` reported `Cannot find module '../lib/assert-exhaustive' or its corresponding type declarations.` The CLI image only copies `src/cli`, `src/types`, `src/validation`, and `src/lib/guide-stats`, so the new `src/lib/assert-exhaustive.ts` dependency is absent from that build.

Push a fix and CI will rerun; we will review once it is green.

loglapa:
Fixed the CLI build closure in `e8083c6` after rebasing onto current `main`:

- copy `src/lib/assert-exhaustive.ts` into the CLI Docker builder
- allow it through `.dockerignore`
- include it in both CLI and E2E runner workflow path filters

Validation:

- `cli-build-contract.test.ts` + unit tests: 11/11 pass
- `tsc -p tsconfig.cli.json`: pass
- workflow YAML Prettier check: pass

A local Docker build confirmed the helper is present in the build context, then stopped during `npm ci` because Docker Desktop ran out of space (`ENOSPC`) before compilation. The repository build-contract test that originally reproduced the missing-file failure now passes.

moxious (approved):

## Nits

1. [nit] **PR1734-N1 — New exhaustiveness-check test lives under a security-labeled describe block** (low · testing-and-verification)
   `src/validation/security-lint-config.test.ts`'s `describe('Security lint configuration', ...)` previously asserted only the F5 DOM-sink eslint selectors. This PR adds an assertion on `@typescript-eslint/switch-exhaustiveness-check` inside that same describe block, but that rule is a correctness/architecture rule, not a security control, so the suite's label no longer matches its full contents.
   Suggested: Move the new assertion to its own describe block (or a general eslint-config test) so the security-labeled suite stays scoped to F1/F5 rules.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1734
Summary: Enforce exhaustive TypeScript switches
Verdict: Approve with Minor

PR #1737, opened 2026-08-29
test(go): stabilize RawMessage contract type on Go 1.27

### LINKED ISSUE #1724: test(go): reflected contract golden fails under Go 1.27
The documented `npm run test:go` command fails under Go 1.27: `json.RawMessage` reflects as `jsontext.Value`, which changes the generated tag golden in the reflected contract inventory. The repository's own Go 1.26.5 toolchain passes, and the failure reproduces at heads that predate any current work, so this is pre-existing rather than introduced by a change.

It matters twice over: anyone already on 1.27 sees the documented test command fail today, and the failure will become everyone's the moment the toolchain moves.

Suggested fix: normalize `json.RawMessage` and `jsontext.Value` to the same wire descriptor in the reflected contract inventory, and add a regression test that pins the descriptor across toolchains so the next reflection change is caught rather than discovered.

Raised as a follow-up during review of #1719 and deferred there deliberately; it is not that PR's to fix.

BLCCoreStudio:
## Summary

Go 1.27 reflects `json.RawMessage` as `jsontext.Value`, which makes the contract golden test fail.

Handle `json.RawMessage` before the generic named-type case. Added cases for `json.RawMessage` and `[]json.RawMessage` to keep the descriptor consistent on Go 1.26.5 and Go 1.27.

## How to verify

1. Run the focused contract tests with Go 1.26.5 and Go 1.27.
2. Confirm the existing struct-tag golden stays unchanged.

## Breaking changes

None.

Fixes #1724

### DIFF
--- a/pkg/plugin/contract_fixtures_test.go
+++ b/pkg/plugin/contract_fixtures_test.go
@@ -352,6 +352,9 @@ func contractNestedStruct(typ reflect.Type) (reflect.Type, bool) {
 }
 
 func renderContractType(typ reflect.Type) string {
+	if typ == reflect.TypeOf(json.RawMessage{}) {
+		return "json.RawMessage"
+	}
 	if typ.Name() != "" {
 		return typ.String()
 	}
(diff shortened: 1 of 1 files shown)


moxious (approved):
Thanks for stabilizing the RawMessage contract descriptor across Go toolchains — this keeps the repository’s contract golden reliable as Go 1.27 lands.

PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1737
Summary: stabilize RawMessage contract type on Go 1.27
Verdict: Approve

Once this approval is published, a maintainer can take the normal merge decision.

PR #1738, opened 2026-08-29
fix(custom-guides): finalize decode warning budget

### LINKED ISSUE #1725: fix(custom-guides): decode-warning summary is lost when a drain stops at its entry budget
`ListPage` emits the decode-warning suppression summary only when the continue token is empty (`pkg/plugin/custom_guide_repository_client.go:297-301`), but `drainCustomGuides` also terminates at its aggregate entry budget with a non-empty token (`pkg/plugin/custom_guide_repository.go:215-228`). On that path the detail logs are emitted and the summary never is, so a namespace with widespread schema drift produces the capped detail lines and no statement of how much was suppressed.

A probe on that path reproduced ten detail logs and zero summary logs.

Suggested fix: finalize the warning budget on every terminal path of a drain, including aggregate-budget truncation, and cover that path through `drainCustomGuides` rather than only through `ListPage`.

Known and stated when the per-drain budget landed in #1719 — closing it means touching the drain, which was outside that PR's scope.

BLCCoreStudio:
## Summary

The decode-warning suppression summary was previously emitted from `ListPage`, which meant it only appeared when pagination reached the final page.

A catalogue drain can also stop early when it reaches the aggregate entry budget, or return because a later page fails. In those cases the drain had already accumulated suppressed decode warnings, but the summary was never emitted.

Move that finalization to `drainCustomGuides` so the summary runs once whenever the drain finishes, regardless of which terminal path is taken.

Add regression coverage for normal pagination, aggregate-budget truncation, and page errors using the real `customGuideHTTPClient`.

## How to verify

- `env GOTOOLCHAIN=go1.26.5 go test ./pkg/plugin -count=1`
- `env GOTOOLCHAIN=go1.26.5 npm run lint:go`
- `env GOTOOLCHAIN=go1.26.5 npm run test:go`
- `LC_ALL=C env GOTOOLCHAIN=go1.26.5 npm run check`

## Breaking changes

None.

Fixes #1725

### DIFF
--- a/pkg/plugin/custom_guide_repository.go
+++ b/pkg/plugin/custom_guide_repository.go
@@ -202,6 +202,10 @@ func (a *App) writeCustomGuideUnavailable(w http.ResponseWriter) {
 // drainCustomGuides drains the namespace LIST across pages — up to the
 // aggregate entry budget — and returns the shaped catalogue entries.
 func drainCustomGuides(ctx context.Context, namespace string, lister customGuideLister) ([]customGuideRepositoryEntry, int, error) {
+	if finalizer, ok := lister.(customGuideDrainFinalizer); ok {
+		defer finalizer.finalizeDrain(namespace)
+	}
+
 	entries := []customGuideRepositoryEntry{}
 	continueToken := ""
 	pages := 0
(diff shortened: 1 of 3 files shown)


moxious (approved):
Thanks for making the decode-warning suppression summary run on every catalogue-drain exit and for covering both entry-budget and page-error paths.

PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1738
Summary: finalize decode warning budget
Verdict: Approve

Next: after this approval is published, a maintainer can merge the PR.

PR #1741, opened 2026-08-29
fix(block-editor): remove unreachable metadata modal

Nikshay1:
## What this changes

- remove the unreachable `GuideMetadataForm`
- remove the unused `metadata` modal state and related `BlockEditorModals` wiring
- remove metadata-only test IDs and unused guide metadata styles
- update `useModalManager` tests to use the remaining active modal names

Guide title and ID handling continue through the existing editor flow.

## Verification

- `npx jest src/components/block-editor/hooks/useModalManager.test.ts --coverage=false`
- `npm run check`
- `git diff --check`

Closes #1677

PR #1742, opened 2026-08-30
chore(lint): enforce unreachable and vacuous code rules

moxious:
## Summary

Unreachable code has never had a mechanical detector in this repository. `@grafana/eslint-config` does not extend `eslint:recommended`, so `no-unreachable` and its siblings have never run here; TypeScript's `noUnusedLocals` reports unused *bindings* inside a dead branch but never the dead branch itself, and a `@ts-expect-error` silences even that. This adds the five ESLint rules that close the gap.

Four of the five had zero violations at introduction — this is a ratchet, not a cleanup. The fifth, `no-useless-return`, had five, all fixed here. **No behaviour change is intended anywhere in this diff.**

## What to look at

- [`eslint.config.mjs`](https://github.com/grafana/grafana-pathfinder-app/blob/7ad1917bbbfc24c2c0430da409bcdc733bd2ac0a/eslint.config.mjs) — the new rule block. `no-unreachable`, `no-unreachable-loop`, `no-constant-condition` and `no-dupe-else-if` are unscoped, so they cover every file ESLint already lints, not just `src/`. Note what that set does **not** include: `tests/` is linted by nothing today. The `files: ['./tests/**/*']` block in `.config/eslint.config.mjs:25` matches no files — the leading `./` appears to be why — and no other block declares TypeScript under `tests/`, so `npx eslint tests/open-close.spec.ts` reports "File ignored because no matching configuration was supplied". These rules inherit that blind spot. Fixing it is a config-block change with its own violation fallout, tracked in #1744 rather than smuggled in here.
- [`src/components/BubbleTour/BubbleTour.tsx`](https://github.com/grafana/grafana-pathfinder-app/blob/7ad1917bbbfc24c2c0430da409bcdc733bd2ac0a/src/components/BubbleTour/BubbleTour.tsx) — the one site where the autofix was **wrong**, and the bullet worth the most reviewer attention. `paint` has one path that returns a promise and two that fall through; deleting its trailing `return;` broke `noImplicitReturns` (TS7030), which `@grafana/tsconfig` already enables. `paint` now declares `Promise<void>` and returns explicitly on every path. This is behaviour-identical for the `paintQueue.current.then(paint)` chain that awaits it — `.then()` wraps a bare `undefined` into a resolved promise anyway, so the ordering invariant documented above that line is untouched.
- [`src/completion-records/completion-write-storage.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/7ad1917bbbfc24c2c0430da409bcdc733bd2ac0a/src/completion-records/completion-write-storage.ts) — three `catch { return; }` blocks at the end of void functions become `catch {}`. These are the deliberate "localStorage is unavailable, the volatile map is authoritative" swallows; no serialization, storage key, or envelope shape is touched.
- [`src/components/block-editor/hooks/useRecordingPersistence.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/7ad1917bbbfc24c2c0430da409bcdc733bd2ac0a/src/components/block-editor/hooks/useRecordingPersistence.ts) — the autofix exposed an effect whose body had already degenerated to a no-op: it loaded persisted state, branched on it, and did nothing on either path. The `load()` call is retained because `load()` emits a `logger.error` on corrupt storage, which is observable; the vacuous guard and the three comments describing the clearing it never did are removed. Whether the effect should exist at all is a behaviour question and is deliberately left alone.

## Why

Part of [#603 — mechanical constraint enforcement for agents](https://github.com/grafana/grafana-pathfinder-app/issues/603), which argues that a constraint living only in prose can drift or be missed, while one living in `eslint.config.mjs` is verified on every lint run and every agent interaction. That epic's own reliability hierarchy ranks lint rules above documentation and compiler errors above lint rules; unreachable code currently sits below all three, detected by nothing.

The immediate prompt: [#1741](https://github.com/grafana/grafana-pathfinder-app/pull/1741) removed an unreachable metadata modal from the block editor, found by a human reading the code. Nothing in the toolchain would have flagged it.

`no-useless-return` is the one judgement call in here, and reviewers may reasonably disagree with it. It is the only rule in the set that partially contradicts a constraint the repo already enforces: a trailing bare `return;` is *required* by `noImplicitReturns` in any function with a value-returning path, and `no-useless-return` calls it useless. It fired at one such site out of five. Dropping the rule is a one-line revert of the config block if that trade reads badly — the other four rules stand on their own.

## How to verify

The rule set is enforced by CI on this very diff, so the config's own effect needs no manual step. What is worth checking by hand:

1. Open the guide editor, start recording into a section, refresh the page mid-recording, and confirm the recording is restored — this exercises `useRecordingPersistence`'s restore path, which sits next to the effect that was trimmed.
2. Stop recording and confirm no console or Faro error appears — the trimmed effect still runs on the `isRecording` transition.
3. Start a bubble tour, advance through a step whose target element does not exist (so `showCenteredComment` runs) and one whose target does exist, then close the tour mid-step. Confirm no orphaned highlight or bubble survives the close — this is the ordering invariant `paint`'s promise chain protects.
4. Complete a guide and confirm the completion is recorded and survives a reload, with `localStorage` both available and (via DevTools) blocked.

## Breaking changes

None. No storage key, serialized shape, telemetry event, `data-test-` value, or public contract changes.

## Reversibility

Fully reversible, with one caveat. The four dead-branch rules revert with the config block alone. Backing out `no-useless-return` is no longer a one-line revert, because the new corrupt-storage test pins the retained `load()` call in `useRecordingPersistence` — that test would have to come out with it. No migration runs, so nothing needs manual recovery on revert.

## Test coverage added here

The tests in this PR were written during validation, not by hand alongside the config change, and they go wider than the config:

- `src/validation/dead-code-lint-config.test.ts` — a real-consumer guard that drives the ESLint API over this repo's own `eslint.config.mjs`. Five of its seven cases fail against the base config and all seven pass against this one, so it tests the thing this PR actually adds. Deliberately not the grep-the-config-string pattern used by the neighbouring `security-lint-config.test.ts`.
- `completion-write-storage.test.ts` and `useRecordingPersistence.test.ts` — behaviour tests for the five `no-useless-return` sites, which had no coverage at all. They pass identically against base and target source, which is the behaviour-equivalence claim made concrete. The corrupt-storage case specifically fails if the `load()` call is dropped, so the one deliberate deviation from the raw autofix is now guarded.

The second group is adjacent coverage for behaviour this PR does not change, which sits outside the usual one-ratchet-plus-forced-fixes shape. It is here on purpose and the call on whether to keep it belongs to review — say the word and it comes out.

## Out of scope

- `@typescript-eslint/no-unused-vars` — `@grafana/eslint-config` explicitly disables it, and re-enabling it surfaces ~76 violations (50 unused parameters, 20 unused `catch` bindings, 6 unused values) that TypeScript's `noUnusedLocals` structurally cannot see. Filed separately.
- Unused exports and orphaned modules — the largest bucket by far, but it needs an entrypoint config or a reachability ratchet built on `buildModuleGraph()` before any number from it is trustworthy. Filed separately.
- `eslint.config.mjs` has no `trigger_paths` entry in `docs/design/CONCERNS.md`, so changes to it route only through always-on concerns. Noted, not addressed here.

Refs #603


leslie-heinzen (approved):



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1742
Summary: enforce unreachable and vacuous code rules
Verdict: Approve

PR #1745, opened 2026-08-30
docs(external-api): correct the unknown-field contract to prune, not 422

### LINKED ISSUE #1676: EXTERNAL_API.md claims unknown fields get a 422; they are silently pruned with a 201
## Summary

`docs/developer/EXTERNAL_API.md:311-314` says:

> The CRD schema **is the validator**. Submit unknown fields and you'll get a `422 Unprocessable
> Entity` with a K8s `Status` envelope explaining which field is wrong.

**This is wrong.** Unknown fields are **pruned with a `201`/`200`**, not rejected. The only signal is
a `Warning:` response header, which `getBackendSrv()` does not surface to callers.

This is the most dangerous shape of doc drift: it tells an implementer the platform will catch their
mistake, and it will not. Anything that writes an undeclared key reports success and loses the data.

## Evidence

Real generated CRD (`grafana-pathfinder-backend/jsonnet/pathfinder-backend/kinds/v1alpha1/interactiveguide.pathfinderbackend.ext.grafana.app.json`)
applied to a throwaway k3d cluster, then `kubectl create --raw` against the collection path:

**Undeclared keys nested in `spec.manifest`:**

```
$ kubectl create --raw ".../interactiveguides" -f probe.json
Warning: unknown field "spec.manifest.language"
Warning: unknown field "spec.manifest.recommends"
Warning: unknown field "spec.manifest.startingLocation"
Warning: unknown field "spec.manifest.stats"
Warning: unknown field "spec.manifest.targeting"
Warning: unknown field "spec.manifest.testEnvironment"
$ kubectl get interactiveguide spike-stats -n stacks-1 -o jsonpath='{.spec.manifest}'
{ "description": "typed field", "repository": "app-platform", "type": "guide" }
```

201 created. Six keys silently gone.

**Undeclared key at the top level of `spec`:**

```
$ kubectl create --raw ".../interactiveguides" -f probe.json   # spec.bogusTopLevel: "kept?"
Warning: unknown field "spec.bogusTopLevel"
$ kubectl get interactiveguide spike-topunknown -n stacks-1 -o jsonpath='{.spec}' | jq 'keys'
["blocks","id","title"]
```

Also pruned. So the claim is wrong at both nesting levels, not just inside `spec.manifest`.

## What the CRD *does* reject with a 422

Worth stating positively in the same edit, because these are real:

| Shape | Result |
|---|---|
| `manifest` with no `type` | `422 spec.manifest.type: Required value` |
| `manifest.type: "tutorial"` | `422 Unsupported value … supported values: "guide", "path", "journey"` |
| `manifest.depends: ["bare-string"]` | `422 spec.manifest.depends[0] in body must be of type array` |
| PUT with a stale `metadata.resourceVersion` | `409 Conflict` |

A **declared** field with a wrong value is a hard 422. An **undeclared** field is a silent prune.
That is the distinction the doc should draw.

## Suggested fix

Replace the 422 paragraph with the prune/reject distinction above, and state explicitly that
`additionalFields` is the only durable home for a key `#Manifest` does not declare — which the doc
already says at `:331-338` but frames as a stylistic choice rather than a hard requirement.

## Provenance

Measured while designing block-editor manifest support (scout `pf-editor-manifest-s1`).
See also the sibling drift in `grafana-pathfinder-backend/AGENTS.md:191-193`.

MFA-G:
Fixes #1676.

## What

`docs/developer/EXTERNAL_API.md` told implementers that the CRD rejects unknown fields with a `422`:

> The CRD schema **is the validator**. Submit unknown fields and you'll get a `422 Unprocessable Entity` with a K8s `Status` envelope explaining which field is wrong.

That is the opposite of what happens. Undeclared fields are **pruned** on a successful `200`/`201`; the only signal is a `Warning:` response header, which `getBackendSrv()` does not surface to callers. `422` is reserved for **declared** fields carrying a bad value.

This is the dangerous shape of doc drift: it promises the platform will catch a mistake that it will not catch, so a tool that writes an undeclared key reports success and loses the data.

## The change

Two edits, both in `EXTERNAL_API.md`:

1. Replaced the 422 paragraph with the declared-vs-undeclared distinction — rejection for a bad value in a declared field (with the three concrete `manifest` cases from the issue), silent pruning for anything undeclared, at the top level of `spec` and inside `spec.manifest` alike. It closes by naming `additionalFields` as the only durable home for an undeclared manifest key, which the doc previously framed at `:331-338` as a stylistic choice rather than a requirement.
2. Sharpened the `422` row in the errors table so the summary table doesn't restate the old claim.

Both point at the pruning behaviour the same document already documents for block fields under "Block fields the CRD doesn't declare" — the two sections previously contradicted each other, and now the new text cross-links to it.

## Consistency with the rest of the repo

The prune contract the new text describes is the one the codebase already encodes:

- `scripts/upsert-learning-path.sh:171-220` warns per resource about pruning and sweeps extension keys into `additionalFields`.
- `src/validation/upsert-script-crd-fields.test.ts` asserts that undeclared block fields disappear after a successful write.

No non-doc file needed to change; the code was right and the doc was wrong.

The sibling drift in `grafana-pathfinder-backend/AGENTS.md` is tracked separately at grafana/grafana-pathfinder-backend#84 and is out of scope here.

## Validation

- `npx prettier --check docs/developer/EXTERNAL_API.md` — passes (the errors-table column widths in the diff are prettier's own reflow after the row text changed).
- `npx jest --coverage=false src/validation/` — 914 passed. One unrelated pre-existing failure in `import-graph.test.ts` (`isNodeBuiltin('node:test')`), which is a Node-version artifact of my local Node 22 vs. the repo's `>=24` floor and fails identically on an unmodified checkout.
- Docs-only change: no runtime code, no types, no behaviour.

## AI disclosure

Drafted with AI assistance (Claude), reviewed and verified by me against the repo sources cited above before opening.

### DIFF
--- a/docs/developer/EXTERNAL_API.md
+++ b/docs/developer/EXTERNAL_API.md
@@ -308,9 +308,25 @@ The wire format is the standard Kubernetes envelope:
 | `spec.blocks`              | yes      | Array of content blocks. The full schema is owned by the CUE definition in [grafana-pathfinder-backend/kinds/interactiveguide.cue](https://github.com/grafana/grafana-pathfinder-backend/blob/main/kinds/interactiveguide.cue) — that file is the source of truth. |
 | `spec.manifest`            | no       | Package metadata: grouping, sequencing, dependencies. Absent for content-only guides. See [Manifest](#manifest).                                                                                                                                                   |
 
-The CRD schema **is the validator**. Submit unknown fields and you'll
-get a `422 Unprocessable Entity` with a K8s `Status` envelope explaining
-which field is wrong.
+The CRD schema **is the validator**, and it treats two kinds of mistake
+very differently:
+
+- A **declared** field with a bad value is rejected with a `422` and a
+  K8s `Status` envelope naming the offending field. A `manifest` without
+  `type`, a `manifest.type` outside `guide`/`path`/`journey`, and a
+  `manifest.depends` entry that is a bare string instead of a clause
+  array all fail this way.
+- An **undeclared** field is **silently pruned**, not rejected. The write
+  returns `200`/`201`, and the key is gone on the next GET. The only
+  signal is a `Warning:` response header, which `getBackendSrv()` does not
+  surface to callers. This holds at the top level of `spec` and inside
+  `spec.manifest` alike, and it is the same pruning described for blocks
+  in [Block fields the CRD doesn't declare](#block-fields-the-crd-doesnt-declare).
+
+So a successful write is not evidence that everything you sent was
+persisted. For manifest keys `#Manifest` doesn't declare,
+`additionalFields` is the only durable home — not a stylistic
+preference, but the difference between keeping the value and losing it.
 
 ### Manifest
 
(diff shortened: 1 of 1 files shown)


moxious:
Hi @MFA-G  - we'd like to take up this PR, but we require first that the CLA is signed, and that all github commits are cryptographically signed as well.  Thank you for contributing and we look forward to considering this PR.

moxious:
Hi @MFA-G -- we are interested in accepting this contribution, but per repo rules mentioned above, we can't review this PR or get it merged in until all commits are signed and the CLA is signed as well.  If you have time to return to this in the future, please feel free to re-open this PR.

Issue #1456, opened 2026-07-29
pathfinder-cli validate: cross-check snippet-ref IDs against the snippets catalog


moxious:
**Summary**

`pathfinder-cli validate` is snippet-blind: a `snippet-ref` block passes `--strict` validation with any `snippetId` string. The interactive-tutorials CI validates snippet *bodies* (`build-snippets` exits non-zero on invalid bodies) and guide *schemas* (`validate --strict` per `content.json`), but nothing cross-checks that a guide's referenced `snippetId` exists in the snippets catalog. A typo'd ref — or a guide referencing a snippet that a later PR renames or deletes — ships cleanly through CI today.

**Consequence today**

Bounded, thanks to two runtime behaviors: the snippet inliner reports `unresolvedSnippetIds`, and `prepareGuideLaunch` fails safe (`requiresGrafanaUi(...) || unresolvedSnippetIds.length > 0`), so a dangling ref renders as a broken placeholder in the sidebar — never dead interactive buttons in full screen. It's a content bug, not a surface bug (see the scoping note on #1446). But it's a content bug that publish-time validation could catch for free.

**Proposed fix**

1. Add an optional `--snippets-dir <path>` (or `--snippets-catalog <index.json>`) flag to `pathfinder-cli validate`. When present: walk the guide's blocks (including nested containers), collect `snippet-ref` `snippetId`s, and fail validation for any ID absent from the catalog built from that directory.
2. Wire it into interactive-tutorials' `validate-json.yml`: the workflow already builds the catalog to a throwaway path in the "Validate snippet bodies" step — pass that to the per-guide `validate` invocations.

Runtime resolution failures (transient network, CDN drift between guide and snippet deploys) remain possible by construction and are covered by the runtime fail-safe; this issue only closes the authoring-time door.

Found while verifying review claims on #1446.

Nikshay1:
I’ll take this. I’ll add snippet catalog cross-checking to pathfinder-cli validate, including nested snippet-ref blocks, and cover it with regression tests.

moxious:
## 🧭 Implementation readiness

**Stakes:** A guide whose `snippet-ref` names a typo'd or deleted snippet still ships through the content repository CI. A learner then sees an inert placeholder where the snippet content belongs.

**Sanity check:** The in-repository half of this issue merged as PR #1750 and its tests pass at this base. The remaining work is a CI wiring change in `grafana/interactive-tutorials`, so nobody needs to write code here.

[...]

PR #1750, opened 2026-08-31
fix(validation): validate snippet references against catalog

Nikshay1:
## What does this PR do?

Adds optional snippet-reference validation to `pathfinder-cli validate` through:

`--snippets-catalog <index.json>`

When the option is supplied, the CLI validates the generated snippets catalog and checks every `snippet-ref` in a guide against its catalog keys. References are checked at the top level and when nested inside sections, assistants, and both conditional branches.

Unknown snippet IDs are reported as normal validation errors with their exact guide path, for example:

`blocks[0].whenTrue[0].snippetId`

The catalog is loaded once per `validate` invocation and the option is forwarded through file, stdin, bundled, package, and package-tree validation.

Without `--snippets-catalog`, validation behaves exactly as before, so existing callers remain snippet-blind unless they explicitly provide a catalog.

This only adds authoring-time validation. Runtime snippet resolution and its existing fail-safe behavior are unchanged.

The corresponding `interactive-tutorials` workflow change is being made separately because it lives in another repository. It will reuse the catalog already generated by `build-snippets` and pass it to strict guide validation when available.

## Linked issue(s)

Addresses #1456

## Verification

- `npm run test:ci -- --coverage=false src/validation/snippet-references.test.ts src/cli/__tests__/validate-snippet-catalog.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm run prettier-test`
- `npm run build:cli`
- `npm run check`
- `git diff --check`
- manually verified that a known catalog reference passes
- manually verified that a missing nested reference fails with its exact `snippetId` path
- manually verified that the same guide passes when `--snippets-catalog` is omitted

## Checklist

- [x] This PR addresses a **single concern**: validating authored `snippet-ref` IDs against a supplied snippets catalog.
- [x] All commits are signed.

### DIFF
--- a/docs/developer/CLI_TOOLS.md
+++ b/docs/developer/CLI_TOOLS.md
@@ -79,6 +79,7 @@ node dist/cli/cli/index.js validate [options] [files...]
 - `--format <format>`: Output format. Options are `text` (default) or `json`.
 - `--package <dir>`: Validate a single package directory (expects `content.json` and optionally `manifest.json`).
 - `--packages <dir>`: Validate a tree of package directories recursively.
+- `--snippets-catalog <file>`: Optional generated snippet `index.json` to check every `snippet-ref` ID against. The catalog must satisfy the snippet catalog schema; a missing or invalid catalog fails validation. This option applies to file, stdin, bundled, package, and package-tree validation. Omit it when no local catalog is available.
 - File arguments accept explicit paths to JSON guide files.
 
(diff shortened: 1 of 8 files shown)


moxious (approved):



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1750
Summary: validate snippet references against catalog
Verdict: Approve

PR #1754, opened 2026-08-31
Make experiment variant projections total

### LINKED ISSUE #1548: Make the highlighted-guide variant union total over its four projections
## What silently breaks

`HighlightedGuideConfig['variant']` is a three-member string union with four hand-maintained projections, and none is total by construction. Add a fourth experiment arm to the union, forget one projection, and `tsc` says nothing:

- Miss `VALID_VARIANTS` → `validateHighlightedGuideValue` rejects the whole payload and every user in the new arm falls back to `DEFAULT_HIGHLIGHTED_GUIDE_CONFIG` (`variant: 'excluded'`). No auto-open, no Featured-slot injection, no exposure event, no `experiments` property. The arm looks like it enrolled nobody. Since #1529 this at least logs `reason: 'unknown_variant'` once per page load, so it is diagnosable — but only if someone reads the console.
- Miss `rollUpVariant` → users are enrolled and behave normally, but the scalar RudderStack `variant` property reports `'excluded'` while the sibling `experiments` array reports the true arm. `experiments` is stripped from the Faro mirror (`src/lib/analytics.ts:270`), so Faro carries only the wrong scalar with no corrective sibling. An analyst slicing by `variant` attributes new-arm users to the excluded bucket. **This is the quietest failure — wrong-but-plausible data, zero warning.**
- Miss `TRACKED_EXPERIMENT_VARIANTS` → enrollment works, but no `pathfinder_feature_flag_evaluated` exposure event fires for the arm. The analyst sees downstream activity from an arm with no exposure, so the A/B denominator is missing entries.

Correcting the review that found this: `VALID_VARIANTS` is the most user-visible failure but the loudest; `rollUpVariant` is the most insidious. Rank by silence, not blast radius.

## Verified inventory

As of `ab3a4cfd` (head of #1529):

| #   | Projection                    | Location                               | Shape                                                                                                                 |
| --- | ----------------------------- | -------------------------------------- | --------------------------------------------------------------------------------------------------------------------- |
| 1   | The union                     | `src/utils/openfeature.ts:37`          | `variant: 'excluded' \| 'control' \| 'treatment'` on `ExperimentConfig`; `HighlightedGuideConfig extends` it at `:63` |
| 2   | `VALID_VARIANTS`              | `src/utils/openfeature.ts:565`         | `ReadonlySet<HighlightedGuideConfig['variant']>` — typed, not total. Gates `:582` and `:613`                          |
| 3   | `TRACKED_EXPERIMENT_VARIANTS` | `src/utils/openfeature-tracking.ts:44` | untyped `Set<string>` — would not even catch a typo. Gates `:117`                                                     |
| 4   | `rollUpVariant`               | `src/lib/analytics.ts:167`–`175`       | inline literals, no shared projection                                                                                 |

Three more sites test the _complement_ (`variant === 'excluded'`) rather than enumerating arms, so a new arm falls through to "enrolled" — safe for a new treatment arm, wrong for a holdback-style arm: `src/utils/openfeature.ts:642` (`getActiveExperiments`), `src/context-engine/context.service.ts:753` (Featured-slot injection), `src/utils/experiments/highlighted-guide-orchestrator.ts:114` (auto-open).

Docs restate the arm list at `docs/developer/FEATURE_FLAGS.md:136` and `docs/developer/EXPERIMENT_TESTING.md:11`, `:187`. The Go backend has no variant references. No existing test asserts totality.

**How `main` differs:** `VALID_VARIANTS` does not exist on `origin/main` — #1529 introduces it, so `main` has three projections, not four. On `main`, `validateHighlightedGuideValue` only checks `typeof record.variant === 'string'`, so an unknown arm sails through and is treated as enrolled. #1529 correctly trades "unknown arms silently enroll" for "unknown arms rejected and warned" — and in doing so creates the fourth list that must now stay total. That is precisely why it is worth ratcheting now. (#1529 also hoisted `VALID_DOC_TYPES` to module scope at `:567`, so line numbers from earlier analysis of `main` are stale.)

## Evidence: the gap is real

Baseline on `ab3a4cfd` is clean (`npm run typecheck` → exit 0).

**Probe A — drop a member from `VALID_VARIANTS`, leave the union intact:**

```
const VALID_VARIANTS: ReadonlySet<HighlightedGuideConfig['variant']> = new Set(['excluded', 'control']);
```

```
> tsc --noEmit
PROBE_A_EXIT=0
```

Compiles clean. The `ReadonlySet<...>` annotation does not enforce totality.

**Probe A2 — for contrast, a typo _is_ caught:**

```
const VALID_VARIANTS: ReadonlySet<HighlightedGuideConfig['variant']> = new Set(['excluded', 'control', 'treatmnet']);
```

```
src/utils/openfeature.ts:565:7 - error TS2322: Type 'Set<"control" | "excluded" | "treatmnet">' is not assignable to type 'ReadonlySet<"control" | "treatment" | "excluded">'.
Found 1 error in src/utils/openfeature.ts:565
```

So the annotation buys typo-safety, not totality — exactly the half-guarantee that makes the gap easy to miss in review.

**Probe B — add a fourth arm to the union, touch none of the four projections (the crux):**

```diff
 export interface ExperimentConfig {
-  variant: 'excluded' | 'control' | 'treatment';
+  variant: 'excluded' | 'control' | 'treatment' | 'holdback';
```

```
> tsc --noEmit
PROBE_B_EXIT=0
```

Compiles clean. Nothing in the type system connects the union to any of its projections.

## Proposed fix

Invert the direction so the runtime list is the source of truth, then apply the house `satisfies` pattern per projection. Not everything should be derived — two projections encode a _deliberate_ difference from the full arm list, and flattening them would be a behavior change:

1. **Derive the union** (`src/utils/openfeature.ts`). Add `const HIGHLIGHTED_GUIDE_VARIANTS = ['excluded', 'control', 'treatment'] as const;` and set `variant: (typeof HIGHLIGHTED_GUIDE_VARIANTS)[number]`. Safe: the derived type is the identical literal union, union member order is irrelevant to assignability, and both external consumers reference it indirectly as `ExperimentConfig['variant']` (`src/utils/openfeature.ts:197`, `src/lib/analytics.ts:167`).
2. **Genuinely derive `VALID_VARIANTS`**: `new Set(HIGHLIGHTED_GUIDE_VARIANTS)`. Same module, no new import. Omission becomes structurally impossible.
3. **Do not derive `TRACKED_EXPERIMENT_VARIANTS`** — it intentionally omits `'excluded'` so the exposure stream holds only real enrolments. Replace the `Set<string>` with a total lookup, `satisfies Record<ExperimentConfig['variant'], boolean>`, with an explicit `excluded: false`. A new arm then fails to compile until someone decides whether it is an exposure. Same shape as `FORM_COMPONENTS` in the block-type ratchet.
4. **Do not derive `rollUpVariant`** — it encodes a _precedence_ (treatment beats control beats excluded), which a flat list cannot express. Replace the `if` chain with a priority map `satisfies Record<ExperimentConfig['variant'], number>`. This keeps the reference type-only, so `src/lib/analytics.ts` needs no new value import.
5. Optionally centralise the three complement checks behind an `isEnrolledVariant()` helper, so a future holdback arm forces one decision instead of three.

Note on 4: a value import from `src/lib/analytics.ts` to `src/utils/openfeature.ts` would _not_ fail the cycle ratchet — verified empirically, the footprint stays `clusters=2 filesInCycles=14 largestCluster=9` because both files already sit in the allowlisted telemetry cluster. Type-only is still preferable, since it avoids deepening that cluster.

## Severity

**Nothing is broken today** — verified: the union has exactly three members and all four projections agree (`TRACKED_EXPERIMENT_VARIANTS`'s omission of `'excluded'` is intentional and documented at `src/utils/openfeature-tracking.ts:40`–`43`). This is a latent-defect / maintainability issue, not a live bug.

On likelihood: the three arms have been stable since the union was introduced, and the one experiment-churn event (#1215, retiring `pathfinder.experiment-variant` and `pathfinder.after-24h-experiment`) removed whole flags without touching the arm vocabulary. So there is no evidence of arm churn. Honest read: **unlikely to trigger, but expensive when it does** — a mis-projected arm produces a silently invalid experiment result, and #1529's own history shows how a variant-handling change can add a projection without anyone noticing there are now four. The fix is a handful of lines and mechanically prevents the whole class, which is the argument for doing it rather than the probability.

## References

- Epic #603 — mechanical constraint enforcement; this is squarely the totality theme.
- Found while reviewing #1529, which introduces projection 2.
- Prior art, same class, same house pattern: `86d54816` "fix(types): make block-type registries total over the schema union" — `satisfies Record<...>` in `src/types/json-guide.schema.ts` and `src/types/package.schema.ts`. Note `src/utils/openfeature.ts:176` already uses `as const satisfies Record<...>` for the flag registry, so the pattern is established in this very file.

loglapa:
## What does this PR do?

Makes the experiment arm vocabulary a single runtime tuple and derives both its TypeScript union and the banner flag registry from it. The two deliberately different projections remain explicit total records: exposure tracking classifies every arm, and analytics roll-up assigns every arm a precedence.

This preserves the current three-arm behavior while making a future arm fail type-checking until its exposure and roll-up policies are decided. The parser and tracking boundary also reject inherited object property names rather than treating them as variants.

## Linked issue(s)

Closes #1548

## Verification

- focused OpenFeature, exposure-tracking, and analytics tests: 83 tests
- regression coverage for the excluded roll-up and prototype-property variant case

## Checklist

### DIFF
--- a/.cursor/skills/create-experiment/SKILL.md
+++ b/.cursor/skills/create-experiment/SKILL.md
@@ -37,7 +37,7 @@ Add to `pathfinderFeatureFlags` in `src/utils/openfeature.ts`. This is the one p
 ```typescript
 'pathfinder.<experiment-name>-experiment': {
   valueType: 'object',
-  values: [{ variant: 'excluded' }, { variant: 'control' }, { variant: 'treatment' }],
+  values: EXPERIMENT_VARIANTS.map((variant) => ({ variant })),
   defaultValue: { variant: 'excluded' },
   trackingKey: '<experiment_name>_experiment',
 },
@@ -49,6 +49,8 @@ Add to `pathfinderFeatureFlags` in `src/utils/openfeature.ts`. This is the one p
 - Keep `defaultValue` an inline literal so the registry does not import your experiment module — that import direction causes a cycle (see step 3).
 
 Three arms, always: `excluded` (not in the experiment), `control` (in it, current behaviour), `treatment` (in it, new behaviour). `excluded` and `control` must render identically; the difference is only whether the user is counted.
+
+`EXPERIMENT_VARIANTS` in `src/utils/openfeature.ts` is the runtime vocabulary and derives the TypeScript union. Adding an arm also requires an explicit policy in both total projections: `EXPERIMENT_VARIANT_EMITS_EXPOSURE` in `src/utils/openfeature-tracking.ts` and `EXPERIMENT_VARIANT_PRECEDENCE` in `src/lib/analytics.ts`. The compiler should fail until both decisions are made.
 
 ### 2. Create the experiment module
 
--- a/.cursor/skills/create-experiment/SKILL.md
+++ b/.cursor/skills/create-experiment/SKILL.md
@@ -52,6 +52,8 @@ Three arms, always: `excluded` (not in the experiment), `control` (in it, curren
 
 `EXPERIMENT_VARIANTS` in `src/utils/openfeature.ts` is the runtime vocabulary and derives the TypeScript union. Adding an arm also requires an explicit policy in both total projections: `EXPERIMENT_VARIANT_EMITS_EXPOSURE` in `src/utils/openfeature-tracking.ts` and `EXPERIMENT_VARIANT_PRECEDENCE` in `src/lib/analytics.ts`. The compiler should fail until both decisions are made.
 
+The compiler does not cover the runtime arm gates in `src/utils/experiments/highlighted-guide-orchestrator.ts` and `src/context-engine/context.service.ts`; review both explicitly when adding an arm.
+
 ### 2. Create the experiment module
 
 Everything else goes in `src/utils/experiments/<experiment-name>.ts` so retirement is a directory delete.
--- a/src/lib/analytics.test.ts
+++ b/src/lib/analytics.test.ts
@@ -185,6 +185,14 @@ describe('reportAppInteraction experiment enrichment', () => {
     expect(mockReportInteraction.mock.calls[0][1].variant).toBe('control');
   });
 
+  it('rolls variant up to excluded when every experiment is excluded', () => {
+    bindExperimentsProvider(() => [{ flag: HIGHLIGHTED, variant: 'excluded', pages: [], guideId: 'g' }]);
+
+    reportAppInteraction(UserInteraction.SummaryClick, {});
+
+    expect(mockReportInteraction.mock.calls[0][1].variant).toBe('excluded');
+  });
+
   it('omits variant/experiments when the user is enrolled in nothing', () => {
     bindExperimentsProvider(() => []);
 
--- a/src/lib/analytics.test.ts
+++ b/src/lib/analytics.test.ts
@@ -185,6 +185,19 @@ describe('reportAppInteraction experiment enrichment', () => {
     expect(mockReportInteraction.mock.calls[0][1].variant).toBe('control');
   });
 
+  it.each([
+    ['control before treatment', ['control', 'treatment']],
+    ['treatment before control', ['treatment', 'control']],
+  ] as const)('rolls multiple experiments up to treatment with %s', (_order, variants) => {
+    bindExperimentsProvider(() =>
+      variants.map((variant) => ({ flag: HIGHLIGHTED, variant, pages: [], guideId: 'g' }))
+    );
+
+    reportAppInteraction(UserInteraction.SummaryClick, {});
+
+    expect(mockReportInteraction.mock.calls[0][1].variant).toBe('treatment');
+  });
+
   it('rolls variant up to excluded when every experiment is excluded', () => {
     bindExperimentsProvider(() => [{ flag: HIGHLIGHTED, variant: 'excluded', pages: [], guideId: 'g' }]);
 
--- a/src/lib/analytics.test.ts
+++ b/src/lib/analytics.test.ts
@@ -189,9 +189,7 @@ describe('reportAppInteraction experiment enrichment', () => {
     ['control before treatment', ['control', 'treatment']],
     ['treatment before control', ['treatment', 'control']],
   ] as const)('rolls multiple experiments up to treatment with %s', (_order, variants) => {
-    bindExperimentsProvider(() =>
-      variants.map((variant) => ({ flag: HIGHLIGHTED, variant, pages: [], guideId: 'g' }))
-    );
+    bindExperimentsProvider(() => variants.map((variant) => ({ flag: HIGHLIGHTED, variant, pages: [], guideId: 'g' })));
 
     reportAppInteraction(UserInteraction.SummaryClick, {});
 
--- a/src/utils/openfeature-tracking.ts
+++ b/src/utils/openfeature-tracking.ts
@@ -35,20 +35,19 @@ function markReportedExposure(flagKey: string, variant: string): void {
 }
 
 /**
- * Variants for which we emit a FeatureFlagEvaluated exposure event.
+ * Experiment-arm membership and exposure policy at the analytics boundary.
  *
- * We intentionally skip 'excluded' (user isn't in the experiment) so the
- * event stream only contains real experiment exposures (control + treatment),
- * which is what downstream A/B analysis needs.
+ * Every valid arm is an own key; `false` keeps `excluded` out of the event
+ * stream while control and treatment emit the exposures used by A/B analysis.
  */
-const TRACKED_EXPERIMENT_VARIANTS = {
+const EXPERIMENT_VARIANT_EMITS_EXPOSURE = {
   excluded: false,
   control: true,
   treatment: true,
 } satisfies Record<ExperimentConfig['variant'], boolean>;
 
 function isExperimentVariant(variant: string): variant is ExperimentConfig['variant'] {
-  return Object.hasOwn(TRACKED_EXPERIMENT_VARIANTS, variant);
+  return Object.hasOwn(EXPERIMENT_VARIANT_EMITS_EXPOSURE, variant);
 }
 
 function extractExperimentVariant(value: JsonValue): ExperimentConfig['variant'] | null {
@@ -123,7 +122,7 @@ export function reportFeatureFlagExposure(flagKey: string, value: JsonValue): vo
   }
 
   const variant = extractExperimentVariant(value);
-  if (!variant || !TRACKED_EXPERIMENT_VARIANTS[variant]) {
+  if (!variant || !EXPERIMENT_VARIANT_EMITS_EXPOSURE[variant]) {
     return;
   }
 
--- a/src/utils/openfeature-tracking.ts
+++ b/src/utils/openfeature-tracking.ts
@@ -2,7 +2,7 @@ import type { Hook, HookContext, EvaluationDetails, JsonValue } from '@openfeatu
 
 import { reportAppInteraction, UserInteraction } from '../lib/analytics';
 import { StorageKeys } from '../lib/storage-keys';
-import { pathfinderFeatureFlags, type FeatureFlagName } from './openfeature';
+import { pathfinderFeatureFlags, type ExperimentConfig, type FeatureFlagName } from './openfeature';
 
 /**
  * In-memory fast path so the same flag never fires twice within one page load
@@ -41,18 +41,27 @@ function markReportedExposure(flagKey: string, variant: string): void {
  * event stream only contains real experiment exposures (control + treatment),
  * which is what downstream A/B analysis needs.
  */
-const TRACKED_EXPERIMENT_VARIANTS = new Set(['control', 'treatment']);
+const TRACKED_EXPERIMENT_VARIANTS = {
+  excluded: false,
+  control: true,
+  treatment: true,
+} satisfies Record<ExperimentConfig['variant'], boolean>;
 
-/**
- * Safely extract a string `variant` field from a JSON flag value.
- * Returns null if the value isn't an object or has no string variant.
- */
-function extractVariant(value: JsonValue): string | null {
-  if (value && typeof value === 'object' && !Array.isArray(value) && 'variant' in value) {
-    const raw = (value as { variant: unknown }).variant;
-    return typeof raw === 'string' ? raw : null;
+function isExperimentVariant(variant: string): variant is ExperimentConfig['variant'] {
+  return Object.hasOwn(TRACKED_EXPERIMENT_VARIANTS, variant);
+}
+
+function extractExperimentVariant(value: JsonValue): ExperimentConfig['variant'] | null {
+  if (!value || typeof value !== 'object' || Array.isArray(value) || !('variant' in value)) {
+    return null;
   }
-  return null;
+
+  const variant = (value as { variant: unknown }).variant;
+  if (typeof variant !== 'string' || !isExperimentVariant(variant)) {
+    return null;
+  }
+
+  return variant;
 }
 
 /**
@@ -113,8 +122,8 @@ export function reportFeatureFlagExposure(flagKey: string, value: JsonValue): vo
     return;
   }
 
-  const variant = extractVariant(value);
-  if (!variant || !TRACKED_EXPERIMENT_VARIANTS.has(variant)) {
+  const variant = extractExperimentVariant(value);
+  if (!variant || !TRACKED_EXPERIMENT_VARIANTS[variant]) {
     return;
   }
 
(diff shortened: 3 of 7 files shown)


src/utils/openfeature.ts
```diff
@@ -33,8 +33,10 @@ type FeatureFlag =
  * @param pages - Target pages where sidebar should auto-open (for treatment)
  * @param resetCache - When toggled true, clears session storage to allow sidebar to auto-open again
  */
+export const EXPERIMENT_VARIANTS = ['excluded', 'control', 'treatment'] as const;
```
moxious: N1: this const now sits between the `ExperimentConfig` docblock and the interface, so the `@param variant / pages / resetCache` docs bind to a string tuple and `ExperimentConfig` loses its hover documentation. Moving the const above the docblock fixes both.

src/utils/openfeature.ts
```diff
@@ -579,7 +581,7 @@ export const getHighlightedGuideConfig = (): HighlightedGuideConfig => {
   }
 };
 
-const VALID_VARIANTS: ReadonlySet<HighlightedGuideConfig['variant']> = new Set(['excluded', 'control', 'treatment']);
+const VALID_VARIANTS: ReadonlySet<string> = new Set(EXPERIMENT_VARIANTS);
```
moxious, line 584: N2: now that this is a `ReadonlySet<string>`, the `variant as ExperimentConfig['variant']` casts on lines 605 and 618 are dead — `has()` takes the narrowed `string` directly. Worth dropping them so the remaining cast (the `parseExperimentVariant` return, which is still load-bearing) is the only assertion left.

src/utils/openfeature-tracking.ts
```diff
@@ -41,18 +41,27 @@ function markReportedExposure(flagKey: string, variant: string): void {
  * event stream only contains real experiment exposures (control + treatment),
  * which is what downstream A/B analysis needs.
  */
-const TRACKED_EXPERIMENT_VARIANTS = new Set(['control', 'treatment']);
+const TRACKED_EXPERIMENT_VARIANTS = {
```
moxious: N3: the record is correct and the `satisfies` keeps it total, but it now answers two questions — which strings are arms (via `isExperimentVariant`/`Object.hasOwn`) and which arms emit exposure — while the docblock above still only describes the second and says it skips `excluded`, which is now an explicit `false` entry. A rename (`EXPERIMENT_VARIANT_EMITS_EXPOSURE`) plus a one-line comment would match the value.

moxious (changes requested):
Thanks for this — making the arm vocabulary one runtime tuple and deriving both the union and the flag registry from it is the right move, and keeping the two projections as separate total records rather than collapsing them into one lookup is the judgement call I'd have wanted.

The three things I most wanted to confirm all hold:

- **Exposure semantics are unchanged.** The new `!TRACKED_EXPERIMENT_VARIANTS[variant]` filter is truth-table identical to the old `Set.has`. No arm starts or stops emitting, and the dedup keys, marker keys, and emitted property set are untouched.
- **Totality is genuinely compiler-enforced,** not just asserted. Both projections use `satisfies Record<ExperimentConfig['variant'], _>` over the tuple-derived union, so a fourth arm fails to compile at each site. Your claim that a new arm can't be added without deciding its exposure and roll-up policy is accurate — and no `assertExhaustive` is needed here, since neither projection is a switch.
- **The prototype guard is correct and the new test is non-vacuous.** Removing `Object.hasOwn` makes `TRACKED_EXPERIMENT_VARIANTS['toString']` a truthy Function and the test fails, so it's really exercising the guard. The other boundary, `parseExperimentVariant`, is Set-based and inherently immune.

Telemetry is clean too: no new field, no new event, nothing new reaching Faro or RudderStack, and the filter-versus-activity-gate path is untouched.

Requesting changes rather than approving, because none of the four items below is technically blocking and that's exactly why they'd never get done otherwise. All four are mechanical and land in files this PR already opens.

- `.cursor/skills/create-experiment/SKILL.md` — still teaches the hand-listed `values: [{variant:'excluded'},…]` template this PR replaces, and nothing documents `EXPERIMENT_VARIANTS` or the two projections a new arm must extend. Someone following that skill after this merges writes the shape you just removed. Same-PR doc updates are the house rule.
- `openfeature-tracking.ts` — `TRACKED_EXPERIMENT_VARIANTS` now answers two questions (which strings are arms, via `isExperimentVariant`/`Object.hasOwn`, and which arms emit exposure) while its docblock describes only the second and says it skips `excluded`, which is now an explicit `false` entry. A rename to something like `EXPERIMENT_VARIANT_EMITS_EXPOSURE` plus a line about the membership role would match what the value does.
- `openfeature.ts` — now that `VALID_VARIANTS` is a `ReadonlySet<string>`, the `variant as ExperimentConfig['variant']` casts on lines 605 and 618 are dead. Dropping them leaves the `parseExperimentVariant` return cast as the only assertion, which is the one that's still load-bearing.
- `openfeature.ts` — the new const sits between the `ExperimentConfig` docblock and the interface, so the `@param variant / pages / resetCache` docs now bind to a string tuple and `ExperimentConfig` loses its hover documentation. Moving the const above the docblock fixes both.

---

Reviewed at `0aebe333e8e2b02800472f41839ce44d3156e6e5`. Typecheck and tests were deliberately withheld — community PR content is not executed locally, so CI is the authority. The claim most dependent on the compiler is that `values: EXPERIMENT_VARIANTS.map(...)` still satisfies `readonly JsonValue[]` under the registry's `as const satisfies`; CI is green on this head, so it does. The only failure is the known fork-permission `publish-report` signature, which is infrastructure, not yours.

loglapa:
Addressed all review notes in 6600a50: restored the ExperimentConfig docblock association, removed the dead set-membership casts, renamed/documented the exposure projection, and updated the experiment creation skill. Focused experiment/OpenFeature suites pass (83/83). Ready for re-review.

src/lib/analytics.test.ts
```diff
@@ -185,6 +185,14 @@ describe('reportAppInteraction experiment enrichment', () => {
     expect(mockReportInteraction.mock.calls[0][1].variant).toBe('control');
   });
 
+  it('rolls variant up to excluded when every experiment is excluded', () => {
```
moxious, line 199: F1: this pins the reduce's `excluded` seed, which the other two cases don't — good. The gap is one line further on: no test in the repo ever hands `rollUpVariant` more than one entry, so the precedence *ordering* is unverified. `satisfies Record<..., number>` checks the keys and the value type, never that treatment's number is larger. Swap the table to `control: 2` / `treatment: 1` and every test here still passes, because each single entry only has to beat the `excluded: 0` seed — while a user who is control in one experiment and treatment in the other now reports `variant: 'control'`. That user is reachable: `active-experiments.ts` pushes both entries from independent blocks. Worth adding `[{variant:'control'},{variant:'treatment'}]` and its reverse.

.cursor/skills/create-experiment/SKILL.md
```diff
@@ -50,6 +50,8 @@ Add to `pathfinderFeatureFlags` in `src/utils/openfeature.ts`. This is the one p
 
 Three arms, always: `excluded` (not in the experiment), `control` (in it, current behaviour), `treatment` (in it, new behaviour). `excluded` and `control` must render identically; the difference is only whether the user is counted.
 
+`EXPERIMENT_VARIANTS` in `src/utils/openfeature.ts` is the runtime vocabulary and derives the TypeScript union. Adding an arm also requires an explicit policy in both total projections: `EXPERIMENT_VARIANT_EMITS_EXPOSURE` in `src/utils/openfeature-tracking.ts` and `EXPERIMENT_VARIANT_PRECEDENCE` in `src/lib/analytics.ts`. The compiler should fail until both decisions are made.
```
moxious, line 53: F3: accurate for the two records, but an author will read "the compiler should fail" as end-to-end coverage. A fourth arm also lands at two sites the compiler is silent about — `highlighted-guide-orchestrator.ts:114` and `context.service.ts:754` both return early only on `variant === 'excluded'`, so an unknown arm falls through to auto-open and Featured-slot injection. Centralising those is correctly out of scope here (the readiness comment on #1548 rules it out), but one sentence naming them as decisions the compiler will not force would keep this paragraph from over-promising.

moxious (dismissed):
Thanks for the fix pass on #1548 — all four items from the last round are resolved at head. Inverting the arrangement so the union derives from the tuple, rather than sitting beside it under a `satisfies` check, is a better answer than the issue asked for: it makes drift structurally impossible instead of merely detected.

Approving. One follow-up and two suggestions below, none of which blocks merge — the follow-up is worth folding in whenever you next touch this area.

## Follow-ups

1. [follow_up] **F1 — Nothing checks the precedence numbers, and no test rolls up more than one arm** (medium · testing-and-verification)
   The if-chain that encoded treatment-beats-control in control flow is now a numeric table, and `satisfies Record<ExperimentConfig['variant'], number>` checks key totality and value type but never the ordering of the numbers. Every `bindExperimentsProvider` call in the repository sits in `src/lib/analytics.test.ts` (167, 181, 189, 197, 207, 222) and supplies zero or one entry, so swapping the values to `control: 2` / `treatment: 1` passes the whole suite while returning `control` for a user enrolled as control in one experiment and treatment in the other. That input is production-reachable: `active-experiments.ts:19-35` pushes both entries from independent, non-exclusive blocks, and `active-experiments.test.ts:129` already proves a two-entry array. `analytics.ts:297` strips `experiments` from the Faro mirror, so the rolled-up scalar is the only arm signal a mirrored event carries.
   Follow-up: Add one two-entry case asserting that control plus treatment rolls up to treatment in both orders.

## Suggestions

1. [suggestion] **F2 — The arm passthrough test still hand-enumerates the vocabulary** (low · testing-and-verification)
   `src/utils/openfeature.test.ts:884` is `it.each(['excluded', 'control', 'treatment'])`, the last hand-written arm list left in the repository and the one the compiler cannot see. A fourth arm forces both `satisfies` records to be updated but leaves this suite silently exercising only the original three through `getHighlightedGuideConfig`.
   Suggested: Pass `EXPERIMENT_VARIANTS` to `it.each` instead of the literal.

2. [suggestion] **F3 — The new skill paragraph reads as if the compiler covers a fourth arm end to end** (low · ai-subsystem)
   The added line in `.cursor/skills/create-experiment/SKILL.md` says the compiler should fail until both projection decisions are made. That is true of the two records, but a fourth arm also changes behaviour where the compiler is silent: `highlighted-guide-orchestrator.ts:114` and `context.service.ts:754` return early only on `variant === 'excluded'`, so an unknown arm falls through to the auto-open and Featured-slot injection paths. Centralising those checks is correctly out of scope per the decision of record on the issue; only the skill's claim is broader than what ships.
   Suggested: Add one sentence naming those two sites as decisions the compiler will not force.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1754
Summary: Make experiment variant projections total
Verdict: Approve with Minor

src/lib/analytics.test.ts
loglapa: Addressed in 59c72a3. The analytics test now passes both control→treatment and treatment→control entries through the real enrichment path and asserts that treatment wins in either order. I also folded in F2: the OpenFeature pass-through test now iterates the exported EXPERIMENT_VARIANTS vocabulary instead of hand-enumerating arms.

.cursor/skills/create-experiment/SKILL.md
loglapa: Addressed in 59c72a3. The create-experiment skill now explicitly names src/utils/experiments/highlighted-guide-orchestrator.ts and src/context-engine/context.service.ts as runtime arm gates that the compiler does not cover and that must be reviewed when adding an arm.

moxious:
Thanks for the fix pass — I verified all three items and they are properly closed, F1 especially. The two-entry test goes through the real enrichment path, and since `rollUpVariant` applies no per-flag dedupe, both entries genuinely fold even though they share the `HIGHLIGHTED` flag. I checked the mutation directly: swapping the precedence table to `control: 2` / `treatment: 1` now returns `control` and fails both rows, while every pre-existing single-entry case still passes. So that test is the only thing in the repo that catches it, which is exactly the hole I was worried about.

F2 and F3 both land too. Deriving the loop from the real `EXPERIMENT_VARIANTS` inside `isolateModules` rather than `it.each` is the right call given this file has no top-level import of `./openfeature` — I would have made the same choice.

CI is failing on one thing only:

```
[warn] src/lib/analytics.test.ts
[warn] Code style issues found in the above file. Run Prettier with --write to fix.
```

It is `src/lib/analytics.test.ts:192-194`. Collapsed onto one line that `bindExperimentsProvider` call measures exactly 120 characters, and `printWidth: 120` is an inclusive maximum, so Prettier joins it rather than keeping your three-line break. `Prettier` was green at the previous head and `CI Gate` needs it, which is why the PR shows as blocked.

Run `npm run prettier` and push — nothing else in the change needs to touch. I will approve once it is green.

loglapa:
Formatted the flagged `bindExperimentsProvider` call in d7f0628. Local Prettier check is green; no other code changed.

moxious (approved):

PR #1755, opened 2026-08-31
Ratchet eslint-disable comments without justifications

### LINKED ISSUE #1752: Ratchet undescribed eslint-disable comments (29 sites, 24 of them exhaustive-deps)
Part of #603 (mechanical constraint enforcement).

## The problem

The first progress comment on #603 records an unexpected side benefit of this epic:

> if eslint checks the right things, then eslint is a code smell for tech debt. Since architecture.test.js tests the right things, exclusions in that file are a code smell for tech debt.

That only holds while the exclusions say *why*. 29 lint suppressions in `src/` carry no `--` justification, so the map they form has no legend. 54 others do carry one, which is the pattern this issue makes mandatory.

24 of the 29 are `react-hooks/exhaustive-deps`. That is the suppression where the missing reason costs the most: with a justification it records a deliberate decision about which dependency is intentionally stale; without one it is indistinguishable from a warning somebody silenced to get green, and no later reader can tell which it was without re-deriving the whole hook.

## Evidence

29 undescribed suppressions. `react-hooks/exhaustive-deps` unless noted:

- `src/global-state/completion-store.ts:482`
- `src/requirements-manager/step-checker.hook.ts:836`, `:959`
- `src/utils/usePublishedGuides.ts:90`
- `src/utils/devtools/element-inspector.hook.ts:114`
- `src/docs-retrieval/learning-journey-helpers.ts:20` — `no-restricted-imports`
- `src/integrations/assistant-integration/AssistantSelectionPopover.tsx:88` — `react-hooks/set-state-in-effect`
- `src/integrations/assistant-integration/AssistantCustomizable.tsx:233` — `react-hooks/set-state-in-effect`
- `src/components/App/ContextPanel.tsx:31`
- `src/components/interactive-tutorial/interactive-step.tsx:475`, `:925`
- `src/components/interactive-tutorial/datasource-check-step.tsx:228`
- `src/components/interactive-tutorial/interactive-quiz.tsx:222`
- `src/components/interactive-tutorial/hooks/use-section-requirements.ts:192` — `react-hooks/set-state-in-effect`
- `src/components/interactive-tutorial/hooks/use-section-persistence.ts:141`
- `src/components/full-screen/FullScreenPanel.tsx:130`
- `src/components/content-renderer/content-renderer.tsx:374`, `:603`
- `src/components/block-editor/GitHubPRModal.tsx:227`
- `src/components/block-editor/LintBadge.tsx:68`
- `src/components/block-editor/hooks/useBlockPersistence.ts:225`
- `src/components/block-editor/hooks/useBackendGuides.ts:391`
- `src/components/floating-panel/FloatingPanelManager.tsx:183`
- `src/components/docs-panel/hooks/useSessionJoinUrlCheck.ts:42`
- `src/components/docs-panel/hooks/useTabRestoration.ts:80`
- `src/learning-paths/useDiscoverMore.ts:126`
- `src/learning-paths/learning-paths.hook.ts:137`, `:243`
- `src/package-engine/recommender-resolver.test.ts:387` — `@typescript-eslint/no-namespace`, test file

Two further grep hits are documentation examples inside a JSDoc block at `src/test-utils/interactive-section-harness.tsx:16` and `:20`, and one is prose mentioning the word at `src/requirements-manager/step-checker.hook.ts:395`. Those three are not suppressions and should not be counted or "fixed".

## Proposed work, in two parts

**Part 1 — the mechanism (maintainer, this issue).** A governance test under `src/validation/` that scans tracked sources for suppression comments lacking a `--` justification and compares the result against a checked-in baseline that may only shrink.

This is deliberately a governance test rather than an ESLint rule, because a rule cannot be baselined per site and would turn the whole list red at once. It reuses the repository's existing ratchet idiom directly: `assertRatchet` in `src/validation/import-graph.ts:647` already implements the exact semantics needed — reject new entries, and reject stale entries whose violation was fixed but which were left in the list. No new mechanism has to be invented.

**Negative-test it.** Per the correction comment on #603, a proposed constraint must demonstrate it fails when it should. The failure mode to test for specifically is the one that bit `verifyFieldsFromEffects`: a guard that silently passes because it found nothing to check.

**Part 2 — the drain (follow-up issues, filed once part 1 lands).** Write a real justification for each baselined suppression, one directory per pull request. Natural slices: `components/interactive-tutorial/` (5), `components/block-editor/` (3), `components/docs-panel/` + `full-screen/` + `floating-panel/` + `App/` (5), `learning-paths/` (3), `requirements-manager/` + `global-state/` + `utils/` + `docs-retrieval/` + `integrations/` (8 across the remaining engines).

## A warning for whoever does part 2

A plausible-sounding but wrong justification is worse than no justification, because it stops anyone else from looking. The reason must state what is actually true of that hook — which dependency is intentionally excluded and why re-running on it would be wrong — not a generic restatement that the dependencies are intentional. Where the honest answer turns out to be "this suppression is not justified", the correct outcome is to fix the hook and delete the suppression, not to invent a reason.

## Acceptance

- The baseline test is in place, negative-tested, and wired into the existing `test:governance` target.
- The baseline exactly matches the 29 sites above at install time, with the three false positives excluded.
- No new undescribed suppression can be added without the test failing.

loglapa:
## What does this PR do?

Adds a governance ratchet for `eslint-disable` comments that have no `--` justification. It scans tracked JavaScript and TypeScript sources, pins the current 29-site baseline exactly, rejects new undescribed suppressions, and requires stale entries to be removed as existing comments are explained or deleted.

The scanner explicitly distinguishes the three known documentation/prose examples from executable ESLint directives, and a credibility assertion prevents an accidentally empty scan from passing.

## Linked issue(s)

Closes #1752

## Verification

- focused governance test: 4 tests, including new-violation, false-positive, and non-empty-scan coverage
- wired the new test into `test:governance`

## Checklist

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -20,6 +20,8 @@ Never paste a raw control byte into a tracked file — write it as an escape (`\
 
 **Default to no comments.** Add one only when removing it would confuse a reader who can already read the surrounding code. The narrow band that earns one: counterintuitive-but-correct code, hidden invariants the type system can't express, external-bug workarounds (with an upstream link), and security or correctness warnings. If the comment won't fit on one short line, rename or restructure instead.
 
+Every `eslint-disable` directive requires a narrow explanation after `--`. Legacy directives without one are tracked by the shrink-only #1752 baseline and must not grow.
+
 **Trim on touch.** When editing a function, also trim bad-shape comments inside it and on adjacent declarations in the same file. Do not sweep whole files or grep the repo for cleanup — comment removal rides along on code changes, never as a standalone PR.
 
 The keep-list above is the whole of it. The eight bad shapes (QC8), with worked before/after examples, live in the `comment-hygiene` skill.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -20,7 +20,7 @@ Never paste a raw control byte into a tracked file — write it as an escape (`\
 
 **Default to no comments.** Add one only when removing it would confuse a reader who can already read the surrounding code. The narrow band that earns one: counterintuitive-but-correct code, hidden invariants the type system can't express, external-bug workarounds (with an upstream link), and security or correctness warnings. If the comment won't fit on one short line, rename or restructure instead.
 
-Every `eslint-disable` directive requires a narrow explanation after `--`. Legacy directives without one are tracked by the shrink-only #1752 baseline and must not grow.
+Every `eslint-disable` directive requires a narrow explanation after `--`. `src/validation/eslint-disable-justifications.test.ts` enforces this; legacy directives without one are tracked by `UNDESCRIBED_ESLINT_DISABLE_BASELINE` and must not grow.
 
 **Trim on touch.** When editing a function, also trim bad-shape comments inside it and on adjacent declarations in the same file. Do not sweep whole files or grep the repo for cleanup — comment removal rides along on code changes, never as a standalone PR.
 
--- a/src/validation/eslint-disable-justifications.test.ts
+++ b/src/validation/eslint-disable-justifications.test.ts
@@ -8,6 +8,7 @@ const REPO_ROOT = path.resolve(__dirname, '../..');
 const SOURCE_EXTENSIONS = new Set(['.js', '.jsx', '.ts', '.tsx']);
 const DIRECTIVE_PATTERN = /(?:\/\/|\/\*)\s*eslint-disable(?:-next-line|-line)?(?:\s|$)/;
 const JUSTIFICATION_SEPARATOR = ' -- ';
+const ESLINT_DISABLE_DIRECTIVE = ['eslint', 'disable'].join('-');
 
 const UNDESCRIBED_ESLINT_DISABLE_BASELINE = new Set([
   'src/learning-paths/learning-paths.hook.ts:137',
@@ -101,7 +102,7 @@ describe('eslint-disable justifications', () => {
   it('detects a newly introduced undescribed suppression', () => {
     const result = scanEslintDisableComments(
       'src/example.ts',
-      '// eslint-disable-next-line react-hooks/exhaustive-deps\nuseEffect(() => {}, []);'
+      `// ${ESLINT_DISABLE_DIRECTIVE}-next-line react-hooks/exhaustive-deps\nuseEffect(() => {}, []);`
     );
 
     expect(result).toEqual({ directiveCount: 1, undescribed: new Set(['src/example.ts:1']) });
@@ -111,7 +112,7 @@ describe('eslint-disable justifications', () => {
     const result = scanEslintDisableComments(
       'src/example.ts',
       [
-        '// eslint-disable-next-line react-hooks/exhaustive-deps -- dependency is intentionally stable',
+        `// ${ESLINT_DISABLE_DIRECTIVE}-next-line react-hooks/exhaustive-deps -- dependency is intentionally stable`,
         '// This prose says eslint-disable but is not a directive.',
       ].join('\n')
     );
@@ -122,11 +123,11 @@ describe('eslint-disable justifications', () => {
   it('excludes the known documentation and prose examples', () => {
     const docExample = scanEslintDisableComments(
       'src/test-utils/interactive-section-harness.tsx',
-      ' *   // eslint-disable-next-line @typescript-eslint/no-require-imports'
+      ` *   // ${ESLINT_DISABLE_DIRECTIVE}-next-line @typescript-eslint/no-require-imports`
     );
     const proseExample = scanEslintDisableComments(
       'src/requirements-manager/step-checker.hook.ts',
-      '// eslint-disable on its dep array). The callback closes over the value.'
+      `// ${ESLINT_DISABLE_DIRECTIVE} on its dep array). The callback closes over the value.`
     );
 
     expect(docExample).toEqual({ directiveCount: 0, undescribed: new Set() });
--- a/src/validation/eslint-disable-justifications.test.ts
+++ b/src/validation/eslint-disable-justifications.test.ts
@@ -6,93 +6,147 @@ import { assertRatchet } from './import-graph';
 
 const REPO_ROOT = path.resolve(__dirname, '../..');
 const SOURCE_EXTENSIONS = new Set(['.js', '.jsx', '.ts', '.tsx']);
-const DIRECTIVE_PATTERN = /(?:\/\/|\/\*)\s*eslint-disable(?:-next-line|-line)?(?:\s|$)/;
+const DIRECTIVE_PATTERN =
+  /(?:\/\/\s*(eslint-disable-(?:next-line|line))|\/\*\s*(eslint-disable(?:-(?:next-line|line))?))(?=\s|\*\/|$)/;
 const JUSTIFICATION_SEPARATOR = ' -- ';
 const ESLINT_DISABLE_DIRECTIVE = ['eslint', 'disable'].join('-');
 
+/**
+ * Shrink-only baseline for #1752. Keys are `path::rule#occurrence`, where the
+ * occurrence counts undescribed suppressions for that rule within the file.
+ * Line numbers stay diagnostic-only so unrelated edits cannot invalidate it.
+ */
 const UNDESCRIBED_ESLINT_DISABLE_BASELINE = new Set([
-  'src/learning-paths/learning-paths.hook.ts:137',
-  'src/learning-paths/learning-paths.hook.ts:243',
-  'src/learning-paths/useDiscoverMore.ts:126',
-  'src/components/docs-panel/hooks/useTabRestoration.ts:80',
-  'src/components/docs-panel/hooks/useSessionJoinUrlCheck.ts:42',
-  'src/components/content-renderer/content-renderer.tsx:374',
-  'src/components/content-renderer/content-renderer.tsx:603',
-  'src/package-engine/recommender-resolver.test.ts:387',
-  'src/components/interactive-tutorial/interactive-step.tsx:475',
-  'src/components/interactive-tutorial/interactive-step.tsx:925',
-  'src/components/interactive-tutorial/hooks/use-section-requirements.ts:192',
-  'src/components/interactive-tutorial/hooks/use-section-persistence.ts:141',
-  'src/components/interactive-tutorial/interactive-quiz.tsx:223',
-  'src/components/interactive-tutorial/datasource-check-step.tsx:228',
-  'src/components/App/ContextPanel.tsx:31',
-  'src/components/full-screen/FullScreenPanel.tsx:130',
-  'src/components/block-editor/hooks/useBackendGuides.ts:391',
-  'src/requirements-manager/step-checker.hook.ts:836',
-  'src/requirements-manager/step-checker.hook.ts:959',
-  'src/components/block-editor/hooks/useBlockPersistence.ts:225',
-  'src/components/block-editor/GitHubPRModal.tsx:227',
-  'src/components/block-editor/LintBadge.tsx:68',
-  'src/docs-retrieval/learning-journey-helpers.ts:20',
-  'src/global-state/completion-store.ts:482',
-  'src/components/floating-panel/FloatingPanelManager.tsx:183',
-  'src/integrations/assistant-integration/AssistantCustomizable.tsx:233',
-  'src/integrations/assistant-integration/AssistantSelectionPopover.tsx:88',
-  'src/utils/usePublishedGuides.ts:90',
-  'src/utils/devtools/element-inspector.hook.ts:114',
+  'src/learning-paths/learning-paths.hook.ts::react-hooks/exhaustive-deps#1',
+  'src/learning-paths/learning-paths.hook.ts::react-hooks/exhaustive-deps#2',
+  'src/learning-paths/useDiscoverMore.ts::react-hooks/exhaustive-deps#1',
+  'src/components/docs-panel/hooks/useTabRestoration.ts::react-hooks/exhaustive-deps#1',
+  'src/components/docs-panel/hooks/useSessionJoinUrlCheck.ts::react-hooks/exhaustive-deps#1',
+  'src/components/content-renderer/content-renderer.tsx::react-hooks/exhaustive-deps#1',
+  'src/components/content-renderer/content-renderer.tsx::react-hooks/exhaustive-deps#2',
+  'src/package-engine/recommender-resolver.test.ts::@typescript-eslint/no-namespace#1',
+  'src/components/interactive-tutorial/interactive-step.tsx::react-hooks/exhaustive-deps#1',
+  'src/components/interactive-tutorial/interactive-step.tsx::react-hooks/exhaustive-deps#2',
+  'src/components/interactive-tutorial/hooks/use-section-requirements.ts::react-hooks/set-state-in-effect#1',
+  'src/components/interactive-tutorial/hooks/use-section-persistence.ts::react-hooks/exhaustive-deps#1',
+  'src/components/interactive-tutorial/interactive-quiz.tsx::react-hooks/exhaustive-deps#1',
+  'src/components/interactive-tutorial/datasource-check-step.tsx::react-hooks/exhaustive-deps#1',
+  'src/components/App/ContextPanel.tsx::react-hooks/exhaustive-deps#1',
+  'src/components/full-screen/FullScreenPanel.tsx::react-hooks/exhaustive-deps#1',
+  'src/components/block-editor/hooks/useBackendGuides.ts::react-hooks/exhaustive-deps#1',
+  'src/requirements-manager/step-checker.hook.ts::react-hooks/exhaustive-deps#1',
+  'src/requirements-manager/step-checker.hook.ts::react-hooks/exhaustive-deps#2',
+  'src/components/block-editor/hooks/useBlockPersistence.ts::react-hooks/exhaustive-deps#1',
+  'src/components/block-editor/GitHubPRModal.tsx::react-hooks/exhaustive-deps#1',
+  'src/components/block-editor/LintBadge.tsx::react-hooks/exhaustive-deps#1',
+  'src/docs-retrieval/learning-journey-helpers.ts::no-restricted-imports#1',
+  'src/global-state/completion-store.ts::react-hooks/exhaustive-deps#1',
+  'src/components/floating-panel/FloatingPanelManager.tsx::react-hooks/exhaustive-deps#1',
+  'src/integrations/assistant-integration/AssistantCustomizable.tsx::react-hooks/set-state-in-effect#1',
+  'src/integrations/assistant-integration/AssistantSelectionPopover.tsx::react-hooks/set-state-in-effect#1',
+  'src/utils/usePublishedGuides.ts::react-hooks/exhaustive-deps#1',
+  'src/utils/devtools/element-inspector.hook.ts::react-hooks/exhaustive-deps#1',
 ]);
+
+type DirectiveForm = 'block' | 'block-next-line' | 'line' | 'line-next-line';
 
 interface ScanResult {
   directiveCount: number;
+  directiveForms: Record<DirectiveForm, number>;
   undescribed: Set<string>;
+  undescribedLocations: Map<string, string>;
 }
 
-function isDocumentedExample(relPath: string, line: string): boolean {
-  if (relPath === 'src/test-utils/interactive-section-harness.tsx' && line.trimStart().startsWith('*')) {
-    return true;
-  }
+function emptyDirectiveForms(): Record<DirectiveForm, number> {
+  return { block: 0, 'block-next-line': 0, line: 0, 'line-next-line': 0 };
+}
+
+function directiveForm(lineDirective: string, blockDirective: string | undefined): DirectiveForm {
+  const prefix = blockDirective ? 'block' : 'line';
+  return lineDirective.endsWith('-next-line') || blockDirective?.endsWith('-next-line')
+    ? `${prefix}-next-line`
+    : prefix;
+}
+
+function directiveRule(line: string, match: RegExpMatchArray): string {
+  const remainder = line.slice((match.index ?? 0) + match[0].length);
+  const beforeJustification = remainder.split(JUSTIFICATION_SEPARATOR, 1)[0] ?? '';
   return (
-    relPath === 'src/requirements-manager/step-checker.hook.ts' && line.includes('eslint-disable on its dep array')
+    beforeJustification
+      .replace(/\*\/.*$/, '')
+      .trim()
+      .replace(/\s*,\s*/g, ',') || '<all>'
   );
 }
 
 export function scanEslintDisableComments(relPath: string, content: string): ScanResult {
   let directiveCount = 0;
+  const directiveForms = emptyDirectiveForms();
   const undescribed = new Set<string>();
+  const undescribedLocations = new Map<string, string>();
+  const occurrences = new Map<string, number>();
 
   content.split(/\r?\n/).forEach((line, index) => {
-    if (!DIRECTIVE_PATTERN.test(line)) {
+    const match = line.match(DIRECTIVE_PATTERN);
+    if (!match || line.trimStart().startsWith('*')) {
       return;
     }
 
-    if (isDocumentedExample(relPath, line)) {
-      return;
-    }
-
-    const location = `${relPath}:${index + 1}`;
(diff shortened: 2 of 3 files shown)


src/validation/eslint-disable-justifications.test.ts
```diff
@@ -0,0 +1,148 @@
...
+  const combined: ScanResult = { directiveCount: 0, undescribed: new Set() };
+
+  for (const relPath of trackedSourceFiles()) {
+    const result = scanEslintDisableComments(relPath, fs.readFileSync(path.join(REPO_ROOT, relPath), 'utf-8'));
+    combined.directiveCount += result.directiveCount;
+    result.undescribed.forEach((location) => combined.undescribed.add(location));
+  }
+
+  return combined;
+}
```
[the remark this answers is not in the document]
loglapa: Fixed in 5663b80. The fixture directives are now assembled at runtime, so the tracked test file cannot match its own source scan. I reran the focused governance test with the file tracked and the full `npm run check`; both pass.

src/validation/eslint-disable-justifications.test.ts
```diff
@@ -0,0 +1,149 @@
...
+
+import { assertRatchet } from './import-graph';
+
+const REPO_ROOT = path.resolve(__dirname, '../..');
+const SOURCE_EXTENSIONS = new Set(['.js', '.jsx', '.ts', '.tsx']);
+const DIRECTIVE_PATTERN = /(?:\/\/|\/\*)\s*eslint-disable(?:-next-line|-line)?(?:\s|$)/;
+const JUSTIFICATION_SEPARATOR = ' -- ';
+const ESLINT_DISABLE_DIRECTIVE = ['eslint', 'disable'].join('-');
+
+const UNDESCRIBED_ESLINT_DISABLE_BASELINE = new Set([
```
moxious, line 20: F1 (blocking): keying on `path:line` makes this gate fire on merges that touch no suppression. Replaying the ratchet over main from 2026-05-01 to 2026-08-31, 59 of 509 first-parent commits would have shifted at least one baselined line — 96 pure-shift events against 5 genuine content changes. It has already happened once: #1734 moved the interactive-quiz.tsx site 222 -> 223, which is why #1752 lists `:222` and this pins `:223`. Rekey on `path::rule-name` plus a per-file count and keep the line number in the failure message only. Path alone would let a second undescribed suppression into an already-baselined file.

F5 (follow-up): no sibling baseline ships without a docblock. `ALLOWED_CYCLES` states the key format, the shrink-only direction, and names #1359 for the paydown; this one should name #1752 / #603.

src/validation/eslint-disable-justifications.test.ts
```diff
@@ -0,0 +1,149 @@
+import { execFileSync } from 'child_process';
+import * as fs from 'fs';
+import * as path from 'path';
+
+import { assertRatchet } from './import-graph';
+
+const REPO_ROOT = path.resolve(__dirname, '../..');
+const SOURCE_EXTENSIONS = new Set(['.js', '.jsx', '.ts', '.tsx']);
+const DIRECTIVE_PATTERN = /(?:\/\/|\/\*)\s*eslint-disable(?:-next-line|-line)?(?:\s|$)/;
```
moxious: F2 (follow-up): ESLint honours the bare range directives (`eslint-disable` / `eslint-enable`) only in block comments, so accepting `//` for the bare form matches prose rather than directives. The single bare `// eslint-disable` hit in all of src/ is prose at step-checker.hook.ts:395, which is exactly what `isDocumentedExample` then hardcodes. Allow `//` only with `-next-line` / `-line`, require `/*` for the bare form, and the allowlist entry disappears.

F4 (suggestion): the trailing `(?:\s|$)` also misses `/*eslint-disable*/` and `/* eslint-disable-next-line*/`, both of which ESLint accepts. Neither occurs in src/ today.

src/validation/eslint-disable-justifications.test.ts
```diff
@@ -0,0 +1,149 @@
...
+  'src/utils/usePublishedGuides.ts:90',
+  'src/utils/devtools/element-inspector.hook.ts:114',
+]);
+
+interface ScanResult {
+  directiveCount: number;
+  undescribed: Set<string>;
+}
+
+function isDocumentedExample(relPath: string, line: string): boolean {
```
moxious: F2 (follow-up): this couples the governance test to the wording of an unrelated file's comment — reword that sentence, or write similar prose elsewhere, and the ratchet fails on a PR that added no suppression. Fixing DIRECTIVE_PATTERN to require `/*` for the bare form lets the step-checker branch go entirely. The harness branch generalises too: a line whose trimmed start is `*` is inside an existing block comment in any file, so it need not name a path.

src/validation/eslint-disable-justifications.test.ts
```diff
@@ -0,0 +1,149 @@
...
+    if (!line.includes(JUSTIFICATION_SEPARATOR)) {
+      undescribed.add(location);
+    }
+  });
+
+  return { directiveCount, undescribed };
+}
+
+function trackedSourceFiles(): string[] {
+  return execFileSync('git', ['ls-files', '-z', 'src'], { cwd: REPO_ROOT })
```
moxious: F6 (nit): the sibling this mirrors, src/validation/control-bytes.test.ts, wraps this call in try/catch with an explanatory rethrow and filters through `fs.existsSync && fs.statSync().isFile()` before reading. Without the second guard, a file removed with `rm` rather than `git rm` is still listed by `git ls-files` and `npm run test:governance` dies with a raw ENOENT.

For the record, `maxBuffer` is not a problem here even though the sibling sets it: `git ls-files -z src` emits 62,516 bytes, about 6% of the 1 MB default.

src/validation/eslint-disable-justifications.test.ts
```diff
@@ -0,0 +1,149 @@
...
+    );
+
+    expect(docExample).toEqual({ directiveCount: 0, undescribed: new Set() });
+    expect(proseExample).toEqual({ directiveCount: 0, undescribed: new Set() });
+  });
+
+  it('matches the checked-in baseline and proves the scanner found real directives', () => {
+    const result = scanTrackedSources();
+
+    expect(result.directiveCount).toBeGreaterThan(UNDESCRIBED_ESLINT_DISABLE_BASELINE.size);
```
moxious, line 241: F3 (follow-up): the real post-exclusion count is 83, so this floor of 29 leaves a scanner covering a third of the directive forms passing. Concretely, all 29 undescribed sites come from the two `//` forms (21 `-next-line` + 8 `-line`) and none from either `/*` form — every one of the 7 block sites already carries a `--` reason. Drop `/*` from the pattern and the undescribed set stays byte-identical at 29 while `directiveCount` falls only to 76: nothing in this file fails. The unit tests above use `//` fixtures only. That is the silent-pass mode #1752's acceptance text asked to negative-test, and it lands on the form that suppresses a whole region (step-checker.hook.ts:979 through :992).

moxious (changes requested):
Thanks for taking #1752 on. Two things I want to credit before the problems, because they're the parts that are easy to get wrong and you got right: the scanner is **accurate** — I reimplemented it independently and the 29-entry baseline is exact, with zero missed token-bearing lines — and you included a credibility assertion at all, which is more than most ratchets ship with. The `directiveCount` floor is the right *idea*.

Requesting changes. One blocker, and five smaller items I'd like folded in here rather than deferred, since all six live in the single file this PR adds.

**Blocking — the baseline keys on line numbers, so the gate fires on unrelated merges**

`UNDESCRIBED_ESLINT_DISABLE_BASELINE` keys on `path:line` and `assertRatchet` enforces exact set equality, so any edit that shifts a line above a pinned suppression fails the gate on a PR that touched no suppression. I replayed the ratchet over main from 2026-05-01 to 2026-08-31: **59 of 509 first-parent commits — about one in nine — would have shifted at least one baselined line**, 96 pure-shift events against only 5 genuine content changes. The pins sit mid-file in hot files: `FloatingPanelManager.tsx` (13 shifts), `interactive-step.tsx` (12), `content-renderer.tsx` (11), `step-checker.hook.ts` (6), and one Faro telemetry commit moved baselined lines in 8 files at once.

This has already happened to you: #1734 merged the day you opened this, added 4 lines above the `interactive-quiz.tsx` suppression and moved it 222 → 223 without touching it — which is exactly why #1752 enumerates `:222` and this PR pins `:223`.

What makes it a blocker rather than an annoyance is the advice the innocent author gets. Pass one prints the new location under "Explain the suppression after `--`, or remove it. Do not grow the baseline." — telling them to justify a suppression they never wrote — and after they renumber, pass two throws again on the now-stale old keys. Two passes to clear a gate they didn't trip. Note also that this is a merge gate independently of the `test:governance` wiring: it matches `src/**/*.test.ts`, so `npm run test:ci` picks it up in CI's `test` job, which is in `ci-gate`'s `needs`.

No sibling ratchet embeds a line number — the four `assertRatchet` calls in `architecture.test.ts` key on `directionKey`, `barrelKey`, and `cycleKey`, all paths or path pairs.

Please rekey on `path::rule-name` plus a per-file occurrence count, and keep line numbers in the failure *message* rather than the pinned key. Path alone isn't enough: it would let a second undescribed suppression into an already-baselined file.

**Please also fold in**

- **The floor can't detect losing two-thirds of the directive forms.** Post-exclusion matches over tracked `src/` total 83: 66 `// eslint-disable-next-line`, 10 `// eslint-disable-line`, 5 `/* eslint-disable`, 2 `/* eslint-disable-next-line`. All 29 undescribed sites come from the two `//` forms; none from either block form, because all 7 block sites already carry a `--` reason. So **drop `/*` from `DIRECTIVE_PATTERN` and the undescribed set stays byte-identical at 29 while `directiveCount` only falls to 76 — nothing in this file fails.** The unit tests use `//` fixtures exclusively. That's the silent-pass mode #1752's acceptance text specifically asked to negative-test, and it lands on the form that suppresses a whole region (`step-checker.hook.ts:979`–`:992`). Block-form fixtures plus a per-form assertion close it.
- **Accepting `//` for the bare directive matches prose, not directives.** ESLint honours bare `eslint-disable` / `eslint-enable` only in block comments. The single bare `// eslint-disable` hit in all of `src/` is prose at `step-checker.hook.ts:395` — which is precisely what `isDocumentedExample` then hardcodes. Requiring `/*` for the bare form deletes the false-positive class *and* that allowlist entry. Worth doing together with the item above, since both are one change to `DIRECTIVE_PATTERN`.
- **The prose allowlist couples this test to an unrelated file's wording.** Reword that sentence in `step-checker.hook.ts`, or write similar prose anywhere else, and the ratchet fails on a PR that added no suppression. The fix above removes the need for the path hardcode entirely. The harness branch generalises too: a line whose trimmed start is `*` is inside a block comment in any file, so it needn't name a path.
- **The trailing `(?:\s|$)` misses `/*eslint-disable*/` and `/* eslint-disable-next-line*/`,** both of which ESLint accepts. Neither occurs in `src/` today, so this is latent — but it's the same regex you'll be editing anyway.
- **The baseline ships without the docblock every sibling carries.** `ALLOWED_CYCLES` states its key format, the shrink-only direction, and names #1359 for paydown; this one should name #1752 or #603. Relatedly, the ` -- ` convention appears nowhere in AGENTS.md or `.cursor/rules/` — a ratchet that enforces a rule the docs never state will keep surprising people.
- **`execFileSync` drops the guards its sibling has.** `control-bytes.test.ts` wraps the same git-plus-fs walk in try/catch with an explanatory rethrow, and filters through `existsSync && statSync().isFile()` before reading. Without the second guard, a file removed with `rm` rather than `git rm` is still listed by `git ls-files` and `npm run test:governance` dies with a raw ENOENT. For the record, `maxBuffer` is *not* a problem here even though the sibling sets it — `git ls-files -z src` emits 62,516 bytes, about 6% of the 1 MB default.

---

Reviewed at `5663b8055a9915c34d515b3f0cea8a816fda0e3b`. Test execution was deliberately withheld — community PR content is not executed locally, so CI is the authority on the suite. I confirmed the `package.json` change is one script line with no dependency added.

loglapa:
Addressed the review in cfb0723 and follow-up a5373e1: stable path/rule/occurrence keys, correct directive forms (including immediate */), generic block-doc exclusion, safe git enumeration, per-form fixture coverage, and AGENTS guidance. The repository-wide exact form counts were removed after Bugbot correctly noted they would reject legitimate justified suppressions. Focused governance tests pass.

src/validation/eslint-disable-justifications.test.ts
```diff
@@ -0,0 +1,248 @@
...
+      undescribedLocations.set(key, `${relPath}:${index + 1}`);
+    }
+  });
+
+  return { directiveCount, directiveForms, undescribed, undescribedLocations };
+}
+
+function trackedSourceFiles(): string[] {
+  try {
+    return execFileSync('git', ['ls-files', '-z', 'src'], { cwd: REPO_ROOT })
```
moxious: Blocking (F1). The `'src'` pathspec plus the four-extension filter make this cover a strict subset of what ESLint actually lints. `npm run lint` is `eslint --cache .`; neither the root ignore list nor `.config/eslint.config.mjs` excludes `.config/`, `scripts/`, `.github/scripts/`, `tests/`, or the root configs, so an undescribed suppression there is honoured by ESLint and invisible here. The sibling `control-bytes.test.ts:116` runs the same walk with no pathspec.

Drop the `'src'` argument and add `.mjs`/`.cjs` to `SOURCE_EXTENSIONS`. I checked that this is free: the 81 tracked non-src files matching those extensions contain zero directives, so the baseline stays at exactly 29 and `directiveCount` stays at 83. Whole-tree `git ls-files -z` is 75,643 bytes, ~7% of the 1 MB default, so `maxBuffer` still does not need touching.

src/validation/eslint-disable-justifications.test.ts
```diff
@@ -0,0 +1,248 @@
...
+    const match = line.match(DIRECTIVE_PATTERN);
+    if (!match || line.trimStart().startsWith('*')) {
+      return;
+    }
+
+    directiveCount++;
+    const lineDirective = match[1] ?? '';
+    const blockDirective = match[2];
+    directiveForms[directiveForm(lineDirective, blockDirective)]++;
+    if (!line.includes(JUSTIFICATION_SEPARATOR)) {
```
moxious: Blocking (F2). This tests the whole line, not the text after the directive — and `directiveRule` just above already computes the right suffix and throws it away. So:

```ts
const cmd = 'git log --oneline -- src/'; // eslint-disable-line no-restricted-syntax
```

is recorded as justified. The `' -- '` is a git pathspec separator in a string literal; no reason was written, the site never enters the baseline, and nothing reports it. This is the one predicate the ratchet exists to evaluate.

Compute the remainder once — `const after = line.slice((match.index ?? 0) + match[0].length)` — and use it for both `directiveRule` and this check. Latent today: I found no tracked line with the mismatch, so the 29-entry baseline does not move.

src/validation/eslint-disable-justifications.test.ts
```diff
@@ -0,0 +1,248 @@
...
+const SOURCE_EXTENSIONS = new Set(['.js', '.jsx', '.ts', '.tsx']);
+const DIRECTIVE_PATTERN =
+  /(?:\/\/\s*(eslint-disable-(?:next-line|line))|\/\*\s*(eslint-disable(?:-(?:next-line|line))?))(?=\s|\*\/|$)/;
+const JUSTIFICATION_SEPARATOR = ' -- ';
+const ESLINT_DISABLE_DIRECTIVE = ['eslint', 'disable'].join('-');
+
+/**
+ * Shrink-only baseline for #1752. Keys are `path::rule#occurrence`, where the
+ * occurrence counts undescribed suppressions for that rule within the file.
+ * Line numbers stay diagnostic-only so unrelated edits cannot invalidate it.
```
moxious, line 17: Nit (F4). The rekey is right and it resolves my blocker — both sides of `assertRatchet` resolve locations from the same current scan, so a pure line shift cancels out. Worth recording the one residual here, though: because the key is a per-file *count*, justifying `content-renderer.tsx::react-hooks/exhaustive-deps#1` while adding a fresh undescribed `exhaustive-deps` suppression in the same file leaves the key set unchanged and the gate green. That is the accepted cost of the shape I asked for, not a flaw in your implementation — one docblock line saying so, and that it closes as the baseline drains to zero, is all I'm after.

AGENTS.md
```diff
@@ -20,6 +20,8 @@ Never paste a raw control byte into a tracked file — write it as an escape (`\
 
 **Default to no comments.** Add one only when removing it would confuse a reader who can already read the surrounding code. The narrow band that earns one: counterintuitive-but-correct code, hidden invariants the type system can't express, external-bug workarounds (with an upstream link), and security or correctness warnings. If the comment won't fit on one short line, rename or restructure instead.
 
+Every `eslint-disable` directive requires a narrow explanation after `--`. Legacy directives without one are tracked by the shrink-only #1752 baseline and must not grow.
```
moxious: Nit (F3). Thanks for adding this — it closes the half of my round-1 point about the convention being undocumented. One thing to match the neighbours: the control-bytes rule twelve lines above ends by naming its enforcing test ("`src/validation/control-bytes.test.ts` enforces this over every tracked file"), and the tier-model and skills sections do the same. Name `src/validation/eslint-disable-justifications.test.ts` and `UNDESCRIBED_ESLINT_DISABLE_BASELINE` here, so an agent that trips the gate has a route from the rule to the guard and the baseline.

moxious (changes requested):
Thanks for the fix pass — **all seven items from my last round land at head**, and the rekey is a better answer than I asked for. Resolving the location for both sides of `assertRatchet` from the same current scan means a pure line shift cancels out algebraically rather than by convention, and a genuinely fixed entry surfaces as stale instead of vanishing. I re-implemented the scanner independently to check the rekey did not quietly move the baseline: identical 29-key set, `directiveCount` 83, forms `{line-next-line 66, line 10, block 5, block-next-line 2}` — matching my round-1 tally exactly.

Requesting changes on two items, both of which I verified are free to fix: neither moves the 29-entry baseline. Then this is ready. Two nits after them, one of which is a residual of a shape I prescribed rather than anything you got wrong.

## Merge contract

1. [blocking] **F1 — The ratchet only walks `src` and only four extensions, so most of the tree ESLint lints is ungated** (medium · testing-and-verification)
   `trackedSourceFiles()` passes a literal `src` pathspec to `git ls-files` (line 114) and filters to `.js/.jsx/.ts/.tsx` (line 118). `npm run lint` is `eslint --cache .`, and neither `eslint.config.mjs`'s ignore list nor `.config/eslint.config.mjs` excludes `.config/`, `scripts/`, `.github/scripts/`, `tests/`, or the root configs — so a new undescribed suppression in any of them, or in any `.mjs`/`.cjs` file, is honoured by ESLint and invisible to the gate. That contradicts #1752's acceptance criterion and the PR description's claim to scan "tracked JavaScript and TypeScript sources". The sibling guard `control-bytes.test.ts:116` runs the same walk with no pathspec, over every tracked file.
   Required: Drop the `'src'` argument from the `git ls-files` call and add `.mjs` and `.cjs` to `SOURCE_EXTENSIONS`. I verified this is free at head: the 81 tracked non-src files matching those extensions contain zero directives, so the baseline stays at exactly 29 keys and `directiveCount` stays at 83, and whole-tree `git ls-files -z` output is 75,643 bytes — about 7% of the 1 MB `execFileSync` default, so no `maxBuffer` change is needed.

2. [blocking] **F2 — The justification check reads the whole line, so a spaced double dash in the code launders a suppression** (low · correctness-and-reliability)
   Line 99 branches on `line.includes(JUSTIFICATION_SEPARATOR)` against the entire source line rather than the text after the directive — and `directiveRule` two lines earlier already computes exactly the right suffix and then discards it. So `const cmd = 'git log --oneline -- src/'; // eslint-disable-line no-restricted-syntax` is recorded as justified: the site never enters the baseline, no reason was ever written, and nothing reports it. This is the single predicate the whole ratchet exists to evaluate.
   Required: Compute the remainder once — `const after = line.slice((match.index ?? 0) + match[0].length)` — and use it for both `directiveRule` and the `includes` check. A fixture pairing a code-side ` -- ` with a trailing `-line` directive pins it. Verified latent at head: no tracked line exhibits the mismatch today, so the 29-entry baseline does not move.

## Nits

1. [nit] **F4 — The docblock does not record that occurrence-count keys miss a same-file same-rule swap** (low · testing-and-verification)
   Keys are path + rule + a per-file occurrence counter (lines 100-105), and both sides of `assertRatchet` resolve locations from the same current scan (232-239). So justifying `content-renderer.tsx::react-hooks/exhaustive-deps#1` while adding a fresh undescribed `exhaustive-deps` suppression in that same file leaves the key set unchanged and the gate green. This is the accepted cost of the key shape I asked for, not a defect in it — but the docblock at 14-18 states the key format and the shrink-only direction without stating this residual.
   Suggested: Add one docblock line saying the key is a count, so a same-file same-rule swap is not detected, and that the residual closes as the baseline drains to zero.

2. [nit] **F3 — The new AGENTS.md rule does not name the test that enforces it** (low · ai-subsystem)
   AGENTS.md line 23 states the ` -- ` requirement and the shrink-only baseline but names neither `src/validation/eslint-disable-justifications.test.ts` nor `UNDESCRIBED_ESLINT_DISABLE_BASELINE`. Twelve lines above, the control-bytes rule ends by naming its enforcing test, and the tier-model and skills sections do the same — so an agent that trips this gate has no route from the guidance to the guard or the baseline.
   Suggested: Extend the sentence in the control-bytes style, naming the enforcing test file and the baseline constant.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1755
Summary: Ratchet eslint-disable comments without justifications
Verdict: Request Changes
Results: 2 blockers, 2 non-blocking findings, 0 follow-ups

src/validation/eslint-disable-justifications.test.ts
loglapa: Computed `afterDirective` once and now use only that suffix for both rule parsing and the ` -- ` justification check. Added the exact regression case where `git log -- src/` appears before an undescribed directive.

src/validation/eslint-disable-justifications.test.ts
loglapa: Dropped the `src` pathspec, added `.cjs` and `.mjs`, and now scan every tracked JS/TS source file. The whole-tree ratchet remains at the same 29-entry baseline; focused governance tests pass (6/6).

AGENTS.md
loglapa: Updated AGENTS.md to name both `src/validation/eslint-disable-justifications.test.ts` and `UNDESCRIBED_ESLINT_DISABLE_BASELINE`.

src/validation/eslint-disable-justifications.test.ts
loglapa: Documented the accepted same-file count-key replacement residual in the baseline docblock, including that it disappears once the baseline drains to zero.

loglapa:
Addressed the full review round in 7547ae1: whole-tree JS/TS scan including cjs/mjs, suffix-only justification parsing with regression coverage, documented count-key residual, and direct AGENTS.md pointers. Focused suite passes 6/6; TypeScript, full ESLint, Prettier, and diff checks are clean.

moxious (approved):
Thanks for the third pass — both blockers land, and you fixed them the narrow way rather than the broad one. Dropping the `src` pathspec and adding `.mjs`/`.cjs` closes F1 exactly as scoped, and computing `afterDirective` once and threading it through both the `includes` check and `directiveRule` closes F2 while deleting a redundant slice. Both nits are addressed as well.

I re-derived the numbers independently rather than trusting the diff. A from-scratch reimplementation of the head scanner reproduces the 29-key baseline byte-for-byte at `directiveCount` 83, and the widened walk newly scans 27 non-`src` files that contain zero directives — so the baseline provably does not move. Running the pre-fix and post-fix guards over the same tree also yields identical key sets, which rules out a silent baseline shift hiding inside the F2 fix. Extension coverage is now a strict superset of what ESLint lints, and nothing in the repo's ignore list is tracked, so there is no gap in either direction.

One residual I looked at and deliberately dropped rather than filing: for the block form, `afterDirective` still runs past the closing `*/`, so trailing code containing a spaced double dash would read as a justification. It is unreachable — zero instances across 1352 files at head — and since the check is presence-only by design (`-- x` passes), closing it would not change the gate's strength. Not worth tracking.

Nice piece of work. Merging is yours.

---

PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1755
Summary: Ratchet eslint-disable comments without justifications
Verdict: Approve

Issue #1525, opened 2026-08-07
Guided blocks accept navigate/popout at validation, then throw at runtime


moxious:
## Summary

The interactive action verb (`targetAction`) is enumerated **nine times** across the codebase and no two enumerations agree. The hub declaration is a bare `string`, and a canonical union already exists — declared 14 lines above the field it should type, and imported by **nothing**.

## The evidence

`src/types/interactive.types.ts`:

```ts
export type InteractiveActionType =        // ← line 1. Zero references repo-wide.
  | 'button' | 'highlight' | 'formfill' | 'navigate' | 'hover'
  | 'sequence' | 'multistep' | 'guided' | 'popout';

export interface InteractiveElementData {
  refTarget: string;
  targetAction: string;                    // ← line 15. Should be the type above.
```

The nine declarations, with their member sets:

| Location | Notable divergence |
|---|---|
| `types/interactive.types.ts:1` `InteractiveActionType` | no `noop`; **unused** |
| `types/interactive.types.ts:15` | bare `string` |
| `types/json-guide.types.ts:272` / `json-guide.schema.ts:88` | no `sequence`/`multistep`/`guided` |
| `types/component-props.types.ts:29` | no `guided`/`multistep` |
| `types/interactive-actions.types.ts:26` `GuidedAction` | 5 members |
| `types/collaboration.types.ts:128` | no `noop`/`popout`/`guided`/`sequence` |
| `types/cross-tab.types.ts:197` `KNOWN_TARGET_ACTIONS` | no `noop`/`popout`/`sequence` |
| `lib/dom/selector-validator.ts:7` `ActionType` | no `noop`/`popout` |
| `constants/interactive-config.ts:236` | adds `QUIZ: 'quiz'`, in **no** union |

## Why it's a live bug, not just untidiness

**1. Three verbs are silently dropped.** `src/interactive-engine/interactive.hook.ts:174-187` — `dispatchInteractiveAction` is a 7-branch if/else with **no terminal `else`**. `sequence`, `multistep` and `noop` fall straight through. Because the chain sits inside a telemetry-wrapped async callback, the step then reports **success**, and `sequence-manager.ts` returns `'completed'` for a step that did nothing.

**2. The code has already surrendered in writing.** `src/interactive-engine/action-handlers/guided-handler.ts:647`:

```ts
const targetAction = action.targetAction as 'hover' | 'button' | 'highlight' | 'formfill';
[...]

PR #1757, opened 2026-08-31
Type interactive actions at the DOM ingress boundary

loglapa:
## What does this PR do?

Implements PR 1 of the three-part split recorded on #1525: a canonical `INTERACTIVE_ACTION_TYPES` tuple now derives the runtime union, and `InteractiveElementData.targetAction` uses that union instead of `string`.

Raw `data-targetaction` values are decoded at the DOM boundary. Missing, empty, `quiz`, and unknown values are rejected there without casts or sentinel values; trusted dispatch APIs are now exhaustive over the canonical vocabulary. String-typed authored and cross-tab inputs are narrowed before they enter the trusted runtime type.

Contract evolution: this **extends** the established interactive-engine contract by making `INTERACTIVE_ACTION_TYPES` the ingress authority for runtime interactive verbs. The corresponding `interactive-engine` anchor in `docs/design/CONCERN_DETAILS.md` is updated in this PR. Component-owned requirement pseudo-actions such as `section` remain deliberately string-widened through the named `InteractiveRequirementsData` type.

This deliberately leaves the other two planned changes separate: it does not delete the dead sequence path or narrow the guided schema. It also leaves `KNOWN_TARGET_ACTIONS` and `GUIDED_VERBS` unchanged, so the cross-tab receive gate is not widened.

## Linked issue(s)

Addresses #1525

## Verification

- focused interactive/DOM suites: 191 tests passed, 5 todo
- canonical-action decoder coverage plus regressions for missing, empty, unknown, and `quiz` values
- rejection coverage for requirement decoding, section show/do routing, multi-step aborts, and cross-tab replay
- architecture/control-byte governance tests: 38 tests

## Checklist


src/components/interactive-tutorial/interactive-section.tsx
```diff
@@ -626,11 +628,17 @@ export function InteractiveSection({
         }
       }
 
+      const { targetAction } = stepInfo;
+      if (targetAction === undefined || !isInteractiveActionType(targetAction)) {
+        logger.warn(`Unknown sequence action: ${targetAction ?? 'missing'}`);
+        return false;
+      }
```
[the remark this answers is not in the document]
loglapa: Fixed in 120bd02. Only canonical actions enter the typed dispatcher; missing or component-owned action strings now preserve the legacy successful fallthrough (including post-verification) instead of returning `action_error`. Focused section-runner tests and the full `npm run check` pass.

src/components/interactive-tutorial/interactive-multi-step.tsx
```diff
@@ -385,6 +387,12 @@ export const InteractiveMultiStep = forwardRef<{ executeStep: () => Promise<bool
         for (let i = 0; i < internalActions.length; i++) {
           const action = internalActions[i]!;
 
+          if (!isInteractiveActionType(action.targetAction)) {
+            logger.error(`Unknown multi-step action: ${action.targetAction}`);
```
moxious, line 394: F-1757-2 (follow-up): this early exit returns false without setting setFailedStepIndex(i) or setExecutionError, unlike the three sibling failure paths below it, so the learner sees a step that refuses to run with no explanation. On base an unrecognized verb was skipped and the multi-step still completed, so failing the whole step is new behaviour; if that is intended, give it the same UI signal the other failures get. The guard also runs before the isCancelledRef check, so a cancelled run reports a verb error instead of cancellation.

src/interactive-engine/interactive.hook.ts
```diff
@@ -474,6 +483,10 @@ export function useInteractiveElements(options: UseInteractiveElementsOptions =
                 sequenceResult = await interactiveSequence(elementData, isShowMode);
                 break;
 
+              case 'multistep':
+                logger.warn(`Unknown interactive action: ${targetAction}`);
```
moxious: F-1757-4 (nit): 'multistep' is now a member of INTERACTIVE_ACTION_TYPES, so "Unknown interactive action" contradicts the type this case was added to satisfy. This warn is the only signal an author gets when a multistep verb reaches the element path — a message like "multistep is executed by InteractiveMultiStep, not the element path" (the shape live-tab-executor.ts already uses for composites) would point debugging the right way.

src/interactive-engine/interactive.hook.ts
```diff
@@ -44,15 +45,10 @@ export interface CheckResult {
   targetHref?: string;
 }
 
-/**
- * This function is a guard to ensure that the interactive element data is valid.  It can encapsulte
- * new rules and checks as we go.
- * @param data - The interactive element data
- * @returns boolean - true if the interactive element data is valid, false otherwise
- */
+type InteractiveRequirementsData = Omit<InteractiveElementData, 'targetAction'> & { targetAction: string };
```
moxious: F-1757-3 (suggestion): this type is declared byte-identically in interactive-multi-step.tsx too, and neither copy is exported, so a tier 4 component restates a tier 2 parameter type. It is also load-bearing rather than incidental: interactive-section.tsx:717 and hooks/use-section-requirements.ts:116 still pass targetAction: 'section', which is not an INTERACTIVE_ACTION_TYPES member. Exporting one named type — named for the 'section' pseudo-action it admits — would stop the two copies drifting and tell the next reader this widening is deliberate.

src/integrations/cross-tab/live-tab-executor.ts
```diff
@@ -193,6 +199,10 @@ export function installLiveTabExecutor(
           `[Pathfinder] cross-tab executor: composite action "${action.targetAction}" carried no internalActions to replay`
         );
         break;
+      case 'sequence':
```
moxious, line 203: F-1757-5 (nit): with these cases the switch enumerates every InteractiveActionType, but the default below still only warns, while the hook's switch in this same PR gained assertExhaustive. AGENTS.md asks for assertExhaustive from src/lib/assert-exhaustive.ts in fail-safe defaults; without it this switch is no longer compile-checked against the union, in the one place that replays actions into another tab.

src/components/interactive-tutorial/interactive-section.tsx
```diff
@@ -627,15 +629,21 @@ export function InteractiveSection({
...
+            buttonType: 'do',
+          });
+          if (actionOutcome === 'error') {
+            logger.warn(`Sequence action did not complete for ${stepInfo.stepId}`);
+            return false;
+          }
+        } else {
+          // Preserve the legacy section-runner behavior for component-owned
+          // steps while their pause semantics are handled separately.
+          logger.warn(`Unknown interactive action: ${targetAction}`);
```
moxious, line 646: F-1757-6 (suggestion): on base this branch still called executeInteractiveAction, which emitted a DoItButtonClick Faro span via withFaroUserAction (critical: true) even when the switch matched nothing. Component-owned steps — the population this commit exists to preserve — now emit no span at all, so section runs containing code-block/terminal-connect/datasource-check steps will show a Do it count drop with nothing in the dashboard to explain it. Worth deciding explicitly whether a verb-less step counts as a Do it.

moxious (changes requested):
Thanks for this — closing the `data-targetaction` decode at the DOM boundary instead of casting downstream is the correct fix, and keeping it to PR 1 of the #1525 split (leaving the dead sequence path and the guided schema alone) made this reviewable in a way the combined change wouldn't have been.

I spent most of the review on the claims in your description, and they check out:

- The decode boundary genuinely rejects missing, empty, `unknown`, and `quiz` with no cast, assertion, or sentinel value, and both of its only two consumers handle `null`.
- `KNOWN_TARGET_ACTIONS` and `GUIDED_VERBS` are byte-unchanged, and the new `isInteractiveActionType` guard sits *behind* the wire gate, so the extra verbs are inert. The cross-tab receive gate is not widened, as you said.
- The hook switch is exhaustive with a fail-safe default, tier flow stays downward-only with no new cycle, and no authored content is stranded: HTML interactive parsing was already removed at `html-parser.ts:573`, so JSON guides are the only ingress and `JsonInteractiveAction` is a strict subset.
- Three regressions I went looking for aren't there — the multi-step abort can't strand the blocking overlay (the `finally` releases it), the section runner's unknown-verb path already returned `'ok'` at base, and all 26 `InteractiveElementData` sites plus every `executeInteractiveAction` caller are consistent, so the split boundary holds and this is independently revertible.

Requesting changes. One item is a user-visible behaviour change I don't think was intended, one is a test gap that got worse rather than better, and the rest are small enough that I'd rather sweep them here than file backlog.

**A step can now fail silently (please fix)**

`interactive-multi-step.tsx` — the new early exit returns `false` without setting `setFailedStepIndex(i)` or `setExecutionError`, unlike the three sibling failure paths right below it. The learner sees a step that refuses to run with no explanation. On base an unrecognized verb was skipped and the multi-step still completed, so failing the whole step is new behaviour; if that's deliberate, it needs the same UI signal the other failures get. Separately, the guard runs *before* the `isCancelledRef` check, so a cancelled run now reports a verb error instead of cancellation.

**Test coverage went backwards (please fix)**

Five new reject branches, one covered. Only `dom-utils.test.ts` gains behaviour tests; the `checkElementRequirements` null path, `interactive-section`'s do/show branches, the multi-step abort, and `live-tab-executor`'s `runAction` guard are all untested — and the one pre-existing unknown-verb test was deleted without a replacement. For a PR whose whole point is that ingress now rejects things, the rejections are the part that needs proving.

**Contract statement is missing (please add)**

`INTERACTIVE_ACTION_TYPES` becomes the ingress authority for interactive verbs, which trips the contract-evolution rule: the PR body needs to say whether this follows, extends, or replaces the established contract, and an implementation that establishes one updates its anchor in the same PR. I'm filing a separate issue for the three siblings that restate rather than derive the vocabulary (`InteractiveStepProps.targetAction`, `InternalAction.targetAction: string`, and the non-exhaustive `dispatchInteractiveAction` if/else) since that's PR 2/3 territory, not yours to fix here.

**Cheap, folding in here**

- `live-tab-executor.ts` — the default omits `assertExhaustive` while the hook's switch gained it in this same PR. This is the one place actions replay into another tab, and without it the switch is no longer compile-checked against the union.
- `interactive.hook.ts` — `case 'multistep'` logs a now-recognized verb as "Unknown interactive action". Something like "multistep is executed by InteractiveMultiStep, not the element path" (the shape `live-tab-executor` already uses) points debugging the right way.
- `interactive.hook.ts` / `interactive-multi-step.tsx` — `InteractiveRequirementsData` is declared byte-identically in tier 2 and tier 4 with neither copy exported, and it's load-bearing: `interactive-section.tsx:717` and `use-section-requirements.ts:116` still pass `targetAction: 'section'`, which isn't an `INTERACTIVE_ACTION_TYPES` member. Export one named type so the copies can't drift and the next reader knows the widening is deliberate.
- `docs/developer/engines/interactive-engine.md:265` — the action list omits `multistep` and `popout`, which matters more now that the vocabulary is load-bearing.

**Deferring to an issue, needs a product call**

On base, the component-owned section branch still called `executeInteractiveAction`, which emitted a `DoItButtonClick` Faro span via `withFaroUserAction` even when the switch matched nothing. Component-owned steps — the exact population this commit preserves — now emit no span, so section runs containing code-block, terminal-connect, or datasource-check steps will show a Do-it count drop with nothing in the dashboard explaining it. I haven't verified the magnitude. Whether a verb-less step counts as a "Do it" is a product question rather than a code one, so I'll file it rather than block on it.

---

Reviewed at `120bd0253aea79e42bd24d92c670346bca2c3c23`. Typecheck, lint, and coverage were deliberately withheld — community PR content is not executed locally, so CI is the authority. CI is green on this head (the only failure is the known fork-permission `publish-report` signature, which is infrastructure, not yours). I did not independently verify the "PR 1 of three" framing against #1525.

loglapa:
Addressed the requested changes in 27f6162: unknown multi-step actions now surface UI failure after the cancellation check; all five rejection paths have behavior coverage; cross-tab dispatch is exhaustive; the shared requirements pseudo-action type is canonical; the multistep log and action docs are corrected; and the PR body plus interactive-engine contract anchor now state the contract evolution. The telemetry product question remains intentionally deferred as requested. Focused suites pass (91 tests, 5 existing todo). Ready for re-review.

moxious (approved):
Thanks — this is a disciplined slice. The fix pass resolved every item from the last round, and the deliberate restraint (leaving `KNOWN_TARGET_ACTIONS` and `GUIDED_VERBS` byte-unchanged, keeping the sequence deletion and the guided schema for PRs 2 and 3) is exactly right.

Eleven separate claims were verified at head. One candidate finding was raised during review and then refuted: the cross-tab test you replaced was already vacuous — its synchronous `uninstall()` set `cancelled` before the awaited auth gate resumed, so both assertions passed on nothing. Your replacement is weak but real, so no coverage was lost.

Approving.

PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1757
Summary: Type interactive actions at the DOM ingress boundary
Verdict: Approve

PR #1758, opened 2026-08-31
Use request-scoped logging for custom guide truncation

### LINKED ISSUE #1403: Custom guide catalogue drain: per-item decode-failure skip and request-scoped truncation logging
## What

Two small hardening items in the custom guide catalogue drain path, from the #1400 review. Same file/function area, so one small PR:

### 1. Per-item skip on spec decode failure

`customGuideHTTPClient.ListPage` (`pkg/plugin/custom_guide_repository_client.go`) aborts the **whole page** when one `items[].spec` fails `json.Unmarshal`. A status-less decode error classifies as _transient_, so a single type-drifted guide (e.g. after CRD schema evolution in pathfinder-backend) turns the entire catalogue into a permanently retried `503 + Retry-After` — the wrong failure class for a condition retrying will never fix.

Fix: skip the malformed entry and continue, like the existing `entry.ID == ""` branch, logging the skip (with the guide's metadata name if available) so drift is diagnosable.

### 2. Thread the request logger into `drainCustomGuides`

The aggregate-budget truncation warning in `drainCustomGuides` (`pkg/plugin/custom_guide_repository.go`) uses `log.DefaultLogger`, while every other log in the route goes through `a.ctxLogger(r.Context())`. Pass the logger in (the sibling completion-records drain already threads its logger) so the warning carries request context.

## Notes

- Both are Low severity; neither blocks the feature. Filed as follow-up per the #1400 review disposition.
- Not in scope: the per-request `context.WithoutCancel` drain — deliberate and commented in `custom_guide_repository.go`, bounded by the 60s aggregate deadline; wontfix unless traffic patterns change.

Refs #1400.

loglapa:
## What does this PR do?

Routes custom-guide aggregate-budget truncation warnings through the request-scoped logger already used by the handler instead of `log.DefaultLogger`. The logger is captured before the detached, deadline-bounded drain and passed explicitly, matching the completion-index sibling pattern.

Pagination, aggregate caps, `context.WithoutCancel`, per-item decode skipping, and response behavior are unchanged. The aggregate-budget test now proves the warning reaches the drain's capturing logger.

## Linked issue(s)

Closes #1403

## Verification

- focused custom-guide drain tests with real `httptest` listeners
- `go build ./...`
- full `npm run check`: all 9 stages passed, including Go lint/tests and 505 Jest suites / more than 8,100 tests
- no runtime smoke required: this changes only logger routing and does not alter the proxy route, credential path, or wire contract

## Checklist

### DIFF
--- a/pkg/plugin/custom_guide_repository_test.go
+++ b/pkg/plugin/custom_guide_repository_test.go
@@ -917,7 +917,7 @@ func TestCustomGuideHTTPClient_DecodeWarnSummaryOnAggregateBudget(t *testing.T) 
 		t.Fatalf("drainCustomGuides: %v", err)
 	}
 	if !logger.warnedWith("custom guide catalogue LIST truncated at aggregate budget") {
-		t.Fatal("aggregate-budget truncation warning did not use the drain's request-scoped logger")
+		t.Error("aggregate-budget truncation warning did not use the drain's request-scoped logger")
 	}
 
 	details, summaries := 0, 0
--- a/pkg/plugin/custom_guide_repository_test.go
+++ b/pkg/plugin/custom_guide_repository_test.go
@@ -869,7 +869,7 @@ func TestCustomGuideHTTPClient_DecodeWarnBudgetSpansPages(t *testing.T) {
 
 	logger := newCapturingLogger()
 	c := newCustomGuideHTTPClient(srv.URL, &stubMinter{token: "at-xyz"}, "id-token-abc", logger)
-	if _, _, err := drainCustomGuides(context.Background(), testNamespace, c); err != nil {
+	if _, _, err := drainCustomGuides(context.Background(), testNamespace, c, logger); err != nil {
 		t.Fatalf("drainCustomGuides: %v", err)
 	}
 
@@ -913,8 +913,11 @@ func TestCustomGuideHTTPClient_DecodeWarnSummaryOnAggregateBudget(t *testing.T) 
 
 	logger := newCapturingLogger()
 	c := newCustomGuideHTTPClient(srv.URL, &stubMinter{token: "at-xyz"}, "id-token-abc", logger)
-	if _, _, err := drainCustomGuides(context.Background(), testNamespace, c); err != nil {
+	if _, _, err := drainCustomGuides(context.Background(), testNamespace, c, logger); err != nil {
 		t.Fatalf("drainCustomGuides: %v", err)
+	}
+	if !logger.warnedWith("custom guide catalogue LIST truncated at aggregate budget") {
+		t.Fatal("aggregate-budget truncation warning did not use the drain's request-scoped logger")
 	}
 
 	details, summaries := 0, 0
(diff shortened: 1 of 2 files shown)


pkg/plugin/custom_guide_repository_test.go
```diff
@@ -913,9 +913,12 @@ func TestCustomGuideHTTPClient_DecodeWarnSummaryOnAggregateBudget(t *testing.T)
 
 	logger := newCapturingLogger()
 	c := newCustomGuideHTTPClient(srv.URL, &stubMinter{token: "at-xyz"}, "id-token-abc", logger)
-	if _, _, err := drainCustomGuides(context.Background(), testNamespace, c); err != nil {
+	if _, _, err := drainCustomGuides(context.Background(), testNamespace, c, logger); err != nil {
 		t.Fatalf("drainCustomGuides: %v", err)
 	}
+	if !logger.warnedWith("custom guide catalogue LIST truncated at aggregate budget") {
```
moxious, line 919: N1 (nit): this new guard is a real regression detector, but t.Fatal aborts before the decode-warn detail/summary counts below, which verify unrelated invariants. t.Error would keep both signals visible on failure.

moxious (changes requested):
Thanks for finishing this one off — routing the aggregate-budget truncation warning through the request-scoped logger is exactly right, and matching the `getCompletionIndex` sibling contract rather than inventing a second shape is the part I appreciate most. After this, `log.DefaultLogger` appears on no request path in `pkg/` at all.

I verified the substance and it holds: all three `drainCustomGuides` call sites are updated, the logger is captured before the detached `context.WithoutCancel` drain, warn level is preserved so the Faro-visible signal is unchanged, and the new assertion is genuinely reachable (`customGuideListMaxTotalEntries = 1` with an 11-item stub and a continue token makes `truncated` true on both clauses). No behaviour, wire, credential, or identity change. I also checked the "Closes #1403 is premature" worry and it's unfounded — the issue's per-item spec-decode item is already implemented at base, so this completes the only remaining one.

Requesting changes for a single one-word fix rather than approving, because a nit noted on an approved PR is a nit that never gets made:

- `custom_guide_repository_test.go` — `t.Fatal` → `t.Error` on the new `warnedWith` guard, so the decode-warn detail and summary assertions below it still report on failure.

That's the whole list. Push that and I'll merge.

---

Reviewed at `2d45cb8bd4134e726bf7a2dffb60ca1e8deb3a55`. Go toolchain verification was deliberately withheld — community PR content is not executed locally, so CI is the authority on `lint:go`, `test:go`, and `go build`. CI is green on this head (the only failure is the known fork-permission `publish-report` signature, which is infrastructure, not yours).

loglapa:
Changed the guard from t.Fatal to t.Error in f73e9c1 so the downstream decode-warning assertions still run. The focused Go test passes. Ready for re-review.

moxious (approved):
Thanks for the quick turnaround — `t.Fatal` → `t.Error` is exactly what the last round asked for, and the diff is that one line and nothing else.

I checked the two things this shape usually gets wrong and both are clean: `FromContext` snapshots the attribute slice rather than retaining the context, and the drain runs synchronously on the handler goroutine anyway, so there is no post-return logger use. No identity attribute can reach the sink either — `uname` is gated on `DataSourceInstanceSettings != nil`, unreachable for an app plugin.

Approving; this supersedes my earlier changes-requested.

PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1758
Summary: Use request-scoped logging for custom guide truncation
Verdict: Approve

PR #1759, opened 2026-08-31
test(validation): guard PresentationalBlock type drift

### LINKED ISSUE #1705: The PresentationalBlock drift guard does not detect drift
## What is wrong

`src/validation/type-coupling.test.ts:157-159` carries this comment:

> Drift guard: `PresentationalBlockSchema` (collapsible children) and the `PresentationalBlock` type must list the same block types. If the union gains/loses a member on one side only, one of these assertions fails.

The assertions below it do not do that. They are two hardcoded lists of sample blocks, and both call `PresentationalBlockSchema.safeParse`. Nothing reads the `PresentationalBlock` TypeScript type at all, so the two sides can drift apart with no error.

I confirmed this. Adding a member to `PresentationalBlockSchema` alone produced no failure. Adding one to the `PresentationalBlock` type alone produced no failure either.

## Why it matters

`PresentationalBlock` sets what a `collapsible` block can hold. When the two sides disagree, the schema accepts a block the type rejects, or the reverse. The comment tells the next reader a safety net exists, so nobody looks.

## The fix

Add a type-level assertion to `src/validation/type-coupling.test.ts`. Both directions are needed, because each catches drift on one side:

```ts
import type { z } from 'zod';
import type { PresentationalBlock } from '../types/json-guide.types';

type InferredPresentational = z.infer<typeof PresentationalBlockSchema>;

// Drift guard: each direction catches an addition the other side did not get.
export const _schemaMatchesType: PresentationalBlock = {} as InferredPresentational;
export const _typeMatchesSchema: InferredPresentational = {} as PresentationalBlock;
```

Keep the existing `safeParse` tests. They check runtime acceptance, which the type assertion does not.

## How to test your change

1. Run `npm run typecheck`. It must pass.
2. Add `JsonInputBlock` to the `PresentationalBlock` union in `src/types/json-guide.types.ts`.
3. Run `npm run typecheck` again. It must now fail with `TS2322`.
4. Undo step 2.

## Two things to know

- Test files are typechecked. `.config/tsconfig.json` sets `include: ["../src"]`, so an assertion in a `.test.ts` file is live.
- Use `export const`, not `const`. A bare `const` triggers `TS6133`, unused variable.

## Context

Found during review of #1702, which edits both sides of this union. That PR's edit is correct — the unions were checked by hand. This issue is about the missing guard, not about that PR.

sehrawatvinay71-svg:
Fixes #1705

## What changed

- Added bidirectional type-level assertions between the Zod-inferred presentational block type and `PresentationalBlock`.
- Kept the existing runtime `safeParse` membership tests separate.

## Why

The existing runtime tests did not compare the Zod-inferred union with the handwritten TypeScript union, so the two definitions could drift without failing.

## Testing

- `npm run typecheck`
- `npm run test:ci -- src/validation/type-coupling.test.ts --coverage=false`
- `npm run lint`
- Prettier check
- Confirmed that temporarily adding `JsonInputBlock` to only one side fails with `TS2322`

### DIFF
--- a/src/validation/type-coupling.test.ts
+++ b/src/validation/type-coupling.test.ts
@@ -1,4 +1,4 @@
-import type { JsonGuide, JsonBlock } from '../types/json-guide.types';
+import type { JsonGuide, JsonBlock, PresentationalBlock } from '../types/json-guide.types';
 import {
   JsonGuideSchema,
   JsonMarkdownBlockSchema,
@@ -22,7 +22,13 @@ import {
   KNOWN_FIELDS,
   type InferredJsonGuide,
 } from '../types/json-guide.schema';
-import { z } from 'zod';
+import type { z } from 'zod';
+
+type InferredPresentational = z.infer<typeof PresentationalBlockSchema>;
+
+// Each direction catches an addition the other side did not get.
+export const _schemaMatchesType: PresentationalBlock = {} as InferredPresentational;
+export const _typeMatchesSchema: InferredPresentational = {} as PresentationalBlock;
 
 describe('Type Coupling: TypeScript <-> Zod', () => {
   it('JsonGuide types should be assignable', () => {
@@ -159,9 +165,6 @@ describe('KNOWN_FIELDS sync', () => {
     verifyFields(JsonDividerBlockSchema, 'divider');
   });
 
-  // Drift guard: PresentationalBlockSchema (collapsible children) and the
-  // PresentationalBlock type must list the same block types. If the union
-  // gains/loses a member on one side only, one of these assertions fails.
   describe('PresentationalBlockSchema membership', () => {
     it('accepts the content block types', () => {
       const accepted: unknown[] = [


moxious:
Thanks for adding the bidirectional type-level assertions between the Zod-inferred presentational block type and `PresentationalBlock` — that closes a real drift gap the runtime `safeParse` tests couldn't catch on their own.

One thing to clear before we review: the Contributor License Agreement isn't signed yet, and we can't accept a contribution until it is. The [contributing guide](https://github.com/grafana/grafana-pathfinder-app/blob/37d755c5f67cad6268d8ee38f64bf06313acf4b9/CONTRIBUTING.md) walks through it, and the `cla-assistant` comment above links straight to the signing page.

Once that's done, we'll pick this up for review.

moxious (approved):
Thanks for this, and welcome — first contribution here.

A drift guard is only worth having if it actually fails on drift, so I did not take the assertions on faith: I worked through member-add, member-remove, required-field and narrowed-type drift on both sides. All of them produce a genuine `TS2322`. Details in a separate comment below.

Approving.

PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1759
Summary: guard PresentationalBlock type drift
Verdict: Approve

moxious:
A bit more detail on the verification, since the review body itself is terse by design.

The thing I most wanted to rule out is a guard that compiles clean forever and protects nothing. Several ways that happens here, and none of them apply:

- **`z.infer` degrading to `unknown[]`.** `PresentationalBlockSchema` is a flat `z.union` with no `z.lazy`, so it resolves to a real union. The older `JsonGuide` coupling test in the same file has to work around exactly that degradation, which is why it is deliberately weaker.
- **Trivial assignability.** No `z.any()`, no `.passthrough()`, no `.catchall()`, and no index signature on either side, so nothing is assignable by accident.
- **Erasure.** These are annotated `export const` values rather than bare type aliases, and `.config/tsconfig.json` includes `../src` with no test exclusion — so `tsc --noEmit` really does check them, and `Typecheck` is in `ci-gate`'s `needs`. The guard is merge-blocking, not decorative.

Both directions are independently load-bearing. The case I expected to find a hole in was member *removal*, since `JsonDividerBlock` is nearly all-optional — but it holds, because every other member carries a required field (`content` / `src` / `title`) that the divider lacks.

One gap, which is outside this issue's scope and not something to fix here: an **optional** field added to only one side still compiles in both directions. Union membership is guarded; per-member optional drift is not. The pre-existing `KNOWN_FIELDS sync` suite covers part of that on the schema side.

Removing the old three-line comment was the right call, by the way — it claimed the two `safeParse` assertions enforced membership parity, which is the exact false claim #1705 was filed about. Leaving it would have been worse than the missing guard.

PR #1760, opened 2026-08-31
chore(review): rank fired contract-evolution gates deterministically

moxious:
## Summary

`.cursor/skills/review/SKILL.md` tells the review orchestrator to run the contract-evolution gate for every activated subsystem and cross-cutting concern with concrete routing paths, but `buildReviewPlan` accepted a single `contract_evolution` object. When two gates fired there was one specialist slot and no defined winner: which contract got audited against its prior semantic PRs was undefined, and the gate that lost left no trace in the plan.

`buildReviewPlan` now accepts a list of fired gates, ranks them in a documented order, and routes every unselected gate to the root synthesizer through the overflow path that already existed, so `coverage` names it as root-owned.

## What to look at

- **[`concern-context.mjs`](https://github.com/grafana/grafana-pathfinder-app/blob/8f15f745d5b71533af2f20d9540b4e7169adfd5d/.cursor/skills/review/scripts/concern-context.mjs) — `firedContractGates`.** Normalizes the singular object and the array into one list, rejects duplicate `concern_id`s, then sorts by anchor touch, then higher `prior_semantic_pr_count`, then ascending `concern_id`. The id comparison is a plain `<`/`>` rather than `localeCompare`, so ordering cannot shift with locale.
- **The `listed` flag is the backward-compatibility boundary.** Field validation and the empty-packet guard apply to the array form only. A singular object still ignores both ranking fields, because one gate ranks against nothing — that is why the singular path needs no migration.
- **The slot-award loop.** `specialistTaken` means at most one specialist worker can exist, so the hardcoded `contract-evolution` worker id cannot collide. A gate that cannot use the slot — invalid `concern_id`, over the packet envelope, or an array-form gate with an empty packet — falls through to `rootConcernIds` and the slot passes to the next ranked gate rather than being wasted.
- **[`SKILL.md`](https://github.com/grafana/grafana-pathfinder-app/blob/8f15f745d5b71533af2f20d9540b4e7169adfd5d/.cursor/skills/review/SKILL.md) is the prose contract the code implements.** It states the selection order, that the caller supplies `touches_anchor_with_consumers` because the planner never infers anchor reach, and that only one *contract* specialist runs — a dedicated security worker still runs alongside it.

Worker and packet budgets are deliberately untouched: two general workers plus one gated contract specialist, eight files and 30,000 characters per packet. This change adds at most one worker to the plan, which is the slot that already existed.

## Why

The contract-evolution specialist is the only reviewer that reads prior semantic PRs, their reviews, and directly linked follow-up issues to judge whether a change is branching a contract that already has consumers. With one slot and no ordering, that check silently did not run for one of the two contracts, and nothing recorded which one lost.

The motivating case is #1750, which changed `src/validation/**` and `src/cli/**` in one 8-file diff: `guide-schema-and-contracts` fired with 30 prior semantic PRs and `cli-and-e2e-runner` fired with 21. Both gates triggered into a single slot.

Ranking is stated by the caller rather than inferred. The gate script already emits `prior_semantic_pr_count`, and anchor reach is a judgement from the diff that the planner cannot make on its own. A listed gate that omits or mistypes either field is rejected instead of coerced, because a coerced value would silently invert the order — the exact failure mode this change exists to remove. Anchor-first is not a new concept: `SKILL.md` already names a changed hunk touching a named anchor with existing consumers as an independent specialist trigger.

This change is deliberately minimal; a refactor of the review skill is out of scope.

## How to verify

The plan builder is a script, so it can be exercised directly. Steps 1–5 use `node .cursor/skills/review/scripts/concern-context.mjs --plan <file>`.

1. **Two gates, one slot.** Write a plan with `contract_evolution` as an array of two gates — `guide-schema-and-contracts` with `prior_semantic_pr_count: 30` and `cli-and-e2e-runner` with `21`, both `touches_anchor_with_consumers: false`, each with a non-empty `context`. Expect one `contract_evolution` worker owning `guide-schema-and-contracts`, and `coverage["contract-evolution:cli-and-e2e-runner"] === "root"`.
2. **Order independence.** Reverse the array and re-run. The winner should not change.
3. **Anchor beats history.** Set `touches_anchor_with_consumers: true` on the gate with the *lower* count. The slot should move to it. Move the flag to the other gate and confirm the winner moves back — if it does not, the top tier is not being read.
4. **A gate that cannot use the slot yields it.** Give the top-ranked gate an empty `context`, or a `concern_id` with underscores instead of hyphens. An empty array-form packet routes that gate to `root` and the slot falls through; an invalid `concern_id` fails the build loudly rather than quietly transferring the slot.
5. **Singular form unchanged.** Pass `contract_evolution` as a bare object with only `concern_id` and `context` — no ranking fields — and confirm it still takes the specialist slot and produces the same `coverage` key as before.
6. **The guards are not decorative.** Delete any one guard from `firedContractGates` or the slot-award loop and run `node --test .cursor/skills/review/scripts/concern-context.test.mjs`. A guard whose removal leaves the suite green is not doing anything.

The 12 added tests cover each ranking tier, both yield paths, the duplicate-id rejection, the malformed-packet error naming which gate to repair, and the singular-form path. The test file is additions only — no existing assertion was changed or removed, which is the backward-compatibility evidence.

## Breaking changes

None. The singular `contract_evolution` object keeps its existing shape, its worker id, its `coverage` key, and its overflow behaviour, and ignores the two ranking fields exactly as it did before. The array form and its stricter validation are new surface, so no existing caller can trip on them.

## Out of scope

- When the gate fires is unchanged. Its trigger signals are not tightened here, so a gate that over-fires today still over-fires.
- `docs/design/CONCERNS.md` and `docs/design/CONCERN_DETAILS.md` are not touched.
- No ranking-basis field is added to the plan or debug output. `coverage` already distinguishes the audited contract from the root-owned ones, and the lowest-`concern_id` tie-break stays as documented.

Refs #1750


moxious:
@cursoragent review

Jayclifford345 (approved):

PR #1766, opened 2026-09-02
fix(security): reject Unicode formatting characters

### LINKED ISSUE #1547: Control-byte guard gives no Trojan Source (bidi) protection
Follow-up from #1537 (review of the control-byte guard added in `src/validation/control-bytes.test.ts`).

## What the current guard does and does not cover

The guard added in #1537 rejects raw C0 control bytes (everything below `0x20` except tab, newline, carriage return) plus `0x7f`. It operates on **raw file bytes**, which is exactly right for its stated purpose: that byte set is precisely what makes `file(1)` classify a file as `data` and makes `grep -r` / `rg` skip it silently (the failure documented in #1536).

**It therefore gives no protection against Trojan Source** ([CVE-2021-42574](https://nvd.nist.gov/vuln/detail/CVE-2021-42574)). The bidi-override and invisible-formatting characters used in those attacks are all above `0x20` once UTF-8 encoded, so the predicate cannot match them:

| Character | UTF-8 bytes | Matched by current guard? |
| --- | --- | --- |
| U+202A–U+202E (bidi embedding/override) | `e2 80 aa`–`e2 80 ae` | no |
| U+2066–U+2069 (bidi isolate) | `e2 81 a6`–`e2 81 a9` | no |
| U+200B ZWSP, U+200E LRM, U+200F RLM | `e2 80 8b` / `8e` / `8f` | no |
| U+00AD soft hyphen | `c2 ad` | no |
| U+FEFF BOM | `ef bb bf` | no |
| U+2060 word joiner | `e2 81 a0` | no |

This is a **different threat model** (source that renders differently than it compiles, i.e. a code-review integrity attack) from the one #1537 addresses (source that is invisible to search tooling), and it needs a different predicate — one that decodes UTF-8 and matches code points rather than scanning bytes.

## Current state

Verified during the #1537 review: **zero occurrences** of any character in the table above across all 22 entries under `src/locales/`, and none elsewhere in the tracked tree. There is no RTL locale in the repo (no `ar`, no `he`), so there is no legitimate need for bidi marks today. A guard added now would start green.

## Suggested shape

- Decode each tracked text file as UTF-8 and reject the bidi-override / invisible-formatting set above.
- Unlike the control-byte case there is a plausible legitimate need later (an RTL locale would want U+200F), so this guard probably wants a scoped allowlist — for example permitting the set under `src/locales/**` while rejecting it in `.ts` / `.tsx` / `.go`.
- Reuse the failure-message shape from `control-bytes.test.ts`: name `file:line:column`, name the offending code point, and cite the doc. Per #603, the message is the deliverable.

Filed as a separate issue rather than folded into #1537 because it is a distinct predicate, a distinct threat model, and would need its own allowlist design. Part of the #603 mechanical-constraint-enforcement family.

ravik453:
## What does this PR do?

Adds a repository-wide guard against invisible Unicode formatting and bidirectional control characters that can make source text misleading or alter its display.

The guard:

- Scans decoded tracked source text by Unicode code point.
- Rejects the 15 forbidden Unicode formatting/bidirectional code points defined by issue #1547.
- Reports the file, line, column, and offending code point.
- Scans tracked text outside `src/` as well.
- Avoids embedding the forbidden characters in the source itself by constructing test fixtures at runtime.

## Linked issue(s)

Fixes #1547

## Testing

- Unicode formatting character tests: 9 passed
- Governance tests: 185 passed
- Full Jest suite: 8,135 passed, 18 skipped, 5 todo
- Go lint: 0 issues
- Go tests: passed
- Review contract tests: 64 passed
- Script tests: 86 passed

`test:coverage` could not complete locally because the WSL environment was OOM-killed during the coverage run.

## Notes

The repository was verified to contain no occurrences of the forbidden code points before adding the guard.


moxious:
Thanks for the Unicode formatting character rejection fix — format- and control-character hygiene in tracked source is something this repository cares about, and you brought tests with it.

One thing to clear before we review: both commits on this branch are unsigned, so they can't be verified. This repository requires verified signed commits on every commit in a PR. The setup steps are in our contributor guide: https://github.com/grafana/grafana-pathfinder-app/blob/97cf2ef4729f4beb2bf62f9f87a15a19bd10be4d/CONTRIBUTING.md — once signing is configured you can re-sign the two existing commits and force-push. Your CLA is already signed, so that side is clear.

ravik453:
Thank you, I've re-signed both commits on this branch, and force-pushed them up to the PR.The commits should now be showing up with the green 'Verified' status badge. The branch is ready for your review.

docs/developer/bugfix-patterns.md
```diff
@@ -12,3 +12,4 @@ Patterns from resolved robustness issues. Loaded during `/bugfix` Phase 2 to inf
...
 | #1351 | UX behavior bug | Raw browser pathnames include Grafana's `appSubUrl` while plugin routes are basename-relative; use `locationService`'s normalized pathname for route identity checks                                                                                                                      | `src/utils/pathfinder-deep-link-handler.ts`, `src/components/docs-panel/hooks/usePanelMode.ts`                                                                                                                                                                                      | 2026-07-15 |
+| #1547 | Security robustness bug | Unicode formatting and bidirectional control code points can bypass byte-oriented control-character checks; scan decoded source text by Unicode code point and reject the explicit forbidden set | `src/validation/unicode-format-characters.ts`, `src/validation/unicode-format-characters.test.ts` | 2026-09-02 |
```
moxious: blocking — this row is not padded to the table column widths, so `prettier --check` fails here and the required CI Gate check is red. The base blob is prettier-clean, this row is the only cause, and the failed job log names only this file. `Security robustness bug` is wider than any existing Category value, so prettier re-pads all eleven table lines.

Run `npm run prettier` and commit the result. Consider adding a `<!-- prettier-ignore -->` line above the table, as `docs/design/CONCERNS.md` does, so future single-row appends stop re-padding the whole table.

Separately: `Security robustness bug` is a sixth Category value. `.cursor/skills/bugfix/SKILL.md` defines a closed five-value vocabulary that the `/bugfix` Phase 2 classifier reads back from this table, and issue #1547 recorded severity:LOW with "the value is hygiene, not security".

src/validation/unicode-format-characters.test.ts
```diff
@@ -0,0 +1,220 @@
...
+  return [
+    `Unicode format character${hits.length === 1 ? '' : 's'} found in tracked source:`,
+    '',
+    ...locations,
+    '',
+    'These invisible Unicode formatting characters can make source text misleading or alter its display.',
+    'Remove the raw character from the tracked file. If the character is required at runtime, encode it',
+    'as an escape or construct it in code rather than embedding the raw code point in source.',
+    '',
+    'See AGENTS.md § "Control characters in source".',
```
moxious, line 100: blocking — this citation points at a section that documents a different character class. At this head, AGENTS.md § "Control characters in source" covers raw control bytes only: it prescribes `\x00`-style escapes, says tab, newline and carriage return are fine, and names `control-bytes.test.ts` as the sole enforcer. Nothing there states the invisible-Unicode rule or names this guard, and the PR does not touch AGENTS.md.

PR #1537 recorded this guard pattern as two halves — the failure message as the teaching surface, plus an AGENTS.md entry carrying the rule, the reason, the spelling convention and a pointer. This ships the message half only, and CLAUDE.md is literally `@AGENTS.md`, so every agent working here reads guidance that omits an enforced constraint.

Add the rule to AGENTS.md in this PR, naming this test as its enforcer, and keep this citation pointing at whichever heading ends up carrying it.

src/validation/unicode-format-characters.ts
```diff
@@ -0,0 +1,34 @@
+export interface UnicodeFormatHit {
+  relPath: string;
+  line: number;
+  column: number;
+  codePoint: number;
+}
+
+const FORBIDDEN_UNICODE_FORMAT_CODE_POINTS = new Set([
```
moxious, line 8: suggestion — this set is declared again in the test file, which builds its `should detect all forbidden Unicode format characters` expectation from its own copy. Removing a value here fails the test, but adding one here passes silently, so that test name claims a completeness it does not verify. The drift widens detection rather than narrowing it, so guard coverage is never silently weakened — but exporting this set and deriving the fixture from it removes the second edit site.

While here: U+FEFF has no offset-zero carve-out, so a legitimate leading BOM is reported at line 1 column 1 with advice ("remove the raw character", "encode it as an escape") that does not fit a byte-order mark. No tracked file begins with EF BB BF today, so either policy is cheap to encode — but pick one and add the test.

moxious (changes requested):
## Merge contract

1. [blocking] **prettier-bugfix-patterns-row — Appended bugfix-patterns.md row is not prettier-formatted, so the required CI Gate check fails** (high · build-and-ci)
   The row added at docs/developer/bugfix-patterns.md:15 is not padded to the table's column widths, and its Category value 'Security robustness bug' is wider than any existing one, so prettier wants to re-pad all eleven table lines. The base blob passes `prettier --check` (exit 0) and the head blob fails (exit 1) with the repo's pinned prettier 3.9.6. Markdown is in the `prettier-test` glob, `.prettierignore` does not exclude docs/, the ci.yml `prettier` job runs `prettier-test`, and `ci-gate` hard-fails on it with no skip tolerance. The live CI run on this head shows Prettier fail and CI Gate fail, and the job log names only this file. `scripts/check.js` includes prettier-test, so `npm run check` is red too.
   Required: Run `npm run prettier` (or `prettier --write docs/developer/bugfix-patterns.md`) and commit the re-padded table. Optionally add a `<!-- prettier-ignore -->` line above the table, as docs/design/CONCERNS.md does, so future single-row appends stop re-padding the whole thing.

2. [blocking] **agents-md-rule-missing-for-new-guard — The enforced rule is never written into AGENTS.md, and the guard's message cites a section documenting a different character class** (medium · cross-cutting-architecture)
   The guard's failure message ends with `See AGENTS.md § "Control characters in source".`, but the PR does not touch AGENTS.md. At this head that section covers raw control bytes only: it prescribes `\x00`-style escapes, says tab, newline and carriage return are fine, and names control-bytes.test.ts as the sole enforcer — so for U+200B, U+200F or U+FEFF it reads as a different rule with a different remedy. PR #1537 recorded this guard pattern as two halves: the failure message as the teaching surface, plus an AGENTS.md entry carrying the rule, the reason, the spelling convention and a pointer. This PR ships the message and omits the AGENTS.md half. CLAUDE.md is literally `@AGENTS.md`, so that file is loaded into every agent's context here — a rule CI enforces but AGENTS.md never states is one agents keep violating until CI stops them. No other doc surface states it either: no hit for unicode, bidi, zero-width or U+200 across AGENTS.md, CLAUDE.md, CONTEXT_INDEX.md, docs/design or .cursor/rules. The five author-facing guards in src/validation each carry such a prose rule.
   Required: Extend AGENTS.md § 'Control characters in source' (or add an adjacent subsection) to state the invisible-Unicode rule, name src/validation/unicode-format-characters.test.ts as its enforcing test, and give the escape-or-construct remedy — then keep the failure-message citation pointing at whichever heading carries the rule.

## Follow-ups

1. [follow_up] **walker-copy-drops-binary-arm — The copied tracked-file walker drops the sibling guard's binary-asset arm, so an unlisted binary extension will fail opaquely** (low · testing-and-verification)
   BINARY_EXTENSIONS, listTrackedFiles, collectScannableFiles and scanFiles are duplicated near-verbatim from control-bytes.test.ts, but its content-derived binary heuristic (BINARY_DENSITY_THRESHOLD, BINARY_MIN_HITS) and the failure section that names BINARY_EXTENSIONS and the guard file are not carried over. The new formatFailure has one branch whose only advice is to remove the character. The direction is over-report, not under-report: U+00AD is the two-byte sequence C2 AD, so decoding a blob as UTF-8 yields roughly one hit per 65,536 bytes, and the five tracked PNGs would produce 1, 3, 4, 5 and 5 hits respectively if they were scanned — they are excluded only by the .png entry. BINARY_EXTENSIONS omits .bmp, .icns, .bin, .dat, .jar, .so, .dylib, .tar, .tgz, .mov and .ogg. Nothing fails today, and the extension list now has two edit sites that must stay in sync.
   Follow-up: Extract the shared walker — BINARY_EXTENSIONS, the enumeration helpers and the binary-density arm — into one module both guards import; or at minimum add the sibling's 'add the extension to BINARY_EXTENSIONS in <guard path>' remediation line, with a test asserting it.

2. [follow_up] **bugfix-row-category-and-security-framing — The new pattern row introduces a sixth Category value and a security framing issue #1547 decided against** (low · cross-cutting-architecture)
   .cursor/skills/bugfix/SKILL.md defines a closed five-value Category vocabulary — Logic bug, State bug, Rendering bug, UX behavior bug, Visual bug — and all nine existing rows use one of them. The new row's 'Security robustness bug' is a sixth value, and bugfix-patterns.md states the table is loaded during /bugfix Phase 2 to inform classification, so the column is read back by a classifier that knows only the five. Separately, issue #1547's decision comment records 'Severity is LOW, and the value is hygiene, not security', and the issue carries severity:LOW — while the PR title is 'fix(security): ...', which the changelog skill propagates into release notes by conventional-commit prefix. Sentence case on the row is correct.
   Follow-up: Use one of the five existing Category values, or add the new one to .cursor/skills/bugfix/SKILL.md in the same change. Consider the hygiene framing that issue #1547 suggested for the title.

## Suggestions

1. [suggestion] **forbidden-set-duplicated-in-test — The forbidden code-point set is declared twice, so a later addition to the module would be exercised by no test** (low · testing-and-verification)
   unicode-format-characters.ts declares FORBIDDEN_UNICODE_FORMAT_CODE_POINTS without exporting it, and the test re-declares the same 15 values, building the 'should detect all forbidden Unicode format characters' expectation from its own copy. Removing a value from the module fails the test, and adding one to the test only fails the test — but adding one to the module only passes silently. Silent drift therefore widens detection rather than narrowing it, so coverage is never silently weakened; the cost is two edit sites and a test whose name claims completeness it does not verify.
   Suggested: Export FORBIDDEN_UNICODE_FORMAT_CODE_POINTS from the module and derive the test fixture from it.

2. [suggestion] **bom-at-offset-zero-policy-unstated — A leading UTF-8 byte-order mark is rejected with no test and no stated policy** (low · testing-and-verification)
   Node's fs.readFileSync(path, 'utf-8') does not strip a UTF-8 BOM, so the string starts with U+FEFF and the detector reports it at line 1 column 1. 0xfeff is in the set with no offset-zero carve-out. The four detector tests cover mid-line, multi-line, astral-adjacent and escaped cases but not an offset-zero BOM, and no tracked file begins with EF BB BF today. A leading BOM is a legitimate encoding artifact rather than a smuggling attempt, and the message's advice — remove the character or encode it as an escape — does not fit one. Rejecting it may well be the right policy; as shipped the decision is implicit and untested.
   Suggested: Pick one and encode it: either add a test asserting an offset-zero BOM is flagged plus a message line saying to re-save without a BOM, or exempt U+FEFF at line 1 column 1 with a test proving a mid-file U+FEFF is still flagged.

3. [suggestion] **set-omits-tag-character-block — The set omits the U+E0000-E007F tag block, the smuggling vector most relevant to an agent-instruction repository** (low · security)
   The module lists exactly the 15 code points issue #1547 enumerated, so the set is in spec rather than wrong. But this repository tracks AGENTS.md, CLAUDE.md, .cursor/rules/** and .cursor/skills/**, which agents consume as instructions — precisely the surface tag-character ASCII smuggling targets. A full-tree scan at this head found zero occurrences of U+E0000-E007F, U+200C, U+200D, U+2028, U+2029, U+2061-2064, U+180E, U+3164 and U+034F, so the block is free to add. Note the counter-evidence that a fixed set was deliberate: U+FE00-FE0F variation selectors appear in 32 tracked files as legitimate emoji presentation selectors and must not be added.
   Suggested: Consider adding U+E0000-E007F as a range check in a follow-up, and record why U+FE00-FE0F, U+200C and U+200D are deliberately allowed so the next reader does not re-litigate it.

## Nits

1. [nit] **stale-base-unrelated-lint-red — The Lint job is red from a stale base, not from this change** (low · build-and-ci)
   Lint fails with a single error at .cursor/skills/review/scripts/concern-context.mjs:255:3 (no-unreachable-loop). That file is not among the PR's four changed files, and commit 3073f0d1 ('fix(review): remove unreachable duplicate gate loop', #1769) fixes exactly that line but is not an ancestor of this branch's base. Two CI jobs are red and only the Prettier one belongs to this change.
   Suggested: Merge origin/main to clear the Lint failure; only the Prettier failure needs a change here.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1766
Summary: reject Unicode formatting characters
Verdict: Request Changes
Results: 2 blockers, 4 non-blocking findings, 2 follow-ups

src/validation/unicode-format-characters.ts
ravik453: Addressed the two blocking review findings:

- Ran Prettier on `docs/developer/bugfix-patterns.md` and committed the corrected table formatting.
- Extended the `AGENTS.md` control-character guidance to cover invisible Unicode formatting and bidirectional control code points, including the forbidden set, the escape/construct remedy, and the enforcing test path.

The full Jest suite passes: 515 suites passed, with 8,198 tests passed, 18 skipped, and 5 todo.

I left the non-blocking follow-ups unchanged to keep this PR focused, including the shared binary-file heuristic, category/security framing, forbidden-set reuse, BOM policy, and possible tag-character coverage.

Would you prefer these to be handled in this PR, or tracked as follow-up issues?

Documentation fixes have been pushed in commit f5f17dce.

The remaining blockers appear to be:
- Maintainer approval for the pending GitHub Actions workflows.
- The branch being out of date with the base branch.
- The existing review request for changes.

Please let me know if you would like me to update the branch as well, or if the workflow approvals and branch update should be handled by a maintainer.

moxious (approved):
Thanks for turning this around so quickly. The re-padded table is prettier-clean, and the AGENTS.md rule now names both the enforcing test and the escape-or-construct remedy, which is exactly what was missing last round.

---

---

Both prior blockers are cleared and this is good to merge. The two carried follow-ups — the missing binary-asset arm and the Category value outside the bugfix skill's five — can be tracked separately rather than holding this up.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1766
Summary: reject Unicode formatting characters
Verdict: Approve

PR #1768, opened 2026-09-02
Wire requirements/objectives/skippable through ChallengeBlock

### LINKED ISSUE #1544: convertChallengeBlock drops requirements, objectives and skippable, so they are inert on every challenge block
`JsonChallengeBlockSchema` accepts `requirements`, `objectives` and `skippable` (`src/types/json-guide.schema.ts:575-577`), but `convertChallengeBlock` never maps them into element props:

```ts
// src/docs-retrieval/json-parser.ts:987-999 — 11 props, none of these three
props: {
  title, ...(stepId ? { stepId } : {}), mode, vmTemplate, vmScenario, vmApp,
  setupCommands, setupScript, successCriteria, hintLevels, failureMessage,
},
```

`convertCodeBlockBlock`, immediately below at `:1006-1008`, _does_ map all three — so this reads as an oversight in the challenge converter rather than a design decision. `content-renderer.tsx:1208-1224` forwards only the same 11, and `ChallengeBlockProps` (`src/components/interactive-tutorial/challenge-block.tsx:45-75`) declares no such props.

## Why it matters

The three fields validate, so there is no signal anywhere that they do nothing:

- `"requirements": ["is-admin"]` on a destructive challenge does not gate it — the block renders for everyone.
- `"objectives": [...]` never credits an objective, so guide progress accounting silently under-counts.
- `"skippable": true` renders no skip control, so a learner who cannot complete the challenge is stuck.

`npm run validate:strict` passes, the block editor shows nothing, and the only symptom is an absence.

## Distinct from #1543

#1543 covers `ChallengeBlockForm` dropping these fields (plus `id` and `authorNote`) **on save**. This issue is the converter half: a hand-written JSON guide that never touches the block editor still has inert `requirements` / `objectives` / `skippable`. Fixing #1543 alone would let an author round-trip the fields through the editor while they remain non-functional at runtime.

## Fix

Mirror `convertCodeBlockBlock`: join `requirements` / `objectives` and pass `skippable` through, then thread them into `ChallengeBlock` the way the other step components do.

## Interim

#1538 documents the three fields as accepted-by-the-schema-but-not-honoured, so authors are not misled in the meantime. That note should come out when this lands.

Refs #1530, #1543

Nikshaan:
Fixes #1544

The `convertChallengeBlock` received `requirements`, `objectives`, and `skippable` parameters, but did not pass it down to the component. It is similar to how `convertCodeBlockBlock` → `CodeBlockStep` works: parser → renderer → props → `useStepChecker` wiring for `ChallengeBlock`.

**Decisions from the issue discussion:**
- Skips are done whenever `skippable: true`, whether or not the gating is active (this works like a quiz, unlike CodeBlockStep's "if blocked" rule).
- The `objectives` provide metadata only to `useStepChecker` logic; the `successCriteria` via `checkPostconditions` is the only completion logic.
- The `CHALLENGE_BLOCK_SCHEMA` now has `pausesSectionRun: true` because otherwise section automation would continue past the gated/skippable challenge until an undefined action error.

No longer leaves an interim "accepted but not honoured" comment about this in documentation.

`npm run check` passes locally except `lint:go` (no `mage` binary available in this environment); no `.go` files are touched by this PR.


moxious (changes requested):
Thanks for wiring requirements and skippable through the challenge block. The `pausesSectionRun: true` registry change and the parser plumbing are both right, and they close a real forge-completion hazard.

---

## Merge contract

1. [blocking] **F2 — Gating hides Check my work, Check again and Try again after the VM is provisioned, and there is no Cancel in those states** (high · coda-terminal)
   The buttons after Start are the only exit from a provisioned sandbox VM. Requirements are a precondition for starting a challenge, not for finishing one, but they now gate the finish path too. When eligibility or a requirement regresses mid-challenge the block renders a warning banner and, for the default skippable: false, no button at all — no Check my work, no Try again, no Cancel — so the learner can neither finish nor abandon and the VM stays provisioned with no teardown affordance. When the challenge is skippable the only remaining button is Skip, which takes durable credit rather than resuming.
   Required: Gate only Start challenge on isEnabled. Leave Check my work, Check again and Try again to the block's own state machine, or render them disabled with the checker explanation the way interactive-quiz.tsx does, and render Cancel for any state where a session was provisioned. Add a test for the ready-then-requirements-regress transition, not only setup-failed.

2. [blocking] **F1 — objectives is still inert on challenge blocks while the doc is rewritten to say it works** (medium · guide-schema-and-contracts)
   Two of the three fields in the PR title are wired; objectives is not. The doc previously told authors plainly not to rely on it and now tells them the opposite, so an author who sets objectives on a challenge gets a silent no-op with no error and no warning — the same accepted-but-inert failure mode issue #1544 was filed to end.
   Required: Either route objectives into the checker while keeping successCriteria as the completion gate, or keep the deliberate empty string and restore an accurate doc sentence saying objectives are accepted but not honoured on a challenge, with the reason stated at the call site rather than only in a test name.

3. [blocking] **F3 — The transient checking status blanks every action button with no message, because the banner needs an explanation that SET_ENABLED just cleared** (medium · requirements-manager)
   Hiding on !isEnabled while the banner requires an explanation produces a state with no button and no message — the block is a title and a brief. On a challenge with a fragile requirement the primary action blinks in and out with nothing said, which is the failure mode the quiz's disable-plus-fallback shape exists to avoid.
   Required: Disable rather than hide, and give the banner a fallback string so !isEnabled always explains itself; or derive the gate from checker.isEnabled || checker.isChecking so an in-flight recheck does not blank the UI.

4. [blocking] **F4 — Skip is available during connecting and preparing but never sets the cancellation flag, so setup keeps running behind a solved badge** (medium · coda-terminal)
   Skip is a new exit from a state only Cancel could leave before. Using it mid-provision leaves the setup script executing against a live VM with no remaining way to stop it, and removes the only cooperative teardown affordance — the completion path and the provisioning path stop agreeing about whether work is in flight.
   Required: Have the skip handler set cancelRequestedRef.current (or call handleCancel) before marking the step skipped, and consider hiding Skip while the state is connecting or preparing so Cancel stays the single mid-provision exit.

5. [blocking] **F5 — A skipped challenge renders Challenge solved; the completion reason is available and ignored** (medium · interactive-engine)
   The store, the FSM and the progress event all record the challenge as skipped, but the only surface the learner sees asserts they solved it. That is a factual misstatement in learner-facing UI and it diverges from the skipped-versus-completed distinction every other step type maintains.
   Required: Read reason from useStepCompletion (or checker.completionReason) and render a distinct skipped affordance, matching interactive-step.tsx and datasource-check-step.tsx.

6. [blocking] **F6 — A sequentially blocked challenge still renders Skip, so it can be skipped for durable credit before it is reachable** (medium · requirements-manager · partially reversible)
   Before this PR a challenge had no skip control at all. It now has one that renders even while the step is blocked by a sequential dependency, so a learner can take durable skipped credit for a challenge they were never eligible to attempt, and the record survives reload. The PR description states the skip timing matches Quiz; Quiz at least reads canSkip, and the checker states outright that a sequentially blocked step must not be skippable.
   Required: Hide Skip while the step is blocked by sequential ineligibility — gating on checker.isEnabled is sufficient at this call site. Note that reading checker.canSkip alone would not work today: SET_BLOCKED drops canSkip (step-state.ts:104-113), so createBlockedState's false never reaches the hook's return, which is worth a separate follow-up. Add a test with a blocked checker asserting Skip is hidden; the current mock derives canSkip from skippable, so it cannot catch this.

## Follow-ups

1. [follow_up] **F8 — The skip path revives skipStateUpdate, branching the section completion contract every sibling follows** (medium · guide-schema-and-contracts)
   The skip-completion seam now has two incompatible representations and the new one is the only caller of a parameter the section runner owns. Nothing breaks today because onComplete has no live consumer, but a section whose last step is a skipped challenge would silently stop reporting completion the moment one appears, and the next step type copied from this block inherits the branch.
   Follow-up: Call onStepComplete?.(stepId) without the flag so the skip path matches quiz and interactive-step; the section's manual write is a no-op against the sticky skipped entry markSkipped already made.

2. [follow_up] **F7 — tracked-step-types.mdc still says challenge does not set pausesSectionRun** (low · cross-cutting-architecture)
   That rule file is the agent-facing contract for the step-type registry, and the sentence is now false for one of the four types it enumerates, so the next agent reading it will believe Do Section can still force-complete a challenge.
   Follow-up: Update the sentence in .cursor/rules/tracked-step-types.mdc in this PR so challenge is listed among the types that set pausesSectionRun: true.

## Suggestions

1. [suggestion] **F9 — Already-published challenges change behaviour even with no requirements authored, and neither the doc nor a test says so** (medium · guide-schema-and-contracts)
   A published guide whose challenge sits after other steps previously showed Start challenge immediately and now shows a Complete previous step banner with no action button. That is probably the intended semantics, but it changes already-shipped authored content and neither the doc edit nor any test states it.
   Suggested: State the sequential and disabled gating explicitly in the doc paragraph, since it is not requirements-driven, and add one section-level test that a challenge placed after an incomplete step renders blocked.

2. [suggestion] **F10 — The new tests assert against a fully mocked requirements-manager, and the parser hop and the runner pause are untested** (medium · testing-and-verification)
   The changed semantics are parse-time joining, three renderer props, checker gating and a Do Section pause. The added tests exercise only the gating, and only against a stub of the checker, so they would survive a regression in the plumbing, in the skip invariant, or in the parser.
   Suggested: Add a parser test for a challenge block carrying all three fields, assert the requirements string is forwarded to useStepChecker, drive Skip from canSkip in the mock, and add a runner tripwire that a challenge stops Do Section without being credited.

---

The six blocking items mostly trace back to one seam — a single `isEnabled` flag gating the finish path as well as the start path — so they are likely to fall out together once that is split. Push when you have had a pass at them and we will re-review. Ask here if any of them is unclear; six is a lot to get in one go, and none of it reflects on the direction of the change.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1768
Summary: Wire requirements/objectives/skippable through ChallengeBlock
Verdict: Request Changes
Results: 6 blockers, 2 non-blocking findings, 2 follow-ups

moxious:
Merge blockers - conflicts need resolving, then re-review.

moxious (changes requested):
Thanks for working through round 1 — all six blockers are genuinely fixed, and I checked each at head rather than trusting the vanished code anchor. The finish-path buttons are disabled rather than hidden with Cancel reachable in every provisioned coda state; the requirement banner always says something; Skip is hidden mid-provision and sets the cancellation flag first; a skipped challenge renders "Challenge skipped" and the reason survives the section's follow-on `manual` write because the store makes `skipped` sticky; and `isSequentialBlock` is a real reachable FSM field. `pausesSectionRun: true` is honoured end to end by the section run loop. Both deferred items are closed too.

Four things to fix before this can merge. All four are in code this PR adds.

## 1. Skip never renders for a skippable challenge with no requirements (high)

This is the important one, because it breaks the plain configuration the PR description and the rewritten doc both promise.

`hasResolvedOnce` only becomes true after a committed render where `checker.isChecking` was true. On the no-requirements path the real `useStepChecker` never commits one. The block always passes `objectives: ''`, so `hasObjectives` is false; with no `requirements`, `decideStepCheck` returns `enabled` on its last line. That branch runs `safeDispatch({ type: 'START_CHECK' })`, then `decideStepCheck` synchronously, then falls straight through to `createEnabledState` and the enabled dispatch — every `await` in `checkStep` sits inside the `check-objectives` or `check-requirements` branches, and neither is taken. React batches both reducer dispatches into one commit, so the status goes idle -> enabled without ever committing as checking. `wasCheckingRef` never flips, `hasResolvedOnce` stays false, and the Skip gate has it as a hard conjunct with no alternate branch.

So `skippable: true` with no `requirements` renders no Skip control. Because this PR also sets `pausesSectionRun: true`, Do Section now stops at that challenge as well, which leaves a learner who cannot solve it with no exit at all.

Worth calling out why the suite is green: every skip test hand-drives `isChecking: true` through the mocked requirements-manager barrel, so it is asserting a state the real hook cannot produce.

Required: derive the gate from checker state at render time instead of latching on an observed `isChecking` edge — `!checker.isChecking && !checker.isSequentialBlock` is enough at this call site, or expose the FSM status from `useStepChecker` if you need to distinguish an unchecked idle state from a resolved one. Then add one test that drives the real `useStepChecker` for `skippable: true` with no `requirements` and asserts Skip renders. While you are in the doc paragraph, it currently attributes gating only to `requirements`; the block now also honours `isEligibleForChecking`, so a published guide whose challenge sits after any other step shows a "Complete previous step" banner where it previously showed "Start challenge". That behaviour change is worth stating explicitly.

## 2. The objectives probe accepts on `result.pass`, which fails open (medium)

The probe sets `objectivesMet` from `result.pass` of `checkRequirements`, which evaluates in `pre` mode. In that mode `routeUnifiedCheck` returns `pass: true` with `verdict: 'invalid'` for any token that fails `isValidRequirement`, ends with a colon, or is bare `var-`. `isParameterizedRequirement` is a `startsWith` test and `ObjectiveTokenSchema` is deliberately permissive, so a bare prefix such as `has-role:` or `coda-exit-zero:` survives `executableObjectives` and reaches the probe — and the block then tells the learner the objective is already met for a condition that was never evaluated.

`useStepChecker` is the established owner of this question and uses `checkVerdict(result) === 'satisfied'`, with a comment giving exactly this reason. This does not reach completion today, but it leaves two runtime evaluators of objectives with different acceptance rules, and the isolation is invisible to the type system because it rests on a hard-coded `objectives: ''`.

Required: use `checkVerdict(result) === 'satisfied'` instead of `result.pass`, and set `objectivesMet` from that verdict on every resolution so the note also clears when the objective stops holding. Please also record the challenge-block branch in the `guide-schema-and-contracts` anchor in `docs/design/CONCERN_DETAILS.md` in this PR: objectives on a challenge are informational only, `successCriteria` alone completes it, and the satisfied-only rule binds both evaluators.

## 3. A `coda-exit-zero:` objective execs at mount against whatever session is live (medium)

The probe effect has dependencies `[objectives, stepId, checkRequirements]` and no gate on `isEligibleForChecking`, `checker.isEnabled`, `mode`, or the block's own state. `coda-exit-zero:` is a first-class requirement prefix, so it survives `executableObjectives` and routes to `codaExitZeroCheck`, which resolves the session from the module-level `getTerminalSessionId()` and execs there. That contradicts the block's own stated invariant that every exec is pinned to the session the caller resolved.

The preconditions are reachable in an ordinary guide: an earlier challenge or a terminal-connect step leaves one shared connected session, and any completed setup has already written the ready file the exec is gated on. An author-authored command then runs in that VM before the learner starts anything, including for a step they are not yet eligible to attempt.

Required: gate the probe on the block owning a live session, or on `checker.isEnabled` plus a coda-mode ready state, or exclude `coda-exit-zero:` tokens from the informational probe entirely.

## 4. Cancel and Skip disconnect the shared terminal session (low, but introduced here)

`handleCancel` gains a branch calling `terminalCtx?.disconnect()` for coda-mode `ready`, `checking`, `failed-check` and `setup-failed`. At base, Cancel existed only in `connecting` and `preparing` and never referenced the terminal context.

The terminal is a process-wide singleton: `disconnect()` clears the active VM options, resolves every pending session waiter with null, closes the live session, and leaves `getTerminalSessionId()` returning null for every other step. `openTerminal` reuses the live session when VM options match, so a challenge sitting in `ready` can be on a session a terminal step or a sibling challenge opened. The Skip handler calls `handleCancel()` first, so the new skip control tears it down too. The sibling standard-mode branch states the opposite invariant — never tear down a session another step may be using — in the same function.

Required: disconnect only when the block provisioned the session itself. Track the session id `runSetup` pinned to and compare it against `staleSessionIdRef` before calling `disconnect()`, and have Skip stop short of the disconnect when the session is not the block's own.

## Two small things while you are in there

- `JsonChallengeBlockSchema.objectives` still carries `.describe('Conditions that automatically complete this block')`. The `pathfinder schema` command exports that as the JSON Schema description an author or authoring agent reads, so it is now the only surface still stating the old contract after this PR corrected the prose. One string.
- `stepIndex` and `totalSteps` are newly destructured and never read, and `runSetup` declares its own `totalSteps` that shadows the prop. The widened `onStepComplete` signature carrying `skipStateUpdate` is dead the same way — the only call site passes the step id alone. Nothing in the toolchain reports any of it.

---

The shape of this change is right, and the coda and completion semantics you have wired up are the hard part. Push those fixes and I will re-review promptly.

Nikshaan:
Fixed all four. 

Skip now checks gates on checker.status !== 'idle' rather than an observed isChecking transition, which was actually the bug. The no-requirements path returns synchronously in one React commit and never paints the intermediate checking state, so the latch never fired. The objectives probe now checks checkVerdict(result) === 'satisfied' rather than result.pass, excludes coda-exit-zero: tokens (the only side-effecting requirement prefix confirmed in the codebase), and checks for session ownership/eligibility, so it cannot exec against a live shared session. Cancel/Skip now check session ownership before disconnect(), so that they cannot teardown a terminal another step opened. Also removed the unused stepIndex/totalSteps destructuring, totalSteps shadow in runSetup, and the unused skipStateUpdate argument on onStepComplete, and fixed the schema objectives description to match the fixed contract.

src/components/interactive-tutorial/challenge-block.tsx
```diff
@@ -594,31 +751,63 @@ export const ChallengeBlock: React.FC<ChallengeBlockProps> = ({
       )}
 
       <div className={styles.actions}>
-        {state === 'idle' && !configGateMessage && (
+        {state === 'idle' && !configGateMessage && isEnabled && (
```
moxious: B1 (blocking): gating Start on `isEnabled` makes a `coda-exit-zero:` requirement unsatisfiable — that check needs `/tmp/pathfinder-ready`, which only this block's own `runSetup` writes, and Start is the only way to reach `runSetup`. With `skippable: false` there is no other control in `idle`, so the block is permanently dead. Either reject `coda-exit-zero:` in a challenge's `requirements` in `validateBlockConditions`, or keep Start reachable and gate only Check my work.

src/components/interactive-tutorial/challenge-block.tsx
```diff
@@ -225,10 +281,60 @@ export const ChallengeBlock: React.FC<ChallengeBlockProps> = ({
...
   const isStandalone = !onStepComplete;
   const isCompleted = storedCompleted || state === 'solved';
+  const isSkipped = storedReason === 'skipped';
+
+  // Objectives are probed read-only here, never through the gating checker:
+  // a satisfied objectives string on useStepChecker would SET_COMPLETED and
+  // persist to this block's store key, solving the challenge without successCriteria.
+  const [objectivesMet, setObjectivesMet] = useState(false);
+  useEffect(() => {
+    const safeObjectives = conditionTokens(objectives).filter((token) => !token.startsWith('coda-exit-zero:'));
```
moxious, line 304: F2 (suggestion): this filter needs its one-line reason — it is the only thing keeping the informational probe from exec'ing author commands in the learner's VM. Note also that after filtering, a mixed `objectives` array can still render "Objective already met" on the strength of the surviving tokens alone.

src/components/interactive-tutorial/challenge-block.tsx
```diff
@@ -594,31 +751,63 @@ export const ChallengeBlock: React.FC<ChallengeBlockProps> = ({
...
+            state === 'checking' ||
+            state === 'failed-check' ||
+            state === 'setup-failed') && (
+            <Button variant="secondary" icon="times" onClick={handleCancel}>
+              Cancel
+            </Button>
+          )}
+        {skippable &&
+          !isCompleted &&
+          checker.status !== 'idle' &&
```
moxious, line 816: F1 (follow-up): `checker.status !== 'idle' && !checker.isSequentialBlock` is stricter than decision 3 on #1544 ("Skip is available whenever `skippable: true` ... not gated on ... unmet prerequisite") and than quiz, which shows Skip while blocked because no reducer case clears `canSkip`. This gate is also the only consumer of the new `isSequentialBlock` field on `StepState` / `toLegacyState` / `UseStepCheckerReturn`.

src/components/interactive-tutorial/challenge-block.tsx
```diff
@@ -594,31 +751,63 @@ export const ChallengeBlock: React.FC<ChallengeBlockProps> = ({
...
-          </Button>
-        )}
+        {mode === 'coda' &&
+          (state === 'connecting' ||
+            state === 'preparing' ||
+            state === 'ready' ||
+            state === 'checking' ||
+            state === 'failed-check' ||
+            state === 'setup-failed') && (
+            <Button variant="secondary" icon="times" onClick={handleCancel}>
```
moxious, line 810: F3 (follow-up): Cancel now renders through `ready`/`checking`/`failed-check` too and calls `terminalCtx.disconnect()`, which closes the session and leaves the VM holding its quota slot; `resetToIdle` clears `setupStartedRef`, so Start re-provisions from scratch. One unconfirmed click on a button labelled Cancel discards the learner's environment. The tests that need this (`:1302`, `:1342`) are the blocked-after-ready cases only.

moxious (changes requested):
Thanks for wiring `requirements`, `objectives`, and `skippable` through `ChallengeBlock` — a challenge block was the one tracked step type that ignored the gating every other step honours, and closing that is real work. The `pausesSectionRun` addition is the right call too, and the `.mdc` note you updated alongside it is accurate.

---

## Merge contract

1. [blocking] **B1 — A coda challenge with a `coda-exit-zero:` requirement can never be started** (medium · coda-terminal)
   `convertChallengeBlock` now forwards `requirements`, and `challenge-block.tsx` gates the whole block on the checker: Start challenge renders only when `state === 'idle' && !configGateMessage && isEnabled` (`challenge-block.tsx:754`, `isEnabled` at `:256`), coda mode starts at `idle` (`:262`), Cancel deliberately excludes `idle` (`:774-784`), and Skip needs `skippable: true` (`:785`, default `false` at `:219`). `coda-exit-zero:<cmd>` is a valid `RequirementTokenSchema` token, and nothing downgrades it on the way in — `executableObjectives` filters objectives only, and `validateInteractiveRequirements` checks `exists-reftarget` only. But `codaExitZeroCheck` execs with `readyFile: PATHFINDER_READY_FILE` (`src/requirements-manager/checks/coda.ts:60`) and that sentinel is written only by the challenge's own `SENTINEL_WRITE_COMMAND` (`challenge-block.tsx:49`), which runs only inside `runSetup` (`:454`), so the requirement returns `pass: false` even from a manually connected terminal panel. The result is a non-skippable challenge that renders only the requirement banner — no Start, no Cancel, no Skip, no fix affordance — under the unactionable fallback text "Requirements not met. Please check the page state and try again." At base `convertChallengeBlock` forwarded no requirements, so this is newly reachable. The same ordering hazard is already handled on the informational path: the objectives probe filters exactly this prefix at `:296`. The related `is-terminal-active` case is not affected — `TerminalPanel` mounts independently (`docs-panel.tsx:1405`) and `step-checker.hook.ts:1063-1084` rechecks on terminal status change, so that token recovers.
   Required: Refuse the impossible authoring instead of rendering a dead block. `validateBlockConditions` (`src/validation/condition-validator.ts`) already visits a challenge's `requirements`, so reject `coda-exit-zero:` there for challenge blocks, since only the challenge's own setup can satisfy it. If the token must stay authorable, filter it out of the gating requirements the way the objectives probe already does, or keep Start reachable in coda mode and gate only Check my work. Add a test for a coda challenge whose requirement can only pass after Start.

## Follow-ups

1. [follow_up] **F1 — Skip visibility adds two gates the recorded decision rules out, and one of them is the only reason `UseStepCheckerReturn` grows new fields** (medium · requirements-manager)
   Decision 3 recorded on #1544 says "Skip is available whenever `skippable: true`, matching Quiz's existing behavior. Not gated on a prior failure or unmet prerequisite", and the PR body repeats it. The code renders Skip only when `skippable && !isCompleted && checker.status !== 'idle' && !checker.isSequentialBlock && state !== 'connecting' && state !== 'preparing'` (`challenge-block.tsx:785-790`). Quiz renders Skip on `canSkip && !isCompleted` (`interactive-quiz.tsx:569-585`), and no reducer case in `step-state.ts` touches `canSkip`, so quiz does keep Skip visible while sequentially blocked — quiz parity is not what this ships. The `!checker.isSequentialBlock` gate is also the sole consumer of new shared surface: `isSequentialBlock` is added to `StepState`, to every reducer case, to `toLegacyState` and to the public `UseStepCheckerReturn`, alongside `status`, a `StepStatus` relocation into `src/types/requirements.types.ts`, and a widened `updateManager` signature. Neither the PR body nor the challenge section of `json-guide-format.md` mentions that Skip is withheld while blocked or while the first check is unresolved.
   Follow-up: Pick one and make the record agree with the code. Either drop the `checker.status !== 'idle'` and `!checker.isSequentialBlock` gates so Skip matches quiz as decided, which also removes the `isSequentialBlock` plumbing from `step-state.ts`, `hooks.types.ts` and `step-checker.hook.ts`; or keep the gates, get decision 3 on #1544 amended, and state the behavior in the PR body, in the challenge section of `json-guide-format.md`, and in the `guide-schema-and-contracts` anchor.

2. [follow_up] **F3 — Cancel now sits beside Check my work through a healthy coda challenge and discards the sandbox session** (medium · coda-terminal)
   At base, Cancel rendered only for `connecting` and `preparing`, and `handleCancel` handled only those two states. It now renders for coda mode in `connecting`, `preparing`, `ready`, `checking`, `failed-check` and `setup-failed` (`challenge-block.tsx:774-784`) — the whole life of a provisioned, working challenge — and the new branch calls `terminalCtx.disconnect()` when this block owns the live session (`:640-649`). `cleanup()` in `useTerminalLive.hook.ts:135-155` closes the session and clears `currentVmIdRef`, with the comment that this "releases the terminal only — the VM keeps its quota slot until expiry", and `resetToIdle` clears `setupStartedRef`, so Start re-provisions and re-runs setup rather than returning the learner to their environment. One unconfirmed click on a button labelled Cancel therefore throws away the learner's work in the VM and burns its quota slot. The tests that motivate the wider render are the blocked-after-ready cases (`challenge-block.test.tsx:1302` and `:1342`) where the primary button is disabled, which is a much narrower condition than "always".
   Follow-up: Render Cancel only where the primary action is unavailable — `connecting`, `preparing`, `setup-failed`, and `ready`/`failed-check` while `!isEnabled` — or keep it always and give it a label that says what it does plus a confirmation, e.g. "End challenge". Note the behavior in the challenge section of `json-guide-format.md` either way.

## Suggestions

1. [suggestion] **F2 — The objectives note can claim an objective is met after silently dropping the sandbox objectives, and the drop has no stated reason** (low · coda-terminal)
   The probe filters its input with `conditionTokens(objectives).filter((token) => !token.startsWith('coda-exit-zero:'))` (`challenge-block.tsx:296`) and no comment says why, then renders the definite statement "Objective already met — Check my work is still required to finish the challenge." (`:714-719`) when the survivors all come back `satisfied`. `challenge-block.test.tsx:1161` pins the mixed case, so `objectives: ["coda-exit-zero:...", "has-datasource:prom"]` shows the note on the strength of the datasource alone. The coda branch of `isReadyForProbe` (`:301-306`) additionally waits for `status === 'connected'` and session ownership, which is close to redundant once the session-dependent tokens are already gone, and it suppresses the note for ordinary Grafana objectives until the VM is provisioned.
   Suggested: Show the note only when no token was filtered out, or word it as "some objectives already met". Add the one-line reason the filter exists — that probing a `coda-exit-zero:` objective would exec author-supplied commands in the learner's VM outside the Check my work flow — then reconsider whether the coda session-ownership branch of `isReadyForProbe` still earns its complexity.



---

Next step: B1 is the only item standing between this and merge — the two follow-ups and the suggestion do not need to land here, and I am happy to file them separately. CI is green on this head (the two `publish-report` / Bundle Stats failures are a known fork-permission limitation on our side, not yours). Push a fix and CI will rerun; ping me and I will re-review.

Nikshaan:
Fixed B1. Start challenge no longer gates on isEnabled, because Start challenge itself sets up conditions rather than checking that they've been met. Requirement banner still shows along with it so that a learner can see an unfulfilled precondition before provisioning starts. Check my work/Check again/Try again still gate on isEnabled as previously. Also added a validateBlockConditions rule that rejects coda-exit-zero: requirement for a challenge (objectives still accepts it, is unchanged). Added the comment to explain the objective probe filter per F2.

src/components/interactive-tutorial/challenge-block.tsx
```diff
@@ -507,10 +646,24 @@ export const ChallengeBlock: React.FC<ChallengeBlockProps> = ({
       // No setup is in flight yet — bail immediately. (Setup will see the
       // flag if it starts after this and skip out before running anything.)
       resetToIdle();
+    } else if (
+      mode === 'coda' &&
+      (state === 'ready' || state === 'checking' || state === 'failed-check' || state === 'setup-failed')
+    ) {
+      const liveSessionId = terminalCtx?.sessionId;
+      const isOwner = Boolean(provisionedSessionIdRef.current) && provisionedSessionIdRef.current === liveSessionId;
```
moxious, line 672: When `openTerminal` reuses an existing session with matching VM options, `runSetup` still records that ID in `provisionedSessionIdRef`. This equality therefore proves the challenge used the live session, not that it created it. Cancel—and Skip through `handleCancel()`—then disconnects the terminal another step opened. Please distinguish a newly provisioned session from one already present at Start, and cover the same-options reuse case. This is the remaining F15 ownership issue.

moxious (changes requested):
Thanks for making Start reachable for setup-dependent requirements and adding the regression coverage. That resolves B1. The earlier shared-session ownership finding still needs a fix before merge; the other items below remain follow-ups. Please cover reuse of an already connected session when fixing it.

CI Gate, tests, builds, and the E2E matrix passed. The two reporting failures are the known fork-permission limitation and require no change from you. This follow-up was verified by reading the current code and CI results; no PR code was run locally.

Blockers:

1. [blocking] **F15 — Cancel and Skip still disconnect a session reused from another step** (medium · coda-terminal)
   The ownership fix still misses session reuse. With a terminal already connected using the same VM options, openTerminal returns that existing session (TerminalContext.tsx:249-250). runSetup records it as provisioned (challenge-block.tsx:401), so the comparison at :654 treats it as owned and Cancel disconnects it at :656. Skip calls the same handler at :806. This interrupts the terminal another step opened. The existing tests cover a changed live session ID, but not reuse of the same ID. This is the unresolved shared-session finding from my earlier review.
   Required: Track whether this Start actually provisioned the session, and do not disconnect a session that already existed at Start. Cover same-options session reuse, not only the case where another step later changes the live session ID.

Follow-ups:

1. [follow_up] **F3 — Cancel still ends a healthy challenge session without explaining the effect** (medium · coda-terminal)
   The previously noted Cancel behavior is unchanged: it remains beside Check my work during a healthy challenge (challenge-block.tsx:784-793), and closes the session and resets setup at :648-658. The label does not explain that effect. Closing the terminal is established; deleting the underlying VM is not.
   Follow-up: Limit Cancel to states where an exit is needed because the main action is unavailable, or label the action to explain that it ends the challenge and confirm the session-ending action.

2. [follow_up] **CODA-ASYNC-1 — Restarting after Cancel revives the previous asynchronous success check** (medium · correctness-and-reliability)
   Check my work → Cancel → Start can accept the previous check response. Cancel sets the shared flag at challenge-block.tsx:644, but Start clears it at :579. The old successful response then passes :615 and marks the restarted challenge complete at :620, even before its new work is checked. The cancellation test at challenge-block.test.tsx:454 covers staying idle, not restarting before the response arrives.
   Follow-up: Give each challenge run/check a generation token and ignore results belonging to a cancelled or reset generation. Cover Cancel followed by Start before the old check resolves.

3. [follow_up] **SCHEMA-DOC-START-GATE — Challenge authoring docs still describe a Start gate that the latest change removed** (low · guide-schema-and-contracts)
   The latest fix deliberately leaves Start reachable before prerequisites or sequential eligibility pass (challenge-block.tsx:764-765; section test:103-127), but json-guide-format.md:1197 still describes those gates as preventing execution and replacing Start. Check my work, rather than Start, remains gated.
   Follow-up: Update the challenge format paragraph to say that Start remains available before prerequisites and sequential eligibility are satisfied, and that these gates apply to Check my work.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1768
Summary: Wire requirements/objectives/skippable through ChallengeBlock
Verdict: Request Changes
Results: 1 blocker, 0 non-blocking findings, 3 follow-ups

Nikshaan:
Fixed. provisionedSessionIdRef will be set only if openTerminal actually returns a session which is different from that which was alive before Start was clicked. In no way will the reused session (having the same VM options) be assigned, hence Cancel/Skip will not cause its disconnection. Regression test for the reuse scenario has been added to the existing ownership test suite.

moxious (changes requested):
First, an apology. This is the fifth round on a PR whose production diff is under 300 lines, and that is on the review side, not yours. Looking back at the session-teardown finding, I asked for a narrow fix three times instead of stating the rule once: round 2 said "track the session id runSetup pinned to", round 12 said "do not disconnect a session that already existed at Start". You implemented both correctly. Neither instruction described the whole invariant, so each fix exposed the next uncovered path of the same rule. That is a specification failure on my part and it cost you three rounds.

So this round states the rule, not a patch, and it is the only thing left.

## The invariant

A challenge block may tear down a sandbox session only when no other block holds it. That has to hold in both directions:

- you did not provision the session, so it is not yours to close — fixed at this head, with a regression test
- you did provision it, but another block has since joined it, so it is no longer yours alone — still open

`handleCancel` gates teardown on `provisionedSessionIdRef.current === liveSessionId`, which answers "did I provision this id" and never "is anyone else using it". There is no refcount or lease in `src/integrations/coda`, and `openTerminal` returns the live session whenever template, app and scenario compare equal — each coerced with `|| ''` — so two challenges with no VM options share one session.

The path: challenge A provisions S1 and owns it. Challenge B reuses S1, correctly declines ownership under your new guard, and runs its setup inside S1. The learner navigates back to A and clicks Cancel, or Skip, which calls `handleCancel` first. A is still the owner, so `disconnect()` resolves through `useTerminalLive` cleanup to `session.close()` and `DELETE /sessions/{id}`. B is left on a dead session: its objectives probe goes quiet because it requires `terminalCtx.sessionId === setupSessionIdRef.current`, and Check my work fails with `terminal_disconnected`. The VM keeps its quota slot, so in-VM work survives — this is interruption, not loss. Recovery is Cancel then Start, which re-runs B's whole setup.

## What has to change

Either one is fine:

- **Simplest:** stop disconnecting on Cancel and Skip in the post-provision states — ready, checking, failed-check, setup-failed. Let those reset without teardown, and keep teardown only for the connecting and preparing window this block provisions for itself. This is a deletion, and it also resolves the standing Cancel-beside-Check-my-work item.
- **Or:** make session holding refcounted in `TerminalContext` so `disconnect` becomes a release.

Plus one test: put a second block on the reused session, cancel the provisioning block, assert the session survives.

## Commitment

Fix that, and this merges. I will not add a new blocker on the next round. Anything else I find — including the four items below — gets filed as a follow-up issue against the merged code, not held over this PR:

- challenge-arm coverage in `content-renderer.prop-forwarding.test.ts`, and a challenge case in the `pausesSectionRun` tripwire
- the `coda-exit-zero:` remedy landing as a warning outside `--strict`, and pointing at `objectives`, which filters the token out
- the cross-block session-ownership protocol being recorded only in test titles
- the appended contract-anchor clause overstating what completes a challenge

None of those is harm this PR introduces, so none of them should be your problem before merge.

## Why this is worth merging

Two things here fix real defects in shipped behaviour, which is why I would rather land it than keep iterating. `pausesSectionRun: true` stops a forged completion: on main, a challenge inside a section run gets `markStepCompleted(stepId, sectionId, 'manual')` written durably without the learner doing anything, and that fake record feeds `interactive-section-completed`, `sectionDoneStorage`, guide progress and downstream `section-completed:` gating. And honouring `requirements`, `objectives` and `skippable` closes #1544, where those author fields were accepted and silently ignored.

## CI

CI Gate, Lint, Lint backend, Typecheck, Test, Test backend, Test scripts, Prettier, the E2E matrix, and the bundled-guide and package validation all pass on this head. The two red checks are `compare` and `publish-report`, both failing only at "Creating PR comment... Resource not accessible by integration" — the known fork-permission limit, not something for you to fix; the bundle comparison itself succeeded. This review was done by reading code and blobs at this head; no PR code was run locally.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1768
Summary: Wire requirements/objectives/skippable through ChallengeBlock
Verdict: Request Changes
Results: 1 blocker, 0 non-blocking findings, 4 follow-ups (all to be filed as issues)

moxious (dismissed):
Thanks for landing the session-teardown fix exactly as scoped last round: gating Cancel/Skip's disconnect to the connecting/preparing window closes F15's provisioner-side hazard, and it's cleared below alongside the reuse-side fix from two rounds back.

While verifying that fix, I found a new gap: the refcounted holder-tracking this fix introduced (`activeSessionHolders`/`claimSession`/`hasOtherSessionHolders`) has no liveness guard in `handleStart`'s post-`await openTerminal()` continuation, so an unmount during provisioning (without clicking Cancel) can register a phantom session holder. That finding is real and still stands as written below — I'm not softening or dropping it.

Here's the call on it: I'm approving and merging this now rather than holding for another round, as a deliberate trade. F15 was the broader hazard — Cancel/Skip on one learner's challenge block could kill a *different* learner's live session outright, and that's now closed with a regression test. F17 is narrower: it leaks one Coda sandbox VM on a specific navigate-away-during-provisioning path, recoverable once the VM's own lifecycle reclaims it, and it doesn't touch another learner's live work. Closing the broad hazard now is worth more than holding the PR open for the narrow one. F17 is tracked as a follow-up issue rather than held over this PR: https://github.com/grafana/grafana-pathfinder-app/issues/1896.

This also keeps last round's commitment intact: I said F15 would be the last blocker I'd add, and this PR is not being held up by F17 — it's resolved by trade-and-track, not by another review round.

Follow-ups:

1. [follow_up] **F17 — handleStart's post-openTerminal continuation has no liveness guard, so an unmounted instance can register a phantom session holder** (medium · coda-terminal · partially reversible)
   handleStart's only guard after `await terminalCtx.openTerminal(vmOpts)` is `cancelRequestedRef.current`, which only Cancel sets. If a ChallengeBlock unmounts while openTerminal() is still pending — the learner navigates away before a slow VM provision resolves, without clicking Cancel first — the unmount cleanup effect only runs `releaseSession()`; it does not stop the pending continuation. When that continuation resolves, it unconditionally calls `claimSession(nextSessionId)` and `runSetup(nextSessionId)`, re-registering the dead instanceId in the module-level `activeSessionHolders` map. `hasOtherSessionHolders`/`hasOtherConnectingInstances` — the exact functions this round's fix added so `handleCancel` can tell whether another block still needs a shared session — then see this phantom holder and can make a later, live block's legitimate Cancel/Skip silently skip its `disconnect()` call, leaking the Coda sandbox VM. No test in challenge-block.test.tsx exercises unmount-while-provisioning; the file's only `unmount()` calls cover an unrelated counter-reset scenario.
   Follow-up: Track liveness (a mounted ref, or an AbortController wired through openTerminal) and check it in handleStart's post-await continuation before calling claimSession/runSetup; on unmount, if this instance was still awaiting openTerminal(), also run the same hasOtherSessionHolders/hasOtherConnectingInstances-gated disconnect that handleCancel performs, instead of only releasing local bookkeeping. Tracked at https://github.com/grafana/grafana-pathfinder-app/issues/1896.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1768
Summary: Wire requirements/objectives/skippable through ChallengeBlock
Verdict: Approve with Minor

moxious:
(facepalm) -- this now needs merge conflicts reconciled.  Sorry.

Nikshaan:
> (facepalm) -- this now needs merge conflicts reconciled. Sorry.

No worries! Resolving this right now.

moxious (approved):
Thanks for sticking with it, appreciate the contribution!

PR #1770, opened 2026-09-04
fix(telemetry): pause replay when Pathfinder closes

d0ugal:
## Summary
- Start Pathfinder session replay only after an explicit surface open.
- Pause recording five seconds after the panel closes and resume immediately when reopened.
- Cover the replay lifecycle across Pathfinder surfaces.

## Test plan
- [x] Focused Faro and replay lifecycle tests
- [x] ESLint
- [ ] CI


src/lib/telemetry/faro-adapter.ts
```diff
@@ -159,17 +166,46 @@ function startSessionReplayOnFirstOpen(faro: Faro, samplingRate?: number): void
...
+      }, REPLAY_CLOSE_PAUSE_DELAY_MS);
+    }
+  };
   const start = () => {
-    if (started || activating || attempts >= MAX_REPLAY_ACTIVATION_ATTEMPTS || !isPathfinderOpen()) {
+    if (
+      started ||
+      activating ||
+      attempts >= MAX_REPLAY_ACTIVATION_ATTEMPTS ||
+      !hasReportedPathfinderSurface() ||
```
moxious, line 217: F-1770-1 (blocking): this clause makes activation depend on an explicit `reportPathfinderSurface` call, and no boot path makes one for floating or fullscreen. `module.tsx:338-340` mounts `FloatingPanelManager` when `panelModeManager.getMode() === 'floating'`, and `FloatingPanelManager.tsx:38` seeds its state from the same read, so on a reload the panel is visibly open while `hasReportedPathfinderSurface()` is false. The only reporter for those modes is `panel-mode.ts:232` inside `applyModeChange`, which is not invoked at boot and returns early on `mode === previous`; `FullScreenPanel.tsx:71-73` skips `setModeTransient` for the same reason. Before this PR the cold `readPathfinderSurface()` read of PANEL_MODE started the recording here — the deleted test 'records straight away when a surface was already open at init' was asserting exactly that. Suggest reporting the surface from whatever actually opens the panel at boot, and covering it with a reload test.

src/lib/telemetry/replay.ts
```diff
@@ -103,6 +108,16 @@ class ContinuousReplayInstrumentation extends ReplayInstrumentation {
       this.metas = metas;
     }
   }
+
+  public pauseRecordingForSurface(): void {
```
moxious, line 113: F-1770-2 (blocking): the pause these methods drive is undone by the SDK itself. `defaultReplayInstrumentationOptions.inactivityThresholdMs` is `6e4` and `buildReplayOptions` above never overrides it, so `startRecording()` calls `setupInactivityTracking()`, which attaches capture-phase `pointermove`/`pointerdown`/`scroll`/`keydown`/`input` listeners on `document` whose handler calls `resumeRecording()` whenever `isPaused`. `pauseRecording()` clears only the inactivity timer — `teardownInactivityTracking()` runs only from `stopRecording()` — so those listeners survive the pause and the first mouse move anywhere in Grafana resumes recording. Confirmed against the locked 2.11.0, not just the installed 2.10.0. Pass `inactivityThresholdMs: 0` in `buildReplayOptions` so the surface lifecycle is the sole pause authority.

F-1770-4 (follow-up): `pauseRecording`/`resumeRecording` are declared `private` in the SDK's `.d.ts`, and these casts have no runtime feature detection. Combined with the unguarded `setTimeout` at faro-adapter.ts:182-187, an SDK rename becomes an uncaught TypeError in the host page.

src/lib/telemetry/faro-adapter.ts
```diff
@@ -159,17 +166,46 @@ function startSessionReplayOnFirstOpen(faro: Faro, samplingRate?: number): void
...
+      if (pauseTimer !== undefined) {
+        clearTimeout(pauseTimer);
+        pauseTimer = undefined;
+      }
+      controller?.resume();
+      start();
+      return;
+    }
+    if (controller !== null && pauseTimer === undefined) {
+      pauseTimer = setTimeout(() => {
```
moxious: F-1770-4 (follow-up): `controller?.pause()` runs in a bare `setTimeout` callback with no try/catch and outside `guardTelemetry`, which this file uses at 14 other sites and whose invariant is stated at line 39. The optional chain guards a null controller, not a missing method, and the surface listener's try/catch cannot reach a callback on its own macrotask. Wrapping both controller calls in `guardTelemetry` and feature-detecting the methods would make a missing one degrade to no pausing.

src/lib/telemetry/faro-adapter.ts
```diff
@@ -185,8 +221,9 @@ function startSessionReplayOnFirstOpen(faro: Faro, samplingRate?: number): void
         pushFaroEvent(TELEMETRY_EVENTS.sessionReplayActivationFailed, { exhausted: String(exhausted) });
       });
   };
+  onPathfinderSurfaceChange(handleSurfaceChange);
   const unsubscribe = onPathfinderSurfaceChange(start);
```
moxious: F-1770-6 (suggestion): `handleSurfaceChange` already calls `start()`, so this second subscription makes every non-closed transition enter `start` twice — harmless, since `activating` is set synchronously before the import. The part worth fixing is that `unsubscribe()` now removes only this listener while `start` stays reachable through `handleSurfaceChange` (which must stay subscribed for pause/resume), so the bound described above `MAX_REPLAY_ACTIVATION_ATTEMPTS` rests solely on the attempts counter. Dropping this subscription and the `unsubscribe` indirection would make the code say what it does.

src/lib/faro.test.ts
```diff
@@ -1531,14 +1533,14 @@ describe('session replay activation', () => {
     expect(mockInstrumentationsAdd).toHaveBeenCalledTimes(1);
   });
 
-  it('records straight away when a surface was already open at init', async () => {
+  it('does not record from a persisted surface mode before it opens', async () => {
```
moxious, line 1559: F-1770-3 (follow-up): this rename encodes the premise that a persisted panel mode is not an open surface. For floating and fullscreen that premise does not hold — see F-1770-1. Separately, no test in this file covers the behaviour the PR adds: `settleReplayImport` (line 1488) is a real 0ms timeout and no replay test uses fake timers, so the 5s pause can never fire, and tests that report `closed` without reopening leave a live real timer holding the module-level controller past the end of the test. The mock's public `pauseRecording`/`resumeRecording` no-ops also mean the private-method cast in replay.ts is never checked against the real shape. Worth adding fake-timer tests for pause-after-close, cancel-on-reopen, resume-on-reopen and close-during-import, and revisiting the 'keeps recording after Pathfinder is closed again' test at line 1558, whose name now states the opposite of this PR's intent.

moxious (commented):
Thanks for spotting that session replay keeps recording after Pathfinder is closed — that is a real privacy-shaped gap rather than a cosmetic one, and `hasReportedPathfinderSurface` is the right seam to fix it at.

---

## Merge contract

1. [blocking] **F-1770-1 — Replay never starts when a reload lands straight in a persisted floating or fullscreen panel** (high · analytics-and-telemetry)
   The new `!hasReportedPathfinderSurface()` clause in `start()` (faro-adapter.ts:195) makes activation depend on an explicit `reportPathfinderSurface` call. No boot path makes that call for floating or fullscreen. `module.tsx:338-340` mounts `FloatingPanelManager` when `panelModeManager.getMode() === 'floating'`, and `FloatingPanelManager.tsx:38` seeds its state from the same read, so the panel renders visibly open with no user action. The only floating/fullscreen reporter is `panel-mode.ts:232` inside `applyModeChange`, which is not invoked at boot and returns early when `mode === previous`; `FullScreenPanel.tsx:71-73` skips `setModeTransient` for the same reason on a fullscreen reload. Before this PR, `isPathfinderOpen()` fell through to `readPathfinderSurface()`, which reads PANEL_MODE from localStorage, and recording started. The PR deletes the test that asserted exactly this ('records straight away when a surface was already open at init'). Net effect: for a user who reloads and keeps working in the floating or fullscreen panel, session replay is off for that page load, while the activity gate is latched open so everything else still reports. It fails silently — no error, no event.
   Required: Report the surface from the boot path that actually opens the panel — for example have `FloatingPanelManager` call `reportPathfinderSurface(mode)` on mount for floating and fullscreen — or keep the cold read as an activation trigger for those two modes. Cover it with a test that reloads with PANEL_MODE persisted and no explicit report.

2. [blocking] **F-1770-2 — The SDK's own inactivity tracker resumes recording on the first pointer move after the pause** (high · analytics-and-telemetry)
   `defaultReplayInstrumentationOptions.inactivityThresholdMs` is `6e4`, and `buildReplayOptions` (replay.ts:61-75) never overrides it, so the SDK merges the 60s default in. `startRecording()` therefore calls `setupInactivityTracking()`, which attaches capture-phase `pointermove`, `pointerdown`, `scroll`, `keydown` and `input` listeners on `document` whose handler is `if (this.isPaused) this.resumeRecording()`. `pauseRecording()` clears only the inactivity timer — `teardownInactivityTracking()` runs only from `stopRecording()` — so those listeners stay attached across the pause this PR introduces. Five seconds after Pathfinder closes the recorder pauses, and the user's very next mouse move anywhere in Grafana resumes it. Verified against the locked 2.11.0 from package-lock.json, not just the installed 2.10.0. The stated goal is not reached. What the change does deliver is churn: `resumeRecording()` re-runs `record()`, so every close-then-interact cycle costs a fresh full-DOM snapshot plus `faro.session_recording.paused` and `resumed` events, which `filtering.ts:149-151` documents as passing through unfiltered on a collector whose volume is already the reason the sampling dial exists.
   Required: Set `inactivityThresholdMs: 0` in `buildReplayOptions` so the SDK never installs its interaction listeners and the surface lifecycle is the sole pause authority. Add a test that a pointer event dispatched after the pause does not resume recording.

## Follow-ups

1. [follow_up] **F-1770-3 — No test covers the pause, the delay, or the resume** (high · testing-and-verification)
   The test diff only reshapes existing assertions for the new `activateSessionReplay` return type and adds no-op `pauseRecording`/`resumeRecording` to the mock. Nothing asserts that pause runs after `REPLAY_CLOSE_PAUSE_DELAY_MS`, that reopening inside the window cancels the timer, that resume runs on reopen, or that a close during the dynamic import schedules the pause. `settleReplayImport` (faro.test.ts:1488) is a real 0ms timeout and no replay test uses fake timers, so the 5s timer can never fire in a test — and tests that report `closed` without reopening leave a live real timer holding the module-level controller after the test ends. Because the mock declares both methods as public no-ops, the `as unknown as` cast in replay.ts:112-119 is never exercised against the SDK's real private shape, so a rename would pass CI. The pre-existing test named 'keeps recording after Pathfinder is closed again' (faro.test.ts:1558) is left untouched and now states the opposite of the PR's intent, though it only ever asserted the recorder is not re-registered. Both blocking findings above are the kind of thing a behavioural test would have caught.
   Follow-up: Add fake-timer tests for pause-after-close, cancel-on-reopen, resume-on-reopen, and close-during-import, which also keeps the 5s timer from outliving the test. Assert against a mock that reflects the SDK's real method surface, and rename or rewrite the now-contradictory 'keeps recording after Pathfinder is closed again' test.

2. [follow_up] **F-1770-4 — Unguarded reach into SDK private methods from a bare setTimeout** (medium · correctness-and-reliability)
   `instrumentation.d.ts` declares `pauseRecording` and `resumeRecording` as `private`, and replay.ts:112-119 reaches them through `as unknown as` casts with no runtime feature detection. `controller?.pause()` then runs inside a bare `setTimeout` callback (faro-adapter.ts:182-187) with no try/catch and outside `guardTelemetry`, which the same file uses at 14 other sites and whose invariant is stated at line 39: telemetry must never break the app it observes. The optional chain guards a null controller, not a missing method, and the surface listener's try/catch cannot help because the timer runs on its own macrotask — a TypeError after an SDK bump would be an uncaught error in the host Grafana page. Separately, the success handler calls `handleSurfaceChange(...)` after `started = true` (faro-adapter.ts:204-213), so a throw from `resume()` there lands in the chain's `.catch()`, which resets `activating = false` and emits `sessionReplayActivationFailed` even though the instrumentation was already added. Both methods exist at the locked 2.11.0, so this is prospective rather than a live crash.
   Follow-up: Wrap both controller calls in `guardTelemetry`, and feature-detect the methods before casting so a missing one degrades to no pausing instead of throwing.

3. [follow_up] **F-1770-5 — TELEMETRY.md still documents the recorder as running for the rest of the page** (medium · analytics-and-telemetry)
   `docs/developer/TELEMETRY.md:78` states that 'the open that latches the activity gate is also what starts the recording, and once started it runs for the rest of the page, including after Pathfinder is closed again'. Both halves are what this change sets out to alter, and F-1770-1 invalidates the first half independently. The diff touches four source files and no documentation. That paragraph was authored by the same PR that shipped the recorder (#1482), so it is the contract prose for this behaviour, and it is what an operator reads to reason about what a recording covers. After this change the doc and the code disagree about the recorder's lifecycle, and because of F-1770-2 it is not even settled which of the two currently describes reality.
   Follow-up: Update the recorder-lifecycle paragraph in `docs/developer/TELEMETRY.md` in this PR to say when recording pauses, after how long, and what resumes it. Line 84's 'no removal path' can stay — it is about mid-session flag re-evaluation, and pausing is not removal.

## Suggestions

1. [suggestion] **F-1770-6 — start() is subscribed twice and unsubscribe() no longer bounds retries** (low · correctness-and-reliability)
   faro-adapter.ts:224-225 registers `handleSurfaceChange` and `start` as two separate listeners, while `handleSurfaceChange` already calls `start()` itself, so every non-closed transition enters `start` twice. The second entry is a guaranteed no-op because `activating` is set synchronously before the dynamic import, so there is no live defect. But `unsubscribe()` now removes only the second listener, and `start` stays reachable through `handleSurfaceChange`, which has to remain subscribed for pause and resume to work at all — so the bound described in the comment above `MAX_REPLAY_ACTIVATION_ATTEMPTS` rests solely on the attempts counter, while the `unsubscribe()` reads as enforcing something it does not.
   Suggested: Drop the second subscription and the `unsubscribe` indirection, leaving `handleSurfaceChange` as the single listener and the attempts counter as the stated bound.

---

Leaving this as a comment rather than a formal review, since the PR is still a draft — the findings are the same either way. Two things need to change before this can merge: report the surface from the boot path that actually opens a persisted panel, and pass `inactivityThresholdMs: 0` so the SDK stops resuming on its own. Mark it ready for review when you have had a pass at those and we will do a formal round. If either finding is unclear or you disagree with it, say so here — happy to work through it.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1770
Summary: pause replay when Pathfinder closes
Verdict: Request Changes
Results: 2 blockers, 1 non-blocking finding, 3 follow-ups

src/lib/telemetry/faro-adapter.ts
d0ugal: Fixed in src/components/floating-panel/FloatingPanelManager.tsx lines 127-128 and src/components/full-screen/FullScreenPanel.tsx lines 71-72. Runtime pre-mount reporting was removed from src/global-state/panel-mode.ts. Persisted-mode replay coverage is in src/lib/faro.test.ts lines 1559-1567, and the general activity gate now requires an explicit surface report in src/lib/telemetry/filtering.ts lines 207-224.

src/lib/telemetry/replay.ts
d0ugal: Fixed in src/lib/telemetry/replay.ts line 71, which sets inactivityThresholdMs to 0. The option is asserted in src/lib/faro.test.ts, and controller calls are guarded in src/lib/telemetry/faro-adapter.ts lines 160-167.

src/lib/telemetry/faro-adapter.ts
d0ugal: The replay controller boundary now checks method availability and wraps calls with guardTelemetry in src/lib/telemetry/faro-adapter.ts lines 160-167. The failure regression is covered in src/lib/faro.test.ts lines 1619-1632.

src/lib/telemetry/faro-adapter.ts
d0ugal: I removed the redundant second listener and unsubscribe indirection. src/lib/telemetry/faro-adapter.ts lines 198-243 now use handleSurfaceChange as the single listener; MAX_REPLAY_ACTIVATION_ATTEMPTS still bounds retries.

src/lib/faro.test.ts
d0ugal: Added fake-timer coverage in src/lib/faro.test.ts lines 1594-1617 for delayed pause, cancel-on-reopen, and immediate resume, plus lines 1824-1856 for close-during-import. The persisted-mode test is now named does not record from a persisted surface mode before it opens, and controller failures are covered at lines 1619-1632.

d0ugal:
All review feedback is addressed in commit 08741a0c.

- F-1770-1: mounted floating/fullscreen components report their surfaces in src/components/floating-panel/FloatingPanelManager.tsx:127-128 and src/components/full-screen/FullScreenPanel.tsx:71-72. Pre-mount reports were removed from src/global-state/panel-mode.ts.
- F-1770-2: src/lib/telemetry/replay.ts:71 sets inactivityThresholdMs to 0.
- F-1770-3: src/lib/faro.test.ts adds fake-timer coverage for delayed pause, cancel-on-reopen, immediate resume, and late activation.
- F-1770-4: src/lib/telemetry/faro-adapter.ts:160-167 feature-detects controller methods and wraps calls in guardTelemetry.
- F-1770-5: docs/developer/TELEMETRY.md:78 documents the five-second pause and instant resume lifecycle.
- F-1770-6: removed the redundant second surface subscription and unsubscribe indirection; retry limits remain enforced by MAX_REPLAY_ACTIVATION_ATTEMPTS.
- Additional lifecycle feedback: src/lib/telemetry/filtering.ts:207-224 now prevents persisted localStorage modes from opening the general activity gate before a real surface report.

Verification: 208 focused tests passed across src/lib/faro.test.ts and src/global-state/panel-mode.test.ts, and lint passed.

Jayclifford345 (approved):

PR #1771, opened 2026-09-04
chore(devtools): remove the legacy interactive-HTML authoring surface

moxious:
## Summary

Deletes the dead code that generated and configured interactive guides as HTML. Pathfinder stopped parsing `class="interactive"` / `data-targetaction` markup in #550, but two producers of that markup survived: the legacy WYSIWYG editor's config and the dev-tools tutorial exporter. Neither had a live caller. A third file, `step-parser.util.ts`, became fully dead as a direct consequence and goes with them. Net −649 / +17 lines, no behaviour change.

This is the first PR in a short series sweeping the residue left by #550.

## What to look at

- **`src/utils/devtools/`** — the load-bearing change. `tutorial-exporter.ts` is deleted outright; its only genuinely-used export, the `RecordedStep` type, moved to [`dev-tools.types.ts`](https://github.com/grafana/grafana-pathfinder-app/blob/973696590789bcba7efd07bc38e2759c8ba4f077/src/utils/devtools/dev-tools.types.ts) in its own commit so the move reads independently of the deletion. Also removes `useActionRecorder`'s `exportSteps`, which was the sole caller of `exportStepsToHTML` and had no caller of its own — worth confirming that read, since it's the one deletion that touches a live hook's public shape.
- **`src/utils/devtools/step-parser.util.ts`** — whole-file delete, folded in during review. Removing `exportSteps` orphaned `formatStepsToString`, its last live consumer; its two siblings `parseStepString` and `extractSelector` were already unreferenced. `StepDefinition` in `dev-tools.types.ts` went with it as its only consumer. Worth a second pair of eyes on the claim that nothing outside the barrel used these.
- **`src/components/block-editor/`** — import-line changes only (`useRecordingPersistence` and its test now take `RecordedStep` from the `utils/devtools` barrel rather than reaching into `tutorial-exporter` directly). No recording or persistence logic is touched.
- **`src/constants/editor-config.ts`** — whole-file delete, zero importers anywhere in `src/` or `tests/`. Note Tiptap stays a dependency: `block-editor/forms/MarkdownBlockForm.tsx` still uses it for the markdown field, so this delete does not drag the editor stack with it.
- **`docs/developer/`** — the constants README documented `editor-config.ts` as "Used By: `src/components/block-editor/`", which was already drifted before this PR.

## Why

Pathfinder JSON packages are the only guide format in use — all 38 bundled guides are JSON, and the authoring docs under `docs/developer/interactive-examples/` already contain zero HTML examples. #550 removed the parser, so anything still emitting interactive HTML produces output nothing can render. `tutorial-exporter.ts` was 304 lines generating exactly that.

These two deletions were originally scoped as separate PRs and deliberately combined: they are the same ratchet (the legacy HTML authoring and export surface), both are dead-code removals with no runtime behaviour change, and both are verified the same way. Splitting them would have doubled the review overhead for one story. The type relocation is kept as its own commit so the mechanical part stays separable.

`exportSteps`' string branch went with it rather than leaving a one-branch wrapper around `formatStepsToString`; that function is untouched and still exported from the barrel.

## How to verify

The real risk surface is record mode, since that is what consumed the exporter.

1. Open the block editor and start record mode. Perform a few actions in Grafana (a click, a form fill).
2. Confirm recorded steps still appear in the editor's step list with correct selectors and descriptions.
3. Stop recording and confirm the steps convert into JSON blocks — this path runs through `recorded-steps-processor.ts` and never touched the HTML exporter.
4. Reload the page mid-recording and confirm recording state still restores (exercises `useRecordingPersistence`, whose import changed).
5. Confirm no UI anywhere still offers a "Copy HTML" or "Download HTML" action — those labels lived in the deleted `editor-config.ts` and had no live consumer.

## Breaking changes

None for guide authors or end users. Several symbols leave the `utils/devtools` barrel — `exportStepsToHTML`, `combineStepsIntoMultistep`, `combineStepsIntoGuided`, `detectMultistepGroups`, `exportAsFullHTML`, `parseStepString`, `formatStepsToString`, `extractSelector`, plus the `MultistepGroup` / `ExportStep` / `ExportOptions` / `StepDefinition` types — and `useActionRecorder` no longer returns `exportSteps`. All were unreferenced outside the deleted files; this is internal surface with no external consumer.

## Reversibility

Fully reversible by revert — no storage format, telemetry, API contract, or manifest field is touched.

## Out of scope

Deliberately left for follow-up PRs in this series:

- **The `data-targetaction` DOM contract stays.** These attributes are emitted onto rendered DOM by `interactive-step.tsx` and read by E2E discovery, `interactive.styles.ts`, and dev tooling. They are a live runtime contract, not authoring syntax — a reviewer scanning for "interactive HTML" will hit them, and they are correctly documented as DOM attributes in `STEP_MODEL.md` and `E2E_TESTING_CONTRACT.md`.
- **`unstyled.html` support is untouched**, by design. The `content.json` → `unstyled.html` ladder in `content-fetcher.ts` serves docs pages loaded from the website and is unrelated to interactive HTML.
- **The DOMPurify allowlist** still permits `data-targetaction`, `data-reftarget`, and friends. Raw HTML in a JSON guide's markdown body currently keeps those attributes and gets styled like a real step while doing nothing. Fixing that inverts existing assertions in `security.test.ts`, so it belongs in its own PR.
- **Stale prose** in `ASSISTANT_INTEGRATION.md` (an HTML authoring example) and `docs/developer/styles/README.md`, plus a dangling reference to a never-committed `REMOVE-HTML-DYNAMICS.md` at `html-parser.ts:576`.
- **The `sequence` action runner** (`interactiveSequence` + `sequence-manager.ts`) appears to be unreachable from JSON guides — `sequence` is absent from the JSON guide action union — but that needs a dedicated verification pass before deletion.

Refs #550


Jayclifford345 (approved):

## Nits

1. [nit] **decorative-jsdoc-example-left-on-touched-hook — Partially-edited useActionRecorder JSDoc keeps a restating @example and stays detached from the hook** (low · correctness-and-reliability)
   The PR edited this docblock at src/utils/devtools/action-recorder.hook.ts:137-163 (dropping exportSteps from the destructure and the string-export lines) but left the rest: four comments that each restate the identically-named call beneath them, plus two @param options.* lines already declared on UseActionRecorderOptions. The block is also separated from the useActionRecorder declaration by DEFAULT_EXCLUDE_SELECTORS, so it documents the const and never surfaces in editor hovers. The misattachment predates the PR (verified at base c2d96f7d), but the AGENTS.md trim-on-touch rule covers the edited function and adjacent declarations in the same file.
   Suggested: While already in this docblock, drop the @example body and the two @param options.* lines, and move the surviving one-line summary below DEFAULT_EXCLUDE_SELECTORS so it attaches to useActionRecorder. Zero behavioral risk.

2. [nit] **pr-description-contradicts-its-own-formatstepstostring-claim — The PR description Why section still claims formatStepsToString survives, but head deletes it** (low · testing-and-verification)
   The Why paragraph says the exportSteps string branch went rather than leaving a wrapper around formatStepsToString, and that formatStepsToString is untouched and still exported from the barrel. That was true at commit f71b6b47, but the later review-sweep commit 9736965 deletes step-parser.util.ts and removes the barrel export, so the sentence is false at head. This is prose only: the code is consistent (zero references to formatStepsToString, parseStepString or the singular extractSelector; tsc --noEmit clean), and the same body describes the deletion correctly under What to look at and Breaking changes. The cost is that a reviewer trusting the paragraph could mistake the missing module for accidental over-deletion.
   Suggested: Edit the Why paragraph to match the sweep commit, noting that formatStepsToString was the last live consumer of exportSteps and that it and the rest of step-parser.util.ts are deleted here. No code change needed.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1771
Summary: remove the legacy interactive-HTML authoring surface
Verdict: Approve with Minor
Results: 0 blockers, 2 non-blocking findings, 0 follow-ups

PR #1774, opened 2026-09-04
refactor(types): Type Pathfinder window globals

### LINKED ISSUE #1751: Type the plugin's 12 window globals and stop reaching them through `as any` (24 sites)
Part of #603 (mechanical constraint enforcement). This is the **shape 3** case from the type-design analysis comment on that epic — a contract across a boundary, where TypeScript provably cannot help until the boundary is declared.

## The problem

The plugin uses 12 `window.__*` globals as a side channel between subsystems that do not import each other. Every one of the 24 access sites reaches them through `(window as any)`, which switches type checking off at exactly the point where two subsystems agree on a contract. There is no declaration for any of them.

Concretely, the renderer writes the active content key and analytics reads it; a tutorial hook writes the step counters and analytics reads those; the tab-exposure hook writes the active tab URL and both the renderer and analytics read it. The compiler sees none of these agreements.

## Evidence

The 12 keys, and the fact that nothing declares them:

| Key | Written at | Read at |
| --- | --- | --- |
| `__DocsPluginActiveTabUrl` | `src/components/docs-panel/hooks/useGlobalActiveTabExposure.ts:37` | `src/components/content-renderer/content-renderer.tsx:298`, `src/lib/analytics.ts:579` |
| `__DocsPluginActiveTabId` | `src/components/docs-panel/hooks/useGlobalActiveTabExposure.ts:36` | — |
| `__DocsPluginContentKey` | `src/components/content-renderer/content-renderer.tsx:348` | `src/lib/analytics.ts:580` |
| `__DocsPluginTotalSteps` | `src/components/interactive-tutorial/hooks/use-document-step-progress.ts:53` | `src/lib/analytics.ts:692` |
| `__DocsPluginCurrentStepIndex` | `src/components/interactive-tutorial/hooks/use-document-step-progress.ts:61` | `src/lib/analytics.ts:691` |
| `__pathfinderPluginConfig` | `src/hooks/usePathfinderPluginConfig.ts:59` | `src/utils/dev-mode.ts:191`, `:234`, `src/components/SelectorDebugPanel/SelectorDebugPanel.tsx:32`, `src/components/full-screen/FullScreenPanel.tsx:54`, `src/components/floating-panel/FloatingPanelManager.tsx:115`, `src/pages/docsPage.ts:18` |
| `__pathfinderKioskConfig` | `src/module.tsx:287` | `src/components/docs-panel/components/TabBarActions.tsx:109` |
| `__pathfinderKioskSessionId` | `src/utils/pathfinder-deep-link-handler.ts:81` | `src/lib/analytics.ts:272` |
| `__pathfinderExperiment` | `src/utils/experiments/experiment-debug.ts:47` | debug surface only |
| `__pathfinderAutoOpenUnlisten` | `src/utils/sidebar-auto-open.ts:76` | — |
| `__pathfinderDeepLinkNavUnlisten` | `src/utils/pathfinder-deep-link-handler.ts:200` | — |
| `__pathfinderHighlightedGuideNavUnlisten` | `src/utils/experiments/highlighted-guide-orchestrator.ts:245` | — |

The only two `declare global { interface Window }` blocks in `src/` describe third-party libraries — Prism in `src/interactive-engine/action-handlers/code-block-handler.ts:14` and the YouTube iframe API in `src/docs-retrieval/components/docs/youtube-video-renderer.tsx:19`. None of the 12 keys above is declared anywhere.

**One reader/writer disagreement is already live.** `src/hooks/usePathfinderPluginConfig.ts:12` publishes `__pathfinderPluginConfig` as `ResolvedPathfinderConfig`, which is `ReturnType<typeof getConfigWithDefaults>` — the shape *after* defaults are applied. `src/utils/dev-mode.ts:191` and `:234` read the same global as `DocsPluginConfig | undefined`, the raw interface from `src/constants.ts:110`, in which those defaulted fields are still optional. The other four readers assert no type at all. Nothing today can detect that these disagree.

## Three partial solutions already exist, and one of them is blind

**1. A typed accessor, for one key.** `src/hooks/usePathfinderPluginConfig.ts:11-23` does this properly:

`src/hooks/usePathfinderPluginConfig.ts:11-23` does this properly for its own key:

```ts
interface PathfinderConfigWindow extends Window {
  __pathfinderPluginConfig?: ResolvedPathfinderConfig;
}

function configWindow(): PathfinderConfigWindow {
  return window as PathfinderConfigWindow;
}
```

**2. A narrower read helper, for two keys.** `src/global-state/content-key.ts:30` has a `readGlobal(name: string)` that returns `string | undefined`, covering `__DocsPluginActiveTabUrl` and `__DocsPluginContentKey` on the read side only. It takes the key as a plain `string`, so a typo in a key name is not a compile error there either.

**3. An owner manifest, for three keys — and it cannot detect what it claims to.** `src/components/docs-panel/docs-panel.contract.test.tsx:287-298` defines `WINDOW_GLOBAL_OWNERS` for `__pathfinderPluginConfig`, `__DocsPluginActiveTabId` and `__DocsPluginActiveTabUrl`, and asserts each owner file assigns its global. The assertion is `expect(src).toContain(global)` — a substring match over the whole file.

`src/components/docs-panel/hooks/useGlobalActiveTabExposure.ts` mentions `__DocsPluginActiveTabId` twice: once in a doc comment at line 15, once in the actual assignment at line 36. **Delete the assignment and the test still passes**, because the comment satisfies the substring. This is the same failure shape as `verifyFieldsFromEffects` and as #1705 — a guard that looks like a ratchet and is not one. A declaration plus a lint rule replaces it with something the compiler enforces; the manifest test should then be narrowed to what it can actually prove, or removed as redundant.

So: three partial mechanisms, covering at most 3 of the 12 keys between them, and 24 access sites that use none of them. This issue generalises the accessor pattern in (1) to the whole surface.

## Why now

#1519 fixed one instance of this class — guide identity riding a race-prone window singleton — and closed. Nothing was installed to stop the next one. Per the epic's exemplar rule, the mechanism that closes the class is what is missing.

## Proposed work, in two parts

**Part 1 — the mechanism (maintainer, this issue).** Declare the 12 keys once with real types, in one module, using the `PathfinderConfigWindow` pattern above. Add a `no-restricted-syntax` rule to `eslint.config.mjs` rejecting `(window as any).__*` with a remediation message naming the declaration module. **Negative-test the rule**: per the correction comment on #603, an issue proposing a new constraint has to demonstrate the constraint fails when it should, not merely describe what it would catch.

**Part 2 — the drain (follow-up issues, filed once part 1 lands).** Remove the 24 casts, one consumer area per pull request: `lib/analytics.ts` (5 sites), the `components/` readers (6), `utils/` (6), the writers in `module.tsx` / `pages/` / hooks (7). The compiler adjudicates every disagreement each slice uncovers — starting with the `ResolvedPathfinderConfig` vs `DocsPluginConfig` split above, which is a real decision, not a mechanical rename.

Part 2 is behaviour-preserving and graded by `npm run typecheck`, so it does not need to be batched with part 1.

## Out of scope

Removing the globals themselves. Several are a genuine cross-subsystem side channel and replacing them with props or context is a separate architectural question. This issue makes the existing contract visible to the compiler; it does not relitigate whether the contract should exist.

## Acceptance

- No `(window as any).__*` remains in `src/` outside test files.
- `npm run typecheck` and `npm run lint` pass.
- The new lint rule has a test proving it fails on a reintroduced cast.
- Any reader/writer type disagreement the declaration surfaces is resolved deliberately and recorded, not papered over with a widened type.
- `WINDOW_GLOBAL_OWNERS` in `docs-panel.contract.test.tsx` is either narrowed to an assertion it can actually fail, or removed as superseded. It must not be left in place looking like coverage it does not provide.

loglapa:
## Summary

Centralize Pathfinder-owned `window` globals in one typed contract and replace production `(window as any).__*` accesses with typed properties.

The change also adds an ESLint restriction that prevents new untyped Pathfinder window-global accesses from entering production code.

Fixes #1751.

## Why

Pathfinder's runtime globals were declared implicitly at individual call sites. This made their ownership and value types difficult to discover, allowed casts to drift independently, and left new globals outside any enforceable contract.

A single `Window` augmentation makes the complete integration surface visible while preserving existing runtime behavior.

## What to look at

- `src/types/window-globals.ts` defines the shared contract for Pathfinder-owned globals.
- Existing consumers now access those properties directly through the typed `Window` interface.
- The ESLint rule rejects new `(window as any).__*` production accesses.
- The validation test exercises the real ESLint configuration and verifies both rejected and allowed forms.

## Out of scope

- Removing the window globals or changing their runtime ownership.
- Changing session, kiosk, experiment-debugger, or docs-panel behavior.
- Introducing a new state or dependency-injection mechanism.

## How to verify

- Run the window-global lint configuration test and confirm that an untyped Pathfinder global is rejected while typed access remains valid.
- Run the focused plugin-config, content-key, experiment-debugger, and docs-panel contract tests.
- Exercise the docs panel and plugin configuration flows and confirm that their exposed globals behave as before.

## Breaking changes

None. This is a type-safety refactor with no intended runtime behavior change.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -177,6 +177,13 @@ export default defineConfig([
             'Use DOM methods for structure, or sanitizeDocumentationHTML() if HTML structure is required.',
         },
         {
+          selector:
+            "MemberExpression[object.type='TSAsExpression'][object.expression.name='window'][object.typeAnnotation.type='TSAnyKeyword'][property.name=/^__/]",
+          message:
+            'Do not bypass the typed Pathfinder window-global contract with window as any. ' +
+            'Declare the global in src/types/window-globals.ts and access it through window directly.',
+        },
+        {
           selector: "CallExpression[callee.property.name='insertAdjacentHTML']",
           message:
             'Avoid insertAdjacentHTML() — it parses strings as HTML and risks XSS (F5). ' +
--- a/src/components/SelectorDebugPanel/SelectorDebugPanel.tsx
+++ b/src/components/SelectorDebugPanel/SelectorDebugPanel.tsx
@@ -29,7 +29,7 @@ export function SelectorDebugPanel({ onOpenDocsPage, onOpenLearningJourney }: Se
   const handleLeaveDevMode = useCallback(async () => {
     try {
       // Get current user ID and user list from global config
-      const globalConfig = (window as any).__pathfinderPluginConfig;
+      const globalConfig = window.__pathfinderPluginConfig;
       const currentUserId = (window as any).grafanaBootData?.user?.id;
       const currentUserIds = globalConfig?.devModeUserIds ?? [];
 
--- a/src/components/content-renderer/content-renderer.tsx
+++ b/src/components/content-renderer/content-renderer.tsx
@@ -295,7 +295,7 @@ export const ContentRenderer = React.memo(function ContentRenderer({
         return; // This handler was created for different content
       }
       const detail = (event as CustomEvent).detail;
-      const currentTabUrl = (window as any).__DocsPluginActiveTabUrl as string | undefined;
+      const currentTabUrl = window.__DocsPluginActiveTabUrl;
       if (detail?.completionPercentage >= 100 && detail?.contentKey && currentTabUrl) {
         // Only trigger if the event is for the current page (strict equality after normalization).
         // Bidirectional startsWith would produce false matches when URLs share a common prefix
@@ -345,7 +345,7 @@ export const ContentRenderer = React.memo(function ContentRenderer({
   // (progress restoration) runs — prevents stale key from a previous milestone.
   useLayoutEffect(() => {
     try {
-      (window as any).__DocsPluginContentKey = content?.url || '';
+      window.__DocsPluginContentKey = content?.url || '';
     } catch {
       // no-op
     }
--- a/src/components/docs-panel/components/TabBarActions.tsx
+++ b/src/components/docs-panel/components/TabBarActions.tsx
@@ -106,7 +106,7 @@ export const TabBarActions: React.FC<TabBarActionsProps> = ({
     locationService.push(PLUGIN_BASE_URL);
   };
 
-  const [kioskEnabled, setKioskEnabled] = useState(!!(window as any).__pathfinderKioskConfig);
+  const [kioskEnabled, setKioskEnabled] = useState(!!window.__pathfinderKioskConfig);
 
   useEffect(() => {
     const onReady = () => setKioskEnabled(true);
--- a/src/components/docs-panel/docs-panel.contract.test.tsx
+++ b/src/components/docs-panel/docs-panel.contract.test.tsx
@@ -275,29 +275,6 @@ describe('E2E Contract: Scroll-restoration DOM id', () => {
 });
 
 /**
- * Window globals assigned by the docs-panel surface for cross-component
- * communication. These are read by dev-mode utilities, SelectorDebugPanel,
- * interactive-section, and analytics. Do not remove during refactoring —
- * document for future migration to React Context.
- *
- * Owner manifest: each global is assigned in exactly one file. As the
- * renderer is decomposed, hooks take ownership of individual assignments —
- * update the owner path here in the same commit that moves the source.
- */
-const WINDOW_GLOBAL_OWNERS: Array<{ global: string; ownerFile: string }> = [
-  { global: '__pathfinderPluginConfig', ownerFile: '../../hooks/usePathfinderPluginConfig.ts' },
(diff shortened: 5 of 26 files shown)


moxious (approved):
Thanks for typing the Pathfinder window globals. Swapping the scattered `window as any` casts for one tier-0 declaration plus a lint rule that stops them coming back is exactly the ratchet shape we want, and keeping the call-site churn to what the rule forces makes it easy to read — 16 of the 26 files move by four lines or fewer.

Nice catch on the live reader/writer disagreement, and on resolving it toward the published shape rather than widening the type.

Approving. Two non-blocking items are filed separately rather than held against this: #1776 (the lint selector's coverage gap) and #1777 (two stale doc pointers). Details recorded on #1751.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1774
Summary: Type Pathfinder window globals
Verdict: Approve with Minor

PR #1788, opened 2026-09-06
Publish internal docs to EngHub with the shared TechDocs workflow

### LINKED ISSUE #445: enhancement:Use shared workflows to publish docs to enghub
Workflow here:

https://github.com/grafana/shared-workflows/blob/main/.github/workflows/publish-techdocs.yaml
https://github.com/grafana/shared-workflows/tree/main/actions/techdocs-rewrite-relative-links

Related: https://github.com/grafana/grafana-recommender/issues/82

rajathpi:
Closes #445

Wires up the shared [publish-techdocs workflow](https://github.com/grafana/shared-workflows/blob/main/.github/workflows/publish-techdocs.yaml) so the internal engineering docs (`docs/developer`, `docs/design`, `docs/history`) get published to EngHub.

- `.github/workflows/publish-techdocs.yml` — runs on pushes to main that touch docs, plus a manual dispatch with an instance picker so it can be tried against dev before ops
- `catalog-info.yaml` — registers the component with `techdocs-ref: dir:.`
- `mkdocs.yml` — techdocs-core config. `docs_dir` is set explicitly because the rewrite-relative-links action reads that key to find the docs root (without it the rewrite silently does nothing). `docs/sources` is excluded since the product docs are already published to grafana.com
- `docs/index.md` — small landing page so the site root resolves

`rewrite-relative-links: true` handles the ~10 docs with links that escape the docs tree (`AGENTS.md`, `.cursor/rules/*`, `src/...`) by rewriting them to GitHub URLs at build time.

Verified locally with mkdocs + techdocs-core 1.3.5 (the versions the shared workflow pins): builds clean, `sources/` excluded, only the expected out-of-tree link warnings that the rewrite step covers.

Two things I couldn't check from outside the org: the owner group in `catalog-info.yaml` (guessed `pathfinder` — please correct) and whether the repo is already registered with EngHub.

### DIFF
--- a/.github/workflows/publish-techdocs.yml
+++ b/.github/workflows/publish-techdocs.yml
@@ -1,0 +1,40 @@
+name: Publish TechDocs
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'docs/**'
+      - 'mkdocs.yml'
+      - 'catalog-info.yaml'
+      - '.github/workflows/publish-techdocs.yml'
+  workflow_dispatch:
+    inputs:
+      instance:
+        description: EngHub instance to publish to
+        required: true
+        type: choice
+        default: 'ops'
+        options:
+          - 'dev'
+          - 'ops'
+
+concurrency:
+  group: '${{ github.workflow }}-${{ github.ref }}'
+  cancel-in-progress: true
+
+permissions: {}
+
+jobs:
+  publish-docs:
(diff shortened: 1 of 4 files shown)


Jayclifford345:
Hi @rajathpi thank you for the contribution, but our dev docs will not be something we are currently looking to publish. Thank you for the awesome work, though.  If you would like to contribute again in the future checkout our issues list and that should show our focus areas

PR #848, opened 2026-05-06
feat(block-editor): authoring resilience + sidebar UX overhaul

tomglenn:
image

## Summary

Stacked-but-cohesive PR covering three rounds of authoring-resilience work plus a sidebar-density UX overhaul. 37 commits, 54 files, +5461 / -584. Branched in three phases originally (`phase-1` → `phase-2` → `phase-3`) but lands as a single PR.

### Authoring resilience (phases 1–3)

- **Phase 1 — inline lint primitives.** Real-time per-field condition validation with `<LintBadge>` per block, shared `lintGuide` pipeline, parity guard against the canonical CLI validator.
- **Phase 2 — chip picker.** Structured chip-based editor for `requirements` / `objectives` (replaces freeform Inputs). Parameterized conditions get per-prefix helpers; `has-datasource:` is split into type vs. name groups; "Remove" quick-fix for fully-unknown tokens.
- **Phase 3 — guide-level Health surface.** Editor-only cross-block checks (`firstStepMissingOnPage`, `orphanSectionReference`, `destructiveActionWithoutObjective`, `unusedSection`, `requirementsImpliedByActionButNotDeclared`); context-aware requirements suggester; new `HealthStatusBar` (IDE-style sticky bottom bar) replacing the original right-side aside; in-form `SuggestedRequirementsRow` companion.

### Sidebar UX overhaul (4 phases of cleanup, single PR)

The block editor was styled like a desktop tool but lives in a ~360–500px sidebar. Three-level nesting consumed ~33% of horizontal space and the always-visible action pill burned ~110px on every block.

- Tightened nested indent (40 → 10 px / level) — `marginLeft` 24 → 0, `paddingLeft` 16 → 10, only the 3px coloured `borderLeft` indicates nesting.
- Hover-revealed secondary actions (eye / copy / delete / record); Edit stays anchored to the right edge always.
- Single-line block summary — drops the redundant sub-type badge and the separate preview row; LintBadge moves out of the badge cluster to the right of the content column.
- Single-row header — `New` and `Library` move into the kebab menu; saved indicator slimmed to a small `save` icon; selection trigger relocated to the header.
- Inline section / branch `+ Add block` zones now match block-card height (40px) and sit unobtrusively at rest; the footer `+ Add block` matches them. Quieter drop indicators between blocks.
- Health-bar legend collapsed to a `?` icon next to the first severity heading.

### Late-stage polish fixes

[...]

PR #1791, opened 2026-09-07
fix(objectives): reject invalid completion checks and preserve condition arrays

Jayclifford345:
Prevents invalid executable objectives from completing steps and preserves comma-containing conditions through local and cross-tab checks. Existing objectives remain the authoring model; the separate outcomes schema, evidence store, feature flag, pilot UI, and Assistant integration have been removed.

Changes:

- Distinguish satisfied, unsatisfied, unavailable, and invalid checks while retaining the boolean compatibility projection. Unknown prerequisites retain legacy fail-open eligibility; only an explicit `satisfied` verdict is completion evidence, so no invalid objective or postcondition can establish completion.
- Accept condition arrays in the shared checker and cross-tab transport, then preserve arrays through JSON parsing and runtime forwarding. An array element is one condition, so a parameter value may contain commas; the comma-separated string remains a legacy input boundary. A challenge's success criterion is passed as one condition.
- Match conditions per token wherever the runtime inspects them. Substring matching on an array is an exact-element test, so the terminal-status, `on-page:` heartbeat, and `section-completed:` recheck triggers are now token-matched via `hasConditionToken` / `hasConditionPrefix`.
- Enforce the objective vocabulary without stranding published content: `condition-validator` warns, `json-parser` drops the unrecognised token, and the runtime refuses to complete on a non-`satisfied` verdict. `objectives` keeps a permissive `ObjectiveTokenSchema` rather than the blocking `RequirementTokenSchema`, because the field shipped documented as free-text "learning objectives" and guides already published to a backend repository or CDN can hold prose that no in-repo validation can enumerate or migrate.
- Extract the step-decision rules into a pure function while preserving alignment pauses, existing-objective shortcuts, sequential gating, reset behavior, subscriptions, retries, and persistence.
- Bound cross-tab condition arity and per-token length, and strengthen the wire mirrors so optional-only drift fails to compile.

Before: a typo such as `has-dashbord-named:Example` could be accepted as an objective and inherit a successful unknown-check fallback. After: authoring warns, the parser drops the token, and the runtime cannot complete the step on it. An empty-argument token with a known prefix (`has-dashboard-named:`, `var-`) is covered too — `isValidRequirement` accepts the prefix, so the router's verdict is what refuses it.

Before: joining an array containing `has-dashboard-named:Sales, Europe` split the dashboard name during checking. After: the array element reaches one handler intact, including across tabs, and `condition-validator` no longer splits it either. API failures now produce an unavailable verdict rather than appearing as a missing resource.

This extends the existing objective and requirement checking contracts, and records them as the `#1662 → #1791` anchor for `guide-schema-and-contracts`. It introduces no new guide authoring fields or CRD changes, and does not change progress percentages or historical completion records. Resource-specific objective evidence and additional Assistant context are deferred.

Behavior change worth review: `objectives` no longer rejects an unrecognised token at the schema. `validateGuide` returns early on a Zod failure, so a blocking check meant `condition-validator` never ran and one stale token replaced the whole guide with a parse-error panel. Since `upsert-learning-path.sh` forwards `objectives` to the CRD with jq shape checks only and the backend never validates it, that failure would have landed on guides in customer stacks this repo cannot inspect. The runtime verdict gate is what actually prevents a typo from auto-completing a step, so the schema does not need to be the gate that strands content.

Validation:

- `npm run check`: all nine gates pass, including type checking, linting, formatting, Go lint/tests, frontend coverage, review-contract tests, and script validation.
- `npm run build`: passed (existing bundle-size warning).
- Focused regression coverage includes invalid and empty-argument objectives in both the string and array shape, a `condition shape parity` suite running the terminal / `on-page:` heartbeat / `section-completed:` triggers in both shapes, comma-containing parameters and Coda commands, cross-tab transport and arity bounds, unavailable APIs, and step-decision behavior.

Based directly on current `main`. This is the retained PR for the correctness and decision-extraction work; #1792–#1798 remain superseded and closed.


Jayclifford345 (commented):
## Merge contract

1. [blocking] **condition-substring-match-breaks-recheck-triggers — Three recheck triggers still substring-match a value that is now an array, so they are dead for every JSON-authored guide** (high · requirements-manager)
   `json-parser.ts` stopped joining `requirements`/`objectives` and now emits arrays, but three guards in `step-checker.hook.ts` are unchanged from base and match by substring, which silently becomes exact-element matching on an array. (1) Line ~1079 `readsTerminalStatus` is gated on `typeof clause === 'string'`, now permanently false, so the `TERMINAL_STATUS_CHANGED_EVENT` listener is never attached — and it is the only listener for that event in the repo, so an `is-terminal-active` / `coda-exit-zero:` step never re-checks when the terminal connects. (2) Line ~1114 `req.includes('on-page:')` no longer matches a prefix token, so `on-page:` steps lose the heartbeat drift watchdog; the heartbeat is on by default and 6 bundled guides use `on-page:`. The bare `navmenu-open` and `exists-reftarget` tokens still match, which is what makes this easy to miss. (3) Line ~956 `requirements?.includes('section-completed:')` no longer matches, so that listener is dead — largely masked in shipped content by the section-level 5s poll, at up to 5s latency. Both skeptics reproduced all three with string-vs-array controls; the terminal case has zero in-repo content but real exposure in Coda guides outside this repo. TypeScript cannot see any of it because `.includes` exists on both `string` and `readonly string[]`.
   Required: Match on tokens, not substrings: run each value through `conditionTokens()` and compare with `===` / `startsWith` for the parameterized prefixes. Then audit the rest of the widened surface the same way — the type change reaches ~20 declarations and `.includes`, truthiness (`[]` is truthy where `''` was falsy), and `=== '...'` all type-check on both shapes.

2. [blocking] **invalid-objective-with-empty-parameter-still-auto-completes-step — An empty-parameter objective still auto-completes the step, which is the hole this PR set out to close** (high · requirements-manager)
   The hook gates objectives with `objectiveTokens.every(isValidRequirement)`, but `isValidRequirement` is prefix-only, so `has-dashboard-named:` and `var-` pass it. The router added a stricter rule (`check.endsWith(':') || check === 'var-'`) that returns `verdict: 'invalid'` — but in `pre` mode it also returns `pass: true` to preserve fail-open eligibility, and the objectives path reads only `.pass`, never `.verdict`. Verified end to end on the real hook: `objectives: ['has-dashboard-named:']` and `['var-']` both give `isCompleted=true completionReason=objectives`. This is a regression, not a pre-existing hole — at base the same tokens reached their handlers and returned `pass=false` (`Dashboard named '' not found`). It is reachable from authored JSON (`parseJsonGuide` reports `isValid=true`) and the stakes are durable: `completionReason: 'objectives'` writes through `markStepCompleted` and `objectives` is an accepted server-side completion-record source. The PR's own new test picks the misspelled `has-dashbord-named:Example`, so it never covers the valid-prefix/empty-argument case. Cursor Bugbot flagged the same thing independently.
   Required: Delete the hook's `validObjectives` re-derivation and gate on the verdict the router already computes: `objectivesPassed = checkVerdict(objectivesResult) === 'satisfied'`. That collapses three separate validity rule sets to one and implements the invariant this PR's own doc states — "Only an explicit `satisfied` verdict is outcome evidence." Separately, tighten `RequirementTokenSchema` to reject an empty argument after a parameterized prefix, and add both tokens to the new execution-time table.

3. [blocking] **legacy-prose-objectives-reject-entire-guide — A published guide with a prose objective now fails to render entirely, with no fallback or migration** (high · guide-schema-and-contracts)
   The contract-evolution scan found the *meaning* change is legitimate — `follows_contract`: PR #417 declared objectives a closed vocabulary, #848 made it the authoring-UI contract, and `condition-validator.ts` has validated objectives with the requirements grammar all along. The problem is the enforcement level. An unknown objective token was an advisory warning from `condition-validator`; it is now a hard Zod error, and `validate-guide.ts` returns early on a Zod failure, so `condition-validator` never runs and no per-block degradation exists. Verified base-vs-head: `objectives: ['Learn about dashboards']` parses `isValid=true` at base and `isValid=false, hasData=false` at head, after which `content-renderer.tsx` replaces the whole guide with a `ContentParsingError` panel. In-repo blast radius is numerically zero (one bundled `objectives`, valid). The exposure is entirely external and unverifiable from here: `upsert-learning-path.sh` forwards `objectives` to the App Platform CRD on an allowlist with jq shape checks only — never the Zod schema — and `grep -rn objectives pkg/` shows the backend does not validate it either, so prose objectives publish to customer stacks today. `backend-guide.ts` runs `validateGuide` on read, so such a guide fails 100% of the time after this ships. The base `describe()` text actively invited prose ("Learning objectives this block addresses"), and the PR had to rewrite four in-repo fixtures authored under it.
   Required: Pick one and record it. Either soften the objectives token check to a warning that drops the unrecognised token and lets the guide render — the runtime half of this PR already prevents the actual bug, so the schema need not be the gate that strands content — or keep the hard error and state the breaking-content change in the PR body with a recovery path for affected authors. If you keep it hard, note that the whole-guide failure shape is the harshest available; per-block degradation would cost far less.

4. [blocking] **step-checker-hook-tests-exercise-only-the-legacy-string-shape — No hook-level test uses the array shape the PR made canonical, which is why the regressions above passed CI** (high · testing-and-verification)
   `step-checker.hook.test.ts` contains zero array literals for `requirements`/`objectives` — every input is a comma-separated or bare string, including the `objective validation at execution time` table this PR adds. Arrays are now the only production shape for JSON guides. The decisive evidence: a skeptic copied the test file to an untracked scratch copy and changed nothing but the prop shape at three call sites; two existing tests failed immediately (`rechecks a blocked step when the terminal reports a status change` and `subscribes a step whose objectives read the terminal, not only its requirements`). The PR's array-shaped tests all live in the pure/router layers — `condition-input`, `check-verdict`, `controller-requirements`, `cross-tab.types`, `requirements-checker.utils`, `live-tab-executor` — none at the hook level where the three broken guards live. The suite is green at 358 tests, and typecheck passes, so nothing signals the gap.
   Required: Parameterize the hook suite over both shapes (`it.each([['string', 'a,b'], ['array', ['a','b']]])`), at minimum for the heartbeat, terminal-status, and section-completed listener tests, and add one end-to-end assertion that `json-parser` output reaches `useStepChecker` unsplit. Any widened-type call site not covered by a shape-parameterized test has the same blind spot.

5. [blocking] **authoring-docs-objectives-examples-now-invalid — Three worked examples in the canonical authoring reference now produce unparseable guides** (high · guide-schema-and-contracts)
   `docs/developer/interactive-examples/json-guide-format.md` lines ~197 (`visited-dashboards`), ~400 (`completed-tour`) and ~511 (`viewed-logs`) are `objectives` examples for an interactive block, a section, and a conditional section config. All three tokens fail `isValidRequirement`; copying each fence verbatim gives `isValid=false, hasData=false` at head and `isValid=true` at base. `git diff --name-only bd222712 a33d7bfb -- docs/` returns only `requirements-manager.md`, so the PR documents the change in the engine doc while leaving the authoring reference contradicting it. This file is what authors are pointed at from `README.md`, `GETTING_STARTED.md` ("read this before editing any guide"), `CUSTOM_GUIDES.md`, `package-authoring.md`, and `schema-coupling.mdc`, and it sits inside the `prevent-doc-drift` allowed-edit set, so the repo's own per-PR doc discipline expects this update here. Nothing catches it: `json-guide-format-docs.test.ts` only checks block-type coverage against the schema and never parses the fences. The field tables in the same file also still frame objectives as "Objectives marked complete after this step" — author-declared labels credited on completion, the inverse of a condition the runtime evaluates to cause it.
   Required: Replace the three example values with real tokens and reword the objectives rows to the condition framing the new `describe()` uses. Consider extending `json-guide-format-docs.test.ts` to parse the JSON fences through `JsonGuideSchema`, which would have caught this and will catch the next one. `docs/design/AGENT-AUTHORING.md` also still quotes the old schema line and CLI help, but that is a human-curated design doc that `prevent-doc-drift` forbids editing, so it is out of scope here.

6. [blocking] **comma-inside-single-array-element-no-longer-splits — The CLI's own documented usage now writes a permanently unsatisfiable prerequisite** (medium · docs-retrieval-and-rendering)
   `conditionTokens` splits only the string form, so a comma inside an array element is now part of the parameter. That is the intended trade, but it silently breaks the authoring route the product itself documents. `add-block interactive --help` prints `--requirements <item>` — singular, not variadic — and illustrates the value comma-separated twice, with no hint the flag is repeatable; that example text comes from the schema's own `.describe('Prerequisite conditions (e.g., on-page:/dashboards, is-admin)')`, so the MCP/agent surface is invited identically. The tiebreaker ran the real CLI: `add-block interactive --requirements "on-page:/dashboards, is-admin"` reports `package valid: true` and writes `["on-page:/dashboards, is-admin"]` — one element. At base the parser joined and the checker re-split, so this input worked as two conditions; at head it routes to the `on-page:` handler and compares the whole `/dashboards, is-admin` as the required path, blocking the step forever. It fails silently: `validate --strict` prints PASS, because `condition-validator` still splits every element on commas and both halves look valid. The same inconsistency cuts the other way — `condition-validator` now false-positives on the capability this PR adds, warning `Unknown condition type 'EU'` for `has-dashboard-named:Sales, EU`. Zero in-repo content is affected; the reachable path is the shipped CLI and any agent following the schema example.
   Required: Make the new atomic-token contract enforceable and consistent instead of silent. Update `condition-validator` to validate array elements whole (keeping the comma split only for the legacy comma-separated string boundary), fix the `--requirements`/`--objectives` help text and the schema `.describe()` examples so they stop illustrating a comma-separated single value, and make the flags visibly repeatable. A schema refinement rejecting a comma in a token whose prefix does not take free text would catch the rest at authoring time.

## Suggestions

1. [suggestion] **wire-mirror-guards-blind-to-optional-fields — The cross-tab wire-mirror guards cannot see optional-field divergence, which is exactly what this PR adds** (medium · cross-tab-controller)
   `cross-tab.types.ts` carries a comment promising that if `CheckResultError` and the wire type "stop being mutually assignable (field added/removed/retyped on either side), these lines stop compiling", and `live-tab-executor.ts` implements a `MutuallyAssignable` check over the result types. A tsc probe shows both are blind to optional-only divergence: `MutuallyAssignable<{requirement;pass;verdict?}, {requirement;pass}>` compiles cleanly, while the required-field version fails with TS2322. This PR happens to add `verdict?` to all four sides, so nothing is divergent today — but it is the first change to add an optional wire field, and the next one can add it to the internal type only and ship a controller reading a field the live tab never sends, with green CI.
   Suggested: Compare required-key sets and full key sets in both directions (`keyof` equality plus mutual assignability over `Required<A>`/`Required<B>`), and correct the comment to describe what is actually enforced.

2. [suggestion] **conditional-stringify-parse-roundtrip-and-stale-comments — `JSON.parse(JSON.stringify(conditions))` reimplements referential stability, and two adjacent comments now describe code that is gone** (low · correctness-and-reliability)
   `interactive-conditional.tsx` derives `conditionsKey = JSON.stringify(conditions)` and then `requirements = useMemo(() => JSON.parse(conditionsKey), [conditionsKey])` purely to get a value-stable array, with an unchecked cast where `isConditionInput` from the PR's own new `condition-input.ts` already exists. `useMemo(() => conditions, [conditionsKey])` gives the same stability without serialising. Two touched comments are now wrong: one still says "The joined string is referentially stable" over a JSON string, the other says "Convert conditions to requirements string format" directly above a line producing `string[]`. AGENTS.md trim-on-touch applies. Verified harmless: the derived `conditionalId` strips `[`, `]`, `"` and `,` so it is byte-identical to the old joined form across seven representative condition sets, and `conditionsKeyNeedsDomWatch` still works because the JSON string contains the token as a substring.
   Suggested: Keep `conditionsKey` for effect keying, pass `conditions` through memoized on it, drop the `JSON.parse`, and delete or fix the two comments.

## Nits

1. [nit] **conditioninput-imports-above-module-docblocks — The new `ConditionInput` import was inserted above the module doc block in 13 files** (low · correctness-and-reliability)
   In 13 changed files the added import is line 1 and the file's own doc block starts at line 2, detaching each module's header documentation from the module — including `requirements-checker.utils.ts` ("Requirements checking — router and retry harness"), `controller-requirements.ts`, `section-completed-check.ts`, `user-storage.ts`, four `types/*.ts`, and four `interactive-tutorial` components. The two new files this PR authors are ordered correctly, which is what makes the rest read as tool output rather than intent.
   Suggested: Move each inserted import below the leading doc block into the existing import group.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1791
Summary: reject invalid completion checks and preserve condition arrays
Verdict: Request Changes
Results: 6 blockers, 3 non-blocking findings, 0 follow-ups

Jayclifford345:
## Blockers addressed in a4f1b0ce

All six blockers from the review above are fixed on this branch. `npm run check` passes all nine gates (8398 tests).

| Blocker | Fix |
| --- | --- |
| `condition-substring-match-breaks-recheck-triggers` | [471f681c](https://github.com/grafana/grafana-pathfinder-app/commit/471f681c) — new `hasConditionToken` / `hasConditionPrefix` in `src/lib/condition-input.ts`; all three guards now match per token |
| `step-checker-hook-tests-exercise-only-the-legacy-string-shape` | [471f681c](https://github.com/grafana/grafana-pathfinder-app/commit/471f681c) — new `condition shape parity` suite runs the terminal, `section-completed:` and `on-page:` heartbeat behaviours in **both** shapes via `describe.each` |
| `invalid-objective-with-empty-parameter-still-auto-completes-step` | [754792a6](https://github.com/grafana/grafana-pathfinder-app/commit/754792a6) — the hook's `validObjectives` re-derivation is gone; completion now requires `checkVerdict(objectivesResult) === 'satisfied'`. Router verdicts for `has-dashboard-named:` and `var-` are pinned in `requirements-checker.utils.test.ts` |
| `legacy-prose-objectives-reject-entire-guide` | [754792a6](https://github.com/grafana/grafana-pathfinder-app/commit/754792a6) — warn-and-drop, see below |
| `comma-inside-single-array-element-no-longer-splits` | [e7b8f00e](https://github.com/grafana/grafana-pathfinder-app/commit/e7b8f00e) — `validateConditions` validates each element whole; `--requirements` help and the schema `describe()` no longer illustrate a comma-joined value |
| `authoring-docs-objectives-examples-now-invalid` | [a4f1b0ce](https://github.com/grafana/grafana-pathfinder-app/commit/a4f1b0ce) — the three examples now use real tokens and the field tables use the condition framing |

### The one design change worth reviewing

`objectives` no longer uses `RequirementTokenSchema`. It has its own permissive `ObjectiveTokenSchema`, and the vocabulary is enforced one layer later:

- `condition-validator` still **warns** on an unrecognised objective (it already did)
- `json-parser` **drops** the unrecognised token before it reaches the runtime
- the runtime **refuses to complete** on any non-`satisfied` verdict

Rationale: `validateGuide` returns early on a Zod failure, so `condition-validator` never ran and `content-renderer` replaced the *whole guide* with a parse-error panel. Since `upsert-learning-path.sh` forwards `objectives` to the CRD with jq shape checks only and the backend never validates it, guides already published to customer stacks can hold prose from the months when the field was documented as free-text "learning objectives" — and this repo cannot enumerate or migrate them. The runtime gate is what actually prevents a typo from auto-completing a step, so the schema does not need to be the gate that strands content.

Net effect on the original goal, unchanged: a typo'd objective still cannot complete a step. It just no longer takes the guide down with it.

### Two non-blocking items left open

`wire-mirror-guards-blind-to-optional-fields` and the three deferred follow-ups are untouched — happy to take those in a follow-up if you want them here.

### One thing to check yourself

The PR body still says "a typo such as `has-dashbord-named:Example` … authoring rejects it". With warn-and-drop that is now "authoring warns and the runtime refuses it". Worth a quick edit before merge.

Jayclifford345:
## Remaining items closed in d1e6e578, and the description updated

Everything left open from the review above is now addressed. `npm run check` passes all nine gates (8404 tests).

| Item | Fix |
| --- | --- |
| `wire-mirror-guards-blind-to-optional-fields` | [244e47cd](https://github.com/grafana/grafana-pathfinder-app/commit/244e47cd) — new `WireMirrors<A, B>` adds key-set and required-key-set equality on top of mutual assignability |
| `condition-array-unbounded-arity-prevalidated-preauth` | [244e47cd](https://github.com/grafana/grafana-pathfinder-app/commit/244e47cd) — `isBoundedConditionInput` caps arity at 32 tokens and each token at 512 chars, using the existing `isBoundedString` pattern |
| `challenge-form-prose-objectives-unguarded` | [d1e6e578](https://github.com/grafana/grafana-pathfinder-app/commit/d1e6e578) — see below; resolved differently than the review suggested |
| `objectives-contract-anchor-not-recorded` | [d1e6e578](https://github.com/grafana/grafana-pathfinder-app/commit/d1e6e578) — `guide-schema-and-contracts` anchor is now `#1662 → #1791` |

I verified the strengthened wire guard actually catches the case the old one missed: adding an optional field to `CheckResultError` alone now fails `tsc` (`Type 'true' is not assignable to type 'never'`) where it previously compiled clean.

### One item resolved differently than the review proposed

The review suggested rewriting the `ChallengeBlockForm.test.tsx` fixtures off `objectives: ['Create a dashboard']`, on the grounds that the repo was asserting a shape the new schema forbids. Warn-and-drop dissolved that premise — prose objectives are schema-valid again. Those tests are specifically about preserving fields the form does not render, and a value the form cannot interpret is what makes them meaningful, so I left them alone.

What warn-and-drop *did* create is a worse problem the review had not found: `destructiveActionWithoutObjective` counted any objective, so a destructive step whose only objective was prose cleared the warning while the parser dropped the token and the step still re-ran on a second pass — hiding exactly the hazard the rule exists for. That check now counts only executable objectives, with a regression test.

### Description updated

The body no longer claims authoring rejects a typo'd objective. It now states the layered enforcement, calls out the schema change as a behavior change worth review, and notes the empty-argument case (`has-dashboard-named:`, `var-`) that `isValidRequirement` accepts and only the router's verdict refuses.

moxious (commented):

## Follow-ups

1. [follow_up] **PR1791-REQ-1 — Cross-tab requirement check silently degrades when a new controller tab is paired with an old live tab** (medium · requirements-manager)
   The two-tab controller's `check-requirements`/`fix-requirement` wire message widens `requirements` from string to string|array. A pre-PR live tab's validator still requires a plain string, so it silently drops the message; the new controller's request then times out and falls back to a local-only check that skips the tab-local DOM/URL requirements the live tab would have verified. The mechanism was independently confirmed by two skeptics tracing the actual base-vs-head code. It is not a new violation of a live guarantee, though: `src/types/cross-tab.types.ts` and `docs/developer/CROSS_TAB_CONTROLLER.md` already declare, unchanged since before this PR, a 'same-build assumption' — cross-version tab compatibility is explicitly out of scope. This is a new instance of that already-accepted risk category, in a field where array-shaped requirements are now the mainline representation for most guides, and the whole feature remains admin-gated and off by default.
   Follow-up: Tracked in https://github.com/grafana/grafana-pathfinder-app/issues/1808 — no blocking change needed given the standing same-build non-goal.

## Suggestions

1. [suggestion] **PR1791-REQ-2 — Empty-array requirements no longer hit the checker's fast-path** (low · requirements-manager)
   `executeChecksWithRetry`'s `if (!requirements)` fast-path only catches the falsy empty-string case; `requirements` is now typed `ConditionInput` and an empty array `[]` is truthy in JS, so it falls through to the general check path instead. Today both paths coincidentally produce the same `satisfied` result (an empty token list vacuously passes), so there's no live bug, but the divergence is untested and a future edit to either branch could silently apply only to one shape.
   Suggested: Replace `if (!requirements)` with `if (conditionTokens(requirements).length === 0)`, and add a test asserting `requirements: []` short-circuits the same way `''`/`undefined` do.

2. [suggestion] **sec-1791-02 — This PR also closes a pre-existing unbounded-string gap, not just the new array shape** (low · security)
   The commit message frames the new arity/length bounds as guarding only the newly-widened array shape, but the base commit's string-shaped `requirements` field on `check-requirements`/`fix-requirement` had no length bound at all. This PR's `isBoundedConditionInput` bounds both shapes, which is a net-positive fix beyond what the commit message claims.
   Suggested: No code change needed; optionally credit this PR with closing the pre-existing unbounded-string gap in the PR description or changelog.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1791
Summary: reject invalid completion checks and preserve condition arrays
Verdict: Approve with Minor

moxious (approved):
see previous review

PR #1803, opened 2026-09-07
docs(external-api): clarify unknown field pruning

### LINKED ISSUE #1676: EXTERNAL_API.md claims unknown fields get a 422; they are silently pruned with a 201

dvd233:
## Summary

Correct the external API documentation to distinguish validation of declared fields from Kubernetes's default pruning of undeclared fields. A write containing unknown fields can still succeed and lose data, so the document now points manifest extensions to `spec.manifest.additionalFields` and describes the optional warning header consistently.

## What to look at

- [`docs/developer/EXTERNAL_API.md`](https://github.com/dvd233/grafana-pathfinder-app/blob/07f1a4ca5d3e85ff91dcf8adf5d2bfbeac2ab9e6/docs/developer/EXTERNAL_API.md) — the block-field guidance, field overview, manifest table, and HTTP error table now share one boundary: invalid declared fields return 422, while undeclared fields are pruned unless they are stored in the CRD's escape hatch.

## How to verify

1. Submit an `InteractiveGuide` with an undeclared top-level or `spec.manifest` key, capture the response headers, and confirm the write can return 200 or 201 with a `Warning: unknown field ...` header while a subsequent GET omits the key.
2. Submit an `InteractiveGuide` with a missing required declared field or an invalid declared value and confirm the API returns a 422 Kubernetes `Status` response.
3. Read the block-field guidance, field overview, manifest table, and HTTP error table together and confirm they describe the same declared-versus-undeclared behavior.

## Breaking changes

None. This is a documentation-only correction and changes no runtime, schema, or payload behavior.

Fixes #1676

### DIFF
--- a/docs/developer/EXTERNAL_API.md
+++ b/docs/developer/EXTERNAL_API.md
@@ -308,9 +308,16 @@ The wire format is the standard Kubernetes envelope:
 | `spec.blocks`              | yes      | Array of content blocks. The full schema is owned by the CUE definition in [grafana-pathfinder-backend/kinds/interactiveguide.cue](https://github.com/grafana/grafana-pathfinder-backend/blob/main/kinds/interactiveguide.cue) — that file is the source of truth. |
 | `spec.manifest`            | no       | Package metadata: grouping, sequencing, dependencies. Absent for content-only guides. See [Manifest](#manifest).                                                                                                                                                   |
 
-The CRD schema **is the validator**. Submit unknown fields and you'll
-get a `422 Unprocessable Entity` with a K8s `Status` envelope explaining
-which field is wrong.
+The CRD validates declared fields. A required declared field that is missing,
+or a declared value with an invalid type or value, returns a
+`422 Unprocessable Entity` with a K8s `Status` envelope.
+
+Unknown fields are **not rejected by default**. Kubernetes prunes them and
+may still return `200 OK` or `201 Created`. The API can emit a `Warning`
+response header, but callers that do not inspect headers receive no signal;
+`getBackendSrv()` does not surface it. For manifest extensions,
+`spec.manifest.additionalFields` is the only durable home for keys the CRD
+does not declare.
 
 ### Manifest
 
@@ -326,7 +333,7 @@ and what makes a path a path.
 | `author`           | no       | `{ name?, team? }`. The CRD declares no other keys; `upsert-learning-path.sh` moves any it finds (`email`, `github`, …) to `additionalFields.author` instead of dropping them. |
 | `category`         | no       | Free-form grouping label.                                                                                                                                                      |
 | `depends`          | no       | CNF (AND of ORs): an **array of arrays**. A single dependency is a singleton clause — `[["a"], ["b"]]` is "a AND b", `[["a","b"]]` is "a OR b". A bare string is not accepted. |
-| `additionalFields` | no       | Free-form escape hatch, `x-kubernetes-preserve-unknown-fields`. Anything not typed above goes here.                                                                            |
+| `additionalFields` | no       | Free-form escape hatch, `x-kubernetes-preserve-unknown-fields`. This is the only durable home for manifest keys not declared above.                                            |
 
 `recommends`, `suggests`, `provides`, `targeting`, `testEnvironment`,
 `startingLocation`, and the generated `stats` stamp have no typed home yet, so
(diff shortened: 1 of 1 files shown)


moxious (changes requested):
Thank you for the contribution!  Fix this item and this PR is mergeable.

1. [blocking] **docs-backendsrv-header-claim-false — New claim that getBackendSrv() cannot surface the Warning header is false** (medium · cross-cutting-architecture)
   Both edited passages assert unconditionally that `getBackendSrv()` does not surface the `Warning` header to callers. `FetchResponse` declares `readonly headers: Headers` as public API (`@grafana/runtime` `backendSrv.d.ts:90`), and this repository's own InteractiveGuide create/update client reads its responses through `getBackendSrv().fetch({ method: 'POST' | 'PUT' })` (`src/components/block-editor/hooks/useBackendGuides.ts:250,288,331,363`), so those callers do receive the header. The claim holds only for the shorthand `get`/`post`/`put`/`delete` helpers, which resolve to the parsed body. The PR's own 'How to verify' step 1 asks the reader to capture the response headers and confirm the `Warning: unknown field ...` header, contradicting the claim inside the same PR. The second occurrence also sits in the wire-format section addressed to external tooling using service-account tokens, where `getBackendSrv()` is not the caller at all.
   Required: Narrow the claim to the shorthand helpers -- `get`/`post`/`put`/`delete` resolve to the parsed body and drop the header, while `getBackendSrv().fetch()` returns `FetchResponse.headers` and can read it -- or drop the client-library assertion entirely and say only that a caller must inspect response headers to see the warning.

## Suggestions

1. [suggestion] **docs-field-validation-opt-in-unmentioned — Says unknown fields are not rejected 'by default' but never names the non-default** (low · testing-and-verification)
   The section's audience is CI, Terraform, and scripts writing guides unattended, which is exactly where a silent prune goes unnoticed. The only remedy offered is `spec.manifest.additionalFields`, which changes the payload rather than detecting the mistake. Kubernetes exposes `?fieldValidation=Strict` on write verbs to turn unknown-field pruning into a request error, and that is the actionable counterpart to the new 'by default' wording.
   Suggested: Measure whether the Pathfinder aggregator honours `?fieldValidation=Strict` against a real stack, then document it beside the 'not rejected by default' sentence. Do not assert it unmeasured, given the drift this PR is correcting.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1803
Summary: clarify unknown field pruning
Verdict: Request Changes
Results: 1 blocker, 1 non-blocking finding, 0 follow-ups

dvd233:
Thanks — fixed the blocking overbroad claim in `d3434724`. The docs now distinguish `getBackendSrv().fetch()`, whose `FetchResponse` exposes `headers`, from the shorthand `get`/`post`/`put`/`delete` helpers that return the parsed body. The external-tooling section now simply says callers must inspect response headers.

I left `fieldValidation=Strict` undocumented for now because I have not measured whether the Pathfinder aggregator honors it on a real stack, as suggested. Please take another look when convenient.

moxious (approved):
Thank you for the contribution!  LGTM

PR #1804, opened 2026-09-07
fix(snippets): classify and cache missing snippets

### LINKED ISSUE #1546: Unknown snippet ID reports network-error, misleading authors and defeating the no-negative-cache policy
`OnlineSnippetResolver` classifies every non-2xx response as a network error:

```ts
// src/snippet-engine/online-snippet-resolver.ts:47-51
```

So a `snippet-ref` pointing at an ID that does not exist reports `network-error`, indistinguishable from a genuinely transient failure.

## Two consequences

**1. The author-facing placeholder is misleading.** An unresolved ref renders an inert markdown placeholder naming the snippet ID and the error code (`src/snippet-engine/inline-refs.ts:17-22`, `:125-133`). An author who typo'd a snippet ID sees `network-error` and reasonably concludes the catalog fetch is flaky, not that the ID is wrong. That is the opposite of the debugging hint they need.

**2. It interacts badly with the no-negative-caching policy.** `CachingSnippetResolver` deliberately caches only successes (`src/snippet-engine/caching-snippet-resolver.ts:46-53`), which is right for transient failures. But because a permanently-missing snippet is also classed as transient, a typo'd ID is re-fetched on every single guide load, forever, with no backoff.

## Fix

Distinguish 404 from transport and 5xx failures, and give the placeholder a `not-found` code. Once the two are distinguishable, a short negative cache for `not-found` becomes safe and would stop the repeated fetches — worth considering in the same change, but the classification is the part that fixes the misleading message.

## Note

#1538 documents the current behaviour so the reference is not wrong in the meantime; that sentence should come out when this lands.

Refs #1530, #1538

Nikshay1:
## Summary

- classify HTTP 404 snippet responses as `not-found` instead of `network-error`
- cache `not-found` resolutions for the existing TTL while keeping transient failures uncached
- add regression coverage for the `not-found` placeholder and remove the documentation for the old behavior

## Verification

- focused snippet-engine tests
- all steps from the local pre-merge check passed
- `git diff --check`

Closes #1546

### DIFF
--- a/docs/developer/interactive-examples/json-guide-format.md
+++ b/docs/developer/interactive-examples/json-guide-format.md
@@ -1226,7 +1226,7 @@ A ref may sit at the top level of `blocks`, or nested inside a `section`, a `con
 
 Snippets cannot reference other snippets. The snippet schema rejects `snippet-ref` at every supported nesting depth, so a snippet body may contain any other block type but never a ref.
 
-If a ref cannot be resolved — unknown ID, catalog fetch failure — it is replaced with an inert markdown placeholder naming the snippet ID and an error code. The guide still renders, but nothing interactive appears in the ref's place. The resolver does not distinguish a missing snippet from a transient failure, so an unknown ID reports `network-error` rather than a not-found code.
+If a ref cannot be resolved — unknown ID, catalog fetch failure — it is replaced with an inert markdown placeholder naming the snippet ID and an error code. The guide still renders, but nothing interactive appears in the ref's place.
 
 ---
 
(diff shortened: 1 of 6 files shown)


moxious (approved):
Thanks for tracking this down and fixing it, @Nikshay1 — nice catch closing out #1546! Reviewed the resolver/cache reclassification and the new test coverage; no correctness, security, or reliability concerns.

PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1804
Summary: classify and cache missing snippets
Verdict: Approve

Once CI is green we'll get this merged. Thanks again! 🎉

Nikshay1:
Thanks, @moxious! Glad I could help with #1546, and I really appreciate the kind words. Happy to contribute, and I’m looking forward to helping out more! 🙌

PR #1813, opened 2026-09-07
fix: reject Pathfinder globals through window casts

### LINKED ISSUE #1776: Window-global lint rule misses two cast spellings it was written to catch
#1774 added a `no-restricted-syntax` selector rejecting untyped Pathfinder window globals. The selector is:

```
MemberExpression[object.type='TSAsExpression'][object.expression.name='window'][object.typeAnnotation.type='TSAnyKeyword'][property.name=/^__/]
```

It requires `object.expression.name === 'window'` and a `TSAnyKeyword` annotation, so three forms escape it:

1. `(window as unknown as Record<string, unknown>)[name]` — nested `TSAsExpression`, so `.expression` has no `.name`. This exact form was live at `src/global-state/content-key.ts:34` before #1774.
2. `(window as unknown as { __pathfinderPluginConfig?: DocsPluginConfig }).__pathfinderPluginConfig` — same reason. Live at `src/components/docs-panel/docs-panel.tsx:243` before #1774.
3. `(window as any)['__DocsPluginContentKey']` — computed access, so `property.name` is absent on the `Literal`.

Both of the first two are spellings #1774 itself had to delete, so they are demonstrated-reachable, not hypothetical.

**Currently latent.** No non-test `src/` file uses any of them at `d71ea54`; the only remaining window casts are `grafanaBootData` and `DEBUG_HIGHLIGHT_TARGET`, neither Pathfinder-owned. The defect is that the rule's message states the invariant as "do not bypass the typed Pathfinder window-global contract" while enforcing something narrower.

## Proposed fix

Add sibling selectors in the same `eslint.config.mjs` block for nested `TSAsExpression` window casts and for computed member access with a `__`-prefixed string literal. Per the #603 correction rule, add a probe for each to `src/validation/window-global-lint-config.test.ts` proving the new selectors fail when they should.

Follow-up to #1751, found reviewing #1774.

dvd233:
## Summary

The window-global lint contract misses nested casts and computed string access, allowing untyped Pathfinder-owned globals to bypass the rule. This adds sibling selectors for both AST shapes and regression probes while preserving typed direct access and the existing Grafana-owned cast.

## How to verify

1. Lint `const name = '__DocsPluginContentKey'; void (window as unknown as Record<string, unknown>)[name];` and `void (window as any)['__DocsPluginContentKey'];` as non-test TypeScript sources; confirm both emit the typed Pathfinder window-global diagnostic.
2. Lint `window.__pathfinderPluginConfig = undefined` and `(window as any).grafanaBootData`; confirm neither emits that diagnostic.

## Breaking changes

None. This change only tightens an internal lint guard and is fully reversible.

Fixes #1776

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -192,7 +192,14 @@ export default defineConfig([
         },
         {
           selector:
-            "MemberExpression[object.type='TSAsExpression'][object.expression.type='TSAsExpression'][object.expression.expression.type='Identifier'][object.expression.expression.name='window']",
+            "MemberExpression[object.type='TSAsExpression'][object.expression.type='TSAsExpression'][object.expression.expression.type='Identifier'][object.expression.expression.name='window'][property.name=/^__/]",
+          message:
+            'Do not bypass the typed Pathfinder window-global contract with a window cast. ' +
+            'Declare the global in src/types/window-globals.ts and access it through window directly.',
+        },
+        {
+          selector:
+            "MemberExpression[object.type='TSAsExpression'][object.expression.type='TSAsExpression'][object.expression.expression.type='Identifier'][object.expression.expression.name='window'][computed=true][property.type='Literal'][property.value=/^__/]",
           message:
             'Do not bypass the typed Pathfinder window-global contract with a window cast. ' +
             'Declare the global in src/types/window-globals.ts and access it through window directly.',
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -180,7 +180,21 @@ export default defineConfig([
           selector:
             "MemberExpression[object.type='TSAsExpression'][object.expression.name='window'][object.typeAnnotation.type='TSAnyKeyword'][property.name=/^__/]",
           message:
-            'Do not bypass the typed Pathfinder window-global contract with window as any. ' +
+            'Do not bypass the typed Pathfinder window-global contract with a window cast. ' +
+            'Declare the global in src/types/window-globals.ts and access it through window directly.',
+        },
+        {
+          selector:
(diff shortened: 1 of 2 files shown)


moxious (changes requested):
Thanks for tightening the window-global lint guard against double-cast bypasses -- catching these patterns keeps the typed contract in `src/types/window-globals.ts` actually enforceable.

## Merge contract

1. [blocking] **1813-window-cast-selector3-overreach -- New double-cast selector bans any property access, not just Pathfinder dunder globals** (low · correctness-and-reliability)
   The third new no-restricted-syntax selector matches any MemberExpression whose object is a double TSAsExpression rooted at `window`, with no filter on the accessed property's name or value -- unlike the other two selectors in this PR, which both require the property to match /^__/. It will reject any future double-cast window property access regardless of name, which is broader than the rule's own stated contract (protecting __-prefixed globals declared in src/types/window-globals.ts). No test in the diff exercises a non-dunder property after a double cast to confirm this is intentional.
   Required: Add the same __ prefix guard used in selectors 1-2 to selector 3 (covering both literal and identifier-named computed properties), or add a test asserting a non-dunder property access after a double cast is intentionally still rejected.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1813
Summary: reject Pathfinder globals through window casts
Verdict: Request Changes
Results: 1 blocker, 0 non-blocking findings, 0 follow-ups

Once the selector-3 scoping is narrowed (or a test confirms the broader ban is intentional), we'll get this merged.

moxious (approved):
Thanks for the quick turnaround -- scoping both new selectors down to the __-prefix guard, and adding the exact negative test that was missing, closes the gap cleanly.

PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1813
Summary: reject Pathfinder globals through window casts
Verdict: Approve

PR #1814, opened 2026-09-07
docs(package): align resolution contract with discriminated union

### LINKED ISSUE #752: Package engine cleanup: align identity-and-resolution.md with discriminated union types
## Background

Two small alignment gaps between the design spec and the implemented code were noted at the end of the package implementation epic. Neither is blocking; both are cosmetic.

### Resolution type shape in identity-and-resolution.md

`docs/design/package/identity-and-resolution.md` describes a flat `PackageResolution` shape. The implementation (`src/types/package.types.ts`) uses a discriminated union: `PackageResolutionSuccess | PackageResolutionFailure` with an `ok: true/false` discriminant.

The discriminated union is more expressive and should be considered the authoritative shape. The spec should be updated to match.

- [ ] Update `docs/design/package/identity-and-resolution.md` to document the discriminated union shape (`PackageResolutionSuccess`, `PackageResolutionFailure`, `ok: boolean`)

### Tab storage migration note

`PersistedTabData.packageInfo` (`{ packageId?, packageManifest? }`) is serialized to localStorage with `Record<string, unknown>` typing, which tolerates shape drift. No code change is needed now, but if the shape ever changes a migration step will be needed.

- [ ] If `PackageOpenInfo` shape changes in future: add a migration step in `src/utils/tab-storage-restore.ts` to normalize old localStorage data before use

## References

- `src/types/package.types.ts` — `PackageResolutionSuccess`, `PackageResolutionFailure`
- `docs/design/package/identity-and-resolution.md`
- `src/utils/tab-storage-restore.ts`

dvd233:
## Summary

Update the [package resolution design](https://github.com/dvd233/grafana-pathfinder-app/blob/35f8efaa0ef64de13e28f9414bb1ebdf98f783bd/docs/design/package/identity-and-resolution.md#L147) to match the authoritative `PackageResolutionSuccess | PackageResolutionFailure` type. The spec now requires callers to narrow on `ok`, documents the current loading and publication-verification options, and removes the unimplemented navigation shape.

## What to look at

- The revised fence captures the public resolution surface without copying implementation-only detail, and the following prose distinguishes successful and failed results while preserving the dynamic-registry contract.
- The issue's tab-storage item remains a no-op: every field in [`PackageOpenInfo`](https://github.com/dvd233/grafana-pathfinder-app/blob/35f8efaa0ef64de13e28f9414bb1ebdf98f783bd/src/types/content-panel.types.ts#L79) is optional, and [tab restore](https://github.com/dvd233/grafana-pathfinder-app/blob/35f8efaa0ef64de13e28f9414bb1ebdf98f783bd/src/components/docs-panel/utils/tab-storage-restore.ts#L196) can continue forwarding persisted package info without normalization.

## How to verify

1. Compare the documented success arm, failure arm, error codes, and options with the [authoritative type definitions](https://github.com/dvd233/grafana-pathfinder-app/blob/35f8efaa0ef64de13e28f9414bb1ebdf98f783bd/src/types/package.types.ts#L219).
2. Confirm callers are told to narrow on `ok` before reading success-only fields and that no `PackageNavigation` contract remains.
3. Confirm the paragraph about future registry evolution still says the frontend resolver and `PackageResolver` interface remain unchanged.

## Breaking changes

None. This is a documentation-only correction and is fully reversible.

Fixes #752

### DIFF
--- a/docs/design/package/identity-and-resolution.md
+++ b/docs/design/package/identity-and-resolution.md
@@ -141,10 +141,11 @@ The frontend plugin does **not** fetch, store, or reason about repository indexe
 
 The recommender shares its cached repository indexes between the package resolution endpoint and the recommendation endpoints. Both consumers benefit from the same periodic refresh cycle (~20 minutes). The recommendation engine uses the indexes for targeting, dependency graph analysis, and recommendation quality; the resolution endpoint uses them to map bare IDs to CDN content locations.
 
-All tiers return the same resolution shape:
+All resolver implementations return the same discriminated union:
 
 ```typescript
-interface PackageResolution {
+interface PackageResolutionSuccess {
+  ok: true;
   id: string;
   contentUrl: string;
   manifestUrl: string;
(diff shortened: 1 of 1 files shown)


moxious (approved):
Thanks for taking this one — doc-to-type drift is exactly the kind of thing that quietly misleads the next reader, and this section had two pieces of actively wrong documentation in it.

I checked the new fence against `src/types/package.types.ts` at current main and it matches field for field: both arms of the union, all five `ResolutionError.code` members, every optionality marker, `loadContent?: boolean | 'metadata-only'`, and an unchanged `PackageResolver.resolve` signature. Removing the `navigation` block is right too — I went looking for a reason to keep it, and there isn't one: `PackageNavigation` never shipped under `src/`, no consumer reads it, and the decision it described is already recorded in `decision-log.md` D24 and `learning-journeys.md`.

That covers both items in #752, including the tab-storage note — your reading is correct, it stays a no-op because every field in `PackageOpenInfo` is optional.

Approving. Thanks again.

Issue #1815, opened 2026-09-07
Step components ignore section-injected props; clear the no-unused-vars grandfather baseline


moxious:
Three step components receive props from the section and silently ignore them. This is the consumption-side twin of the drops tracked in #1506 and #1544: delivery is already contract-guaranteed, and nothing checks that the receiver honours what arrives.

It is filed as the pay-down issue for the grandfather baseline introduced when `@typescript-eslint/no-unused-vars` was enabled (#1744). That ratchet flagged all 19 bindings below; rather than delete them (which erases the evidence) or `_`-prefix them (which marks a real defect as intentional), the ratchet PR excluded these files and pointed here. **The goal is zero grandfathered entries.**

## Delivery is guaranteed, not incidental

`interactive-section.symmetry.tripwire.test.tsx:179-199` pins the per-type enhanced-prop surface. `COMMON_PROPS` — delivered to **every** step type — is:

```
stepId, isEligibleForChecking, onStepComplete, stepIndex,
totalSteps, sectionId, sectionTitle, disabled, resetTrigger
```

`step-type-registry.ts:135-144, 205-212, 274-282` does the injecting. So these props provably arrive.

## Inventory

| Prop | Component(s) | Delivered? | Verdict |
| --- | --- | --- | --- |
| `resetTrigger` | CodeBlockStep, TerminalStep, TerminalConnectStep | yes (`COMMON_PROPS`) | arrives, ignored — **bug** |
| `stepIndex` | same three | yes | arrives, ignored — **bug** |
| `totalSteps` | same three | yes | arrives, ignored — **bug** |
| `sectionTitle` | same three | yes | arrives, ignored — **bug** |
| `isEligibleForChecking` | TerminalConnectStep | yes | arrives, ignored — **bug** |
| `onStepReset` | CodeBlockStep, TerminalStep | **no** | props type over-promises — see below |
| `objectives` | InteractiveStep | author-set | dropped en route to `useStepChecker` — **bug** |
| `hints` | CodeBlockStep, TerminalStep | author-set | dropped en route to `useStepChecker` — **bug** |

## What each ignored prop costs, per the reference implementation

`InteractiveStep` honours all of them, so its usage defines what the others lose:

- `resetTrigger` → drives the step-reset effect (`interactive-step.tsx:468`). The three components have no reset support.
[...]

Issue #1786, opened 2026-09-05
Content Ecosystem Ratchet


moxious:
Our goal is to make the content ecosystem around Pathfinder increasingly tight, specific, high quality, and monitored.

To do this, there are several different prongs of attack.

- Good schema in guides; allow users to express things very specifically with features like requirements, objectives, `startingLocation` and the like.  This makes guides **specific in what they say/do**
- Good validation software in `pathfinder-cli` so that validate which does static analysis of guides makes it **hard to write guides that are unlikely to function correctly**
- Good testing software in the E2E suite that **proves guides work against a relevant test environment**, giving us confidence that we can suggest
- Good failure messages & documentation that **makes it easy for authors to do the right things**

The value of the overall package is derived indirectly from the total population of guides.  So if we make the content ecosystem both ridiculously good & safe, this is a major win condition for pathfinder.

This issue is held as an epic to link together associated work on these different prongs, and to loop in issues that are related so they are properly motivated by these design goals.

moxious:
## Work in flight and on the backlog, grouped by prong

A sweep of open and recently merged work that belongs to this epic. Each item is filed under the prong it advances. Open items are candidates for sequencing; merged items are recorded so the ratchet's history is visible in one place.

### Prong 1 — schema lets authors say precisely what they mean

Open:

- https://github.com/grafana/grafana-pathfinder-app/pull/1714 — declare a runtime Grafana floor in the guide manifest, so a guide can state the versions it works against
- https://github.com/grafana/grafana-pathfinder-app/pull/1768 — wire `requirements` / `objectives` / `skippable` through `ChallengeBlock`, fixing https://github.com/grafana/grafana-pathfinder-app/issues/1544 (they are declared but inert on every challenge block today)
- https://github.com/grafana/grafana-pathfinder-app/pull/1658 — `has-escalation-chain(s)` requirement check
- https://github.com/grafana/grafana-pathfinder-app/issues/1365 — support `targeting.requires` in the manifest schema so eligibility gates survive `build-repository`
- https://github.com/grafana/grafana-pathfinder-app/issues/1780 — tripwire: nested manifest keys are stripped before `repository.json`
- https://github.com/grafana/grafana-pathfinder-app/issues/1781 — tripwire: Go response struct drops fields the TypeScript package schema declares
- https://github.com/grafana/grafana-pathfinder-app/issues/1782 — tripwire: an author can publish a mode the target stack cannot run
[...]

Issue #1743, opened 2026-08-30
Detect unused exports and orphaned modules (step 3 of 3)


moxious:
Third and last step of the dead-code enforcement work. Steps 1 and 2 cover code that is unreachable *inside* a function and bindings that are never read *inside* a module. Neither can see the larger shape: a module nothing imports, or an export nothing consumes.

This step needs a human decision before anyone writes code. See "what a human must settle first" at the bottom.

## Lineage

- Extends the closed epic https://github.com/grafana/grafana-pathfinder-app/issues/603 ("Epic: Mechanical constraint enforcement for Agents"), which is about turning conventions agents can forget into checks that fail loudly — this adds reachability to the set of things a machine enforces rather than a reviewer notices.
- Preceding step: https://github.com/grafana/grafana-pathfinder-app/pull/1742 enabled the unreachable and vacuous code lint rules (step 1 of 3).
- A sibling issue covers `@typescript-eslint/no-unused-vars` (step 2 of 3).

## What we have today

`eslint-plugin-import` is not installed. Verified by `npx eslint --print-config src/lib/logging.ts`: zero rules in the effective config start with `import/`, and the package appears nowhere in `package.json`. So `import/no-unused-modules` is not an option we can just switch on — adopting it means adding a dependency and paying its flat-config and TypeScript-resolver setup cost.

## The measurement, and how far to trust it

An **unconfigured** `npx knip@5` run reports 13 unreferenced files, 638 unused exports and 717 unused exported types. These are an **upper bound, not a finding**. Do not quote the headline numbers as if they were verified dead code — an unconfigured run does not know this repository's entrypoints, and it is provably wrong about several of them.

Confirmed false positives in that same run:

- `src/cli/index.ts` — this is the published `pathfinder-cli` binary (`package.json#bin` points at `./dist/cli/cli/index.js`) and the target of ten or more npm scripts.
- `tests/e2e-runner/playwright.config.ts` — passed explicitly as `--config=` in `.github/workflows/e2e-runner-publish.yml`. Knip only knows about the root `playwright.config.ts`.
- `tests/e2e-runner/auth/auth.setup.ts` — reached through a `testMatch: /auth\/auth\.setup\.ts$/` regex inside that second config. No static import exists to find.

Three of thirteen wrong at the file level, for one reason: an unstated entrypoint set. That is the whole design problem in miniature.

Spot-checks that **do** hold, re-verified by grep against `main`:

[...]

PR #1816, opened 2026-09-07
chore(lint): re-enable @typescript-eslint/no-unused-vars with a grandfathered baseline

### LINKED ISSUE #1744: Re-enable @typescript-eslint/no-unused-vars and fix the 76 violations it surfaces
### The gap

`@grafana/eslint-config` explicitly turns this rule off. `node_modules/@grafana/eslint-config/index.js:86` reads:

```js
"@typescript-eslint/no-unused-vars": "off",
```

and `npx eslint --print-config src/lib/logging.ts` confirms it resolves to severity `0` in this repo today. The reasoning is sound as far as it goes: TypeScript's `noUnusedLocals` covers the same ground more reliably, and we do inherit `noUnusedLocals: true` from `@grafana/tsconfig` (`node_modules/@grafana/tsconfig/base.json`), confirmed present in `npx tsc --showConfig -p tsconfig.json`.

But `noUnusedLocals` is a *locals* flag. It structurally cannot see two categories:

- **Unused function parameters** — that is `noUnusedParameters`, a separate flag, and it is not set anywhere in our config chain. `npx tsc --showConfig -p tsconfig.json` shows `noUnusedLocals` and no `noUnusedParameters`.
- **Unused `catch` bindings** — `catch (err) {}` where `err` is never read. Neither compiler flag reports this.

There is a live example of the compiler being talked out of even the part it *can* see, at `src/components/EnableRecommenderBanner/EnableRecommenderBanner.tsx` lines 2-3:

```tsx
// @ts-expect-error - Icon kept available as core Grafana UI component
import { Button, Alert, Icon, useStyles2, Tooltip } from '@grafana/ui';
```

`Icon` is imported and never used. `noUnusedLocals` would raise TS6133 — and a `@ts-expect-error` comment exists solely to suppress it. ESLint would still flag it, because a TypeScript suppression directive does not silence an ESLint rule. This is the reliability hierarchy from #603 working in reverse: a compiler error got demoted to a comment.

### Measured violations

Re-measured on `main` at f5549ef, running the rule at the proposed configuration over `src/**/*.{ts,tsx}` and `tests/**/*.ts`:

**76 violations across 43 files.**

| Category | Count |
| --- | --- |
| Unused function parameters | 50 |
| Unused `catch` bindings | 19 |
| Genuinely unused values | 7 |

Only 10 of the 76 are in test files (`*.test.*`, `*.spec.*`, or under `tests/`), so this is overwhelmingly production code. 39 of the 43 files are under `src/`.

The 7 genuinely unused values in full:

```
src/cli/mcp/tools/finalize.ts:139            '_packageType'
src/components/EnableRecommenderBanner/...:3 'Icon'
src/docs-retrieval/json-parser-datasource-check.test.ts:137 '_dropped'
src/package-engine/app-platform-resolver.ts:212 '_synthetic'
src/types/cross-tab.types.ts:16              'requirements'
tests/block-editor-title-row.spec.ts:19      'NARROW_WIDTH'
tests/e2e-runner/guide-runner.spec.ts:25     'testIds'
```

### Proposed rule configuration

```js
'@typescript-eslint/no-unused-vars': ['error', {
  args: 'after-used',
  argsIgnorePattern: '^_',
  caughtErrors: 'all',
  caughtErrorsIgnorePattern: '^_',
}],
```

`args: 'after-used'` only flags parameters after the last used one, so a callback that must accept `(value, index)` and only reads `value` is fine. `^_` is the escape hatch for a parameter or `catch` binding that must exist but is deliberately unread.

Two configuration decisions the implementer has to make, both surfaced by the measurement:

1. **`varsIgnorePattern` is not in the config above**, so three `_`-prefixed *locals* (`_packageType`, `_dropped`, `_synthetic`) are flagged. Either add `varsIgnorePattern: '^_'` for consistency with the args and catch patterns, or delete those three bindings. Adding the pattern is the consistent choice; the underscore prefix already reads as "deliberately unused" everywhere else in the config.
2. **`tests/` is not linted at all today.** `npx eslint tests/open-close.spec.ts` returns *"File ignored because no matching configuration was supplied"* — the `files: ['./tests/**/*']` block in `.config/eslint.config.mjs` does not match, and nothing else declares TypeScript files under `tests/`. Covering the 5 violations there means adding a config block with a TypeScript parser for `tests/**`, which is a slightly wider change than a rule flip. Splitting `tests/` out to a follow-up and landing `src/` alone is a reasonable scope call.

### Why the ESLint rule rather than the tsconfig flag

Flipping `noUnusedParameters: true` in `tsconfig.json` looks like the cheaper move, and per #603's hierarchy a compiler error outranks a lint rule. Three reasons it is the wrong tool here:

- **It misses `tests/` entirely.** `tsconfig.json` extends `.config/tsconfig.json`, whose `include` is `["../src", "./types"]`. `tests/` is outside the compiler's view, so a tsconfig flag can never cover it. The resolved config confirms this.
- **It misses `catch` bindings entirely.** No TypeScript flag reports an unused `catch` parameter. That is 19 of the 76 — a quarter of the finding — invisible to the compiler under any configuration.
- **It has no escape hatch.** `noUnusedParameters` honours a leading underscore, but it has no equivalent of `args: 'after-used'`, so every intermediate parameter in a fixed-arity callback signature would need renaming.

`noUnusedParameters` remains worth enabling on its own merits later; it is not a substitute for this.

### Shape of the fix

The fixes are mechanical — prefix with `_`, delete the binding, or drop the parameter — but they touch 43 files. Per the repo's ratchet convention, this should land as **one ratchet PR: the config change plus only the fixes the ratchet forces**, and nothing else. No behaviour change is intended at any site; any violation whose fix is not obviously behaviour-neutral (the `cross-tab.types.ts` and `app-platform-resolver.ts` ones deserve a look) should be called out in the PR body rather than quietly bundled.

### Lineage

This extends phase 3 of [#603 — epic: mechanical constraint enforcement for agents](https://github.com/grafana/grafana-pathfinder-app/issues/603) (now closed), which is about mechanically enforcing constraints currently expressed only as prose: "do not leave dead bindings behind" is exactly such a constraint, enforced today only by review attention and by a compiler flag that cannot see two thirds of the cases.

It is step 2 of 3. The preceding step is [#1742 — chore(lint): enforce unreachable and vacuous code rules](https://github.com/grafana/grafana-pathfinder-app/pull/1742), which enables `no-unreachable`, `no-unreachable-loop`, `no-constant-condition`, `no-dupe-else-if` and `no-useless-return`, and which names this work in its own out-of-scope section. That PR is still open at time of filing; this one should land after it to avoid config conflicts in `eslint.config.mjs`.

Refs #603

moxious:
Closes #1744. Refs #603.

## Summary

`@grafana/eslint-config` turns `@typescript-eslint/no-unused-vars` off in favour of TypeScript's `noUnusedLocals`, which this repo does inherit. But `noUnusedLocals` is a *locals* flag, and it is structurally blind to two things:

- **unused function parameters** — that is `noUnusedParameters`, a separate flag, unset anywhere in our config chain
- **unused `catch` bindings** — no compiler flag reports these at all

A `@ts-expect-error` also silences the compiler while leaving an ESLint rule intact. That is how the dead `Icon` import in `EnableRecommenderBanner.tsx` survived: a real TS6133 got demoted to a comment.

The forward value is narrow and specific: a refactor that stops using a parameter now fails `npm run lint` instead of surviving review. That is the residue agents and humans both leave most often.

## This PR uses the grandfather-and-land pattern

Per the decision-of-record on #1744 (theme T6): **land the ratchet, fix only what the ratchet forces, grandfather the rest into a baseline, and file a follow-up to reach zero grandfathered entries.**

63 violations surfaced. **44 are fixed here. 19 are grandfathered** behind an explicit block in `eslint.config.mjs` plus one justified inline disable, with **#1815** filed to empty it.

The grandfathered 19 are not dead code. They are props that `step-type-registry.ts` provably injects — `interactive-section.symmetry.tripwire.test.tsx:179-199` pins the per-type prop surface as a contract — and that three step components never read. All three available "fixes" were wrong for a mechanical PR:

| Option | Why not |
| --- | --- |
| delete the binding | erases the only in-code evidence of a real pre-existing defect |
| prefix `_` | permanently marks a real defect as deliberate |
| honour the prop | behaviour change; for `objectives` it activates five dormant `completionReason === 'objectives'` branches (`interactive-step.tsx:353, 365, 1066, 1073, 1230`) and needs its own tests |

So they are recorded, not resolved. This follows the convention PR #1493 established for this bug family: **document the drop, don't erase it.**

## The baseline cannot grow

Review's one finding was that a file-level `rules: { ... : 'off' }` is not a baseline — it has no growth guard, unlike the repo's three other grandfather lists (`UNDESCRIBED_ESLINT_DISABLE_BASELINE`, `KNOWN_UNFORWARDED`, `ALLOWED_VERTICAL_VIOLATIONS`). Fixed: `unused-bindings-lint-config.test.ts` now walks every `.ts`/`.tsx` under `src/`, asks ESLint's own resolver (`calculateConfigForFile`) which files have the rule at non-error severity, re-enables it over exactly those, and asserts strict equality against an enumerated baseline — both the file set and the bindings each file exempts.

**Growth and drift both fail.** A 19th dead binding in those files is caught; a fourth exempted file cannot be appended silently; and clearing one of the 18 also fails until the baseline is updated, which is what drives #1815 to zero instead of letting it rot. Asking the resolver rather than parsing the config text means narrowing via `ignores`, via `files`, or via any future mechanism is all caught equally.

## What to look at

- `eslint.config.mjs` — the rule block and the grandfather block. `ignoreRestSiblings: true` was chosen deliberately **over** `varsIgnorePattern: '^_'`: it absorbs all six deliberate omit-style rest-sibling destructures with zero code edits, where `varsIgnorePattern` covers only the three that already carry an underscore and would force renames in deliberate code.
- `src/validation/unused-bindings-lint-config.test.ts` — drives the real ESLint API over the repo's own config, mirroring #1742's `dead-code-lint-config.test.ts`. Its clean-source probe includes an unused *leading* parameter, which is what actually discriminates `args: 'after-used'` from `args: 'all'`; verified by flipping the config to `'all'` and confirming failures.
- The 19 `} catch {` conversions — already the house idiom, with 236 existing uses.
- Two deliberate `_` retentions. `createImageLightbox`'s `_theme`: the call site comment claims "theme awareness" the function never implements, so the wiring is preserved for whoever fixes that latent bug, and removing it would cascade into a hook dependency array. `handleAutoDetectedMatch`'s `_detectedAction`: removing the sole parameter let prettier collapse an indent level across ~100 lines of untouched body.

## Deliberately out of scope

- **`tests/**` is not covered.** Nothing lints TypeScript there today, and bringing it under lint surfaces further errors from unrelated rules — one ratchet becomes three.
- `startSectionBlocking` still takes a `data` param nothing reads; `args: 'after-used'` does not flag it because a used param follows.
- `ContentProcessor`'s `onReady` / `contentType` prop declarations and pass-through wiring are now redundant plumbing that could go entirely.
- The lint-probe child-process harness is now a third near-identical copy alongside `dead-code-lint-config.test.ts` and `window-global-lint-config.test.ts`. Extracting a shared helper would touch two pre-existing test files.
- `.husky/pre-commit` is mode `100644`, so git skips it — and `lint-staged` only runs prettier anyway. **This and every other lint ratchet only fires on an explicit `npm run lint` or in CI.** Worth fixing separately; it would raise the value of the whole existing ratchet stack.

## How to verify

```sh
npm run lint            # 0 errors with the rule active on src/**
npm run typecheck
npx jest src/validation/unused-bindings-lint-config.test.ts --coverage=false   # 9 tests
```

The two claims worth checking rather than trusting, both confirmed on this branch:

- **The guard is not vacuous.** Add a dead binding to any of the three grandfathered files — the baseline test fails as growth. Remove one entry from `GRANDFATHERED` — it fails as drift. Both were run and both fail.
- **`args: 'after-used'` is really pinned.** Flip the option to `'all'` in `eslint.config.mjs` and two tests in the ratchet suite fail, because the clean-source probe carries an unused *leading* parameter.

## Related work

This is step 2 of a 3-step lint series, and it lands into an established bug family.

| Ref | Relationship |
| --- | --- |
| #1742 | step 1 of 3 — unreachable/vacuous code rules; this PR mirrors its config-block and companion-test pattern |
| #1743 | step 3 of 3 — unused exports and orphaned modules |
| **#1815** | **filed by this PR** — clears the grandfather baseline; the unhonoured step-component props |
| #1506 | same bug family — `content-renderer` drops `openGuide` / `formHint` / `validateInput` |
| #1544 | same bug family — `convertChallengeBlock` drops `requirements` / `objectives` / `skippable` |
| #1543 | same bug family, closed — the editor half of #1544 |
| #1493 | established the prop-forwarding tripwire and the document-the-gap convention this PR follows |
| #893 | InteractiveSection bug hunt; finding B3 overlaps the objectives-reset behaviour |
| #603 | epic: mechanical constraint enforcement for agents |
| #1786 | Content Ecosystem Ratchet epic |

## Reversibility

Fully reversible. Reverting restores the rule to `off` and puts back 44 dead bindings; no data, storage, or wire format is touched. No user-visible behaviour changes in this PR.


loglapa:
Coordination with #1832: this PR's `generateUserFriendlyError(..., _url)` rename touches the old `src/docs-retrieval/content-fetcher/fetch-raw.ts`, which #1832 turns into a re-export facade. I have carried the `_url` rename into the new `src/lib/content-fetcher/fetch-raw.ts` implementation on our branch.

I tested a temporary combination of both PRs, retaining the facade when resolving that one conflict. The moved Tier 1 files pass this PR's ESLint configuration, and the unused-binding baseline/architecture tests pass. Whichever lands second should keep the facade and the renamed parameter in the Tier 1 implementation; no baseline exception is needed.

dgehring-grafana (approved):
Simple enough, agentic review only asked that we pull the lint config into concerns.md, which I disagree with (and it's non-blocking) - LGTM

PR #1820, opened 2026-09-08
perf(review): cut the serial root load that dominates review wall clock

moxious:
## Summary

`/review` had been getting slow. Examining the skill from the outside — measuring its packets, running its gates against real merges, and simulating its planner — the wall clock turned out to be dominated by one thing: **the planner delegates the cheap work and keeps the expensive work.**

On PR #1791 (62 files, 16 activated concerns) it assigned 4 concerns to observation workers and 12 to `root`, which reviews them serially in the single most context-loaded agent. This PR fixes that and four smaller things found along the way, then adds two changes to what the review *outputs*: it stops the skill reporting low-value adjacent findings at all, and flattens the rendered sections.

## What to look at

**1. Dispatch ranking — `concern-context.mjs:307`.** `dispatchOrder` broke ties on registry position. The five always-on concerns are listed first, so they greedily filled both worker slots and every subsystem concern overflowed to `root`. Routed entries now carry their packet's own `category` and rank subsystem depth ahead of always-on breadth. No existing test broke, which was itself the finding — nothing covered dispatch ordering.

**2. Worker slots — `concern-context.mjs:306`, `:373`.** The general cap was a flat `2`. It bounded *parallelism*, not total work: overflow fell to `root` (`:316`, `:329`), so the cap bought nothing and serialized everything. It now scales with routed breadth — 2 up to six concerns, then one per three, capped at six. Reviews of six or fewer concerns produce byte-identical plans to today.

Fixes 1 and 2 are coupled. Ranking alone reorders without reducing; scaling alone delegates the wrong concerns.

**3. Verification livelock — `adversarial-policy.mjs:58`.** A high-risk defect holding exactly one verdict returned `needs_verification` with `{ role: null, count: 0 }`. `SKILL.md` §3 instructs the orchestrator to *"run exactly the returned role/count, append verdicts, and call again"* — count 0 appends nothing, so the next call returns the identical state, forever. Reachable whenever verdicts arrive one at a time: one skeptic returns while the other is still running, or one dies. The batch path already guarded `dispatch.count === 0` at `review-policy.mjs:248`; the per-observation path at `:220` did not.

**4. The gate did not gate — `contract-evolution-gate.mjs:194`.** `at_least_two_prior_semantic_prs` fired for **20 of 22** subsystem and cross-cutting concerns against a no-op diff. The threshold is ≥2 semantic PRs in 30 days on the concern's paths; this repository merges 91–157 first-parent commits per 30 days across six measured windows, so any concern with real paths clears it. `fix_heavy_history` fires for 2/22 and is the real discriminator.

The cost was never the 130 ms per gate call — it was that root must set `touches_anchor_with_consumers` from the diff for *every* fired gate (`SKILL.md:90`: the planner never infers anchor reach), then rank ~20 of them, to award exactly one specialist slot. `prior_semantic_pr_count` is untouched and still ranks gates that already fired, so `concern-context.mjs:244`/`:267` needed no change.

**5. Always-on doc pointers — `CONCERN_DETAILS.md`.** All five always-on concerns co-activate on every PR, so any doc on more than one of their rows was duplication in root's view, and two pointers were circular: `reversibility` named `CONCERNS.md` (which Route already reads at step 1) and `correctness` named `PR_REVIEW.md` (the skill's own standards doc). Reviewing a one-line docs fix was asking root to consider `E2E_TESTING_CONTRACT.md` and `systemPatterns.mdc` three times over.

`SKILL.md` now says explicitly that `load_docs`/`load_code` are an index of where evidence lives, not a read list.

**6. `follow_up` was the only sink — `review-policy.mjs:150`.** A `pre_existing` or `latent_unreachable` defect became a published follow-up at *any* severity; the rule never consulted severity. `low` defects reach disposition with no skeptic at all (`adversarial-policy.mjs:36`), and `dropped` existed only for a refuted verdict or round-three optional work. So a low-value problem in a file the PR merely touches had exactly one representable outcome: a durable follow-up that enters `next_deferred`, is re-checked every subsequent round, and downgrades the verdict to `Approve with Minor` (`review-report.mjs:94`). Scope-widening suggestions and nits were *promoted* into the same bucket rather than dropped.

New rule 2 drops an adjacent condition below `high` severity, applied before verification so it does not consume a skeptic either. Protected impact, `breaks_shipped_path`, and a `clearance_contradiction` are exempt — a severity mislabel must not silence real harm, and a reviewer must always be able to retract a wrong prior clearance. `high` and `critical` adjacent findings still surface, having passed two skeptics and a tiebreaker, and now state filing a separate issue as their suggested action rather than work for this PR.

**7. Rendered sections were mostly chrome — `review-report.mjs:411`.** Each group led with a Markdown heading over an explanatory subtitle (`## Merge contract` / "Fix this item and this PR is mergeable"; `## Follow-ups` / "These are tracked separately and do not block merge"), and optional work split across `## Suggestions` and `## Nits`. Findings now render under flat `Blockers:`, `Follow-ups:`, and `Suggestions & nits:` labels, each independently numbered, suggestions before nits, empty labels omitted. `## Review incomplete` stays the one heading the renderer emits, because it marks a report that states no mergeability at all.

## Why

Measured effect:

| | before | after |
|---|---|---|
| PR #1791 worker / root split | 4 / 12 | **12 / 4** |
| contract gate firing rate | 20 / 22 | **2 / 22** |
| always-on `load_docs` | 12 docs, 278 KB summed | **10 docs, 151 KB** |

Context for why this surfaced now: production script lines in this skill went **465 → 1,887 between 2026-08-11 and 2026-09-03**. `SKILL.md` prose was compressed 317 → 193 lines over the same period, so the instructions got tighter while the machinery they drive quadrupled — which hid the growth.

## How to verify

```bash
npm run test:review-contract   # 83 pass (was 76: 6 assertions updated, 7 tests added)
npm run test:governance        # 188 pass
npx prettier --check .cursor/skills/review/ docs/design/
```

Reproduce the planner change directly — build a plan with the 16 concerns PR #1791 activates and compare `coverage` before and after; the gate rate reproduces by running `contract-evolution-gate.mjs` for each subsystem and cross-cutting concern with `--base HEAD~1 --head HEAD`.

## Breaking changes

None for narrow reviews. Plans routing ≤6 concerns are unchanged. Two behavioral changes worth naming:

- **Large reviews now spawn up to 6 observation workers instead of 2.** Wall clock drops; total parallel token spend rises. If spend matters more than latency, lower `MAX_GENERAL_WORKERS` — the formula is isolated in `generalWorkerLimit`.
- **The contract-evolution specialist will run far less often.** That is the intent, but it does mean genuine contract-accretion in a concern with balanced fix/feat history now only gets caught by the anchor path, not the history path.
- **Low and medium adjacent findings no longer appear anywhere.** They are dropped, not deferred, so they leave no trace in the report or in `next_deferred`. If an adjacent problem matters, it has to be labelled `high` or carry protected impact. Adjacent entries already sitting in a prior review's `deferred` state are not cleaned up retroactively; they carry as bare IDs, which the recap already excludes from the results line.
- **The report's section headings changed.** Any tooling that greps a published review for `## Merge contract`, `## Follow-ups`, `## Suggestions`, or `## Nits` will stop matching. The trailing state marker and the four-line operator recap are untouched, so re-review state parsing is unaffected.

## Reversibility

Fully reversible. Every change is a constant, a one-line predicate, or a section label in the skill's own scripts, plus four table cells in `CONCERN_DETAILS.md`. Nothing persists state, no schema changed, and `category` is optional on routed entries so existing plan inputs remain valid.


leslie-heinzen (changes requested):
Blockers:

1. [blocking] **contract-evolution-ai-subsystem-worker-budget-anchor — The worker-budget owner conflicts with its recorded contract anchor** (high · ai-subsystem)
   `concern-context.mjs` and `SKILL.md` now allow six general workers, but `CONCERN_DETAILS.md:57` still records two. Contract scans receive the stale limit.
   Required: Update the `ai-subsystem` anchor to record the scaled general-worker budget and the specialist slot.

2. [blocking] **AI-PLAN-CATEGORY-INHERITED-PROPERTY — Inherited property names bypass the new category validator** (medium · ai-subsystem)
   `category in CONCERN_DEPTH` accepts inherited names such as `constructor`. These invalid values can take worker slots and move valid concerns to serial root review.
   Required: Use an own-property or `Set` membership check. Add rejection cases for inherited property names.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1820
Summary: cut the serial root load that dominates review wall clock
Verdict: Request Changes

moxious:
Thanks for the review, @leslie-heinzen — both fixed in 094c7d39:

1. **`contract-evolution-ai-subsystem-worker-budget-anchor`**: reworded the `ai-subsystem` anchor cell in `CONCERN_DETAILS.md:57` to record the scaled formula — 2 general workers for ≤6 routed concerns, +1 worker per 3 additional concerns up to a cap of 6 — plus the unchanged specialist slot and the unchanged incremental-mode limit (1 general worker + 1 specialist).

2. **`AI-PLAN-CATEGORY-INHERITED-PROPERTY`**: swapped the `category in CONCERN_DEPTH` check (and the parallel lookup in `concernDepth`) for `Set.has()` on `Object.keys(CONCERN_DEPTH)`, so inherited property names no longer validate. Added a test asserting `constructor`, `toString`, `hasOwnProperty`, and `__proto__` are all rejected as unknown categories.

Verified: `npm run test:review-contract` (84 pass, was 83), `npm run test:governance` (188 pass), `npx prettier --check .cursor/skills/review/ docs/design/` (clean), `npm run lint:fix` (clean).

Ready for re-review.

leslie-heinzen (approved):

PR #1821, opened 2026-09-08
fix(e2e): add atomic milestone reset contract

leslie-heinzen:
## Summary

Shared E2E milestone runs now reset the previous guide through a versioned, plugin-owned browser control before they close its tab. This removes the timing-dependent reset UI from the primary transition path while retaining that path as a rollout fallback.

Transition faults now use `TRANSITION_FAILED` and a bounded `transitionKind`. Reports preserve the original transition classification for every unrun milestone.

PR #1821 replaces #1728's runner-only reset ownership and unchanged plugin frontend clauses. `resetGuideProgress` is the plugin-side reset authority for the normal UI and the URL-scoped browser control.

## What to look at

- **Docs panel reset contract** — The [browser capability](https://github.com/grafana/grafana-pathfinder-app/blob/6cd662aea410fac329c8d65275485501373a5ef5/src/components/docs-panel/hooks/useE2EResetGuideCapability.ts) exists only for the exact `bundled:e2e-test` URL. The [shared reset helper](https://github.com/grafana/grafana-pathfinder-app/blob/6cd662aea410fac329c8d65275485501373a5ef5/src/components/docs-panel/hooks/resetGuideProgress.ts) clears scoped progress and the in-memory completion cache before it dispatches the progress-cleared event.
- **Shared milestone runner** — The [replacement flow](https://github.com/grafana/grafana-pathfinder-app/blob/6cd662aea410fac329c8d65275485501373a5ef5/tests/e2e-runner/utils/guide-runner/milestone-replacement.ts) prefers control version 1 and checks empty scoped storage before tab closure. It retains the legacy fallback when the capability is absent. All injected E2E guides deliberately share one URL, so the stable tab ID prevents closure of the wrong tab.
- **Report and publish contract** — The [report schema](https://github.com/grafana/grafana-pathfinder-app/blob/6cd662aea410fac329c8d65275485501373a5ef5/src/cli/e2e/schemas/e2e-report.schema.ts) moves to version 1.1.0. The [publish workflow](https://github.com/grafana/grafana-pathfinder-app/blob/6cd662aea410fac329c8d65275485501373a5ef5/.github/workflows/e2e-runner-publish.yml) checks the exported version before image publication.

## Why

The #1775 fallback depended on a visible **Reset guide** control. Cloud cycles produced eight reset-button timeouts and four unsafe post-close states across 12 learning journeys.

The plugin owns the completion cache and storage lifecycle. A mounted plugin method can clear that state in one acknowledged operation before the runner closes the tab.

## How to verify

1. Run the shared-session fixture against a local Grafana instance with the branch plugin. Confirm that both milestones pass without a page reload.
2. Complete an injected `bundled:e2e-test` guide. Transition to the next milestone and confirm that only the prior guide progress is cleared.
3. After the plugin release, rerun the 12 affected learning journeys on Cloud stacks. Confirm that no milestone stops during reset or tab replacement.

Local validation used temporary local-tier copies of the 12 failing journey structures. The copies preserved all journey IDs, 96 milestone IDs, milestone order, and chain boundaries.

The temporary guides replaced Cloud-dependent actions with deterministic local actions. The first milestone wrote unsaved browser state, and every later milestone required that state.

The branch plugin passed all 96 milestones and 84 transitions. A capability-only guard prevented use of the legacy fallback during these runs.

A negative HAProxy mirror run used the old #1775 plugin. It passed milestone one, then failed transition one with `TRANSITION_FAILED`, `reset-ambiguous`, and an unavailable-control error.

The temporary mirror and guard were removed after validation. They are not part of this PR.

The full repository gate, plugin builds, CLI build, shared-session fixture, and runner-image smoke tests also passed locally.

## Breaking changes

The E2E report schema version changes from 1.0.0 to 1.1.0. Deploy consumer support before publishing a runner digest that emits the new schema.

The plugin browser control is additive. The runner retains the legacy fallback during rollout.

## Reversibility

The browser control and primary runner path can be reverted together. Consumers must retain version 1.1.0 support for reports that already use the new transition fields.

## Out of scope

- Deployment of the version 1.1.0 consumer change in `pathfinder-e2e-platform`.
- Final Cloud verification of the 12 original learning-journey bodies. This verification requires the released plugin.

Co-Authored-By: Warp <agent@warp.dev>


moxious (dismissed):
## Merge contract

1. [blocking] **contract-anchor-retains-superseded-runner-only-clauses — The rewritten contract anchor still asserts the reset is runner-only and the installed plugin frontend is unchanged** (medium · cli-and-e2e-runner)
   The `cli-and-e2e-runner` row in `docs/design/CONCERN_DETAILS.md` was updated to evidence `#1372 → #1821` and its report-version clause replaced, but it still contains, verbatim and unchanged from base, both `the runner-only progress reset` and `The installed plugin frontend and bundled loader remain unchanged.` The second sentence is affirmatively false at this head: the PR adds `useE2EResetGuideCapability.ts`, wires it at `docs-panel.tsx:1138`, exports it from the hooks barrel, adds `resetGuideButton` to `src/constants/testIds.ts:46`, renders that id in two components, and declares `__pathfinderE2E` on `Window`. The author already accepted the sentence is dead, deleting the equivalent line from `docs/developer/E2E_TESTING_CONTRACT.md` and replacing `This flow is runner-only.` in `E2E_TESTING.md` — only the anchor row keeps it, so the two documents now contradict each other and the stale one is the registry the review gate reads. This is machine-extracted: `concern-context.mjs --worker cli-and-e2e-runner` emits the row verbatim as `contract_anchor.contract`, and the skill mandates that packet as the only contract text an observation worker sees, so a future reviewer will be told the plugin frontend is out of scope for this concern when it no longer is. `npm run test:review-contract` only exercises the extractor scripts and cannot detect a false claim. The same registry-versus-doc divergence was the blocking finding on the immediate predecessor, #1775.
   Required: Replace both stale sentences in this PR: state that `resetGuideProgress.ts` is the single plugin-side reset authority reached by both `useContentReset` and the URL-scoped versioned `window.__pathfinderE2E` control, that `milestone-replacement.ts` prefers that control and retains the legacy UI-click and residue paths only while it is absent, and that this replaces those clauses of the #1728 contract. Per AGENTS.md 'Extending existing capabilities', also state explicitly in the PR body whether this follows, extends, or replaces the established contract — the body currently does neither.

## Follow-ups

1. [follow_up] **orphaned-step-comments-after-reset-extraction — Three numbered step comments in useContentReset now describe code that moved to resetGuideProgress** (low · correctness-and-reliability)
   After the extraction, `useContentReset.ts:56-64` holds nine consecutive lines of `Step 2` / `Step 2b` / `Step 3` comments with no code beneath them, followed by the single `await resetGuideProgress(progressKey)` call. At base each statement sat directly under its comment. The destination file carries none of the rationale, so the one genuinely load-bearing invariant — that clearing storage alone does not invalidate the in-memory completion snapshot, which is why eviction is required — is now separated from the code it protects. The function's JSDoc still enumerates steps 1-4 as though this function performed them. AGENTS.md requires trimming bad-shape comments on touch.
   Follow-up: Delete the three step comments and correct the JSDoc enumeration; move a one-line version of the eviction invariant next to `evictContentCache(contentKey)` in `resetGuideProgress.ts`.

2. [follow_up] **reset-testid-declared-runner-contract-but-unused — docs-panel-reset-guide-button is declared part of the shared runner contract but no runner code references it** (low · reversibility-and-one-way-door)
   `E2E_TESTING_CONTRACT.md` now states that the reset control uses `docs-panel-reset-guide-button` and that both test IDs are part of the shared runner contract. A full-tree grep on the PR branch finds the id only in `src/` — `testIds.ts`, the two components, and their tests — and nothing under `tests/`. The runner still selects the legacy control by accessible role at `milestone-replacement.ts:371` and `:470`, unchanged from base, and `E2E_TESTING.md` in the same PR says the fallback uses the accessible control, contradicting the contract doc. A test ID promoted to a public runner contract that nothing on the runner side consumes is a compatibility obligation with no enforcement: it can be renamed and every runner test still passes.
   Follow-up: Either switch the legacy fallback selector to `page.getByTestId(testIds.docsPanel.resetGuideButton)`, or downgrade the doc sentence to say the id is published for future runner use while the current selector is accessible-name based. If you take the first option, note that `LearningJourneyMilestoneToolbar.tsx:217` renders the same id in its kebab overlay, so a `getByTestId` selector can match twice on a package-backed interactive journey tab with the kebab open.

3. [follow_up] **e2e-guide-url-constant-triplicated-without-parity-assertion — bundled:e2e-test now has three independent definitions with nothing asserting they agree** (low · guide-schema-and-contracts)
   `useE2EResetGuideCapability.ts:7` adds a third literal definition of `bundled:e2e-test`, alongside the runner's own `E2E_GUIDE_URL` at `milestone-replacement.ts:9` and the loader's independent resolution of the `e2e-test` content id in `bundled.ts:29-30`. The exposure gate is exact-match on the plugin side and the runner reads the global with no URL cross-check. `docs-panel.contract.test.tsx` pins the shared test IDs but nothing pins this string across the plugin and runner boundary. If the plugin copy drifts, the control silently never appears and the runner quietly takes the legacy path; if the runner copy drifts, storage inspection targets a different content key than the reset does. Neither failure surfaces as an error.
   Follow-up: Import the constant from one shared Node-safe home and add a parity assertion beside the existing test-ID assertions in `docs-panel.contract.test.tsx`.

## Suggestions

1. [suggestion] **reset-guard-covered-only-by-non-production-state-channel — The new control's active-guide guard is covered only through a state channel production never uses** (medium · correctness-and-reliability)
   `resetActiveGuide` guards on `getActiveTabUrl() !== E2E_GUIDE_URL`. `setActiveTabUrl` has no production caller anywhere in `src/`, so in the running app that guard resolves exclusively through the legacy `window.__DocsPluginActiveTabUrl` fallback written by `useGlobalActiveTabExposure`. Both new test files drive the guard with `setActiveTabUrl` and never write the window global, so the only production resolution path has no coverage. `content-key.ts` documents that fallback as load-bearing and slated for removal once remaining consumers migrate; that routine cleanup would make `getActiveTabUrl()` return undefined, causing the control to reject every reset, while all five new unit tests stayed green. The install predicate and the guard cannot disagree today — both derive from `activeTabCurrentUrl || activeTabBaseUrl` and the exposure hook's layout effect runs first in the same component — so this is a coverage gap, not a live bug.
   Suggested: Either have `useGlobalActiveTabExposure` also call `setActiveTabUrl` so the typed channel is real, or add one test that sets `window.__DocsPluginActiveTabUrl` instead of calling `setActiveTabUrl`, pinning the path production actually takes.

2. [suggestion] **capability-path-failure-modes-without-tests — Three newly reachable classification branches have no test, including the AUTH_EXPIRED versus TRANSITION_FAILED precedence** (medium · testing-and-verification)
   A control present but with `resetActiveGuide` missing or non-callable returns `rejected` at `milestone-replacement.ts:140-142`, and the test harness's `capability` option always installs a `resetActiveGuide` mock, so that branch is never entered. A non-numeric `version` produces `unsupported version undefined` at :137-138; only `version: 2` is tested. Separately, `playwright-runner.ts:286-292` orders `AUTH_EXPIRED` ahead of `fatalTransition`, so a `TRANSITION_FAILED` last partial result combined with an `AUTH_EXPIRED` abort file drops `transitionKind` entirely, and neither existing recovery test covers that co-occurrence. The first two are exactly the shapes a partially-implemented or renamed control produces, and they are the branches that decide fatal-versus-fallback; the third decides which structured code and detail a consumer receives when two abort causes race. All affected suites pass at head, so these are coverage gaps rather than failures.
   Suggested: Extend the `replacementHarness` `capability` option with an omitted `resetActiveGuide` and a non-numeric `version` case, and add a `processPlaywrightChainResults` case carrying both an `AUTH_EXPIRED` abort file and a `TRANSITION_FAILED` last partial result, asserting the resulting `errorCode` and `transitionKind`.

## Nits

1. [nit] **recovered-milestone-message-contradicts-its-transition-code — Partial-result recovery stamps TRANSITION_FAILED onto milestones whose errorMessage still says no report was written** (low · cli-and-e2e-runner)
   `playwright-runner.ts:286-292` sets `errorCode` to `TRANSITION_FAILED` and copies `transitionKind`, but `errorMessage` at :293-294 keeps the generic 'ended before this milestone produced a result report' text. The shared-chain equivalent keeps the two aligned: `unrunSharedSessionResult` selects a fatal-transition message on that branch. The new recovery test asserts only outcome, code, kind, and results, so the mismatch is unpinned. An operator reading a recovered report sees a `reset-ambiguous` transition code beside prose asserting the milestone simply never produced a report — the exact ambiguity `transitionKind` was added to remove.
   Suggested: Use the transition-specific message on the `fatalTransition` branch, mirroring `unrunSharedSessionResult`, and assert the message in the new recovery test.

2. [nit] **post-close-sweep-blames-the-legacy-reset-on-the-capability-path — The post-close safety block runs on the capability path but reports failures as an acknowledged legacy reset** (low · cli-and-e2e-runner)
   `milestone-replacement.ts:514-531` gates the post-close block on `storage.hasStoredCompletion` alone rather than on `!usedE2ECapability`, so it also runs after a successful plugin-owned reset — which is good, since it keeps the bounded absence check and residue sweep on the primary path. But its catch throws with the message 'The acknowledged legacy reset did not reach a safe post-close state', naming the legacy path even when the capability path ran, and `E2E_TESTING_CONTRACT.md` lists that bounded check and residue re-clear as steps 5 and 6 of the legacy sequence only. A capability-path failure here therefore points a maintainer at the wrong reset mechanism.
   Suggested: Make the message reflect which path ran, and record in `E2E_TESTING_CONTRACT.md` that the bounded absence check and residue re-clear apply to both paths.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1821
Summary: add atomic milestone reset contract
Verdict: Request Changes
Results: 1 blocker, 4 non-blocking findings, 3 follow-ups

leslie-heinzen:
Addressed the blocker, suggestions, nits, and follow-ups 1 and 2 in three commits:

- `2ad25ac0` updates the living reset contract. It now describes the plugin reset authority, browser control, and legacy fallback.
- `82d4214a` adds production-guard and capability-error coverage. It also aligns recovery messages, error precedence, and post-close diagnostics.
- `e8d66ee4` removes stale reset comments. It moves the cache invariant to its owner and makes the legacy runner use `resetGuideButton`.

Deferred:

- The duplicated E2E guide URL identity and parity assertion remain for a separate PR. That work spans the frontend, bundled loader, and Node runner. It is not required for the atomic reset contract.

Validation passed:

- Focused reset tests
- TypeScript
- ESLint
- Prettier
- Review-contract tests
- `git diff --check`

Jayclifford345 (approved):

## Follow-ups

1. [follow_up] **report-1-1-old-schema-compatibility-claim — The 1.1.0 report is not accepted by the older schema as the contract claims** (medium · guide-schema-and-contracts)
   docs/developer/E2E_TESTING.md says an orchestrator can validate a newer report with an older open-world schema, but the 1.0.0 schema requires schemaVersion 1.0.0 and its closed errorCode enum lacks TRANSITION_FAILED. That claim is false, although the immutable image is not automatically adopted and the documented consumer-first rollout keeps this from causing current runtime harm.
   Follow-up: In a follow-up, state that report schemas are version-specific and that consumers must select validation by schemaVersion before using a 1.1.0 runner. Add a compatibility test that proves the documented rule.

2. [follow_up] **e2e-guide-url-constant-triplicated-without-parity-assertion — The E2E guide identity still has independent plugin and runner owners** (low · guide-schema-and-contracts)
   The plugin capability, Playwright runner, and bundled loader still own separate representations of bundled:e2e-test, while the contract test pins only shared test IDs. They agree at this head, but a future one-sided rename can silently disable the capability or inspect the wrong storage key.
   Follow-up: Move the identity to one Node-safe shared constant and add a parity assertion for the plugin, loader, and runner.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1821
Summary: add atomic milestone reset contract
Verdict: Approve with Minor
Results: 0 blockers, 0 non-blocking findings, 2 follow-ups

Jayclifford345 (approved):
Mergeable with a few nits from automated reviews to clear up after but all good to test in release

PR #1823, opened 2026-09-08
fix(husky): make the pre-commit hook run, and give it the right work

### LINKED ISSUE #1817: Nothing runs at commit time: .husky/pre-commit is not executable and lint-staged omits eslint
`.husky/pre-commit` has never run. It is tracked at mode `100644`, and git silently skips a hook that is not executable — it only prints a hint:

```
$ git commit
hint: The '.../.husky/pre-commit' hook was ignored because it's not set as executable.
```

```
$ git ls-files -s .husky/pre-commit
100644 c61f95f20a4c8e885166db569f6a5790eef53674 0	.husky/pre-commit
```

The hook's contents are:

```sh
npx lint-staged
npm run typecheck
npm run test:governance
```

So `typecheck` and `test:governance` have never gated a commit.

## Why this matters more than it looks

There is a second, independent gap. Even with the hook executable, `lint-staged` runs **only prettier**:

```json
"lint-staged": { "*.{ts,tsx,js,mjs,json,yaml,md}": "prettier --config .prettierrc.js --write" }
```

No `eslint`. So **every lint ratchet this repo has built fires only on an explicit `npm run lint` or in CI** — #1742 (unreachable/vacuous code), #1744 (unused bindings), #1752 (undescribed eslint-disables), #1524 (switch exhaustiveness), #1751 (window globals), plus the import-boundary rules. The mechanical enforcement those PRs installed under #603 works, but its feedback loop is minutes-to-CI rather than seconds-to-commit.

That is the whole value proposition of #603 — mechanically enforce constraints so an agent cannot leave residue — running at a fraction of its intended speed. Fixing this raises the value of the existing ratchet stack more than any single new ratchet does.

## Fix

1. `chmod +x .husky/pre-commit` and commit the mode change (`git update-index --chmod=+x`).
2. Decide whether `eslint` belongs in `lint-staged`. Adding `eslint --fix` for `*.{ts,tsx}` gives per-file lint at commit time; the tradeoff is commit latency, since this repo's config is type-aware.
3. Check the hook's steps are still the ones wanted at commit time. `npm run typecheck` over the whole project on every commit may be too slow now — measure before enabling, and consider moving it to pre-push.

Worth measuring the resulting commit latency before landing, since a hook slow enough to be bypassed with `--no-verify` is no better than one git ignores.

Refs #603

dchaudhari7177:
Closes #1817. Refs #603.

Round 2: this is option 2 from the previous round — the mode fix and the lint-staged split, without the `pre-push` hook. All three blockers were properties of that hook and leave with it.

## 1. The hook never ran

`git ls-files -s .husky/pre-commit` was `100644`, so git skipped it and only printed a hint. It is now committed `100755`.

`.gitattributes` had `eol=lf` for `*.sh`, `*.bash`, `*.zsh` and the source extensions, but a git hook has no extension, so nothing covered `.husky/*`. A hook checked out with CRLF fails to execute — the same silent skip, one layer down. Added `.husky/* text eol=lf`.

## 2. `lint-staged` ran only prettier

The lint ratchets in this repo (#1742, #1744, #1752, #1524, #1751, plus the import-boundary rules) apply to `src/`, and fired only on `npm run lint` or in CI. `eslint --cache --fix` now runs on `*.{ts,tsx,js,mjs}` ahead of prettier; `*.{json,yaml,md}` keeps prettier alone. So the coverage this buys is on `src/`, which is where the ratchets live — it is not repo-wide.

## 3. Where each step belongs — measured first

The issue asks to measure before enabling, so I did, on this checkout with a warm eslint cache:

| step | time |
| --- | --- |
| `eslint`, one changed file | 4s |
| `eslint`, five changed files | 14s |
| `prettier`, one file | 2s |
| `npm run typecheck` (whole project) | 25s |
| `npm run test:governance` | 27s |

The hook as written was a **52s commit before eslint was even added** — squarely in "gets bypassed with `--no-verify`" territory, which the issue rightly calls no better than a hook git ignores.

So `pre-commit` runs `lint-staged` only. A real commit on this branch takes ~3s.

`typecheck` and `test:governance` are **not** relocated to a `pre-push` hook here. Round 1 did that, and the review found the reason not to: `npm run test:governance` includes `src/validation/import-graph.test.ts`, which has 4 separator-dependent assertions and fails on Windows. Gating `git push` on a suite that cannot pass off posix is how `--no-verify` becomes a habit. Those two stay on `npm run check` and CI until the separator problem is fixed on its own.

## Proof it now gates

Staged a `.ts` file and ran lint-staged for real:

```
export const probe   =  {a:1,   b:2};      # staged
→ eslint --cache --fix, then prettier
export const probe = { a: 1, b: 2 };       # committed
```

And against a ratchet rule:

```
import moment from 'moment';

✖ eslint --cache --fix
  1:1  error  'moment' import is restricted from being used  no-restricted-imports
✖ 1 problem (1 error, 0 warnings)
✖ Failed to run tasks for staged files!
⋯ Reverting to original state because of errors…
```

The commit does not land. Before this change it did.

## Docs

`LOCAL_DEV.md:146` and `GETTING_STARTED.md:99` both described lint-staged as prettier-only; both now name `eslint --fix` then prettier on `*.{ts,tsx,js,mjs}` and prettier alone on `*.{json,yaml,md}`. Both troubleshooting headings are widened to "Husky pre-commit hook fails or blocks the commit", since the failure people will now hit is a lint error stopping a commit that used to go through.

`GETTING_STARTED.md:40` said `npm install` triggers husky hook setup. `.npmrc` sets `ignore-scripts=true`, so it does not, and `LOCAL_DEV.md:24` already said so. It now shows `npm run prepare`.

## Not done, deliberately

The `--no-warn-ignored` follow-up. The flag *suppresses* the ignored-file notice, so adding it is what would make an uncovered staged file pass silently. On this repo's eslint 9.39.1:

```
$ eslint --cache package.json
  0:0  warning  File ignored because no matching configuration was supplied     exit 0

$ eslint --cache --no-warn-ignored package.json
                                                                                exit 0   # silent
```

Left off, so the notice surfaces. Say the word if the intent was to quiet it instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -175,7 +175,11 @@
     }
   },
   "lint-staged": {
-    "*.{ts,tsx,js,mjs,json,yaml,md}": "prettier --config .prettierrc.js --write"
+    "*.{ts,tsx,js,mjs}": [
+      "eslint --cache --fix",
+      "prettier --config .prettierrc.js --write"
+    ],
+    "*.{json,yaml,md}": "prettier --config .prettierrc.js --write"
   },
   "packageManager": "npm@12.0.2"
 }
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,6 +2,9 @@
 *.sh text eol=lf
 *.bash text eol=lf
 *.zsh text eol=lf
+# Git hooks have no extension, so no rule above covers them. Checked out
+# with CRLF they fail to execute.
+.husky/* text eol=lf
 
 # Ensure consistent line endings for other text files
 *.ts text eol=lf
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,3 +1,5 @@
+# Fast, per-file checks only. Measured on this repo: eslint over one changed
+# file is ~4s warm, prettier ~2s. `npm run typecheck` (25s) and
+# `npm run test:governance` (27s) are whole-project; a 52s commit gets bypassed
+# with --no-verify, which is no better than the hook git was silently skipping.
 npx lint-staged
-npm run typecheck
-npm run test:governance
--- a/docs/developer/GETTING_STARTED.md
+++ b/docs/developer/GETTING_STARTED.md
@@ -37,7 +37,8 @@ node -v && npm -v && go version && docker --version && mage --version
 ### Build and run
 
 ```bash
-npm install            # installs frontend deps and triggers husky hook setup
+npm install            # frontend deps only; .npmrc sets ignore-scripts=true
+npm run prepare        # install the husky git hooks (once, after a fresh clone)
 npm run build:all      # frontend + Linux/ARM64 backend (what docker-compose mounts)
 npm run server         # build:all + docker compose up --build
 ```
@@ -96,7 +97,9 @@ The project ships with `.eslintrc`, `.prettierrc.js`, and `tsconfig.json` config
 - **Prettier — Code formatter** (`esbenp.prettier-vscode`) — set as default formatter, format on save.
 - **Go** (`golang.go`) — for backend work.
 
-A husky pre-commit hook runs `lint-staged`, which applies Prettier to staged `.ts`/`.tsx`/`.js`/`.json`/`.yaml`/`.md` files automatically.
+A husky pre-commit hook runs `lint-staged`, which applies `eslint --fix` then Prettier to
+staged `.ts`/`.tsx`/`.js`/`.mjs` files, and Prettier alone to staged
+`.json`/`.yaml`/`.md` files.
 
 ## First-week reading list
 
@@ -135,7 +138,7 @@ go install github.com/magefile/mage@latest
 export PATH="$PATH:$(go env GOPATH)/bin"
 ```
 
-### Husky pre-commit hook fails
+### Husky pre-commit hook fails or blocks the commit
 
 `npm run check` reproduces the failure locally. Fix the root cause; do not bypass with `--no-verify`.
 
--- a/docs/developer/LOCAL_DEV.md
+++ b/docs/developer/LOCAL_DEV.md
@@ -143,7 +143,8 @@ npm run prettier-test    # check formatting only
 npm run lint:go          # golangci-lint via mage
 ```
 
-Husky runs `lint-staged` on commit (Prettier on staged `.ts`/`.tsx`/`.js`/`.json`/`.yaml`/`.md`).
+Husky runs `lint-staged` on commit: `eslint --fix` then Prettier on staged
+`.ts`/`.tsx`/`.js`/`.mjs`, and Prettier alone on staged `.json`/`.yaml`/`.md`.
 
 ## IDE setup
 
@@ -188,7 +189,7 @@ go install github.com/magefile/mage@latest
 export PATH="$PATH:$(go env GOPATH)/bin"
 ```
 
-### Husky pre-commit hook fails
+### Husky pre-commit hook fails or blocks the commit
 
 `npm run check` reproduces the failure locally. Fix the underlying issue; do not bypass with `--no-verify`.
 


.husky/pre-push
```diff
@@ -0,0 +1,5 @@
+# Whole-project gates. These ran on every commit, where their combined ~52s
+# was slow enough to invite --no-verify. Once per push is the right cadence:
+# still ahead of CI, and it cannot be outrun by a fast commit loop.
+npm run typecheck
+npm run test:governance
```
moxious: **Blocking — this blocks `git push` on Windows.**

`npm run test:governance` includes `src/validation/import-graph.test.ts`, and four assertions there are separator-dependent, so gating the push on this suite makes a pre-existing test bug a hard push blocker on a platform CI never runs.

- `import-graph.ts:383` returns `toPosixPath(relative)` — always `/` — while tests at `:308`/`:314`/`:320` assert `path.join('utils','helpers')`, `path.join('utils','types')`, `path.join('lib','analytics')`. The same file at `:332` asserts the posix literal for the identical input, so the two forms are mutually exclusive off posix.
- `ALIAS_RESOLUTION_SUFFIXES` at `:211` holds `'/index.ts'`, concatenated at `:215` into an otherwise native absolute path, against a `path.join` expectation at test `:223`.

Your Windows report is confirmable from the source — two reviewers reproduced all four with `path.win32`, no Windows host needed. There is no platform guard or skip, and all 24 `runs-on:` lines are `ubuntu-latest`.

Cheapest resolution is your option 2: land the mode change and the lint-staged split now, hold this hook until the four assertions are separator-clean.

.husky/pre-push
```diff
@@ -0,0 +1,5 @@
+# Whole-project gates. These ran on every commit, where their combined ~52s
+# was slow enough to invite --no-verify. Once per push is the right cadence:
+# still ahead of CI, and it cannot be outrun by a fast commit loop.
+npm run typecheck
```
moxious: **Blocking — this gate fails open.**

Two bare commands with no `set -e` and no `&&` means the script exits with the status of its *last* command. So a failing `npm run typecheck` followed by a passing `npm run test:governance` exits 0 and the push proceeds.

Under the husky 9 layout `.husky/_/h` runs `sh -e "$s"` and masks this. Under `core.hooksPath=.husky` git execs the hook directly with no `-e` — and that is precisely the layout your mode change exists to repair, so the fail-open is live exactly where the fix is load-bearing. Both reviewers reproduced it: a mode-755 hook containing `npm run boom` (exit 3) then `npm run ok` let `git push` complete with exit 0, `ls-remote` confirming the ref moved.

The ordering is the common case, not contrived: `.config/jest.config.js` transforms with `@swc/jest`, which strips types without checking them, so `tsc --noEmit` can fail while the six governance suites stay green.

Fix: `set -e` as the first line here (and in `.husky/pre-commit` for consistency), or chain with `&&`.

.husky/pre-push
```diff
@@ -0,0 +1,5 @@
+# Whole-project gates. These ran on every commit, where their combined ~52s
+# was slow enough to invite --no-verify. Once per push is the right cadence:
+# still ahead of CI, and it cannot be outrun by a fast commit loop.
```
moxious: **Blocking — the guarantee claimed here is weaker than what runs.**

Git passes a pre-push hook only the remote name and URL as argv, with the ref lines on stdin; both reviewers confirmed `ARGC=2` with a probe hook. `tsc --noEmit` and `jest` read neither, so this asserts a property of the current *working tree*, not of the history reaching the remote.

Reproduced: pushing a branch while `HEAD` is elsewhere typechecks the wrong tree; a dirty tree fails a push whose commits are clean; a clean tree passes even when an older pushed commit does not typecheck. A branch-delete push, a tag-only push and an already-up-to-date push all fire the hook and pay the full cost for zero code.

In fairness this is inherent to any pre-push hook shelling a whole-project command, not something you did uniquely wrong — but the mitigation is cheap: read the stdin ref lines and exit 0 when every local SHA is the all-zeroes delete sentinel. Then either document the working-tree semantics or drop the "cannot be outrun" claim, which the implementation does not support.

package.json
```diff
@@ -175,7 +175,11 @@
     }
   },
   "lint-staged": {
-    "*.{ts,tsx,js,mjs,json,yaml,md}": "prettier --config .prettierrc.js --write"
+    "*.{ts,tsx,js,mjs}": [
+      "eslint --cache --fix",
```
moxious, line 179: **Follow-up — this task silently does nothing for 66 tracked TS files.**

lint-staged 17.2.0 sets `matchBase: !pattern.includes('/')` (`lib/matchFiles.js:16`), so `*.{ts,tsx,js,mjs}` matches at any depth and does hand these files to eslint. But those paths resolve to no ESLint configuration, so the task exits 0 without linting them.

Measured read-only — `npx eslint --no-fix` on `.config/webpack/webpack.config.ts`, `tests/e2e-runner/utils/preflight.ts` and `playwright.config.ts` each reports the file ignored for want of a matching configuration, at exit 0. `git ls-files` for `*.ts`/`*.tsx` outside `src/` counts exactly 66, plus 28 more `js`/`mjs`. The no-op is invisible: a warning at exit 0, no `--max-warnings`, and lint-staged suppresses task output on success.

The change genuinely closes the gap for `src/`, which is the bulk of the code — just worth scoping the claim to `src/` in the PR body. Optionally `--no-warn-ignored` makes an uncovered staged file surface instead of passing silently.

For the record, three adjacent worries were checked and cleared: `.eslintcache` is gitignored (`.gitignore:47`); sharing it with `npm run lint` is safe because ESLint refuses to cache a result carrying an `output` property, so `--fix` cannot poison it; and `eslint --fix` before `prettier --write` is the correct ordering, with no prettier plugin or stylistic rules in either config to fight over.

moxious (changes requested):
Thanks for fixing this — and for disclosing the Windows failure and the `--no-verify` push rather than quietly working around them. That disclosure is what let the review confirm the platform problem from the source instead of guessing at it, and it is the right instinct on a change that touches every contributor's local workflow.

## Merge contract

1. [blocking] **pre-push-blocks-non-linux-push — pre-push gates git push on a suite with four separator-dependent assertions, so merging blocks push on Windows** (high · testing-and-verification)
   `.husky/pre-push` runs `npm run test:governance`, which includes `src/validation/import-graph.test.ts`, and four assertions there are separator-dependent. `import-graph.ts:383` returns `toPosixPath(relative)`, always forward slashes, while the tests at `:308`/`:314`/`:320` assert `path.join('utils','helpers')`, `path.join('utils','types')` and `path.join('lib','analytics')`; the same file at `:332` asserts the posix literal for the identical input, so the two forms are mutually exclusive off posix. The fourth is `ALIAS_RESOLUTION_SUFFIXES` at `:211`, whose `'/index.ts'` entry is concatenated at `:215` into an otherwise native absolute path, against a `path.join` expectation at test `:223`. So your Windows report is not just plausible, it is confirmable from the source: two reviewers reproduced all four with `path.win32` without needing a Windows host. There is no platform guard or skip in either file, all 24 `runs-on:` lines in `.github/workflows/` are `ubuntu-latest`, and `test:governance` is not a CI job at all, so CI cannot see this. Windows is a documented dev platform (`npm run build:backend:windows`, `LOCAL_DEV.md:48`). `git push` works for those contributors today and stops working when this lands, and it needs no reinstall to activate: husky already wrote `.husky/_/pre-push`, and the `[ ! -f "$s" ]` guard is the only thing keeping pre-push inert. You disclosed this and pushed the branch with `--no-verify`, which is the gate failing its first real user.
   Required: Pick one of your three options explicitly. The cheapest is option 2: land the mode change and the lint-staged split now, hold `.husky/pre-push` until the assertions are separator-clean. If you would rather land the gate, fix the four first: assert the posix literal in the three `resolveImportToRelative` tests, and use `path.join` instead of `+ suffix` for the index branch of `findExistingSourceFile`.

2. [blocking] **pre-push-fails-open-no-set-e — New pre-push has no set -e, so a failing typecheck followed by a passing governance run exits 0 and the push proceeds** (high · correctness-and-reliability)
   `.husky/pre-push` is two bare commands with no `set -e` and no `&&`, so the script exits with the status of its last command. Under the husky 9 layout `.husky/_/h` supplies the `-e` and masks this. Under `core.hooksPath=.husky` git execs the tracked hook directly with no `-e`, and that is precisely the layout your mode change exists to repair, so the fail-open is live exactly where the fix is load-bearing. Both reviewers reproduced it independently in throwaway repos: a mode-755 shebang-less hook containing `npm run boom` (exit 3) then `npm run ok` let `git push` complete with exit 0, and `ls-remote` confirmed the ref moved. The ordering is the common case rather than a contrived one, because `.config/jest.config.js` transforms with `@swc/jest`, which strips types without checking them, so `tsc --noEmit` can fail while the six governance suites stay green. A gate that fails open is worse than no gate, because the contributor reasonably believes typecheck ran.
   Required: Add `set -e` as the first line of `.husky/pre-push`, and for consistency `.husky/pre-commit`, or chain the commands with `&&`. Both hooks then fail closed whether git invokes them directly or through husky's `sh -e` wrapper.

3. [blocking] **pre-push-validates-working-tree — pre-push ignores argv and stdin, so it checks the working tree rather than the pushed commits and fires on pushes carrying no code** (medium · correctness-and-reliability)
   Git passes a pre-push hook only the remote name and URL as argv, with the ref lines on stdin; both reviewers confirmed `ARGC=2` with a probe hook. `tsc --noEmit` and `jest` read neither, so the gate asserts a property of the current working tree, not of the history reaching the remote. Reproduced consequences: pushing a branch while `HEAD` is elsewhere typechecks the wrong tree; a dirty tree fails a push whose commits are clean; a clean tree passes even when an older pushed commit does not typecheck. A branch-delete push, a tag-only push and an already-up-to-date push all invoke the hook and pay the full cost for zero code. In fairness this is inherent to any pre-push hook that shells a whole-project command rather than something you did uniquely wrong, and the empty-stdin case shows the mitigation is cheap and available. But the new file's comment asserts the gate `cannot be outrun by a fast commit loop`, and what the implementation actually guarantees is weaker than that. False blocks from unrelated dirty state, plus a full suite on no-op pushes, are the usual reason people reach for `--no-verify`.
   Required: Read the stdin ref lines and exit 0 when every line's local SHA is the all-zeroes delete sentinel, which removes the no-op-push cost for one short block. Then either accept and document the working-tree semantics, or drop the cannot-be-outrun claim, which the implementation does not support.

## Follow-ups

1. [follow_up] **hook-mode-premise-scoped — The 644 to 755 change is inert in both supported hook environments; it repairs only a stale husky-8 core.hooksPath** (medium · build-and-ci)
   `node_modules/husky/index.js` unconditionally runs `git config core.hooksPath "<dir>/_"` and writes all 14 wrappers into the gitignored `.husky/_/` at mode 0o755. `.husky/_/h` gates on `[ ! -f "$s" ]`, a file-existence test, then passes the tracked hook to `sh -e` as an argument, so its mode is never consulted. Verified across all three environments with reviewer-authored hook bodies: with `core.hooksPath=.husky/_` a mode-644 shebang-less hook ran at both 644 and 755; with `core.hooksPath=.husky` mode 644 was skipped with git's not-executable hint, which is the exact transcript in #1817, and 755 ran; with `core.hooksPath` unset nothing under `.husky/` ran at any mode. `.husky/_/pre-commit` is present at 755 in this tree, so under the supported layout the old three-command hook was already running all three steps. The mode bit is therefore load-bearing in exactly one environment, a legacy `core.hooksPath` that `npm run prepare` would itself overwrite. For a fresh clone the real cause is `.npmrc`'s `ignore-scripts=true` plus the manual `npm run prepare` step, which this PR does not touch. The chmod is still correct hygiene and the hook bodies are benign, invoking only pre-existing package.json scripts with no new command, network call or env read.
   Follow-up: Keep the chmod, but restate the mechanism in the PR body and at issue close: it repairs the legacy `core.hooksPath=.husky` layout, and under a canonically prepared checkout the hook was already running. Then decide explicitly whether the fresh-clone gap is in scope here or a follow-up, so #1817 does not close on a premise that holds in one of three environments.

2. [follow_up] **eslint-noop-on-unlinted-paths — The added eslint task silently no-ops on the 66 tracked TS files outside src/, so the commit-time coverage claim is broader than the change delivers** (medium · build-and-ci)
   lint-staged 17.2.0 sets `matchBase: !pattern.includes('/')` at `lib/matchFiles.js:16`, so the new `*.{ts,tsx,js,mjs}` task matches at any depth and does hand these files to eslint. But those paths resolve to no ESLint configuration in this repo, so the added task exits 0 without linting them. Measured read-only: `npx eslint --no-fix` on `.config/webpack/webpack.config.ts`, `tests/e2e-runner/utils/preflight.ts` and `playwright.config.ts` each reports that the file is ignored because no matching configuration was supplied, at exit 0. `git ls-files` for `*.ts` and `*.tsx` outside `src/` counts exactly 66 files, and a further 28 tracked `js`/`mjs` files outside `src/` are in the same position. The no-op is invisible: ESLint reports it as a warning at exit 0, no `--max-warnings` is set anywhere, and lint-staged suppresses task output on success, so the hook prints nothing and passes. Three adjacent worries were checked and cleared, not findings: `.eslintcache` is gitignored at `.gitignore:47`; sharing that cache with `npm run lint` is safe because ESLint refuses to cache a result carrying an `output` property, so `--fix` cannot poison it; and `eslint --fix` before `prettier --write` is the correct ordering, with no prettier plugin or stylistic rules in either config to fight over.
   Follow-up: Scope the claim to src/ in the PR body so the ratchet-coverage argument matches what the task actually does. Optionally pass --no-warn-ignored so an uncovered staged file surfaces instead of passing silently.

3. [follow_up] **hook-docs-stale — Both contributor-facing hook descriptions become wrong, and no doc mentions the new pre-push gate** (medium · cross-cutting-architecture)
   `docs/developer/LOCAL_DEV.md:146` and `docs/developer/GETTING_STARTED.md:99` both say the hook applies Prettier to staged files. After this PR lint-staged also runs `eslint --cache --fix` on `*.{ts,tsx,js,mjs}` and the single glob is split in two, so both sentences are wrong in a new way; at base they were already incomplete, since neither mentioned the typecheck and governance steps. `grep -rn pre-push` over `docs/`, `.cursor/`, `AGENTS.md` and `.github/` returns zero hits, so a contributor's `git push` starts running a ~52s gate that no document describes. Both troubleshooting sections are headed `Husky pre-commit hook fails` (`LOCAL_DEV.md:191`, `GETTING_STARTED.md:138`), so a blocked pusher finds no matching entry, though the advice itself still holds because `scripts/check.js` covers typecheck and the governance suites transitively. `prevent-doc-drift/SKILL.md:21` puts `docs/developer/**` in the allowed-edit list and forbids `docs/design/**`, so these two updates belong in this PR while `AGENT_HARDENING.md` correctly does not. The first unexplained 52s stall on `git push` reads as a hang, not a gate, which is how `--no-verify` becomes habit.
   Follow-up: Rewrite `LOCAL_DEV.md:146` and `GETTING_STARTED.md:99` to state that lint-staged runs `eslint --fix` then prettier on `*.{ts,tsx,js,mjs}` and prettier alone on `*.{json,yaml,md}`; add one sentence in each naming the pre-push hook, what it runs and its rough cost; and broaden the two troubleshooting headings to cover both hooks. `GETTING_STARTED.md:40`, which claims `npm install` triggers hook setup and is contradicted by `.npmrc` and `LOCAL_DEV.md:24`, is a one-line fix worth riding along.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1823
Summary: make the pre-commit hook run, and give it the right work
Verdict: Request Changes
Results: 3 blockers, 0 non-blocking findings, 3 follow-ups


Next step: of the three options you laid out, option 2 is the one I would take — land the mode change and the lint-staged split now, and hold `.husky/pre-push` until the separator assertions are clean and the hook fails closed. That gets the real fix for #1817 in without gating anyone's `git push` on a suite that cannot pass off posix. If you would rather land the gate in one go, the three blocking items above are what it needs first. Push when ready and we will re-review on the new head — CI is already authorized, and the one red check (`publish-report`) is a known fork-permission limitation on community PRs, not anything you did.

dchaudhari7177:
Took option 2. `.husky/pre-push` is gone from this PR; what is left is the mode fix and the lint-staged split, which is the actual fix for #1817. All three blockers were properties of the pre-push hook, so they leave with it — the separator work in `import-graph.test.ts` is a separate problem and does not belong in front of a hook fix.

All three follow-ups are in.

**`hook-mode-premise-scoped`** — the commit message now says the ratchets apply to `src/`, and the PR body below is rewritten the same way.

**`hook-docs-stale`** — `LOCAL_DEV.md:146` and `GETTING_STARTED.md:99` both now say `eslint --fix` then prettier on `*.{ts,tsx,js,mjs}`, prettier alone on `*.{json,yaml,md}`. Both troubleshooting headings are now "Husky pre-commit hook fails or blocks the commit", since the failure mode people will actually hit is a lint error stopping a commit that used to go through. No pre-push sentence, since there is no pre-push hook. `GETTING_STARTED.md:40` rode along: it claimed `npm install` triggers hook setup, which `.npmrc`'s `ignore-scripts=true` has made untrue, and it now shows `npm run prepare` the way `LOCAL_DEV.md:24` does.

**`eslint-noop-on-unlinted-paths` → `--no-warn-ignored`** — I did not add this one, because it does the opposite of what the follow-up wants. The flag *suppresses* the ignored-file notice; leaving it off is what surfaces one. Checked on this repo's eslint 9.39.1:

```
$ eslint --cache package.json
  0:0  warning  File ignored because no matching configuration was supplied
  ✖ 1 problem (0 errors, 1 warning)          exit 0

$ eslint --cache --no-warn-ignored package.json
                                             exit 0     # silent
```

So the default already prints the warning through lint-staged, and adding the flag would be the change that makes an uncovered staged file pass silently. Happy to add it if the intent was actually to quiet the notice — say the word.

**One extra, same bug class.** `.gitattributes` had `eol=lf` for `*.sh`, `*.bash`, `*.zsh` and the source extensions, but git hooks have no extension, so nothing covered `.husky/*`. A hook checked out with CRLF fails to execute for the same silent reason the mode did. Added `.husky/* text eol=lf`.

### Verified

Ran lint-staged against real staged files rather than reasoning about the glob:

- A `.ts` file with `export const probe   =  {a:1,   b:2};` → eslint then prettier ran, file came back `export const probe = { a: 1, b: 2 };`.
- The same file importing `moment` → `✖ eslint --cache --fix` with `'moment' import is restricted from being used  no-restricted-imports`, tasks failed, staged state reverted. That rule is one of the ratchets that previously only fired in CI.
- The commit for this round was itself gated by the hook — it ran `prettier` over the two changed `.md` files.

moxious (approved):
Approving. All three round-1 blockers are resolved — `.husky/pre-push` is gone from the head tree and `pre-commit` holds a single command — and five of the six earlier items are closed out.

This is a good change, and worth saying why, because the diff is small enough to undersell it:

- `.gitattributes` genuinely had **no** rule covering hook files. Every existing rule is extension-based (`*.sh`, `*.ts`, …) and git hooks have no extension, so a CRLF checkout could leave the hook unexecutable. `.husky/* text eol=lf` closes a real gap.
- The `GETTING_STARTED.md` correction is the most valuable part. `.npmrc` sets `ignore-scripts=true` as deliberate supply-chain hardening, so `npm install` does not install the hooks — while the docs said it did. Every fresh clone following the documented path had no hooks at all. Adding the `npm run prepare` step fixes that.
- `lint-staged` was Prettier-only; it now runs `eslint --cache --fix` on staged `.ts/.tsx/.js/.mjs` first. That is new enforcement, not a reshuffle.
- Dropping `typecheck` and `test:governance` from the hook is the right trade. Both still run in CI inside `ci-gate`, which is required on `main`, so nothing stops being enforced — and a 52-second commit hook is one people learn to `--no-verify` past, at which point it enforces nothing at all.

I also declined to re-raise the `set -e` item from round 1: the restructure left the hook at one command, so its exit status *is* the hook's exit status. That ask became moot rather than unaddressed.

One note tracked separately as #1860, not a change request against this PR: the hook comment's closing clause says the old hook was one "git was silently skipping". That is true only under a legacy `core.hooksPath=.husky` checkout. Under husky 9's own default the shim gates on file existence rather than the executable bit and invokes the hook through `sh -e`, so the three-command hook was running and this PR narrows a live gate rather than reviving a dead one. The incentive argument stands without that clause, and the premise came from how the work was framed upstream, so correcting it is ours to do.

On CI: the `CI` run shows a failure, and it is **not** yours. 38 of 39 jobs passed; `publish-report` could not post its results comment because a fork PR lacks that permission (`Resource not accessible by integration`). That is a known limitation of running this workflow from a fork.

PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1823
Summary: make the pre-commit hook run, and give it the right work
Verdict: Approve

PR #1824, opened 2026-09-08
docs(cross-tab): document requirement shape mismatch

### LINKED ISSUE #1808: docs(cross-tab): note requirements/objectives in the same-build assumption
### Context

Filed as a follow-up from review of https://github.com/grafana/grafana-pathfinder-app/pull/1791 ("fix(objectives): reject invalid completion checks and preserve condition arrays"). Not a blocker on that PR — see reasoning below — but worth a durable record.

### What's true

That PR widens the `requirements` field of the two-tab controller's `check-requirements`/`fix-requirement` wire messages from `string` to `string | readonly string[]` (`ConditionInput`). `src/requirements-manager/step-checker.hook.ts` now forwards a real array to `controllerChannel.requestRequirementCheck` whenever a guide's JSON declares `requirements`/`objectives` as an array — which, after that PR, is the mainline shape for essentially all guides (`src/docs-retrieval/json-parser.ts` no longer joins them into a comma string).

An **old (pre-PR) live tab's** validator (`isValidCheckRequirements` / `isValidFixRequirement` in `src/types/cross-tab.types.ts`) still requires `typeof message.requirements === 'string'`, so it rejects the new array shape outright. `CrossTabTransport.handleMessage` drops any message that fails `validateCrossTabMessage` before it reaches a listener — silently, with no reply. The requesting (new) controller tab's `request()` then times out (`REQUEST_TIMEOUT_MS`) and falls back to `checkHere(stripTabLocalRequirements(remaining))`, a local-only check that does **not** re-verify the tab-local DOM/URL requirements the live tab would have checked.

Net effect: a controller tab on a newer plugin build, paired with a live tab still running an older build (realistic during a rolling deploy, or a tab left open across a version bump), silently degrades requirement checking for `requirements`/`objectives` instead of genuinely verifying them on the live tab. It looks identical to an ordinary disconnected-tab timeout, so it's easy to miss in the field.

Verified end-to-end (base-vs-head code trace) independently by two reviewers during the PR #1791 review.

### Why this isn't blocking

`src/types/cross-tab.types.ts` and `docs/developer/CROSS_TAB_CONTROLLER.md` already declare, **unchanged since before PR #1791**, a "same-build assumption": *"the controller and live tabs are the same plugin build in the same browser/origin/session ... there is no protocol-version negotiation. Cross-version compatibility is not a goal; a mismatched build is out of scope."* This finding is a new instance of that already-accepted risk category (any wire-shape change can silently degrade across a mismatched build), not a new violation of a live guarantee. The two-tab controller also remains admin-gated and off by default (`DEFAULT_ENABLE_TWO_TAB_CONTROLLER = false`).

### Suggested action

Small, non-urgent:

- Note in `docs/developer/CROSS_TAB_CONTROLLER.md`'s existing "Same-build assumption" section that `requirements`/`objectives` is now one of the fields affected by it (previously this field couldn't diverge in a way that changed validation outcome, since both sides always spoke plain strings).
- If/when `enableTwoTabController` moves toward default-on, revisit whether the same-build non-goal is still acceptable for this specific field, given array-shaped requirements are now the mainline representation for guides.

No code fix is being deferred here — this is a documentation gap plus a standing architectural note for whenever the same-build assumption gets revisited.

dvd233:
## Summary

The two-tab controller's same-build assumption now names array-shaped guide `requirements` as a concrete cross-version failure mode. Older live tabs that accept only strings can drop a `check-requirements` command; it then times out and uses the documented local-only fallback, while a dropped `fix-requirement` returns `No live tab responded` and is not retried locally.

This records the existing cross-version non-goal; it does not change wire types, validation, or runtime behavior. Guide `objectives` are evaluated in the controller tab and never cross this wire.

## What to look at

- [Developer contract](https://github.com/dvd233/grafana-pathfinder-app/blob/d19ca7943672cc3a224dae37efe1167251194d08/docs/developer/CROSS_TAB_CONTROLLER.md) distinguishes the `check-requirements` local fallback from the `fix-requirement` blocked outcome and records that guide objectives stay local.
- [Protocol-side contract](https://github.com/dvd233/grafana-pathfinder-app/blob/d19ca7943672cc3a224dae37efe1167251194d08/src/types/cross-tab.types.ts) keeps both messages on the existing `ConditionInput` shape and explains the same-build boundary.

## How to verify

1. Read both linked paragraphs and confirm they describe array-shaped guide requirements, controller-local objectives, and the distinct timeout behavior for checks versus fixes.
2. Trace `CheckRequirementsMessage` and `FixRequirementMessage`; confirm their `requirements` field remains `ConditionInput`, neither message carries `objectives`, and no validator or signed-message kind changes in this PR.

## Breaking changes

None. This change is documentation-only and fully reversible.

Closes #1808

### DIFF
--- a/docs/developer/CROSS_TAB_CONTROLLER.md
+++ b/docs/developer/CROSS_TAB_CONTROLLER.md
@@ -155,7 +155,11 @@ Defense in depth on top of authentication:
 - **Same-build / same-origin / one-session assumption.** Controller and live
   tabs are the same plugin build in the same browser profile and session; there
   is no protocol-version negotiation and cross-version compatibility is not a
-  goal.
+  goal. Requirement traffic is a concrete example: `check-requirements` can now
+  carry array-shaped guide `requirements` and `objectives`, and
+  `fix-requirement` also accepts a `ConditionInput` in its `requirements` field.
+  An older validator that accepts only strings drops the command; the controller
+  times out and falls back to its local-only check.
 
 ### Known limitations / future work
 
--- a/docs/developer/CROSS_TAB_CONTROLLER.md
+++ b/docs/developer/CROSS_TAB_CONTROLLER.md
@@ -155,11 +155,12 @@ Defense in depth on top of authentication:
 - **Same-build / same-origin / one-session assumption.** Controller and live
   tabs are the same plugin build in the same browser profile and session; there
   is no protocol-version negotiation and cross-version compatibility is not a
-  goal. Requirement traffic is a concrete example: `check-requirements` can now
-  carry array-shaped guide `requirements` and `objectives`, and
-  `fix-requirement` also accepts a `ConditionInput` in its `requirements` field.
-  An older validator that accepts only strings drops the command; the controller
-  times out and falls back to its local-only check.
+  goal. Requirement traffic is a concrete example: `check-requirements` can
+  carry array-shaped guide `requirements`, and `fix-requirement` carries the
+  same `ConditionInput` shape. Guide `objectives` are evaluated in the
+  controller tab and never cross this wire. An older validator that accepts
+  only strings can drop a command; the timeout behavior differs by message kind
+  as described below.
 
 ### Known limitations / future work
 
@@ -264,10 +265,13 @@ Controller (useStepChecker, controller mode)        Live tab (installLiveTabExec
   `requestRequirementCheck` posts regardless of connection state (the channel
   value omits `connected`), so the check fires the moment a step renders;
   `connected` only drives the presence badge.
-- If no live tab answers within the timeout, the round-tripped half falls back to
-  stripping the tab-local tokens and evaluating the rest locally —
-  session/permission requirements still gate, and a disconnected controller never
-  hangs. The guide-scoped half is unaffected: it never depended on the live tab.
+- If no live tab answers within the 4-second request timeout, a
+  `check-requirements` request falls back to stripping the tab-local tokens and
+  evaluating the rest locally — session/permission requirements still gate, and
+  a disconnected controller never hangs. A `fix-requirement` request instead
+  returns `No live tab responded`, is not retried locally, and leaves the step
+  blocked while logging a warning. The guide-scoped half is unaffected: it
+  never depended on the live tab.
 
 ## Presence
 
--- a/src/types/cross-tab.types.ts
+++ b/src/types/cross-tab.types.ts
@@ -210,7 +210,11 @@ export const SIGNED_MESSAGE_KINDS: ReadonlySet<CrossTabMessage['kind']> = new Se
 // Same-build assumption: the controller and live tabs are the same plugin
 // build in the same browser/origin/session, so there is no protocol-version
 // negotiation. Cross-version compatibility is not a goal; a mismatched build
-// is out of scope. See docs/developer/CROSS_TAB_CONTROLLER.md.
+// is out of scope. Requirement traffic is one concrete case: check-requirements
+// can now carry array-shaped guide requirements/objectives, and fix-requirement
+// also accepts ConditionInput. Older validators accept only strings and drop
+// these commands; the controller then times out and falls back locally. See
+// docs/developer/CROSS_TAB_CONTROLLER.md.
 
 // Recognized interactive action verbs. Kept as a literal set (not derived
 // from InteractiveAction) so the receive gate stays decoupled from the
--- a/src/types/cross-tab.types.ts
+++ b/src/types/cross-tab.types.ts
@@ -211,9 +211,11 @@ export const SIGNED_MESSAGE_KINDS: ReadonlySet<CrossTabMessage['kind']> = new Se
 // build in the same browser/origin/session, so there is no protocol-version
 // negotiation. Cross-version compatibility is not a goal; a mismatched build
 // is out of scope. Requirement traffic is one concrete case: check-requirements
-// can now carry array-shaped guide requirements/objectives, and fix-requirement
-// also accepts ConditionInput. Older validators accept only strings and drop
-// these commands; the controller then times out and falls back locally. See
+// can carry array-shaped guide requirements, and fix-requirement carries the
+// same ConditionInput shape. Guide objectives stay local to the controller and
+// never cross this wire. A dropped check falls back locally after 4s with
+// tab-local tokens stripped; a dropped fix returns "No live tab responded", is
+// not retried locally, and leaves the step blocked. See
 // docs/developer/CROSS_TAB_CONTROLLER.md.
 
 // Recognized interactive action verbs. Kept as a literal set (not derived


docs/developer/CROSS_TAB_CONTROLLER.md
```diff
@@ -155,7 +155,11 @@ Defense in depth on top of authentication:
 - **Same-build / same-origin / one-session assumption.** Controller and live
   tabs are the same plugin build in the same browser profile and session; there
   is no protocol-version negotiation and cross-version compatibility is not a
-  goal.
+  goal. Requirement traffic is a concrete example: `check-requirements` can now
+  carry array-shaped guide `requirements` and `objectives`, and
```
moxious: `objectives` never crosses this boundary. `CheckRequirementsMessage` (src/types/cross-tab.types.ts:124-132) has no objectives field, and the only producer (src/global-state/controller-channel.tsx:332) is reached only from src/requirements-manager/step-checker.hook.ts:313 behind `isRemote` (:266). The two `remote: true` call sites (:545, :630) both pass the step's `requirements`; the objectives branch (:494-513) passes `requirements: objectives ?? []` and omits `remote`, so it evaluates locally via `checkHere`. `interactive-section.tsx:380-388` supplies objectives with no requirements, so `decideStepCheck` returns `enabled` (src/requirements-manager/step-decision.ts:24) and never reaches the remote branch.

Drop `objectives` here. Stating positively that guide `objectives` are always evaluated in the controller tab and never round-trip would be more useful than just removing the word.

src/types/cross-tab.types.ts
```diff
@@ -210,7 +210,11 @@ export const SIGNED_MESSAGE_KINDS: ReadonlySet<CrossTabMessage['kind']> = new Se
 // Same-build assumption: the controller and live tabs are the same plugin
 // build in the same browser/origin/session, so there is no protocol-version
 // negotiation. Cross-version compatibility is not a goal; a mismatched build
-// is out of scope. See docs/developer/CROSS_TAB_CONTROLLER.md.
+// is out of scope. Requirement traffic is one concrete case: check-requirements
+// can now carry array-shaped guide requirements/objectives, and fix-requirement
```
moxious: Same issue as the doc: no objectives value can reach a posted message. Both wire types carry only `requirements: ConditionInput` (:128, :145), and `step-command` carries `Omit<InternalAction, 'requirements'>` (:13-24), which has no objectives field either. Please drop `objectives` from this sentence.

src/types/cross-tab.types.ts
```diff
@@ -210,7 +210,11 @@ export const SIGNED_MESSAGE_KINDS: ReadonlySet<CrossTabMessage['kind']> = new Se
...
+// also accepts ConditionInput. Older validators accept only strings and drop
+// these commands; the controller then times out and falls back locally. See
```
moxious: `these commands ... falls back locally` is true for `check-requirements` only. The check fallback is `null` (src/global-state/controller-channel.tsx:339), which src/requirements-manager/step-checker.hook.ts:317 turns into a real local re-check. The fix fallback is `{ ok: false, error: 'No live tab responded' }` (:355), and remote-vs-local fix is a mutually exclusive ternary (step-checker.hook.ts:688-704), so `dispatchFix` is never reached after a timeout — there is no local fallback for a fix. It is also quieter than this implies: `SET_ERROR` leaves `explanation` undefined (src/requirements-manager/step-state.ts:151-161) and interactive-step.tsx:1233 gates the `Fix this` block on `explanationText`, so the step goes blocked and the affordance unmounts while the message reaches only `logger.warn`.

Suggest splitting the consequence per kind.

moxious (changes requested):
Thanks for documenting the cross-tab requirement shape mismatch — the same-build assumption is exactly the kind of thing that costs a debugger hours when it is not written down, and pairing the doc with the source comment is the right instinct.

## Merge contract

1. [blocking] **crosstab-objectives-never-on-wire — The added same-build example names guide `objectives` as a `check-requirements` payload, but objectives never cross the cross-tab boundary** (medium · cross-tab-controller)
   Both added blocks state that `check-requirements` can now carry array-shaped guide `requirements` and `objectives`. The `requirements` half is correct; the `objectives` half is false, and three independent traces agree. Neither wire type has an objectives field — `CheckRequirementsMessage` (src/types/cross-tab.types.ts:124-132) and `FixRequirementMessage` (:141-149) carry only `requirements: ConditionInput`. The only producers are controller-channel.tsx:332 and :348, whose only non-test callers are step-checker.hook.ts:313 and :690, both gated on `isRemote` (step-checker.hook.ts:266). Exactly two call sites pass `remote: true` (:545, :630) and both pass the step's `requirements`. The objectives branch (:494-513) passes `requirements: objectives ?? []` but omits `remote`, so it evaluates locally via `checkHere`. No consumer launders objectives into the `requirements` prop, `interactive-section.tsx:380-388` supplies objectives with no requirements so `decideStepCheck` returns `enabled`, and `step-command` carries `Omit<InternalAction, 'requirements'>`, which has no objectives field. These two blocks are the canonical statement of what crosses this trust boundary, so the wrong half will send a future debugger hunting a cross-version drop that cannot occur.
   Required: Drop `objectives` from both sentences and keep `requirements` as the affected field. If the array-shape origin is worth naming, cite the JSON-guide producer (`requirements?: string[]`, forwarded unchanged by src/docs-retrieval/json-parser.ts). Better still, state positively that guide `objectives` are always evaluated in the controller tab and never round-trip — that is the fact a debugger actually needs.

2. [blocking] **crosstab-fix-requirement-has-no-local-fallback — The source comment attributes the timeout-then-local-fallback consequence to `fix-requirement`, which has no local fallback** (low · cross-tab-controller)
   The comment reads `drop these commands; the controller then times out and falls back locally` — plural, directly after a sentence naming both `check-requirements` and `fix-requirement`. It is accurate for the check only. `request()` resolves a per-caller fallback after REQUEST_TIMEOUT_MS = 4000 (controller-channel.tsx:24, :312-325). The check fallback is `null` (:339), which step-checker.hook.ts:317 turns into a genuine local re-check with tab-local tokens stripped. The fix fallback is `{ ok: false, error: 'No live tab responded' }` (:355), and the remote/local fix choice is a mutually exclusive ternary (step-checker.hook.ts:688-704), so `dispatchFix` is never reached after a remote timeout. The failure is also quieter than the prose implies: it reaches `logger.warn` plus a `SET_ERROR` that leaves `explanation` undefined (step-state.ts:151-161), and interactive-step.tsx:1233 gates the block hosting the `Fix this` button on `explanationText` — so the step goes blocked, the affordance unmounts, and nothing renders `checker.error`. The doc's generic singular `the command` can fairly be read as scoping to the check alone; only the source comment is unambiguously wrong.
   Required: Split the consequence per kind in the source comment, and consider mirroring it in the doc: a dropped `check-requirements` times out after 4s and falls back to a local check with tab-local tokens stripped; a dropped `fix-requirement` times out to `No live tab responded`, is never retried locally, and surfaces only as a console warning with the step blocked.

## Suggestions

1. [suggestion] **crosstab-same-build-comment-duplicates-doc — The source comment now duplicates the doc it already points to, and both copies inherited the same error** (low · cross-tab-controller)
   The note at src/types/cross-tab.types.ts:210-217 is a floating comment attached to no declaration; the base was four lines ending in `See docs/developer/CROSS_TAB_CONTROLLER.md.` The added clause `fix-requirement also accepts ConditionInput` restates the declarations 80 lines above it (:128, :145), and the rest now exists as prose in the very document the last line links to. `can now carry` is a temporal anchor that goes stale on the next wire change. The `objectives` error landing identically in both copies is the drift cost made concrete on the first edit. This subsystem does keep multi-line correctness-warning comments that end in a doc pointer (controller-channel.tsx:247-252 is a six-line example), so length alone is not the objection — duplicating the linked prose is.
   Suggested: Keep the source comment at its base four-line form, whose doc pointer already routes the reader, and carry the corrected concrete example only in CROSS_TAB_CONTROLLER.md. If a source-side hint is wanted, one short line naming `requirements` as the field the assumption now bites is enough.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1824
Summary: document requirement shape mismatch
Verdict: Request Changes
Results: 2 blockers, 1 non-blocking finding, 0 follow-ups


Next step: drop `objectives` from both sentences, and split the timeout consequence per message kind so the check's local fallback is not attributed to the fix. Push when ready and we will re-review on the new head — CI is already authorized, and the one red check (`publish-report`) is a known fork-permission limitation on community PRs, not anything you did.

dvd233:
@moxious Thanks for the precise review. `d19ca794` addresses both blockers:

- `objectives` are now described as controller-local; only array-shaped guide `requirements` and the existing `ConditionInput` requirement field are named on the wire.
- The docs and source comment distinguish the 4-second `check-requirements` local fallback from `fix-requirement` returning `No live tab responded`, without a local retry and with the step remaining blocked.

I also corrected the general requirement-evaluation paragraph so it no longer attributes the check fallback to fixes. Verification completed: the five related cross-tab suites (141 tests), typecheck, full lint, Prettier, and terms-sync check. Ready for another look.

moxious (approved):
Thanks for turning both blockers around so precisely — you corrected the `objectives` half of the wire statement, split the timeout consequence per message kind, and dropped the temporal "can now" anchor, all without touching a line of runtime code. I re-traced every factual claim in the new prose against the code at `d19ca794` and all of it holds.

Approving. Nothing further is needed from you on this one; a maintainer will handle the merge.

PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1824
Summary: document requirement shape mismatch
Verdict: Approve

PR #1716, opened 2026-08-27
refactor(docs-panel): extract pure tab-state and docs-load seams

moxious:
## Intent

Address the sole review comment on PR 1716 by preventing the getPackageRenderType factory mock in docs-panel.alignment.test.ts from leaking its learning-journey return value into later tests, while preserving the reviewed test's behavior. Commit and push the focused fix to the existing PR branch from its isolated worktree.

## What Changed

- Moved four decisions out of `docs-panel.tsx` into two new pure modules: close-tab focus adjacency, gated-tab pruning, and persisted-tab projection now live in `utils/tab-state-transitions.ts` (`closeTabState`, `pruneGatedTabState`, `projectPersistedTabs`), and the implied-0th-step alignment decision plus the load-success tab patch live in `utils/docs-load-finalizer.ts` (`resolveDocsLoadAlignment`, `buildDocsLoadSuccessPatch`). The panel delegates to them and keeps the impure bits — `Date.now()` stamping, `locationService`, `setState`, storage writes — at the call site; `withoutUnauthorizedGatedTabs` is gone and the internal `didPrune` flag is now `changed`.
- Narrowed two consumers off the concrete panel class: `pendingGuideRouter.ts` takes a structural `PendingGuidePanel` interface it declares itself, and `FloatingPanelContent` types its `model` prop as `DocsPanelModelOperations` instead of importing `CombinedLearningJourneyPanel`.
- Added unit tests for both new modules and extended the panel's characterization suites (alignment commit ordering, close-tab, tab-restore guard, `FloatingPanelContent` forwarding), including a `getPackageRenderType` mock reset so the `learning-journey` return value does not leak into later tests. Docs updated to match: `AGENT_HARDENING.md` F-11/E3 size facts, the corrected path-member starting-location precedence in `EXTERNAL_API.md`, the `docs-panel.tsx` size marker in `systemPatterns.mdc`, and README type/API inventories replaced with pointers to their owning files.

## Risk Assessment

✅ Low: The fix-round changes are a single correct test-mock default plus documentation corrections I verified line-by-line against `starting-location.ts`, `CustomGuidesSection.tsx`, and the Go manifest struct, and the underlying refactor is a branch-for-branch behavior-preserving extraction with dedicated unit tests for every extracted seam.

## Testing

[...]

PR #894, opened 2026-05-17
refactor(docs-panel): extract renderer into components + hooks (sweet-spot slice)

moxious:
## Summary

Refactor `src/components/docs-panel/docs-panel.tsx` per the [High-Risk Refactor Guidelines](https://github.com/grafana/grafana-pathfinder-app/wiki/High-Risk-Refactor-Guidelines). Renderer drops from 2,681 → **1,555 LoC** (-42%); the SceneObject class (`CombinedLearningJourneyPanel`) is **untouched**. Renderer logic moved into 17 new files (5 components, 10 hooks, 2 pure utilities) plus 18 new test files pinning behavior. 18 commits — one extraction per commit.

## Guideline scope (what was applied)

- **Pattern A** — pure utility extraction (Step 1.1)
- **Pattern B** — simple hook extraction (Steps 1.2–1.7)
- **Pattern E** — interface-first component extraction (Steps 1.8–1.12)
- **Pattern C** — DI hooks for event-listener effects (Steps 2.1–2.5)
- **Pattern J** — pin-only: contract surfaces preserved; two missing tripwires (`auto-open-event`, `panel-mode`) added in Phase 0

## Where we stopped, and why (risk tolerance)

Deferred to a future refactor:

- **Pattern G — async state-machine decomposition** of `loadTabContent` / `loadDocsTabContent`. Multi-state transitions + alignment-evaluation side path + the `_pendingLaunchSource` consume-once carrier make this the single riskiest change in the file. Out of scope for this slice.
- **Pattern I — module-level registry / lifecycle guard** rework of `_hasRestoredTabs` and `_pendingLaunchSource`. StrictMode and fullscreen re-mount semantics are load-bearing; rotating them needs its own design.
- **Pattern H — singleton-to-injectable** migration of `panelModeManager`, `linkInterceptionState`, `tabStorage`.
- **Pattern F (Step 2.6) — `useLiveSessionLifecycle`** explicitly skipped per plan. Three live-session effects share `actionCaptureRef` / `actionReplayRef` across boundaries with load-bearing cleanup ordering — the plan called out "skip if cleanup-order tests prove fiddly," so deferred rather than rushed.
- **Renames of contract surfaces** (`testIds`, CustomEvent names, window globals, `tabStorage` keys, permanent tab IDs) — guideline forbids during structural refactor.

## Code change characterization

| Bucket | Change |
|---|---|
| `docs-panel.tsx` net change | **-1,126 LoC** (2,681 → 1,555) |
[...]

Issue #1763, opened 2026-09-01
Pay down remaining lateral Tier 2 boundary violations


loglapa:
PR #1756 adds accountable metadata to the architecture allowlists and separates five existing Tier 2 lateral edges that are not part of the circular-dependency clusters tracked by #1359.

Track and remove these exceptions as their boundaries are extracted:

- [x] requirements-manager/checks/grafana-api.ts -> context-engine
- [ ] context-engine/context.service.ts -> docs-retrieval
- [ ] docs-retrieval/learning-journey-helpers.ts -> learning-paths
- [x] requirements-manager/requirements-checker.hook.ts -> context-engine
- [x] requirements-manager/step-checker.hook.ts -> context-engine

For each edge, prefer extracting a shared dependency below both engines or injecting the dependency. Remove the matching ALLOWED_LATERAL_VIOLATION_ENTRIES item when the edge disappears; the architecture ratchet will report stale entries.

Progress: #1832 merged and removed the three checked boundaries. The two unchecked boundaries remain in scope here. The small API deduplication follow-up is tracked separately in #1898.

moxious:
## 🧭 Implementation readiness

**Stakes:** Nothing changes for the user. This is internal work. Five cross-engine imports keep four Tier 2 engines coupled, so a change inside one engine can break another.

**Sanity check:** All five listed edges exist at this base in ALLOWED_LATERAL_VIOLATIONS, and none is an edge of a tracked cycle cluster. One edge also has a dynamic import that the ratchet counts, so deleting the static import alone will not clear it.

**Severity:** 🟡 Low — No user-visible behavior changes. The work removes internal import coupling and allowlist entries only.

**Code basis:**
- `src/validation/architecture.test.ts:123-135` — ALLOWED_LATERAL_VIOLATIONS holds all eight lateral entries. The five in scope are on lines 127, 130, 132, 133 and 134.
- `src/requirements-manager/checks/grafana-api.ts:9-10` — Static import of ContextService, used for fetchPlugins, fetchDashboardsByName and fetchDataSources.
- `src/context-engine/context.service.ts:1029-1067` — The three fetch helpers are thin getBackendSrv wrappers. They depend only on logger and types, so a move to `src/lib/` is compatible.
- `src/context-engine/context.service.ts:16-23` — Static import of six functions from the docs-retrieval barrel. This is the heaviest of the five edges.
[...]

PR #1825, opened 2026-09-08
docs(review): anchor docs retrieval ownership

### LINKED ISSUE #1806: Record the docs-retrieval-and-rendering contract anchor with sufficient review history
## Context

Raised by Jay Clifford (`Jayclifford345`) as a non-blocking follow-up on [PR #1716](https://github.com/grafana/grafana-pathfinder-app/pull/1716) (`contract-evolution-docs-retrieval-and-rendering-insufficient-history`, severity: low):

> The extraction is coherent with the current owner, but the review history is partial and this concern has no recorded contract anchor, so the scan cannot rule out older competing ownership conventions.
>
> Follow-up: Record the current docs retrieval and rendering contract in `docs/design/CONCERN_DETAILS.md` in a separate change.

`docs-retrieval-and-rendering` already has a row in `docs/design/CONCERN_DETAILS.md`, but per the contract-evolution gate's own signal, its review history isn't deep enough yet for the gate to treat that row as an authoritative anchor it can diff future extractions against.

## What to do

Working from `docs/design/CONCERN_DETAILS.md`'s `docs-retrieval-and-rendering` row and the current shape of `src/docs-retrieval/**` + `src/components/docs-panel/**`, write down the contract this concern is meant to hold PRs to — ownership boundaries (what "coherent extension" vs. "competing convention" looks like for this subsystem), and enough PR history/precedent (e.g. #894, the pure-seam extraction from #1716, #1692, #1712) that the `/review` contract-evolution gate can anchor against it with confidence next time this area changes.

## Not in scope

This is deliberately separate from [PR #1716](https://github.com/grafana/grafana-pathfinder-app/pull/1716), which the reviewer explicitly scoped this out of.

dvd233:
## Summary

The contract-evolution gate can now load a durable ownership anchor for docs retrieval and rendering instead of reconstructing that contract from raw PR history on every review. The anchor records the fetch, parse, render, and panel-lifecycle boundaries, including the panel's two existing loading paths, the accepted legacy residual, and the conditional App Platform identity rule.

## What to look at

- [Contract anchor](https://github.com/dvd233/grafana-pathfinder-app/blob/b27c7c08bb5c55dd911560f82816718b26a22349/docs/design/CONCERN_DETAILS.md) names `shouldUseDocsLoader`, the `loadDocsTabContentResult` package/docs arm, and the accepted legacy `loadTabContent` arm; it also distinguishes plain-guide `spec.id` preference from the path/journey `packageId` exception.
- [Extraction guard](https://github.com/dvd233/grafana-pathfinder-app/blob/b27c7c08bb5c55dd911560f82816718b26a22349/.cursor/skills/review/scripts/concern-context.test.mjs) pins the owner symbols and the conditional identity wording without relying on bare issue-number or generic-prose matches.

## Why

Without this row, `concern-context.mjs docs-retrieval-and-rendering` returns `contract_anchor: null`, so each future review has to infer the subsystem contract from raw history. A simple implementation inventory would age quickly; this instead records ownership and extension boundaries, with historical implementations serving only as evidence.

## How to verify

1. Run `node .cursor/skills/review/scripts/concern-context.mjs docs-retrieval-and-rendering`; confirm the packet contains evidence `#894 → #1716` and a non-null contract.
2. Trace `shouldUseDocsLoader`, `loadDocsTabContentResult`, and `loadTabContent`; confirm the anchor discloses both existing loading arms, keeps top-level React composition in `components/content-renderer`, and states the path/journey `packageId` exception to plain-guide `spec.id` preference.

## Breaking changes

None. This is additive review metadata and a parser regression test; reverting it restores the previous unanchored behavior.

Closes #1806


docs/design/CONCERN_DETAILS.md
```diff
@@ -58,6 +58,7 @@ Rules:
...
... The docs panel owns tab lifecycle, not a second content pipeline: `loadDocsTabContentResult` selects package loading (`fetchPackageContent` / `fetchPackageById`) versus plain `fetchContent`; `CombinedLearningJourneyPanel` retains time, state, storage, navigation, and other effects, while `resolveDocsLoadAlignment` and `buildDocsLoadSuccessPatch` own only pure post-load decisions (#1716). #894 established this seam by extracting UI and hooks while leaving the Scene model and asynchronous load orchestration intact. #1692 established manifest precedence: the catalogue `packageManifest` wins when it declares a starting location; otherwise the fetched manifest may supply the fallback. #1712 made `package-info-from-url.ts` the shared URL → package-info owner, including `backend-guide:`, and aligned App Platform manifest identity to the current `spec.id`. A coherent extension adds a fetch strategy, parser, or rendered block behind these owners while preserving `ContentFetchResult` and fallback semantics; direct fetch, parsing, or package resolution inside a panel component, a second stateful load finalizer, or moving React rendering back into `docs-retrieval` is a competing convention. |
```
moxious: **blocking · `docs-panel-second-pipeline-undisclosed`**

"The docs panel owns tab lifecycle, not a second content pipeline" does not hold at head. The panel's real selector is `shouldUseDocsLoader` (`docs-panel.tsx:494`), branching to `loadDocsTabContent` (:496) or to a live `loadTabContent` (:504). That second arm calls `fetchContent` directly (:522), reshapes content in-panel (:527-556), finalizes with an inline patch instead of `buildDocsLoadSuccessPatch` (:561), and writes completion storage itself (:566-572). `loadDocsTabContentResult` is only the selector inside the *other* arm (its single call site is :851).

It is not dead code: `shouldUseDocsLoader` is false for a `learning-journey` tab with no `packageInfo`, which `openLearningJourney` creates (:419-429), and the panel's own comment at :409-411 depends on that path. `tab-validation.ts:16-21` names both pipelines in source.

So the closing clause — "a second stateful load finalizer ... is a competing convention" — condemns code already on main without disclosing it. Since the contract-evolution gate diffs future PRs against this row, the next change to `loadTabContent` gets scored `contract_branching` for a residual it did not introduce.

Please name `shouldUseDocsLoader` as the panel selector, scope `loadDocsTabContentResult` to the docs/package arm, and disclose `loadTabContent` as an accepted residual using the table's existing convention (peer rows use "Fracture: …", "Not established: … an accepted residual", "is not contracted here").

One narrowing in your favour: the transform helpers are imported from `docs-retrieval` (:59), so *parsing* is not happening in the panel — only the direct fetch and the second finalizer are.

moxious: **blocking · `spec-id-alignment-omits-path-journey-exception`**

"and aligned App Platform manifest identity to the current `spec.id`" is unconditional, but the head rule is not: `app-platform-resolver.ts:190-191` is `const resolvedId = isPathOrJourney ? packageId : spec.id || packageId;`. The exception is load-bearing and the code explains why at :158-172 — `fetchPackageContent`'s baseUrl hydration re-resolves `manifest.id` in URL-only mode by string-templating `backend-guide:<id>` without verifying, so a drifted `spec.id` would reconstruct a URL the loader cannot fetch. It is pinned twice: `app-platform-resolver.test.ts:466` (path) and :488 (journey).

Reading the clause as provenance rather than current rule does not rescue it — it makes it worse. #1712 (`f454b7bf`) contains a fourth sub-commit, "keep path/journey manifest.id resource-name-equal", whose message reads "the prior commit made buildManifest prefer spec.id unconditionally … Narrow the spec.id preference to non-path/journey manifests". The unconditional form existed only transiently inside that PR and was backed out before merge as a 404-producing defect, so the row records the intermediate state #1712 corrected.

Because the gate diffs conformance against this text, a future PR flattening :191 to prefer `spec.id` unconditionally would read as *conforming* while reintroducing exactly that regression. Please qualify: `spec.id` for a plain guide only; `packageId` wins for `path`/`journey`; `packageId` remains the fallback where no `spec.id` exists. The `block-editor-authoring` row's "`startingLocation` is owned CONDITIONALLY" is the model for recording this.

.cursor/skills/review/scripts/concern-context.test.mjs
```diff
@@ -64,6 +64,23 @@ test('includes the concern anchor and named invariants', () => {
   assert.equal(context.named_invariants[0].name, 'payload-boundary-normalization');
 });
 
+test('records the docs retrieval and rendering ownership chain', () => {
```
moxious, line 67: **suggestion · `prose-pin-assertions-cement-imprecise-wording`**

To be clear up front: this test matches the file's established convention exactly — the adjacent `ai-subsystem` test (:67-82 at base) uses the same `assert.equal(evidence, …)` plus `assert.match(contract, /…/)` shape, prose pin included. Consistency is not the problem.

The problem is what the pins buy. Only the `evidence` equality and one contract-cell match actually exercise the extractor; the other six re-assert substrings of the markdown this same commit wrote. `validateConcernRegistry` (`concern-context.mjs:170-178`) already fails on an unknown or duplicated anchor concern id, so the id-join half is pre-covered.

`assert.match(contract, /#1692/)` and `/#1712/` are unanchored matches over one ~1,600-character cell: they pass identically if the two claims are swapped or attached to the wrong owner — which is precisely the error class the two blocking findings on the anchor row describe. And since `npm run test:review-contract` runs this file in CI, `/competing convention/` plus the `fetchContent`-bearing sentence pins are exactly the text those fixes must reword, so the test as written reds CI on its own fix.

Suggest keeping the `evidence` equality and the owner-symbol matches, dropping `/competing convention/` and the bare PR-number regexes — or pinning the PR numbers positionally so attribution, not presence, is what fails.

moxious (changes requested):
Thanks for taking on the `docs-retrieval-and-rendering` anchor — this is the row that closes the `insufficient_history` gap #1806 was filed for, and writing an ownership chain this dense from four separate PRs is genuinely hard work.

## Merge contract

1. [blocking] **docs-panel-second-pipeline-undisclosed — The anchor asserts the docs panel is not a second content pipeline, but head has two and the row names the pre-existing one a competing convention without disclosing it** (medium · correctness-and-reliability)
   The row's load-bearing ownership sentence reads "The docs panel owns tab lifecycle, not a second content pipeline: `loadDocsTabContentResult` selects package loading versus plain `fetchContent`". At head the panel's actual selector is `shouldUseDocsLoader` (docs-panel.tsx:494), which branches to `loadDocsTabContent` (:496) or to a live sibling pipeline `loadTabContent` (:504). `loadTabContent` calls `fetchContent` directly (:522), reshapes content in-panel (:527-556), finalizes with an inline patch rather than `buildDocsLoadSuccessPatch` (:561), and writes completion storage itself (:566-572). `loadDocsTabContentResult` is only the selector inside the other arm (called once, at :851). The branch is not vestigial: `shouldUseDocsLoader` is false for a `learning-journey` tab with no `packageInfo`, which `openLearningJourney` creates (:419-429), and the panel's own comment at :409-411 depends on that path; `tab-validation.ts:16-21` names both pipelines in source. So the row's closing clause, which lists "a second stateful load finalizer" as a competing convention, condemns code already on main without disclosing it — and the contract-evolution gate diffs future PRs against this text. The next PR touching `loadTabContent` will either be scored `contract_branching` for a residual it did not introduce, or a reviewer will find the anchor incoherent and spend the specialist slot proposing a replacement. That is a worse outcome than the `insufficient_history` finding on #1716 that commissioned this row, because it converts a known-unknown into a confident wrong answer. Two independent skeptics confirmed the control flow; one narrowing they established is that parsing is NOT happening in the panel (the transform helpers are imported from `docs-retrieval`), so the undisclosed residual is the direct fetch plus the second stateful finalizer, not in-panel parsing.
   Required: Reword the ownership sentence to match head: name `shouldUseDocsLoader` as the panel's pipeline selector and `loadDocsTabContentResult` as the selector within the docs/package arm, then disclose the legacy `loadTabContent` arm as an accepted residual using the table's existing convention — peer rows carry "Fracture: #1599, #1600.", "Not established: ... an accepted residual", and "is not contracted here" for exactly this. Keeping "a second stateful load finalizer is a competing convention" is fine once the known one is named, because then the gate can tell a new one from the residual.

2. [blocking] **spec-id-alignment-omits-path-journey-exception — "aligned App Platform manifest identity to the current `spec.id`" is stated unconditionally, but paths and journeys deliberately keep the resource name** (medium · correctness-and-reliability)
   The row's #1712 clause records the `spec.id` rule with no qualifier. The head rule is conditional: `app-platform-resolver.ts:190-191` is `const resolvedId = isPathOrJourney ? packageId : spec.id || packageId;`. The exception is load-bearing and the code says why at :158-172 — `fetchPackageContent`'s baseUrl hydration re-resolves `manifest.id` in URL-only mode by string-templating `backend-guide:<id>` without verifying, so a drifted `spec.id` would silently reconstruct a URL the loader cannot fetch. Two tests pin it (app-platform-resolver.test.ts:466 and :488). The historical-framing defence — that the clause reports what #1712 did rather than the current rule — was tested by both skeptics and inverts: #1712 (f454b7bf) contains a fourth sub-commit titled "keep path/journey manifest.id resource-name-equal", whose message says "the prior commit made buildManifest prefer spec.id unconditionally ... Narrow the spec.id preference to non-path/journey manifests". The unconditional version existed only transiently inside that PR and was corrected before merge as a 404-causing defect, so the row restates the exact intermediate state #1712 fixed. Because an anchor is what the gate diffs conformance against, a future PR flattening :191 to prefer `spec.id` unconditionally would read as conforming while reintroducing the unfetchable-`backend-guide:`-URL regression. Recording only the positive half of a two-sided invariant is how an anchor blesses the regression the code went out of its way to prevent.
   Required: Qualify the clause: `spec.id` is preferred for a plain guide only; `packageId` (the immutable resource name) wins for `path` and `journey` because URL-only re-resolution string-templates `backend-guide:<id>`; and `packageId` remains the fallback where no `spec.id` exists. The table already records this kind of conditionality at length — the `block-editor-authoring` row's "`startingLocation` is owned CONDITIONALLY" is the model.

## Follow-ups

1. [follow_up] **react-in-docs-retrieval-boundary-overstated — The React-boundary clause contradicts the 15 components already in `docs-retrieval` and the row's own coherent-extension clause** (low · cross-cutting-architecture)
   The row's closing sentence blesses "a coherent extension adds a fetch strategy, parser, or rendered block behind these owners" and, in the same breath, calls "moving React rendering back into `docs-retrieval`" a competing convention. But `src/docs-retrieval/components/docs/` holds 15 React components at head and the barrel deliberately re-exports nine of them (index.ts:57-68), so adding a rendered block means adding a React component exactly where the row says not to. The real invariant is narrower and the barrel states it precisely at index.ts:48-51: only top-level composition (`ContentRenderer`) moved out, to keep `docs-retrieval` a pure parser tier. As written a gate can reach either verdict on a future block-renderer PR from the same text, which is the failure mode an anchor exists to prevent.
   Follow-up: Narrow the claim to top-level content composition: `ContentRenderer` lives in `src/components/content-renderer/` and is deliberately not re-exported through the barrel, while leaf block renderers remain in `src/docs-retrieval/components/docs/` and are re-exported. The anti-pattern then becomes "moving top-level content composition back into `docs-retrieval`".

## Suggestions

1. [suggestion] **prose-pin-assertions-cement-imprecise-wording — Six of the eight assertions pin prose rather than extractor behaviour, and the PR-number pins verify presence rather than attribution** (low · testing-and-verification)
   The new test follows the file's established convention exactly — the adjacent `ai-subsystem` test uses the same `assert.equal(evidence, ...)` plus `assert.match(contract, /.../)` shape, including a prose-substring pin — so consistency is not the issue and the test should not be rejected on that ground. The issue is what the pins buy. Only the `evidence` equality and one contract-cell match exercise the extractor; the other six re-assert substrings of the markdown the same commit wrote, and `validateConcernRegistry` (concern-context.mjs:170-178) already fails on an unknown or duplicated anchor concern id. `assert.match(contract, /#1692/)` and `/#1712/` are unanchored matches over a single 1,600-character cell, so they pass identically if the two claims are swapped or attached to the wrong owner — which is precisely the error class the two blocking findings above describe. Because `npm run test:review-contract` runs this file in CI, the `/competing convention/` pin and the sentences carrying `fetchContent` are exactly the text the blocking fixes must reword, so the test as written reds CI on its own fix.
   Suggested: Keep the `evidence` equality and the owner-symbol matches (`fetchContent`, `ContentRenderer`, `loadDocsTabContentResult`, `resolveDocsLoadAlignment`), and drop `/competing convention/` and the bare `/#1692/` `/#1712/` regexes. If the PR references must be pinned, pin them positionally so that attribution, not mere presence, is what fails.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1825
Summary: anchor docs retrieval ownership
Verdict: Request Changes
Results: 2 blockers, 1 non-blocking finding, 1 follow-up

Next step: reword the two blocking clauses so the row matches head — name `shouldUseDocsLoader` and disclose the legacy `loadTabContent` arm as an accepted residual, and qualify the `spec.id` rule with its path/journey exception. Worth doing together with the test finding, since two of the current assertions pin the exact prose those fixes must reword and would red CI on their own fix. Push when ready and we will re-review on the new head — CI is already authorized, and the one red check (`publish-report`) is a known fork-permission limitation on community PRs, not anything you did.

loglapa:
Coordination note: #1832 implements the Tier 2 boundary cleanup approved in #1763 and moves shared content resolution into Tier 1 (`src/lib/content-fetcher*`, relative-URL resolution, and pure journey helpers). Existing `docs-retrieval` exports remain re-export facades over the same implementation; panel lifecycle and React composition ownership remain unchanged.

This matters for the proposed ownership anchor and the verification step saying fetch/parse stays in `docs-retrieval`: the public facade remains there, while shared resolution implementation moves to Tier 1. Please account for that distinction when these PRs land, so the anchor does not classify the approved extraction as a competing content pipeline.

dvd233:
@moxious Thanks for the precise review. `110461ea` addresses both blockers and the test suggestion:

- The anchor now names `shouldUseDocsLoader` and the `loadDocsTabContentResult` package/docs arm, discloses the existing `loadTabContent` arm as an accepted residual, and narrows the competing-convention wording to new pipelines.
- The App Platform identity rule now states that plain guides prefer `spec.id`, while `path` and `journey` retain the immutable `packageId` because URL-only resolution string-templates `backend-guide:<id>`.
- The extraction test pins the owner symbols and the conditional identity wording rather than bare issue numbers or generic prose.

Verification completed: review-contract (77 tests), typecheck, full lint, and Prettier. Ready for another look.

moxious (approved):
Thanks for turning this around so precisely. Both blockers are fixed exactly as scoped, and the `#1712` clause in particular now reproduces the code's own reason for the path/journey exception (`app-platform-resolver.ts:159-167`) rather than a plausible-sounding substitute — that is the hard half of writing an anchor, and it is right. The `loadTabContent` disclosure holds on all four sub-claims I raised, and the test now pins attribution instead of mere presence, so it no longer reds CI on its own fix.

I re-reviewed at `b27c7c08` after the `upstream/main` merge landed. That merge is a pure catch-up: your net contribution against the new base is byte-identical to what I reviewed before it — the same two files and the same 25 added lines. The only difference in the diff is an adjacent context line, because #1841 rewrote the neighbouring `interactive-engine` row. Nothing of yours was dropped in the resolution, and all three re-verified anchors still hold at the new head.

Next step: nothing is required from you. One wording refinement is tracked separately as #1859 — the row names `shouldUseDocsLoader` as the selector, which is true but is one of three dispatch inputs, and since this row is what the contract-evolution gate diffs against it is worth tightening whenever the row is next touched. That is a note for the anchor, not a change request against this PR. CI on this head is still held pending maintainer authorization, which is routine for a community PR and unrelated to your change.

PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1825
Summary: anchor docs retrieval ownership
Verdict: Approve

PR #1833, opened 2026-09-08
fix: preserve pending request fallbacks on disconnect

### LINKED ISSUE #1828: Cross-tab: failAllPending resolves in-flight fix requests with null, swallowing the disconnect error
`failAllPending` in `src/global-state/controller-channel.tsx` resolves **every** pending request with `null`, including in-flight `fix-requirement` requests that were promised a `FixOutcome` fallback. A live tab that closes or reloads mid-fix therefore produces a swallowed `TypeError` reported as a generic message, on exactly the disconnect branch this code exists to handle.

## Mechanism

- `failAllPending` resolves each pending entry with `null` (`controller-channel.tsx:197-206`).
- `PendingRequest.resolve` is typed to accept `null` (`:58-61`), and the `resolve as PendingRequest['resolve']` cast at `:319` erases the generic — so `Promise<FixOutcome>` is not enforced at the call site.
- The per-caller fix fallback is `{ ok: false, error: 'No live tab responded' }` (`:355`), but `failAllPending` never consults it.
- `step-checker.hook.ts:705` then evaluates `!result.ok` on `null`, throwing a `TypeError` into the catch at `:739-750`, which reports the generic `Failed to fix requirements` instead of `No live tab responded`.

## Triggers

Both are ordinary, not edge cases:

- stale heartbeat at `:284` (`HEARTBEAT_STALE_MS = 5000`)
- unconditional unmount at `:304`

A fix action frequently navigates, which is precisely what closes or reloads the live tab — so the common case for a fix is the case that hits this.

## Suggested fix

Store each request's fallback alongside its pending entry, or give `failAllPending` a per-kind failure value, so a fix request always settles as a `FixOutcome`. Add a `controller-channel` test covering a stale heartbeat with a fix in flight.

## Provenance

Found during review of https://github.com/grafana/grafana-pathfinder-app/pull/1824 and confirmed by two independent verifiers. Pre-existing on `main` in files that PR does not touch, so it was withheld from the contributor's review and filed here instead.

ravik453:
## Summary

Fixes #1828

Pending cross-tab requests were always resolved with `null` when the controller channel was torn down. This caused `requestFix()` callers to receive an unexpected result and could lead to a runtime error when accessing `result.ok`.

This change stores the fallback associated with each pending request and uses it when resolving requests during channel cleanup, including stale-heartbeat handling.

## Changes

- Store the request-specific fallback in each pending request entry.
- Resolve pending requests with their stored fallback in `failAllPending()`.
- Add a regression test covering an in-flight `requestFix()` when the provider unmounts.

## Testing

- `npm test -- --runInBand`
- 55 test suites passed
- 816 tests passed

### DIFF
--- a/src/global-state/controller-channel.test.tsx
+++ b/src/global-state/controller-channel.test.tsx
@@ -350,6 +350,45 @@ describe('ControllerChannelProvider', () => {
     await waitFor(() => expect(screen.getByTestId('fix')).toHaveTextContent('ok:true'));
   });
 
+  it('resolves an in-flight requestFix with the fix fallback when the provider unmounts', async () => {
+    let resolveOutcome: ((value: string) => void) | undefined;
+    const outcome = new Promise<string>((resolve) => {
+      resolveOutcome = resolve;
+    });
+
+    function PendingFixProbe() {
+      const channel = useControllerChannel();
+
+      return (
+        <button
+          onClick={() =>
+            channel
+              ?.requestFix('s1', { requirements: 'navmenu-open', fixType: 'navigation' })
+              .then((result) => resolveOutcome?.(`ok:${result.ok}`))
+          }
+        >
+          fix
+        </button>
+      );
+    }
+
+    const transport = new FakeCrossTabTransport();
+    const { unmount } = render(
+      <ControllerChannelProvider transport={transport} pairing={TEST_PAIRING}>
+        <PendingFixProbe />
+      </ControllerChannelProvider>
+    );
+
+    await pairWithLive(transport);
+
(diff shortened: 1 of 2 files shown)


moxious (approved):
Thanks for tracking this down and fixing it — the `failAllPending` fallback bug was a real crash risk for `requestFix()` callers on disconnect (accessing `.ok` on `null`), and the added test exercises exactly that path.

PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1833
Summary: preserve pending request fallbacks on disconnect
Verdict: Approve

Nothing further needed here — good to merge.

PR #1841, opened 2026-09-09
fix(interactive-engine): credit guided-tour progress on performance, not display

tomglenn:
## Summary

The progress bar in the guided-tour comment box filled from the step being **displayed** rather than the steps the reader had **performed**. It drew its width from `(current + 1) / total`, so a four-step tour showed 25% before the reader had done anything and read 100% on the last step with nothing completed.

The bar now draws from `completedSteps`, which is the same evidence the step dots already use — `GuidedHandler` pushes to it only on `completed` or `skipped`.

## What to look at

`CommentBoxStepInfo.progress` (`navigation-manager.ts`) is a new discriminator, because the comment box has two consumers that mean different things by `completedSteps`:

| Consumer | Declares | Behaviour |
|---|---|---|
| guided blocks (`guided-handler.ts`) | `'performed'` | fills from steps the reader performed — the fix |
| bubble tours (`BubbleTour.tsx`, e.g. `BlockEditorTour`) | `'position'` | unchanged; for a read-only tour, display *is* the evidence |

Without that split the change would have reached the bubble tours and capped their bar at 83%, so the discriminator is what contains it. `progressBarPercent` is the only place the formula lives, and the `interactive-engine` contract anchor in `docs/design/CONCERN_DETAILS.md` now records that.

The "Step N of M" badge deliberately stays on `current + 1`. That is a position, and positions are 1-based.

## Why

The bar is a tour-local measure: `total` counts the steps inside one `guided` block, which `OPAQUE_PARENT_BLOCK_TYPES` keeps out of the guide-wide denominator. It is not guide progress and is not meant to converge on it — hence the code comment and the contract tests, which pin the tour-local denominator alongside the crediting fix.

## How to verify

- `npx jest --coverage=false src/interactive-engine src/components/BubbleTour` — 22 suites, 423 tests.
- Reverting the `'performed'` branch to `(current + 1) / total` fails three of the four new bar tests, so the regression is genuinely guarded.
- Each consumer has its own test, so a future change to one cannot silently move the other.

## Known follow-up

Two adjacent issues are out of scope here and want their own work:

- **Restart keeps stale credit.** `GuidedHandler.resetProgress()` is called only from `live-tab-executor.ts`; the in-app guided path does not call it on restart or retry. Perform steps 0–2 of a four-step block, cancel, then "↻ Restart tour" and the bar shows 75% next to "Step 1 of 4". The step dots already carry the same stale state, so fixing it is a behaviour change of its own.
- **100% is not painted live.** `finishGuidedStep` tears the comment box down as each step settles, so the highest value a reader sees is `(total - 1) / total`. A filled bar on completion would need a repaint in the guided loop, not a change to this formula.

## Breaking changes

None. `progress` is required on `CommentBoxStepInfo`, and both call sites set it; no other caller passes `stepInfo` at all.

## Reversibility

Fully reversible — revert the commits. No persistence, no durable records, no config.


moxious (changes requested):
Blockers:

1. [blocking] **guided-restart-paints-stale-credit — Restart, try again, redo, and section reset paint the previous run's credit as if it were earned** (medium · correctness-and-reliability)
   `GuidedHandler.completedSteps` is only cleared by `resetProgress()`, and the sole production caller is `live-tab-executor.ts:246`. The in-app path never calls it: the handler is a `useMemo(..., [])` at `interactive-guided.tsx:276-280`, a section reset is a prop bump against a stable `key: stepInfo.stepId` (`interactive-section.tsx:1260-1266`) rather than a remount, and `cancel()` preserves the array. So every in-app re-entry - `handleRetry` (wired to both '↻ Try again' at :1014 and '↻ Restart tour' at :1064), `executeStep` at :388, `handleStepRedo` at :703, and the `resetTrigger` effect at :292 - replays the block against stale indices. Two skeptics reproduced it: running two 2-step sequences on one handler, run B's first paint carried `completedSteps: [0,1]`, painting a 100% bar next to the 'Step 1 of 2' badge in the same box. The stale array pre-dates this PR and already tinted the step dots, but the previous bar was position-derived and stayed correct through a restart, so this change is what turns that state into a quantified false claim - in the one indicator the PR set out to make truthful, and against durable completion records that read zero.
   Required: Call `guidedHandler.resetProgress()` where an in-app run begins - alongside `setCurrentStepIndex(0)` in `executeStep` at `interactive-guided.tsx:388`, which covers retry, redo and the reset trigger together - matching `live-tab-executor.ts:246`. Add a guided-handler test that runs two sequences on one instance and asserts the second run's first paint carries `completedSteps: []`; today's only `resetProgress` test just asserts the method is defined.

Follow-ups:

1. [follow_up] **guided-overlay-bar-cannot-complete — The bar credits a step only after its box is destroyed, so it tops out at (total-1)/total and never signals completion** (medium · interactive-engine)
   `finishGuidedStep` calls `cleanupListeners(true)` - which removes every `.interactive-comment-box` via `navigation-manager.ts:93` - before `completedSteps.push(stepIndex)` (`guided-handler.ts:206-212`). The box painted for step N can therefore only carry credit for steps 0..N-1, a one-step lag both skeptics reproduced by execution. On the two shipped guided blocks (2 steps each) the bar now reads 0% then 50% where it read 50% then 100%, and a one-step block - legal per `json-guide.schema.ts:422` - paints 0% for its whole life. The PR's own 100% contract test only passes by hand-feeding `completedSteps: [0,1,2,3]`, a state no live paint can hold. This is not misinformation: the value matches the in-panel bar at `interactive-guided.tsx:972` exactly and correctly withholds credit for unperformed work, and a 'Completed' badge renders at :1090. But the crediting fix is only half-delivered while the bar cannot reach the state its own test describes.
   Follow-up: Move the `completedSteps.push` in `finishGuidedStep` above `cleanupListeners(true)`, or repaint once after the final step settles so the bar can fill. You already track this as a known follow-up; if it stays out of scope, record that in the contract anchor so the 100% test is not read as describing a reachable state.

2. [follow_up] **anchor-evidence-not-bumped — The interactive-engine anchor gains a clause but keeps its old evidence range** (low · ai-subsystem)
   The `interactive-engine` row's evidence cell still reads `#1616 → #1757` while its contract cell gains the `CommentBoxStepInfo.progress` clause. #1757 set the precedent inside this same row, bumping the cell from `#1616` to `#1616 → #1757` in the same commit that extended the prose, and every multi-touch row carries that origin-to-most-recent form. The contract-evolution adapter returned `coherent_extension` with `requires_observation: false`, so no automated check reports this - a later scan resolving the anchor's evidence lands on the DOM-ingress PR rather than on the PR that established this contract.
   Follow-up: Extend the evidence cell to `#1616 → #1841` in the same edit that adds the clause.

Suggestions & nits:

1. [suggestion] **panel-bar-duplicates-unanchored-formula — A second progress bar for the same guided block keeps its own inline formula the anchor does not cover** (medium · interactive-engine)
   `interactive-guided.tsx:969-980` renders `.interactive-guided-progress-fill` with an inline width of `(currentStepIndex / internalActions.length) * 100`, a real styled bar (`interactive.styles.ts:1003-1020`) measuring the same quantity as the comment-box bar for the same block, on screen at the same time. For a fresh run the two are now numerically identical - during step i, `completedSteps.length === i === currentStepIndex` - which is worth stating plainly: this PR removes a visible one-step disagreement between two bars rather than creating one. But that alignment is coincidental, not enforced: nothing ties the panel bar to `CommentBoxStepInfo`, no test pins them together, and the anchor a future agent reads names only the comment-box bar.
   Suggested: Name the in-panel bar in the same anchor sentence, or have it derive from the same helper, so the two bars for one guided block cannot drift apart again.

2. [nit] **progress-bar-comment-misplaced — The new inline comment annotates a line that no longer mentions total, and restates the interface JSDoc** (low · correctness-and-reliability)
   `navigation-manager.ts:961` puts `// total is the consumer's own step set ... never the guide.` above a statement in which `total` no longer appears. The same invariant is already in the `CommentBoxStepInfo` JSDoc six lines of code away (:36-42), in the contract anchor, and again as a `Guard:` comment in the contract test (:876) - four copies, none of them on the `total` field they constrain. Separately, `guided-handler.ts:557`'s pre-existing `// Build step info for progress display in comment tooltip` restates the declaration directly below it and sits inside a hunk this PR edits, so the AGENTS.md trim-on-touch rule applies.
   Suggested: Drop the comment at `navigation-manager.ts:961` and move the `total` sentence onto the `total` field in the `CommentBoxStepInfo` JSDoc; delete `guided-handler.ts:557` while editing that hunk.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1841
Summary: credit guided-tour progress on performance, not display
Verdict: Request Changes
Results: 1 blocker, 2 non-blocking findings, 2 follow-ups

moxious (approved):

Follow-ups:

1. [follow_up] **guided-reset-wiring-unguarded — The delta's only production line is unverified - deleting the resetProgress() call keeps every test green** (medium · testing-and-verification)
   The round-1 blocker is genuinely fixed and I confirmed the invariant rather than the anchor: `guidedHandler.resetProgress()` at `interactive-guided.tsx:388` sits ahead of the loop in `executeStep`, `handleRetry` (both '↻ Try again' and '↻ Restart tour') calls `executeStep`, and `handleStepRedo` and the `resetTrigger` effect only reset local UI - they cannot paint a box without a later `executeStep`. The hook's own `GuidedHandler` never runs steps, so the component and `live-tab-executor.ts:246` are the only two paint paths and both reset. What is missing is a guard on that. Three independent mutation experiments deleted only line 388: over the four directly related suites, 103 of 103 stayed green; over all of `src/components/interactive-tutorial` + `src/interactive-engine`, 52 suites and 941 tests stayed green; over the whole repository, 535 of 535 suites and 8567 tests stayed green, with `eslint` and `tsc --noEmit` also exiting 0. As a control, emptying `resetProgress`'s body (`guided-handler.ts:75-77`) does turn the new test at `guided-handler.test.ts:115` red - so the delta's tests guard the method, which was already correct before this PR and was never the broken half. The missing call was the broken half, and it is the half still unverified. `interactive-guided.test.tsx:137` even mocks `resetProgress`, so the observation point exists and is simply unobserved.
   Follow-up: Assert the call site, not only the method: hoist the existing `resetProgress: jest.fn()` mock at `interactive-guided.test.tsx:137` to a module-level fn and assert it fires on a run start - ideally ordered before the first `executeGuidedStep`, using the `callOrder` harness the file already applies to `waitForReactUpdates` vs `executeGuidedStep`.

Suggestions & nits:

1. [suggestion] **guided-reset-invariant-caller-owned — The freshness invariant stays caller-owned, and the new characterization test makes the handler-side fix a red test** (low · correctness-and-reliability)
   `resetProgress()` (`guided-handler.ts:75-77`) remains the only writer that clears `completedSteps`, and `executeGuidedStep` never notices that a `stepIndex === 0` arrived with credit already banked - so `interactive-guided.tsx:388` and `live-tab-executor.ts:246`, the only two in-app callers, must each remember it independently. Placement is correct today and I found no live regression. But the bug class round 1 caught is 'a caller forgot', and the delta leaves that class open for a third caller while adding a test - `carries stale credit into a second sequence when it is not called` at `guided-handler.test.ts:125-132` - that asserts `completedSteps: [0, 1]` on run B's first paint. That pins the absence of a handler-side guard as expected behaviour, so the natural defence-in-depth fix becomes a test-breaking change rather than a green one.
   Suggested: Pick one rather than both: reframe that second test so it reads as characterization a future self-healing handler is free to break, or move the invariant into `executeGuidedStep` (clear on `stepIndex === 0`) and let the two caller-side calls become belt-and-braces.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1841
Summary: credit guided-tour progress on performance, not display
Verdict: Approve with Minor

PR #1842, opened 2026-09-09
refactor(guide-stats): key completion positions by runtime step id

tomglenn:
## Intent

Extend the shared guide-stats module (work item A2 of the Pathfinder completion-model cutover plan), with the amendment that changes its size.

Authority: docs/design/COMPLETION-MODEL.md on main.

Scope, and nothing else:

1. Step-id positions in src/lib/guide-stats/. The completion numerator needs a position for a 'Do it', and the runtime dispatches completion under a runtime step id, not an author block id — only 1 of 137 completable blocks in the bundled library carries an author id, so positionsById alone resolves essentially nothing. computeGuideBlockIndex gains an injected resolveStepId option and publishes positionsByStepId alongside positionsById; progress.ts's evidencedPosition consults the step id first and falls back to the author id. The resolver is injected rather than imported so lib/guide-stats stays dependency-free.

2. The one file move the plan names: resolveStepIdForBlock moves out of src/integrations/assistant-integration/ai-fix-step-id.ts (tier 3, so a tier-1 consumer cannot import it) into src/global-state/guide-step-id-resolver.ts (tier 1, beside step-id.ts). This is the only file move in the plan.

3. THE AMENDMENT, and the reason this item is bigger than the plan says. The plan assumed the existing resolver could be reused as-is. It could not: it resolved 3 of the 9 block types that can emit completion evidence — interactive, multistep and guided only — with quiz, terminal, terminal-connect, code-block, challenge and the blocking-datasource shape of input all returning undefined, while the parser resolves all nine. Shipping that unamended ships a silent zero: a quiz- or terminal-driven guide would evidence nothing and read 0% while looking healthy. So the resolver had to cover all nine completable shapes before anything depends on it, deriving each one the way json-parser.ts does. That is six new deriveStepId call shapes plus their parity cases, not the six lines the plan implied.

4. The parity test must iterate the completable set rather than a sample — this was explicitly required, because a sampled test is exactly what let the 3-of-9 gap survive. The new src/lib/guide-stats/progress.parity.test.ts drives its cases from COMPLETION_AFFORDANCE_BLOCK_TYPES plus input, so a newly tracked completable type fails the test until a fixture exists, and it asserts the resolver derives the same value json-parser.ts assigns to props.stepId in each namespace the parser uses (top level, section, assistant), plus a sweep over the bundled corpus. It lives in src/lib/guide-stats/ deliberately: docs/design/CONCERNS.md maps that directory to the completion-records concern, so a PR weakening the file activates that reviewer.

5. meanOfMemberPercentages for path rollup, in a new src/lib/guide-stats/rollup.ts. A path's percentage is the arithmetic mean of its members' percentages, weighted equally regardless of member length (decision 4), and the same formula recurses for any level above a path (decision 5). Members contribute a percentage and never a block count, which is what removes path-wide totals entirely. 'complete' is deliberately every member at 100 and never percent === 100, because guideProgressAtPosition reserves 100 for completion and caps everything else at 99, so a genuinely finished path can legitimately round to 99. An empty member list is {percent: 0, complete: false}. It has no consumer yet — the consumers are later work items — and it lives in the module because a later architecture test needs exactly one place where a progress percentage is computed.

Deliberate decisions made while implementing, which a reviewer reading only the diff would not know:

- The author-id short-circuit in the resolver was moved from a single check at the top of the function to a per-case 'block.id ?? derive(...)'. That is a behaviour change and it is intended: the parser assigns no props.stepId at all to a passive input or to a non-step block type, so returning an author id for those would put a key in positionsByStepId that the runtime never dispatches. The bundled-corpus sweep caught this on json-guide-demo's 'advisory-data-check' block.
- The blocking datasource input is the one completable shape whose step id is always the author's, because the schema requires an explicit id on a blocking data check so its completion records survive an edit.
- The counter's traversal now carries the parser's step-id namespace and JSON path alongside the existing numeric path, because the derived id is a hash over {sectionId, index, action, refTarget} and a divergence in how those are spelled would silently rekey every anonymous block's progress. It reproduces the parser's spellings exactly: __standalone__ at the top level, section-<id> or section:<path> for a section, assistant:<path> for an assistant container.
- No step id is resolved for blocks under a collapsible, matching the parser, whose collapsible converter passes no step context at all.
- resolveCountedBlockStepId is a thin adapter carrying a single documented cast, because the counter deliberately walks a structural CountableBlock rather than the JsonBlock union the resolver switches over, and the alternative was a cast at every future injection site.
- Widening the resolver also makes materializeStepIds address more block types on the AI-fix apply path. That is strictly additive and cannot change what the apply path asks of it; its existing parity tests pass unchanged.

Explicitly out of scope, deferred to later work items in the same plan: any consumer of positionsByStepId or meanOfMemberPercentages; the universal 'Mark complete' button and its analytics event; rewriting path and journey progress to use the mean; the docstring and contract-anchor sweep for the retired finalCompletablePosition predicate; the guided-tour bar; the frozen block index and content-renderer seam; making the completion store a producer; removing navigation completion credit; the teaching nudge; and the architecture test forbidding private percentage calculations. Nothing else was to be touched.

This change extends the completion-denominator contract; it does not replace it. There are no event payload changes, no runtime behaviour changes and no percentage moves — nothing user-visible to verify.

Known pre-existing local failures, unrelated to this change and present on main: src/integrations/coda/coda-api.test.ts fails because the installed @grafana/coda-client lacks exports the source imports, and src/validation/import-graph.test.ts fails one case because this Node build does not list node:test in builtinModules.

## What Changed

- `computeGuideBlockIndex` takes an injected `resolveStepId` option and publishes `positionsByStepId` alongside `positionsById`, carrying the parser's step-id namespace and JSON path through its traversal (`__standalone__`, `section-<id>` / `section:<path>`, `assistant:<path>`, and no step context under a `collapsible`). `progress.ts`'s `evidencedPosition` now consults the step id first and falls back to the author id. The resolver is injected rather than imported so `lib/guide-stats` stays dependency-free.
- `resolveStepIdForBlock` moves from `src/integrations/assistant-integration/ai-fix-step-id.ts` (tier 3) to the new `src/global-state/guide-step-id-resolver.ts` (tier 1, beside `step-id.ts`) and is widened from 3 to all 9 completable block shapes — quiz, terminal, terminal-connect, code-block, challenge and the blocking-datasource `input` join interactive, multistep and guided — with the author-id short-circuit moved from a single top-of-function check to a per-case `block.id ?? derive(...)` so passive and non-step blocks no longer emit a key the runtime never dispatches. `resolveCountedBlockStepId` adapts the structural `CountableBlock` for the counter.
- Adds `src/lib/guide-stats/rollup.ts` with `meanOfMemberPercentages` (equal-weight mean of member percentages; `complete` is every member at 100, never `percent === 100`; empty list is `{percent: 0, complete: false}`), currently unconsumed. New `progress.parity.test.ts` drives cases from `COMPLETION_AFFORDANCE_BLOCK_TYPES` plus `input` and asserts the resolver matches `json-parser.ts`'s `props.stepId` in each namespace, with a bundled-corpus sweep; `rollup.test.ts` and extra `progress.test.ts` cases cover the rest. Docs updated across `COMPLETION-MODEL.md`, `STEP_MODEL.md`, `AI_FIX.md`, `CONTEXT_INDEX.md`, `CONCERNS.md`, `CONCERN_DETAILS.md` and `tracked-step-types.mdc`.

## Risk Assessment

✅ Low: The change is additive and well-bounded — a new optional `positionsByStepId`/injected resolver, a like-for-like file move, and a new consumerless `meanOfMemberPercentages` — with no production caller injecting the resolver yet, derived-id parity directly pinned against `json-parser.ts` across all nine completable shapes and three namespaces plus a bundled-corpus sweep, the stamped-summary shape and counting rule untouched, and the one known keying gap (snippet-ref) documented at the call site with a measured justification for deferring; both remaining findings are unreachable-today robustness and test-ratchet tightening.

## Testing

`Installed the missing dependencies, then ran the targeted guide-stats, step-id, AI-fix apply-path, and architecture suites (all pass, 130 + 27 tests) plus the review-contract suite (76 pass). To show the intent rather than just green tests, I swapped the pre-change 3-of-9 resolver back in and the new parity test fails 17 of 43 cases, then passes 43/43 with the shipped resolver — the fail-before/pass-after the amendment claims. I also produced a reviewer-readable before/after transcript driving guideProgress with the parser's own step ids over the bundled json-guide-demo: quiz clicks went from a lost 0% to 72%/75%, step-id coverage from 10/12 to 12/12, and the stale advisory-data-check author-id key disappeared, with a terminal-only guide going from 0% end to end to 50% -> 75%. The path rollup and the new-completable-type ratchet were exercised the same way, and the concern-routing doc change was verified by parsing it with the review skill's own extractor. No visual evidence: the change is confined to tier-1 pure functions with no consumer yet, and the intent states there are no runtime, event-payload, or percentage changes visible in the UI — the CLI transcripts above are the end-user-facing surface available. Only pre-existing failure reproduced is the node:test builtin case in import-graph.test.ts; coda-api.test.ts passes on a fresh install.`

<details>
<summary>Evidence: Completion numerator before/after: bundled json-guide-demo and a terminal-only guide</summary>

<code>guide: json-guide-demo denominator: 33 blocks completable: 12&#10;&#10;A reader clicks the affordance on each block below. The runtime dispatches&#10;completion under the parser&#39;s props.stepId; the numerator looks that key up.&#10;&#10;block type parser stepId before after&#10;----------------------------------------------------------------------------------&#10;interactive-step __standalone__-step-12oopkb 33% 33%&#10;interactive-step section-interactive-demo-step-h9tdq039% 39%&#10;interactive-step section-interactive-demo-step-14xhpcn42% 42%&#10;interactive-step section-button-demo-step-85ht1f 48% 48%&#10;interactive-step section-formfill-demo-step-1ghgrv9 54% 54%&#10;interactive-multi-step __standalone__-step-2i6dkt 60% 60%&#10;interactive-guided __standalone__-step-ft77cj 66% 66%&#10;quiz-block section-quiz-section-step-1hw1dkb 0% (lost) 72%&#10;quiz-block section-quiz-section-step-13zy8lj 0% (lost) 75%&#10;interactive-step section-quiz-section-step-5oy3cx 78% 78%&#10;interactive-step section-admin-section-step-16lhsb1 84% 84%&#10;datasource-check-step blocking-data-check 93% 93%&#10;----------------------------------------------------------------------------------&#10;step ids the parser assigns: 12&#10;resolved before this change: 10&#10;resolved after this change: 12&#10;keys the runtime never dispatches, before: [&#34;advisory-data-check&#34;]&#10;keys the runtime never dispatches, after: []&#10;&#10;&#10;guide: a terminal-only guide (4 blocks, 2 completable)&#10;&#10;Reader runs both commands, in order. Percentage after each click:&#10;&#10;click before after&#10;----------------------------------------------------------&#10;terminal-step &#34;__standalone__-s…&#34; 0% 50%&#10;terminal-step &#34;__standalone__-s…&#34; 0% 75%&#10;----------------------------------------------------------&#10;Before: every click was lost, so the guide sat at 0% while looking healthy.</code>

```text
guide: json-guide-demo   denominator: 33 blocks   completable: 12

A reader clicks the affordance on each block below. The runtime dispatches
completion under the parser's props.stepId; the numerator looks that key up.

block type              parser stepId                       before     after
----------------------------------------------------------------------------------
interactive-step        __standalone__-step-12oopkb         33%        33%
interactive-step        section-interactive-demo-step-h9tdq039%        39%
interactive-step        section-interactive-demo-step-14xhpcn42%        42%
interactive-step        section-button-demo-step-85ht1f     48%        48%
interactive-step        section-formfill-demo-step-1ghgrv9  54%        54%
interactive-multi-step  __standalone__-step-2i6dkt          60%        60%
interactive-guided      __standalone__-step-ft77cj          66%        66%
quiz-block              section-quiz-section-step-1hw1dkb   0% (lost)  72%
quiz-block              section-quiz-section-step-13zy8lj   0% (lost)  75%
interactive-step        section-quiz-section-step-5oy3cx    78%        78%
interactive-step        section-admin-section-step-16lhsb1  84%        84%
datasource-check-step   blocking-data-check                 93%        93%
----------------------------------------------------------------------------------
step ids the parser assigns: 12
resolved before this change:  10
resolved after this change:   12
keys the runtime never dispatches, before: ["advisory-data-check"]
keys the runtime never dispatches, after:  []


guide: a terminal-only guide (4 blocks, 2 completable)

Reader runs both commands, in order. Percentage after each click:

click                             before     after
----------------------------------------------------------
terminal-step "__standalone__-s…" 0%         50%
terminal-step "__standalone__-s…" 0%         75%
----------------------------------------------------------
Before: every click was lost, so the guide sat at 0% while looking healthy.
```
</details>
<details>
<summary>Evidence: progress.parity.test.ts against the pre-change 3-of-9 resolver (fails 17 of 43)</summary>

<code>● step-id parity with the parser, over the completable set › quiz › derives the id the parser assigns, at the top level&#10;● step-id parity with the parser, over the completable set › quiz › derives the id the parser assigns, inside a section&#10;● step-id parity with the parser, over the completable set › quiz › derives the id the parser assigns, inside an assistant block in a section&#10;● step-id parity with the parser, over the completable set › terminal › derives the id the parser assigns, at the top level&#10;● step-id parity with the parser, over the completable set › terminal › derives the id the parser assigns, inside a section&#10;● step-id parity with the parser, over the completable set › terminal › derives the id the parser assigns, inside an assistant block in a section&#10;● step-id parity with the parser, over the completable set › terminal-connect › derives the id the parser assigns, at the top level&#10;● step-id parity with the parser, over the completable set › terminal-connect › derives the id the parser assigns, inside a section&#10;● step-id parity with the parser, over the completable set › terminal-connect › derives the id the parser assigns, inside an assistant block in a section&#10;● step-id parity with the parser, over the completable set › code-block › derives the id the parser assigns, at the top level&#10;● step-id parity with the parser, over the completable set › code-block › derives the id the parser assigns, inside a section&#10;● step-id parity with the parser, over the completable set › code-block › derives the id the parser assigns, inside an assistant block in a section&#10;● step-id parity with the parser, over the completable set › challenge › derives the id the parser assigns, at the top level&#10;● step-id parity with the parser, over the completable set › challenge › derives the id the parser assigns, inside a section&#10;● step-id parity with the parser, over the completable set › challenge › derives the id the parser assigns, inside an assistant block in a section&#10;● step-id parity on the bundled corpus › block-editor-tutorial resolves every completable block to a parser step id&#10;● step-id parity on the bundled corpus › json-guide-demo resolves every completable block to a parser step id&#10;Test Suites: 1 failed, 1 total&#10;Tests: 17 failed, 26 passed, 43 total</code>

```text
  ● step-id parity with the parser, over the completable set › quiz › derives the id the parser assigns, at the top level
  ● step-id parity with the parser, over the completable set › quiz › derives the id the parser assigns, inside a section
  ● step-id parity with the parser, over the completable set › quiz › derives the id the parser assigns, inside an assistant block in a section
  ● step-id parity with the parser, over the completable set › terminal › derives the id the parser assigns, at the top level
  ● step-id parity with the parser, over the completable set › terminal › derives the id the parser assigns, inside a section
  ● step-id parity with the parser, over the completable set › terminal › derives the id the parser assigns, inside an assistant block in a section
  ● step-id parity with the parser, over the completable set › terminal-connect › derives the id the parser assigns, at the top level
  ● step-id parity with the parser, over the completable set › terminal-connect › derives the id the parser assigns, inside a section
  ● step-id parity with the parser, over the completable set › terminal-connect › derives the id the parser assigns, inside an assistant block in a section
  ● step-id parity with the parser, over the completable set › code-block › derives the id the parser assigns, at the top level
  ● step-id parity with the parser, over the completable set › code-block › derives the id the parser assigns, inside a section
  ● step-id parity with the parser, over the completable set › code-block › derives the id the parser assigns, inside an assistant block in a section
  ● step-id parity with the parser, over the completable set › challenge › derives the id the parser assigns, at the top level
  ● step-id parity with the parser, over the completable set › challenge › derives the id the parser assigns, inside a section
  ● step-id parity with the parser, over the completable set › challenge › derives the id the parser assigns, inside an assistant block in a section
  ● step-id parity on the bundled corpus › block-editor-tutorial resolves every completable block to a parser step id
  ● step-id parity on the bundled corpus › json-guide-demo resolves every completable block to a parser step id
Test Suites: 1 failed, 1 total
Tests:       17 failed, 26 passed, 43 total
```
</details>
<details>
<summary>Evidence: Path rollup: mean of member percentages</summary>

<code>path rollup: mean of member percentages&#10;&#10;a short guide and a long one weigh the same (decision 4)&#10;3-block quickstart 100%&#10;80-block deep dive 0%&#10;=&gt; path reads 50% complete: false&#10;&#10;every member finished — reads 100 and complete&#10;quickstart 100%&#10;deep dive 100%&#10;=&gt; path reads 100% complete: true&#10;&#10;members capped at 99 mean the path rounds to 99, not 100&#10;guide with trailing prose 99%&#10;another 99%&#10;=&gt; path reads 99% complete: false&#10;&#10;an empty path&#10;(no members)&#10;=&gt; path reads 0% complete: false</code>

```text
path rollup: mean of member percentages

a short guide and a long one weigh the same (decision 4)
    3-block quickstart           100%
    80-block deep dive             0%
  => path reads 50%   complete: false

every member finished — reads 100 and complete
    quickstart                   100%
    deep dive                    100%
  => path reads 100%   complete: true

members capped at 99 mean the path rounds to 99, not 100
    guide with trailing prose     99%
    another                       99%
  => path reads 99%   complete: false

an empty path
    (no members)
  => path reads 0%   complete: false
```
</details>
<details>
<summary>Evidence: Parity ratchet fires for a newly tracked completable type</summary>

<code>FAIL src/lib/guide-stats/progress.parity.test.ts&#10;● step-id parity with the parser, over the completable set › covers every block type that can emit completion evidence&#10;&#10;expect(received).toEqual(expected) // deep equality&#10;&#10;- Array []&#10;+ Array [&#10;+ &#34;video&#34;,&#10;+ ]&#10;&#10;135 | const uncovered = [...COMPLETION_AFFORDANCE_BLOCK_TYPES, &#39;input&#39;].filter((type) =&gt; !covered.has(type));&#10;136 |&#10;&gt; 137 | expect(uncovered).toEqual([]);</code>

```text
FAIL src/lib/guide-stats/progress.parity.test.ts
  ● step-id parity with the parser, over the completable set › covers every block type that can emit completion evidence

    expect(received).toEqual(expected) // deep equality

    - Expected  - 1
    + Received  + 3

    - Array []
    + Array [
    +   "video",
    + ]

      135 |     const uncovered = [...COMPLETION_AFFORDANCE_BLOCK_TYPES, 'input'].filter((type) => !covered.has(type));
      136 |
    > 137 |     expect(uncovered).toEqual([]);
          |                       ^
      138 |   });
      139 |
      140 |   it('treats every fixture shape as completable, so the sweep measures what it claims to', () => {
--
Snapshots:   0 total
Time:        0.796 s, estimated 2 s
Ran all test suites matching src/lib/guide-stats/progress.parity.test.ts with tests matching "covers every block type".
```
</details>
<details>
<summary>Evidence: Concern routing parsed by the review skill&#39;s own extractor</summary>

<code>concern: completion-records (subsystem, activation strong, min_signals 1)&#10;&#10;trigger paths, parsed from docs/design/CONCERNS.md by the review skill&#39;s own extractor:&#10;src/completion-records/**&#10;src/lib/guide-stats/**&#10;src/global-state/guide-step-id-resolver.ts&#10;pkg/plugin/completion_records*.go&#10;src/cli/commands/build-stats.ts&#10;&#10;files this change touches:&#10;src/global-state/guide-step-id-resolver.ts -&gt; activates completion-records via src/global-state/guide-step-id-resolver.ts&#10;src/lib/guide-stats/progress.parity.test.ts -&gt; activates completion-records via src/lib/guide-stats/**&#10;src/lib/guide-stats/rollup.ts -&gt; activates completion-records via src/lib/guide-stats/**&#10;src/lib/guide-stats/block-index.ts -&gt; activates completion-records via src/lib/guide-stats/**&#10;&#10;named invariant: completion-denominator-authority&#10;states the mean-of-members rule: true&#10;no longer promises a rollup index: true</code>

```text
concern: completion-records (subsystem, activation strong, min_signals 1)

trigger paths, parsed from docs/design/CONCERNS.md by the review skill's own extractor:
  src/completion-records/**
  src/lib/guide-stats/**
  src/global-state/guide-step-id-resolver.ts
  pkg/plugin/completion_records*.go
  src/cli/commands/build-stats.ts

files this change touches:
  src/global-state/guide-step-id-resolver.ts       -> activates completion-records via src/global-state/guide-step-id-resolver.ts
  src/lib/guide-stats/progress.parity.test.ts      -> activates completion-records via src/lib/guide-stats/**
  src/lib/guide-stats/rollup.ts                    -> activates completion-records via src/lib/guide-stats/**
  src/lib/guide-stats/block-index.ts               -> activates completion-records via src/lib/guide-stats/**

named invariant: completion-denominator-authority
  states the mean-of-members rule:  true
  no longer promises a rollup index: true
```
</details>

## Pipeline

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>
<summary>⚠️ **Review** - 2 infos</summary>

- ⚠️ `src/lib/guide-stats/block-index.ts:247` - positionsByStepId is keyed with pre-inlining sibling indices, but the parser derives props.stepId from post-inlining indices, so any guide containing a snippet-ref gets wrong keys. Concrete trace: authored blocks [markdown, snippet-ref(resolves to 2 markdowns), interactive(anonymous, action=highlight, reftarget=X)]. computeGuideBlockIndex walks the authored tree (its own docstring at block-index.ts:70 mandates the PRE-INLINING tree for the denominator) and calls the resolver with index=2, deriving `__standalone__-step-&lt;hash(…|2|highlight|X|)&gt;`. snippet-engine/inline-refs.ts spliceBlocks() replaces the ref before the parser runs, so json-parser sees the interactive at index 3 and assigns `__standalone__-step-&lt;hash(…|3|highlight|X|)&gt;`. The runtime dispatches the index-3 key; positionsByStepId only holds the index-2 key, so evidencedPosition returns 0 and the guide reads 0% while looking healthy — the exact silent-zero failure this work item exists to prevent. A section following a ref shifts too, rekeying its whole subtree via `section:blocks[n]`. The bundled corpus has no snippet-refs, so progress.parity.test.ts cannot catch this. No consumer exists yet, so nothing regresses today, but the map is wrong for this class of guide the moment a consumer lands. Needs a decision: index the post-inlining tree for step ids, or explicitly document/exclude snippet-ref guides from step-id keying.
- ⚠️ `src/global-state/guide-step-id-resolver.ts:54` - The per-case short-circuit uses `block.id ?? derive(...)` (lines 54, 66, 79, 90, 94, 100, 111, 117), but the parser&#39;s resolveStepId (json-parser.ts:265) uses a truthiness check `if (authorId) return authorId;`. `id` has no min-length in the schema (json-guide.schema.ts:405/420/450/622/640/667/712), so `id: &#34;&#34;` is valid guide JSON. For `{ type: &#39;interactive&#39;, id: &#39;&#39;, action: &#39;highlight&#39;, reftarget: &#39;X&#39; }` the parser derives a real hash and the runtime dispatches under it, while the resolver returns `&#39;&#39;`, which block-index.ts:249 (`if (stepId &amp;&amp; …)`) discards and positionsById also skips (`block.id.length &gt; 0`) — the block becomes wholly unresolvable, position 0. This also regresses the pre-existing AI-fix apply path: ai-fix-step-id.ts:14-18 does `if (!block.id) { const derived = resolveStepIdForBlock(...); if (derived) block.id = derived; }`, so an empty-id block that previously got a materialized id (old code fell through the `if (block.id)` guard and derived) now gets none, and applyPatchToGuide&#39;s selector-patch can no longer address it. Use `block.id ||` in each case (or normalize the empty string once) to match the parser.
- ⚠️ `src/global-state/guide-step-id-resolver.ts:64` - `(block as JsonMultistepBlock).steps[0]` (line 64) and `(block as JsonGuidedBlock).steps[0]` (line 77) dereference `steps` without a guard, while every other field access in the function is optional-chained. computeGuideBlockIndex documents that CountableBlock also accepts &#34;a raw parsed manifest payload whose blocks have not been narrowed&#34; (block-index.ts:30-33), and resolveCountedBlockStepId casts that straight to JsonBlock. A payload containing `{ type: &#39;multistep&#39; }` with no `steps` array turns a previously total, throw-free pure function into a TypeError at block-index.ts:248. Unreachable today (no current caller injects the resolver on unvalidated input), but the fix is `steps?.[0]`.
- ℹ️ `src/lib/guide-stats/progress.parity.test.ts:23` - `import type { ParsedElement } from &#39;../../types/documentation.types&#39;` references a module that does not exist — there is no src/types/documentation.types.ts, and tsconfig&#39;s only path mapping (`&#34;*&#34;: [&#34;../src/*&#34;]`) does not apply to a relative specifier. ParsedElement is exported from src/types/content.types.ts (line 198), which is what json-parser.ts itself imports. Jest erases type-only imports so the suite still runs, which is why this survived a test run, but tsc/build resolves it and fails.
- ℹ️ `src/global-state/guide-step-id-resolver.ts:40` - `StepIdBlockContext` is field-for-field identical to `BlockStepIdContext` (block-index.ts:102), which this file already type-imports via the `BlockIndexOptions` import on line 24. Reusing `BlockStepIdContext` and deleting the local duplicate removes a second declaration of the same contract that has to be kept in sync by hand, at no cost to the module&#39;s dependency-free property (the import is already type-only).

🔧 Fix: fix step-id resolver parity, typing, and snippet-ref note
3 issues (1 warning, 2 infos) still open:
- ⚠️ `docs/design/CONCERN_DETAILS.md:82` - The `completion-denominator-authority` contract anchor still states &#34;`rollUpGuideStats` publishes a path&#39;s or journey&#39;s totals only, so a consumer recording a percentage against a rolled-up `blockCount` still needs its positions from somewhere — the rollup index lands with the first path-level consumer.&#34; `meanOfMemberPercentages` retires that statement: under decision 4 the path-level consumer consumes per-guide percentages and never a rolled-up `blockCount`, so no rollup index arrives. `COMPLETION-MODEL.md:224-233` says so explicitly (&#34;Under this decision that consumer does not arrive... its own docstring&#39;s warning ... should be read as a reason not to, rather than as a gap to fill&#34;). AGENTS.md requires that &#34;when an implementation establishes or replaces one, update the contract anchor in the same PR&#34; — this PR lands the implementation that replaces it and leaves the anchor unamended. The same stale framing sits in `src/lib/guide-stats/summary.ts:68-72`, which still points a would-be consumer at issue #1666 for &#34;the first consumer&#34; of the summed `blockCount`. The intent&#39;s out-of-scope list defers rewriting path progress to use the mean and the `finalCompletablePosition` docstring sweep, but not this anchor statement, so a future reader has two exported path-aggregation models in one directory with the doc naming the retired one as the path denominator.
- ℹ️ `src/lib/guide-stats/rollup.ts:40` - `complete` is decided on the clamped value, so the two flavours of invalid input are handled in opposite directions. Trace `meanOfMemberPercentages([150, 100])`: 150 is finite, clamps to 100, `clamped &lt; 100` is false, so `complete` stays true and the function returns `{percent: 100, complete: true}` — a path reported as finished from a member percentage that could never come from `guideProgressAtPosition`. `meanOfMemberPercentages([Number.POSITIVE_INFINITY, 100])` takes the opposite branch (clamps to 0, `complete: false`). The module&#39;s own docstring justifies the clamp because &#34;the percent reaches a durable record&#34;, and `complete` is the stronger of the two signals to publish from garbage. Deciding the flag on the raw value — `if (value !== 100) { complete = false; }` — is exactly the stated rule (&#34;every member at 100&#34;) and makes both invalid shapes pessimistic. `rollup.test.ts:40` asserts the clamped percent for `[200, 0]` but never the `complete` flag for an over-range member, so the asymmetry is unpinned. No consumer exists yet, so nothing is wrong today.
- ℹ️ `src/lib/guide-stats/progress.test.ts:181` - The test is named &#34;still credits an author id, which the resolver returns unchanged&#34;, but the injected stub at line 174 is `(block, context) =&gt; block.type === &#39;interactive&#39; ? `derived-${context.index}` : undefined` — it returns `derived-2` for the authored block at index 2, never the author id. So `positionsByStepId` holds `{derived-1: 2, derived-2: 3}` and `positionsById` holds `{authored: 3}`; the assertion passes only through the `?? index.positionsById.get(...)` fallback on `progress.ts:104`. The assertion is worth keeping — it is the only direct coverage of that fallback — but the name and the adjacent comment describe author-id passthrough in the real resolver, which is covered elsewhere (the `input` fixture in `progress.parity.test.ts`). Rename to say it falls back to the author id when the resolver keys the block under something else, so a reader does not mistake this for resolver-passthrough coverage.

🔧 Fix: fix rollup completion flag and stale rollup-index docs
- ⚠️ `src/global-state/guide-step-id-resolver.ts:11` - This change adds a fourth site that must be edited when a tracked step type is added, and the registry doc that tells a developer where to edit still says there are three. `.cursor/rules/tracked-step-types.mdc:16` and its heading at :18 state &#34;Adding a new step type requires editing **three sites**&#34;, its checklist ends at &#34;7. **Registry site 3**&#34; (:46) naming only `COMPLETION_AFFORDANCE_BLOCK_TYPES`, and its `globs:` list does not include `src/global-state/guide-step-id-resolver.ts`. `docs/developer/CONTEXT_INDEX.md:25` repeats the &#34;3-site registry&#34; framing. As of this change, adding `&#39;foo&#39;` to `COMPLETION_AFFORDANCE_BLOCK_TYPES` without adding a `case &#39;foo&#39;:` here means `positionsByStepId` holds no key for it, so a foo-driven guide evidences nothing and reads 0% - the exact silent zero this file&#39;s own docstring (lines 11-16) warns about. Two ratchets do catch it (`assertExhaustive` at :143 becomes a type error once the block type is added to the `JsonBlock` union, and `progress.parity.test.ts:133-138` fails until a `COMPLETABLE_SHAPES` fixture exists), so this is guidance drift rather than an unguarded gap - but a developer following the numbered checklist is not routed to either. Flagging rather than fixing because the intent states &#34;Nothing else was to be touched&#34;, and amending a numbered registry contract is the author&#39;s call.
- ℹ️ `src/global-state/guide-step-id-resolver.ts:18` - `docs/design/CONCERNS.md:72` routes the `completion-records` concern at `src/completion-records/**`, `src/lib/guide-stats/**`, `pkg/plugin/completion_records*.go`, `src/cli/commands/build-stats.ts`. The resolver&#39;s new home is not in that list - `src/global-state/**` appears only under `security` and `reversibility-and-one-way-door`. Intent item 4 argues the parity test belongs in `src/lib/guide-stats/` precisely so &#34;a PR weakening the file activates that concern&#39;s reviewer&#34;, but the file whose weakening actually produces the silent zero is this one, and a PR touching only it activates neither the concern nor its contract anchor. Adding `src/global-state/guide-step-id-resolver.ts` to that row&#39;s globs (and optionally `deriveStepId`/`resolveStepIdForBlock` to its signal list) closes the routing gap. Not a regression - the resolver&#39;s previous home in `src/integrations/assistant-integration/` was not routed there either - but it is newly load-bearing for the completion numerator.

🔧 Fix: route step-id resolver to completion-records concern
2 infos still open:
- ℹ️ `src/global-state/guide-step-id-resolver.ts:123` - Residual of the round-1 `unnarrowed-steps-throw` hardening, missed in the same function. `Boolean(b.dataCheckQuery?.trim())` optional-chains the property but not its type, so a raw payload block `{ type: &#39;input&#39;, inputType: &#39;datasource&#39;, dataCheckQuery: 5 }` throws `b.dataCheckQuery.trim is not a function` at block-index.ts:263, turning the previously total pure counter into a crash. This is exactly the shape `computeGuideBlockIndex` documents accepting (block-index.ts:30-33, &#34;a raw parsed manifest payload whose blocks have not been narrowed&#34;) and that `resolveCountedBlockStepId` casts straight to `JsonBlock`. The precedent for distrusting it is in the same traversal: `emitsCompletionEvidence` (completion-affordance.ts:78-80) guards the identical field with `typeof block.dataCheckQuery === &#39;string&#39; &amp;&amp; block.dataCheckQuery.trim().length &gt; 0`. Unreachable today — no production caller injects the resolver — but the fix is to mirror that guard so the two readers of the same field agree on how untrusted it is.
- ℹ️ `src/lib/guide-stats/progress.parity.test.ts:35` - The coverage ratchet the intent singles out as load-bearing (&#34;the parity test must iterate the completable set rather than a sample — this was explicitly required, because a sampled test is exactly what let the 3-of-9 gap survive&#34;) compares a hand-written `shape` label against `COMPLETION_AFFORDANCE_BLOCK_TYPES` (line 134-137), but `shape` is declared independently of `block.type`. A fixture entered as `{ shape: &#39;quiz&#39;, block: { type: &#39;interactive&#39;, action: &#39;highlight&#39;, reftarget: &#39;x&#39;, content: &#39;Do it&#39; } }` satisfies both the coverage assertion (label &#39;quiz&#39; is present) and the completability assertion at line 141 (the block is an interactive, so `emitsCompletionEvidence` is true), while `quiz` is in fact untested and its three namespace cases silently re-test `interactive` twice. Deriving `shape` from `entry.block.type` — with `input` the one named exception, since its tracked form is an authored shape rather than a type — or asserting `entry.block.type === entry.shape` for the other eight, closes the one hole in the ratchet that exists to stop a missed block type.

</details>

<details>

✅ No issues found.
- <code>`npm install` (worktree had no node_modules; package-lock.json restored afterwards)</code>
- <code>`npx jest --coverage=false src/lib/guide-stats src/integrations/assistant-integration/ai-fix-step-id.test.ts src/global-state/step-id.test.ts` — 8 suites / 130 tests pass</code>
- <code>`npx jest --coverage=false src/validation/architecture.test.ts` — 27 pass; confirms the tier-3 to tier-1 resolver move is legal, lib/guide-stats stays dependency-free, ratchet unchanged (vertical=2 lateral=8)</code>
- <code>Regression proof: temporarily swapped the base-commit 3-of-9 resolver into the injected shape and ran `npx jest --coverage=false src/lib/guide-stats/progress.parity.test.ts` — 17 of 43 cases fail (quiz, terminal, terminal-connect, code-block, challenge in all three namespaces, plus block-editor-tutorial and json-guide-demo corpus sweeps); 43/43 pass with the shipped resolver</code>
- <code>Product-level before/after: drove `guideProgress` with the parser&#39;s own `props.stepId` values for every step in `src/bundled-interactives/json-guide-demo/content.json`, old resolver vs new — quiz clicks go from `0% (lost)` to 72%/75%, resolved step ids 10/12 -&gt; 12/12, and the stale `advisory-data-check` author-id key disappears from positionsByStepId</code>
- `Product-level before/after: a synthetic terminal-only guide (4 blocks, 2 terminal steps) read 0% end to end before and reads 50% -> 75% now`
- <code>Ratchet check: temporarily added `video` to `COMPLETION_AFFORDANCE_BLOCK_TYPES` and ran `npx jest src/lib/guide-stats/progress.parity.test.ts -t &#39;covers every block type&#39;` — fails with `uncovered: [&#34;video&#34;]`; restored</code>
- <code>Rollup behaviour transcript for `meanOfMemberPercentages`: [100,0] -&gt; 50% not complete, [100,100] -&gt; 100% complete, [99,99] -&gt; 99% not complete, [] -&gt; 0% not complete</code>
- <code>Concern routing via the review skill&#39;s real extractor: `node .cursor/skills/review/scripts/concern-context.mjs` (`extractConcernContext`) over docs/design/CONCERNS.md + CONCERN_DETAILS.md — confirms `src/global-state/guide-step-id-resolver.ts` and `src/lib/guide-stats/**` both activate `completion-records`, and the `completion-denominator-authority` invariant now states the mean-of-members rule and drops the rollup-index promise</code>
- <code>`npm run test:review-contract` — 76/76 pass</code>
- <code>`npx jest --coverage=false src/validation/import-graph.test.ts src/integrations/coda/coda-api.test.ts` — confirms the claimed pre-existing failures: only the `node:test` builtin case fails; coda-api passes on a fresh install</code>
- <code>Cleanup: removed the throwaway evidence-demo test file and confirmed `git status --porcelain` is empty</code>

</details>

<details>
<summary>⚠️ **Document** - 2 infos</summary>

- ℹ️ `.cursor/rules/systemPatterns.mdc:113` - Pre-existing drift, not caused by this change: the `global-state/` bullet enumerates 4 files (`sidebar.ts`, `panel-mode.ts`, `link-interception.ts`, `alignment-pending-context.ts`) while the directory holds 18, including load-bearing ones like `completion-store.ts`, `step-id.ts`, and now `guide-step-id-resolver.ts`. Left alone deliberately — bringing the list up to date is a standalone maintenance edit, and an exhaustive list will re-rot; a follow-up should either replace it with a one-line characterisation plus a pointer, or drop the enumeration.
- ℹ️ `src/lib/guide-stats/block-index.ts:247` - Judgment call: the known limitation that `positionsByStepId` keys off pre-inlining sibling indices while the parser assigns `props.stepId` post-inlining (so every block after a `snippet-ref` is keyed under an id the runtime never dispatches) is documented only in the code comment at the injection site. That is the right owner today because nothing consumes the map yet, and no bundled or published guide contains a `snippet-ref`. When the first consumer lands, this should be promoted to a line in the `completion-denominator-authority` invariant next to the existing pre-inlining statement, since by then it becomes a reader-visible 0%.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>


moxious (changes requested):
Blockers:

1. [blocking] **B1 — The completable-set coverage ratchet counts hand-written labels, not block types, so a resolver miss on four uncorpused shapes passes the whole suite** (high · testing-and-verification)
   `progress.parity.test.ts` builds its covered set from each fixture's hand-written `shape` label — `new Set(COMPLETABLE_SHAPES.map((entry) => entry.shape))` — and diffs that against `COMPLETION_AFFORDANCE_BLOCK_TYPES` plus `input`. The neighbouring guard reads `entry.block` instead, so the two assertions never meet, and `shape` is typed plain `string` with the table closed by an `as unknown as` cast, so nothing ties a label to its block's type. Reproduced at this head: relabelling only the `challenge` entry's block as an `interactive` leaves the suite at 43/43 with `challenge` exercised by none of the 27 namespace cases; adding a real `case 'challenge': return undefined` silent-zero regression to the resolver on top of that still leaves `src/lib/guide-stats` + `src/integrations/assistant-integration` + `src/global-state` at 362/362, and `tsc --noEmit` clean. Nothing else catches it: the resolver has only two consumers repo-wide, `block-index.test.ts` injects no resolver, and `completion-affordance.parity.test.ts` never calls one. The bundled corpus cannot substitute — a recursive walk of all ten guides finds interactive 112, multistep 18, quiz 4, guided 2, input 2 and zero occurrences of terminal, terminal-connect, code-block or challenge, so for four of the nine shapes this table is the only coverage that exists. The intent makes this ratchet the reason the work item grew ("a sampled test is exactly what let the 3-of-9 gap survive"), and the `completion-denominator-authority` anchor amended in this same PR records that the test "pins it to the parser over the whole completable set rather than a sample" — the guarantee this hole falsifies.
   Required: Derive the covered set from the fixture itself — `new Set(COMPLETABLE_SHAPES.map((entry) => entry.block.type))` — keeping `shape` only as the `describe.each` title, or drop `shape` and title with `'$block.type'`. Verified: with that one line changed, the same mislabel fails loudly with `uncovered: ["challenge"]` (1 failed / 42 passed).

Suggestions & nits:

1. [suggestion] **S1 — The anonymous-section step-id namespace is live but exercised by neither the shape table nor the corpus** (low · testing-and-verification)
   `childSectionId` has two live section branches — `section-${block.id}` and `section:${jsonPath}` — and `id` is optional on `JsonSectionBlock`, so the anonymous form is a valid guide shape. Every parity fixture's section carries `id: 'setup'`, so all 27 namespace cases go through the id-bearing branch only, and the corpus offers 21 sections with none anonymous. `block-index.test.ts` never mentions `resolveStepId` either. Probed directly, the branch is correct today (parser and counter agree on `section:blocks[0]-step-95jfya`, and on nested anonymous sections), so this is a coverage gap rather than a live defect — but it decides the numerator key for every completable inside an anonymous section, and its failure mode is the same silent 0% the file exists to prevent.
   Suggested: Add one case to the existing `describe.each` using a section without an `id`, asserting the same parser-parity equality the id-bearing case already asserts.

2. [nit] **N1 — The tracked-step-type registry routes to the resolver through the checklist, but every count phrase and the in-file back-pointer convention still say three sites** (low · ai-subsystem)
   Checklist routing is sound — step 7 now names `resolveStepIdForBlock` and the intro adds a forward pointer to it — so a developer reading top to bottom does reach the new file. But `tracked-step-types.mdc` still says "Three-site registry" in its frontmatter description, "requires editing **three sites**" in the intro, "## The three registry sites" as a heading and "Three sites, not one." in the body, with a three-row table and no row for the resolver; and `CONTEXT_INDEX.md` still reads "Lists the 3-site registry (STEP_TYPE_SCHEMAS + INTERACTIVE_STEP_COMPONENT_TYPES + COMPLETION_AFFORDANCE_BLOCK_TYPES)" beside a glob column this same PR extended to five files including the resolver. The repository's convention is also an in-file back-pointer — `completion-affordance.ts` carries `⚠ TRACKED STEP TYPE REGISTRY — site 3 of 3` — and the new resolver carries none, so a developer who opens it directly gets no route back to the rule.
   Suggested: Pick one framing and apply it consistently: either promote the resolver to a fourth table row and update the frontmatter description, the intro count, the heading, the "Three sites, not one." line and `CONTEXT_INDEX.md`, or keep it an explicit satellite of site 3 and say so in that row. Either way add the `⚠ TRACKED STEP TYPE REGISTRY` back-pointer to `guide-step-id-resolver.ts` to match `completion-affordance.ts`.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1842
Summary: key completion positions by runtime step id
Verdict: Request Changes
Results: 1 blocker, 2 non-blocking findings, 0 follow-ups

moxious (approved):

Suggestions & nits:

1. [suggestion] **S2 — The sweep's conditional tail is hand-written, so the checklist's promise that the ratchet demands a fixture does not hold for a conditionally-completable type** (low · testing-and-verification)
   `progress.parity.test.ts:138` builds the expected set as `[...COMPLETION_AFFORDANCE_BLOCK_TYPES, 'input']` — the always-completable half derived from the exported constant, the conditional half a single hand-written literal. `completion-affordance.ts` deliberately keeps `input` out of both lists ("it is the one conditional case") and hard-codes its blocking-datasource predicate in `emitsCompletionEvidence`; I confirmed that is the only conditional branch at head. A second type made conditionally completable the same way would appear in neither the constant nor the test's literal tail, so `uncovered` would stay `[]`, no fixture would ever be demanded, and a missing resolver case would produce the same silent 0% that B1 was about — one step further out. The rule actively invites that shape: checklist step 7 says "If only some authored shapes do, branch in `emitsCompletionEvidence` the way `input` does." Step 8 then claims "`progress.parity.test.ts` iterates the completable set, so it fails until both the resolver case and a fixture exist", and the amended `completion-denominator-authority` anchor says the test "pins it to the parser over the whole completable set rather than a sample". Both are true for the eight unconditional types — I proved it by falsification — and neither is enforced on the step-7 path. Nothing is broken today: `input` is the only conditional type and it is covered by the fixture and by the resolver at `guide-step-id-resolver.ts:122-131`, so the sweep is genuinely whole at this head. The doc over-claim is the sharper half, because an engineer reading step 8 will believe the ratchet is watching them when on that path it is not.
   Suggested: Cheapest honest fix, in files this PR already edits: narrow step 8 in `.cursor/rules/tracked-step-types.mdc` to say the sweep demands a fixture only for a type listed in `COMPLETION_AFFORDANCE_BLOCK_TYPES`, note in step 7 that a conditionally-completable type must also be added to the parity test's tail by hand, and align the anchor sentence. The stronger option removes the hand-maintenance instead of documenting it: derive the tail structurally — either export a conditional set that `emitsCompletionEvidence` and the test both read, or probe `emitsCompletionEvidence` across the exhaustive `JsonBlock['type']` union so any type that can return true for some shape is demanded.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1842
Summary: key completion positions by runtime step id
Verdict: Approve with Minor

PR #1843, opened 2026-09-09
fix(learning-paths): join path members to their saved percentage by content key

tomglenn:
## Summary

A path's percentage is the mean of its milestones' percentages, so each member has to be joined to the record it persisted under. That key is stored nowhere: a member is keyed by the sanitized URL it was launched from, while a path definition carries only ids. This adds that join as a pure module. Nothing consumes it yet — the path rollup builds on top of it.

Two things a reviewer should see explicitly:

- **A member whose record cannot be found is excluded from the mean and counted, never scored zero.** A zero is indistinguishable from a real result and would silently drag a path's number down, so the result carries `unresolvedCount` and `unresolvedMemberIds` instead. The exclusion is visible rather than silent.
- **Both bundled key forms are read.** A guide is launched as `bundled:<id>` from My Learning and as `bundled:<id>/content.json` from the context panel package surface, and each keeps its own progress. Reading only the bare form made a member progressed from the context panel resolve as a real zero. Within a scheme the furthest record wins; across schemes the composite resolver's bundled-first precedence is honoured, so one guide's progress is never reported as another's.

The percentage comes from `interactiveCompletionStorage` and is read by key presence, because that storage's `get` returns 0 for a missing key — which would collapse the very distinction the module exists to keep.

Two limits are recorded in the doc rather than papered over: a supplied member URL is trusted verbatim (this module is pure and cannot consult the bundled repository), and `interactiveCompletionStorage` evicts by insertion order at its 100-entry cap, so an evicted record is indistinguishable from never-opened at this boundary. Both name their follow-on fix.

## What changed

- Adds `src/global-state/path-member-join.ts`, a pure tier-1 module (`pathMemberContentKeys`, `pathMemberIdSchemeKeys`, `resolvePathMemberPercentage`, `resolvePathMemberPercentages`) that builds the content keys a path member may have persisted under — a resolved launch URL sanitized plus its sibling `bundled:` package shape, otherwise both the `bundled:` and `backend-guide:` id schemes in composite-resolver precedence — and reads the caller-supplied `interactiveCompletionStorage` record by key presence (`Object.hasOwn`), taking the furthest value in `[0, 100]` within a scheme. Members with no formable key or an unreadable value are excluded from the result's `resolvedPercentages` and surfaced via `unresolvedCount`/`unresolvedMemberIds` rather than scored zero. `sanitizeContentKey` is now exported from `content-key.ts` so the join reduces member URLs through the launch-side sanitizer. Nothing consumes the module yet.
- `resetPath` in `learning-paths.hook.ts` now derives its key lists from the join instead of restating the scheme pair, splitting them by key space: `pathMemberContentKeys` for the sanitized namespaces (`interactiveStepStorage`, `interactiveCompletionStorage`) and `pathMemberIdSchemeKeys` for the raw-launch-URL namespaces (`milestoneCompletionStorage`, `journeyCompletionStorage`). This widens reset coverage to the `bundled:<id>/content.json` package-launch shape, which was previously left behind.
- Records the rationale as decision 9 in `docs/design/COMPLETION-MODEL.md` (with the open follow-ons: bundled-first URL resolution in the caller, and insertion-order eviction in `interactiveCompletionStorage`), updates `CONTEXT_INDEX.md` and the learning-paths README, corrects the `BoundedRecordStorage.cleanup` doc comment to say insertion order rather than recency, and adds a unit suite plus a storage round-trip suite that drives the real `getContentKey` for each launch shape and asserts the persisted key is one the join looks up.

## Risk assessment

✅ Low: The join is a pure, well-tested tier-1 module with no consumer yet, and its only production-reachable effect — resetPath now clearing a third key shape — is a strict superset of the prior behavior; the surviving findings are a documentation claim that overstates what the scheme precedence protects and one missing regression key, neither of which changes a computed value.

## Testing

<code>I exercised the join the way a reader reaches it: real step completions written through the production completion store into real localStorage, then resolved back from a path definition that carries only ids. That round-trip is now a committed test file (5 cases) alongside a new reset-path case covering the `bundled:&lt;id&gt;/content.json` key the join added; both were proven meaningful by reverting the corresponding production code and watching them fail. The transcript artifact shows the persisted record and the resolved per-member percentages, including a member correctly reported at the furthest of its two launch surfaces and two members excluded-and-counted rather than silently scored zero. Existing join, content-key, completion-store, learning-paths hook, round-trip and architecture suites all pass, as does a full typecheck. There is no screenshot because the change is a tier-1 pure module with deliberately no UI consumer yet — the reviewer-visible surface is the persisted record and join output, which the transcript captures directly. The only red test locally is a Node-22 `import-graph` case, which is environmental and untouched by this diff.</code>

<details>
<summary>Evidence: Reader progress → persisted record → resolved percentage (real storage transcript)</summary>

<code>=== Path member content-key join: reader progress -&gt; persisted record -&gt; resolved percentage ===&#10;&#10;-- 1. A reader makes real progress (real completion-store, real user-storage) --&#10;reader opens &#34;bundled:first-dashboard&#34; from My Learning, completes 1 of 4 steps&#10;reader opens &#34;bundled:first-dashboard/content.json&#34; from the context panel package surface, completes 3 of 4 steps&#10;reader opens &#34;backend-guide:fe-alerting-01&#34; from an App Platform path, completes 2 of 4 steps&#10;&#10;-- 2. What is actually persisted (localStorage, interactive completion namespace) --&#10;grafana-pathfinder-app-interactive-completion = {&#34;bundled:first-dashboard&#34;:25,&#34;bundled:first-dashboard/content.json&#34;:75,&#34;backend-guide:fe-alerting-01&#34;:50}&#10;journey completion namespace = null&#10;(decision 4: the journey namespace holds nothing for these members, so the join reads only the interactive one)&#10;journeyCompletionStorage.getAll() -&gt; {}&#10;&#10;-- 3. The path definition the rollup will start from (ids only, no keys) --&#10;members: first-dashboard, fe-alerting-01, fe-alerting-02, welcome-to-grafana&#10;completed set: welcome-to-grafana&#10;keys looked up for &#34;first-dashboard&#34;: bundled:first-dashboard, bundled:first-dashboard/content.json, backend-guide:first-dashboard&#10;keys looked up for &#34;fe-alerting-01&#34;: bundled:fe-alerting-01, bundled:fe-alerting-01/content.json, backend-guide:fe-alerting-01&#10;keys looked up for &#34;fe-alerting-02&#34;: bundled:fe-alerting-02, bundled:fe-alerting-02/content.json, backend-guide:fe-alerting-02&#10;keys looked up for &#34;welcome-to-grafana&#34;: bundled:welcome-to-grafana, bundled:welcome-to-grafana/content.json, backend-guide:welcome-to-grafana&#10;&#10;-- 4. Join result --&#10;first-dashboard 75% source=persisted via &#34;bundled:first-dashboard/content.json&#34;&#10;fe-alerting-01 50% source=persisted via &#34;backend-guide:fe-alerting-01&#34;&#10;fe-alerting-02 0% source=unopened&#10;welcome-to-grafana 100% source=completed&#10;path percentage (mean of the scored members) = 56%&#10;unresolvedCount = 0 unresolvedMemberIds = []&#10;note: first-dashboard reports 75% — the furthest of its two launch surfaces, not the 25% one.&#10;&#10;-- 5. The exclusion case: a URL-based path whose milestone URLs never resolved --&#10;collect-logs EXCLUDED source=unresolved&#10;define-rules EXCLUDED source=unresolved&#10;path percentage = n/a (no member could be scored)&#10;unresolvedCount = 2 unresolvedMemberIds = [collect-logs, define-rules]&#10;note: excluded and counted — never scored 0%, which would silently drag the path number down.</code>

```text
=== Path member content-key join: reader progress -> persisted record -> resolved percentage ===

-- 1. A reader makes real progress (real completion-store, real user-storage) --
  reader opens "bundled:first-dashboard" from My Learning, completes 1 of 4 steps
  reader opens "bundled:first-dashboard/content.json" from the context panel package surface, completes 3 of 4 steps
  reader opens "backend-guide:fe-alerting-01" from an App Platform path, completes 2 of 4 steps

-- 2. What is actually persisted (localStorage, interactive completion namespace) --
  grafana-pathfinder-app-interactive-completion = {"bundled:first-dashboard":25,"bundled:first-dashboard/content.json":75,"backend-guide:fe-alerting-01":50}
  journey completion namespace = null
  (decision 4: the journey namespace holds nothing for these members, so the join reads only the interactive one)
  journeyCompletionStorage.getAll() -> {}

-- 3. The path definition the rollup will start from (ids only, no keys) --
  members: first-dashboard, fe-alerting-01, fe-alerting-02, welcome-to-grafana
  completed set: welcome-to-grafana
  keys looked up for "first-dashboard": bundled:first-dashboard, bundled:first-dashboard/content.json, backend-guide:first-dashboard
  keys looked up for "fe-alerting-01": bundled:fe-alerting-01, bundled:fe-alerting-01/content.json, backend-guide:fe-alerting-01
  keys looked up for "fe-alerting-02": bundled:fe-alerting-02, bundled:fe-alerting-02/content.json, backend-guide:fe-alerting-02
  keys looked up for "welcome-to-grafana": bundled:welcome-to-grafana, bundled:welcome-to-grafana/content.json, backend-guide:welcome-to-grafana

-- 4. Join result --
  first-dashboard      75%        source=persisted  via "bundled:first-dashboard/content.json"
  fe-alerting-01       50%        source=persisted  via "backend-guide:fe-alerting-01"
  fe-alerting-02       0%         source=unopened
  welcome-to-grafana   100%       source=completed
  path percentage (mean of the scored members) = 56%
  unresolvedCount = 0  unresolvedMemberIds = []
  note: first-dashboard reports 75% — the furthest of its two launch surfaces, not the 25% one.

-- 5. The exclusion case: a URL-based path whose milestone URLs never resolved --
  collect-logs         EXCLUDED   source=unresolved
  define-rules         EXCLUDED   source=unresolved
  path percentage = n/a (no member could be scored)
  unresolvedCount = 2  unresolvedMemberIds = [collect-logs, define-rules]
  note: excluded and counted — never scored 0%, which would silently drag the path number down.
```
</details>
<details>
<summary>Evidence: Negative controls proving the new tests are real regressions</summary>

```text
# backend-guide scheme removed from idSchemeKeyGroups
$ npx jest --coverage=false src/global-state/path-member-join.storage-round-trip.test.ts
Tests: 2 failed, 3 passed, 5 total

# learning-paths.hook.ts restored from base commit 677a6aa
$ npx jest --coverage=false src/learning-paths/reset-path-completion.test.ts
● resetPath — App Platform path (no url) › clears the package-launch shape of a member, which the join also reads
Tests: 1 failed, 7 passed, 8 total

# both reverts undone, production code back to target commit
$ npx jest --coverage=false src/global-state/path-member-join src/global-state/content-key.test.ts src/global-state/completion-store.test.tsx src/learning-paths/reset-path-completion.test.ts src/learning-paths/learning-paths.hook.test.ts src/learning-paths/learning-paths.completion-round-trip.test.ts src/validation/architecture.test.ts src/validation/control-bytes.test.ts src/validation/unicode-format-characters.test.ts
Test Suites: 10 passed, 10 total
Tests: 175 passed, 175 total
```
</details>
- Outcome: ⚠️ 2 infos across 1 run (5m50s)

## Pipeline

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>

<details>
<summary>⚠️ **Review** - 3 infos</summary>

- ⚠️ `src/global-state/path-member-join.ts:111` - The candidate key list for a member with no resolved URL covers only `bundled:&lt;id&gt;` and `backend-guide:&lt;id&gt;`, but production also persists interactive completion under the package-format key `bundled:&lt;id&gt;/content.json`. Concrete path: `BundledPackageResolver.resolve` returns `contentUrl = &#39;bundled:&#39; + entry.path + &#39;content.json&#39;` (src/package-engine/resolver.ts:55); `createCompositeResolver` puts the bundled resolver first (src/package-engine/composite-resolver.ts:115); `resolvePackageNavLinks` carries that URL into the context panel, which calls `openDocsPage(link.contentUrl, ...)` (src/components/docs-panel/context-panel.tsx:637); `openDocsPage` sets `currentUrl: url` verbatim (src/components/docs-panel/docs-panel.tsx:800), so `getContentKey()` returns `bundled:first-dashboard/content.json` and `completion-store.refreshGuidePercentage` writes the percentage under that key. `first-dashboard` (and every other member of the `getting-started` path in paths.json) is a bundled package id in src/bundled-interactives/repository.json, and My Learning launches the same guide as bare `bundled:first-dashboard` (MyLearningTab.tsx:271). So a reader who reaches 50% on `first-dashboard` from the context panel joins to nothing: `resolvePathMemberPercentage` returns `{ percent: 0, source: &#39;unopened&#39; }`, which enters `resolvedPercentages` as a genuine zero rather than being excluded and counted — the precise failure decision 9 says must never occur (&#39;a zero silently drags a path&#39;s number down and looks exactly like a real result&#39;). The earliest shared boundary is this key list: add the package form as a third candidate (and consider that `member.url`, when it is already a `bundled:&lt;id&gt;/content.json` URL, needs the bare form as a sibling candidate for the reverse case).
- ℹ️ `src/learning-paths/learning-paths.hook.ts:442` - The intent states this swap is &#39;provably identical output&#39;, but `pathMemberContentKeys` runs each key through `sanitizeContentKey` (strips `..`, truncates to 200 chars) whereas the removed literals did not. The four namespaces resetPath clears are split: `interactiveCompletionStorage` and `interactiveStepStorage` are keyed by the sanitized `getContentKey()` output (so sanitizing here is a net improvement), while `milestoneCompletionStorage` and `journeyCompletionStorage` are keyed by the raw `journeyBaseUrl` (src/docs-retrieval/learning-journey-helpers.ts:442, 651). For any id where sanitization changes the string, resetPath now stops clearing the milestone/journey records it previously cleared. Reachability is narrow (App Platform ids are RFC1123 CR names, so `..` cannot appear; only a &gt;186-char id trips the 200-char truncation), but the claim of identical output is not exact and the mixed key space is worth an explicit decision rather than an implicit one.
- ℹ️ `src/global-state/path-member-join.ts:131` - `persistedPercentages` is typed `Record&lt;string, number&gt;`, but its production source is `interactiveCompletionStorage.getAll()`, which is an unchecked cast of parsed persisted JSON (`storage.getItem&lt;Record&lt;string, number&gt;&gt;` in src/lib/storage/bounded-record-storage.ts:110, backed by a raw `JSON.parse` in createLocalStorage). A key present with a non-numeric value passes `Object.hasOwn`, is returned as `percent`, survives the `percent !== undefined` filter in `resolvePathMemberPercentages`, and lands in `resolvedPercentages` — poisoning the very mean this module exists to protect. A `typeof value === &#39;number&#39; &amp;&amp; Number.isFinite(value)` guard at the read (treating a non-numeric record as absent) closes it cheaply. Related nit at line 148: `unresolvedMemberIds` is derived from `percent === undefined` rather than `source === &#39;unresolved&#39;`, so the two derived fields can disagree if a percent ever arrives undefined from a present key.
- ℹ️ `src/global-state/path-member-join.test.ts:27` - The &#39;launch round-trip&#39; suite is described as evidence that &#39;the persisted key is one the join looks up&#39;, but `keyPersistedByLaunch` sets `activeTabUrl` to a string literal that the test itself also assigns to `member.url` (or reconstructs as `bundled:&lt;id&gt;`). It therefore proves `sanitize(x) === sanitize(x)` — real value against a one-sided sanitizer change, but none against the launch surface producing a different URL shape for the same member. That is exactly the gap the package-format finding above exercises: no test here fails when `openDocsPage` is handed `bundled:first-dashboard/content.json`. Driving one case through an actual launch surface (or asserting against `BundledPackageResolver.resolve(id).contentUrl`) would make the round-trip claim hold.

🔧 Fix: fix path member join key coverage and value guards
3 issues (1 warning, 2 infos) still open:
- ⚠️ `src/global-state/path-member-join.ts:158` - The fix round widened the candidate set to cover both live bundled launch shapes, but left the lookup first-hit-wins on a fixed order (`bundled:&lt;id&gt;`, then `bundled:&lt;id&gt;/content.json`, then `backend-guide:&lt;id&gt;`, line 116-122). That order was justified by the premise the module and doc still assert — &#34;A reader can only have progressed under one of them&#34; (path-member-join.ts:19-20, COMPLETION-MODEL.md:313) — and the widening makes that premise false for the two bundled shapes, because both are independent, simultaneously reachable launch surfaces for the same guide, and each keeps its own step progress.

Concrete trace, using `first-dashboard` (a member of the static `getting-started` path in paths.json, and a bundled package in repository.json with `targeting.match` urlPrefix `/`):
1. My Learning: `resolveGuideMetadata(&#39;first-dashboard&#39;,&#39;getting-started&#39;)` returns the static entry, which carries no `url`, so MyLearningTab.tsx:271 falls through to `bundled:first-dashboard`. `openDocsPage` sets `currentUrl: url` (docs-panel.tsx:795), and `fetchBundledInteractive` echoes the requested URL back as `rawContent.url` (content-fetcher/bundled.ts:151), so `buildDocsLoadSuccessPatch`&#39;s `currentUrl: fetchedContent.url` keeps it bare. Steps persist under content key `bundled:first-dashboard`; say 30.
2. Context panel: the same guide is recommended, and `openDocsPage(link.contentUrl, ...)` (context-panel.tsx:637) is handed `bundled:first-dashboard/content.json` from `BundledPackageResolver.resolve`. `interactiveStepStorage` is keyed `INTERACTIVE_STEPS_PREFIX + contentKey + &#39;-&#39; + sectionId` (user-storage.ts:972), so this launch starts from zero and progresses independently; say 90.
3. `resolvePathMemberPercentage({id:&#39;first-dashboard&#39;}, ...)` hits `bundled:first-dashboard` first and returns 30. The reader&#39;s furthest progress on that member is 90, so the path mean is under-reported with no error — a wrong value of exactly the kind decision 9 exists to prevent, just arriving as an understatement rather than a zero.

Earliest shared boundary is this loop: pick across the candidate keys deliberately (the maximum finite value is the honest read of &#34;how far has this reader got on this guide&#34;) instead of taking the first key that happens to hold anything, and correct the &#34;only progressed under one&#34; claim in both path-member-join.ts:19 and COMPLETION-MODEL.md:313, since it no longer holds within the bundled scheme. The same ordering question applies to the sibling pairing in the resolved-URL branch (line 140).
- ℹ️ `src/global-state/path-member-join.ts:119` - The package-form candidate is built as `bundled:&lt;memberId&gt;/content.json`, but the key a package launch actually persists under is `bundled:${entry.path}content.json` (resolver.ts:55), derived from the repository entry&#39;s `path`, not from its id. Every one of the 10 entries in src/bundled-interactives/repository.json currently has `path === &#39;&lt;id&gt;/&#39;`, so the two agree today; the moment an entry is added with a nested or renamed path, the join silently reverts to the pre-fix behavior for that member (resolves as `unopened`, enters the mean as a real zero) and the same divergence also makes `resetPath`&#39;s `journeyCompletionStorage.clearMany(rawSchemeKeys)` miss that member&#39;s journey record.

The new round-trip tests only drive `createBundledResolver().resolve(&#39;first-dashboard&#39;)` (path-member-join.test.ts:75-101), so the coupling is unguarded for every other entry. Loop the round-trip assertion over all repository ids (or derive the package form from the resolver&#39;s `contentUrl`) so a divergent `path` fails here rather than in production.
- ℹ️ `src/global-state/path-member-join.ts:163` - The finite-number guard skips a present-but-unreadable value and falls through to `return { percent: 0, source: &#39;unopened&#39; }` (line 169). A key that is present was written by a launch, so the member was opened; scoring it 0 and letting it into `resolvedPercentages` is the &#34;zero that looks exactly like a real result&#34; the module exists to prevent (line 21-25). The candidate-formed-but-unreadable case is distinct from both `unopened` (formed, genuinely absent, 0 is correct) and `unresolved` (no key formable), and excluding-and-counting it would match the stated rule.

The round&#39;s own test makes the mismatch visible: `keeps a non-finite record out of the mean` (path-member-join.test.ts:203-210) asserts `resolvedPercentages` is `[0]` and `source` is `&#39;unopened&#39;` — the value is in the mean, and the test name claims the opposite. Reachability is narrow (the record is unchecked `JSON.parse` output, so only hand-edited or corrupted localStorage produces it), so this is a decision to make explicitly rather than a defect to rush: either exclude-and-count it, or rename the test and say in the doc that a corrupted record is deliberately scored zero.

🔧 Fix: take furthest record and exclude unreadable ones
3 infos still open:
- ℹ️ `src/global-state/path-member-join.ts:196` - `&#39;unopened&#39;` returns `percent: 0` into the mean, and the module (line 29-30) and Decision 9 both assert &#34;A key that is genuinely absent is a different thing: the member was never opened, and zero is the honest answer.&#34; That claim does not hold under the storage&#39;s own cap. `interactiveCompletionStorage` is a bounded record with `limit: LIMITS.MAX_INTERACTIVE_COMPLETIONS = 100` (src/lib/user-storage.ts:164), and `writeWithCap` drops entries with `entries.slice(-limit)` (src/lib/storage/bounded-record-storage.ts:39) — insertion order, so the earliest-inserted keys are evicted, and updating an existing key does not move it forward. Concrete path: a reader who has opened 100+ distinct guides over time progresses `first-dashboard` to 60, then opens enough new guides that the record overflows; `first-dashboard`&#39;s key was inserted early, so it is the one dropped. The join then forms candidates, finds none present, and returns `{ percent: 0, source: &#39;unopened&#39; }`, which enters `resolvedPercentages` as a genuine zero — the exact outcome decision 9 exists to prevent, arriving as a silent understatement. The information needed to tell the two apart is gone by the time the join runs, so this is not fixable at this boundary: the remedy is storage-side (raise the cap, evict least-recently-updated rather than earliest-inserted, or persist an opened-guides set the join can consult). Worth noting that the new dual-shape candidate set also means a bundled guide opened from both surfaces now consumes two of the 100 slots, so cap pressure is slightly higher than before. No action needed in this change; the doc claim is what should be softened if the cap is accepted as-is.
- ℹ️ `src/global-state/path-member-join.ts:181` - The guard rejects a non-number and a non-finite number but accepts any finite value, including one outside `[0, 100]`. `BoundedRecordStorage.set` clamps on write (`Math.max(0, Math.min(100, percentage))`, src/lib/storage/bounded-record-storage.ts:49), so an out-of-range value can only come from the same source the guard was added for — the unchecked `JSON.parse` output of a corrupted or hand-edited record. In that case the value passes, and because the round-2 change made &#34;furthest wins&#34;, an inflated value now *always* wins its candidate set: `{ &#39;bundled:guide-a&#39;: 40, &#39;bundled:guide-a/content.json&#39;: 500 }` yields `percent: 500, source: &#39;persisted&#39;`, which enters the mean and drags the path&#39;s number up with no error. Extending the existing predicate to `persisted &gt;= 0 &amp;&amp; persisted &lt;= 100` (treating out-of-range as `&#39;unreadable&#39;`, or clamping to match the writer) closes it in the same line the guard already occupies. Marginal reachability, mechanical fix.
- ℹ️ `src/global-state/path-member-join.ts:20` - Noting for traceability, not as a defect. The authoritative intent&#39;s decision 1 reads &#34;Read under both key schemes and take whichever holds a record, since a reader can only have progressed under one.&#34; The current code takes the maximum finite value across candidates instead (line 185-188), which contradicts that wording. This was explicitly directed by the round-2 instruction (&#34;Take the maximum finite value across candidate keys rather than first-hit-wins&#34;), the premise is now correctly retracted in both the module docstring (line 18-21) and COMPLETION-MODEL.md decision 9 (&#34;The furthest record wins, not the first one found&#34;), and the retraction is substantiated by source: `bundled:&lt;id&gt;` and `bundled:&lt;id&gt;/content.json` are independently reachable launch URLs (MyLearningTab.tsx:271 vs context-panel&#39;s `link.contentUrl` from resolver.ts:55) and `interactiveStepStorage` keys step progress by content key (user-storage.ts:972), so both can hold a record for one guide. No action — the divergence from the original intent text is the user&#39;s own decision and is documented.

🔧 Fix: bound percentage guard to 0-100 and document eviction
2 infos still open:
- ℹ️ `src/global-state/path-member-join.ts:204` - Furthest-wins is applied across all three candidates, including across the `bundled:` and `backend-guide:` schemes, not just across the two bundled shapes the doc justifies it with. The justification in decision 9 is that a guide &#34;is bundled or it is App Platform, not both&#34;, but `createCompositeResolver` states the opposite as a live possibility and encodes a precedence for it: &#34;Bundled/CDN win any ID collision so today&#39;s fallback behavior is preserved; private guide IDs are expected to carry an `fe-`-style prefix by convention to make collisions vanishingly unlikely&#34; (src/package-engine/composite-resolver.ts:110-112). Nothing enforces that convention — App Platform ids are CR-authored.

Concrete state: a namespace publishes a private guide with id `first-dashboard`, colliding with the bundled package of that name. The reader progresses the private guide to 90 (key `backend-guide:first-dashboard`) and the bundled `getting-started` member to 30 (key `bundled:first-dashboard`). `resolvePathMemberPercentage({id:&#39;first-dashboard&#39;}, ...)` has no url (static path member), forms all three candidates, and returns 90 with `contentKey: &#39;backend-guide:first-dashboard&#39;` — progress from a different guide, reported as this member&#39;s, with no error. First-hit-wins previously matched the resolver&#39;s bundled-first precedence here; the round-2 change inverted it.

The bundled-shape argument does not extend across schemes: two shapes of the same guide are the same guide, two schemes with the same id may not be. Narrowing max to within a scheme and keeping bundled-before-backend-guide precedence across schemes would preserve the round-2 fix and restore the resolver&#39;s own collision rule. Reachability is low (requires a colliding CR id), so this is a decision to make explicitly rather than a defect to rush — but decision 9&#39;s premise sentence currently asserts something the resolver&#39;s comment contradicts.
- ℹ️ `src/global-state/path-member-join.ts:150` - `pathMemberIdSchemeKeys` enumerates the launch shapes of two of the four resolver tiers. `createCompositeResolver` also wires `OnlineCdnPackageResolver` (contentUrl `buildPackageFileUrl(baseUrl, entry.path, &#39;content.json&#39;)` — an https URL, online-cdn-resolver.ts:57) and `RecommenderPackageResolver` (contentUrl straight from the recommender response, recommender-resolver.ts:78). A member id resolved through either tier and opened from the context panel persists under that https key, which none of the three candidates match, so the join returns `{ percent: 0, source: &#39;unopened&#39; }` — a real zero in the mean rather than an exclusion.

I checked reachability and it is closed today: every no-url path member in both static catalogues is a bundled repository id (`getting-started` → welcome-to-grafana, prometheus-grafana-101, first-dashboard; `observability-basics` → prometheus-grafana-101, loki-grafana-101, prometheus-advanced-queries; `cloud-getting-started` → first-dashboard-cloud), the bundled resolver is first in the composite so it wins those ids, and App Platform members always carry `backend-guide:&lt;id&gt;` because `fetchAppPlatformLearningPaths` builds `paths` and `guideMetadata` from the same fetch (app-platform-paths.ts:52-79) — a path whose catalogue has not loaded is not rendered, so the cold-launch-as-bundled case cannot occur.

So: no defect now, but the module docstring presents its scheme list as &#34;the whole of it&#34; while the invariant that makes it complete lives in the static catalogues, unguarded. Adding one non-bundled member id to paths.json would silently understate that path. The cheapest guard is an assertion in the round-trip suite that every static path member id resolves from the bundled repository; noting it here rather than proposing it as work for this change.

🔧 Fix: honour scheme precedence in path member join
2 infos still open:
- ℹ️ `src/global-state/path-member-join.ts:181` - The cross-scheme precedence rule this round introduced is enforced only on the no-url branch. `pathMemberContentKeyGroups` returns a single group for any member carrying a `url` (line 181-183), so whatever scheme that URL is in answers unconditionally — and the URL&#39;s own precedence is decided upstream by `resolveGuideMetadata`, whose order is the inverse of `createCompositeResolver`&#39;s.

Concrete reachable state, using the same collision premise decision 9 is built on:
1. A namespace publishes a private guide with id `first-dashboard`. `fetchAppPlatformLearningPaths` puts every published guide into a flat `guideMetadata` map with `url: &#39;backend-guide:first-dashboard&#39;` (app-platform-paths.ts:52-57) — the map covers all published guides, not just members of App Platform paths.
2. The reader opens the static `getting-started` path (paths.json), whose member `first-dashboard` is a bundled repository id. `resolveGuideMetadata(&#39;first-dashboard&#39;, &#39;getting-started&#39;)` finds no scoped dynamic entry (static path), then hits `if (appPlatformData.guideMetadata[guideId]) return ...` at learning-paths.hook.ts:204 — BEFORE the static `getPathsData()` fallback at line 208. It returns the App Platform entry.
3. `getPathGuides` propagates that as `url: metadata.url` (learning-paths.hook.ts:340) — precisely the `PathMember` shape this module consumes.
4. `resolvePathMemberPercentage` takes the `member.url` branch, forms the single group `[&#39;backend-guide:first-dashboard&#39;]` (the bundled sibling pairing at line 166-173 only fires for `bundled:` URLs), and returns the private guide&#39;s percentage as the bundled member&#39;s, with no error.

That is the exact inversion the module docstring at line 22-27 and decision 9 say the join prevents (&#34;Taking a maximum across schemes would report a private App Platform guide&#39;s progress as a bundled member&#39;s&#34;) — it just arrives through the resolved-URL branch instead of the max. Latent today because the module has no consumer, and it needs a colliding CR id; but the eventual caller is `getPathGuides`, so the path closes on its own once the rollup lands.

This module is pure and cannot consult the bundled repository, so the honest options are (a) scope the precedence claim in path-member-join.ts:22-27 and COMPLETION-MODEL.md to the id-scheme branch and state that a supplied `url` is trusted verbatim, or (b) make the eventual caller resolve member URLs bundled-first so the join is never handed an App Platform URL for a static path member. Recording which one, rather than leaving the unqualified claim standing, is the ask.
- ℹ️ `src/global-state/path-member-join.ts:189` - This round moved the join&#39;s own read path off `pathMemberIdSchemeKeys`. `pathMemberContentKeyGroups` now restates the scheme construction inline (line 187-194: `bundled:&lt;id&gt;`, `bundled:&lt;id&gt;/content.json`, `backend-guide:&lt;id&gt;`), while `pathMemberIdSchemeKeys` (line 157-163) remains exported and is now used only by `resetPath` for the raw-keyed namespaces (learning-paths.hook.ts:443, 446). The two lists are byte-identical today modulo `sanitizeContentKey`, so there is no wrong value now — this is a drift hazard, and one the module&#39;s own documentation denies exists: the docstring at line 13-15 and COMPLETION-MODEL.md (&#34;`resetPath` reads its scheme list rather than restating it&#34;) both assert a single owner.

The failure that follows from divergence is the round-1 bug again: a scheme or launch shape added to `pathMemberIdSchemeKeys` alone would be cleared by `resetPath` but never read by the join, so a member progressed under it resolves `{ percent: 0, source: &#39;unopened&#39; }` and enters the mean as a real zero. Nothing catches it — the `pathMemberIdSchemeKeys` test (path-member-join.test.ts:161-168) asserts that function&#39;s output in isolation, and the launch round-trip asserts `pathMemberContentKeys` contains the launch key, so both still pass while the two lists disagree.

Mechanical fix that restores the single owner: make the grouped list the source of truth (e.g. an unsanitized `idSchemeGroups(memberId)` returning `[[bare, package], [backend-guide]]`), define `pathMemberIdSchemeKeys` as its `.flat()`, and have `pathMemberContentKeyGroups` map `sanitizeContentKey` over each group. Output is unchanged and the grouping semantics the precedence loop depends on stay explicit.

🔧 Fix: scope precedence claim and restore scheme single owner
3 infos still open:
- ℹ️ `docs/design/COMPLETION-MODEL.md:359` - Decision 9 closes the collision discussion with &#34;do not read decision 9 as a guarantee that a private guide&#39;s progress can never surface as a bundled member&#39;s — only that the join&#39;s own scheme fallback will not cause it.&#34; The scheme fallback does cause it, in the case the precedence loop does not cover.

Trace, using the same collision premise the section is built on: a namespace publishes a private guide with id `first-dashboard`, colliding with the bundled package. The reader progresses the private guide to 90 (`backend-guide:first-dashboard`) and has never opened the bundled one. `resolvePathMemberPercentage({id:&#39;first-dashboard&#39;}, ...)` — a static `getting-started` member, so no url — forms both groups (path-member-join.ts:171-176). Group 1 (`bundled:first-dashboard`, `bundled:first-dashboard/content.json`) holds nothing, so the loop falls through (line 232-256). Group 2 answers 90, and the bundled member reports a different guide&#39;s progress with no error. Bundled-first precedence only decides between the schemes when the bundled group actually holds a record; it does nothing when it is empty, and the empty case is the reachable one for a never-opened member. The `answers from the bundled scheme even when backend-guide holds more` test (path-member-join.test.ts:245) exercises only the both-present case, so nothing contradicts the claim.

The behavior itself is intent decision 1 (&#34;take whichever holds a record&#34;) and is required — an App Platform member whose catalogue has not loaded has no url and can only be found under `backend-guide:`. So this is the sentence to correct, not the loop. Two related inaccuracies to fix in the same pass: the contract anchor at COMPLETION-MODEL.md:668 says the join takes the furthest record &#34;across a member&#39;s candidate keys&#34; when the maximum is within-scheme only, and path-member-join.ts:32 scopes the residual collision case to the supplied-url branch when the id-scheme fallback carries the same residual.
- ℹ️ `src/learning-paths/reset-path-completion.test.ts:54` - `resetPath`&#39;s no-base-url branch now clears a third key shape per member — `bundled:&lt;id&gt;/content.json` (learning-paths.hook.ts:447-450, via `pathMemberContentKeys` / `pathMemberIdSchemeKeys`) — across `interactiveStepStorage`, `interactiveCompletionStorage`, `journeyCompletionStorage`, and `evictContentCache`. That is a user-visible behavior change in this diff and nothing exercises it: `MEMBER_KEYS` here is still `GUIDES.flatMap((id) =&gt; [&#39;bundled:&#39; + id, &#39;backend-guide:&#39; + id])`, so `clears every member key in one pass, without restoring siblings, and spares unrelated content` (line 138) passes identically whether or not the package form is cleared.

The uncovered failure is the reset-side twin of the round-1 bug: a reader who progressed a bundled guide from the context panel persists under `bundled:&lt;id&gt;/content.json` (BundledPackageResolver.resolve, resolver.ts:55), hits &#34;Reset path&#34;, and the record survives — the guide reads back as progressed. Adding `&#39;bundled:&#39; + id + &#39;/content.json&#39;` to `MEMBER_KEYS` makes the existing assertions cover it, and it is also the sentinel case for the `clearAllForContent` prefix scan, since `INTERACTIVE_STEPS_PREFIX + &#39;bundled:&lt;id&gt;-&#39;` does not match a package-form step key.
- ℹ️ `src/global-state/path-member-join.ts:232` - Traceability only, no action. Intent decision 1 reads &#34;Read under both key schemes and take whichever holds a record, since a reader can only have progressed under one.&#34; The current loop is neither first-hit-wins nor a flat maximum: it takes the maximum within a scheme group and the first group holding a record across groups. Both halves were directed explicitly by the user (round 2: &#34;Take the maximum finite value across candidate keys rather than first-hit-wins&#34;; round 4: &#34;Narrow furthest-wins to operate WITHIN a scheme, and keep the resolver&#39;s own bundled-before-backend-guide precedence across schemes&#34;), and both retractions are substantiated in source — `bundled:&lt;id&gt;` and `bundled:&lt;id&gt;/content.json` are independently reachable launches for one guide (MyLearningTab.tsx:271 vs `link.contentUrl` from resolver.ts:55) with per-content-key step storage (user-storage.ts:972), while `createCompositeResolver` (composite-resolver.ts:110-112) documents cross-scheme id collisions as possible. The premise is correctly retracted in both the module docstring (line 15-27) and decision 9. A prior round flagged this and it was ignored; noting it again only because the cross-scheme half of the divergence is new since then.

</details>

<details>
<summary>⚠️ **Test** - 2 infos</summary>

- ℹ️ `src/validation/import-graph.test.ts:817` - `src/validation/import-graph.test.ts` › &#39;recognizes prefix-only builtins&#39; fails in this worktree because Node 22.22.0 is active while package.json requires &gt;=24 (`node:test` is absent from builtinModules on 22). Confirmed unrelated to this diff and unfixable inside the worktree boundary since it needs a Node runtime upgrade; CI runs Node 24.
- ℹ️ `src/global-state/path-member-join.storage-round-trip.test.ts` - new test file written by agent: src/global-state/path-member-join.storage-round-trip.test.ts
- <code>`npx jest --coverage=false src/global-state/path-member-join.test.ts src/global-state/content-key.test.ts src/learning-paths/reset-path-completion.test.ts` — existing join, sanitizer and reset coverage</code>
- <code>Wrote `src/global-state/path-member-join.storage-round-trip.test.ts` (5 tests): real `markStepCompleted` → real `interactiveCompletionStorage` → `resolvePathMemberPercentages`, covering App Platform member resolution, the empty journey namespace (decision 4), package-surface bundled launch, furthest-of-two-surfaces, and whole-path exclusion counting</code>
- <code>Negative control: removed the `backend-guide:` group from `idSchemeKeyGroups` → 2 of the 5 new round-trip tests fail; module restored and `git diff` clean</code>
- <code>Added `clears the package-launch shape of a member, which the join also reads` to `src/learning-paths/reset-path-completion.test.ts`</code>
- <code>Negative control: restored `learning-paths.hook.ts` from base commit 677a6aa → that new reset test fails; hook restored and `git diff` clean</code>
- <code>`npx jest --coverage=false src/global-state/path-member-join src/global-state/content-key.test.ts src/global-state/completion-store.test.tsx src/learning-paths/reset-path-completion.test.ts src/learning-paths/learning-paths.hook.test.ts src/learning-paths/learning-paths.completion-round-trip.test.ts src/validation/architecture.test.ts src/validation/control-bytes.test.ts src/validation/unicode-format-characters.test.ts` — 10 suites, 175 tests, all pass</code>
- <code>`npx tsc --noEmit -p tsconfig.json` — clean, covering the new test files</code>
- `Manual evidence run: temporary transcript harness driving real progress writes and printing persisted localStorage plus join output; output saved to the evidence directory and the harness file deleted from the worktree`
- <code>`npx jest --coverage=false src/validation/import-graph.test.ts` — confirmed the single pre-existing Node-22 failure, and verified locally that `node:test` is absent from `builtinModules` on the active runtime</code>

</details>

<details>
<summary>⚠️ **Document** - 2 infos</summary>

- ℹ️ `.cursor/rules/systemPatterns.mdc:113` - The `global-state/` bullet lists 4 of the directory&#39;s 18 modules and does not include the new `path-member-join.ts`. Left unchanged deliberately: the list is exemplar-style (the sibling `utils/` and `lib/` bullets are equally partial) and pre-dates this change, so nothing there became false. A follow-up could decide whether these bullets are inventories or examples and label them accordingly.
- ℹ️ `docs/design/COMPLETION-MODEL.md:41` - The model-on-one-page table&#39;s Path row still reads &#34;the mean of its milestones&#39; percentages&#34; without decision 9&#39;s exclusion qualifier. Left as-is: decision 9 is the owner of that refinement, adding the qualifier would copy the fact into a second prose location, and the change ships no rollup consumer yet. Revisit when the path rollup lands.

</details>

<details>

✅ No issues found.

</details>

<details>

✅ No issues found.

</details>


moxious (changes requested):
Blockers:

1. [blocking] **B1 — resetPath now runs member ids through the lossy content-key sanitizer on the destructive side, so a '..'-bearing id deletes another guide's progress** (medium · security · irreversible without cleanup)
   `sanitize` in `content-key.ts:27` is `value.replace(/\.\./g, '').slice(0, 200)` — a non-injective map, newly exported as `sanitizeContentKey`. `pathMemberContentKeys` maps it over every id-scheme key, and at head `resetPath` feeds those sanitized keys to `interactiveStepStorage.clearAllForContent`, `interactiveCompletionStorage.clearMany` and `evictContentCache` (learning-paths.hook.ts:448-462). At base the same list was built from raw literals and never sanitized, so a malformed id could only ever name its own key. Verified: a member id `welcome..-to-grafana` now produces the clear keys `bundled:welcome-to-grafana` and `bundled:welcome-to-grafana/content.json` — the real shipped guide's keys. A reader who only ever opened the legitimate `welcome-to-grafana` loses that progress when an unrelated path carrying the malformed id is reset. Nothing validates the id on the way in: App Platform `entry.id` and `manifest.milestones` are `z.string()` (backend-api.schema.ts:121,128-133), the Go side rejects only an empty id and reads it from the CR spec rather than `metadata.name` (custom_guide_repository_client.go:302-306), and `app-platform-paths.ts:74` filters members only by `Object.hasOwn(guideMetadata, id)`, which such an id satisfies. The truncation half is reachable without any malformed input at all: `PACKAGE_ID_MAX_LENGTH` is 253 while `MAX_KEY_LENGTH` is 200, so two kebab-legal ids agreeing on their first 192 characters sanitize to one key. The write-side collision is pre-existing in the sanitizer, but base's reset missed it and head's reset deletes it.
   Required: In `pathMemberContentKeyGroups`, refuse to form a candidate whose `sanitizeContentKey(value) !== value`, so an id that does not survive normalization intact yields no key rather than another guide's key — the module already has an `unresolved` outcome for exactly that shape. Applying `packageIdSchema` to `CustomGuideRepositoryEntryWireSchema.id` and `manifest.milestones` would close it at the wire boundary too, but that lands outside this PR's files.

2. [blocking] **F3 — The reset-clears-a-superset-of-what-the-join-reads invariant is asserted in two docs and enforced by nothing** (medium · testing-and-verification)
   `docs/developer/learning-paths/README.md:181` states that sharing the key list with the join "means a reset can never spare a key the join would read a stale percentage back from", and `COMPLETION-MODEL.md:389-397` states that "sanitization never silently narrows what a reset clears". Neither holds structurally. `resetPath`'s no-url branch calls `pathMemberContentKeys({ id: guideId })` with no `url` (learning-paths.hook.ts:449-451) so it forms only the id-scheme keys, while the join given `{ id, url }` short-circuits at path-member-join.ts:203-205 and returns the sanitized `url` alone. `resolveGuideMetadata` tier 3 returns the static `paths.json` guideMetadata entry verbatim, including any `url`, for members of paths that have no `url` of their own. A probe at head shows the two key sets are disjoint for such a member, and `resolvePathMemberPercentage` reads a stale percentage back from a key the reset never touches; the launch does key its record there, because `MyLearningTab.tsx:265-271,304` launches with that url and `getContentKey` is the identity on it. The condition is unenforced rather than presently violated — no shipped guideMetadata entry carries a `url`, verified across both catalogues and their git history — but adding one silently falsifies both doc statements with zero failing tests: `learning-paths.test.ts:67-76` gates its url rule on `if (entry.url)`, and no JSON schema or validator for `paths.json` exists anywhere in the repo.
   Required: Add one test beside the existing "static path catalogues" block asserting that every member of a path with no `url` resolves to metadata carrying no `url`, or only a `bundled:`/`backend-guide:` one. Alternatively soften the two doc sentences to state the containment as a precondition on the data rather than a guarantee.

3. [blocking] **F4 — Decision 9 and the CONTEXT_INDEX row present the exclusion rule as live behavior, but no production code resolves a member percentage** (medium · cross-cutting-architecture)
   Half of decision 9 shipped and half did not, with nothing marking which. `calculatePathProgress` (learning-paths.hook.ts:71-78, reached only through `getPathProgress` at :364) is the only production path-percentage site and is a completed-count fraction; every UI consumer goes through it. `resolvePathMemberPercentage`, `resolvePathMemberPercentages`, `PathMemberPercentage`, `PathMemberJoinContext`, `PathMemberJoinResult` and `PathMemberPercentageSource` appear only in the new module, its two test files, and one JSDoc link — zero production consumers. Yet `COMPLETION-MODEL.md:307,318` states the exclusion rule in the present indicative, the same file's one-page table at :42 still reads "the mean of its milestones' percentages" with no exclusion qualifier, and the Journey row directly below it at :43 carries an explicit "(decision 5 — not built, and not needed to be)" marker that decision 9 has no equivalent of. `CONTEXT_INDEX.md:120` now advertises "the path-member content-key join and its exclusion rule (decision 9)" and adds the module as an auto-load glob, so `AGENTS.md`'s own routing surface points an agent at a module whose percentage half is exercised only by its own tests. The file's global provisional disclaimer softens doc-versus-code generally but does not resolve a half-shipped section when the same file marks decision 5 inline.
   Required: Mark the rollup half of decision 9 as not-yet-built the way decision 5 is marked — one clause on the heading or first paragraph — and add the exclusion qualifier to the Path row at :42. If the intent is that the percentage half lands with its consumer, say so in the same clause so a reader knows the exports are staged rather than live.

Follow-ups:

1. [follow_up] **F5 — The reset key list drives a localStorage prefix sweep, so clearing a member also deletes every guide whose content key begins with that member's key plus a hyphen** (high · security · irreversible without cleanup)
   `clearAllForContent(contentKey)` builds `${INTERACTIVE_STEPS_PREFIX}${contentKey}-` (plus the collapse, acknowledgement and done prefixes) and removes every localStorage key satisfying `startsWith` (user-storage.ts:1042-1069), while a step record is written as `${PREFIX}${contentKey}-${sectionId}` (:986-989). So the sweep prefix for `bundled:first-dashboard` matches every record belonging to `bundled:first-dashboard-cloud`, and both are real shipped bundled packages with their own content.json (repository.json:48, :85), launchable in one profile regardless of catalogue selection. A user resetting one learning path silently destroys a non-member guide's step progress, collapse state, acknowledgements and done markers, with no undo. With author-written App Platform ids the set of hyphen-prefix pairs is not controlled by this repository. This is pre-existing: base made the identical `clearAllForContent` call over a list already containing `bundled:first-dashboard`, and the shapes this PR adds do not widen this particular collision. It surfaces here because the PR makes `pathMemberContentKeys` the single owner of the key list the sweep consumes. The only existing coverage uses the non-colliding pair `guide-a` / `guide-b` (user-storage.test.ts:403-426).
   Follow-up: File a tracked issue to delimit the step, collapse, acknowledgement and done key spaces so one content key cannot be a prefix of another, or to sweep on an exact `${contentKey}-${sectionId}` match set rather than `startsWith`. Do not fix it in this PR — the sweep is in `user-storage.ts` and untouched here.

Suggestions & nits:

1. [suggestion] **S1 — The bundled-first resolution order is mirrored as a literal array order with no test coupling it to createCompositeResolver** (medium · package-engine)
   `idSchemeKeyGroups` hardcodes the bundled group before the backend-guide group (path-member-join.ts:171-176) and the rationale at :22-27 cites `createCompositeResolver` in prose only. Today's mirror is faithful — composite-resolver.ts:114-127 really does put the bundled resolver first and App Platform last. The URL-shape half of that duplication is machine-pinned: `path-member-join.test.ts:77-89` runs `it.each(createBundledResolver().listPackageIds())` against the resolver's own `contentUrl`, and mutating one repository entry's `path` reddens it. The precedence half is only restated in a comment at :245-248. The module is tier 1 and cannot import package-engine at runtime, so duplicating the order is structurally forced; but if `createCompositeResolver` ever reorders its tiers, the join silently keeps bundled-first and the disagreement surfaces as a wrong member percentage rather than a build break.
   Suggested: In the existing test, which already imports package-engine, derive the assertion from `createCompositeResolver`'s resolver list — assert the bundled resolver's index precedes the App Platform resolver's — so an upstream reorder reddens this test.

2. [suggestion] **S2 — unresolvedCount and unresolvedMemberIds also count 'unreadable' members, colliding with the 'unresolved' source value of the same name** (low · correctness-and-reliability)
   `excluded = resolved.filter((entry) => entry.percent === undefined)` (path-member-join.ts:266-274) captures both the `'unresolved'` and the `'unreadable'` source, which the same file declares as distinct, separately documented values at :108-121 — and the field's own docstring at :137-138 already says "excluded", not "unresolved". `path-member-join.test.ts:337-346` asserts an unreadable member into `unresolvedMemberIds`, so the conflation is intended. A first consumer reading `unresolvedCount` will reasonably take it as the `source === 'unresolved'` set and under-report corruption, or show a "could not resolve" message for a member whose record was in fact opened and is corrupt — the two conditions decision 9 spends a paragraph separating. With zero production consumers the name is free to fix now; after the rollup lands it is a rename across call sites.
   Suggested: Rename to `excludedCount` / `excludedMemberIds`, matching the field's own docstring and decision 9's wording, or split the counts per source.

3. [suggestion] **S3 — The sanitized-versus-raw key-space split that is the point of the resetPath change is untested** (low · testing-and-verification)
   The substance of the resetPath change is routing raw scheme keys to `milestoneCompletionStorage` and `journeyCompletionStorage` and sanitized content keys to `interactiveStepStorage` and `interactiveCompletionStorage` (learning-paths.hook.ts:443-460), and the switch of `journeyCompletionStorage.clearMany` from sanitized to raw keys is a real correctness fix — that namespace is keyed by the raw `journeyBaseUrl` (learning-journey-helpers.ts:442,462,688), so under the old code a reset silently spared the journey record for a long id. Nothing pins it. Every id in `reset-path-completion.test.ts` sanitizes to itself (`fe-alerting-01`..`03`, including the new case at :159-176), so the two lists are equal throughout and reverting line 460 to `contentKeys` leaves the suite fully green. The divergence is reachable in production because App Platform members are CR-authored `manifest.milestones` ids that can exceed the 200-character truncation point.
   Suggested: Add one resetPath case with a member id whose sanitized and raw keys differ — seed `journeyCompletionStorage` under the raw key and `interactiveCompletionStorage` under the truncated content key, then assert both are cleared.

4. [nit] **N1 — The 62-line module header restates decision 9's argued rationale, creating a second copy that will drift** (low · correctness-and-reliability)
   The header at path-member-join.ts:1-62 argues furthest-wins-within-scheme, precedence-across-schemes, url-trusted-verbatim, exclusion-not-zero and the eviction boundary. `COMPLETION-MODEL.md:307-419` makes the same five arguments at length, in places near-verbatim — compare path-member-join.ts:30-39 with COMPLETION-MODEL.md:347-357. The header itself defers twice to decision 9 for the follow-on work, so the doc is already the acknowledged owner of the rationale. AGENTS.md keeps comments for hidden invariants, which this module genuinely has; the problem is the argued rationale now living in two places at similar length, and the header is the copy a future editor will not think to update.
   Suggested: Keep in the header only what the code cannot express and a reader needs at the call site — furthest-wins within a scheme, precedence across them, presence-not-value, exclusion-not-zero — and replace the argued paragraphs at :22-39 and :48-58 with the pointer to decision 9 the header already carries.

5. [nit] **N2 — sanitizeContentKey's new JSDoc links a symbol from a module content-key.ts does not and must not import** (low · cross-cutting-architecture)
   The new export's docstring at content-key.ts:30-36 contains `{@link resolvePathMemberPercentage}`. path-member-join.ts:64 imports content-key, not the reverse, so the link has no resolvable target from content-key.ts and points up its own dependency edge. It renders dead in editor hover tooltips and names the sanitizer's consumer inside the sanitizer, so it goes stale as soon as the join is renamed or the rollup consumer replaces it. The rest of the comment earns its place — it records why the key must match what the member persisted under.
   Suggested: Drop the `{@link ...}` wrapper and keep the plain prose, or name the module rather than the symbol.

6. [nit] **N3 — PathMember.url's docstring and one test name a state production cannot produce** (low · correctness-and-reliability)
   path-member-join.ts:70-76 documents `url` as "absent for bundled guides and for App Platform members whose catalogue has not loaded yet". The bundled half is true and load-bearing; the App Platform half is not. `appPlatformData` is one state value holding both `paths` and `guideMetadata`, written once per fetch and read by the paths memo and `resolveGuideMetadata` in the same render (learning-paths.hook.ts:101,155,165-176,190-213), so a path cannot appear while its metadata is absent — and `app-platform-paths.ts:53-58,73` stamps `url: backend-guide:<id>` for every published entry before filtering members to ids with an own metadata key, so every surviving member is guaranteed a url-bearing entry. `path-member-join.test.ts:65-74` pins the id-scheme fallback for that state, and the only place a member reaches the join with `url` undefined is `learning-paths.hook.test.ts:118-135`, which mocks a shape the real adapter's filter cannot produce. That clause is what justifies consulting the backend-guide scheme from the id-scheme branch at all, so a reader auditing the precedence rules will look for a production caller and not find one.
   Suggested: Drop the "App Platform members whose catalogue has not loaded yet" clause, or qualify it as a defensive case with no production caller, and reword the test name to match.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1843
Summary: join path members to their saved percentage by content key
Verdict: Request Changes
Results: 3 blockers, 6 non-blocking findings, 1 follow-up

moxious (approved):

## Follow-ups

1. [follow_up] **F5 — The reset key list drives a localStorage prefix sweep, so clearing a member also deletes every guide whose content key begins with that member's key plus a hyphen** (high · security · irreversible without cleanup)
   Unchanged at head, and correctly so — `src/lib/user-storage.ts` is not in this PR's changed-file set. `clearAllForContent(contentKey)` still builds `${PREFIX}${contentKey}-` for the interactive-steps, section-collapse, section-acknowledged and section-done spaces and removes every localStorage key satisfying `startsWith` (:1042-1071), while a step record is written as `${PREFIX}${contentKey}-${sectionId}`; there is no delimiter, terminator or exact key set. Both halves were re-proved by execution at head: driving the real `useLearningPaths.resetPath` over the verbatim `getting-started` path wiped all four namespaces belonging to `bundled:first-dashboard-cloud` while a non-prefix control survived, and `pathMemberContentKeys` was asserted not to contain the victim key, so the deletion is purely the sweep. The scan reads `localStorage` directly, bypassing the UserStorage abstraction, so the hybrid backend path does not narrow it. One detail worth adding to the tracked issue: the second colliding pair is stronger than the one already recorded — `welcome-to-grafana` is a `getting-started` member in `paths.json` while `welcome-to-grafana-cloud` also ships in `repository.json`, so that collision needs no cross-catalogue mixing, whereas `first-dashboard` / `first-dashboard-cloud` sit in different platform path files.
   Follow-up: Leave it out of this PR and close it under issue #1846, which already records the condition and both remedies: delimit the four key spaces so one content key cannot be a prefix of another, or sweep an exact `${contentKey}-${sectionId}` match set instead of `startsWith`. Add the `welcome-to-grafana` / `welcome-to-grafana-cloud` pair to that issue and use a genuinely colliding pair in the regression test — the existing coverage uses the non-colliding `guide-a`/`guide-b` (`user-storage.test.ts:403-426`).

## Suggestions

1. [suggestion] **S1 — The new scheme-precedence cross-check does not redden when createCompositeResolver reorders its tiers, so the mirrored bundled-first order is still uncoupled** (medium · package-engine)
   `path-member-join.test.ts:182-204` resolves the bundled id `first-dashboard` through `createCompositeResolver` and asserts that `pathMemberIdSchemeKeys(packageId)[0]` starts with the scheme the composite answered under. In jest the other three tiers cannot answer — no backend, no network — so the composite falls through to the bundled resolver and `contentUrl` starts with `bundled:` whatever the list order is. Proved it: patching `composite-resolver.ts:115,123` to put `new AppPlatformPackageResolver()` first and `createBundledResolver()` last — the exact upstream reorder round 1 asked to be detected — leaves the precedence block green (2 passed). The hardcoded bundled-before-backend-guide order at `path-member-join.ts:159-164` therefore remains a mirror of `composite-resolver.ts:114-127` with nothing pinning it, and a future reorder would surface as a wrong member percentage rather than a build break.
   Suggested: Assert on the resolver list rather than on a resolution outcome: derive the ordered resolver identities from `createCompositeResolver` and assert the bundled resolver's index precedes the App Platform resolver's, so a reorder reddens regardless of which tier can answer in a test environment.

2. [suggestion] **S4 — The new normalization guard drops candidates individually, so a partly-dropped group scores the member 0 instead of the exclusion the module's own rule requires** (low · correctness-and-reliability)
   `normalizedCandidates` filters each candidate independently (`path-member-join.ts:141-143`) and only wholly empty groups are removed (:192-198), so a group that keeps one launch shape and loses another is still searched and can fall through to `'unopened'`. The PR's own test pins that shape: for a 185-character id, `pathMemberContentKeys` returns `bundled:<id>` and `backend-guide:<id>` while the 206-character `bundled:<id>/content.json` is dropped (`path-member-join.test.ts:245-249`). Executed at head: with a record present under only that dropped package shape, `resolvePathMemberPercentage` returns `{ percent: 0, source: 'unopened' }` — a real record scored as a zero that enters the mean, which is the one failure the module header (:28-30) and decision 9 call load-bearing to avoid. The same over-broad filter costs read fidelity on the url branch for no destructive benefit, since `resetPath` never builds keys from a member url: a url the sanitizer rewrites now yields no key at all, though `getContentKey` persists the record under exactly that sanitized key (the test at :123-129 flipped from asserting the key is formed to asserting it is not). Not reachable with today's content — the longest shipped bundled id is 27 characters, and an App Platform member's single `backend-guide:` candidate is dropped as a whole group, which correctly yields `'unresolved'`.
   Suggested: Drop the whole group when any of its candidates fails normalization, so the outcome is `'unresolved'` rather than `'unopened'`. That keeps the containment claim the README now makes — reset and join still refuse the same keys — while restoring exclusion-not-zero.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1843
Summary: resolve a path member's saved percentage
Verdict: Approve with Minor

PR #1845, opened 2026-09-09
fix(block-editor): compile-check branch block builders

### LINKED ISSUE #1655: Make the branch add-picker's block-type list compile-checked against createDefaultBlock

dvd233:
## Summary

The branch add picker now accepts only block types backed by an intentional default builder, so removing or forgetting a builder becomes a TypeScript error instead of silently adding an empty markdown block. The same contract is applied to the collapsible picker and the `addableBlockTypes` prop; legacy fallback behavior for non-picker block types remains unchanged.

## What to look at

- [Builder registry and branch picker](https://github.com/dvd233/grafana-pathfinder-app/blob/bccf64e2adcc80ae95f306a1a03059b22b1b3663/src/components/block-editor/forms/BranchBlocksEditor.tsx): `DefaultableBlockType` is derived from `BLOCK_DEFAULT_BUILDERS`; `createDefaultBlock` delegates buildable types while retaining its legacy fallback switch.
- [Collapsible picker](https://github.com/dvd233/grafana-pathfinder-app/blob/bccf64e2adcc80ae95f306a1a03059b22b1b3663/src/components/block-editor/forms/CollapsibleBlockForm.tsx): its presentational list uses the same type, so both nested and branch entry points reject types without builders.

## Why

Issue #1655 identified two independent hand-maintained lists and a broad prop type. Without a shared type, deleting an arm or adding a fallback-only type can compile and produce an empty markdown block at runtime. The registry keeps the existing ten branch and five collapsible choices while making drift fail at compile time.

## How to verify

1. Open the branch block editor and add each default picker option; confirm the created block's type and shape match the selected option.
2. Open a collapsible block, add each presentational option, save, and reopen it; confirm the nested block type and content are preserved.
3. Temporarily remove the `guided` entry from `BLOCK_DEFAULT_BUILDERS` (or add `terminal` to `ALLOWED_BRANCH_BLOCK_TYPES`) and run the TypeScript checker; confirm it reports a type error at the stale list, then restore the entry.
4. Pass a fallback-only block type such as `terminal` to `addableBlockTypes` in a TypeScript call site; confirm it is rejected before runtime.

## Follow-up

The follow-up commit closes the review gaps without widening the runtime surface: builder values are keyed to their JsonBlock discriminants, Object.hasOwn guards the registry, the branch and collapsible pickers use safe builder-backed types (excluding challenge), and the legacy fallback has a regression assertion. The #1542 picker invariant is recorded in a named type and comment. This extends the original compile-check change rather than replacing the existing fallback contract.

## Breaking changes

The TypeScript `addableBlockTypes` prop is intentionally narrower and now rejects non-buildable block types. Existing valid call sites and runtime picker behavior are unchanged.

## Out of scope

- Narrowing `createDefaultBlock`'s own `BlockType` parameter or removing its legacy fallback remains separate, as requested by the issue.

Fixes #1655

### DIFF
--- a/src/components/block-editor/forms/BranchBlocksEditor.test.ts
+++ b/src/components/block-editor/forms/BranchBlocksEditor.test.ts
@@ -1,8 +1,34 @@
-import { ALLOWED_BRANCH_BLOCK_TYPES, createDefaultBlock } from './BranchBlocksEditor';
+import { ALLOWED_BRANCH_BLOCK_TYPES, createDefaultBlock, type BranchBlocksEditorProps } from './BranchBlocksEditor';
+
+type AddableBlockTypes = NonNullable<BranchBlocksEditorProps['addableBlockTypes']>;
+
+const buildablePropTypes: AddableBlockTypes = ['markdown', 'challenge'];
+// @ts-expect-error terminal has no intentional default builder
+const fallbackOnlyPropTypes: AddableBlockTypes = ['terminal'];
+
+void buildablePropTypes;
+void fallbackOnlyPropTypes;
 
 describe('BranchBlocksEditor createDefaultBlock', () => {
-  it('does not offer challenge in the inline add picker', () => {
-    expect(ALLOWED_BRANCH_BLOCK_TYPES).not.toContain('challenge');
+  it('offers exactly the block types the branch add picker can build', () => {
+    expect(ALLOWED_BRANCH_BLOCK_TYPES).toEqual([
+      'markdown',
+      'divider',
+      'interactive',
+      'image',
+      'video',
+      'input',
+      'callout',
+      'quiz',
+      'multistep',
+      'guided',
+    ]);
+  });
+
+  it('constructs every offered block without changing its type', () => {
+    for (const type of ALLOWED_BRANCH_BLOCK_TYPES) {
+      expect(createDefaultBlock(type).type).toBe(type);
+    }
   });
 
   it('builds a challenge block instead of empty markdown if challenged', () => {
@@ -14,17 +40,7 @@ describe('BranchBlocksEditor createDefaultBlock', () => {
     });
   });
 
-  it('still defaults unknown types to empty markdown', () => {
-    // html is legacy / palette-excluded and not in the creatable list
-    expect(createDefaultBlock('html' as any)).toEqual({ type: 'markdown', content: '' });
-  });
-
-  it('offers callout in the inline add picker', () => {
-    expect(ALLOWED_BRANCH_BLOCK_TYPES).toContain('callout');
-  });
-
-  it('offers and builds a divider in the inline add picker', () => {
-    expect(ALLOWED_BRANCH_BLOCK_TYPES).toContain('divider');
+  it('builds an empty divider block', () => {
     expect(createDefaultBlock('divider')).toEqual({ type: 'divider' });
   });
 
--- a/src/components/block-editor/forms/BranchBlocksEditor.test.ts
+++ b/src/components/block-editor/forms/BranchBlocksEditor.test.ts
@@ -2,11 +2,14 @@ import { ALLOWED_BRANCH_BLOCK_TYPES, createDefaultBlock, type BranchBlocksEditor
 
 type AddableBlockTypes = NonNullable<BranchBlocksEditorProps['addableBlockTypes']>;
 
-const buildablePropTypes: AddableBlockTypes = ['markdown', 'challenge'];
+const buildablePropTypes: AddableBlockTypes = ['markdown', 'guided'];
+// @ts-expect-error challenge has a builder but cannot be edited through this picker
+const nonAddableBuilderType: AddableBlockTypes = ['challenge'];
 // @ts-expect-error terminal has no intentional default builder
 const fallbackOnlyPropTypes: AddableBlockTypes = ['terminal'];
 
 void buildablePropTypes;
+void nonAddableBuilderType;
 void fallbackOnlyPropTypes;
 
 describe('BranchBlocksEditor createDefaultBlock', () => {
@@ -31,6 +34,10 @@ describe('BranchBlocksEditor createDefaultBlock', () => {
     }
   });
 
+  it('does not offer challenge without an inline editor', () => {
+    expect(ALLOWED_BRANCH_BLOCK_TYPES).not.toContain('challenge');
+  });
+
   it('builds a challenge block instead of empty markdown if challenged', () => {
     expect(createDefaultBlock('challenge')).toEqual({
       type: 'challenge',
@@ -42,6 +49,10 @@ describe('BranchBlocksEditor createDefaultBlock', () => {
 
   it('builds an empty divider block', () => {
     expect(createDefaultBlock('divider')).toEqual({ type: 'divider' });
+  });
+
+  it('keeps the legacy fallback for non-defaultable block types', () => {
+    expect(createDefaultBlock('terminal')).toEqual({ type: 'markdown', content: '' });
   });
 
   it('builds an empty callout block, not empty markdown', () => {
--- a/src/components/block-editor/forms/BranchBlocksEditor.tsx
+++ b/src/components/block-editor/forms/BranchBlocksEditor.tsx
@@ -263,16 +263,13 @@ const getStyles = (theme: GrafanaTheme2) => ({
 // Helper Functions
 // ============================================================================
 
-/**
- * Create a default block of a given type
- */
 const BLOCK_DEFAULT_BUILDERS = {
-  markdown: (): JsonBlock => ({ type: 'markdown', content: '' }),
-  divider: (): JsonBlock => ({ type: 'divider' }),
-  interactive: (): JsonBlock => ({ type: 'interactive', action: 'highlight', reftarget: '', content: '' }),
-  image: (): JsonBlock => ({ type: 'image', src: '' }),
-  video: (): JsonBlock => ({ type: 'video', src: '' }),
-  quiz: (): JsonBlock => ({
+  markdown: () => ({ type: 'markdown', content: '' }),
+  divider: () => ({ type: 'divider' }),
+  interactive: () => ({ type: 'interactive', action: 'highlight', reftarget: '', content: '' }),
+  image: () => ({ type: 'image', src: '' }),
+  video: () => ({ type: 'video', src: '' }),
+  quiz: () => ({
     type: 'quiz',
     question: '',
     choices: [
@@ -280,18 +277,21 @@ const BLOCK_DEFAULT_BUILDERS = {
       { id: 'b', text: '' },
     ],
   }),
-  input: (): JsonBlock => ({ type: 'input', prompt: '', inputType: 'text', variableName: '' }),
-  multistep: (): JsonBlock => ({ type: 'multistep', content: '', steps: [] }),
-  guided: (): JsonBlock => ({ type: 'guided', content: '', steps: [] }),
-  challenge: (): JsonBlock => ({ type: 'challenge', title: '', brief: '', successCriteria: '' }),
-  callout: (): JsonBlock => ({ type: 'callout', title: '', content: '' }),
-} as const satisfies Partial<Record<BlockType, () => JsonBlock>>;
+  input: () => ({ type: 'input', prompt: '', inputType: 'text', variableName: '' }),
+  multistep: () => ({ type: 'multistep', content: '', steps: [] }),
+  guided: () => ({ type: 'guided', content: '', steps: [] }),
+  challenge: () => ({ type: 'challenge', title: '', brief: '', successCriteria: '' }),
+  callout: () => ({ type: 'callout', title: '', content: '' }),
+} as const satisfies { [K in BlockType]?: () => Extract<JsonBlock, { type: K }> };
 
 /** Block types that have an intentional default builder. */
 export type DefaultableBlockType = keyof typeof BLOCK_DEFAULT_BUILDERS;
 
+/** Builder-backed block types safe to offer in the branch picker (see #1542). */
+export type BranchAddableBlockType = Exclude<DefaultableBlockType, 'challenge'>;
+
 function isDefaultableBlockType(type: BlockType): type is DefaultableBlockType {
-  return type in BLOCK_DEFAULT_BUILDERS;
+  return Object.hasOwn(BLOCK_DEFAULT_BUILDERS, type);
 }
 
 export function createDefaultBlock(type: BlockType): JsonBlock {
@@ -317,7 +317,6 @@ export function createDefaultBlock(type: BlockType): JsonBlock {
   }
 }
 
-// Block types offered by default in the branch add picker.
 export const ALLOWED_BRANCH_BLOCK_TYPES = [
   'markdown',
   'divider',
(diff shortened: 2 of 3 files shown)


moxious (changes requested):
Thanks for picking up #1655 and implementing the registry-plus-derived-type fix — deriving `DefaultableBlockType` from `BLOCK_DEFAULT_BUILDERS` and putting the constraint on the `addableBlockTypes` prop is exactly the right place for it, because it covers both pickers and every future call site at once. The `isDefaultableBlockType` predicate combined with the retained `assertExhaustive` is a nice bonus: the eleven registry keys and ten fallback arms are now a compile-verified total partition of `BlockType`, which is stronger than what base had.

I am requesting all six items below as changes for this round. To be straight with you about their weight: the review policy classified none of them as blocking — two mediums, two lows and a nit — so this is a maintainer decision to have them addressed together rather than carried, not a claim that any one of them breaks the build. They are ordered heaviest first, and each keeps its original classification tag so you can see what I think it is worth.

The first two are the ones I would fix even if the rest were dropped: the collapsible picker's `satisfies` does not bind the constraint that actually matters, and the deleted comment was the repository's only reference to #1542.

## Requested changes

1. [suggestion] **F-1845-collapsible-presentational — The collapsible picker gets the weaker of the two available constraints, leaving its `as PresentationalBlock[]` cast unguarded** (medium · block-editor-authoring)
   `PRESENTATIONAL_ADDABLE_TYPES` is now `as const satisfies readonly DefaultableBlockType[]` (CollapsibleBlockForm.tsx:20-26), but `DefaultableBlockType` is the eleven builder keys, so `interactive`, `quiz`, `input`, `multistep`, `guided` and `challenge` all satisfy it. The list's real invariant is narrower: `JsonCollapsibleBlock.blocks` is `PresentationalBlock[]` (json-guide.types.ts:216, 223-224) and `JsonCollapsibleBlockSchema.blocks` is `z.array(PresentationalBlockSchema)`. `buildBlock` launders the difference with `blocks as PresentationalBlock[]` (CollapsibleBlockForm.tsx:47-58) under a comment claiming the add menu restricts it to content blocks, and nothing validates a collapsible's nested blocks on the editor save path. Adding `'quiz'` to that list would typecheck, satisfy the cast, and surface only as a later schema rejection. Base annotated the same list `BlockType[]`, so this is not a regression - it is the missed half of the improvement, and the cast's comment now reads as if it were proven.
   Suggested: Narrow to the intersection so the cast becomes type-backed: `] as const satisfies readonly (DefaultableBlockType & PresentationalBlock['type'])[];`. `PresentationalBlock` is already imported in that file.

2. [suggestion] **F-1845-picker-invariant-unrecorded — The prop is typed by "has a builder" while the picker invariant is "safe to offer", and the comment recording the difference was deleted** (medium · cross-cutting-architecture)
   `challenge` has a builder, so `DefaultableBlockType` includes it and `addableBlockTypes?: readonly DefaultableBlockType[]` (BranchBlocksEditor.tsx:393-394) accepts `['challenge']`. But `INLINE_EDITABLE_TYPES` (:336) excludes `challenge`, `quiz`, `multistep` and `guided` and gates editing at :627, so a `challenge` offered through the prop is created with empty fields and is then not inline-editable. The deleted comment at base :317-320 recorded exactly that invariant and cited #1542 - and a grep of the whole repository shows base :320 was the single reference to #1542 anywhere in it. The replacement comment, "Block types offered by default in the branch add picker", restates the name and drops both the rationale and the issue reference. AGENTS.md keeps comments that record hidden invariants the type system cannot express and external-issue references; this was one, and the new type does not carry the half it lost.
   Suggested: Restore a one-line comment stating the invariant and the #1542 reference - corrected, since `quiz`, `multistep` and `guided` are offered but not inline-editable - or encode it as a second alias such as `type BranchAddableBlockType = Exclude<DefaultableBlockType, 'challenge'>` and type the prop with that. Consider also keeping a named exclusion assertion (`expect(ALLOWED_BRANCH_BLOCK_TYPES).not.toContain('challenge')`) beside the new list equality, which is order-sensitive and has no oracle independent of the source array.

3. [suggestion] **F-1845-builder-shape-uncheck — The per-builder `(): JsonBlock` annotations defeat the key-to-shape half of the compile check** (low · correctness-and-reliability)
   Every builder is annotated `(): JsonBlock` (BranchBlocksEditor.tsx:269-288) and the constraint is `satisfies Partial<Record<BlockType, () => JsonBlock>>` (:288), whose value type is the same widened union. A builder keyed `image` returning `{ type: 'markdown', content: '' }` therefore typechecks, and `DefaultableBlockType = keyof typeof BLOCK_DEFAULT_BUILDERS` (:291) derives from keys only. Builder presence is now compile-checked, but the higher-value invariant - key K builds a block of type K, which is the shape of the #1655 bug - is carried only by the runtime loop at test.ts:28-32. `JsonBlock` is a discriminated union on `type`, so the type could carry it for free.
   Suggested: Drop the per-builder `(): JsonBlock` annotations and constrain with a mapped type: `} as const satisfies { [K in BlockType]?: () => Extract<JsonBlock, { type: K }> };`. A mismatched body then becomes a compile error and the runtime loop becomes belt-and-braces.

4. [follow_up] **F-1845-fallback-test — The only runtime assertion on createDefaultBlock's legacy fallback is deleted** (low · testing-and-verification)
   Base test.ts:17-20 asserted `createDefaultBlock('html' as any)` returns `{ type: 'markdown', content: '' }`; head's suite exercises only defaultable types, and `createDefaultBlock` has no other test anywhere in `src/` or `tests/`. Verified as no current harm: the ten fallback arms are byte-identical context lines in the diff, and the `isDefaultableBlockType` predicate plus `assertExhaustive` make the eleven registry keys and ten switch arms a compile-verified total partition of all 21 `JsonBlock['type']` members. The fallback is also effectively unreachable at head, since `addableBlockTypes` is typed `readonly DefaultableBlockType[]` and the inline-edit path is gated by `INLINE_EDITABLE_TYPES.includes` at :627. What a compile error cannot catch is a change to the fallback's returned value, or a silent re-partition moving a type between the registry and the fallback - the class the deleted test guarded.
   Follow-up: Re-add one assertion for a non-defaultable type. No cast is needed now that the picker types are narrower: `expect(createDefaultBlock('terminal')).toEqual({ type: 'markdown', content: '' })`.

5. [follow_up] **F-1845-in-operator-guard — `isDefaultableBlockType` uses `in`, so the new guard is not runtime-total where the base switch was** (low · correctness-and-reliability)
   `return type in BLOCK_DEFAULT_BUILDERS` (BranchBlocksEditor.tsx:293-295) consults the prototype chain of a plain object literal, so it reports true for `constructor`, `toString` and `valueOf`. A true result feeds straight into `BLOCK_DEFAULT_BUILDERS[type]()` at :297-300, which for `'constructor'` evaluates `Object()` and yields `{}` with no `type` field while the declared return type still says `JsonBlock`. Base's flat switch with a no-op `assertExhaustive` returned the markdown stub for any string, so it was runtime-total. Not reachable today: none of the 21 block-type literals collide with an `Object.prototype` key, and both live call paths are gated by the combobox options and by `INLINE_EDITABLE_TYPES.includes` at :627. But `createDefaultBlock` is exported and this guard is the new load-bearing gate between the registry and the fallback.
   Follow-up: `return Object.hasOwn(BLOCK_DEFAULT_BUILDERS, type);`

6. [nit] **F-1845-misplaced-jsdoc — createDefaultBlock's JSDoc is left attached to the new builder map** (low · correctness-and-reliability)
   `/** Create a default block of a given type */` now sits immediately above `const BLOCK_DEFAULT_BUILDERS = {` (BranchBlocksEditor.tsx:266-269), while `createDefaultBlock` at :297 has no doc comment. At base the same comment sat directly above the function, where it was correct. Hovering `BLOCK_DEFAULT_BUILDERS` in an editor now reports a description of the function.
   Suggested: Delete the JSDoc. Both declarations are self-describing, and `DefaultableBlockType` already carries an earned one-liner at :290.

Next: this PR is from a fork, so its CI workflows were held pending maintainer authorization. They have since been released and are running — `zizmor`, `TruffleHog`, `Bundle Stats` and the Grafana API compatibility check are already green.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1845
Summary: compile-check branch block builders
Verdict: Request Changes
Results: 6 blockers, 0 non-blocking findings, 0 follow-ups

dvd233:
Thanks for the review. The follow-up commit `0f7b07776b0d2909c01f71cebcaa124b690fa70d` addresses all six requested items: the collapsible picker is constrained to presentational builder types, the #1542 invariant is recorded, builder keys are tied to their block shapes, the legacy fallback has a regression assertion, the registry guard uses `Object.hasOwn`, and the misplaced JSDoc is corrected. The focused BranchBlocksEditor suite passes 7/7, with typecheck and targeted ESLint also clean. Could you re-review the current head?

moxious (approved):
Thanks for closing out all six review items from the last round in one follow-up commit — each fix matches the shape asked for and none of them reopened anything: the collapsible picker now uses the `DefaultableBlockType & PresentationalBlock['type']` intersection, `BranchAddableBlockType` carries the #1542 invariant with a comment and a `@ts-expect-error` regression test, the builder map is checked against `Extract<JsonBlock, { type: K }>` per key, the legacy-fallback assertion and the `Object.hasOwn` guard are both back, and the stray JSDoc is gone.

PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1845
Summary: compile-check branch block builders
Verdict: Approve

Next: `CI Gate` is green and everything else required has passed; the advisory e2e matrix jobs are still finishing but do not block merge. This is ready to merge.

PR #1848, opened 2026-09-09
feat(completion): add the universal mark complete affordance

tomglenn:
## Summary

Every guide now ends in a **Mark complete** control; every milestone gets **Mark complete and continue**. It renders unconditionally — no predicate on whether the guide has interactive content — per decision 2 of `docs/design/COMPLETION-MODEL.md`.

This matters because most of the published library has no completable block, so those guides previously had no way to reach 100% at all. Reaching 100% is what mints a durable completion record, awards badges and moves path progress.

## How to verify

Open a guide with no interactive steps, click **Mark complete**, and confirm the guide reads 100% everywhere — the footer, the panel, my learning — and that it survives a page reload.

Also worth exercising: the control appears on guides that *do* have interactive steps (that is the point of it being unconditional, not an oversight); a milestone's button both completes and continues; and a marked guide can still be reset, after which it is clickable again.

Two tabs on the same guide: mark it in one, and the other stops offering the button rather than showing 100% beside a live control.

## What to look at

- **`src/components/mark-complete/`** — the control. Rendered from inside `ContentRenderer`, one site, so the sidebar, floating, full-screen, guide-reader and block-editor-preview surfaces all get it without each re-deciding. It resolves its content key once into state from an effect, because both producers of that key publish it from a layout effect that runs after render — resolving during render hands the fresh milestone the previous one's key.
- **`triggerGuideComplete` in `content-renderer.tsx`** — the single gate every completion route now passes through, so a guide records exactly one completion per lifetime however it was finished.
- **`guideCompletionMarkStorage`** — new two-state namespace, keyed by content key alone, beside the per-section acknowledgement rather than replacing it. The mark is authoritative for the guide percentage, so a marked guide stays at 100% when the reader later completes or redoes a step, and `getGuideProgress` settles on it before counting anything — it runs as a `useSyncExternalStore` snapshot, so a storage scan there would be a scan per render.
- **The completion store's storage listener** — extended to the mark namespace, matching the content key exactly, so a mark written in one tab reaches the others.
- **`UserInteraction.MarkCompleteClicked`** — a distinct event from the completion it causes, so a high completion rate with no clicks is detectable rather than having to be inferred. Once per click, never on render or on a return visit, and never from a block-editor preview. Payload carries no guide content.

## Deliberate choices

- **No completion route is removed.** The existing automatic routes and the three navigation auto-complete sites are untouched; retiring them depends on this landing first.
- **The cover page is the one place the control is absent.** A table of contents is neither a guide nor a milestone, and marking one complete would record a guide nobody read.
- **`finalCompletablePosition` stays an authoring signal.** No rendering predicate is re-derived from it.
- **The preview guard uses the narrow `isBlockEditorPreviewUrl`, not `isPreviewContentKey`.** The latter's `devtools` arm is a substring test written for a resolved key; at this call site the key is a URL, so it would capture a real guide whose path contains that word.

## Known gaps, not addressed here

- A reset followed by a re-mark in the same session records no new durable completion: `recordGuideCompletion`'s app-load dedupe set is never cleared by any reset path. Pre-existing, and closing it belongs with the completion store.
- The block-editor preview footer always reads 0%, because preview progress deliberately lives only in memory.
- Marking a guide, reloading, then completing one step emits a second completion record. Cross-session replay is the recorder's documented contract and downstream dedupe owns it.

## Breaking changes

None. Additive, plus docstring corrections where this change falsified them.

## Reversibility

Not free, and `docs/design/COMPLETION-MODEL.md` now has a "Rolling this back" section spelling it out. The control is one component rendered from one site, but the mark namespace is entirely new, so a rollback removes its readers, writers and sweepers together while two things stay behind:

- The `100` the mark path writes into `interactiveCompletionStorage`. That is kept on purpose — it is the reader's own statement that they finished, in the namespace recommendation cards and context already read. On a prose-only guide nothing would later overwrite it, and after a rollback the only corrective is the global "Reset all learning progress".
- The `guide-complete-mark-*` keys themselves, swept today by all three reset scopes, none of which survive the rollback.

The namespace is uncapped, matching `sectionAcknowledgementStorage`, the per-content-key namespace it sits beside. The two bounded percentage namespaces are a different shape — one shared record each — and giving the mark that shape would cost the exact-key cross-tab match the storage listener now depends on.


moxious (dismissed):
## Merge contract

1. [blocking] **mark-preview-guard-misplaced-and-overbroad — The click handler's preview guard is wrong in both directions: the durable completion emitter fires outside it, and its predicate matches any guide key containing the substring devtools** (medium · correctness-and-reliability)
   In `MarkCompleteFooter.handleClick`, `onMarkComplete?.()` runs at line 165, five lines above the guard opened at line 170. Everything the guard was written for — the analytics event, the mark persist, the 100% write, the progress dispatch — sits inside it; the one irreversible side effect, the durable completion record, does not. That is inert today only because `BlockPreview` happens to mount `ContentRenderer` with no `onGuideComplete`, and `content-renderer.mark-complete.test.tsx:153-167` actively pins the leak as correct by asserting `onGuideComplete` was called once for a preview. Separately, the guard's predicate is `isPreviewContentKey`, whose `devtools` arm is a substring test — and this PR's own comment on the narrower `isBlockEditorPreviewUrl` explains why that is wrong for a URL: "applied to a URL it would capture any docs page whose path happens to contain the word". At this call site the resolved key IS a URL, so a guide addressed as `backend-guide:<author-chosen-id>` containing `devtools` takes the preview branch: the reader clicks, sees Completed, a durable record is written, and nothing persists — no mark, so the guide reads incomplete on every later visit, and the click never reaches the click-rate metric this control exists to produce. Two skeptics confirmed both halves; the branch has no test coverage because `MarkCompleteFooter.test.tsx:44` mocks the predicate to a constant `false`.
   Required: Move `onMarkComplete?.()` inside the guard, and gate that guard on the narrow `isBlockEditorPreviewUrl` this PR already extracted rather than on `isPreviewContentKey` — the devtools arm is dead weight here, since the devtools tab carries `content: null` and never mounts `ContentRenderer`. Flip the preview test to assert `onGuideComplete` is not called, and stop mocking `isPreviewContentKey` to a constant.

2. [blocking] **mark-cross-tab-invalidation-missing — The newly authoritative mark namespace is excluded from the completion store's cross-tab storage listener, so a second tab shows 100% complete beside an enabled Mark complete button** (medium · completion-records)
   `completion-store.ts:880` drops every `storage` event outside `INTERACTIVE_STEPS_PREFIX`, so a write to `grafana-pathfinder-app-guide-complete-mark-*` never evicts a cache and never calls `notify`. But `getGuideProgress` now depends on `guideCompletionMarkStorage.isMarked`, a direct `localStorage.getItem` on storage that is shared across tabs. The result is a split: tab B's percentage reads tab A's mark, while tab B's `marked` state does not — the footer re-reads the mark only on `[contentUrl, clearedCount]`, and `clearedCount` advances only on a same-tab cleared event. A skeptic reproduced it: after writing the mark key and dispatching the storage event, `subscribeProgress` reported zero notifications, and one unrelated re-render then rendered "100% complete" with the button present and enabled. Clicking it also mints a second durable record: the recorder's exactly-once `emitted` Set is module-level (per tab), and each queued write takes a `crypto.getRandomValues` idempotency key that the backend hashes into the record name, so the two writes cannot collapse. The store already solves exactly this for the interactive-steps namespace, including a count-cache invalidation; the namespace that is now authoritative was not added to it.
   Required: Extend `handleStorageEvent` to recognise `StorageKeys.GUIDE_COMPLETION_MARK_PREFIX` with an exact content-key match and call `notify(contentKey)`, and have `MarkCompleteFooter` re-read the mark on that signal so a marked guide never presents a clickable button.

3. [blocking] **mark-authority-revert-residue — The PR states reversibility is straightforward, but a revert strands the new namespace and leaves the 100% values the mark path wrote with only a global reset able to correct them** (medium · reversibility-and-one-way-door · irreversible without cleanup)
   `GUIDE_COMPLETION_MARK_PREFIX` is a new dynamic per-guide namespace whose entire implementation this PR adds, so a revert removes every reader, writer and sweeper while the keys stay in localStorage; `syncFromGrafanaStorage`'s `keysToSync` is five exact keys and cannot express a prefix, so the Grafana-side copies are unreachable too, and there is no production caller of `UserStorage.clear()` that would sweep either. Meanwhile both `refreshGuidePercentage` and the footer write `interactiveCompletionStorage.set(contentKey, 100)` into a namespace that survives the revert and that `context.service.ts` reads directly for recommendation cards and context. Post-revert the only writer that would overwrite that value is `refreshGuidePercentage`, reached only from step- or section-driven paths — so on a prose-only guide, the 51.5% of the library this control exists for, the 100% is permanent. The reader's own escape hatch also disappears: the per-guide reset affordance is gated on `hasInteractiveProgress`, and it is this PR's `useGuideProgressState` change that makes a mark count as progress, so a revert hides that button again on exactly the guides carrying the stale value. The blunt global "Reset all learning progress" is the only corrective left. The namespace is also uncapped, where both sibling percentage namespaces are `createBoundedRecordStorage` with a 100-entry limit and a quota notifier. `COMPLETION-MODEL.md` gains a "Where the code stands" rewrite with no rollback, migration or cleanup note and no version stamp.
   Required: Record the rollback posture in `docs/design/COMPLETION-MODEL.md`: either accept the stored 100 as the reader's own durable statement and say so, or stop writing `interactiveCompletionStorage` from the mark path and derive the percentage from the mark alone, so a revert restores the derived value with no residue. State how orphaned `guide-complete-mark-*` keys are swept, and give the namespace the bounded-record cap its siblings have.

4. [blocking] **guide-progress-snapshot-scans-localstorage-per-render — getGuideProgress gains its first production caller as a useSyncExternalStore snapshot, putting an uncached full localStorage scan in the render path of exactly the prose-only guides this control targets** (medium · correctness-and-reliability)
   `getGuideProgress` had no production caller before this PR — at the merge base the only matches are its own definition and three prose comments. This PR makes it the `getSnapshot` of a `useSyncExternalStore`, so it runs on every render of the footer (twice per render in dev, and again on React's post-commit check). Its `total < 1` branch — the all-passive, prose-only case — calls `sectionAcknowledgementStorage.countAllAcknowledged` unconditionally, and that is a bare `for (let i = 0; i < localStorage.length; i++)` with a `getItem` per matching key and no cache. Worse, the result is discarded whenever the guide is marked: `marked` is read at line 437 but only consulted at 445 and 447, where the sibling writer `refreshGuidePercentage` short-circuits on `isMarked` before counting anything. A skeptic measured it with instrumented `Storage.prototype`: a marked all-passive guide returned `{completed:0,total:0,percentage:100}` after 6 `getItem` and 12 `key()` calls, and ack scans went from 5 after mount to 11 after three unrelated re-renders. The repo already treats this cost as unacceptable at a far lower frequency — `completedCountCache` exists expressly "to avoid O(n) localStorage scan on every step completion" — and R18 forbids storage access in a render path.
   Required: Hoist the `marked` check to the top of `getGuideProgress` and return `percentage: 100` before any counting, mirroring `refreshGuidePercentage`; or cache `countAllAcknowledged` the way `countAllCompleted` already is.

5. [blocking] **mark-footer-stale-key-snapshot — The footer resolves its content key inside getSnapshot, which runs before the layout effect that publishes that key, so a fresh milestone paints 100% complete** (medium · correctness-and-reliability)
   Both the `subscribe` and `getSnapshot` callbacks call `resolveGuideContentKey(contentUrl)`, and for any non-preview URL that discards `contentUrl` and returns `getContentKey()` — which reads the ambient `__DocsPluginActiveTabUrl`, published from a `useLayoutEffect` that runs after the whole tree has rendered. So the footer's first render for milestone N+1 asks about milestone N's key, and mixes it with the module-level step and section registries that already reflect the new guide. Because the reader has just marked milestone N, `isMarked` is true for that key and the fresh milestone renders "100% complete" — the most misleading readout possible at the moment the reader is looking at it. The component already applies the opposite reasoning to its other read: the comment at lines 90-93 says resolving the key during render "would latch the previous milestone's key", which is why the mark read lives in an effect. One skeptic reproduced the sequence (`m1=>100, m1=>100, m2=>0, m2=>0`); the other closed the obvious escape by confirming React registers the store's snapshot re-check as a passive effect, which runs after layout effects, so the correction is a separate commit and the stale value can paint.
   Required: Resolve the key once into state from an effect, as the mark read already does, and feed that resolved key to both `subscribeProgress` and `getGuideProgress`; or have `resolveGuideContentKey` prefer the passed `contentUrl` over the ambient global.

## Follow-ups

1. [follow_up] **mark-prefix-sweep-over-clears-siblings — resetPath sweeps completion marks by an unterminated content-key prefix, so resetting one path deletes the marks of any guide whose key merely starts with its URL** (medium · correctness-and-reliability · irreversible without cleanup)
   `resetPath` strips only the trailing slash from `path.url` and passes that to `clearAllWithPrefix`, which deletes every localStorage key satisfying `startsWith(prefix)` with no separator required after it. Proved by probe: sweeping the `drilldown-logs` path URL also cleared the mark on `drilldown-logs-advanced/step-1/`. The store's own cross-tab handler documents this exact hazard and defends against it by requiring an exact match, citing `bundled:loki-101` vs `bundled:loki-101-extended`; the new sweep does not. The sweep is also unnecessary — its comment justifies it as recovering a marked-but-unstepped milestone that has no interactive completion record, but the footer writes `interactiveCompletionStorage.set(contentKey, 100)` on every non-preview mark, so the narrower `clearMany(milestoneKeys)` already reaches it. Filed as a follow-up rather than a blocker because both skeptics established that the hazard is pre-existing at the identical site: at the merge base the same unterminated `normalizedUrl` already prefix-filters and clears a colliding sibling's step progress and percentages. It is also not reachable with shipped data — the prefix branch runs only for paths carrying a `url`, and no pair among the six shipped slugs collides, so it takes one newly added slug that extends an existing one.
   Follow-up: Require a boundary after a supplied prefix in `clearAllWithPrefix` (accept `key === prefix` or `key.startsWith(prefix + '/')`), which also fixes the two pre-existing prefix filters beside it; or drop the sweep for `clearMany(milestoneKeys)`. Add a prefix-sibling case to `reset-path-completion.test.ts`.

2. [follow_up] **contract-evolution-mark-evidence-branching — The guide-level mark ships as a boolean the guide-stats evidence model cannot consume, creating a second representation of mark-guide-complete with no converter and no parity guard** (medium · completion-records)
   The contract-evolution gate fired on this concern and the adapter returned `contract_branching`. The live owner of "what percentage is this guide at" is now the completion store, persisted in `interactiveCompletionStorage`; `guide-stats/progress.ts` is dead code at both base and head, with no production caller for `guideProgress` or `furthestEvidencedPosition`. So `mark-guide-complete` now has two representations — a live stored boolean and a declared-but-unfed `CompletionEvidenceKind` — with nothing converting one to the other and no test relating them, in contrast with `completion-affordance.parity.test.ts`, the existing ratchet on the adjacent seam. They already disagree on one computable case: a marked guide with no sections and no steps reads 100 from the store while `guideProgressAtPosition` at `total === 0` gives `complete: false, percent: 0`, unobservable only because the guide-stats owner is dead. The recorded anchor is not violated — it scopes the completion-store per-step model out explicitly, the PR computes no stamped denominator, the recorded `completionPercent` stays the pre-existing literal, the one new render predicate is `isJourneyCoverPage` rather than anything re-derived from `finalCompletablePosition`, and PR #1840 is not contradicted. But the convergence work is named in prose twice with no issue number anywhere, so nothing outside those two comments will surface the divergence later.
   Follow-up: File the derivation-convergence issue and reference it from `docs/design/COMPLETION-MODEL.md` and the two comments that name it, and state in the contract anchor which representation is authoritative until then.

3. [follow_up] **mark-continue-drops-focus — Mark complete and continue leaves keyboard focus on document.body after navigating, so the control's own focus-claim does not survive its one navigating form** (medium · cross-cutting-architecture)
   The control deliberately manages focus — the comment says the activated button "is about to unmount, and React would drop focus to document.body" — but the milestone form navigates, and the navigation unmounts the element focus was moved to. A skeptic traced `document.activeElement` with the button focused first, as a real click leaves it: the reduced-motion path is on the button at `onContinue` and on `BODY` after the navigation commit, and the celebration path is on the completed div at `onContinue` and on `BODY` after. So a keyboard or screen-reader user who activates "Mark complete and continue" lands on `document.body` with no announcement that the milestone changed, and no `.focus()` restoration exists anywhere in the docs-panel, content-renderer, floating-panel or full-screen surfaces. Filed as a follow-up rather than a blocker because both skeptics refuted the regression framing: the pre-existing next-milestone button is itself omitted from the loading branch and additionally becomes `disabled`, which browsers blur, so it loses focus the same way. The gap belongs to the panel, not to this PR — but this PR adds a second, more prominent entry into it.
   Follow-up: File a separate issue for milestone-navigation focus management covering both entry points: on milestone load, move focus to the new content heading or the panel content region and announce the change. At most, add a test here recording where focus lands after continue so the behaviour is pinned.

4. [follow_up] **mark-prose-overclaims-progress-wiring — Four passages this PR adds describe behaviour the same PR does not implement, including a docstring saying the mark feeds the guide-stats arithmetic** (low · cross-cutting-architecture)
   The new component's docstring says the click produces "`mark-guide-complete` evidence … which the arithmetic in `lib/guide-stats/progress.ts` already understands", and the PR body repeats it as "This is the producer." `COMPLETION-MODEL.md`, added in the same PR, states the opposite and is correct: nothing converts the stored mark into evidence, and `guideProgress` / `furthestEvidencedPosition` have no production caller. What is persisted is a boolean; `progress.ts` consumes a shape nothing in the tree constructs. Three more: that same doc names the type `CompletionSignal`, which does not exist — it is `CompletionEvidence`. The comment "Temporary until the completion store derives the percentage from the mark" is stale on arrival, because the store derives exactly that in this PR, making the write it guards duplication rather than a placeholder. And the test named "records a completion again after a reset re-arms the control" asserts calls to its own mock, not a recorded completion — the recorder dedupes per guide for the whole app load and no reset path clears that Set, which the PR body discloses as a known gap. No current behaviour harm; each one sends the next maintainer looking for code that is not there.
   Follow-up: Correct the four passages: say the mark is authoritative for the completion store's percentage and is not yet converted to guide-stats evidence; fix `CompletionSignal` to `CompletionEvidence`; drop or restate the "Temporary until" comment and the write it guards; and rename the re-arm test to what it asserts.

## Suggestions

1. [suggestion] **mark-continue-wiring-untested — The onContinueToNextMilestone wiring is untested on both surfaces that added it; only the branch producing an undefined callback is exercised** (low · testing-and-verification)
   The whole test change for the sidebar surface is `canNavigateNext: jest.fn(() => false)`, which covers only the branch yielding `onContinue === undefined`, and `FloatingPanelContent` adds the same wiring with no test change at all. The renderer test that checks the "Mark complete and continue" label passes `onContinueToNextMilestone` directly, so it cannot catch a mistake in the `model.canNavigateNext()` derivation. That derivation is the only part of this change deciding whether a click takes the reader anywhere, it is duplicated across two surfaces, and a regression in it is silent — the label degrades to "Mark complete" and the reader simply never advances. Worth noting by contrast: the re-arm state machine is genuinely well pinned, with cases for reset-then-re-mark, clear-then-automatic-then-click, click-then-automatic, and "no clear can re-complete the automatic route".
   Suggested: Add one case per surface with `canNavigateNext: () => true`, asserting the button reads "Mark complete and continue" and that clicking it calls `navigateToNextMilestone`, plus a last-milestone case asserting the label degrades.

## Nits

1. [nit] **mark-completed-live-region-redundant — role=status on the completed state has no case it helps and one it harms — a spurious Completed announcement on a return visit** (low · cross-cutting-architecture)
   The completed state is a `role="status"` div that is inserted at the moment its content appears, and is simultaneously given `tabIndex={-1}` and programmatic focus. Every path that flips `marked` true from a click also claims focus, so the announcement is already carried by the focus move and the live region adds a second one. The one path reaching `marked === true` without claiming focus is hydration of a pre-existing mark — exactly the case where nothing happened and nothing should be announced, and the only case where the live region does the announcing. A live region created together with its content is unreliable in the direction it is wanted and reliable in the direction it is not. The test titled "announces reaching complete exactly once" only counts `[role=status], [aria-live]` nodes, so it can observe neither the double announce nor the spurious one.
   Suggested: Drop `role="status"` and keep the focus move, which announces on exactly the path that warrants it. If a live region is wanted, pre-render an empty one that persists across both states rather than creating it with its content.

2. [nit] **mark-focus-effect-missing-dep-array — R13: the focus-claim useEffect has no dependency array, and no linter catches it** (low · cross-cutting-architecture)
   The focus-claim effect is written with no second argument, so it runs after every render of the footer. R13's agent behaviour is explicit: every `useEffect` must have a dependency array, even an empty one. ESLint over the changed frontend files is clean, confirming `react-hooks/exhaustive-deps` does not flag a missing array, so review is the only gate. It is behaviourally benign today — the `claimFocusRef` guard makes it idempotent and it is consumed on the first run after the click — but it is the one hook in the new file that opts out of the repo's hard rule, in the file most likely to be copied as the pattern for the next footer-style control.
   Suggested: Add `[marked]` — refs attach before effects, so `completedRef.current` is populated on the render where `marked` flips true.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1848
Summary: add the universal mark complete affordance
Verdict: Request Changes
Results: 5 blockers, 3 non-blocking findings, 4 follow-ups

Jayclifford345 (dismissed):

Jayclifford345 (approved):

Issue #1764, opened 2026-09-01
Promote the duplicate-title heading check from WARN to ERROR


moxious:
## Summary

`validateGuide` reports a warning when `blocks[0]` starts with a `# Heading` that duplicates the guide title. Make that one message an error.

Nothing else changes. The match rule stays exactly as it is today.

> [!IMPORTANT]
> This ticket can now be taken because grafana/interactive-tutorials#564 is now merged
> That PR removes the last 69 occurrences. If this change merges first, CI fails on 69 guides in that repository.

## Why

Pathfinder shows the guide title as an `<h1>` above the block content. Two render paths do this:

- A normal guide, at `src/components/content-renderer/content-renderer.tsx:552`
- A learning path cover page, at `src/components/LearningPaths/LearningPathTableOfContents.tsx:102`

No guide shape omits the metadata title. A leading `# Heading` in `blocks[0]` therefore always makes a second `<h1>` with the same text. The reader sees the title two times.

A warning was correct while the content still had 69 occurrences. PR #564 removes them. An error keeps the count at zero.

## What to change

The check is in `src/validation/validate-guide.ts` at lines 100 to 111. It pushes the message into an `advisories` array. A comment there states that strict mode never promotes this message.

Move the message into `errors` and delete that comment. The rule returns `isValid: false` for every caller, with or without `--strict`.

Do not change `headingDuplicatesTitle` at line 65. Do not change `extractLeadingH1` at line 50. The set of guides that trigger the message must stay the same.

## Tests

`src/validation/validate-guide.test.ts` has a `describe('leading heading duplicates title')` block near line 715.

- [ ] Update the 4 tests that expect `isValid: true` and `type === 'suggestion'`.
- [ ] Keep the 3 negative tests unchanged: an h2 heading, a non-markdown `blocks[0]`, and an unrelated heading. These must still pass with no error.

## Two consequences to expect

`block-import.ts:90` gates an import on `isValid`. After this change, the block editor refuses to import a guide with a duplicate heading.

Other repositories that use this CLI fail their CI on the day this change merges. `interactive-tutorials` has 0 occurrences after PR #564. I did not audit the other consumers.

## Sequencing

Pull requests are welcome. This change is small.

These numbers come from the CI run of PR #564, across 667 packages:

| | Before #564 | After #564 |
|---|---|---|
| Duplicate heading messages | 69 | **0** |
| Packages that pass `validate --package` | 667/667 | 667/667 |

[...]

PR #1849, opened 2026-09-09
fix(mcp): let help state what it requires, and what tokens it accepts

Jayclifford345:
## What

Two defects in what the authoring MCP says about itself. Both were found while authoring a real guide against the hosted MCP (`pathfinder-cli-mcp 1.1.0`, guide schema 1.1.0) with the [`create-interactive-guide` skill](https://github.com/grafana/ai-kit) — a pipeline that distils `pathfinder_help` and `pathfinder_get_schema` into the authoring vocabulary handed to LLM sub-agents as their only spec. Both make that distilled spec wrong in ways that produce invalid guides.

Runtime validation is untouched. This is only about the help output telling the truth about rules that already exist.

## Defect A — a required parameter that is in neither requiredness bucket

Measured against the hosted server:

```
pathfinder_help {command: "add-block"}                          → requiredByType.section   = ["id"]
                                                                  requiredByType.multistep = ["id", "content"]
pathfinder_help {command: "add-block", subcommand: "section"}   → required = ["type"]
                                                                  optional = ["ifAbsent","title","requirements",
                                                                              "objectives","autoCollapse","authorNote"]
```

`id` is in neither `required` nor `optional`, so a client that renders "the required fields and the optional fields" — the part that reads like the authoritative spec — produces a section field list with no `id` in it, and every section the author writes is missing one. `parent` and `branch` go missing the same way.

### Which source of truth is correct

**`requiredByType` is right, and the two views do not in fact disagree.** `id` *is* published for `section` — in `addressing`, carrying `required: true`. Both views come from the same `requiredNames(spec, view)` call over the same `specFields` walk, so they cannot report different facts. What they can do is *look* inconsistent, and that is the whole defect: `required` and `optional` are named for requiredness, `addressing` is named for role and overlaps both, and nothing in the response or the tool description tells a reader that the third bucket exists or why it matters.

Concretely, across the three layers:

| layer | `id` on a container | 
| --- | --- |
| `src/types/json-guide.schema.ts` (`SectionProps`) | **optional** — already-published guides predate the rule |
| `src/cli/commands/add-block.ts` (`structuralShape`) | **required**, `role: 'addressing'`, `missingCode: 'CONTAINER_REQUIRES_ID'` |
| `pathfinder_help` / `requiredByType` | **required** — both, correctly |

So nothing was made newly required, and nothing needed to be: the authoring command already required `id` for containers and the preflight already rejected its absence with `CONTAINER_REQUIRES_ID`.

### Fix

`requiredParams` on every rendered interface — every parameter the caller must supply, whatever bucket holds it. It is the same `requiredNames` call `requiredByType` uses, and for a group variant it is exactly what `commandArgViolations` will demand (discriminator first, then the variant's own obligations), so the published obligation and the enforced one are one list. Emitted even when empty, so an absent key can never be read as "this renderer does not report requiredness". The three buckets keep their existing shape — additive only.

The `pathfinder_help` tool description now says the bucket rule outright, and that `requiredByType` on a group root is the same list as each subcommand's own `requiredParams`.

## Defect B — the requirement vocabulary is not discoverable from the MCP at all

The `requirements` field arrived as:

> `Prerequisite conditions (e.g., on-page:/dashboards, is-admin) | valid tokens include is-admin, on-page:/dashboards`

Two examples, phrased as if they were the enumeration, and that is the whole of what a client ever saw. `pathfinder_get_schema {opts: {name: "guide", includeVersion: true}}` contains none of `exists-reftarget`, `has-datasource:`, `min-version:`, `plugin-enabled:`, `section-completed:` either — `RequirementTokenSchema` is a refined `z.string()`, so the tokens are invisible in the converted schema. And `agentView` deliberately withholds the CLI recipe (`pathfinder-cli requirements list`), correctly: an agent has no shell, and `requirements` is not a bound command, so there is no MCP route to it.

That leaves a step with a `reftarget` needing `exists-reftarget` in its `requirements` — which downstream house-rule linting treats as a hard error — with no way to learn the token exists.

Meanwhile the repo already owns the complete list in `src/types/requirements.types.ts`: `FixedRequirementType` (9), `ParameterizedRequirementPrefix` (15), and `REQUIREMENT_DESCRIPTIONS` mapping all 24 to a description.

### Fix

A new `REQUIREMENT_TOKEN_CATALOGUE`, **enumerated from the two enums** rather than written out, joining `REQUIREMENT_DESCRIPTIONS` and `PARAMETERIZED_REQUIREMENT_EXAMPLES` — so a new enum member reaches every surface without a second edit, and cannot drift. Published two ways:

- **`requirementTokens`** on a help response, whenever the rendered interface publishes a parameter that takes tokens *and* the reader has no way to print them for itself. That second condition is a new `SurfaceView.publishesRequirementVocabulary` — the other half of the existing `describe` trade-off. `agentView` opts in; `CLI_VIEW` keeps its pointer to `requirements list` and gets no copy, so `--help` stays short. Attached where it is actionable rather than on every command: `create` has no such parameter and gets nothing.
- **`x-requirement-tokens`** on the exported JSON Schema, alongside `x-refinements` / `x-schema-version`, for consumers that read only the schema. Which schemas get it is *walked* rather than declared per registry entry — a schema carries the vocabulary because it has a `requirements` / `conditions` property to put it in — so `guide` / `block` / `content` gaining or losing one needs no second edit.

The `requirements` description now points at `requirementTokens` instead of offering two tokens as the list.

Neither addition is free but both are small: the `add-block section` help goes 3.3 KB → 5.5 KB, the `guide` schema export 45.5 KB → 47.7 KB (well inside the existing recursive-inlining bound test).

## Tests

The most valuable part of the change — these are the tests that would have caught both defects.

**`src/cli/mcp/__tests__/command-interface.test.ts`** — parameterized over *every* `add-block` variant, so a new block type is covered on arrival:
- every name in `requiredByType[type]` appears in that type's own published field list;
- and carries `required: true` there;
- `requiredParams` for a variant equals `["type", ...requiredByType[type]]`;
- whatever `requiredParams` names, the preflight actually rejects the call for omitting — the published obligation and the enforced one held against each other;
- the concrete case: `section` publishes `id` as required, in `requiredParams` and in the flag list;
- every token in `REQUIREMENT_DESCRIPTIONS` appears in the advertised `requirementTokens`, and the counts match;
- `exists-reftarget`, `has-datasource:`, `min-version:`, `plugin-enabled:`, `section-completed:` named outright, so losing one reads as this defect returning rather than as a count changing;
- the description no longer offers its examples *as* the vocabulary (`/valid tokens (include|are)\b/`).

**`src/types/requirements.types.test.ts`** (new) — the catalogue covers every enum member and nothing else, classifies each by declaring enum, every entry has a non-blank description matching `REQUIREMENT_DESCRIPTIONS`, every published example passes `isValidRequirement`.

**`src/cli/__tests__/schema-command.test.ts`** — `x-requirement-tokens` complete on `guide` / `block` / `content`, every `REQUIREMENT_DESCRIPTIONS` token serializes into the guide export, absent from `repository` / `graph` / `e2e-report`.

**`src/cli/__tests__/output.test.ts`** — the help-shape stability contract gains `requiredParams`, asserts it holds every flag marked `required` regardless of bucket, and that the CLI reader gets no vocabulary copy.

## Verification

`npm run check` — all 9 steps green (typecheck, eslint, prettier, docs:sync-terms:check, lint:go, test:go, test:coverage, test:review-contract, test:scripts).


moxious (changes requested):
## Merge contract

1. **[blocking] PR1849-F1 — pathfinder_help and AGENT-AUTHORING.md claim `requiredByType` equals a subcommand's own `requiredParams`, but they differ by the leading discriminator** (medium · mcp-authoring-server)
   `help.ts`'s tool description and `AGENT-AUTHORING.md` both state, unqualified, that the group root's `requiredByType` entry for a variant is "the same list" as that subcommand's own `requiredParams`. In the shipped code, `renderGroupInterface` (src/cli/contracts/group.ts) prepends the discriminator (e.g. `type`) to a variant's `requiredParams`, while `requiredByVariant` computes `requiredByType[variant]` from the bare variant schema with no discriminator — so the two lists differ by exactly one element, and the PR's own test (`src/cli/mcp/__tests__/command-interface.test.ts:333-338`) encodes that inequality directly (`['type', ...requiredByType[type]]`). An agent or engineer who takes the published claim literally and substitutes `requiredByType[variant]` for that variant's `requiredParams` builds a per-call obligation list missing the discriminator — the exact class of authoring defect this PR exists to eliminate. Verified independently by two skeptic reviewers (both confirmed, no dissent).
   **Required:** Reword both descriptions to state the actual relationship (variant `requiredParams` = discriminator followed by `requiredByType`'s list for that variant) instead of asserting the two lists are the same.

## Follow-ups

_(tracked separately, do not block merge)_

2. **[follow_up] PR1849-F2 — `x-requirement-tokens` presence/absence tests don't cover the `manifest` or `e2e-multi-report` schema registry entries** (medium · testing-and-verification)
   The new `x-requirement-tokens` test block in `schema-command.test.ts` only asserts presence for guide/block/content and absence for repository/graph/e2e-report, leaving `manifest` and `e2e-multi-report` — two of the eight `SCHEMA_REGISTRY` entries — unasserted by either table, even though neither schema declares a `requirements`/`conditions` field today. A future change that accidentally attaches (or fails to attach) `x-requirement-tokens` on either entry would ship undetected.
   **Follow-up:** add `manifest` and `e2e-multi-report` to the absence-test table, or add one registry-wide test asserting every `SCHEMA_REGISTRY` name is covered.

3. **[follow_up] PR1849-F3 — Parameterized-example fallback string isn't actually guarded by a failing test, contrary to the adjacent code comment** (medium · correctness-and-reliability)
   The comment above `REQUIREMENT_TOKEN_CATALOGUE`'s fallback (`${token}<value>`) claims a missing `PARAMETERIZED_REQUIREMENT_EXAMPLES` entry is "a gap `requirements.types.test.ts` fails on." In fact the validator's parameterized check is a bare `startsWith(prefix)`, which the fallback trivially satisfies against its own token, so every existing test passes on the placeholder. Every prefix has a real example today, so this is dormant — but the next contributor who adds a prefix without an example will silently ship a literal placeholder into the published vocabulary with green tests and a comment saying it can't happen.
   **Follow-up:** make a missing example a hard error, or add a test asserting every prefix has one.

## Suggestions

4. **[suggestion] PR1849-F4 — `declaresRequirementTokens`/`carriesRequirementTokens` key on property name only, not the field's actual schema type** (low · correctness-and-reliability)
   The check is a pure name match (`'requirements'`/`'conditions'`) with no relation to the field's actual zod type. Correct for every field today; a future schema with an unrelated same-named field would silently inherit the vocabulary.
   **Suggested:** key on the field's zod type/brand instead, or pin the heuristic's invariant with a test.

**Verdict: Request Changes** — 1 blocker, 1 non-blocking suggestion, 2 follow-ups.

---

### Non-issues / explicitly dismissed (not carried as findings)

- **Contract-evolution check** (mcp-authoring-server anchor, evidence #1639): verdict `coherent_extension` — `requiredParams`/`requirementTokens` are genuinely lock-step with the MCP preflight and with `requiredByType` by construction (both derive from the same `requiredNames()`/`REQUIREMENT_TOKEN_CATALOGUE`), no competing owner exists, and CLI `--help --format json` picks up the new field for free through the shared renderer. One disclosure (not a blocking finding, since the adapter returns no observation for `coherent_extension`): this PR updates `docs/design/AGENT-AUTHORING.md` but not the `mcp-authoring-server` anchor text in `docs/design/CONCERN_DETAILS.md` to mention the new fields — worth a follow-up edit per this repo's "extending existing capabilities" policy, but not required to merge.
- **Reversibility-and-one-way-door**: reviewed clean. The requirement-token vocabulary was already effectively public (CLI's `requirements list`, existing docs) before this PR; the new fields are additive/optional on `HelpJson` and documented in the same PR.
- **Security**: reviewed clean. No new secrets/paths/internal detail exposed; `SCHEMA_REGISTRY` is a fixed internal registry, not attacker-influenced, so the new recursive schema walk has no DoS surface.
- **Cross-cutting-architecture**: reviewed clean. The new `SurfaceView.publishesRequirementVocabulary` flag is consistent with the existing per-reader-hook pattern (`describe`, `publishes`, `spell`) rather than a competing mechanism; the catalogue's source enums are provably complete by construction (`Object.values()`, not hand-maintained).
- **ai-subsystem**: reviewed clean — no findings beyond the F1 defect already captured under mcp-authoring-server.
- **cli-and-e2e-runner**: reviewed clean — no CLI exit-code or command-semantics changes; only additive JSON keys.
- Two structural non-findings verified but not flagged: `renderGroupInterface`'s root/variant paths can't duplicate the discriminator (`defineCommandGroup` throws at load time if a variant schema declares it), and the recursive/`$ref`-based JSON Schema shape doesn't cause `declaresRequirementTokens` to miss nested occurrences.

### Method note

Routed as `mixed` (contracts-and-schemas + docs + tests). Full first round: 3 general observation workers (mcp-authoring-server+ai-subsystem; cli-and-e2e-runner+testing+correctness; reversibility+security+cross-cutting) plus 1 gated contract-evolution specialist, then 4 independent skeptics (2 for the provisionally-blocking F1, 1 each for F2/F3 — all confirmed, no refutations, no tiebreakers needed).

Jayclifford345:
Addressed the blocker and one of the two follow-ups in 79ed9f3.

## PR1849-F1 — fixed (blocker)

The claim was an equality; the relationship is off-by-the-discriminator. `renderGroupInterface` prepends the discriminator to a variant's `requiredParams`, `requiredByVariant` computes `requiredByType[variant]` from the bare variant schema, and this PR's own test already asserts the difference (`['type', ...requiredByType[type]]`). Reworded to state the relationship in the three places that carried the equality:

- **`src/cli/mcp/tools/help.ts`** — "a subcommand's own `requiredParams` is the parameter that selects it followed by the root response's `requiredByType` entry for it, so `requiredByType` alone is that list minus the selector." The trailing clause names the misuse, rather than leaving the reader to derive that substituting one for the other drops a name.
- **`docs/design/AGENT-AUTHORING.md`** — same relationship, plus "`requiredByType` is that list without the discriminator, so it is the per-variant summary and not a substitute for the variant's own `requiredParams`."
- **`src/cli/utils/output.ts`** — `HelpJson.requiredParams`'s own doc comment carried the identical false claim ("exactly what `requiredByType` reports for the same variant"). Not named in the finding, but it is the same sentence on the field being documented, so leaving it would have preserved the defect at its most authoritative site.

The `group.ts` comment that already stated it correctly ("discriminator plus the variant's own obligations") is unchanged.

## PR1849-F3 — fixed

You were right that this was unguarded, and the proof is sharp: temporarily deleting the `min-version:` entry from `PARAMETERIZED_REQUIREMENT_EXAMPLES` on the pre-fix tree left **all 42 assertions in `requirements.types.test.ts` passing** on the `min-version:<value>` placeholder — non-blank, so the description/example check passed, and `startsWith` its own prefix, so `isValidRequirement` passed. With the new test in place the same deletion fails exactly one case:

```
● REQUIREMENT_TOKEN_CATALOGUE › has an authored example for min-version:
    Expected value: "min-version:"
    Received array: ["on-page:", "has-datasource:", ...]
Tests: 1 failed, 42 passed, 43 total
```

Restored, 43/43 green.

I went with the test rather than the hard error. `src/types/requirements.types.ts` is Tier 0 and is imported by the browser bundle (the chip picker reaches it through `ConditionChipsField`), so a load-time throw would take the plugin down for what is a source-constant mistake — and one that cannot escape CI, since the catalogue is built from constants in the same file and every suite importing the module would go red. A named test failure is the same tripwire with the blast radius of a mistake kept in CI. Happy to switch it to a throw if you'd rather.

Both comments now describe what actually happens: the catalogue's doc comment distinguishes the blank-description case from the placeholder-example case instead of lumping them, and names the fallback as a placeholder that "reads as an example and is not one".

## Deliberately not fixed here

- **PR1849-F2** (`x-requirement-tokens` coverage for the `manifest` / `e2e-multi-report` registry entries) — respecting your triage as a separate item. It is a test-coverage gap on registry entries this PR does not otherwise touch, and widening into the registry would put it past the single-concern line.
- **PR1849-F4** (`declaresRequirementTokens` / `carriesRequirementTokens` keying on property name rather than zod type) — likewise. That one is a change to the walk's detection rule, not to a published claim, so it is a different defect class from this PR's subject and deserves its own review.

## On `CONCERN_DETAILS.md`

I made the edit. The anchor column already uses a `#A → #B` chain for later extensions (`#1372 → #1821`, `#1662 → #1791 → #1764 → #1640`), so the in-form edit was `#1639` → `#1639 → #1849` plus the contract sentence, which is what I did. The added prose states the two published-interface rules this PR establishes — `requiredParams` as the single statement of requiredness whatever bucket holds a parameter (worded with the corrected F1 relationship, so the anchor does not reintroduce the claim being fixed above), and the vocabulary attaching to a surface that publishes a `requirements` / `conditions` parameter and cannot print it.

Per the extending-existing-capabilities policy: **this change extends the #1639 contract, it does not replace it.** Nothing in the existing anchor became untrue — the three buckets keep their shape and `validateCommandArgs` still preflights against the exact interface `pathfinder_help` publishes; the additions are new published fields.

## Gate

`npm run check` — all 9 steps passed.

| step | result |
| --- | --- |
| typecheck, lint, prettier, docs:sync-terms:check | pass |
| lint:go, test:go | pass (`pkg/plugin`, `pkg/plugin/auth` ok) |
| test:coverage | 536 suites, 8710 passed / 18 skipped / 5 todo, 8733 total |
| test:review-contract | 76 pass, 0 fail |
| test:scripts | 64 passed, 0 failed |

No version bumps, no unrelated changes: 6 files, +20 / −10.

moxious (approved):

Follow-ups:

1. [follow_up] **PR1849-F2 — Cover the two remaining schema exports** (medium · testing-and-verification)
   `src/cli/__tests__/schema-command.test.ts:98` and `:107` cover six schema names, leaving `manifest` and `e2e-multi-report` without assertions for `x-requirement-tokens`. Both currently omit the vocabulary correctly; a change to either export would escape these assertions.
   Follow-up: Add `manifest` and `e2e-multi-report` to the existing absence table.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1849
Summary: let help state what it requires, and what tokens it accepts
Verdict: Approve with Minor

PR #1852, opened 2026-09-09
fix(storage): evict stale completion entries, not the ones in use

tomglenn:
## What changed

Completion percentages live in one bounded record keyed by content key, trimmed to a cap when it overflows. The trim ran on **key insertion order**, and updating an existing key did not move it. So the first guide you ever opened was always the first one evicted — including when you had just earned progress on it, and even when the entry being kept was one you opened once and abandoned. Exactly backwards.

Because a missing key reads back as `0`, the loss was invisible. An evicted guide is indistinguishable from one you never opened, on every surface. And a path's percentage is the mean of its milestones' percentages, so a single evicted milestone quietly drags the whole path number down.

Two changes in `src/lib/storage/bounded-record-storage.ts`:

- `set` deletes the key before re-adding it, so the record's key order *is* write recency and the existing `slice(-limit)` trim becomes least-recently-written-first. Earning progress on a guide now protects it.
- Before falling back to recency, the trim drops zero-progress entries. Dropping a `0` is free: `get()` already returns `0` for a missing key, so nothing is lost.

The interactive completion cap goes from 100 to 250 (`src/lib/user-storage.ts`). Journey completions stay at 100.

## Why this shape

Three fixes were on the table: raise the cap, evict by last update instead of first insert, or persist a separate set of opened guides so eviction could at least be told apart from never-opened.

The third was rejected outright — it makes the loss legible but still loses the progress.

**A cap raise alone would not have been enough**: it moves the same bug to a bigger number. Whatever the limit, the guide you use most is still the first one thrown away. That is why the raise rides on top of the ordering fix rather than instead of it, and why it is modest. What justifies raising it at all is that 100 slots is nowhere near 100 guides: this record is keyed per guide **and per milestone**, so one ten-milestone path consumes ten slots, and a bundled guide opened from two surfaces occupies two. A reader with a legitimately small library can reach the cap. At 250, worst case is 250 keys against the existing 200-character key limit, about 52KB, and realistically nearer 20KB — well inside localStorage, with the existing `QuotaExceededError` retry as the backstop.

The persisted shape is unchanged — still `{ contentKey: percentage }` — so existing records are read as-is and there is nothing to migrate.

## What to look at

- `trimToLimit` in `bounded-record-storage.ts` — the eviction order.
- The `delete data[key]` before the re-add in `setInternal`. It looks redundant and is not: it is what makes key order mean write recency.
- The record is shared with journey completion, which gets the same ordering fix through the same building block.
- The `cleanup()` docstring. #1843 landed wording asserting insertion-order eviction, which this change makes false; it is rewritten to describe the new order rather than spliced.

One limitation worth your judgement: a guide at 100% stops being written, so over time it drifts toward the front of the eviction queue. The cap raise softens that but does not cure it.

## How to verify by hand

The record is plain JSON in localStorage, so you can put a reader past the cap without clicking through 250 guides. In devtools, on a Grafana instance with the plugin loaded:

```js
const KEY = 'grafana-pathfinder-app-interactive-completion';
const seeded = {};
for (let i = 0; i < 250; i++) { seeded[`filler-${i}`] = 10; }
localStorage.setItem(KEY, JSON.stringify(seeded));
```

Then reload, open **one guide** in the sidebar and complete a step or two in it, and finally open a **second, different** guide and complete a step there — the second one is what pushes the record over budget.

Re-read the record and look for the first guide you worked in:

```js
JSON.parse(localStorage.getItem(KEY));
```

- **Before this change**: the guide you just earned progress in is gone, and the sidebar shows it at 0% as if you had never opened it.
- **After this change**: both guides you actually used are present with their real percentages, and a `filler-*` entry was dropped instead.

Below the cap nothing changes — a reader with fewer than 250 guides sees identical behaviour, and there is a test pinning that.

## Correction: one guide really can occupy two slots

An earlier version of this description claimed that nothing reads two bundled key shapes for one guide's percentage. **That was wrong**, and it predates this change rather than being introduced by it.

`bundled:<id>` and `bundled:<id>/content.json` are both live launch shapes for the same guide: My Learning opens it bare, the package resolver opens the `/content.json` form. Each is read on its own — `context.service.ts:844` reads `rec.url`, `context.service.ts:894` reads the resolver's `contentUrl` — so the same guide is asked about under either shape with nothing pairing them, and each shape that gets written occupies its own slot. `pathMemberContentKeys` (`src/global-state/path-member-join.ts:206`) does pair them, for the path join. I checked the `bundled:`/`backend-guide:` pairing and `guideMetadata.url`, never the resolver's shape, so I reported a real multiplier as absent.

**It does not change the cap.** 250 still holds, because the doubling is bounded by the bundled catalogue — seven guides — so it adds at most seven slots rather than halving the effective budget. The dominant multiplier is still per-milestone keying, where one ten-milestone path consumes ten slots, and that is what 250 is sized against.

**It does change which entry gets evicted, and that is worth a reviewer's attention.** Storage sees the two shapes as unrelated keys, so it keeps whichever was written more recently. The join prefers the *higher* of the pair. Those disagree: a reader who reaches 80% under `bundled:<id>/content.json` and later does one step under bare `bundled:<id>` has a fresh 10% and a stale 80%, and the join reports 80% until the 80% entry ages out, then 10%.

Two things bound that. The zero-first pass covers the common case, where the twin holds no progress and is dropped for free. And it is not a regression from this change: under insertion order the survivor was whichever happened to be written first, unrelated to either recency or value, so the pair was split arbitrarily. Closing it properly needs the two keys to be known as a pair, which is true in `path-member-join.ts` and not in storage — storage has no notion of pairs to reason about. Written down in `docs/design/COMPLETION-MODEL.md` in this PR rather than fixed here.

## Scope

Storage only. Percentage calculation, the completion recorder, and the progress displays are untouched — a separate consolidated change owns those.

## Reversibility

Straightforward revert. The persisted shape is unchanged, so a revert reads existing records fine; a record grown past 100 entries under the higher cap simply gets trimmed back on its next write.

### DIFF
--- a/docs/design/COMPLETION-MODEL.md
+++ b/docs/design/COMPLETION-MODEL.md
@@ -388,19 +388,27 @@ candidate set it appears in and inflate the path's number.
 treats a formable-but-absent key as never-opened and scores it zero, which is
 the honest answer for a reader who has not started that guide. It is _not_
 honest for a reader whose record was evicted: `interactiveCompletionStorage`
-caps at `MAX_INTERACTIVE_COMPLETIONS` (100, `src/lib/user-storage.ts`) and
-`writeWithCap` keeps `entries.slice(-limit)` — insertion order, so the earliest
-keys are dropped and updating an existing key does not move it forward. A reader
-past 100 distinct guides therefore loses real progress on their oldest ones, and
-it arrives at the join as a genuine zero. Reading two candidate shapes per
-bundled member raises the pressure slightly, because a guide opened from both
-surfaces occupies two of the 100 slots.
+caps at `MAX_INTERACTIVE_COMPLETIONS` (250, `src/lib/user-storage.ts`), and past
+that cap `writeWithCap` drops zero-progress entries first and then the least
+recently written of the rest. Reading two candidate shapes per bundled member
+raises the pressure, because a guide opened from both surfaces occupies two
+slots.
 
 The information needed to tell eviction from never-opened is gone by the time
-the join runs, so this is not fixable in `path-member-join.ts`. **Known
-follow-on work, storage-side:** raise the cap, evict least-recently-updated
-rather than earliest-inserted, or persist an opened-guides set the join can
-consult. Until one of those lands, a path mean can understate a heavy reader.
+the join runs, so this is not fixable in `path-member-join.ts`. Eviction now
+falls on the entries a reader has stopped touching rather than the ones they
+opened first, which is what closes the case where a heavy reader lost progress
+on a guide they were actively using.
+
+**Recency and furthest-wins disagree on a split pair.** Storage sees two launch
+shapes of one guide as two unrelated keys, so it cannot prefer the pair's higher
+record — it keeps whichever was written more recently. The join prefers the
+higher. A reader who reaches 80% under `bundled:<id>/content.json` and later
+does one step under bare `bundled:<id>` has a fresh 10% and a stale 80%; the
+join reports 80% until the 80% entry ages out, and 10% after. The zero-first
+pass covers the common case, where the twin holds no progress at all and is
+dropped losslessly. Closing the rest needs the pair to be known where the keys
+are paired, which is here and not in storage.
 
 **The key spaces `resetPath` clears are not one space.**
 `interactiveCompletionStorage` and `interactiveStepStorage` are keyed by the
--- a/src/lib/storage/bounded-record-storage.test.ts
+++ b/src/lib/storage/bounded-record-storage.test.ts
@@ -85,6 +85,59 @@ describe('createBoundedRecordStorage', () => {
     expect(await store.getAll()).toEqual({ b: 2, c: 3, d: 4 });
   });
 
+  describe('eviction order', () => {
+    it('keeps an entry the reader is still updating and evicts an untouched one instead', async () => {
+      const store = makeStore({ storageKey: TEST_KEY, limit: 3, label: 'test' });
+      await store.set('early', 10);
+      await store.set('b', 20);
+      await store.set('c', 30);
+
+      // The reader comes back to `early` and earns more progress on it.
+      await store.set('early', 60);
+
+      // A fourth guide pushes the record over budget.
+      await store.set('d', 40);
+
+      expect(await store.get('early')).toBe(60);
+      expect(await store.getAll()).toEqual({ c: 30, early: 60, d: 40 });
+    });
+
+    it('evicts entries with no recorded progress before entries that hold progress', async () => {
+      const store = makeStore({ storageKey: TEST_KEY, limit: 3, label: 'test' });
+      await store.set('has-progress', 20);
+      await store.set('c', 30);
+      await store.set('opened-only', 0);
+      await store.set('d', 40);
+
+      // Dropping a 0 is lossless: `get()` already returns 0 for a missing key.
+      expect(await store.getAll()).toEqual({ 'has-progress': 20, c: 30, d: 40 });
+    });
+
+    it('reads a record written before the change and keeps updated entries from it', async () => {
+      // Shape written by the previous implementation: a plain key -> percentage
+      // record with no recency metadata, already over the new budget.
+      localStorage.setItem(TEST_KEY, JSON.stringify({ old1: 10, old2: 20, old3: 30, old4: 40 }));
+      const store = makeStore({ storageKey: TEST_KEY, limit: 3, label: 'test' });
+
+      expect(await store.get('old1')).toBe(10);
+
+      await store.set('old1', 55);
+
+      expect(await store.get('old1')).toBe(55);
+      expect(await store.getAll()).toEqual({ old3: 30, old4: 40, old1: 55 });
+    });
+
+    it('changes nothing for a record below the limit', async () => {
+      const store = makeStore({ storageKey: TEST_KEY, limit: 100, label: 'test' });
+      await store.set('a', 0);
+      await store.set('b', 20);
+      await store.set('a', 30);
+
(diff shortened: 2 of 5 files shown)


moxious (dismissed):

Follow-ups:

1. [follow_up] **F1 — The accepted limitation the PR says is recorded in COMPLETION-MODEL.md is recorded nowhere, and two symbols it cites do not exist** (medium · cross-cutting-architecture)
   The Correction section defers the bundled twin-key hazard with "Recorded in COMPLETION-MODEL.md rather than fixed here", and names pathMemberContentKeys in path-member-join.ts as the place where the pairing is known. docs/design/COMPLETION-MODEL.md exists at both base and head but is untouched by this PR, and a concept-level search of it finds no prose about the bounded record, its cap, or its eviction order; no other document records it either. path-member-join.ts and pathMemberContentKeys exist at neither SHA, and nothing else in src pairs bundled:<id> with bundled:<id>/content.json or prefers the higher of the two. The two-key read shape itself is real - context.service.ts:844 and :894 read resolver-shaped URLs while every write goes through a getContentKey()-sanitized key - so the observation is sound; only its stated record and the named code are missing. Net effect: the reviewer is asked to accept a known limitation on the strength of a record that does not exist, and the only written trace of it is the PR body itself.
   Follow-up: Add the limitation to docs/design/COMPLETION-MODEL.md in this PR (bounded record, cap, eviction order, and the bundled:<id> vs bundled:<id>/content.json pair), or correct the PR body to drop the three references and name the real anchor - context.service.ts:844,894 as the unpaired reader.

Suggestions & nits:

1. [suggestion] **S1 — The new eviction semantics are asserted order-insensitively, and the just-written-zero case is untested** (low · testing-and-verification)
   Write recency is carried solely by the record's key enumeration order, but every new eviction assertion uses toEqual on an object, which ignores key order - including the one that reaches into raw localStorage. A future trimToLimit that re-sorted its output (alphabetically, say) would satisfy all five new cases while destroying recency for the next eviction. Separately, the zero-first pass does not exclude the key setInternal just wrote, so at cap set(newKey, 0) completes successfully and leaves the record unchanged; with limit 3 and seed {a:10,b:20,c:30}, set('brand-new',0) yields {a:10,b:20,c:30}. That is a live path - completion-store.ts:303,308 writes Math.round(allCompleted/docTotal*100), which is 0 for a freshly opened guide - and it inverts the surviving key set relative to base, but no case covers it. The five new cases do discriminate the change: run against the base implementation, four of five fail.
   Suggested: Assert Object.keys(JSON.parse(localStorage.getItem(TEST_KEY))) after an overflow trim so the recency order is pinned rather than incidentally satisfied, and add one case asserting that at cap set(newKey, 0) leaves the record unchanged while get(newKey) still returns 0.

2. [suggestion] **S2 — The new set docstring's losslessness claim is not true of the one consumer that reads key presence** (low · reversibility-and-one-way-door)
   The docstring justifies the zero-first pass with "a missing key reads back as 0, so dropping one loses nothing". That holds for every value reader, but resetPath recovers milestone content keys purely by presence in interactiveCompletionStorage.getAll() and is the only driver of interactiveStepStorage.clearAllForContent, which in turn is the only content-scoped remover of SECTION_ACKNOWLEDGED and SECTION_DONE keys. A mixed guide can store 0 beside durable section-done state: completing an all-passive section sets sectionDoneStorage and then takes the docTotal >= 1 branch, whose numerator is countAllCompleted alone. Three independent verification passes agreed the sentence is narrowly false while refuting the harm claim built on it - net eviction incidence goes down under this PR, because the cap raise removes eviction across the whole 100-250 band and, above 250, a zero victim strands strictly less residue than base's oldest-key victim. So this is a wording problem, not a behaviour problem.
   Suggested: Qualify the sentence: dropping a 0 is lossless for value readers, but resetPath reads key presence, so an evicted milestone keeps its ack and section-done bits through a path reset. Making that recovery structural rather than presence-based is separate, pre-existing work worth its own issue.

3. [nit] **N1 — The changed cap line keeps a restating trailing comment beside the new rationale** (low · correctness-and-reliability)
   MAX_INTERACTIVE_COMPLETIONS: 250 carries // Prevent quota exhaustion directly under the new two-line comment explaining why this cap diverges from the journey cap. The two-line comment earns its place and its per-milestone claim checks out; the trailing one adds nothing the constant name and that comment do not already say, and it sits on a line this PR touches.
   Suggested: Drop the trailing // Prevent quota exhaustion on the MAX_INTERACTIVE_COMPLETIONS line only. Leave the untouched sibling lines alone - trim on touch, not a sweep.

4. [nit] **N2 — Recency ordering is only true for keys written after the change, and the journey store has no cap headroom to absorb that** (low · reversibility-and-one-way-door)
   Key order is reinterpreted from insertion order to write recency while the persisted shape stays the same, so a pre-upgrade record's untouched keys are a stale proxy and the first post-upgrade eviction can still discard an actively used entry. For the interactive record the 100 to 250 raise gives roughly 150 writes of headroom in which that window is absorbed. journeyCompletionStorage is built from the same factory and inherits both changes, but its cap stays at 100, so a user already at 100 journey entries hits their first post-upgrade eviction on the very next write, decided entirely by legacy order. The PR body discloses the sharing and the unchanged journey cap; it does not mention this asymmetry. Outcomes are no worse than base and self-heal as writes accumulate, and no migration is needed in either direction.
   Suggested: Optionally add one clause to the set docstring noting that recency ordering only becomes accurate for keys written after this change, so a record already at its cap gets one eviction round decided by legacy order.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1852
Summary: evict stale completion entries, not the ones in use
Verdict: Approve with Minor
Results: 0 blockers, 4 non-blocking findings, 1 follow-up

moxious (approved):

Issue #1448, opened 2026-07-29
Browser Back from a transient prose full-screen launch reopens a closed sidebar


moxious:
**Summary**

PR #1446 launches reading-only guides from My Learning in _transient_ full-screen mode: the surface choice is held in memory and never overwrites the user's persisted panel-mode preference (enforced in `src/global-state/panel-mode.ts`, pinned by `panel-mode.test.ts`). The preference half of the invariant holds. What does not hold is the sidebar's actual open/closed state on the return path.

When anything navigates off `/fullscreen` — including browser Back — `FullScreenPanel`'s history listener (`src/components/full-screen/FullScreenPanel.tsx:212-215`) invokes the generic auto-dock, `dockOnLeavingFullScreen` (`src/components/full-screen/full-screen-autodock.ts:46-100`), which unconditionally switches to sidebar mode and calls `sidebarState.openSidebar`. That behavior predates #1446 and is correct for its original case (an interactive `navigate` step leaving full screen mid-guide needs a panel to continue in). The prose-only My Learning entry path added by #1446 is what exposes the mismatch: content we _classified as a poor sidebar fit_ gets reopened in the sidebar on the most common return gesture, in a sidebar that may have been closed before the launch.

Discovered in the #1446 review (finding 6) and deliberately split out of that PR after risk analysis: the fix requires a primitive that does not exist yet — a way to end the transient session _without_ persisting (today the round-trip ends only via `setModePersisted` or reload; see `_transientActive` at `panel-mode.ts:82,119-156`) — plus threading history action and prior-sidebar-open state into the auto-dock decision. `full-screen-autodock.ts` carries a documented `setTimeout(0)` deferral protecting step-completion writes (`full-screen-autodock.ts:63-72`), so changes there need their own careful regression tests rather than riding at the tail of an already-large PR.

**Use cases / user flows impacted**

1. _Read and go back._ A user with the sidebar closed opens My Learning and clicks a prose guide. It opens full screen (correct). They finish reading and press browser Back, expecting to land on My Learning as they left it. Instead the extension sidebar pops open with the same prose guide squeezed into it — the exact rendering the feature exists to avoid — and the user now has to close a sidebar they never opened.
[...]

PR #1858, opened 2026-09-09
fix(full-screen): quiet exit on transient navigation

### LINKED ISSUE #1472: Grafana nav click (history PUSH) from a transient prose full-screen launch still docks the sidebar
## Summary

Split out of #1448 by the settled product decision (Q2: browser Back / history **POP** only). #1448 quiets the return path when the user presses **browser Back** out of a transient prose full-screen launch. It deliberately leaves **all PUSH paths** on today's auto-dock behavior — including a **Grafana mega-menu / nav click** made while a transient prose guide is up, which still switches to sidebar mode and reopens the extension sidebar with the prose squeezed in.

## Why it was deferred, not folded in

The Back fix keys off a reliable, already-available discriminator: the history `action` (`POP` vs `PUSH`). A nav click is a `PUSH`, indistinguishable at the auto-dock from the interactive `navigate`-step continuation that **must** keep docking (the user needs a panel beside Grafana to finish the step). Telling those two PUSH cases apart needs a signal that does not exist yet: **"an interactive step is in progress."** Building that signal is the actual work here, and it overlaps the auto-open-listener ownership work in #1450.

## Scope

- Introduce an "interactive-step-in-progress" (or equivalent "this fullscreen session is prose-only, not mid-interactive-step") signal the auto-dock can consult.
- On a PUSH off `/fullscreen` during a transient **prose** session, quiet-exit like the Back path (reuse `panelModeManager.endTransientSession()` added in #1448) instead of docking.
- Preserve the interactive `navigate`-step continuation unchanged: an in-progress interactive step leaving full screen must still auto-dock.

## References

- #1448 (browser Back quiet exit) — added `panelModeManager.isTransient()` / `endTransientSession()` and threaded the history `action` into `full-screen-autodock.ts`.
- #1450 / #1460 — auto-open listener ownership across surfaces (adjacent signal work).
- `docs/design/PANEL-MODE-PERSISTENCE.md` — persistence contract; `#1448` is referenced under "Related".

---
_Filed as the deferred follow-up from PR #1471 (fix for #1448)._

dvd233:
## Summary

Leaving a transient prose full-screen session through ordinary Grafana navigation now ends the transient session without forcing the extension sidebar open. Interactive navigation still hands off to the sidebar before its PUSH, and the requirement-fix path now uses that same handoff instead of pushing directly.

## What to look at

- [Auto-dock decision](https://github.com/dvd233/grafana-pathfinder-app/blob/6d2960524fcbc6737af601e657d1d3aac16e35a4/src/components/full-screen/full-screen-autodock.ts) classifies transient POP as `transient_back`, transient PUSH as `transient_navigation`, and leaves REPLACE/non-transient navigation on the existing docking path.
- [Requirement-fix handoff](https://github.com/dvd233/grafana-pathfinder-app/blob/6d2960524fcbc6737af601e657d1d3aac16e35a4/src/interactive-engine/navigation-manager.ts) routes the one remaining direct interactive PUSH through `requestSidebarHandoffAndWait` while preserving the normal direct push outside full screen.
- [Persistence contract](https://github.com/dvd233/grafana-pathfinder-app/blob/6d2960524fcbc6737af601e657d1d3aac16e35a4/docs/design/PANEL-MODE-PERSISTENCE.md) records why the pre-PUSH mode change is the discriminator and why REPLACE still docks.

## Why

History reports both an ordinary Grafana nav click and an interactive navigation as PUSH. The existing click-time handoff changes panel mode before a genuine Grafana-driving step pushes, so Guard 1 removes it from auto-dock; a transient PUSH that still reaches auto-dock has no interactive continuation. `NavigationManager.fixLocationRequirement` was the only direct requirement-fix PUSH and is now routed through the same facade.

## How to verify

1. Open a prose guide through the transient full-screen launch, click a Grafana nav or guide link, and confirm the session exits without reopening the extension sidebar.
2. Trigger an interactive `navigate` action from full screen; confirm the sidebar handoff occurs before navigation and the step can continue at its target.
3. Trigger a location-requirement fix inside and outside full screen; confirm the former hands off with `targetPath` while the latter still calls `locationService.push` directly.
4. Confirm transient browser Back still reports `transient_back`, while REPLACE and non-transient navigation keep the existing docking behavior.

## Follow-up

This follow-up extends, rather than replaces, the established full-screen-sidebar-handoff contract. It adds NavigationManager.fixLocationRequirement as a caller, validates on-page: targets before either navigation branch, preserves query/fragment data, and returns a failed fix when validation refuses a target so the retry affordance remains available. The contract registry and persistence documentation now route and describe this caller; a real panel-mode/auto-dock integration test covers sidebar survival.

## Breaking changes

None. `transient_navigation` is an additive internal outcome and analytics reason; persisted panel-mode state and external contracts are unchanged.

Fixes #1472

### DIFF
--- a/docs/design/PANEL-MODE-PERSISTENCE.md
+++ b/docs/design/PANEL-MODE-PERSISTENCE.md
@@ -116,16 +116,30 @@ auto-dock (`dockOnLeavingFullScreen`) ends the session via `endTransientSession`
 preference — and forces no surface open (outcome `'transient_back'`). The exit
 is **always** quiet: the prior surface is deliberately never captured or
 restored, so Back never reopens a surface the launch didn't durably choose.
-PUSH/REPLACE keep today's dock (an interactive `navigate` step leaving full
-screen still needs a panel to continue in), and a non-transient POP (a
-deliberately-adopted full screen) docks too. history v4 cannot tell Back from
-Forward — both are POP — which is accepted for this quiet exit.
+A non-transient POP (a deliberately-adopted full screen) still docks. history
+v4 cannot tell Back from Forward — both are POP — which is accepted for this
+quiet exit. PUSH behavior is covered by decision 5.
 
 `endTransientSession` neither persists (the launch never chose a preference —
 decision 2) nor dispatches `PANEL_MODE_CHANGE_EVENT` (the surface is already
 unmounted, so no live listener). Its call is deferred so `FullScreenPanel`'s
 unmount cleanup runs first while `getMode()` is still `'fullscreen'`; see the
 code comments for that hazard and the dead-state hazard it also avoids.
+
+### Decision 5 (#1472) — ordinary navigation out of transient prose exits quietly
+
+A Grafana nav or guide link click is a history **PUSH**, the same action emitted
+by an interactive navigation. Treating every PUSH as interactive therefore
+reopened the sidebar when a reader simply left a transient prose guide.
+
+The click-time handoff introduced in #1670 is the discriminator. A real
+Grafana-driving step calls `requestSidebarHandoffAndWait` before navigation;
+`handleExitToSidebar` changes the mode before its PUSH reaches the auto-dock, so
+Guard 1 ignores it. `NavigationManager.fixLocationRequirement` uses that same
+facade rather than pushing directly. A transient PUSH that still reaches the
+auto-dock has no interactive continuation and quietly calls
+`endTransientSession` (`'transient_navigation'`) instead of opening a surface.
+REPLACE retains the docking behavior.
 
 ## What is safe to change vs. load-bearing
 
@@ -140,7 +154,6 @@ code comments for that hazard and the dead-state hazard it also avoids.
 ## Related
 
 - Return-path interaction with browser Back on a transient prose launch:
-  **#1448** (resolved — see decision 4). The Grafana nav-click-with-prose
-  annoyance is a separate follow-up needing an interactive-step-in-progress
-  signal, still to be filed.
+  **#1448** (resolved — see decision 4). Grafana nav clicks from the same
+  transient prose state are covered by **#1472** (see decision 5).
 - Auto-open listener ownership across surfaces: **#1450**.
--- a/src/components/full-screen/full-screen-autodock.test.ts
+++ b/src/components/full-screen/full-screen-autodock.test.ts
@@ -52,8 +52,8 @@ function defaultInputs(overrides: Partial<Parameters<typeof dockOnLeavingFullScr
     myPluginId: PLUGIN_ID,
     guideUrl: baseTab.baseUrl,
     title: baseTab.title,
-    // Default to PUSH (an interactive `navigate` step): exercises the dock
-    // branches. The transient-Back branch is opted into per-test.
+    // A non-transient PUSH exercises the regular dock branches. Transient
+    // quiet-exit cases opt in per test.
     action: 'PUSH' as const,
     ...overrides,
   };
@@ -174,7 +174,7 @@ describe('dockOnLeavingFullScreen', () => {
     });
   });
 
-  describe('transient Back branch (browser Back out of a transient prose launch)', () => {
+  describe('transient quiet-exit branch (leaving a transient prose launch)', () => {
     beforeEach(() => {
       (panelModeManager.isTransient as jest.Mock).mockReturnValue(true);
     });
@@ -207,17 +207,31 @@ describe('dockOnLeavingFullScreen', () => {
       });
     });
 
-    it('leaves PUSH (interactive navigate step) on the docking path even mid-session', () => {
+    it('quietly ends a transient session when an ordinary PUSH reaches auto-dock', () => {
       const outcome = dockOnLeavingFullScreen(defaultInputs({ action: 'PUSH' }));
+
+      expect(panelModeManager.endTransientSession).not.toHaveBeenCalled();
       jest.runAllTimers();
 
-      expect(outcome).toBe('sidebar');
-      expect(panelModeManager.endTransientSession).not.toHaveBeenCalled();
-      expect(panelModeManager.setMode).toHaveBeenCalledWith('sidebar');
-      expect(sidebarState.openSidebar).toHaveBeenCalledWith('Interactive learning');
+      expect(outcome).toBe('transient_navigation');
+      expect(panelModeManager.endTransientSession).toHaveBeenCalledTimes(1);
+      expect(panelModeManager.setMode).not.toHaveBeenCalled();
+      expect(sidebarState.openSidebar).not.toHaveBeenCalled();
+      expect(sidebarState.setPendingOpenSource).not.toHaveBeenCalled();
     });
 
-    it('leaves REPLACE on the docking path (POP-only rule)', () => {
+    it('reports a transient navigation separately from browser Back', () => {
+      dockOnLeavingFullScreen(defaultInputs({ action: 'PUSH' }));
+
+      expect(reportAppInteraction).toHaveBeenCalledWith('full_screen_exit', {
+        destination: 'none',
+        guide_url: baseTab.baseUrl,
+        guide_title: baseTab.title,
+        reason: 'transient_navigation',
+      });
+    });
+
+    it('leaves REPLACE on the docking path', () => {
       const outcome = dockOnLeavingFullScreen(defaultInputs({ action: 'REPLACE' }));
       jest.runAllTimers();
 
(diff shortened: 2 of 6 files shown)


moxious (changes requested):
Thanks for closing the PUSH half of the transient full-screen exit. #1472 sat open because the obvious fix looked like it needed a whole new "interactive step in progress" signal, and spotting that the existing click-time handoff already flips panel mode before its push is the better answer. Rerouting `fixLocationRequirement` so that premise actually holds is the right instinct, and the rewritten integration test wires the real `panelModeManager` instead of mocking it.

Three things to fix, below. The first is the only behavioural one; the other two are the records this change depends on. I have deliberately kept this list short — a few further observations came up that are our work rather than yours, and I am filing those as issues instead of putting them on you.

## Requested changes

1. [blocking] **F3 — In full screen, a requirement target the validator refuses is silently swapped for another page while the fix reports success** (medium · interactive-engine)

   Rerouting `fixLocationRequirement` through the handoff means the requirement's `targetHref` now passes through `resolveSafeTargetPath`, which returns `undefined` indistinguishably for a path lacking a single leading slash and for one `validateRedirectPath` refuses (`src/components/full-screen/resolve-safe-target-path.ts:14-25`). `FullScreenPanel.tsx:297-299` then pushes `safeTargetPath ?? priorPath ?? PLUGIN_BASE_URL` and logs nothing. So the reader clicks "Fix this", gets relocated to the page they were on before full screen — or to Pathfinder home — leaves full screen, and is told the fix succeeded.

   The "reports success" half is structural rather than incidental, which is why this blocks. `fixLocationRequirement` is typed `Promise<void>` (`navigation-manager.ts:1368`), and `locationHandler.execute` awaits it and returns `{ ok: true }` unconditionally (`src/requirements-manager/fix-handlers/location.ts:14-18`). There is no channel through which a refusal could be reported, so `step-checker.hook.ts`'s `if (!result.ok)` branch is unreachable for every location fix. Base pushed the raw target with no substitution, so the silent redirect is new here.

   Reachable from documented authoring: `docs/developer/interactive-examples/requirements-reference.md:409` gives `on-page:/admin/users` as an example, and `/admin` is admin-gated, so it is refused for every non-admin reader. A relative `on-page:dashboards` also loads, since `condition-validator.ts:121-131` flags the missing leading slash only as a warning. Two mitigations are real: no `on-page:` value in `src/bundled-interactives/**` is refusable today, and because the handoff flips mode off `fullscreen`, a retry takes the direct branch and behaves as base did — so this is a one-shot relocation, not a dead end.

   **Required:** validate the target once at the top of `fixLocationRequirement`, before the branch — bail out when validation refuses it, and forward only an already-validated path into the handoff. `validateInternalNavigationPath(targetPath, currentUserIsAdmin())` is the validator every other author-controlled push in the repo uses (`navigate-handler.ts:61`, `starting-location.ts:77`), and it preserves search and hash where `resolveSafeTargetPath` strips them.

   This needs `fixLocationRequirement` to return a result instead of `Promise<void>`, which also touches `FixHandlerNavigationManager` in `src/requirements-manager/fix-handlers/types.ts` and the handler itself — three files. That is deliberate: it makes `ok: false` reachable for the first time, so `step-checker` keeps the retry affordance and the reader sees an error instead of a redirect.

   It also closes a second gap in the same stroke, which is why I am not filing that one separately. `isGrafanaDrivingHandoffNeeded` requires `getMode() === 'fullscreen'` (`panel-mode.ts:410`), so in the default sidebar or floating mode the *unvalidated* branch is the only branch — `on-page:` is accepted on its prefix alone (`requirements.types.ts:55-62`) and `onPageCheck` hands the remainder back verbatim (`checks/location.ts:14,24`). That gap predates this PR, but validating before the branch fixes both at once, whereas validating inside the full-screen branch alone would leave it. For scope: a pushed `/logout` does not sign anyone out, because `locationService.push` is a client-side pushState that never issues the request the backend handler needs. The realistic outcome is an unvalidated same-origin route change.

2. [blocking] **F5 — A fourth caller of the handoff facade lands while the contract anchor, the routing row, and the facade's own doc comments still enumerate exactly three** (low · cross-cutting-architecture)

   `navigation-manager.ts:17,1370` adds a fourth caller of `requestSidebarHandoffAndWait`, but every record of that contract still says three. `isGrafanaDrivingHandoffNeeded`'s doc comment claims to list "every caller" and names only the hook gate, `interactive-guided.tsx`, and `code-block-step.tsx` (`panel-mode.ts:396-397`). `requestSidebarHandoffAndWait`'s doc describes the trigger as a "Do it" click and `targetPath` as a resolved step, milestone, or course location (`panel-mode.ts:325-335`), whereas the new caller is a "Fix this" passing a raw `on-page:` value.

   The load-bearing part is the routing: the `full-screen-sidebar-handoff` contract anchor in `docs/design/CONCERN_DETAILS.md` is not in this PR's diff, and the concern's routing row omits `src/interactive-engine/navigation-manager.ts` — so a future change to this new call site will not route to this concern at all. The anchor's own review question is "does every execution path apply the handoff gate?", and the answer a reviewer reads is now short by one, at exactly the moment the auto-dock started depending on the facade being exhaustive.

   **Required:** AGENTS.md asks for the anchor update to ride in the same PR that extends the contract. Add `fixLocationRequirement` to both `panel-mode.ts` doc comments and to the anchor in `docs/design/CONCERN_DETAILS.md`, add `src/interactive-engine/navigation-manager.ts` to the concern's routing row, note that `targetPath` may now be a requirement target, and state in the PR body that this extends rather than replaces the established contract. These are documentation lines, not logic.

3. [blocking] **F4 — The premise that makes the widened quiet exit safe is asserted only in a comment, and the test that pinned the old behaviour was deleted with nothing equivalent replacing it** (medium · testing-and-verification)

   This PR converts a narrow rule — POP-only quiet exit — into a claim about the rest of the codebase: that no interactive push can reach Guard 3, because every Grafana-driving action flips mode through the facade first. That claim is what keeps a reader from being stranded mid-guide with no surface, and it now rests on four independent call sites (`interactive.hook.ts:440-443`, `interactive-guided.tsx:374-375`, `code-block-step.tsx:235`, and the new `fixLocationRequirement`). Its justification lives only in a comment at `full-screen-autodock.ts:93-98`.

   Meanwhile the base assertion "a PUSH mid-session (interactive navigate step) still docks and DOES reopen the sidebar" was deleted from `transient-back-quiet-exit.integration.test.ts`, and its replacements drive `dockOnLeavingFullScreen` with a hand-built `action: 'PUSH'` input rather than any interactive path. The new `navigation-manager.fullscreen-handoff.test.ts` mocks `../global-state/panel-mode` wholesale, so it proves branch wiring against its own stub and cannot detect the gate changing meaning.

   **Required:** one integration test that wires the real `panelModeManager` and the real `dockOnLeavingFullScreen` to a real `fixLocationRequirement` call from a transient full-screen session, asserting a surface survives. That is the assertion the deleted one was doing, adapted to the new rule.

   I am **not** asking you for the structural tripwire that would enforce the premise rather than describe it — enumerating every `locationService.push` reachable from interactive execution is our infrastructure work, and I am filing it separately.

## Not your work — we are filing these

Recording them so you know they were seen and are not waiting on you: the `full-screen-autodock.ts:1-18` module header still names the two triggers this PR stops docking; `PANEL-MODE-PERSISTENCE.md:131` asserts a guide-content link click produces a history PUSH, which it never does (`link-handler.hook.ts` resolves hrefs against the content base and opens in-panel or via `window.open`, never `locationService`); `endTransientSession`'s doc still scopes it to browser Back and #1448; and the `full-screen-sidebar-handoff` anchor cites #1670, whose commit is not an ancestor of `main` — the facade actually landed squashed inside #1621.

Two further observations I looked at and am deliberately not raising: an unheard handoff would skip the navigation rather than fail visibly, but I could not reach that state through any mounted guide surface at head and the one path that does is behind a default-false admin setting; and the new `transient_navigation` reason moves existing `full_screen_exit` volume off `navigation_away`, which is worth knowing when reading that series but is not a change to make here.

CI on this head is still held pending maintainer authorization, which is routine for a fork PR and unrelated to your change. I will release it so you get CI signal alongside this review.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1858
Summary: quiet exit on transient navigation
Verdict: Request Changes
Results: 3 blockers, 0 non-blocking findings, 0 follow-ups

dvd233:
Thanks for the review. The follow-up commit `f345dd1a8783ce66cc8867f8595a7af1c785d8b0` addresses F3/F5/F4: `fixLocationRequirement` validates and propagates refused targets before either navigation branch, the handoff contract/routing anchor documents the fourth caller and requirement-target semantics, and the real panel-mode integration test covers the Fix this path. The focused full-screen/navigation suite passes 91/91, with typecheck and targeted ESLint clean. Could you re-review the current head?

moxious (approved):
Thanks for pushing the follow-up commit — F3, F4, and F5 all check out at this head.

PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1858
Summary: quiet exit on transient navigation
Verdict: Approve

CI is still finishing up on this head (nothing failed so far) — once it's green this is ready to merge as-is.

PR #1863, opened 2026-09-10
fix: isolate progress reset for prefixed content keys

### LINKED ISSUE #1846: Resetting a path deletes progress for guides whose content key shares a prefix
Resetting a learning path can delete step progress belonging to a guide that is not in it.

`clearAllForContent(contentKey)` builds a prefix of `${PREFIX}${contentKey}-` and removes every localStorage key that starts with it (`user-storage.ts:1042-1069`), while a step record is written as `${PREFIX}${contentKey}-${sectionId}` (`:986-989`). A content key that is itself a prefix of another therefore sweeps the other guide's records.

Reachable with shipped content: `bundled:first-dashboard` and `bundled:first-dashboard-cloud` are both real packages with their own `content.json` (`repository.json:48`, `:85`), launchable in the same profile. Resetting a path containing the first silently destroys the second's step progress, collapse state, acknowledgements and done markers. There is no undo.

The same applies to the collapse, acknowledgement and done key spaces, which share the pattern. With author-written App Platform ids the set of hyphen-prefix pairs is not controlled by this repository.

Pre-existing rather than newly introduced - the identical call is made today over a list already containing `bundled:first-dashboard`. It surfaced during review of #1843, which makes one module the owner of the key list the sweep consumes, and was deliberately left out of that PR because the sweep lives in `user-storage.ts`.

Two ways to close it: delimit the four key spaces so one content key cannot be a prefix of another, or sweep on an exact `${contentKey}-${sectionId}` match set rather than `startsWith`. Existing coverage uses the non-colliding pair `guide-a` / `guide-b` (`user-storage.test.ts:403-426`), so a regression test wants a genuinely colliding pair.

Nikshay1:
## What does this PR do?

Fixes progress resets deleting data for guides whose content keys share a prefix.

Previously, `clearAllForContent()` removed section progress by scanning legacy localStorage keys with a `${contentKey}-` prefix. This could incorrectly match another guide, for example:

- `bundled:welcome-to-grafana`
- `bundled:welcome-to-grafana-cloud`

This PR introduces collision-safe, length-prefixed section keys for content after it has been reset.

Existing guides continue to read their legacy keys until reset. On reset, a per-content v2 marker switches that guide to the versioned key format, making its legacy entries logically inactive without deleting potentially overlapping sibling-guide data.

The change also updates cross-tab cache invalidation and the E2E guide cleanup path to understand the versioned storage format.

Regression coverage verifies that:

- resetting a guide does not remove progress from a prefix-sharing sibling
- step, collapse, acknowledgement, and section-done state remain isolated
- progress written after migration is cleared correctly by subsequent resets
- cross-tab storage events invalidate versioned progress correctly

## Linked issue(s)

Closes #1846

## Checklist

- [x] This PR addresses a **single concern** (one bug fix or feature, or a set of closely-related changes). Unrelated changes belong in separate PRs; see [CONTRIBUTING.md](./CONTRIBUTING.md#pull-request-scope).
- [x] All commits are signed (required, or the checks will not pass). See [CONTRIBUTING.md](./CONTRIBUTING.md#signed-commits).
- [x] `npm run check` passes locally — the pre-merge gate.

### DIFF
--- a/docs/developer/STEP_MODEL.md
+++ b/docs/developer/STEP_MODEL.md
@@ -26,7 +26,7 @@ The hash inputs are documented at `src/global-state/step-id.ts` (`deriveStepId(.
 
 ## Completion store — canonical persistence
 
-Step completion lives in `src/global-state/completion-store.ts`. The store is the canonical persistence layer — `SectionState` no longer carries a parallel `completed` set, and step components no longer maintain a local `isLocallyCompleted` flag. The store backs the existing `interactiveStepStorage` namespace so localStorage shape is preserved.
+Step completion lives in `src/global-state/completion-store.ts`. The store is the canonical persistence layer — `SectionState` no longer carries a parallel `completed` set, and step components no longer maintain a local `isLocallyCompleted` flag. The store uses `interactiveStepStorage`; untouched content keeps the legacy section-key shape, while a reset switches that content key to collision-safe length-prefixed section keys.
 
 The step-checker FSM (`src/requirements-manager/step-checker.hook.ts`) and the `SequentialRequirementsManager` orchestrator remain as mirrors of the store — they hold the in-memory checking state needed to drive the UI. The FSM writes through to the store on every terminal transition (manual completion, skipped, objectives auto-complete, and reset) via `writeStoreCompletion` / `writeStoreReset`, so the orchestration mirrors and the canonical store cannot disagree on either axis. The `step-checker.store-bridge.test.ts` tripwire pins this contract.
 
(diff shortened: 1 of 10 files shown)


tomglenn (approved):
Approving. This is a clean fix for a real data-loss bug, and the approach is the right one.

Encoding the content key length makes the key space unambiguous by construction, so no scan can straddle two guides regardless of how their names overlap - that is a stronger guarantee than tightening the match would have given. Gating reads and writes behind a per-content marker, and switching a guide over only on reset, means no existing progress has to be migrated and none can be lost on upgrade. The cross-tab listener and the e2e cleanup path both being taught the versioned shape is the sort of thing that is easy to miss.

Verified independently on this side: the collision reproduces with the shipped `welcome-to-grafana` / `welcome-to-grafana-cloud` pair, and a set of regression tests written against that pair passes on your branch, including the raw pre-fix key case.

On CI: the one red check is a reporting step that posts results back as a comment, which cannot write from a fork. The browser tests themselves passed on all seven images.

We are picking up the remaining follow-on work ourselves rather than asking you to extend the scope here - notably carrying the versioned format through to the read-side counters. Thanks for this.

PR #1864, opened 2026-09-10
fix(storage): key guide progress so one guide cannot read another's

tomglenn:
## Summary

A guide's completion percentage could include a different guide's steps. A section's progress was stored under a key that joined the guide identifier and the section id with a hyphen — and both halves contain hyphens, since a guide identifier can be a whole tab URL — so nothing marked where the identifier ended. A scan for `bundled:welcome-to-grafana` matched every record belonging to `bundled:welcome-to-grafana-cloud`. Both of those ship, and one is in a learning path.

Every record now keys as `{prefix}{len(contentKey)}:{contentKey}:{sectionId}`. The character count in front of the identifier marks the boundary, so a reader counts to it instead of searching for it.

#1863 introduced that shape for the reset path, behind a per-guide marker that moved a guide onto it the first time it was reset. Now that the shape is used for every read and every write, the marker has nothing left to decide, so it is gone — a simplification the wider scope allows, along with the dual-shape read path it required in the cross-tab listener and the E2E runner.

## What a reader sees

One-time, on first load: **a guide you had started and not finished reopens at the beginning.**

Unaffected: completed guides, badges, streaks, and finished milestones. Those live under keys that carry no guide name and were never ambiguous. A learning path keeps its finished milestones and restarts only the unfinished one. There is an upgrade note in `docs/sources/upgrade-notes/_index.md`.

## Why the old records are discarded rather than migrated

This is a deliberate choice, not something left undone. Moving a record requires knowing where its guide identifier ends, and that is precisely the information the old key never carried. Both available routes fail:

- Rewriting stored records means deciding that boundary per key — the original defect.
- Migrating per guide on load means scanning for keys beginning with that guide's identifier, which collects a longer-named neighbour's records and rewrites them as this guide's. That is not lost data but misattributed data, permanently and silently, which is worse than discarding it.

Records can also exist for guides not installed on a given stack at all — the bundled library, the interactive-tutorials repository, and custom guides on customer stacks — so a load-triggered migration would leave records behind indefinitely anyway.

So they are removed instead, in one pass on load.

## How to verify by hand

1. Open **Welcome to Grafana Cloud** and complete one step.
2. Close it and open **Welcome to Grafana** — a different guide whose name is the start of the first one's.
3. **Before this change:** its progress chip counts the step you completed in the Cloud guide, so it reads a non-zero percentage for a guide you have not started.
4. **After:** it reads 0% with no step marked complete, and the Cloud guide still shows its own step complete.

To see the discard, complete a step in a guide on the previous build, then load this one. That guide reopens at the beginning, while My learning still shows your completed guides, badges, and streak.

## What to look at

- `src/lib/storage/progress-keys.ts` — the single key shape and the discard sweep.
- `src/lib/storage/progress-keys.test.ts` — the sweep's blast radius, pinned by seeding every other value in the `StorageKeys` registry plus badge, streak, completion-queue, and milestone fixtures and asserting `localStorage` is byte-identical afterwards.
- `src/lib/user-storage.hook.test.ts` — that `useUserStorage` actually calls the sweep, and that it runs once per page load. Without this, deleting one line would stop the discard happening for everyone while every other test stayed green.
- `clearAllForContent` in `src/lib/user-storage.ts` — deletes, then re-reads and rejects if any record survived.

## Reset reporting

The reset previously completed by writing a marker; if that write was refused, the failure was logged and swallowed, leaving the old records live and still counted. It now deletes and then re-reads, rejecting if anything survived, so a reset that did not complete reports itself rather than appearing to work.

The re-read deliberately does not depend on knowing why a removal failed. The removal path swallows its own errors, so a refused removal is invisible to the caller; looking again afterwards is what makes the answer correct whatever the cause.

Three code paths can reset progress. Two now report a failure to the reader: the per-guide reset publishes an error, and a learning-path reset attempts every guide in the path and reports once at the end instead of stopping at the first failure. The third — the block editor's guide preview — makes a failure visible by leaving the interface unchanged: the progress stays on screen and the reset control stays available, so it does not appear to have worked. Adding a message there would be an improvement rather than a fix, and separate work is already queued against that surface, so it is better made once there.

## A percentage that self-corrects

A guide's stored percentage lives in its own record, keyed exactly, which was never ambiguous — and it is left alone. So the guide list can keep showing the percentage an unfinished guide had until you open it, at which point it recomputes.

Clearing that record instead would pull down learning-path percentages that this change sets out to preserve, and a number that corrects itself when you open the guide is gentler than one that drops to zero unprompted.

## Breaking changes

The one-time reader-visible effect described above. No API or configuration change.

## Reversibility

Reverting restores the previous code, but records already discarded are gone — a reader whose in-progress position was cleared does not get it back by reverting. Everything durable is untouched either way.


Spidy88 (approved):
LGTM

PR #1866, opened 2026-09-11
feat(completion): base guide, path and journey progress on evidence, not navigation

tomglenn:
## Summary

Guide, path and journey completion now come from one shared calculation, driven by evidence the reader actually produced, rather than from navigation position or a second, independently-maintained count.

- The block index used to compute a guide's denominator and numerator is now computed once per content load and frozen for that load, so both halves of the percentage always come from the same traversal.
- The completion store is the sole producer of a guide's percentage. The older `getGuideProgress` snapshot and its defensive clamps are deleted.
- The three call sites that auto-completed a guide purely from navigating past it are removed. Completion now only happens by evidence or by clicking **Mark complete**.
- Path and journey progress is the mean of their members' own percentages (equal weighting per member), pinned to the same storage the members themselves report from.
- A cross-surface parity test asserts every surface reports the same number for the same guide and evidence, and an architecture test acts as a tripwire against the common single-line `(numerator / denominator) * 100` shape appearing outside the shared module — a catch on the obvious case, not a guarantee against every way a new calculation could be written.
- Fixed a defect where resetting a guide and re-completing it produced no durable record, no badge and no path credit the second time, while the UI still reported success. The write-side dedupe guard is now backed by durable, reload-surviving storage and cleared by every reset path.
- Fixed the block-editor preview always reading 0%, and fixed it going stale while the author kept editing after that.
- Reduced two hand-maintained developer README inventories to short pointers at their generated source, rather than re-synchronizing lists by hand.

## Fixed in review

- **A section with no author id could double its step count.** A pre-registration path in the renderer predicted a different id for such a section than the section's own registration used, so both survived and the displayed "Step N of M" count doubled. Both now use the same id-derivation precedence.
- **The reset-then-re-mark fix above still missed one shape**: milestone identity depended on the caller correctly classifying a tab as a learning journey, which a path member opened through some launch routes does not. The reset path no longer depends on that classification — it derives the same identity the writer does regardless.
- **The recommendation card's journey percentage could read stale** for any journey that is not backend-guide-hosted, since it reads a separately-stored snapshot that nothing but the journey's next load refreshed. It now also refreshes on every milestone completion.
- **The completion write queue's durable guard could outlive a record the queue itself dropped** (evicted over capacity, or past the 30-day retention horizon) — permanently losing the ability to record that identity again. The guard now lifts in exactly those cases, and stays set only when the backend genuinely rejected the record.
- **A percentage persisted under the deleted counting rule could stay stale indefinitely.** Reopening a guide with real evidence now recomputes and re-persists its percentage under the current rule — see the new "Transitional behaviour" note under decision 1 in `COMPLETION-MODEL.md` for the honest limits of that heal.
- The completion-emitted guard is now local-only storage — it never made a readable remote write in the first place, and its deletes were leaving empty-value tombstones behind; both are gone.

## How to verify by hand

1. **See a percentage that only moves on evidence.** Open a multi-milestone guide or path, note its percentage, then navigate forward through several milestones/sections without completing any interactive step or acknowledging any section. The percentage should not move. Now complete a step (or click **Mark complete**) and confirm it moves at that point, not before.
2. **Confirm a guide can still be completed — this is the critical check.** Open any guide, and without necessarily finishing every step, click **Mark complete** at the bottom. Confirm: the guide badge is awarded, the guide's percentage reads 100, and any path or journey it belongs to picks up the credit. Do this on both a guide with interactive steps and an all-prose guide. Navigation alone should never complete anything by itself.

## Behavior changes to expect

**Percentages can now go down, and this is intended.** Because navigating past a milestone no longer counts as progress, a reader who pages through a path or journey without completing anything will see its percentage fall — possibly to zero — and a path's cover-page call to action can revert from "Resume" to "Get started". Navigating without taking any action is browsing, not progress, and the model no longer conflates the two. This includes the segmented milestone progress bar: its "Milestone X of Y" label is still a plain position breadcrumb and is unchanged, but the bar's *segment fill* now reflects earned credit from the shared calculation rather than position, so a reader who pages forward without acting will see earlier segments unfilled. **This is release-note material** — it's a visible, deliberate behavior change for any reader who has been navigating rather than completing.

## Trade-offs already accepted in this model (see `docs/design/COMPLETION-MODEL.md` for the full reasoning)

- **A guide with more prose than interactive steps has a coarse percentage**, and an all-prose guide's only reachable values are 0% and 100% — there is no partial credit for reading. This is roughly half the published library today; it's an authoring limitation, not a bug in the arithmetic.
- **Path and journey milestones are weighted equally regardless of length.** A path of one short milestone and one long one reports 50% at what may really be a fifth of the reading. Simpler and more predictable than weighting by size, at the cost of that imprecision.
- **A path member whose progress can't be resolved is excluded from the mean and counted, never scored zero** — a zero would be indistinguishable from a real result and would silently drag the number down. The exclusion count is what makes that gap visible instead of hidden.
- **Duplicate-write prevention is now best-effort, not a complete guarantee** — deliberately. The write-side guard that stops a guide from recording two completions is only set once the completion has durably reached the write queue (the point that survives a reload), never merely on the attempt to record it. If nothing was listening yet, the completion stays eligible to be recorded again on a later session rather than being silently and permanently lost. Between an occasional duplicate (recoverable — the record is true, and downstream de-duplication can collapse it) and a lost completion (not recoverable — nobody will try again), this deliberately accepts the former.

## Known, deferred limitations (tracked, not fixed in this PR)

- **A guide whose interactive work sits inside a conditional block's branch reports a percentage that cannot rise past the conditional.** The block traversal treats a conditional as a single opaque unit and does not credit evidence produced inside either of its branches. The guide is still completable — completion is a flag set by evidence or by **Mark complete**, not a requirement that the percentage reach 100 — but the percentage undercounts for this content shape until the traversal is taught to look inside a conditional's active branch.
- A path member's resolved percentage can, in one specific case, be looked up through App Platform metadata for a bundled member whose id happens to collide with a customer's own custom guide id, because guide-metadata resolution checks App Platform before the static bundled fallback. Documented as a known follow-on in `docs/design/COMPLETION-MODEL.md` (decision 9) rather than fixed here, since the right fix is in the metadata resolver a layer down, not in this rollup.
- Two narrow, already-tracked edge cases: the Mark-complete button's own click handler can announce a new percentage a moment before its write has landed, and a milestone whose URL yields no usable slug loses a same-value shortcut credit (it still scores correctly from its own persisted record). Both are bounded to their specific call sites and don't affect the general model.
- **This PR must ship in the same release as #1864.** An id-less section's runtime id changes shape here (from a render-order counter to a parser-derived path), which re-keys that section's stored progress under four namespaces. #1864's own key-shape sweep is what makes that re-keying free for every reader already on it — discard-and-restart rather than a silent orphan. Landing this PR in a *later* release than #1864 would open a real window (roughly a third of published sections carry no author id) where a reader's in-guide position for such a section is silently orphaned rather than swept; landing them together closes it, since #1864's own sweep already discards that reader's pre-#1864 progress regardless. Bounded either way to in-guide position — marks, badges, streaks and path credit are keyed by content key or guide id, not section id, so none of those are at risk.

## One pattern worth naming

Three separate identity mismatches surfaced during this change, all the same shape: a writer and a reader each independently derived an identity for the same thing (a milestone, a section's evidence key, an acknowledged section with no author-assigned id) and drifted apart. Each is now fixed the same way — one shared derivation, called by both the writer and the reader — rather than by making either side guess at the other's convention. If this class of bug shows up a fourth time, it's worth asking whether identity derivation for evidence generally needs one home rather than being re-derived per module.

## Left deliberately undisturbed

Backend-guide journeys' historical completion records — write-locked at exactly 100 by an earlier change, and never written any other value — are untouched by this PR. The new journey percentage is computed from each milestone's own evidence rather than from that namespace, so those records simply stop being read for that purpose; they are not migrated, cleared, or reinterpreted. **Also release-note material**, since anyone auditing that storage namespace should know why it goes quiet rather than assuming it broke.


tomglenn:
Hold off reviewing for now — there are fixes still to come and I would rather not waste your time on a version that is about to change.

Two defects this PR claims to fix turned out to be only half-fixed, both confirmed by tracing the code rather than taken from the bot report:

1. **Reset then re-mark still records nothing.** `resolveCompletionIdentity` resolves `guideId` as `packageManifest?.id ?? fallbackId`, so the manifest id wins whenever a manifest is present. `resetGuideProgress` passes `packageManifest`, so it resolves to `manifest.id`; `markMilestoneDone` passes only `repository` plus `fallbackId: milestoneSlug`, so it resolves to the slug. On any milestone carrying a manifest the two diverge, the emitted guard stays set, and the second mark produces no durable record. The earlier fix corrected the fallback, which is not the value that decides.

2. **The block-editor preview index is still frozen across edits.** `publishGuideIndex` ignores a second publish for the same content key while the preview key stays constant while the author types, so the denominator and positions stay at first paint.

Both are being fixed now, with the identity derivation extracted so the write path and the reset path cannot diverge again — this is the third instance of that same class in this change, so it is worth closing at the shared seam rather than per call site.

Two further findings are deliberately not being actioned here, and are worth your view rather than mine:

- **HTML interactive guides get no block index**, so their percentage stays 0 until Mark complete. Establishing whether that is a regression from this change or pre-existing, and how much shipped content it affects, before touching it.
- **`completion_percentage` in analytics has changed meaning** — evidence-based mean rather than navigation position — with no discriminator to separate pre- and post-cutover rows. That is a product/KPI call, not a code one.

Will comment again when the fixes are pushed and CI is green. Everything else in the PR body still stands, including the two intentional behaviour changes called out there.

tomglenn:
Fixes are in — head `3680c41b`, ready for review.

Both defects from the previous comment are closed, and I verified each by reading the mechanism rather than accepting that they were done, since the first attempt at each looked complete and was not:

**Reset/re-mark identity.** There is now a single `resolveMilestoneCompletionIdentity` in `completion-identity.ts`, called by both `markMilestoneDone` and the reset path. It takes no argument that can override the id — a milestone is keyed by its slug by construction, so the two sides cannot diverge again. That is the substantive difference from the earlier attempt, which corrected the fallback while leaving the manifest preference that actually decided the value. The rationale is in the function docstring: a milestone is addressed by URL, not by its journey manifest, so a manifest id winning there would key the writer and a reset differently for the same milestone.

**Preview index staleness.** The freeze is unchanged for published content, where it is correct — content is immutable per content key. Previews get one narrow exception: when the guide object under a stable preview key actually differs from the one last published, the stale index is evicted before republishing. Scoped to preview URLs and triggered by the content genuinely changing, not on every render.

Local suite green at 9079 passing, with the one pre-existing `coda-client` failure that also reproduces on unmodified `main`. CI green so far on this head with nothing failing.

Still worth your view rather than mine, unchanged from the previous comment and deliberately not actioned here:

- **HTML interactive guides get no block index**, so their percentage reads 0 until Mark complete. Whether that is a regression from this change or pre-existing, and how much shipped content it touches, is being established.
- **`completion_percentage` has changed meaning** — evidence-based mean rather than navigation position — with no discriminator separating pre- and post-cutover rows.

The two intentional behaviour changes in the PR body are the ones to weigh hardest: path percentages can go down for readers who navigated without completing anything, and interactive content inside a conditional block credits no position.

moxious (dismissed):
Blockers:

1. [blocking] **section-registry-third-deriver-double-count — A third, unmigrated section-id deriver in content-renderer double-registers every id-less section** (high · cross-cutting-architecture)
   `content-renderer.tsx:828` still predicts a section's id with the old convention (`section-${++sectionCounter}`) and pre-registers it, while `InteractiveSection` now registers under the parser-stamped `sectionId`. For a JSON section with no author `id` the two keys differ, both survive in `globalStepRegistry`, and the totals double. Two independent full-React render probes measured `getTotalDocumentSteps()` at 7 where the truth is 5, and 4 where the truth is 2; the same probes at base return the correct values. It surfaces as the literal "Step N of M" text in full screen, the minimized pill and the floating panel. The bundled library has 21 sections all carrying ids, which is exactly why the new parity sweep cannot see this - but the block editor's own lint advises authors to "Remove its id if it's purely organisational", and 37% of published sections have no id. The completion percentage is unaffected.
   Required: Apply the same precedence `InteractiveSection` now uses - `el.props.sectionId ?? id-based ?? counter` - in the `documentLayout` memo, and add a test asserting one registry entry per section after a full render of a guide with an id-less section.

2. [blocking] **reset-guard-identity-divergence — Reset lifts the durable completion guard under a different identity than the writer used** (high · reversibility-and-one-way-door · partially reversible)
   `markMilestoneDone` calls `resolveCompletionIdentity` without a manifest, so it always records under the milestone slug. `resetGuideProgress`, fed `packageManifest` by `useContentReset`, resolves `guideId = manifest.id`. The keys differ, so reset clears a key that was never written and the real guard stays set - and because this PR makes the guard durable, the miss is permanent for that browser profile rather than clearing on reload. Both skeptics reproduced it against real storage: re-recording after the reset emits zero facts. `MyLearningTab.tsx:301-303` makes it guaranteed rather than incidental by launching a path member with `packageManifest: {...parentPath.manifest, id: parentPath.id}` - the path id - while the member's slug is its own. This is the defect the PR body claims to close, still open on package-sourced journeys. `guideSource` does not diverge; only `guideId`.
   Required: Have both sites share one identity derivation instead of each calling `resolveCompletionIdentity` with different inputs, and add the writer-to-reset agreement test - built on real storage, since `resetGuideProgress.test.ts` asserts arguments against a mock and its manifest fixture picks the one shape where both branches coincide. Also feed the already-computed `milestoneKeys` at `learning-paths.hook.ts:532` into the path-member invalidation so a URL-based `resetPath` does not depend on the dynamic fetch having landed.

3. [blocking] **old-rule-percentages-reinterpreted — Percentages written under the deleted counting rule are read verbatim by the new rollups, and reopening a guide mostly does not heal them** (medium · reversibility-and-one-way-door · partially reversible)
   Base persisted `completedSteps / totalDocumentSteps`; head persists `position / totalBlockCount`, which is systematically lower - measured on the bundled corpus, a reader who completed every step of `block-editor-tutorial` holds 100 where the new rule gives 68. The record carries no version and nothing migrates it. A probe run outside `act()` settled the heal question: the corrective write is dropped, because `ensureHydrated`'s `.then` is a microtask while `publishGuideIndex` is a passive effect that notifies nobody, so `computeGuideProgress` returns `undefined` and `hydratedSections` then blocks a retry. A guide with any fully-completed section does heal via the section-complete effect; a guide partial in every section, or standalone-step-only, does not. The durable edge: `completedGuides` is populated only by `markGuideCompleted`, whose percentage-driven writer returns early for backend-guide journeys, so an App Platform member finished pre-upgrade is not in that set, falls through to the stale record, resolves 100 as `source: 'persisted'`, and the path reads complete.
   Required: Refresh the persisted percentage once per content key when the frozen index publishes and evidence already exists - a guarded `refreshAndNotifyGuideProgress` in the content-load seam, skipped when `collectEvidence` is empty so the capped namespace is not filled with zeros. Alternatively version the namespace and treat an unversioned record as absent. Either way, record the transitional behaviour in `COMPLETION-MODEL.md`, since stale values are systematically higher and a path can read complete off them.

4. [blocking] **journey-percentage-diverges-on-recommendation-card — The recommendation card's journey percentage is a load-time snapshot of a number whose meaning changed, so a finished journey reads below 100 there** (medium · context-engine)
   `docs-panel.tsx:566-572` writes `getJourneyProgress()` into `journeyCompletionStorage` on every journey load, and `context.service.ts:845/893` is the only source of the card's `completionPercentage`. `getJourneyProgress`'s meaning changed underneath that unchanged write. For a public docs journey there is no completion-time refresh - the only whole-journey write is gated on a `backend-guide:` base - so the stored value trails by the credit earned since the last load. Worked trace on a 5-milestone journey with four complete: base stored 100 and head stores 80, and after the reader completes the fifth the toolbar and cover page recompute to 100 while the card stays 80. No refresh corrects it: the context hook's three refresh triggers all re-read the same stale value, and no context-engine subscriber exists for the new revision channel. The gap is 100/N, so 50 points on a two-milestone journey. It self-heals only on a later load of that journey.
   Required: Feed `journeyProgressFromMilestones` through the recommendation enrichment so the card uses the shared calculation, or refresh `journeyCompletionStorage` on completion for non-backend-guide journeys as already happens for backend-guide ones. Add a test pinning the docs-panel write against the context.service read. The CTA concern does not apply - base said "Start" on a finished journey, so head's "Resume" is not worse.

5. [blocking] **owd-guard-outlives-dropped-record — The durable completion guard is set on queue acceptance, so a record the queue later evicts or expires can never be recorded again** (medium · reversibility-and-one-way-door · partially reversible)
   `onFact` returns `queue.enqueue(...)`, i.e. `storage.put` success rather than POST success, and the recorder sets the durable guard on that. The queue can then evict at capacity or drop past the 30-day retention horizon without lifting it, so the identity becomes permanently unrecordable in every future session. A probe over the real recorder, hook, queue and storage confirmed both arms. This contradicts the PR's own stated rule - it deliberately tolerates duplicates because a lost completion is unrecoverable. Bounded honestly: the mainline `created` outcome retains the guard correctly, the record is write-only today (no frontend consumer of the read routes), and base was worse in the common case, since `emitted.add(key)` ran unconditionally before a `void`-returning `emit()`. The argument that an expired record would be rejected anyway was examined and declined - producers stamp `completedAt` at record time, so a base-style re-record carried a fresh acceptable timestamp.
   Required: Lift the guard when a record leaves the queue for a non-terminal reason: call `invalidateEmittedCompletion` from the eviction and `expired-drop` branches, which needs the dedupe key carried on the queued item. Leave `terminal-drop` guarded, since a 4xx means the fact was rejected rather than lost.

6. [blocking] **percentage-ratchet-overstated — The contract anchor says a new percentage calculation fails the build; 10 of 12 obvious rewrites pass the gate** (medium · testing-and-verification)
   `architecture.test.ts:273-289` requires, per line, all of a closing paren immediately before `* 100`, a `/` on the same line, and a match against a fixed keyword list. Two skeptics independently ran the same twelve variants and got the same result: caught only `(completed / total) * 100` and its `100.0` form; passed `completed / total * 100`, `Math.round(100 * done / total)`, a two-statement fraction, a multi-line split, and the identical shape with the denominator named `denominator`, `len`, `size`, `members` or `all`. `collectSourceFiles` also excludes `src/cli/`. Meanwhile `CONCERN_DETAILS.md:83`, rewritten by this PR, asserts without qualification that such a calculation "fails the build", and the PR body repeats it. This is not house style - the only other "fails the build" claim in `docs/` is a genuine compile-time exhaustiveness partition. The test's own "Loose on purpose" comment is accurate; it is the prose that overreaches.
   Required: Soften both sentences to what the gate does - it catches the common single-line `(numerator / denominator) * 100` shape and is a tripwire, not a proof - or strengthen detection by normalizing whitespace across lines and dropping the keyword requirement in favour of a `* 100` plus `/` co-occurrence. Also fold in the adjacent overclaim in the same sentence: guide-stats is not the only place that computes "the number every reading surface displays", since the percentage namespace has two writers and is read outside guide-stats in five places.

7. [blocking] **owd-idless-section-progress-orphaned — Re-keying id-less sections orphans four progress namespaces, which is free only if this ships in the same release as #1864** (low · reversibility-and-one-way-door)
   An id-less section's runtime id moves from `section-N` to `section:blocks[N]`, re-keying the interactive-step, collapse, acknowledged and done namespaces; a probe confirmed `sweepDiscardedProgressRecords` cannot reach the orphans because they remain well-formed. Release timing is what makes this cheap: `git merge-base --is-ancestor 3639d02f v2.17.0` is false, so #1864 - which introduced the current key shape and whose sweep discards the prior one - is unreleased. Under a co-release a shipped user's pre-#1864 progress is discarded by #1864 regardless and `section:blocks[N]` is the first shape they ever see, making incremental orphans zero. The reportable part is the condition: if #1864 ships in release N and this in N+1, the window opens against a population no earlier reviewer measured - 214 of 576 published section blocks carry no id (37%), across 60% of files, where the bundled library's 21-of-21 figure is unrepresentative. Loss is bounded to in-guide position; marks, badges, streaks and path credit key off contentKey or guideId.
   Required: State in the PR body that this must ship in the same release as #1864, since #1864's sweep is what makes the re-keying free. If they might ship separately, extend `isDiscardedProgressKey` to sweep the pre-#1864 shape so the orphans are removed rather than accumulating and feeding the evidence bridge as signals that resolve to position 0.

8. [blocking] **owd-emitted-guard-remote-tombstones — Every remote write of the new completion guard is dead weight in the shared per-user object, and its deletions leave tombstones** (low · reversibility-and-one-way-door)
   `markEmitted` goes through `createUserStorage()`, so each key enters the single per-user `user-storage` resource shared with every other Pathfinder namespace - and `keysToSync` is five fixed keys with no prefix scan, so the namespace is never read back while `isEmitted` reads raw localStorage. One hundred percent of those remote writes can never be consulted. Deletions cannot be deletions either: `GrafanaUserStorage` exposes only `getItem`/`setItem`, so `removeItem` writes an empty-value tombstone, and there is no bulk clear. A path reset fans out to 42 sequential remote PATCHes for a six-member path, most for keys that never held a value. Sizing caps the severity - about 110 bytes per identity with no remote companion doubling, needing roughly 13,000 completions to matter - and the namespace is unreleased, so nothing is stranded today and a pre-release change leaves nothing behind.
   Required: Write the guard with `createLocalStorage()` rather than `createUserStorage()`; its own doc comment says it only needs to survive a reload, not a device change. Sweep it with a direct `localStorage.removeItem` as the progress namespaces already justify, and gate the reset fan-out on a guard key that actually exists so a reset stops writing for combinations that never occurred.

Follow-ups:

1. [follow_up] **design-decision-2-derivation-claim-stale — Decision 2 of the completion model is now false, and contradicts this PR's own edit to the contract anchor** (medium · completion-records)
   `COMPLETION-MODEL.md:193-197` still states that the mark is not wired to the evidence arithmetic and that "`guideProgress` / `furthestEvidencedPosition` still have no production caller. Connecting the two is the later derivation work item, not this one." This PR is that work: `guideProgress` had only a barrel re-export at base and is called at `completion-store.ts:349` at head. The PR edited the four lines immediately above this passage and left it, so the paragraph reads as freshly reviewed while being wrong - and this PR's `CONCERN_DETAILS.md` edit asserts the opposite, so two design documents in one PR now disagree. A skeptic looked for a narrow reading that saves it and found none, since the clause names both functions by identifier. The half still true is that the mark remains a short-circuit rather than a `mark-guide-complete` signal, leaving that evidence kind unexercised.
   Follow-up: Rewrite the paragraph to state what is now true: the store bridges step and section evidence into `guideProgress` against the frozen per-key index, and the mark remains a short-circuit ahead of that arithmetic, so `progress.ts`'s `mark-guide-complete` kind still has no producer. Keep that residual as the named remaining work item if it is still wanted, and mention the design-doc update in the PR body.

2. [follow_up] **frozen-index-telemetry-claims-are-dead-or-false — active-guide-index carries a false telemetry claim, a self-contradictory docstring, an unproduced enum variant, and lost the sweep evidence that justified the snippet-ref gap** (low · completion-records)
   `active-guide-index.ts:31` says `denominatorSource` is "Reported in telemetry", but nothing reads it - one write, one declaration, 21 test lines - so it is a write-only dead field asserting that a divergence is observable when no telemetry exists. The docstring contradicts itself: line 5 says the index runs over the pre-inlining tree while :12-13 say that tree is gone on the launch path, and :20's claim that freezing "removes both" removes only the intra-session half, since `prepare-guide-launch.ts` hands the seam an already-expanded tree. `live-post-inlining-degraded` has zero producers. And the PR deleted the base comment recording that two sweeps found no `snippet-ref` anywhere across 598 published guides, so nothing now documents why the class is uncovered. The behavioural divergence itself is unreachable today - zero bundled guides use `snippet-ref`, and the inliner returns the original array by identity when there are no refs.
   Follow-up: Derive the label from `guideHasSnippetRefs(rawGuide)`, already computed two lines above, so the two arrival shapes are distinguishable; fix the docstring's two wrong clauses; either emit `denominatorSource` or drop the telemetry claim and the unused variant; and restore the sweep-evidence note somewhere durable.

3. [follow_up] **owd-index-publish-no-notify — publishGuideIndex notifies nobody, so a refresh that lands before the index publishes is silently dropped with no re-arm** (low · reversibility-and-one-way-door)
   `computeGuideProgress` returns `undefined` when no index has published, and `refreshGuidePercentage` then returns without writing, announcing or scheduling a retry. Only eviction notifies - `publishGuideIndex` just stores the entry - and the publish sits in a passive effect that can lose the race to hydration's microtask. The percentage is the store's sole product, yet the arrival of its denominator is the one input that produces no signal. Today hydration's own notify and the widened section-complete refresh make a durable stuck-at-zero hard to construct, so this is a gap in the store's contract rather than a live break - but it is the mechanism behind the failed self-heal in the old-rule blocker.
   Follow-up: Have `publishGuideIndex` notify the content key it published for - it already knows it - and make `refreshGuidePercentage`'s undefined branch observable rather than silent.

4. [follow_up] **percentage-owners-not-routable-to-their-concern — Four files this PR makes load-bearing for the completion contract are not trigger paths for that concern** (low · cross-cutting-architecture)
   The rewritten anchor names six percentage owners, but `CONCERNS.md`'s `completion-records` row lists trigger paths that omit four of them - `active-guide-index.ts`, `path-member-join.ts`, `learning-journey-helpers.ts` and `learning-paths.hook.ts` - and the row is unchanged. Subsystem rows never activate on paths alone and the row's keywords omit `calculatePathRollup`, `journeyProgressFromMilestones`, `publishGuideIndex` and `peekGuidePercentage`, so a future PR editing only the path rollup raises no signal for this concern and gets always-on review with none of the denominator questions or the contract anchor. Separately `src/learning-paths/**` appears in exactly one row and `src/components/LearningPaths/**` and `src/components/mark-complete/**` in none, while this PR changes all three.
   Follow-up: Add the four named owners to the `completion-records` trigger paths with the narrowest real contract keywords that co-fire on them, in this PR since it is what made them load-bearing. Consider whether the two component directories warrant bounded context on an existing row.

Suggestions & nits:

1. [suggestion] **global-revision-subscription-on-heaviest-subtree — The content area subscribes to a global progress revision to keep a loading-skeleton bar fresh, re-rendering the whole subtree on every step** (medium · docs-retrieval-and-rendering)
   `DocsPanelContentArea.tsx:127-131` calls `useSyncExternalStore` unconditionally at the top of the component, but the only value it guards sits inside the `isLoading` branch, which is not rendered once content has loaded. The revision is one module global bumped on every `kind: 'guide'` event - i.e. every step write - so the blast radius is the whole `ContentRenderer` subtree, the largest render in the plugin, driven by the highest-frequency event in the app. Nothing else re-rendered this component per step at base. The rollup readers also do a `localStorage` read plus `JSON.parse` per call, amplified on My Learning where the path list filters and sorts by `getPathProgress`, so the comparator alone is two `peekAll()` calls per comparison where base read no storage at all.
   Suggested: Extract the loading-state milestone bar into its own component and move the subscription into it. Separately hoist one `peekAll()` per render into the hook, memoised on the revision, and pass the record into `calculatePathRollup`, which is already pure and takes it as a parameter.

2. [suggestion] **rollup-completeness-boundary-untested-and-doc-ambiguous — An excluded member cannot clear the rollup's complete flag, the boundary is untested, and the design table's two columns disagree about whether it should** (low · completion-records)
   `rollup.ts` clears `complete` only from a value present in the array, and the join drops excluded members before the mean, so a member nobody can answer for vanishes from the question rather than preventing completion. Reproduced twice - the `>= 100` gate is clearable with a member excluded. Two doors reach it: an unresolvable member, and a locked member, since the journey mean filters to unlocked while `totalMilestones` stays locked-inclusive, which is live today for a published path with an unresolvable member. Neither mints anything durable - the badge route needs a `bundled:` base that only in-repo content produces, and all ten bundled manifests are plain guides - and the design permits the code's reading, since `COMPLETION-MODEL.md:42`'s left column already says "resolvable". But `rollup.test.ts` covers only `[100,100]` and `[100,99]`, and that table cell was read two different ways by three reviewers in this review alone.
   Suggested: Pick a reading and pin it: make the table's two columns agree, either "every resolvable milestone at 100%" to match the code or have `complete` require `resolvedPercentages.length === totalMembers`. Add the rollup test either way, and state the rule once for both doors - unresolvable and locked.

3. [suggestion] **preview-denominator-ignores-body-edits — The block-editor preview's denominator is frozen at first paint and does not track guide-body edits** (low · block-editor-authoring)
   The preview is keyed on `block-editor://preview/${guide.id}`, which is authored metadata and stable across add, remove, reorder and rewrite, and `publishGuideIndex` ignores a later publish. Three probes confirmed it: a one-section guide acknowledged reads 100% and still reads 100% after five blocks are appended, where the honest figure is 16%. Bounded three ways - author-only and never persisted, never reaching a badge, rollup or telemetry; cleared by the "Reset guide" affordance that the header renders whenever there is preview progress, always co-present with the wrong number; and cleared by any reload. Not a regression: at base the preview showed a permanent 0% whatever the author completed, so this PR makes the number real for the first time. The PR body's "fixed it going stale" is imprecise rather than unsupported - the numerator is now live, the denominator is fixed at first paint.
   Suggested: Track the preview index identity to the content rather than to `guide.id` - a hash of the serialised guide - or have the editor call `evictContentCache(previewKey)` when the guide body changes, which already drops both the index and the preview ack map. Also tighten that clause of the PR body.

4. [suggestion] **member-exclusion-not-observable — Nothing reads excludedCount, so the PR body's claim that the exclusion count makes the gap visible is not true of the shipped code** (low · analytics-and-telemetry)
   The join computes `excludedCount` and `excludedMemberIds`; both new callers destructure past them and no production code reads either. The design-conformance framing of this was adjudicated and dropped - decision 9's normative sentence opposes "counted" to "scored zero" rather than requiring surfacing, the "visible instead of silent" line sits under a heading that declares itself rationale, the count is absent from the doc's own load-bearing list, and both surfacing-flavoured sentences predate this PR. What stands is narrower: the PR body repeats the visibility claim, and the strongest argument for the counter is prospective - `learning-paths.hook.ts:103-111` records a mass-exclusion hazard where "one finished module out of six would read 100%", and the count is the only signal that would distinguish such a regression from readers genuinely behaving differently.
   Suggested: Add a `logger.warn` from each rollup when `excludedCount > 0` carrying the path id or base URL and the counts but no member ids, which rides the free logger bridge into Faro at zero instrumentation cost - or soften the PR body sentence to match what ships.

5. [suggestion] **completion-percentage-meaning-change-unannounced — completion_percentage changes meaning on ten emit sites with nothing telling a downstream analyst when it changed** (low · analytics-and-telemetry)
   `calculateJourneyProgress` is deleted and `completion_percentage` now carries the evidence mean on every learning-journey interaction, including four sites whose payload meaning changes with no diff line at all. The reviewer has accepted the semantic change itself as desirable, and the partitioning worry does not hold: `reportAppInteraction` injects `plugin_version` unconditionally into every event, so the change lands in one version bump, and `current_milestone` and `total_milestones` still ship unchanged, so the old series is recomputable term-for-term as `round(current_milestone / total_milestones * 100)`. Nothing needs cleanup. What remains is that a dashboard trending the field will show an unexplained step at the release, and the PR body tags two other items release-note material while omitting this one.
   Suggested: One line in `docs/developer/TELEMETRY.md` and in the release notes naming the version at which `completion_percentage` on learning-journey events changed from navigation position to earned progress. A `completion_basis` discriminator is unnecessary.

6. [suggestion] **locked-segment-state-missing — A locked milestone's segment can never fill while the bar is denominated over locked-inclusive totalMilestones** (low · docs-retrieval-and-rendering)
   The fill set comes from unlocked milestones only, while `segments` is built over locked-inclusive `totalMilestones` with no locked state, so a journey can read 100% beside a permanently grey segment. Head is strictly more honest than base here - base filled a segment whenever the reader paged past it with no lock check, painting an unpublished module solid green in the success colour - and the PR body already says segment fill now reflects earned credit. Reachable but rare: `isLocked` is set only on a failed package resolve, which includes a transient resolver failure, so a bundled or CDN package path can hit it in an ordinary session. Cosmetic, on a bar whose label is explicitly a position breadcrumb, and the locked state is already legible on the cover page and the context panel.
   Suggested: Give locked milestones a distinct third segment state so the gap is self-explanatory. Prefer that over re-denominating the bar, which would hide the unpublished module rather than explain it.

7. [suggestion] **owd-section-completed-ordinal-unaddressable — section-completed: can no longer address a section without an author id, by ordinal or by path** (low · reversibility-and-one-way-door)
   `section-completed-check.ts` prefixes a raw id with `section-` unless it already starts that way, so an id-less section's new runtime id `section:blocks[N]` resolves to the nonexistent `section-section:blocks[N]`. The pre-PR render-order id made `section-completed:1` resolve; it no longer does, and nothing warns, so such a guide fails its requirement silently forever. A search across `src/`, `tests/`, `docs/`, the bundled guides and every tracked JSON found zero callers of the ordinal form, so this is a documentation and future-authoring concern rather than a live break - but `STEP_MODEL.md:19` still frames explicit ids as advice.
   Suggested: Upgrade `STEP_MODEL.md:19` from recommended to required for `section-completed:` targets, and consider having the check log a warning when the resolved id matches no element and no stored record, so a broken requirement is visible.

8. [nit] **redundant-double-percentage-refresh-per-section — Widening the acknowledgement refresh double-refreshes the percentage for a section that has interactive steps** (low · correctness-and-reliability)
   The refresh was previously gated on `gateAnalysis.isAllPassive` and now runs for every section shape. For a section with interactive steps, completing the last step already ran `persistSection`, which refreshes - so the section's completion effect refreshes a second time, costing one extra `interactiveCompletionStorage.set` (a localStorage write plus a queued remote write) and one extra global revision bump that wakes every rollup subscriber. Verified harmless: both writes compute the same number and the effect is guarded against looping.
   Suggested: Skip the refresh when the section has interactive steps, keeping the widened call for the all-passive and mixed-ack shapes that genuinely bypass `persistSection`.

9. [nit] **reset-manufactures-guard-key-never-written — Every single-guide reset manufactures a guard key and a timestamp companion for a key that was never written** (low · reversibility-and-one-way-door)
   `invalidateEmittedCompletion` iterates both completion kinds unconditionally, but the two are mutually exclusive per identity - `recordGuideCompletion` bails on a path or journey manifest and `recordJourneyCompletion` fires only on a full milestone set - so one of the two is always guaranteed absent. The hybrid `removeItem` writes a deletion-timestamp companion unconditionally, and the single-key `clear` has no companion sweep, so about 404 bytes per reset is created for an identity that cannot exist. Provably inert rather than harmful, since nothing reads either key.
   Suggested: Skip the clear when the key is not present - `isEmitted` is already a synchronous read - or add the companion sweep to `clear(dedupeKey)` as `clearAll` already does.

10. [nit] **guide-source-list-hand-maintained-untested — KNOWN_PATH_MEMBER_GUIDE_SOURCES is hand-maintained with no test that fails when a new guideSource appears** (low · completion-records)
    The list drives which guard keys a path reset lifts, and its own comment is candid that a member whose real source is absent simply does not get its guard lifted. The set genuinely cannot be derived structurally - `repository` is typed `z.string()` with a default, so a manifest's value is arbitrary remote content - which fairly exempts the derivation. It does not exempt the tripwire: nothing pins the list against the sources first-party code actually produces, so adding a fourth silently reintroduces the reset-then-re-mark defect for every member under it.
    Suggested: Add a test asserting the list covers every `guideSource` the first-party paths can produce - the `resolveCompletionIdentity` fallback, the backend-guide loader's literal, and the package schema default - so a fourth fails the build rather than silently missing its guard.

11. [nit] **test-reset-omits-progress-revision — resetCompletionStoreForTests does not reset the new guideProgressRevision** (low · testing-and-verification)
    The revision is a new module-level counter in `progress-events.ts`, bumped by every guide progress event and every cleared event. `resetCompletionStoreForTests` clears seven other pieces of module state and calls `evictAllGuideIndexes`, but nothing resets the revision - it lives in a different module with no exported reset. Every other module-level counter this PR touches has a test reset path.
    Suggested: Export a test-only reset from `progress-events.ts` and call it from `resetCompletionStoreForTests`, or have tests assert deltas rather than absolute values.

12. [nit] **compute-guide-progress-undefined-comment — computeGuideProgress's doc comment describes an undefined return that does not exist** (low · correctness-and-reliability)
    The comment says it returns `undefined` when the index has not published "or the content has no countable blocks at all". Only the first is true: a guide-shaped JSON with `blocks: []` still publishes, so the index answers, the position calculation returns zero-of-zero, and a 0 is persisted rather than the function returning early. Harmless at runtime since a stored 0 is evictable, but this is the file the contract anchor points at for the producer's behaviour, so a reader reasoning about which content keys get a persisted percentage will get it wrong.
    Suggested: Drop the second clause, or guard the publish on a non-zero block count so the comment becomes true.

13. [nit] **owd-step-model-doc-stale-guideprogress — STEP_MODEL.md still documents a GuideProgress export that this PR deleted** (low · reversibility-and-one-way-door)
    Line 141 lists `GuideProgress` among `completion-store.ts`'s exports and adds a caveat about two distinct same-named models and importing the one for your layer. The interface and `getGuideProgress` are removed in this PR, leaving `src/lib/guide-stats`' as the only one - so an agent following that line writes an import that does not resolve, and the two-models framing is now actively wrong given the PR's whole point is that there is one. The PR updated two other lines in the same file correctly and missed this one.
    Suggested: Drop `GuideProgress` from the completion-store list and delete the same-name caveat sentence.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1866
Summary: base guide, path and journey progress on evidence, not navigation
Verdict: Request Changes
Results: 8 blockers, 13 non-blocking findings, 4 follow-ups

tomglenn:
All eight blockers actioned — head `ee28ac0e`, 36 checks passing, nothing failing.

Six were code fixes. Two were handled without code changes and both are judgement calls, so they are flagged first rather than buried — overrule either and it gets done.

**6 (percentage-ratchet-overstated) — prose corrected, test untouched.** Reading the gate, it does exactly what its own "Loose on purpose" comment says: a tripwire on the single-line `(x / y) * 100` shape with a keyword-gated denominator name. The comment is accurate; the overclaim was elsewhere. `CONCERN_DETAILS.md`'s `completion-denominator-authority` anchor said the calculation "fails the build" unqualified, and the PR body repeated it. Both now describe what the gate actually does. The adjacent claim that guide-stats is the only place the stored result is read or written was also softened. If you would rather the gate were strengthened to match the original wording, say so and it will be.

**7 (owd-idless-section-progress-orphaned) — took the release-ordering statement, not the defensive sweep.** The second option was examined and declined: the old counter-based key shape (`section-N`) is not reliably distinguishable from a genuine author id that happens to be a bare number — a section with `id: "1"` produces an identical key under both conventions — so a pattern-based sweep risks discarding real, intentionally-id-ed progress to clean up orphaned progress. Given the finding is bounded to in-guide position with nothing durable at stake, the PR body now states the release-ordering requirement: this must ship in the same release as #1864. If you would accept a narrower sweep on some signal we have not spotted, that is a better outcome than an ordering constraint and worth hearing.

The six code fixes, each with the condition traced from the real caller rather than confirmed by the fix being present:

**1 (section-registry-third-deriver-double-count)** — `documentLayout` now applies the same `props.sectionId ?? id-based ?? counter` precedence `InteractiveSection` uses, so an id-less section registers once. Test asserts one registry entry per section after a full render with an id-less section.

**2 (reset-guard-identity-divergence)** — the gate was the bug, so it is gone. `resolveActiveMilestoneSlug` no longer consults `contentType`; it derives milestone identity from `journeyBaseUrl` plus the URL-derived slug, neither of which a caller can misclassify. Traced `prepareGuideLaunch`'s own routing: `isLearningJourneyUrl` returns false for a `backend-guide:` scheme URL, which is exactly what `MyLearningTab.tsx:301-303` launches a path member with — and the writer never consulted `contentType` at all, which is why the two sides could disagree. New test mirrors that exact manifest shape, omits `contentType` entirely, and reddens against the previous version. Second half also fixed: `resetPath` depended on `path.guides`, empty until an awaited fetch resolves for URL-based paths, so per-milestone guards went uninvalidated; slugs are now recovered from the already-computed `milestoneKeys` prefix scan. Test calls `resetPath` without awaiting the fetch and asserts the guards are invalidated anyway.

**3, 4, 5, 8** — the healing gap for pre-existing percentages, the recommendation-card journey divergence, the guard outliving a dropped record, and the emitted guard no longer riding remote storage (now `createLocalStorage()` end to end; traced — `isEmitted` already read raw localStorage and the sync allowlist never covered that namespace). Each with a reddening test.

Follow-ups, suggestions and nits from the review are not addressed in this round and are recorded on our side.

src/components/docs-panel/hooks/resetGuideProgress.ts
```diff
@@ -1,17 +1,67 @@
...
+    ? resolveMilestoneCompletionIdentity({
+        repository: identity.repository,
+        packageManifest: identity.packageManifest,
+        milestoneSlug: identity.milestoneSlug,
+      })
+    : resolveCompletionIdentity({
+        packageManifest: identity?.packageManifest,
+        repository: identity?.repository,
+        fallbackId: fallbackGuideIdFromContentKey(contentKey),
+        fallbackSource: 'bundled',
```
Jayclifford345: [P2] Match the standalone writer’s fallback source when resetting

The remaining Bugbot finding reproduces at this head. For a standalone guide with `packageManifest: { id: "g", type: "guide" }` and no repository, `recordStandaloneGuideCompletion` uses `interactive-tutorials`, while this reset clears `bundled`. With the real recorder and localStorage, completing → resetting → clearing recorder memory to simulate a reload → completing again emits only one record in total: the original durable guard survives. Please share the standalone writer’s identity derivation here, retaining the bundled fallback only for bundled content, and add a reset/re-completion regression covering an omitted repository as well as an explicit one.

Jayclifford345 (dismissed):
Focused final review of d1137b34, including the previous feedback and subsequent fixes. One remaining reset identity mismatch, confirmed below with the real recorder and localStorage. The 143 existing focused tests passed; the added reset-and-recomplete reproduction failed (expected two emitted records, received one). CI Gate is green. No new advisory findings.

src/components/docs-panel/hooks/resetGuideProgress.ts
tomglenn: Fixed at 6f513324. Shared the standalone writer's identity derivation with the reset path rather than retaining separate fallbacks: extracted `resolveBundledGuideCompletionIdentity` and `resolveStandaloneGuideCompletionIdentity`, each hard-coding its own fallback source with no argument either caller can override, and switched `resetGuideProgress`'s non-milestone branch to call whichever one matches the actual writer for that content key (the same `bundled:` prefix check the surface emitter already uses to choose between the two writers). `recordBundledGuideCompletion` and `recordStandaloneGuideCompletion` both now go through the same functions.

Added the regression you asked for on both shapes (manifest with an id and no repository, and one with an explicit repository) — mark, reset, re-mark, assert a durable record on the second completion for each. Reddened the omitted-repository case against the pre-fix code first.

tomglenn: Fixed in `6f513324` — and fixed as a class rather than as a one-liner, because this was the fourth instance of it in this PR.

You and Bugbot caught the same thing from different angles: the reset passed `fallbackSource: 'bundled'` while `recordStandaloneGuideCompletion` omitted it entirely and fell through to `DEFAULT_GUIDE_SOURCE` (`'interactive-tutorials'`). For a standalone guide whose manifest carries an id but no repository the two diverge, the durable guard is lifted under the wrong identity, and the second mark is swallowed.

Matching the writer's fallback at the reset site would have closed it. It is not what landed, because the same defect had already appeared three times on two different axes — `guideId` for milestones, then `guideId` for path members, now `guideSource` for standalone guides. As long as either side can pass a fallback, a later caller passes a different one. So standalone guides now get one derivation with the defaults inside it and no argument a caller can vary, the same treatment milestones got earlier in this PR.

Regression test uses the exact shape that slipped past three previous fixes and two reviewers — manifest with an id and no repository — asserting mark, reset, re-mark produces a durable record the second time. It reddens against the previous code.

Two remaining places still derive an identity with their own arguments, both the same shape, neither changed here:

- The whole-journey `journeyIdentity` for the `journey_completed` trigger still passes `fallbackSource: 'bundled'`. Left alone because journeys have no established comparison writer the way guides have `recordBundledGuideCompletion` / `recordStandaloneGuideCompletion`, so the correct default cannot be established without a real journey-manifest-without-repository case — and guessing at an identity is what produced four instances. Not reachable today: `resetPath`'s guard lift tries all three known sources.
- `fallbackGuideIdFromContentKey` strips a trailing `/content.json` that `persistJourneyCompletionPercentage` does not, so a reset landing on the package form could derive a different `guideId`. How reachable that form is at reset time rather than at load is unconfirmed.

Both are tracked. If you have a view on the right journey fallback, that closes the last one properly rather than by inference.

CI green on this head, including Bugbot running to completion rather than skipping.

Jayclifford345 (approved):
Approved pending last workflow changes

PR #934, opened 2026-05-27
Agent hardening Phase B: close the ratchet loopholes

moxious:
## Summary

Second PR in the agent-hardening sequence (`docs/design/AGENT_HARDENING.md`). Phase A — "stop lying to the agent" (#903) — made the prose accurate. This PR closes five known gaps in the existing tier ratchet so the boundaries it claims to enforce are actually enforced everywhere.

Deferred per the design doc: **B1 + B2** (mechanical `/review` + `prevent-doc-drift` Stop-hooks — open design questions per the doc), and **B3** (concentration ratchet — wants B5's sync infrastructure first).

## Per-commit map

| Commit | Item | Finding | What it closes |
| ------ | ---- | ------- | -------------- |
| `dd248565` | **B4** | F-6 | Root-level `src/` files (`module.tsx`, `constants.ts`, …) now have explicit tier assignments via `ROOT_LEVEL_TIER_MAP`. A new `helpers.ts` at root without a tier entry fails CI. `ROOT_LEVEL_ALLOWED_FILES` is now derived from the map (one source of truth). |
| `60d73ac3` | **B5** | F-5 | New `architecture.test.ts` describe-block parses `eslint.config.mjs` and asserts the tier-constant arrays + `tierBoundaryConfig` source-dir lists match `TIER_MAP`. The test caught pre-existing drift: `hooks` missing from `TIER_2_ENGINES`, `recovery` missing from `TIER_1_PLUS` and the Tier 1 source list — all fixed in the same commit. |
| `c9ae837f` | **B6** | F-7 | Import scanner now reads `tsconfig.compilerOptions.paths` and resolves aliased specifiers to src-relative paths via `resolveAliasedSpecifierWithConfig`. Repo has no aliases today so the behavior is unchanged; the first PR to add one will get ratchet coverage automatically. |
| `77590775` | — | — | Prettier formatting trailing from B6's lint-staged hook. |
| `18fa5fc6` | **B8** | F-12 | `jest.config.js` flipped to `collectCoverage: false`. `npm run test:ci -- <focused>` no longer trips the global threshold; `npm run check` now routes through `test:coverage`, which keeps threshold enforcement intact on the full gate. |
| `98b9fcba` | **B7** | F-9 (partial) | `.husky/pre-commit` now runs `typecheck` + a new `test:governance` subset (architecture + import-graph) after lint-staged. Measured ~8s total on a clean tree. |

## Test plan

[...]

PR #1868, opened 2026-09-11
docs(husky): correct pre-commit rationale

### LINKED ISSUE #1860: Pre-commit hook comment claims git was skipping it; false under husky's default hooksPath layout
`.husky/pre-commit` (as of #1823) carries this justification:

> `npm run typecheck` (25s) and `npm run test:governance` (27s) are whole-project; a 52s commit gets bypassed with `--no-verify`, which is no better than the hook git was silently skipping.

The incentive argument is sound and #1823 is the right change. The trailing clause is not: **whether git was skipping the hook depends on `core.hooksPath`, and under husky's own default it was not.**

## Evidence

husky 9.1.7's `prepare` sets `core.hooksPath` to `.husky/_` (`node_modules/husky/index.js`: `core.hooksPath', ${d}/_`). Git then executes the mode-755 shim `.husky/_/pre-commit`, which sources `.husky/_/h`. That shim gates on:

```sh
[ ! -f "$s" ] && exit 0
...
sh -e "$s" "$@"
```

That is a **file-existence** test, not an executable-bit test, and the hook is then invoked through `sh -e`. So on the default layout the tracked mode `100644` is irrelevant and the three-command hook **was running**.

The claim holds only under the legacy `core.hooksPath=.husky`, where git executes `.husky/pre-commit` directly and refuses a non-executable file. That layout still exists on developer machines set up under older husky versions.

## Why it matters

The sentence lives in tracked source, so it is what the next reader reasons from. It makes #1823 look like it revived a dead gate when in fact it **narrowed a live one** — the opposite framing, and the one that matters when deciding whether to restore `typecheck` / `test:governance` to the hook later.

It also silently reverses `docs/developer/AGENT_HARDENING.md:121` item B7, which is not mentioned in #1823.

## Suggested fix

1. Reword the comment in `.husky/pre-commit` to drop the "silently skipping" clause and keep the incentive argument, which stands alone. Optionally note that the mode fix matters for legacy `core.hooksPath=.husky` checkouts.
2. Reconcile `AGENT_HARDENING.md:121` item B7 with the new hook contents.

Both are ours to fix, not the contributor's — the premise came from how the work was framed upstream, and #1823 implemented that framing faithfully.

## Provenance

Delegated follow-up from the round-2 review of #1823 (finding `hook-mode-premise-scoped`, medium). Shim behaviour verified directly against husky 9.1.7 at `3e7b2567`.

Nikshay1:
## What does this PR do?

Corrects stale documentation around the Husky pre-commit hook.

The current comment says a ~52s pre-commit gate is "no better than the hook git was silently skipping." That overstates the executable-bit problem: under Husky 9's default `.husky/_` hooksPath layout, the shim invokes `.husky/pre-commit` through `sh`, so the tracked hook was not skipped solely because it was mode `100644`.

This PR:

- removes the inaccurate "silently skipping" rationale while preserving the reason for keeping whole-project checks out of pre-commit
- updates the F-9 hardening note to reflect that pre-commit now runs per-file lint/format checks through `lint-staged`
- records B7's actual history: #934 added `typecheck` and `test:governance`, then #1823 deliberately narrowed the hook after those checks measured ~52s combined

The executable-bit fix from #1823 remains unchanged because it still matters for legacy `core.hooksPath=.husky` setups.

No runtime behavior changes.

## Linked issue(s)

Closes #1860

## Verification

- `git diff --check`
- `npx prettier --check docs/design/AGENT_HARDENING.md`

No automated test was added because this is a comment/documentation-only correction.

## Checklist

- [x] This PR addresses a **single concern**.
- [x] All commits are signed.
- [x] Tests are not required for this documentation-only change.


moxious (approved):
Thanks for correcting this rationale — it is a small diff that removes a genuinely misleading claim, and you picked the right half to keep.

The old comment's assertion that git was "silently skipping" the hook was conditional on `core.hooksPath`, not universal: under the legacy `.husky` layout git refuses a non-executable hook, but under husky 9's `.husky/_` default the shim gates on file existence and the mode bit is irrelevant. Your replacement makes no layout claim at all, so it reads correctly under both — and the incentive argument it keeps (a 52s commit gets bypassed with `--no-verify`) stands on its own. The `#934` → `#1823` attribution in the B7 row checks out, and the ~52s figure matches #1823's measured 25s + 27s exactly.

---

PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1868
Summary: correct pre-commit rationale
Verdict: Approve

---

CI is green on this head. One `publish-report` check shows a failure, but that is a known fork-permission limitation on community pull requests — a report job cannot write its comment from a fork — and not anything in your change. Your `Closes #1860` will retire the issue on merge. A maintainer will take it from here.

PR #1871, opened 2026-09-12
fix(scripts): quote Authorization header in curl config

### LINKED ISSUE #1869: upsert scripts send no Authorization header: ap_auth_init writes an unquoted curl config value
## Summary

`ap_auth_init` in `scripts/lib/app-platform.sh:54-57` writes an **unquoted** header into a curl config file. curl terminates an unquoted config value at the first whitespace, so it never receives the token — and the truncated value it does receive is curl's syntax for *removing* the `Authorization` header. Every request from `upsert-guide.sh` and `upsert-learning-path.sh` therefore goes out unauthenticated and 401s.

Both scripts route all traffic through `ap_curl`, so neither can authenticate against the aggregator. These are the two scripts `docs/developer/EXTERNAL_API.md` documents as the supported external guide-import path.

## The code

```bash
ap_auth_init() {
  AP_CURL_CONFIG=$(mktemp)
  printf 'header = Authorization: Bearer %s\n' "$1" >"$AP_CURL_CONFIG"
}
```

This produces a config file containing:

```
header = Authorization: Bearer glsa_xxx
```

curl parses that as option `header` with the value `Authorization:`, discarding the rest of the line. A header given as a bare name plus colon with no content is curl's instruction to *remove* that header rather than send it. Net effect: no `Authorization` header on the wire.

The comment directly above the function asserts the opposite, which is presumably why it was written this way:

> Unquoted on purpose: curl takes the rest of the line verbatim, which quoting would subject to backslash escaping.

That is not how curl's config parser works. The whitespace-termination rule applies, and quoting is required for any value containing whitespace.

## Reproduction

Against a Grafana Cloud stack with the aggregator enabled, using a valid Editor-role service-account token:

```bash
source scripts/lib/app-platform.sh
ap_auth_init "$PATHFINDER_SA_TOKEN"
ap_detect_namespace <stack>.grafana.net
# curl: (56) The requested URL returned error: 401

ap_curl -o /dev/null -w '%{http_code}\n' "$(ap_guides_url <stack>.grafana.net stacks-XXXXX)"
# 401
```

The same token returns 200 when passed with a direct `-H`.

A `-v` dump of a request built from that config file lists `GET`, `Host`, `User-Agent` and `Accept`, and no `authorization` line at all.

Syntax matrix against `/api/org` on the same stack with the same token:

| config file line | result |
| --- | --- |
| `header = Authorization: Bearer X`  *(current code)* | **401** |
| `header = "Authorization: Bearer X"` | 200 |
| `header: Authorization: Bearer X` | **401** |
| `header "Authorization: Bearer X"` | 200 |
| `--header "Authorization: Bearer X"` | 200 |

## Why this was not caught

Two layers hide it:

- **The tests stub curl.** `scripts/tests/*.test.sh` replace `curl` with a shell stub, so no test exercises real config-file parsing. A regression test has to invoke a real curl against a real config file.
- **The error message misdirects.** `upsert-guide.sh` calls `ap_detect_namespace` with `|| true`, so the 401 surfaces as `could not auto-detect namespace from /api/frontend/settings; pass --namespace explicitly`. Passing `--namespace` gets you to `Authentication failed (HTTP 401). Check the service-account token.` — which blames the token. Nothing points at the config file.

## Fix

Quote the value in `ap_auth_init`:

```bash
printf 'header = "Authorization: Bearer %s"\n' "$1" >"$AP_CURL_CONFIG"
```

The security property the comment is protecting — keeping the token out of the process table for the length of a multi-resource run — is preserved by the quoted form, since the token still never reaches curl's argv. Update the comment at `scripts/lib/app-platform.sh:50-53`, which currently documents the broken rationale.

Worth adding a test that runs a real curl against a real `ap_auth_init` config file and asserts the header arrives, rather than extending the stub-based suite.

## Scope of verification

Reproduced on curl 8.7.1 (Apple system curl, `x86_64-apple-darwin25.0`, macOS). The whitespace-termination rule is longstanding documented curl behaviour rather than a version quirk, so this is very likely broken on every platform — but only the one curl version was tested, so treat "broken everywhere" as inference.

Found while verifying whether Terraform can replace these scripts for #1233.

Nikshay1:
## Summary

Fixes authentication for the external API upsert scripts by quoting the `Authorization` header value written to curl's config file.

`ap_auth_init` currently writes:

```text
header = Authorization: Bearer <token>
```

curl does not treat the rest of an unquoted config value as verbatim text. The value is terminated at the first whitespace, so curl parses the line as an `Authorization:` header without the bearer token.

For curl, a bare header name followed by `:` means that the header should be removed. As a result, requests made by both `upsert-guide.sh` and `upsert-learning-path.sh` are sent without an `Authorization` header and fail authentication.

The config now writes:

```text
header = "Authorization: Bearer <token>"
```

so curl parses and sends the complete header.

## Why

Both upsert scripts use the shared `ap_auth_init` / `ap_curl` path from `scripts/lib/app-platform.sh`, so the bug affects authentication for both supported external guide upload flows.

The surrounding comment also documented the previous value as intentionally unquoted. That comment was incorrect according to curl's config parsing behaviour, so this PR updates it alongside the fix.

The token continues to be stored in the temporary curl config file rather than passed through `-H` or another command-line argument. This preserves the existing security property of keeping the service-account token out of curl's process arguments during a multi-resource upload.

## Regression coverage

The existing upsert tests use a stubbed curl implementation. That is useful for validating request construction, but the stub does not run curl's real config parser, which is why the malformed config value was not caught.

This PR adds `scripts/tests/app-platform.test.sh`, which exercises the config with real curl:

1. Starts a temporary HTTP server bound to `127.0.0.1`.
2. Calls `ap_auth_init` with a sentinel token.
3. Makes a request using real curl and the generated config file.
4. Captures the `Authorization` header received by the server.
5. Verifies that the value is exactly:

```text
Bearer sentinel-config-token
```

The regression test was also checked against the previous implementation. It fails with the unquoted config value:

```text
FAIL: Authorization header was not parsed from the curl config
```

and passes once the header value is quoted.

The test invokes real curl directly rather than the existing curl stub so that curl's config-file parsing is part of the test.

## How to verify

The script test suite passes:

```bash
npm run test:scripts
```

The full pre-merge check also passes:

```bash
npm run check
```

Result:

```text
check: all 9 steps passed
```

The regression test can also be run directly:

```bash
bash scripts/tests/app-platform.test.sh
```

which reports:

```text
ok   real curl receives the Authorization header from its config
```

## Breaking changes

None.

The request format, external API contract, token source, and upsert behaviour are unchanged. This only corrects how the existing `Authorization` header is serialized into curl's config file.

## Reversibility

Fully reversible. The production change is limited to quoting the curl config value and correcting the accompanying comment.

The added test has no runtime effect. Reverting the production change would restore the previous authentication failure but would not require any data migration or cleanup.

## Linked issue(s)

Closes #1869

## Checklist

- [x] This PR addresses a **single concern** (one bug fix or feature, or a set of closely-related changes).
- [x] All commits are signed.

### DIFF
--- a/scripts/lib/app-platform.sh
+++ b/scripts/lib/app-platform.sh
@@ -49,11 +49,11 @@ AP_CURL_CONFIG=
 
 # curl reads the Authorization header from a 0600 config file rather than an
 # argv -H, so the token is not exposed in the process table for the length of a
-# multi-resource run. Unquoted on purpose: curl takes the rest of the line
-# verbatim, which quoting would subject to backslash escaping.
+# multi-resource run. Quote the value because curl config values containing
+# whitespace must be quoted.
 ap_auth_init() {
   AP_CURL_CONFIG=$(mktemp)
-  printf 'header = Authorization: Bearer %s\n' "$1" >"$AP_CURL_CONFIG"
+  printf 'header = "Authorization: Bearer %s"\n' "$1" >"$AP_CURL_CONFIG"
 }
 
 ap_auth_cleanup() {
--- a/scripts/tests/app-platform.test.sh
+++ b/scripts/tests/app-platform.test.sh
@@ -1,0 +1,79 @@
+#!/usr/bin/env bash
+
+set -uo pipefail
+
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)
+REPO_ROOT=$(cd -- "${SCRIPT_DIR}/../.." && pwd)
+
+# shellcheck source=scripts/lib/app-platform.sh
+source "${REPO_ROOT}/scripts/lib/app-platform.sh"
+
+WORK=$(mktemp -d)
+SERVER_PID=
+
+cleanup() {
+  [[ -z "$SERVER_PID" ]] || kill "$SERVER_PID" 2>/dev/null || true
+  ap_auth_cleanup
+  rm -rf "$WORK"
+}
+trap cleanup EXIT
+
+PORT_FILE="${WORK}/port"
+HEADER_FILE="${WORK}/authorization"
+SERVER_FILE="${WORK}/server.js"
+
+cat >"$SERVER_FILE" <<'JS'
+const fs = require('node:fs');
+const http = require('node:http');
+
+const [portFile, headerFile] = process.argv.slice(2);
+
+const server = http.createServer((req, res) => {
+  fs.writeFileSync(headerFile, req.headers.authorization ?? '');
+  res.writeHead(200);
+  res.end('ok');
+  server.close();
+});
+
+server.listen(0, '127.0.0.1', () => {
+  const address = server.address();
+  if (!address || typeof address === 'string') {
+    process.exit(1);
+  }
+  fs.writeFileSync(portFile, String(address.port));
+});
+JS
+
+node "$SERVER_FILE" "$PORT_FILE" "$HEADER_FILE" &
+SERVER_PID=$!
+
+for _ in {1..100}; do
+  [[ -s "$PORT_FILE" ]] && break
+  sleep 0.05
+done
+
+if [[ ! -s "$PORT_FILE" ]]; then
+  echo "FAIL: local HTTP server did not start" >&2
+  exit 1
+fi
+
+TOKEN="sentinel-config-token"
+ap_auth_init "$TOKEN"
+
+PORT=$(cat "$PORT_FILE")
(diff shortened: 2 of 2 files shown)


moxious (changes requested):
@Nikshay1 in this particular case, just remove the test file, it's more weight than it's worth. Stick with the fix, remove the tests.

Nikshay1:
Removed the test file and kept the fix only. Thanks!

moxious (approved):

PR #1876, opened 2026-09-13
fix(coda): surface terminal protocol mismatches

### LINKED ISSUE #1527: Terminal protocol mismatches are invisible to the user
## Summary

A Coda terminal protocol mismatch is invisible to the learner. `onProtocolError` writes one warning to the app logger and nothing to the terminal, so a session stops behaving with no explanation on screen.

Scope is one handler. Everything else this issue originally asked for now ships inside `@grafana/coda-client`.

## Current behaviour

`src/integrations/coda/useTerminalLive.hook.ts:308-315`:

```ts
onProtocolError: ({ detail, sessionId: sid, vmId }) => {
  connectionLogRef.current.warn('Coda protocol mismatch', {
    detail,
    sessionId: sid,
    vmId,
    category: 'protocol_error',
  });
},
```

Two faults:

1. **Nothing reaches the user.** No xterm line, no `setError`, no status change.
2. **`warn` is the wrong level to be observable.** The activity gate at `src/lib/telemetry/filtering.ts:216` passes exceptions and `error`-level logs only, and drops everything else until a Pathfinder surface opens. A mismatch during connect is exactly the case that goes dark in Faro.

## The fix

`onError`, six lines above at `:272-289`, already does this correctly. Follow it:

- `connectionLogRef.current.error(...)` instead of `.warn(...)`
- `terminal.writeln()` a visible, styled line
- `setError(message)` and `setStatus('error')` so the surrounding UI reacts

Word the message so it names a framing problem. The failure mode worth not repeating is the old one, where a protocol mismatch surfaced 35 seconds later as "SSH handshake timed out — the VM may be unreachable" and blamed the VM.

This issue also records the rule in `.cursor/rules/systemPatterns.mdc`, so later agents find it: at a boundary Pathfinder does not own, do not mirror the other side's logic, and do report one visible error. Emit it at `error` level so it survives the activity gate.

## What is no longer in scope

PR #1468 replaced the hand-rolled Grafana Live protocol with `@grafana/coda-client`. Every other ask in the original body now belongs to the SDK:

| Original ask | Where it lives now |
| --- | --- |
| Validate inbound frames | SDK. Pathfinder receives a typed `SessionEvent`. |
| Add a `default` arm for unknown types | SDK. `SessionEventType` is deliberately open, and its own contract says an unrecognised type "must never be fatal", so unknown types never reach Pathfinder. |
| Give the outbound path a type, drop the `as any` | SDK `SessionInput`. |
| Keep `state` an open string, not a closed enum | SDK. `state?: string`, documented as a VM state or a backend phase. |
| Channel path is positional and duplicated | SDK. `channel` is documented opaque: split it into a Live address, never construct or parse it. |
| Mirror state tokens as Go constants, validate Go-side | Gone. `pkg/plugin/stream.go`, `terminal.go` and `wsconn.go` no longer exist. |
| Fix the `SendJSON` comments | Gone with the hand-rolled transport. |

## Notes for the implementer

- `src/integrations/coda/useTerminalLive.hook.test.ts` never mentions `onProtocolError`. This path has no test.
- `SessionEvent.code` is absent on older backends. Fall back to `error` for display, never for branching.
- The terminal is gated by the `pathfinder.coda-terminal` flag, or by dev mode plus `enableCodaTerminal` (`src/utils/coda-enablement.ts:30`). Blast radius is guide authors and dev-mode users, not general learners.

## History

Rewritten on 2026-09-05. The original body and the earlier comments described the pre-#1468 hand-rolled protocol and no longer match the code, so they are collapsed above as outdated. Closed attempt: #1579, which targeted the deleted parser. Related: epic #603; #1526 was the preventive half and is closed.

dvd233:
## What

Surface Coda terminal protocol mismatches in the terminal UI and connection state.

## Why

`CodaSession` reports a malformed frame through `onProtocolError`. The hook previously logged it at warn level only, so users saw no explanation and the warning could be filtered before a Pathfinder surface opened. The change logs at error level, renders a visible terminal diagnostic, and sets error state while leaving the session object alive for the SDK's non-fatal contract.

## Verification

- Added regression coverage for error-level logging, visible terminal output, error state, and the non-closing session behavior.
- `npx jest src/integrations/coda --coverage=false --runInBand` (141 passed)
- `npm run typecheck`
- Target ESLint and Prettier checks, `git diff --check`, and `npm run test:review-contract` (85 passed)
- The governance import-graph path assertions have existing Windows separator failures, reproduced on a clean baseline.

## Follow-up

The follow-up restores the SDK's documented non-fatal onProtocolError contract: it preserves the connected status and session id and allows later compatible frames, while keeping the diagnostic visible. The onError and onClosed paths remain the terminal teardown boundary.

Fixes #1527

### DIFF
--- a/src/integrations/coda/useTerminalLive.hook.test.ts
+++ b/src/integrations/coda/useTerminalLive.hook.test.ts
@@ -29,6 +29,10 @@ jest.mock('../../lib/logging', () => ({
 }));
 
 const mockedCreateSession = createSession as jest.MockedFunction<typeof createSession>;
+const mockedLogger = jest.requireMock('../../lib/logging').logger as {
+  error: jest.Mock;
+  warn: jest.Mock;
+};
 
 const SESSION_ID = 's_0123456789abcdef0123456789abcdef';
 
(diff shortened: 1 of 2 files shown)


moxious (approved):



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1876
Summary: surface terminal protocol mismatches
Verdict: Approve

PR #1881, opened 2026-09-13
test(guide-stats): derive conditional completion coverage

### LINKED ISSUE #1854: Completable-type coverage ratchet misses conditionally-completable types, and the rule doc over-claims that it doesn't
Raised from the review of #1842, which is approved and merging — follow-up work, nothing is broken today.

## The coverage ratchet does not cover conditionally-completable types

`progress.parity.test.ts:138` builds its expected set as `[...COMPLETION_AFFORDANCE_BLOCK_TYPES, 'input']` — the always-completable half derived from the exported constant, the conditional half a single hand-written literal.

`completion-affordance.ts` deliberately keeps `input` out of both lists ("it is the one conditional case") and hard-codes its blocking-datasource predicate in `emitsCompletionEvidence`. That is the only conditional branch today, so the sweep is genuinely whole at this head.

The problem is the next one. A second type made conditionally completable the same way would appear in neither the constant nor the test's hand-written tail. `uncovered` would stay `[]`, no fixture would ever be demanded, and a missing resolver case would produce the same silent 0% the original fix was about — one step further out.

## The documentation actively invites that shape

`.cursor/rules/tracked-step-types.mdc` step 7 says: "If only some authored shapes do, branch in `emitsCompletionEvidence` the way `input` does." Step 8 then claims the parity test "iterates the completable set, so it fails until both the resolver case and a fixture exist", and the `completion-denominator-authority` anchor says it "pins it to the parser over the whole completable set rather than a sample".

Both claims hold for the eight unconditional types and neither is enforced on the step-7 path. The over-claim is the sharper half of this: an engineer following step 7 will believe the ratchet is watching them when it is not.

## Two ways to close it

**Cheapest honest fix:** narrow step 8 to say the sweep demands a fixture only for a type listed in `COMPLETION_AFFORDANCE_BLOCK_TYPES`, note in step 7 that a conditionally-completable type must also be added to the parity test's tail by hand, and align the anchor sentence.

**Stronger:** remove the hand-maintenance instead of documenting it — either export a conditional set that both `emitsCompletionEvidence` and the test read, or probe `emitsCompletionEvidence` across the exhaustive `JsonBlock['type']` union so any type that can return true for some shape is demanded.

dvd233:
## Summary

Make conditionally completable block types an explicit shared registry with a required evidence predicate, and make both completion parity sweeps consume it. A future conditional step type now has to declare its predicate and representative fixture instead of disappearing behind a hand-written test literal.

## What to look at

- [Completion affordance registry](https://github.com/dvd233/grafana-pathfinder-app/blob/3391020233abfdb9ee1044f70b5209f12ffd8c91/src/lib/guide-stats/completion-affordance.ts) separates unconditional types from the conditional registry and routes shape checks through a mapped predicate set.
- [Progress parity sweep](https://github.com/dvd233/grafana-pathfinder-app/blob/3391020233abfdb9ee1044f70b5209f12ffd8c91/src/lib/guide-stats/progress.parity.test.ts) derives its completable population from both registries, so missing resolver coverage or a fixture becomes a failing test.
- [Registry parity tests](https://github.com/dvd233/grafana-pathfinder-app/blob/3391020233abfdb9ee1044f70b5209f12ffd8c91/src/lib/guide-stats/completion-affordance.parity.test.ts) and the [tracked-step rule](https://github.com/dvd233/grafana-pathfinder-app/blob/3391020233abfdb9ee1044f70b5209f12ffd8c91/.cursor/rules/tracked-step-types.mdc) document and assert the shared contract.

## Why

The previous progress parity test appended input as a hand-written literal because input is completable only when it carries a blocking datasource check. If another block type gained a conditional completion shape, an implementation branch could emit evidence while neither parity test nor the resolver fixture sweep knew that type existed; the guide would then read 0% while looking healthy. The shared tuple plus required predicate map makes the conditional population explicit at the production boundary, while the existing tier separation remains intact.

## How to verify

1. Exercise a blocking datasource input and a passive input; confirm only the blocking shape emits completion evidence.
2. Add a temporary conditional type to the shared registry without a corresponding parity fixture and run the progress parity suite; confirm the coverage assertion fails until the fixture and resolver case are present.
3. Inspect the completion-affordance parity output and confirm conditional types are excluded from the unconditional comparison but included in the schema classification.

## Breaking changes

None. Runtime completion behavior for existing block types is unchanged; this adds an internal registry constraint and test coverage.

## Out of scope

- Adding new tracked block types or changing their completion semantics remains a separate change.
- The runtime step-id resolver cases are unchanged; this PR makes missing cases fail visibly when a new conditional type is introduced.

## Validation notes

- guide-stats Jest suites: 6 passed, 128 tests passed.
- Typecheck, target ESLint, Prettier, review-contract tests (85/85), and git diff check passed.
- The repository full-check wrapper could not run on this Windows host because it invokes npm via spawnSync and npm is not an executable child name; no source workaround was applied.

Fixes #1854


moxious (changes requested):
CI is failing on this PR: the new `Object.hasOwn(...)` call in `src/lib/guide-stats/completion-affordance.ts` is unsupported under `tsconfig.cli.json`'s `ES2020` lib target, breaking `Build CLI, build image, smoke test`, `Validate packages`, `Build and smoke-test E2E runner`, and `CI Gate`. This needs either an ES2022+ lib target for that build, or avoiding `Object.hasOwn` in this file.

Code review to date: I ran a full pass across security, correctness-and-reliability, testing-and-verification, reversibility-and-one-way-door, cross-cutting-architecture, completion-records (including a contract-evolution check against the `completion-denominator-authority` anchor in `docs/design/CONCERN_DETAILS.md`, which this PR extends), and ai-subsystem. 0 other findings — the change is otherwise clean: the registry refactor is behavior-preserving, the `input` predicate is unchanged, the parity tests were updated correctly, and the docs update is a coherent extension of the existing contract anchor.

Please fix the CI failure above; happy to take another look once it's green.

dvd233:
Addressed the CI blocker in `87566aad9178bbb60f06a1a509689d76cf144e6e` and brought the branch current with signed merge commit `35b22e0eaae47fa22a35736035d224816a177063`.

- Replaced `Object.hasOwn(...)` with the ES2020-compatible `Object.prototype.hasOwnProperty.call(...)` without changing the prototype-safe predicate lookup.
- Added inherited-key coverage for `constructor`, `toString`, and `__proto__`.
- Resolved the two upstream conflicts by retaining the new unified completion-percentage model and its `guideProgress`/rollup tests while wiring the shared conditional-affordance registry into the same parity sweep and contract anchor.

Verification:
- Reproduced the original `tsconfig.cli.json` `TS2550` failure before the fix; the same CLI typecheck now passes.
- `npm run build:cli` passes.
- `npm run validate:packages`: 10 / 10 packages valid.
- `src/lib/guide-stats`: 6 suites / 157 tests passed.
- Full TypeScript, targeted ESLint/Prettier, `git diff --check`, and review-contract 85 / 85 passed.

The PR is mergeable against current `main`.

moxious (approved):
Thanks for the quick fix — the `Object.prototype.hasOwnProperty.call` swap keeps this ES2020-compatible and CI is green now (only the known Bundle Stats fork-permission check fails, which is expected infrastructure noise for community PRs, not something in your control).

PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1881
Summary: derive conditional completion coverage
Verdict: Approve

Ready for a maintainer to merge once you're happy with CI.

moxious:
@dvd233 apologies, pre-merge this picked up conflicts from another branch, if you can resolve those with a simple commit we'd merge

PR #1883, opened 2026-09-13
test(interactives): guard renderer attribute executor reachability

### LINKED ISSUE #1779: Tripwire: renderer writes a data attribute the executor never reads
## The seam

The renderer writes an author's field to a DOM attribute, and the executor never reads it. The field is spelled correctly, typed, registered in `KNOWN_FIELDS`, and forwarded. It still does nothing.

`openGuide` is the live instance. `src/components/interactive-tutorial/interactive-step.tsx:990` writes `data-openguide`. It is read at `src/lib/dom/dom-utils.ts:81`, reached from the sequence selector at `src/interactive-engine/interactive.hook.ts:343`. That selector requires the bare class token `interactive`, which top-level React steps never carry. (`extractInteractiveDataFromElement` has a second call site at `interactive.hook.ts:378`, but that is the requirements check and never consumes `openGuide`.) See #1506.

## The tripwire

Assert that every field the renderer writes to a `data-*` attribute reaches the executor for the block shapes that can carry it. Carry a shrink-only baseline of the gaps found on the first run, exactly as `content-renderer.prop-forwarding.test.ts` does for the seam above it.

## Why this shape

`src/components/content-renderer/content-renderer.prop-forwarding.test.ts` proves the pattern works. PR #1493 added it after `targetstate` "shipped broken exactly this way and was only caught by loading a guide in a browser". This is the same test one layer down.

## Origin

Theme T1, seam S2, from a review of the whole `agent/needs-decision` queue on 2026-09-05. Six seams have no tripwire. This is one of them.

dvd233:
## Summary

Add a contract tripwire for authoring data attributes written by interactive step renderers but not fully understood by the action executor. The test records the two verified gaps, data-openguide and data-targetcomment, as a shrink-only baseline so future renderer or executor changes cannot silently add another disconnected field.

## What to look at

- [Attribute executor contract test](https://github.com/dvd233/grafana-pathfinder-app/blob/c20c01e4441f7017e0c77c38143b77fb6c9f473d/src/components/interactive-tutorial/data-attribute-executor.contract.test.ts) derives renderer attributes from both interactive step components and compares them with the DOM extractor and InteractiveActionRequest surfaces.
- The test excludes data-testid, data-step-id, and multistep data-internal-actions because those are identity, test, or composite driver metadata rather than individual action fields.
- The non-empty baseline is intentional: data-targetcomment is not promoted by the DOM extractor, while data-openguide is not accepted by the direct React action request. Shrinking either entry belongs to a focused production fix.

## Why

An authoring field can be typed, rendered, and forwarded without ever reaching the executor that performs the action. That failure is silent: a sequence path selects only elements with the bare interactive class, and the direct React path builds its request separately. The existing prop-forwarding tripwire catches parser-to-renderer loss one layer up; this test covers the renderer-to-executor seam and keeps the known gaps visible.

## How to verify

1. Run the new contract test and confirm the computed unreachable set is exactly the documented baseline.
2. Add a temporary authoring data attribute to a renderer without adding it to both executor surfaces; confirm the baseline assertion fails.
3. Temporarily remove data-targetcomment from the DOM extractor or openGuide from InteractiveActionRequest; confirm the corresponding baseline entry remains live and the test catches the regression.
4. Confirm test, identity, and composite metadata are excluded from the action-field population.

## Follow-up

The follow-up commit bounds the source scan at the InteractiveActionRequest declaration and adds a range assertion so later type declarations cannot influence the baseline. The documented gaps and the renderer/extractor comparison remain unchanged.

## Breaking changes

None. This PR adds tests only and does not change renderer, extractor, or executor behavior.

## Validation notes

- New tripwire plus adjacent prop-forwarding/data-attribute suites: 64 passed, 12 skipped.
- Interactive-tutorial and DOM suites: 42 suites, 846 passed (12 skipped, 5 todo).
- Typecheck, target ESLint, Prettier, and git diff check passed.

Fixes #1779


moxious (approved):
Thanks for adding this tripwire -- it's a nicely structural way to guard against the renderer/executor "unhonoured attribute" gap, and the baseline-honesty tests are a thoughtful touch.

Suggestions & nits:

1. [suggestion] **S1 -- RENDERER_FILES is a hand-maintained population, not a structurally derived one** (low - testing-and-verification)
   The new tripwire hardcodes RENDERER_FILES = ['interactive-step.tsx', 'interactive-multi-step.tsx']. These are currently the only two files writing an action-shaped data-* attribute, but a future renderer that writes one would not be auto-added, so the tripwire would silently stop covering it -- the same class of gap it exists to catch.
   Suggested: Optional follow-up: derive the renderer population structurally instead of a hand-maintained array, consistent with the repo's decided tripwire-population pattern.



This is good to merge as-is -- the suggestion above is optional and doesn't need to hold this up.

Issue #1365, opened 2026-07-16
Support `targeting.requires` in the manifest schema so eligibility gates survive build-repository


moxious:
## Problem

grafana/grafana-recommender#241 adds an optional `requires` field to package `targeting` metadata — a hard eligibility gate evaluated before `match` scoring, so Admin-only content (e.g. the Linux server learning path) is never recommended to Viewers/Editors. The recommender-side implementation is in grafana/grafana-recommender#249, which reads `targeting.requires` from `repository.json`.

The Pathfinder package pipeline currently **silently strips** this field before it can reach the recommender:

1. `GuideTargetingSchema` (`src/types/package.schema.ts`) declares only `match`:
   ```ts
   export const GuideTargetingSchema = z.object({
     match: z.record(z.string(), z.unknown()).optional(),
   })
   ```
2. Zod's default object behavior strips unknown keys rather than rejecting them, so a guide author adding `targeting.requires` to `manifest.json` gets **no validation error and no effect**.
3. `build-repository` (`src/cli/commands/build-repository.ts`) copies targeting from the **parsed** manifest (`entry.targeting = manifest.targeting`), so `requires` is already gone by the time `repository.json` is written.

Net effect: the gate fails open at the pipeline level — exactly the failure mode the recommender-side change goes out of its way to prevent (it fail-closes on malformed `requires` in `repository.json`).

## Proposed change

Add `requires` to `GuideTargetingSchema`, following the same loose-typing pattern as `match` (the recommender owns the semantics):

```ts
export const GuideTargetingSchema = z.object({
  match: z.record(z.string(), z.unknown()).optional(),
  requires: z.record(z.string(), z.unknown()).optional(),
})
```

Because this schema is shared, the one change propagates everywhere it needs to:

- `ManifestJsonObjectSchema` — manifest validation accepts and preserves the field
- `RepositoryEntrySchema` / `RepositoryJsonSchema` — generated repository.json validates
- `schema manifest` CLI output — the canonical JSON Schema documents it
- `build-repository` — no change needed; it copies the whole parsed targeting object

Docs to touch: `docs/developer/package-authoring.md` (targeting field reference) and, in grafana/interactive-tutorials, `docs/manifest-reference.md`.

## Example manifest usage (once supported end-to-end)

```json
{
  "targeting": {
    "requires": { "userRole": "Admin" },
    "match": { "urlPrefix": "/connections" }
  }
}
```

[...]

PR #1884, opened 2026-09-13
test(packages): guard nested targeting field round-trips

### LINKED ISSUE #1780: Tripwire: nested manifest keys are stripped before repository.json
## The seam

A nested key inside `manifest.json` is stripped before `repository.json` is written, with no error. The author sees no warning and the gate silently never applies.

`GuideTargetingSchema` at `src/types/package.schema.ts:103-105` is a plain object, so it strips unknown keys. The manifest schema is loose at the top level only, so a top-level unknown key survives and a nested one vanishes. `src/cli/commands/build-repository.ts:207` then copies an object the field already left. See #1365.

## The tripwire

Round-trip a manifest carrying every field declared by `GuideTargetingSchema` through the build and assert that each one appears in the emitted `repository.json`. Carry a shrink-only baseline for anything already dropped.

Drive the field list off `GuideTargetingSchema.shape` rather than a hand-maintained list of documented fields. `match` is the only declared targeting field today and it survives the build, so a literal list passes vacuously. A shape-derived list is live-but-empty now and starts covering `requires` the moment #1365 lands, with no second edit.

## Why this shape

Extension forwarding cannot rescue this, because it iterates the already-parsed manifest. `validate` runs no check over targeting contents at all. It goes through `validatePackage`, whose only targeting signal is a WARN when the whole field is absent (`src/validation/validate-package.ts:310-321`). `MANIFEST_FIELD_VALIDATORS` (`src/cli/utils/cli-validators.ts:270-280`) also has no targeting entry, but that map is read only by `assertCliManifestFields` on the `set-manifest` path, so adding an entry there would harden `set-manifest`, not `validate`.

## Origin

Theme T1, seam S3, from a review of the whole `agent/needs-decision` queue on 2026-09-05.

dvd233:
## Summary

Add a schema-shape-derived tripwire that round-trips every declared targeting field from manifest.json into repository.json. The empty shrink-only baseline makes a future nested field such as targeting.requires fail loudly if Zod strips it before build-repository can forward it.

## What to look at

- [Build-repository round-trip test](https://github.com/dvd233/grafana-pathfinder-app/blob/73a77ffcc6ef96ea6585461b8db91721bda72cc3/src/validation/build-repository.test.ts) derives its targeting fixture keys from GuideTargetingSchema.shape, checks the emitted repository entry, and records dropped fields explicitly.
- The fixture uses a per-field sentinel object rather than a hand-maintained list, so the existing match field and future record-shaped targeting fields are covered by the same assertion.

## Why

GuideTargetingSchema is strict inside the otherwise loose manifest envelope. A nested targeting field can therefore disappear silently during parsing, while build-repository faithfully copies the already-stripped object. This is the same failure shape that motivates #1365: an authored eligibility gate can vanish before the recommender sees it. The tripwire records the current healthy state without changing the schema; a later field addition automatically enlarges the checked population.

## How to verify

1. Create a package whose manifest targeting object contains every key currently exposed by GuideTargetingSchema.shape and run build-repository.
2. Confirm every declared key and its sentinel value appears under the corresponding repository entry.
3. Temporarily add a nested targeting key to the schema but omit its preservation path, then run the test and confirm the empty dropped-field baseline fails.
4. Confirm the test does not assert or invent any targeting field that the schema does not declare.

## Breaking changes

None. This is test-only coverage; schema parsing, repository generation, and runtime targeting behavior are unchanged.

## Validation notes

- build-repository.test.ts: 35 passed.
- package-schema.test.ts: 85 passed.
- Typecheck, target ESLint, Prettier, and git diff check passed.

Fixes #1780


moxious (approved):



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1884
Summary: guard nested targeting field round-trips
Verdict: Approve

PR #1885, opened 2026-09-13
docs(security): record OBO fail-open exposure window

### LINKED ISSUE #1563: Correct security-model overstatements in CODA.md / BACKEND_PROXY_PATTERN.md (#1408 follow-up)
Follow-up from the second-round review of PR #1408 (https://github.com/grafana/grafana-pathfinder-app/pull/1408). Doc-only; no code change.

## Severity

**Low-medium.** These are security-model statements agents and reviewers treat as ground truth. Both overstatements were identified in the first-round review of #1408 (in the "dropped by adversarial verification" notes — the _finding_ was refuted, but these doc residuals were explicitly called out as surviving); neither was claimed fixed and no issue covered them. Verified verbatim-present at head `16e5bc4b`.

## What to correct

1. **"Not independently reachable with a client-set `X-Grafana-Id`"** — `docs/developer/CODA.md:164-166` and `docs/design/BACKEND_PROXY_PATTERN.md:104-108` both phrase the inbound trust boundary as though Grafana _strips_ the header. In fact core's `ForwardIDMiddleware` only **overwrites** it (`SetHTTPHeader` replaces; it never calls `DeleteHTTPHeader`), so a client-supplied value survives on a request where the requester's ID token is empty. Reword to "Grafana overwrites `X-Grafana-Id` from the server-side authenticated requester on every CallResource; a client-set value survives only when the requester has no ID token, in which case token exchange fails closed."
2. **"No hop on the outbound path accepts [the ID token] as authentication"** — `docs/developer/CODA.md:172-173` still carries the unqualified form, which is false: auth-api's `/v1/sign-access-token` accepts the ID token as the exchange `SubjectToken` that determines whom the minted token impersonates (`pkg/plugin/auth/auth.go:81-86`). `BACKEND_PROXY_PATTERN.md` was already corrected to "nothing on the outbound path accepts one **on its own**" (~`:83`) — mirror that qualifier in CODA.md, e.g. "…which nothing on the outbound path accepts on its own as authentication (auth-api consumes it only as the exchange subject, authenticated by the CAP token)."
3. **Missing fail-open note** — add one sentence near CODA.md's JWKS future-hardening paragraph (`docs/developer/CODA.md:176-179`): if the inbound trust boundary were ever broken, the OBO design fails open for up to the minted token's lifetime rather than closed; this is the concrete argument for the JWKS local-verification item already listed there. (Grep for "fail-open"/"token lifetime" confirms no such note exists in either doc.)

## Acceptance

All three edits are one-to-two-sentence rewordings at the cited lines; keep the two docs consistent with each other (CODA.md and BACKEND_PROXY_PATTERN.md currently describe the same fact at different strengths). No behavior change; `npm run check` prose-only.

## Links

- Root PR: https://github.com/grafana/grafana-pathfinder-app/pull/1408 (first-round review, "Dropped by adversarial verification" section, documents both residuals)
- Related doc-drift umbrella: https://github.com/grafana/grafana-pathfinder-app/issues/1556

dvd233:
## Summary

Document the conditional exposure window if the inbound identity trust boundary is ever defeated: the on-behalf-of access token could remain valid for up to the configured 600-second lifetime. This completes the canonical security-model note without changing authentication code or claiming a present vulnerability.

## What to look at

- [Canonical identity trust boundary](https://github.com/dvd233/grafana-pathfinder-app/blob/a154af7f3f0aee77470cef8e00d92be245eedb4a/docs/design/BACKEND_PROXY_PATTERN.md) now states why the JWKS check is load-bearing even though the current implementation fails closed.
- [Token exchange configuration](https://github.com/dvd233/grafana-pathfinder-app/blob/a154af7f3f0aee77470cef8e00d92be245eedb4a/pkg/plugin/auth/auth.go) is the source of the 600-second requested token lifetime.

## Why

The identity section already records the server-forwarding caveat and the two properties the gate enforces: stack namespace binding and ES256 signature verification. It did not state the consequence of a future regression. An agent treating local JWKS verification as optional could underestimate the blast radius; the new conditional sentence makes the hardening rationale explicit while preserving the current fail-closed description.

## How to verify

1. Trace the canonical identity section and confirm the new sentence is conditional on the inbound boundary being broken, not a statement about current behavior.
2. Check pkg/plugin/auth/auth.go and confirm tokenTTLSeconds remains 600 seconds.
3. Confirm no authentication, token exchange, or proxy runtime code changes are included.

## Breaking changes

None. This is a documentation-only clarification; proxy routes, token formats, and authentication behavior are unchanged.

## Reversibility

Reverting this commit removes the explanatory note without changing deployed state or credentials.

## Out of scope

- Implementing local JWKS verification or changing the identity gate remains tracked separately in #1568.
- The already-corrected forwarding and outbound-credential wording is intentionally not duplicated or altered.

## Validation notes

- Markdown Prettier check and git diff check passed.
- The 600-second value was verified against the current auth.go source.

Fixes #1563


moxious (approved):
LGTM, thanks!

Issue #1693, opened 2026-08-24
Block editor's Create GitHub PR action emits no manifest.json, so contributed guides arrive as incomplete packages


tomglenn:
## What happens

The block editor's **Create GitHub PR** action prepares a contribution into `grafana/interactive-tutorials` containing only the guide content. It writes no `manifest.json`, so every guide contributed through the editor arrives as an incomplete package.

`src/components/block-editor/utils/github-pr.ts:18` sets:

```ts
const CONTENT_FILENAME = 'content.json';
```

and prepares exactly that one file.

## Why it matters

A package without a manifest has no stats stamp, so the tutorials repository's `stats:check` gate flags it — against the contributor who opened the PR, not against the editor that produced the incomplete package. The contributor then has to work out what is missing and add it by hand, with nothing in the editor telling them it was needed.

This is the same defect that was fixed for private guides in https://github.com/grafana/grafana-pathfinder-app/pull/1682 (stamps a derived manifest on editor-authored guides), but on the public contribution path, which that PR deliberately left alone because it changes a public output.

## Suggested fix

Emit `manifest.json` alongside `content.json`, reusing the same derivation the private save path now uses (`deriveManifest`). The derivation already works out everything the content determines, so this should need no extra input from the author.

Depends on #1682 landing, since it introduces the derivation being reused.

## Alternatives considered

- Leave the public path content-only and have contributors add the manifest by hand in the PR. Rejected: the CI gate stays a tripwire for every contributor.
- Decide the editor should not author public packages at all, and remove or relabel the action. Rejected: the action is wanted.

moxious:
## 🧭 Implementation readiness

**Stakes:** Someone who authors a guide and clicks the create pull request button gets a contribution containing one file. The receiving repository's check then fails their pull request with nothing having warned them.

**Sanity check:** Every citation still resolves and the defect is unchanged. The maintainer answered one open question on 2026-09-01 and widened the other from a binary choice into a multi-file handoff redesign.

**Severity:** 🟠 Medium — the public contribution flow reliably produces a package that fails validation, though a contributor can hand-write the missing file.

**Code basis:**
- `src/components/block-editor/utils/github-pr.ts:18` — The only filename the helper knows. A second emitted file starts here.
- `src/components/block-editor/utils/github-pr.ts:143-173` — One serialisation, one path, one clipboard payload, one URL. Every step is single-file by construction.
[...]

PR #1887, opened 2026-09-13
test(block-editor): guard package emission surfaces

### LINKED ISSUE #1783: Tripwire: the editor can emit an incomplete package
## The seam

The block editor's public contribution flow emits an incomplete package. `Create GitHub PR` writes only `content.json` into `grafana/interactive-tutorials`, with no `manifest.json`, so every contributed guide fails that repository's stats-check gate and the contributor has to hand-author the missing file. See #1693.

## The tripwire

Assert that every package the editor emits for a package repository carries `manifest.json` alongside `content.json`. A new export path that emits a partial package fails the test.

`manifest.json` is optional to the package format itself: `src/validation/validate-package.ts:69` accepts a `content.json`-only standalone guide, and the authoring docs say the same. It is `discoverPackages` (`src/cli/commands/build-repository.ts:97-131`) that makes it mandatory in practice, because it recognises a directory as a package only when `manifest.json` is present. A repository-bound emission without one is therefore never discovered, and the failure is silent rather than loud.

## Why this shape

The guard is separable from the open product call in #1693, which is whether the handoff opens two GitHub new-file forms or offers one two-file download. The tripwire holds either way.

## Origin

Theme T1, seam S6, from a review of the whole `agent/needs-decision` queue on 2026-09-05.

dvd233:
## Summary

Add a source-level contract ratchet for every guide-emission path from the block editor.

- Inventory Copy JSON, Download JSON, Create GitHub PR, and backend save/publish at their source and UI callback boundaries.
- Keep intentional standalone content-only exports and the known single-file GitHub gap (#1693) as separate, reasoned baselines that become stale when behavior changes.
- Require the backend save path to retain its `preservedSpec` / `deriveManifest` package boundary and fail closed when a new emission callback appears.

This changes no user behavior. The multi-file GitHub handoff design remains scoped to #1693.

## Why

`manifest.json` is optional for a standalone content artifact, but `discoverPackages` requires it before a repository entry exists. The current public contribution handoff emits only `content.json`, so the receiving repository can drop that guide silently. This tripwire keeps the gap visible without pre-empting the handoff design.

## Verification

- New contract plus backend save flow, manifest derivation, header, and surface-coverage suites: 161 tests passed.
- TypeScript `--noEmit`, targeted ESLint, Prettier, and `git diff --check` passed.
- Review-contract suite: 85 tests passed.

Fixes #1783


moxious (approved):
Thanks for adding this test coverage — pinning the block editor's content-only vs. complete-package export boundary down with explicit contract tests keeps that distinction from drifting silently as the code around it changes. Nothing further needed from you here; this looks ready to merge.

PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1887
Summary: guard package emission surfaces
Verdict: Approve

PR #1888, opened 2026-09-13
fix(coda): scope gcx completion to the requesting step

### LINKED ISSUE #1772: gcx: one credential install completes every co-mounted gcx terminal-connect step
Follow-up from the round-3 review of https://github.com/grafana/grafana-pathfinder-app/pull/1668 (finding `gcx-install-completes-every-comounted-gcx-step`). Not a merge blocker there — no guide uses `gcx` yet and `pathfinder.coda-terminal` defaults off — but it becomes reachable the moment an author writes a guide with two gcx steps.

## What happens

A `terminal-connect` step with `gcx: true` completes itself when a credential lands for its session:

```ts
// src/components/interactive-tutorial/terminal-connect-step.tsx:170
} = useGcxCredential(gcx ? markComplete : undefined, terminalCtx?.sessionId);
```

`useGcxCredential` scopes on the session alone:

```ts
// src/integrations/coda/useGcxCredential.hook.ts:58-76
stored.sessionId === sessionId ? stored : { sessionId, state: 'idle', ... }
...
notifiedRef.current = snapshot.credential;
onReadyRef.current?.(snapshot.credential);
```

`TerminalContext` exposes exactly one live `sessionId`, so **every mounted gcx step names the same session at the same time**. Installing the credential for the first gcx step fires `onReady` for all of them, and each one calls `markComplete`.

In a guide with two gcx `terminal-connect` steps — the natural shape when the second needs a different `vmTemplate` — the second ticks off before its own VM exists, and its connect control never renders. The learner has no route to provision it.

## Reproduction

A throwaway case built on the existing `terminal-connect-step.test.tsx` harness, rendering two steps instead of one:

```tsx
mockTerminalStatus = 'connected';
mockSessionId = 's_abc';
mockProvisionGcx.mockResolvedValue(CREDENTIAL);
const onCompleteA = jest.fn();
const onCompleteB = jest.fn();
render(
  <>
    <TerminalConnectStep stepId="step-a" gcx onComplete={onCompleteA} vmTemplate="vm-aws" />
    <TerminalConnectStep stepId="step-b" gcx onComplete={onCompleteB} vmTemplate="vm-aws-sample-app" />
  </>
);
await act(async () => {
  await runGcxCredential('s_abc');
});
```

Result: `A completed: 1 B completed: 1`. Only A asked for the install.

## Why the existing test does not catch it

`terminal-connect-step.test.tsx:382` — *"leaves a gcx step targeting another VM alone"* — sets the module-level `mockSessionId` to `s_other` and renders a single step. That simulates a state two co-mounted steps cannot occupy: they read the same `terminalCtx.sessionId`. The test passes while the runtime case it stands for fails.

## Suggested fix

Scope `onReady` to the step that requested the run, not to the live session alone. The credential store already tracks which run is current (`generation`) — the missing piece is *who asked*. Options, roughly in order of preference:

1. Have `runGcxCredential` carry a requester id, and have `useGcxCredential` fire `onReady` only when the snapshot's requester matches the caller. Keeps one store, one credential per session, and the intended cross-surface sharing (a toolbar install still satisfies the step that is waiting).
2. Failing that, gate the step's completion on its own `handleConnect`/`handleGcxOnly` having initiated the run.

Note that option 2 gives up a behaviour the PR deliberately shipped: a toolbar install completing a waiting gcx step. Prefer option 1 unless that turns out not to be worth keeping.

Replace the `mockSessionId = 's_other'` case with the two-co-mounted-steps case above, since the current one asserts an unreachable state.

dvd233:
## Summary

- Carry an optional `requesterId` through the shared gcx credential snapshot and bind each hook run to its caller.
- Keep credential readiness shared by sandbox session, but fire `onReady` only for the step whose stable ID started that run.
- Pass `renderedStepId` from `TerminalConnectStep`; toolbar runs carry no requester and complete no step.
- Update the Coda rule and developer contract to distinguish shared readiness from requester-scoped completion.

## Why

All co-mounted gcx terminal-connect steps read the same live `TerminalContext.sessionId`. The previous session-only subscription therefore completed every step after the first credential install, including a later step whose VM had never been requested. That step then lost its own route forward.

## Behavior after this change

- A step still completes after its own minted or pasted credential install.
- A sibling on the same session sees the shared ready credential and its Continue control, but is not completed automatically.
- A toolbar install updates shared readiness but completes no guide step.
- Existing generation and session guards still discard late settlements after a replacement.

## Verification

- Pre-fix regression run: 4 focused failures reproduced the shared-session over-completion.
- Updated hook and component suites: 46 tests passed.
- Coda, terminal, challenge, and terminal-requirement coverage: 14 suites / 217 tests passed.
- TypeScript `--noEmit`, targeted ESLint, Prettier, `git diff --check`, and 85 review-contract tests passed.

Fixes #1772


moxious (approved):
Thanks for tracking down and fixing the co-mounted gcx completion bug (#1772) — scoping completion to the requesting step closes a real gap before any guide author ships two gcx terminal-connect steps.

Suggestions & nits:

1. [suggestion] **gcx-toolbar-no-longer-completes-waiting-step — Toolbar credential install no longer auto-completes a waiting gcx step** (low · coda-terminal)
   The fix scopes `onReady` to an exact requester match, so a toolbar-triggered install (requesterId null) can never satisfy any step's own non-null requesterId. That removes the toolbar-satisfies-a-waiting-step behavior PR #1668 shipped and issue #1772 asked to keep alongside the co-mounted-steps fix. The PR body and doc updates disclose the trade-off, but it is a larger behavior change than the issue's minimal ask.
   Suggested: Confirm with a maintainer that dropping toolbar-driven auto-completion of a waiting gcx step is an accepted product trade-off; if accepted, no further action is needed beyond what the PR already documents.

2. [nit] **gcx-settle-requesterid-check-redundant — settle()'s snapshot.requesterId check appears to never change the outcome** (low · correctness-and-reliability)
   In `gcx-credential-store.ts`, settle()'s discard condition adds `snapshot.requesterId !== requesterId` alongside the existing `run !== generation` and `snapshot.sessionId !== sessionId` checks. Given generation is bumped synchronously on every run start/reset and a single in-flight run per session is enforced before any await, the requesterId clause does not appear reachable as the deciding condition — the other two already cover every path that could change it mid-flight.
   Suggested: Optional: drop the redundant clause, or add a short comment naming the concrete scenario it alone catches if one exists that this review missed.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1888
Summary: scope gcx completion to the requesting step
Verdict: Approve with Minor
Results: 0 blockers, 2 non-blocking findings, 0 follow-ups

Next step: nothing blocks merge — a maintainer can merge once they've glanced at the two optional notes above (particularly confirming the toolbar-completion trade-off is intended).

PR #1890, opened 2026-09-14
feat(block-editor): support bulk block deletion

### LINKED ISSUE #1376: Block Editor: Support multi-select delete for blocks
## Summary

The Block Editor's selection mode (checkboxes) currently only supports merging selected blocks into a Multistep or Guided block. There is no way to delete multiple blocks at once. Please add multi-select delete support.

## Current behavior

- Entering selection mode is labeled **"Select blocks for merging"**.
- Selecting 2+ blocks shows **Merge to Multistep** and **Merge to Guided**.
- Deletion remains per-block only (individual delete control on each block).

This matches the intended design today (selection is for merge), but it is easy to assume the checkboxes are for bulk actions including delete.

## Requested behavior

When one or more blocks are selected in selection mode, offer a **Delete** action (in addition to the existing merge actions) that removes all selected blocks in one step.

Suggested UX details:

- Show Delete when `selectedCount >= 1` (merge can stay at `>= 2`).
- Confirm before deleting (especially for multiple blocks / nested content).
- Clear selection and exit selection mode after a successful delete.
- Keep merge behavior unchanged.

## Why

When editing or cleaning up a guide, authors often need to remove several blocks. Doing that one-by-one is slow and error-prone compared to selecting a range and deleting once.

## Notes

Implementation lives in `BlockEditorHeader` / `BlockEditorContent` selection toolbar and `useBlockSelection`.

Moved from grafana/interactive-tutorials#427 (opened in the wrong repo).

dvd233:
## Summary

- Make every root and rendered nested block selectable in selection mode.
- Add a confirmation-protected bulk delete that handles mixed root, section-child, and conditional-branch selections as one undoable mutation.
- Clear selection mode after deletion and keep merge actions disabled for ineligible selections.
- Document the selection and undo contract.

Fixes #1376

## How to verify

- Block-editor Jest domain: 62 suites, 860 tests passed.
- TypeScript typecheck passed.
- Block-editor ESLint and changed-file Prettier checks passed.
- `npm run test:review-contract` passed (85 tests).

The Windows full-suite run also reports four unrelated path/ratchet baseline failures in `import-graph`, `unused-bindings-lint-config`, `unscoped-checkers`, and `file-loader`; none touch this diff.


moxious (changes requested):
Blockers:

1. [blocking] **bulk-delete-hook-noop-untested — New no-op/skip-history branch in deleteSelectedBlocks is untested at the hook level** (high · testing-and-verification)
   useBlockEditor.ts's deleteSelectedBlocks setState updater returns prev unchanged (skipping notifyChange, and skipping the undo-history push via useGuideHistory's next===prev no-op) when none of the given IDs resolve. No test calls the hook's deleteSelectedBlocks with only stale/unknown IDs and asserts onChange was not called and history/undo state is unchanged; the only hook test covers the happy path, and the bulk-delete.test.ts stale-ID test only exercises the standalone pure function, not the hook wiring. Verified independently by two skeptics.
   Required: Add a hook-level test that calls deleteSelectedBlocks with only unknown/stale IDs and asserts onChange was not called, state.blocks is unchanged, and undoLabel/canUndo did not change.

2. [blocking] **canmergeselection-conditional-branch-untested — canMergeSelection's exclusion of conditional-branch selections has no test coverage** (high · testing-and-verification)
   canMergeSelection in useBlockEditor.bulk-delete.ts deliberately excludes selections resolving to kind === 'conditional' (only root/section are eligible), but no test constructs a conditional-branch child selection ID and asserts the function returns false for it. Verified independently by two skeptics.
   Required: Add a canMergeSelection test case with a conditional-branch child ID (e.g. conditional-1-true-0) in the selection and assert the result is false.

Suggestions & nits:

1. [suggestion] **bulk-delete-no-full-mount-integration-test — No full-mount BlockEditor integration test for the selection-confirm-delete-persist/undo path** (medium · block-editor-authoring)
   Coverage is split into three isolated layers (pure helper, hook with a stubbed onChange, and BlockEditorContent with all handlers mocked); no test mounts the real <BlockEditor /> tree the way BlockEditor.persistence.test.tsx does for other guide-mutating actions, so nothing proves the layers compose correctly (e.g. clearSelection after a real delete, or localStorage/undo after a real click-through).
   Suggested: Add one integration test that mounts <BlockEditor />, enters selection mode, checks two block checkboxes, clicks the real bulk-delete button plus confirm, and asserts the resulting localStorage guide state and that Undo restores the blocks.

2. [suggestion] **bulk-delete-duplicates-parseblockid-resolution — resolveSelectedBlock reimplements existing parseBlockId root/nested-section resolution instead of reusing it** (medium · correctness-and-reliability)
   useBlockEditor.bulk-delete.ts re-derives root-ID lookup and the ${sectionId}-nested-${index} parsing scheme from scratch (adding conditional-branch support on top) instead of building on the existing parseBlockId resolver in useBlockEditor.helpers.ts already used by useBlockEditor.merge.ts; MERGEABLE_TYPES/canMergeSelection likewise duplicates the existing isMergeable type-gate concept. Two independently-maintained resolvers for the same ID scheme can silently diverge on a future change.
   Suggested: Build resolveSelectedBlock on top of parseBlockId/isMergeable rather than re-deriving root/section resolution independently; extract one shared resolver both merge and bulk-delete call.

3. [suggestion] **merge-button-hint-uses-native-title-not-tooltip — Disabled-merge-button explanation uses a raw HTML title attribute instead of Grafana Button's tooltip prop** (medium · cross-cutting-architecture)
   Both merge buttons in BlockEditorContent.tsx pass a native title attribute for the disabled-state explanation instead of Grafana Button's own tooltip prop, which sibling usages in the same subsystem (HeaderKebab.tsx, TypeSwitchDropdown.tsx) already use for exactly this case.
   Suggested: Replace title={...} with tooltip={...} on both merge buttons, matching the convention used elsewhere in block-editor/.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1890
Summary: support bulk block deletion
Verdict: Request Changes
Results: 2 blockers, 3 non-blocking findings, 0 follow-ups

dvd233:
Addressed the review in `8e105039ef74689659f7076441717fa63d0c0724`:

- Added the hook-level stale-selection regression: blocks retain the same reference, `isDirty` stays false, `onChange` is not called, and no undo entry or label appears.
- Added a conditional-branch case whose child is itself mergeable, proving the branch location—not the child type—keeps the selection ineligible.
- Added a full `<BlockEditor />` integration test covering select → confirm delete → persisted one-block guide → selection exit → Undo → persisted restoration.
- Reused `parseBlockId` for root/section resolution and one shared `isMergeableBlock` predicate for both the merge helper and selection eligibility.
- Switched the disabled merge hints to Grafana `tooltip`; the test now verifies `aria-disabled` and that clicks cannot invoke either merge handler.
- Kept the dnd-kit sensor array stable across selection-mode toggles while each root/section/conditional sortable remains disabled, with a regression for the React dependency-length warning.

Verification: 62 block-editor suites / 864 tests passed, plus TypeScript, targeted ESLint, Prettier, and `git diff --check`.

moxious (approved):
Thanks for turning this around so quickly, @dvd233 — bulk block deletion is a solid addition to the editor, and this round closes out every blocker and suggestion from the first pass cleanly.

PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1890
Summary: support bulk block deletion
Verdict: Approve

Nothing further needed from you — this is ready for a maintainer to merge.

moxious:
Smoke tested & verified, demo video attached, thanks for the contribution!

https://github.com/user-attachments/assets/23911cc0-f20d-411e-abb5-25a396e6ddd5

PR #1891, opened 2026-09-14
fix(dom): discover lazy-rendered targets across both scroll axes

### LINKED ISSUE #1865: lazyRender cannot discover targets in horizontally virtualised containers: scrollUntilElementFound only scrolls vertically
## Summary

`lazyRender` + `scrollContainer` can only discover targets that are hidden **below** the fold.
`scrollUntilElementFound` scrolls one axis:

```ts
// src/lib/dom/dom-utils.ts:333
scrollContainer.scrollBy({ top: scrollIncrement, behavior: 'smooth' });
...
const atBottom = scrollContainer.scrollTop + scrollContainer.clientHeight >= scrollContainer.scrollHeight - 10;
```

A container that virtualises **columns** is never scrolled sideways, so a target in a trailing
column never enters the DOM and the step fails with `Element not found - scroll dashboard to
discover` no matter how many attempts it makes.

## Why it matters

Grafana's table panel renders through `react-data-grid`, which virtualises horizontally as well as
vertically. Any control a plugin puts in a table cell is therefore invisible to a guide step
whenever the column sits outside the current viewport — and trailing columns are exactly where
plugins put their action controls, because that is where an appended column lands.

Measured on the Kubernetes Monitoring app (`grafana-k8s-app` 2.38.0) against Grafana 12.3.0 at a
1867 px viewport, all counts `document.querySelectorAll(...).length`:

| target | on load | after `scrollLeft = scrollWidth` on the grid |
|---|---|---|
| the pod-health bar in the Workloads table's `Pods` column | 0 | 17 |
| the forecast buttons in the Namespaces table's `Actions` column | 0 | 24 |

The grid reports `scrollWidth 1950` against `clientWidth 985`, and the column is present in the
column picker and checked — it is simply not rendered. Setting `scrollLeft` makes the targets
appear immediately, so the elements and their test ids are fine; only the discovery axis is
missing.

The practical effect is that a guide cannot teach any in-table control on a wide table. In the tour
we were building this removed four steps and two whole sections, and there is no authoring
workaround: `scrollContainer` selects *which* container to scroll, not which way.

## Suggested fix

Scroll both axes in `scrollUntilElementFound`, and end on whichever axis can still move:

```ts
const canScrollY = scrollContainer.scrollHeight > scrollContainer.clientHeight;
const canScrollX = scrollContainer.scrollWidth  > scrollContainer.clientWidth;

scrollContainer.scrollBy({
  top:  canScrollY ? scrollIncrement : 0,
  left: canScrollX ? scrollIncrement : 0,
  behavior: 'smooth',
});

const atEndY = !canScrollY || scrollContainer.scrollTop  + scrollContainer.clientHeight >= scrollContainer.scrollHeight - 10;
const atEndX = !canScrollX || scrollContainer.scrollLeft + scrollContainer.clientWidth  >= scrollContainer.scrollWidth  - 10;
if (atEndY && atEndX) { break; }
```

Sweeping one axis at a time (exhaust Y, reset, then exhaust X) would be more thorough for a grid
that virtualises both, at the cost of more attempts. Either is a large improvement on today's
behaviour, where the X axis is never touched.

Two smaller things that would help authors regardless:

- The default container is `.scrollbar-view` (`DEFAULT_DASHBOARD_SCROLL_CONTAINER`,
  `dom-utils.ts:313`), which is the dashboard scroller, not the grid. For a table target the author
  must know to pass the `.rdg` element as `scrollContainer`. Worth a line in the authoring guidance
  that `authoring-start.ts:129` already gestures at.
- When `scrollUntilElementFound` gives up, the log says the element was not found after scrolling.
  Saying which axes it actually moved, and whether the container could scroll on each, would have
  made this diagnosable from the console instead of from the source.

## Environment

- `grafana-pathfinder-app` 2.17.0, Grafana 12.3.0 (Enterprise, dev), Chromium via Playwright
- `grafana-k8s-app` 2.38.0 built from source, viewport 1867x863

dvd233:
## Summary

- Extend lazy-render discovery to horizontally virtualized containers.
- Scroll only axes that still have content and stop after both axes reach their boundaries.
- Preserve the existing selector scope, smooth scrolling, early DOM hit, and max-attempt safeguards.
- Add regression coverage for horizontal discovery and independent axis completion.

Fixes #1865

## How to verify

- `src/lib/dom`: 10 suites, 306 tests passed.
- TypeScript typecheck passed.
- Targeted ESLint and Prettier checks passed.
- `npm run test:review-contract` passed (85 tests).

The Windows full-suite run reports four unrelated path/ratchet baseline failures in `import-graph`, `unused-bindings-lint-config`, `unscoped-checkers`, and `file-loader`; none touch this diff.


moxious (changes requested):
Thanks for extending lazy-render discovery to horizontal virtualization -- the dual-axis boundary check is a clean generalization of the existing vertical-only loop, and the new tests for horizontal discovery and independent-axis completion are welcome.

Blockers:

1. [blocking] **1891-lazy-scroll-grace-wait-removed -- Boundary short-circuit removes the one guaranteed wait+recheck cycle for non-scrollable containers** (medium * interactive-engine)
   When both scroll axes already read at their boundary before the first scroll (e.g. a container with no vertical or horizontal overflow), the new getScrollBoundaries() check breaks out of the loop immediately, so scrollBy, the waitTime await, and the recheck never run. The old code always ran that scroll+wait+recheck pass at least once before checking atBottom. The only caller (interactive-step.tsx) reaches this function after a zero-retry resolveWithRetry({ delays: [] }), so this was the sole grace window an asynchronously-rendering element had; it is now removed for any non-overflowing container, and no existing or new test exercises this exact boundary condition (the closest test sets scrollHeight===clientHeight but only asserts the final null return, not scrollBy call count or elapsed wait).
   Required: Keep at least one scroll+wait+recheck pass even when both axes already read at-boundary before the first scroll, or confirm the narrowing is intentional and add a regression test asserting the wait cycle still runs when the container starts fully scrolled on both axes.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1891
Summary: discover lazy-rendered targets across both scroll axes
Verdict: Request Changes
Results: 1 blocker, 0 non-blocking findings, 0 follow-ups

Happy to take another look once this is addressed -- thanks again for the contribution.

dvd233:
Addressed the blocker in `1db032ecb065d126d5af6c65abe45992b6aad389` and brought the branch current with the signed merge commit `22ea1a9141547100db77f8b5f55d59726cb9ea1b`.

- Removed the pre-scroll boundary short-circuit, so a container that starts at both boundaries still performs one scroll/wait/recheck grace cycle before the post-scroll boundary exit.
- Added a regression that asynchronously mounts the target during that grace wait; it verifies the zero-distance scroll call and successful recheck.
- Resolved the sole upstream conflict by preserving the latest docs-retrieval and full-screen handoff anchors alongside #1865's dual-axis lazy-scroll contract.

Verification: `src/lib/dom` 10 suites / 307 tests passed, plus TypeScript, targeted ESLint, Prettier, `git diff --check`, and review-contract 85 / 85. The PR is mergeable against current `main`.

moxious (approved):



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1891
Summary: discover lazy-rendered targets across both scroll axes
Verdict: Approve

Issue #1492, opened 2026-08-03
Differentiate terminal VM timeouts between authoring and viewer modes


Jayclifford345:
## Problem

VM timeouts felt "somewhat aggressive" while *building* a guide, and fine while replaying it in viewer mode. That's the expected outcome of a single budget: authoring means long think-time between terminal interactions, so the VM dies under the author mid-edit. Reading means steady forward progress, so it doesn't.

## What we know

- Coda destroys VMs **30 minutes** after they go `active` — see "VM lifecycle" in [`docs/developer/CODA.md`](docs/developer/CODA.md). Server-side; the plugin can't extend it today.
- The plugin polls VM state every 15s and cancels the stream once the VM enters a terminal state — [`stream.go:956`](pkg/plugin/stream.go#L956).
- Other budgets in play, for reference: frontend SSH-handshake timeout 35s ([`useTerminalLive.hook.ts:307`](src/integrations/coda/useTerminalLive.hook.ts#L307)); backend relay handshake 30s ([`terminal.go:83`](pkg/plugin/terminal.go#L83)); `waitForVMActive` polls 3s × 60 (~3 min); exec default 5s capped at 120s ([`coda_exec.go:35`](pkg/plugin/coda_exec.go#L35)).

## What we want

Distinct budgets by mode — a longer-lived VM while a guide is open in the editor, current behaviour for viewers.

## Shape of the work

- The **caller-side signal is cheap**: the plugin already knows whether it's rendering the block editor or a published guide, so mode can be passed at `CreateVM` time.
- Extending the actual TTL is a **Coda-server change** — a lifetime hint on `CreateVM`, or a heartbeat/extend endpoint. Cross-repo, and the quota question is the real one: authoring VMs living 2–3× longer eats pool capacity, and `vm-aws-sample-app` / `vm-aws-alloy-scenario` are on-demand rather than hot-pool, so the cost is per-VM.
- **Cheaper interim wins** that don't need a Coda change and are worth doing regardless:
  - Surface remaining lifetime — "session ends in N minutes" — instead of letting it expire silently. We already poll expiry every 15s, so the data is on hand.
  - Reconnect without losing place when the VM does die, rather than dropping the author back to a cold terminal.

## Open questions

- Extend a fixed longer TTL, or renew on activity? A heartbeat is friendlier but keeps abandoned editor tabs alive — it needs an absolute cap either way.
[...]

PR #1901, opened 2026-09-14
feat(coda): show remaining VM lifetime

dvd233:
## What

- Show the active sandbox VM's remaining lifetime beside the terminal title in both collapsed and expanded views.
- Join one `GET /v1/vms` lookup to the connected session's `vmId`, then advance the countdown locally without polling the API.
- Hide zero, invalid, missing, or failed expiry lookups without changing terminal connection state.

## Why

Coda VMs have a server-owned fixed lifetime, but Pathfinder currently lets them expire without warning. This implements the remaining-lifetime interim win from #1492 while keeping VM lifetime policy and session lifecycle outside Pathfinder.

## Contract evolution

This extends the existing `coda-terminal` contract. `src/integrations/coda/coda-api.ts` remains the single Coda API adapter, and `grafana-coda-app` remains the owner of VM and session lifecycle. The change does not extend or infer a TTL.

## Verification

- Coda integration suite: 11 suites / 158 tests passed.
- Focused coverage: 4 suites / 39 tests passed; `VmExpiryIndicator.tsx` and `vm-expiry.ts` reached 100% statements, functions, and lines.
- `npm run typecheck`
- Target ESLint, Prettier, and `git diff --check`
- Review-contract suite: 85 / 85 passed.
- A full Jest coverage attempt reached 9,182 passing tests; 10 tests in four unrelated suites hit existing Windows path-separator assertions. All Coda tests were rerun on the final head.
- The `npm run check` wrapper itself could not dispatch its first child `npm` process on this Windows host (`spawnSync npm ENOENT`), so the available frontend checks were run individually.

Related: #1492


Jayclifford345:
Nice wor, LGTM overall.
One nit: the 15s setInterval in VmExpiryIndicator re-renders more often than needed, since the label only shows minute-level granularity (formatVmExpiry). Most of those ticks don't change the displayed text. Could this align to the next minute boundary instead — e.g. a setTimeout computed from expiryMs - clockMs mod 60s, rescheduling each time — rather than a fixed 15s poll? Would cut re-renders without losing accuracy on the "1 min left" → "Session expired" transitions.
Not a blocker, just a polish suggestion.

moxious (approved):

PR #1903, opened 2026-09-15
test(e2e): cover completion tracking in the browser

tomglenn:
## What this adds

Browser coverage for completion tracking, plus a manual checklist for the part that cannot be automated.

Earned progress and the universal **Mark complete** control have dense unit coverage and nothing above it — no browser test asserted a percentage, the Mark complete button, a path rollup, or a queued completion fact. This adds six cases in `tests/completion-tracking.spec.ts`:

- walking a whole four-milestone path with next and back earns no credit — every milestone and the path stay at 0%, nothing is queued, and nothing is persisted for any milestone either;
- completing one interactive step moves the percentage by exactly its share of the guide's block count;
- a path reads the mean of its milestones, over one finished and one partly finished milestone;
- one completion queues one fact under the right identity, for each way a guide can be launched: bundled by id, bundled by package path, an `api:` share link, and as a member of a path;
- reloading after a completion does not queue a second fact;
- a successful write empties the queue, and the POST body carries the idempotency key the backend derives the record name from.

Every assertion is written against `docs/design/COMPLETION-MODEL.md` rather than against what the code happens to produce today.

## Every case was checked against a mutation

A case that cannot fail is worse than no case, so each claim was tested by breaking the source it guards and confirming the case goes red. Two rounds of that found three cases that proved nothing, all now fixed:

- asserting a milestone's percentage before the footer had read its stored mark — the footer reports a hard-coded 0 in that window, so the navigation case was asserting a constant. Both surfaces now declare when their number means something, and the case also asserts the storage underneath.
- the path-mean case asserted 25% with one of four milestones done, which a completed-milestone count also produces. It now takes a second milestone part of the way, so the mean moves while a count stays put, and it asserts the two answers differ before asserting the value.
- nothing ever exercised a successful write. The stub already took a status that no caller passed.

## Two expected failures, both left red on purpose

`test.fail()` on the bundled package-path case: `bundled:<id>` and `bundled:<id>/content.json` are both live launch shapes for one guide and record two different identities — the writer (`persistJourneyCompletionPercentage`) keeps the `/content.json` suffix in `guideId`, the reset path (`fallbackGuideIdFromContentKey`) strips it, so the two disagree about what to key.

`it.failing()` in `learning-journey-helpers.completion-boundary.test.ts`: `markMilestoneDone`'s whole-journey branch derives its identity itself with `fallbackSource: 'bundled'` where the shared derivation would use the schema default. It sits at the unit level because every path launch that reaches that branch through the DOM also resolves a `repository` that pre-empts the fallback; the browser suite keeps a tripwire for the day that changes.

Both fail on the identity comparison itself, not on anything incidental. Picking which spelling wins is a decision about the durable key rather than a test change.

## What to look at

- `tests/completion-tracking.spec.ts` — the cases, and in the comments why each one is shaped the way it is.
- `tests/helpers/completion.fixtures.ts` — guides and paths are fixtures served through `page.route`, so no case depends on the live package index or on network egress. Expected denominators are computed from each fixture's own blocks with `computeGuideBlockIndex`, not written down.
- `tests/helpers/completion.helpers.ts` — how a fact is observed. Facts are read out of the client's write queue in `localStorage` (key prefix imported from `src/lib/storage-keys.ts`). The write route is stubbed absent by default: a structural 404 is the documented condition under which the client stands its network drain down but keeps persisting facts, which is what makes the chain up to the network boundary observable without a cloud stack. One case passes 201 instead.
- `src/` carries four test handles and no behaviour change:
  - `data-test-progress-state` on the Mark complete footer (`src/components/mark-complete/MarkCompleteFooter.tsx`) — the most load-bearing of the four, and the fix for the vacuous-assertion bug above.
  - `data-test-path-percent` on the path table of contents (`src/components/LearningPaths/LearningPathTableOfContents.tsx`) — the path percentage was only rendered inside a ring the cover page hides at 0%, and 0% is the value these cases most need.
  - testids on the next and previous milestone buttons (`src/components/docs-panel/components/LearningJourneyMilestoneToolbar.tsx`) — previously addressable only by a translated `aria-label`.
  - All four are documented in `docs/developer/E2E_TESTING_CONTRACT.md`; the two testids also get a source-level tripwire in `docs-panel.contract.test.tsx`.
- `docs/developer/COMPLETION_RECORDS_CLOUD_CHECKLIST.md` — the cloud round trip stays manual, with no CI job: it needs a real forwarded identity and a namespace that serves the record kind. Worth reading for the delta rule — `/completion-records/my` reports a running count per guide and nothing deletes a durable record, so every assertion has to be a before-and-after delta or it passes once and fails forever after. The replay step confirms the resend actually left the browser before reading any count, because a replay that never sent looks exactly like one that was correctly deduped.

## Scope

Client side, up to the network boundary. Nothing under `pkg/` is touched and no product behaviour changes — the only non-test `src/` edits are the four handles above. Later passes cover reset and re-completion, the analytics event decision 7 exists to emit, and decision 9's excluded-member rule.

The browser suite is not in `ci-gate`'s `needs`, so it is advisory today: it can go red across all seven Grafana matrix legs and a PR still merges on a green CI Gate. That predates this PR and is worth deciding deliberately rather than inheriting.

## How to verify

There is no product behaviour to exercise here; it is tests and docs.

```bash
npm run server                 # or GRAFANA_URL=... against an existing stack
npx playwright test tests/completion-tracking.spec.ts
```

Expect 10 passed, with `the same bundled guide launched by package path` reported as an expected failure. To see what that case asserts, change its `test.fail(` to `test(` — it fails on the `guideId` comparison, with `welcome-to-grafana/content.json` where the model wants `welcome-to-grafana`.

`npm run test:ci`, `npm run typecheck`, `npm run lint` and `npm run prettier-test` are clean. The suite is green on all seven CI matrix legs, and locally over repeated runs including `--repeat-each=3`.


tomglenn:
Both Bugbot findings were right and are fixed in 89331c58. Replying here rather than in the two threads because the tooling I have cannot post threaded replies to review comments.

### Footer percent read before hydration (`tests/completion-tracking.spec.ts` L83-84, L93-94)

Correct, and worse than it looks: that made the most important case in the slice vacuous. Until `MarkCompleteFooter` has read the guide's stored mark it resolves no content key, so `mark-complete-percentage` reads a hard-coded `0% complete` for every guide — the walk-the-path case was asserting a constant. The cover page has the same window: before its stored progress arrives, `completedSlugs` is empty and every path reads 0%.

Both surfaces now say when their number means something:

- `data-test-progress-state` on the footer root — `pending` until the mark is read, `ready` after.
- `data-test-path-percent` on the path table of contents is now **absent** until progress has loaded, so there is no provisional value to mistake for a real one.

The helpers wait on those before reading, and the case polls. The case also asserts the layer underneath the two percentages now: no persisted percentage for any milestone content key, and no completed milestone slugs. A rendered 0 can mean "not read yet"; a stored record cannot.

Verified by mutation locally (not committed): crediting a milestone on arrival turns the case red on the milestone footer assertion, `Received: 100` where 0 is expected. Before the fix that same mutation left it green.

While fixing this I found the same class of hole in `markComplete`: it waited on the button's `disabled` DOM property, which Grafana's `Button` never sets — it uses `aria-disabled`, which this contract deliberately does not select on. So that wait was a no-op too, and a click landing before the mark was read would have been silently dropped by the handler. It waits on the footer's progress state now.

### Reload assertions use short timeout (L318-324)

Correct. Both post-reload assertions now use the same 30s budget `launchDoc` and `openDocsPanel` give the panel on first launch, named as a constant so the two cannot drift apart again.

Also in this push, from a separate review round: the divergence the `test.fail()` case documents has a sibling in `markMilestoneDone`'s whole-journey branch that no DOM-reachable launch can trigger, because every path launch that gets there also resolves a `repository` which pre-empts the hard-coded `'bundled'` fallback. That one now has an expected-failure unit test in `learning-journey-helpers.completion-boundary.test.ts`, with the browser case left as a tripwire.

docs/developer/COMPLETION_RECORDS_CLOUD_CHECKLIST.md
```diff
@@ -0,0 +1,259 @@
...
+```js
+const baseline = before[`${guideSource}\t${guideId}`] ?? 0;
+await waitForDelta(guideSource, guideId, baseline, 1);
+```
+
+6. Force the resend: write the saved item back under its original key and reload the page. The queue drains it again, under the same idempotency key as before.
+
+```js
+localStorage.setItem(saved.key, saved.value);
+location.reload();
```
Jayclifford345, line 186: **[P2] Persist replay-check state before reloading**

Reloading clears the console execution context, including `saved`, `baseline`, `guideSource`, `guideId`, and the previously declared helpers. Following the next snippet therefore fails at `waitForResend(saved.key)` with a ReferenceError, and the final count comparison loses its baseline too. Persist the verification state in `sessionStorage` before reloading, then restore it and re-declare the read helpers afterward so the replay check can be completed as written.

Jayclifford345 (dismissed):
Approving as no major findings but worth a look on the assumption of the AI review findings

tomglenn:
@Jayclifford345 thanks — the reload finding was right and it was the worst possible place for it, in the one step where a wrong result is worse than no result. Fixed in e87b9881. (Replying here rather than in the thread; the tooling I have cannot post threaded replies to review comments.)

**Your finding.** Step 6 now stashes `savedKey`, `guideSource`, `guideId` and `baseline` in one `sessionStorage` entry before reloading. Step 7 opens by saying plainly to re-paste the `PF` line and the helper block — that was the implicit bit worth making explicit — then restores the stash and throws a clear message if it is missing. The last step clears the entry so a second run of the checklist cannot pick up the first one's state.

**Three more of the same shape, found by reading the step cold afterwards.** All of them break for someone pasting snippets in order:

- Steps 2 and 3 both declared `const fact`, `const guideSource`, `const guideId`, `const baseline`. The second declaration throws in a console that has already seen the first, so step 3 was unreachable after step 2. Those working values are `var` now, with a note saying why, since the names are deliberately re-derived.
- Step 2's drained-queue fallback declared `const after`, colliding with the poll below it.
- Step 4 asked you to paste an inline span containing a template literal, and its backticks closed the span early — it rendered as garbled JS. It is a fenced block now, and derives the path's pair off the queued fact instead of having you type it.

`waitForDelta` also moved into the shared helper block. It was declared inside step 2's snippet and used by step 4, so the step 3 reload destroyed it — now there is exactly one block to re-paste.

**Verified by executing the document rather than re-reading it.** Every snippet, in order, in one page context, across the real reload, against a stub that dedupes on `idempotencyKey` the way the write route does. Step 3 came out at 3 POSTs, 2 unique keys, `delta: 1` — the replay confirmed sent, and not counted twice.

**On the automated-review findings you mentioned when approving:** both were already fixed earlier on this PR, in 89331c58, with the evidence in the comment above. The percentage-read-before-hydration one was the more serious of the two — it made the navigation case assert a constant — and the fix added `data-test-progress-state` on the footer plus an absent-until-loaded `data-test-path-percent`, then verified by mutation: crediting a milestone on arrival now turns that case red, where before the same mutation left it green. The short reload timeouts were the second, and all nineteen remaining sites now share `TIMEOUTS.UI_READY`.

Jayclifford345 (approved):

Issue #1665, opened 2026-08-20
guide-stats: preserve canonical snippet counts across direct and prepared launches


moxious:
### Summary

Make snippet-containing guides use the same canonical, pre-inlining completion index through direct opens and prepared launches. Preserve the original counting tree through the prepared-content handoff instead of counting the expanded render tree.

This is now a runtime correctness issue, not just a latent API hazard. Part of #603; originally raised during review of #1661.

### User impact / flow change

A guide using a snippet can currently receive a different completion denominator depending on how it was opened. The frozen index can preserve that difference for the life of the content key, so the first opening path matters.

After this change, opening the same guide directly or through a prepared launch uses the same counting rule and canonical positions, without waiting for snippet resolution on direct open or fetching the guide twice on prepared launch. Snippet rendering and launch-surface classification remain unchanged.

The audit reproduced the arithmetic and frozen-store disagreement with a fixture using the real inliner, counter, and index store; it did not reproduce the full browser flow or establish how many published guides are affected.

### Current failure

Audited at `cdb0befa4209557fe67c277aed282fb73c02c9e1` on 2026-09-15:

1. A guide contains one `snippet-ref` followed by one ordinary block. Its canonical total is **2**, because a snippet reference counts as one block.
2. The snippet resolves to two blocks. `inlineSnippetRefsInGuideWithStatus` replaces the reference with those blocks, so the expanded render tree contains **3** blocks.
3. Direct open gives the renderer the original tree; prepared launch gives it the expanded tree.
4. `ContentProcessor` runs `computeGuideBlockIndex(rawGuide.blocks)` on either input and labels both results `live-pre-inlining`.
5. The real frozen-index store retains totals of **2** and **3**, respectively. Publishing the expanded overlay later does not change the direct-open total.

Freezing prevents within-load drift; it does not make two different initial trees equivalent. The original issue body and September 5 readiness comment predated the renderer consumer and understated the exposure.

### Scope and implementation direction

**Decision: preserve the counting source, rather than reconstruct snippet provenance or refuse valid prepared launches.** This supersedes the earlier refusal-only implementation direction.

1. Keep two purposes explicit at the content boundary: the original pre-inlining guide/blocks for counting, and the expanded guide for rendering and Grafana-UI classification.
[...]

PR #1905, opened 2026-09-15
fix(guide-stats): count the pre-inlining tree on both launch paths

moxious:
## Intent

Implement GitHub issue grafana/grafana-pathfinder-app#1665 ('guide-stats: preserve canonical snippet counts across direct and prepared launches') in an isolated worktree off latest main, and raise a PR. The user asked for the issue's own stated implementation decision, not an alternative design.

The bug: a snippet-ref counts as exactly one block however many blocks it resolves to, so a guide's canonical completion denominator is the PRE-inlining count. A direct open hands ContentRenderer the pre-inlining tree; prepareGuideLaunch expands snippet refs before a display surface is committed and handed the renderer the already-expanded tree. ContentProcessor ran computeGuideBlockIndex over whichever tree arrived and labelled both results 'live-pre-inlining', so one reference expanding into two blocks beside one ordinary sibling froze a total of 2 on the direct path and 3 on the prepared path — and the frozen per-content-key index keeps whichever the reader's first opening path produced.

Decisions the issue mandated and this change follows deliberately:
- Preserve the counting source rather than reconstruct snippet provenance or refuse valid prepared launches. The issue explicitly supersedes an earlier 'refusal-only' direction, so rejecting prepared launches would be wrong.
- Keep two purposes explicit at the content boundary: RawContent.countingSource names the pre-inlining tree; content stays the render tree. PreparedRawContent requires the pre-inlining variant so a producer cannot drop it silently, and createPreparedContent serializes both trees in one call.
- Produce the WHOLE GuideBlockIndex from the preserved tree, never a corrected denominator combined with post-expansion positions.
- Do not infer provenance from a comment, a boolean phase assertion, or the absence of snippet-ref (a fully expanded tree has none).
- A known-expanded payload that lost its counting source keeps rendering but publishes no canonical index and logs a diagnostic through the existing logger — deliberately not a crash, and deliberately not a silent fallback to counting the expanded tree, because the freeze would keep a non-canonical count for the life of the content key. 'unavailable' has no production producer today; it exists so the lost-source state is spellable and testable rather than indistinguishable from a direct open.
- Keep the metadata internal to the content/launch handoff: not author-supplied guide JSON, not CDN content, not tab persistence (projectPersistedTabs already whitelists fields, so nothing new is persisted), and no new global registry.
- rewriteGuideTrees is applied at the one loader site that rewrites a guide body after load (the journey cover-page extras injection in docs-panel.tsx), so a render-tree rewrite cannot silently desync the counting tree. That branch cannot co-occur with prefetched content today (a prepared launch always creates a fresh tab, which has no pathContext yet), so it is defensive at the seam rather than a live bug fix.
- Explicit non-goals kept: the counting rule is unchanged, GUIDE_STATS_VERSION is not bumped, the published schema is unchanged, no per-expanded-block provenance or required author ids are added, pre-launch snippet expansion and Grafana-UI classification still run on the expanded tree, there is no second fetch, no persisted-progress migration, and the existing conservative misses for runtime step/container ids a splice shifts are preserved (an anonymous shifted step still gets no position rather than credit for a different block).
- Documentation was required in the same PR: the frozen-index commentary in global-state/active-guide-index.ts claimed freezing removed cross-path disagreement, which was false; COMPLETION-MODEL.md decision 1 and the completion-denominator-authority anchor in CONCERN_DETAILS.md are the contract anchors the issue names, so both were updated here even though docs/design is normally author-curated.

Tests follow the issue's acceptance criteria: the main regression crosses the real prepareGuideLaunch-to-ContentRenderer handoff with the real inliner, real counter and real frozen-index store, mocking only the network fetch and the snippet CDN resolver; each opening path is exercised with a fresh index-store lifecycle in both orders; coverage includes repeated references to the same snippet, a reference nested in a transparent container, a reference inside an (opaque) conditional, a failed snippet resolution with its fail-safe requiresGrafanaUi, position/percentage parity from the same evidence, overlay stability, and a negative case for an expanded payload missing its counting source. npm run check passed all 9 steps locally before this run.

## What Changed

- Added `src/lib/guide-counting-source.ts` plus `RawContent.countingSource` / `PreparedRawContent` in `src/types/content.types.ts`, so a snippet-expanded payload carries the pre-inlining guide it was expanded from. `prepareGuideLaunch` now builds its payload through `createPreparedContent` (serializing both trees in one call), and the staged-launch, pending-guide, and `OpenDocs*` option types require the prepared variant rather than plain `RawContent`.
- `ContentProcessor` in `content-renderer.tsx` selects the counting tree via `selectCountingTree` and computes the whole frozen `GuideBlockIndex` from it, so a direct open and a prepared launch of the same guide freeze one denominator and one set of positions; a known-expanded payload whose counting source is missing or unparseable still renders but publishes no canonical index and logs a `logger.warn` diagnostic instead of counting the expanded tree.
- Applied `rewriteGuideTrees` at the journey cover-page extras injection in `docs-panel.tsx` so a render-tree rewrite cannot desync the counting tree, and updated the counting contract anchors and routing docs (`COMPLETION-MODEL.md` decision 1, the `completion-denominator-authority` invariant, `CONCERNS.md`, `CONTEXT_INDEX.md`, `systemPatterns.mdc`, the frozen-index commentary in `active-guide-index.ts`, and `guide-stats/block-index.ts`).
- Tests: a new 378-line regression suite crossing the real `prepareGuideLaunch`-to-`ContentRenderer` handoff (real inliner, counter, and frozen-index store; only the fetch and snippet resolver mocked) covers both opening orders, repeated refs, refs nested in transparent and conditional containers, failed snippet resolution, position/percentage parity, overlay stability, and the missing-counting-source case; plus a `prepareGuideLaunch` unit test and prepared-content fixture updates across the launch/panel suites.

## Risk Assessment

✅ Low: The change is tightly bounded to one new pure module plus a single counting-source selection at the one `publishGuideIndex` call site, the launch handoff is type-guarded end to end by `PreparedRawContent`, every other `RawContent` producer legitimately leaves `countingSource` unset, nothing new is persisted, and I could not construct a reachable path where an expanded tree still gets counted.

## Testing

I installed dependencies (the worktree had none, which was the only setup problem) and ran the change's own targeted suites plus every other test file it touched — all green. To prove the reported failure, I reverted the two fix files to the base commit and re-ran the regression file: it now fails 8 of 12 cases with exactly the reported symptom (a one-reference guide freezing a total of 3 on the prepared path against 2 on a direct open, 5 vs 3 for repeated references, plus a phantom position for a splice-shifted anonymous step), and passes fully once the fix is restored. As committed the file only failed 2 of 12 pre-fix because a still-mounted first render republished the correct index on eviction, so I added a `closeAndReset()` helper that unmounts the surface before evicting — the only working-tree change I made. For end-user evidence I drove the real rendered guide surface, including the `MarkCompleteFooter` percentage the reader actually sees, through both opening paths with one recorded \"Do it\" click: before the fix the home-page (prepared) launch showed \"33% complete\" while the identical guide opened directly showed \"50% complete\"; after the fix both show \"50% complete\". Those four rendered surfaces are captured as HTML and as one side-by-side screenshot.

- Evidence: Reader-visible &#34;% complete&#34; on both launch paths, before vs after the fix (screenshot of the real rendered guide surface) (local file: <code>/Users/davidallen/.no-mistakes/evidence/01M2KBXDJMZVNF3Z1QBFV7J3WX/issue-1665-comparison.png</code>)
- Evidence: Side-by-side comparison page (embeds the four rendered surfaces) (local file: <code>/Users/davidallen/.no-mistakes/evidence/01M2KBXDJMZVNF3Z1QBFV7J3WX/issue-1665-comparison.html</code>)

<details>
<summary>Evidence: Rendered guide surface — before fix, prepared launch (33% complete, denominator 3)</summary>

```text
<!doctype html><meta charset="utf-8"><style>body{background:#111217;color:#ccccdc;font-family:Inter,Helvetica,Arial,sans-serif;margin:0;padding:16px}




    /* Blocker overlay visuals (used by GlobalInteractionBlocker) */
    #interactive-blocking-overlay {
      background: transparent !important;
      /* Always-visible subtle border */
      border: 1px solid rgba(170, 170, 170, 0.35);
      /* Breathing pulse layered on top */
      box-shadow: inset 0 0 0 0 rgba(170, 170, 170, 0.22);
      animation: blocker-breathe 3.2s ease-in-out infinite;
    }

    /* When a modal is active, remove breathing pulse to avoid visual clash and keep overlay subtle */
    #interactive-blocking-overlay.no-breathe {
      animation: none !important;
      box-shadow: none !important;
      border: none !important;
    }

    /* Full-screen overlay for modal blocking */
    #interactive-fullscreen-overlay {
      background: transparent !important;
      /* Gray pulse around the outside edge - more visible for full screen */
      border: 2px solid rgba(170, 170, 170, 0.4);
      box-shadow: inset 0 0 0 0 rgba(170, 170, 170, 0.3);
      animation: fullscreen-breathe 2.8s ease-in-out infinite;
    }

    /* Header overlay styling */
    #interactive-header-overlay {
      background: transparent !important;
      border: 1px solid rgba(170, 170, 170, 0.35);
      box-shadow: inset 0 0 0 0 rgba(170, 170, 170, 0.22);
      animation: blocker-breathe 3.2s ease-in-out infinite;
    }

    @keyframes blocker-breathe {
      0% {
        box-shadow: inset 0 0 0 0 rgba(170, 170, 170, 0.18);
      }
      50% {
        box-shadow: inset 0 0 0 6px rgba(170, 170, 170, 0.20);
      }
      100% {
        box-shadow: inset 0 0 0 0 rgba(170, 170, 170, 0.18);
      }
    }

    @keyframes fullscreen-breathe {
      0% {
        box-shadow: inset 0 0 0 0 rgba(170, 170, 170, 0.25);
        border-color: rgba(170, 170, 170, 0.35);
      }
      50% {
        box-shadow: inset 0 0 0 8px rgba(170, 170, 170, 0.30);
        border-color: rgba(170, 170, 170, 0.45);
      }
      100% {
        box-shadow: inset 0 0 0 0 rgba(170, 170, 170, 0.25);
        border-color: rgba(170, 170, 170, 0.35);
      }
    }
    /* Global interactive highlight styles */
    .interactive-highlight-outline {
      position: fixed;
      top: var(--highlight-top);
      left: var(--highlight-left);
      width: var(--highlight-width);
      height: var(--highlight-height);
      pointer-events: none;
      z-index: 9999;
      border-radius: 4px;
      /* Draw border clockwise using four gradient strokes (no fill) */
      --hl-color: rgba(255, 136, 0, 0.85);
      --hl-thickness: 2px;
      background:
        linear-gradient(var(--hl-color) 0 0) top left / 0 var(--hl-thickness) no-repeat,
        linear-gradient(var(--hl-color) 0 0) top right / var(--hl-thickness) 0 no-repeat,
        linear-gradient(var(--hl-color) 0 0) bottom right / 0 var(--hl-thickness) no-repeat,
        linear-gradient(var(--hl-color) 0 0) bottom left / var(--hl-thickness) 0 no-repeat;
      opacity: 0.95;
      /* Draw border animation, then breathing glow (2 cycles), then fade out
       * Total: 1625ms draw + 3s breathe + 0.5s fade = 5.125s
       * JS cleanup at 5625ms (500ms buffer after CSS completes)
       */
      animation:
        interactive-draw-border 1625ms cubic-bezier(0.18, 0.6, 0.2, 1) forwards,
        interactive-glow-breathe 1500ms ease-in-out 1625ms 2,
        interactive-outline-fade 500ms ease-out 4625ms forwards;
    }

    /* Subtle variant to reuse animation cadence for blocked areas */
    .interactive-highlight-outline--subtle {
      border-color: rgba(180, 180, 180, 0.4);
      background-color: rgba(180, 180, 180, 0.08);
      box-shadow: 0 0 0 4px rgba(180, 180, 180, 0.12);
      animation: subtle-highlight-pulse 1.6s ease-in-out infinite;
    }

    /* Hover highlight for element inspector (watch/record mode) */
    .interactive-hover-highlight-outline {
      position: fixed;
      top: var(--highlight-top);
      left: var(--highlight-left);
      width: var(--highlight-width);
      height: var(--highlight-height);
      pointer-events: none;
      z-index: 9999;
      border-radius: 4px;
      /* Same orange color as regular highlights for consistency */
      --hl-color: rgba(255, 136, 0, 0.85);
      --hl-thickness: 2px;
      background:
        linear-gradient(var(--hl-color) 0 0) top left / 0 var(--hl-thickness) no-repeat,
        linear-gradient(var(--hl-color) 0 0) top right / var(--hl-thickness) 0 no-repeat,
        linear-gradient(var(--hl-color) 0 0) bottom right / 0 var(--hl-thickness) no-repeat,
        linear-gradient(var(--hl-color) 0 0) bottom left / var(--hl-thickness) 0 no-repeat;
      /* No animation for hover - instant feedback */
      background-size: 100% var(--hl-thickness), var(--hl-thickness) 100%, 100% var(--hl-thickness), var(--hl-thickness) 100%;
      opacity: 0.95;
      /* Smooth transitions when moving between elements */
      transition: top 0.05s ease-out, left 0.05s ease-out, width 0.05s ease-out, height 0.05s ease-out;
    }

    @keyframes interactive-draw-border {
      0% {
        background-size: 0 var(--hl-thickness), var(--hl-thickness) 0, 0 var(--hl-thickness), var(--hl-thickness) 0;
      }
      25% {
        background-size: 100% var(--hl-thickness), var(--hl-thickness) 0, 0 var(--hl-thickness), var(--hl-thickness) 0;
      }
      50% {
        background-size: 100% var(--hl-thickness), var(--hl-thickness) 100%, 0 var(--hl-thickness), var(--hl-thickness) 0;
      }
      75% {
        background-size: 100% var(--hl-thickness), var(--hl-thickness) 100%, 100% var(--hl-thickness), var(--hl-thickness) 0;
      }
      100% {
        background-size: 100% var(--hl-thickness), var(--hl-thickness) 100%, 100% var(--hl-thickness), var(--hl-thickness) 100%;
      }
    }

    /* Breathing orange glow that activates after border draw completes */
    @keyframes interactive-glow-breathe {
      0%, 100% {
        box-shadow: 0 0 8px 2px rgba(255, 136, 0, 0.3);
      }
      50% {
        box-shadow: 0 0 16px 4px rgba(255, 136, 0, 0.5);
      }
    }

    /* Fade out animation for bounding box after breathing completes */
    @keyframes interactive-outline-fade {
      from {
        opacity: 0.95;
        box-shadow: 0 0 8px 2px rgba(255, 136, 0, 0.3);
      }
      to {
        opacity: 0;
        box-shadow: 0 0 4px 1px rgba(255, 136, 0, 0);
      }
    }

    /* Instant highlight - no draw animation, just breathing glow then fade
     * Instant: 2 breathe cycles (3s) + 0.5s fade = 3.5s
     */
    .interactive-highlight-outline--instant {
      animation: 
        interactive-glow-breathe 1500ms ease-in-out 2,
        interactive-outline-fade 500ms ease-out 3000ms forwards !important;
      background-size: 100% var(--hl-thickness), var(--hl-thickness) 100%, 100% var(--hl-thickness), var(--hl-thickness) 100% !important;
    }

    /* Pulsing dot indicator for small/hidden elements (< 10px width or height)
     * Threshold defined in INTERACTIVE_CONFIG.highlighting.minDimensionForBox
     */
    .interactive-highlight-dot {
      position: fixed;
      top: var(--highlight-top);
      left: var(--highlight-left);
      width: 16px;
      height: 16px;
      border-radius: 50%;
      background: rgba(255, 136, 0, 0.9);
      pointer-events: none;
      z-index: 9999;
      transform: translate(-50%, -50%); /* Center on the point */
      box-shadow: 0 0 12px 4px rgba(255, 136, 0, 0.5);
      /* Finite animation: 2 pulse cycles (3s) then fade out (0.5s) = 3.5s total
       * JS cleanup at 4000ms (500ms buffer after CSS completes)
       */
      animation: 
        interactive-dot-pulse 1500ms ease-in-out 2,
        interactive-dot-fade 500ms ease-out 3000ms forwards;
    }

    @keyframes interactive-dot-pulse {
      0%, 100% {
        transform: translate(-50%, -50%) scale(1);
        opacity: 0.9;
      }
      50% {
        transform: translate(-50%, -50%) scale(1.3);
        opacity: 1;
      }
    }

    @keyframes interactive-dot-fade {
      from {
        opacity: 0.9;
        transform: translate(-50%, -50%) scale(1);
      }
      to

... [14786 bytes truncated] ...

ainer {
      position: relative;
      height: 4px;
      background: var(--pathfinder-comment-progress-bg, rgba(255, 255, 255, 0.1));
      border-radius: 4px;
      overflow: hidden;
      margin-bottom: 16px;
    }

    .interactive-comment-progress-bar {
      position: absolute;
      top: 0;
      left: 0;
      height: 100%;
      background: linear-gradient(90deg, #3871dc 0%, #5a67d8 100%);
      border-radius: 4px;
      transition: width 0.3s ease;
    }

    /* Step dots - centered, no borders */
    .interactive-comment-dots {
      display: flex;
      align-items: center;
      justify-content: center;
      gap: 6px;
      margin: 16px 0 8px;
    }

    .interactive-comment-dot {
      width: 8px;
      height: 8px;
      border-radius: 50%;
      background: var(--pathfinder-comment-dot-bg, rgba(255, 255, 255, 0.2));
      transition: all 0.2s ease;
    }

    .interactive-comment-dot--current {
      background: #3871dc;
      box-shadow: 0 0 0 3px rgba(56, 113, 220, 0.3);
      animation: dotPulse 2s ease-in-out infinite;
    }

    .interactive-comment-dot--completed {
      background: #73bf69;
    }

    @keyframes dotPulse {
      0%, 100% {
        box-shadow: 0 0 0 3px rgba(56, 113, 220, 0.3);
      }
      50% {
        box-shadow: 0 0 0 6px rgba(56, 113, 220, 0);
      }
    }

    /* Title - clean typography */
    .interactive-comment-title {
      font-size: 16px;
      font-weight: 600;
      color: var(--pathfinder-comment-text, #ffffff);
      margin: 0 0 6px 0;
      line-height: 1.3;
    }

    /* Description - secondary text */
    .interactive-comment-description {
      font-size: 14px;
      color: var(--pathfinder-comment-text-secondary, rgba(255, 255, 255, 0.75));
      margin: 0;
      line-height: 1.5;
    }

    /* Navigation buttons - clean, minimal */
    .interactive-comment-nav-btn {
      display: flex;
      align-items: center;
      justify-content: center;
      gap: 4px;
      padding: 8px 16px;
      border: 1px solid var(--pathfinder-comment-btn-border, rgba(255, 255, 255, 0.2));
      border-radius: 6px;
      background: transparent;
      color: var(--pathfinder-comment-btn-text, rgba(255, 255, 255, 0.8));
      font-size: 13px;
      font-weight: 500;
      cursor: pointer;
      pointer-events: auto;
      transition: all 0.15s ease;
      white-space: nowrap;
    }

    .interactive-comment-nav-btn:hover:not(:disabled) {
      background: var(--pathfinder-comment-btn-hover-bg, rgba(255, 255, 255, 0.1));
      border-color: var(--pathfinder-comment-btn-hover-border, rgba(255, 255, 255, 0.3));
      color: var(--pathfinder-comment-btn-hover-text, #ffffff);
    }

    .interactive-comment-nav-btn:disabled {
      opacity: 0.4;
      cursor: not-allowed;
    }

    /* Primary button - gradient */
    .interactive-comment-nav-btn--primary {
      background: linear-gradient(135deg, #3871dc 0%, #5a67d8 100%);
      border-color: transparent;
      color: #ffffff;
    }

    .interactive-comment-nav-btn--primary:hover:not(:disabled) {
      background: linear-gradient(135deg, #4a82e4 0%, #6b74e8 100%);
      border-color: transparent;
      transform: translateY(-1px);
      box-shadow: 0 4px 12px rgba(56, 113, 220, 0.4);
    }

    /* Cancel button - subtle danger styling */
    .interactive-comment-nav-btn--cancel {
      border-color: rgba(255, 100, 100, 0.3);
      color: rgba(255, 180, 180, 0.9);
    }

    .interactive-comment-nav-btn--cancel:hover:not(:disabled) {
      background: rgba(255, 100, 100, 0.15);
      border-color: rgba(255, 100, 100, 0.5);
      color: rgba(255, 220, 220, 1);
    }

    .interactive-comment-nav-spacer {
      flex: 1;
    }

    /* Keyboard hint - subtle, integrated */
    .interactive-comment-keyboard-hint {
      display: flex;
      align-items: center;
      justify-content: center;
      gap: 6px;
      margin-top: 12px;
      font-size: 11px;
      color: var(--pathfinder-comment-kbd-text, rgba(255, 255, 255, 0.35));
    }

    .interactive-comment-kbd {
      display: inline-flex;
      align-items: center;
      justify-content: center;
      min-width: 20px;
      padding: 2px 6px;
      background: var(--pathfinder-comment-kbd-bg, rgba(255, 255, 255, 0.08));
      border: 1px solid var(--pathfinder-comment-kbd-border, rgba(255, 255, 255, 0.12));
      border-radius: 4px;
      font-size: 10px;
      font-family: 'Monaco', 'Menlo', 'Ubuntu Mono', monospace;
    }

    /* Hide interactive comment spans - they're extracted as metadata */
    span.interactive-comment {
      display: none !important;
    }

    /* Spinner animation for section running state */
    @keyframes spin {
      0% { transform: rotate(0deg); }
      100% { transform: rotate(360deg); }
    }

    /* Form validation status in comment box (for guided formfill steps) */
    .interactive-form-validation-status {
      margin-top: 10px;
      padding: 8px 10px;
      border-radius: 4px;
      font-size: 12px;
      display: flex;
      align-items: center;
      gap: 6px;
    }

    .interactive-form-validation-status.form-checking {
      background: var(--pathfinder-form-checking-bg, rgba(255, 255, 255, 0.1));
      border: 1px solid var(--pathfinder-form-checking-border, rgba(255, 255, 255, 0.2));
      color: var(--pathfinder-form-checking-text, rgba(255, 255, 255, 0.8));
    }

    .interactive-form-validation-status.form-hint-warning {
      background: rgba(255, 193, 7, 0.2);
      border: 1px solid rgba(255, 193, 7, 0.4);
      color: rgba(255, 220, 150, 1);
    }

    .interactive-form-validation-status.form-valid {
      background: rgba(115, 191, 105, 0.2);
      border: 1px solid rgba(115, 191, 105, 0.4);
      color: rgba(115, 191, 105, 1);
    }

    .interactive-form-validation-status .interactive-form-spinner {
      animation: spin 1s linear infinite;
    }

    .interactive-form-validation-status .interactive-form-warning-icon {
      flex-shrink: 0;
    }

    .interactive-form-validation-status .interactive-form-success-icon {
      flex-shrink: 0;
      animation: popIn 0.3s ease-out;
    }

    @keyframes popIn {
      0% {
        transform: scale(0);
        opacity: 0;
      }
      50% {
        transform: scale(1.2);
      }
      100% {
        transform: scale(1);
        opacity: 1;
      }
    }
  </style><body><div class="undefined css-1h60in1" data-pathfinder-content="true" style="display: flex; flex-direction: column; min-height: 0; overflow: visible; position: relative;"><div><div class="interactive-step completed" data-test-step-kind="plain" data-test-step-id="connect-datasource" data-targetaction="button" data-reftarget="button[data-testid=&quot;add-datasource&quot;]" data-step-id="connect-datasource" data-testid="interactive-step-connect-datasource" data-test-step-state="completed" data-test-fix-type="none" data-test-requirements-state="unmet"><div class="interactive-step-content"><p>Add a new data source</p></div><div class="interactive-step-actions"><div class="interactive-step-action-buttons"></div><div class="interactive-guided-completed"><div class="interactive-guided-completed-badge"><span class="interactive-guided-completed-icon" data-testid="interactive-step-completed-connect-datasource">✓</span><span class="interactive-guided-completed-text">Completed</span></div><button class="css-1nuwwf8-button" type="button" data-testid="interactive-redo-connect-datasource" title="Redo this step (execute again)" aria-disabled="false"><span class="css-1riaxdn">↻ Redo</span></button></div></div></div><p>Shared snippet step one: open the data sources page.</p><p>Shared snippet step two: confirm the health check passes.</p></div><div class="css-qezvbz" data-testid="mark-complete-footer"><div class="css-s1987a"><div class="css-xr4mdo"><div class="css-14u44gd" style="width: 33%;"></div></div><span class="css-tiv71v" data-testid="mark-complete-percentage">33% complete</span></div><button class="css-3l5qae-button" type="button" data-testid="mark-complete-button" aria-disabled="false"><svg xmlns="http://www.w3.org/2000/svg" data-testid="check" viewBox="0 0 24 24"></svg><span class="css-1riaxdn">Mark complete</span></button></div></div></body>
```
</details>

- Evidence: Rendered guide surface — after fix, prepared launch (50% complete, denominator 2) (local file: <code>/Users/davidallen/.no-mistakes/evidence/01M2KBXDJMZVNF3Z1QBFV7J3WX/after-fix-prepared-launch.html</code>)

<details>
<summary>Evidence: Visible percentage and frozen denominator per opening path</summary>

```text
before-fix: [{"pathLabel":"direct-open","visiblePercentage":"50% complete","frozenTotalBlockCount":2},{"pathLabel":"prepared-launch","visiblePercentage":"33% complete","frozenTotalBlockCount":3}]
after-fix: [{"pathLabel":"direct-open","visiblePercentage":"50% complete","frozenTotalBlockCount":2},{"pathLabel":"prepared-launch","visiblePercentage":"50% complete","frozenTotalBlockCount":2}]
```
</details>
<details>
<summary>Evidence: Regression reproduced against base commit cdb0befa (8 of 12 fail before the fix)</summary>

<code>● counts one reference as one block on both paths, though the expanded tree holds three — Expected: 2, Received: 3&#10;● publishes the whole index the source guide computes, not a corrected denominator&#10;● resolves the same position and percentage from the same evidence&#10;● still withholds a position from an anonymous step a splice shifted — Expected: 0, Received: 1&#10;● agrees on repeated references to the same snippet — Expected: 3, Received: 5&#10;● agrees on a reference nested in a transparent container — Expected: 2, Received: 3&#10;● cannot establish a different total by opening the launch path first — Expected: 2, Received: 3&#10;● declines to publish an index for an expanded payload that lost its counting source&#10;Tests: 8 failed, 4 passed, 12 total</code>

```text
  ● canonical snippet counts across direct and prepared launches › counts one reference as one block on both paths, though the expanded tree holds three
    Expected: 2
    Received: 3
  ● canonical snippet counts across direct and prepared launches › publishes the whole index the source guide computes, not a corrected denominator
  ● canonical snippet counts across direct and prepared launches › resolves the same position and percentage from the same evidence
  ● canonical snippet counts across direct and prepared launches › still withholds a position from an anonymous step a splice shifted
    Expected: 0
    Received: 1
  ● canonical snippet counts across direct and prepared launches › agrees on repeated references to the same snippet
    Expected: 3
    Received: 5
  ● canonical snippet counts across direct and prepared launches › agrees on a reference nested in a transparent container
    Expected: 2
    Received: 3
  ● canonical snippet counts across direct and prepared launches › cannot establish a different total by opening the launch path first
    Expected: 2
    Received: 3
  ● canonical snippet counts across direct and prepared launches › declines to publish an index for an expanded payload that lost its counting source
Tests:       8 failed, 4 passed, 12 total
```
</details>
- Outcome: ⚠️ 1 info across 1 run (7m40s)

## Pipeline

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>
<summary>⚠️ **Review** - 3 infos</summary>

- ℹ️ `src/components/content-renderer/content-renderer.tsx:816` - Product consequence worth stating explicitly: switching the prepared path to the pre-inlining tree also removes numerator resolution it previously had. Concretely, for `blocks: [snippet-ref, interactive(anonymous, action: &#39;button&#39;, reftarget: ...)]` the old prepared path indexed the expanded tree (no snippet-ref present), so `computeGuideBlockIndex`&#39;s `sawSnippetRefSibling` guard never fired and the anonymous step got `positionsByStepId` position 3 of 3 -&gt; a &#39;Do it&#39; read 100%. Now the index comes from the pre-inlining tree, the guard fires, `positionsByStepId` is empty (asserted in content-renderer.canonical-snippet-counts.test.tsx), and the same evidence resolves to position 0 -&gt; 0%. Same for any id-bearing block that lives *inside* a snippet: it is absent from the counting tree, so its evidence resolves to nothing. Both are exactly the pre-existing direct-open behaviour and both are explicitly authorized in the intent (&#39;an anonymous shifted step still gets no position rather than credit for a different block&#39;), so this is parity at the lower value by design, not a defect. Flagging only so the visible effect - snippet-heavy guides that used to report progress on a home-page launch now report 0% there too - is a recorded, deliberate trade rather than a surprise in the field.
- ℹ️ `src/components/docs-panel/types.ts:80` - `DocsPanelModelOperations.loadTab`&#39;s `prefetched?: RawContent` (and the matching params on `loadTab`, `loadTabContent`, `loadDocsTabContent` in docs-panel.tsx:484/509/764) were left as `RawContent` while `OpenDocsOptions`, `OpenLearningJourneyOptions`, `PendingGuide`, `StagedLaunchPayload` and `PreparedGuideLaunch` were all narrowed to `PreparedRawContent`. Every caller today passes prepared content or omits it, so there is no live bug - but this is the one remaining seam where a future internal caller could hand an already-expanded body straight to `finishTabSuccess` -&gt; `ContentRenderer` with no `countingSource` and no compile error, which is the failure mode `PreparedRawContent` was introduced to make unspellable. Narrowing these four declarations to `PreparedRawContent` extends the guarantee to the seam that actually feeds the renderer.
- ℹ️ `src/lib/guide-counting-source.ts:55` - `selectCountingTree(renderedGuideJson, source)` takes the render JSON in only to hand the same string back, which forces the caller to re-detect &#39;this is the tree I already parsed&#39; with a `selection.guideJson === html` string comparison before deciding whether to `JSON.parse` again. A result that names the choice instead of echoing the input - e.g. `{ use: &#39;rendered&#39; } | { use: &#39;preserved&#39;; guideJson } | { use: &#39;none&#39; }` - drops the parameter, drops the identity comparison at content-renderer.tsx:741, and makes the three-way decision readable at the call site without knowing that `processedContent` is reference-equal to `content.content` for JSON guides. Behaviour-preserving.
</details>

<details>
<summary>⚠️ **Test** - 1 info</summary>

- ℹ️ `src/components/content-renderer/content-renderer.canonical-snippet-counts.test.tsx:113` - The committed regression suite passed against the buggy code for 10 of its 12 cases: `act(() =&gt; evictAllGuideIndexes())` between the two opening paths left the first render mounted, so the eviction bump made that still-mounted tree re-publish the correct pre-inlining index, and publishGuideIndex&#39;s first-wins idempotence then swallowed the expanded tree&#39;s wrong count. Auto-fixed by adding a `closeAndReset()` helper that unmounts the surface (the reader closing the guide) before evicting; 8 of 12 now fail pre-fix and all 12 pass post-fix.
- <code>`npm install` (worktree had no node_modules, so jest-environment-jsdom was missing)</code>
- <code>`npx jest --coverage=false src/components/content-renderer/content-renderer.canonical-snippet-counts.test.tsx src/lib/guide-counting-source.test.ts src/components/docs-panel/utils/prepare-guide-launch.test.ts` — 38 tests pass on the target commit</code>
- <code>Fail-before check: reverted `content-renderer.tsx` + `prepare-guide-launch.ts` to base `cdb0befa` and re-ran the regression file — 8 of 12 fail (totals 3 vs 2, 5 vs 3, a phantom position from a shifted anonymous step, and an index published for a counting-source-less expanded payload); restored the fix and all 12 pass</code>
- <code>Strengthened `content-renderer.canonical-snippet-counts.test.tsx` with a `closeAndReset()` helper (unmount the surface before evicting the index store) so the second opening path is a genuinely fresh lifecycle</code>
- <code>Manual end-user evidence harness (transient, since removed): rendered `ContentRenderer` + `MarkCompleteFooter` for a guide of one interactive step followed by one `snippet-ref` (2 blocks), recorded the &#34;Do it&#34; completion via the real `markStepCompleted`/`refreshAndNotifyGuideProgress`, and serialized the rendered surface on both paths against pre-fix and post-fix source</code>
- <code>Screenshotted the four rendered surfaces side by side with Playwright chromium (`issue-1665-comparison.png`)</code>
- <code>`npx jest --coverage=false src/components/Home/HomePanel.test.tsx src/components/docs-panel/docs-panel.load-tab-content.test.ts src/components/docs-panel/hooks/useAutoOpenListener.cold-sidebar-relay.test.ts src/components/docs-panel/hooks/useAutoOpenListener.test.ts src/components/docs-panel/pendingGuideRouter.test.ts src/global-state/guide-launch.test.ts src/global-state/active-guide-index.test.ts src/lib/guide-stats` — 210 tests pass</code>
- <code>Read `projectPersistedTabs` to confirm the tab persistence whitelist still projects a fixed field list, so `countingSource` is not persisted</code>
</details>

<details>

- ℹ️ `src/global-state/active-guide-index.ts:36` - `ActiveGuideIndexDenominatorSource` still declares `live-post-inlining-degraded` under the comment &#34;Where the frozen denominator came from&#34;, but this change&#39;s decision is that a payload which lost its counting source publishes no index at all rather than a degraded one — so that variant now has no plausible producer by design, not just none today (it was already producerless before this change, referenced only by active-guide-index.test.ts). Removing it would be a code and test change, which is outside this documentation phase. Worth a follow-up: either delete the variant or record in the type&#39;s comment why a degraded stamp is retained as spellable-but-unreachable.
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>


moxious:
Follow-up filed for the one `ask-user` finding from the documentation gate: #1906 — `live-post-inlining-degraded` has no producer by design after this change, so the variant should either be deleted or documented as deliberately retained. Deliberately not done here: it is a code and test change to a shipped type, outside this PR's subject.

tomglenn (changes requested):
## Summary

This PR ensures guide progress percentages stay consistent across opening paths by introducing a **counting tree** that tracks the canonical pre-inlining guide structure while the renderer paints an expanded version. Previously, direct opens and prepared launches produced different denominators for the same guide (snippet refs expand to multiple blocks in the render tree but count as one position), and the frozen index locked in whichever denominator the user's first path chose. The core fix is sound; a few hardening steps would make the new seam safe against future changes.

## Findings

### Correctness
- `selectCountingTree` (`src/lib/guide-counting-source.ts:55–67`) — both branches are sound: absent `countingSource` means the rendered tree is itself pre-inlining; `kind: 'pre-inlining'` surfaces the preserved canonical tree from a prepared launch.
- `rewriteGuideTrees` (`src/lib/guide-counting-source.ts:75–84`) — applying one rewrite to both trees prevents the cross-path desync reappearing when later loaders inject blocks. Clean.
- Edge cases partially covered — repeated snippet refs and snippet-in-snippet are not explicitly tested; there's no integration test confirming the two trees stay in agreement under these patterns.

### Design & typing
- `createPreparedContent` (`src/lib/guide-counting-source.ts:37–45`) serializes both trees in one call, which *should* prevent swapping expanded ↔ counting guides — but nothing enforces it at runtime. A hand-constructed `PreparedRawContent` could swap `content` ↔ `countingSource.guideJson` and produce inverted counts. Consider a runtime assertion, or document the contract explicitly on the type.

### Tests
- The integration test (`src/components/content-renderer/content-renderer.canonical-snippet-counts.test.tsx`) exercises the real `prepareGuideLaunch`, inliner, counter, and frozen-index store, demonstrates the 2-vs-3 bug, and confirms the fix. Good.
- `rewriteGuideTrees` is only tested with trivial rewrites — not a representative structural mutation (e.g. `substituteVariables`).
- No tests for persistence across content-key evictions, or for the no-usable-counting-tree path.

### Security & risk
- Unvalidated `JSON.parse` (`src/components/content-renderer/content-renderer.tsx:745`) — `countingSource.guideJson` is parsed with no schema validation; the contract that it is well-formed is a policy assumption with no enforcement. Add a schema check or document the trust boundary.
- Silent degradation (`src/components/content-renderer/content-renderer.tsx:786–793`) — when the counting tree is unavailable, rendering proceeds with no canonical index and only a `logger.warn`; there's no telemetry, so a broken prepared payload is invisible in production. Add telemetry.
- No integrity check that `countingSource` is the tree `expandedGuide` was expanded from (block-count/ID/structure).

## Verdict: Request changes

Top blockers:
1. Unvalidated `JSON.parse` trust boundary (`content-renderer.tsx:745`) — add a schema check or document the assumption.
2. Silent degradation with no telemetry (`content-renderer.tsx:786–793`) — emit telemetry when a prepared payload has no usable counting tree.
3. `PreparedRawContent` construction is trusted but not verified — add an integrity check or document that `createPreparedContent` must be the sole producer.

Before approval:
- Runtime validation or a documented trust boundary for `countingSource.guideJson`.
- Telemetry when a prepared payload has no usable counting tree.
- An integration test for `rewriteGuideTrees` with a real structural mutation, or a note that trivial rewrites are sufficient.

The core fix is correct and the test proves it resolves the reported bug — these are hardening steps to keep the seam from becoming a future silent-failure or trust-boundary issue.

moxious:
Thanks for the read @tomglenn — the `rewriteGuideTrees` and `selectCountingTree` analysis is right, and one of the test observations found a real gap that I've now closed. I verified each of the three blockers against the source and against #1665's stated design, and I don't think any of them hold. Detail below so you can check my working.

### Blocker 1 — unvalidated `JSON.parse` at `content-renderer.tsx:745`

The string is already schema-validated one hop earlier, and it never crosses a boundary:

- `prepare-guide-launch.ts:105-122` — `JSON.parse`, **then `validateGuide()`**, with an early return on failure. Only a validated guide reaches `createPreparedContent`.
- `guide-counting-source.ts:42` — `countingSource.guideJson` is `JSON.stringify` of that validated object.
- `global-state/guide-launch.ts:1-22` — the payload **never rides a public channel**. Only an opaque `crypto.randomUUID()` key goes over the `pathfinder-auto-open-docs` CustomEvent and the cold-sidebar queue; the payload sits in a module-owned `Map`, consumed once, with URL-match and a 60s TTL. `panelModeManager.setPendingGuide` is module-owned memory too.

So line 745 parses a same-realm re-serialization of an object validated a moment earlier. Anyone who can write that `Map` already has arbitrary JS in the page, at which point schema-checking a string buys nothing. Worth noting the render tree gets the same bare `JSON.parse` at line 716 (pre-existing) — validating only the counting tree would be asymmetric, and the parse already fails safe to `null`.

On documenting it: `content.types.ts:19-21` states the boundary — *"never author-supplied guide JSON, never fetched from a CDN, never persisted with the tab"* — which is #1665 item 7 verbatim.

### Blocker 2 — silent degradation with no telemetry

`logger.warn` **is** telemetry. `src/lib/logging.ts:50-53`:

```ts
warn(message, context) {
  console.warn(sanitized, context ?? '');
  getFaroTelemetry()?.pushFaroLog('warn', sanitized, sanitizeContext(context));
}
```

It ships to Faro with the `content_key`. `passesActivityGate` (`filtering.ts:216`) latches open once a Pathfinder surface opens, which is necessarily true when `ContentRenderer` is painting a guide, so it isn't gated off. #1665 item 5 asked for exactly this: *"Surface a diagnostic through existing logging."*

Separately, that branch is unreachable today. `UnavailableCountingSource` is never constructed in production source — `createPreparedContent` is the sole producer of `countingSource` and always writes `kind: 'pre-inlining'`. The only other route to `countingGuide === null` is a parse failure on a string that came from `JSON.stringify`. The type exists so the state is representable and the guard is testable, which is what the negative test at `:361` does.

### Blocker 3 — construction trusted but not verified

The "document that `createPreparedContent` is the sole producer" half is already there (`content.types.ts:65-70`).

The integrity-check half runs into #1665's non-goals. Verifying that `countingSource` is the tree `expandedGuide` came from needs a map from expanded children back to snippet occurrences, and item 6 says *"Mapping every expanded child back to a snippet occurrence is a separate capability, not required here."* A block-count comparison would be worse than useless — the two counts **legitimately differ**, which is the entire purpose of the seam. Item 2 specifies the mechanism that shipped: *"Make a prepared/expanded payload require that counting input in its internal type/constructor contract."*

### Tests

Three of the five observations don't match the file:

| Observation | File |
|---|---|
| repeated snippet refs not tested | `:237` "agrees on repeated references to the same snippet" |
| no integration test that the trees agree | that's the whole 12-test file, through the real `prepareGuideLaunch` + inliner + counter + store |
| no test for the no-usable-counting-tree path | `:361` constructs `{kind:'unavailable'}` and asserts both survival and the warn; also `guide-counting-source.test.ts:61` |

Snippet-in-snippet isn't a supported feature — `spliceBlocks` (`inline-refs.ts:135`) pushes `...resolution.snippet.blocks` without re-splicing, so expansion is single-pass.

**The `rewriteGuideTrees` observation was a real gap, and thank you for it.** You named `substituteVariables`, which is text-level and never touches the guide tree — but the actual call site, `injectJourneyExtrasIntoJsonGuide` at `docs-panel.tsx:539`, *does* append a counted block, and it was only covered by synthetic rewrites. Pushed in 5d94fe89:

1. **The real structural rewrite through both paths.** Removing the counting-tree rewrite from `rewriteGuideTrees` makes it fail 2-vs-3, so it locks the regression rather than just exercising the function.
2. **The inliner's single pass**, as an explicit invariant: a reference inside a snippet body survives into the render tree and is still one position in the counting tree.
3. **A store reset between paths in the ordering test at `:333`.** This one is mine — without the reset the second assertion only re-proved `publishGuideIndex` idempotency and would have held even if the direct path counted 3, which is not what its comment claimed.

### Related decision of record

On the producerless-variant point, #1906 already records the sibling case for `denominatorSource` and notes that this PR is what makes `live-post-inlining-degraded` producerless *by design* — because it settled that a payload which lost its counting source publishes nothing rather than a degraded index. Same reasoning applies to `UnavailableCountingSource`: the type makes the state representable and the guard testable, and the decline-to-publish contract is deliberate rather than an oversight. Whichever disposition #1906 lands on should probably cover both.

Re-requesting review. Happy to be wrong on any of the above if you can point at a path I've missed — particularly on blocker 1, since that's the one where I'd rather over-correct than under-.

tomglenn (approved):
# PR #1905 Re-Review – Final Assessment

## Summary

David has addressed the test-coverage gap and responded substantively to the three blockers. The production happy path is now defensible: guides validate before serialization, telemetry covers degradation cases, and the sole-producer pattern is documented. Runtime guarantees remain softer than ideal—no schema validation at the parse site, no integrity checks on handoff—but acceptable for the intended threat model (normal app code, not adversarial corruption).

---

## Blocker Resolution

### 1. Unvalidated JSON.parse at trust boundary (content-renderer.tsx:745)
**Status: MOSTLY RESOLVED**

The primary production path (prepareGuideLaunch) validates guides via `validateGuide()` before serializing to `countingSource.guideJson`, so the parse consumes pre-validated data. The parse site is wrapped in try/catch.

**Caveat:** The parse site itself has no local schema validation—it trusts provenance. If alternate producers or corrupted state reached it, there's no self-defense. Warden confirms this is acceptable for the current threat model; Delver notes the gap.

### 2. Silent degradation with no telemetry (no-usable-counting-tree path)
**Status: RESOLVED**

`logger.warn` at content-renderer.tsx:816-827 ships to Faro (logging.ts:50-53). The UnavailableCountingSource case appears unreachable in normal production (createPreparedContent always produces 'pre-inlining' kind), but telemetry is in place if it occurs.

### 3. PreparedRawContent construction trusted but not verified
**Status: PARTIALLY ADDRESSED**

Sole-producer documentation exists (content.types.ts:62-65), and `createPreparedContent` is the disciplined constructor. However, there's no runtime integrity check (no block-count comparison, no structure validation). David's rebuttal is correct: block counts legitimately differ post-transform, so comparison isn't a signal. The gap is robustness against accidental alternate producers or corrupted handoff state. Acceptable IF the threat model is "normal app code," but TypeScript alone doesn't enforce sole production at runtime.

---

## Test Coverage

Integration tests for `rewriteGuideTrees` with structural mutations now exist (content-renderer.canonical-snippet-counts.test.tsx). 12-test suite covers snippet expansion, nested refs, journey extras. Before-approval ask is met.

---

## New Issues

None. The 60s TTL Map concern from the original review doesn't apply—it's a consume-once private field.

---

## Recommendation

**Approve.** The production path is sound for the intended use case, tests are comprehensive, and David's rebuttals are well-reasoned. The remaining gaps (no local schema validation, no runtime integrity checks) are robustness issues, not security holes. They don't justify blocking this PR, but consider follow-up hardening if the threat model expands.

**Top drivers:**  
1. Primary blocker (unvalidated parse) is resolved via upstream validation.  
2. Integration test coverage now exists for the rewrite transforms.

Verdict: approve

Issue #1811, opened 2026-09-07
Block editor: JSON-mode exit trap for guides with a duplicate-heading error


moxious:
## Summary

Follow-up from #1764 (promote the duplicate-title heading check from WARN to ERROR).

An author who opens JSON mode in the block editor on an existing, already-published guide that trips the (now hard-error) duplicate-heading check gets stuck: `handleExitJsonMode` re-validates at the strict default and refuses to switch back to visual/preview mode, and `handleJsonUndo` re-validates `originalJson` too, so undo doesn't rescue them either. The only way out is hand-editing the heading out of the raw JSON.

This happens even though the guide renders fine at runtime (the runtime loaders — `json-parser.ts`, `content-fetcher.ts`, `backend-guide.ts`, `prepare-guide-launch.ts` — all opt into `allowDuplicateHeading: true`) and loads into the block editor fine — it's specifically the JSON-mode pane's exit/undo/change/restore path (`src/components/block-editor/hooks/useJsonModeHandlers.ts`) that's affected.

## Why this needs real design, not a quick flag

An initial fix (applying `allowDuplicateHeading: true` to `handleExitJsonMode`/`handleJsonUndo`, then consistently to `handleJsonChange`/`restoreJsonMode` too) was attempted and reverted during #1764's review, because `handleExitJsonMode` is the same function that gates applying newly-typed/pasted JSON content back to the visual editor. Making it lenient unconditionally means an author can select-all + paste in a **brand-new** guide that has this problem and have it silently accepted — defeating the "block editor should block on this error" decision from #1764.

A correct fix needs to distinguish:

- "I'm just leaving a pre-existing guide's JSON view alone (or making an unrelated edit)" → should not be trapped by a cosmetic, already-present duplicate heading.
- "I typed or pasted in materially new content that (re)introduces or worsens this problem" → should still be blocked, matching every other authoring gate.

That likely means snapshotting whether the duplicate-heading condition was already present when JSON mode was entered (or diffing the specific heading text), not a blanket per-call-site option flag.

## Scope

- `src/components/block-editor/hooks/useJsonModeHandlers.ts` — `handleExitJsonMode`, `handleJsonChange`, `restoreJsonMode`, `handleJsonUndo`.
- Out of scope: `block-import.ts` (new-guide import) should stay strict regardless — this is only about a session already open on an existing guide.

moxious:
## 🧭 Implementation readiness

[...]

PR #1907, opened 2026-09-15
fix(validation): gate unsupported guided step actions at authoring time

### LINKED ISSUE #1525: Guided blocks accept navigate/popout at validation, then throw at runtime

moxious:
## Bug Fix

A guided block is the kind that highlights something on screen and waits for the reader to do it themselves — click a button, hover a menu, type into a field. Two of the step actions an author could choose, navigate and popout, aren't things a reader does to anything on screen: the app performs them on its own, so a guided step using one has nothing to wait for. Validation accepted them anyway, so an author could write a guided step meaning "now go to Explore", publish it, and never be told anything was wrong — and the reader would work through the guide, reach that step, and watch it simply fail, with the guided block stopping there and no way to continue. This PR rejects those two actions inside a guided block while the guide is being written, in the CLI validator and the block editor, so the author learns about it before publishing and can use a separate step or an automated multistep block instead. Guides already published keep loading and behaving exactly as they do today — the change only prevents new guides from being written that way.

## Intent

Closes #1525 by implementing PR 3 of the three-PR split that the issue's own 'Decisions of record' comment recommends. PR 1 already merged as #1757 (canonical INTERACTIVE_ACTION_TYPES tuple, hub typing, DOM ingress decoder); PR 2 (deleting the dead 'sequence' path) is deliberately NOT in scope for this PR and stays a separate follow-up. The user chose PR 3 specifically because it is the only user-facing bug of the three.

## Key Deliberate Decisions & Choices

- 'author-time strict, runtime graceful'. I found that ContentJsonSchema.safeParse is a HARD RUNTIME LOAD GATE (package-engine/loader.ts:73, online-cdn-resolver.ts:108, recommender-resolver.ts:125) which returns validation-error for the WHOLE package. The issue's decisions-of-record assumed rejection landed only at 'pathfinder validate' time and therefore proposed a tightened Zod schema. The user was shown this and deliberately rejected 'strict everywhere', because tightening the shared schema would make any already-published guide with one bad guided step stop loading ENTIRELY rather than stalling on that one step. So: do NOT expect a Zod schema change. The restriction is implemented as a post-Zod semantic check that mirrors the allowDuplicateHeading option already in validate-guide.ts — error for authoring gates (CLI validate, block editor), downgraded to an advisory for the three runtime guide loaders which now pass allowUnsupportedGuidedAction: true. This deviates from the literal text of the issue's decision §2 by design, with the user's explicit approval.
- JsonStepSchema is deliberately UNCHANGED. A 'popout' step inside a *multistep* is valid and is pinned by validate-guide.test.ts:289. Narrowing the shared step schema would break that correct test. Only guided blocks are wrong.
- Runtime outcome deliberately PRESERVES today's observable behaviour: an unsupported verb still resolves the step as 'error'. The user explicitly rejected the alternatives of skipping the step or degrading it to a noop, because degrading to noop would report a step complete that never did what it said. logger.error became logger.warn since a broken authored verb is not an engine fault.
- KNOWN_TARGET_ACTIONS (cross-tab.types.ts) and GUIDED_VERBS (live-tab-executor.ts) are deliberately left as hand-written literals, per decision §5 on the issue: deriving a cross-tab wire receive-gate from a local action union lets a refactor silently widen what one tab accepts from another. live-tab-executor.ts is byte-unchanged. Instead of deriving them, the executor test pins the two lists in agreement BEHAVIOURALLY.
- 'quiz' in constants/interactive-config.ts is treated as dead (added with WYSIWYG badge metadata in #380, editor removed in #414). ACTION_TYPES is deliberately left alone; whether ACTION_BADGES/getActionBadge get deleted is a separate filed cleanup, not folded in.
- AuthoredGuidedAction was introduced because json-parser builds guided internalActions through loosely-typed ParsedElement props, so the 7-verb reality genuinely reaches the handler. Without it the runtime guard would be redundant per the type while still necessary at runtime. This is what makes the guard honest rather than dead.
- assertExhaustive returns void by design in this repo, so the switch default keeps a fallback 'return error' rather than throwing.

## Pre-Existing Unrelated Gaps

found and pinned rather than fixed: 'noop' is a verb the guided handler drives but it is absent from the receive gate's KNOWN_TARGET_ACTIONS, so a guided block containing a noop step cannot be relayed cross-tab at all. Fixing that means changing a wire contract, which §5 puts out of scope here; it is recorded as a test plus a comment so it is visible rather than surprising, and will be filed as a follow-up issue.

Docs were updated in the same PR (json-guide-format.md gains an 'Actions a guided step accepts' section and table; AGENT-AUTHORING.md notes the narrower guided parent for add-step), since authors and authoring agents need to know the new rule.

## What Changed

- Adds `src/validation/guided-action-validator.ts`, a post-Zod walk over guided blocks (recursing through `section`, `assistant`, and `conditional` containers) that reports any guided step whose verb `GuidedHandler` cannot drive. `validate-guide.ts` raises it as an `unsupported_guided_action` error by default and downgrades it to an advisory when the new `allowUnsupportedGuidedAction` option is set, mirroring the existing `allowDuplicateHeading` split; the four runtime loaders (`json-parser.ts`, `content-fetcher.ts`, `content-fetcher/backend-guide.ts`, `prepare-guide-launch.ts`) pass it so an already-published guide keeps loading. `build-snippets` errors on the same check over a snippet body's blocks, while the online snippet resolver warns without dropping the snippet. `JsonStepSchema` is unchanged, so `popout` inside a multistep stays valid.
- Introduces `GUIDED_DOM_ACTION_TYPES`, `GUIDED_ACTION_TYPES` (constrained by `satisfies readonly InteractiveActionType[]`), and the `AuthoredGuidedAction` type in `src/types/interactive-actions.types.ts`. `guided-handler.ts` replaces the `as 'hover' | 'button' | 'highlight' | 'formfill'` casts with an `isGuidedDomActionType` guard that logs a warning and settles the step as `skipped` when skippable or `error` otherwise; the completion listener now takes the narrowed action type as a parameter and its switch default uses `assertExhaustive`. `interactive-guided.tsx` skips auto-detection for a verb no detector matches.
- Tests: new `guided-action-validator.test.ts`; guided-handler, build-snippets, snippet-resolver, and prepare-guide-launch cases for unsupported verbs; and `live-tab-executor.test.ts` pins `KNOWN_TARGET_ACTIONS` against `GUIDED_VERBS` behaviourally, including the pre-existing gap that `noop` is absent from the receive gate. Docs updated: an "actions a guided step accepts" section in `json-guide-format.md`, the narrower guided parent in `AGENT-AUTHORING.md`, the new checks in `CLI_TOOLS.md`, the receive-gate note in `CROSS_TAB_CONTROLLER.md`, and the contract anchor in `CONCERN_DETAILS.md`.

## Risk Assessment

⚠️ Medium: the authoritative intent text still says an unsupported verb resolves as 'error' while the shipped (separately authorized) code resolves a skippable one as 'skipped', and the prose reconciliation the user directed in the prior round was not applied, so a human should confirm the record before merge.

## Testing

Author a guide with a guided block using one of the two verbs:

```json
{ "type": "guided", "content": "Now do it yourself.",
  "steps": [{ "action": "navigate", "reftarget": "/explore" }] }
```

1. **Authoring rejects it.** `pathfinder-cli validate <file>` exits 1, naming the step path and the remedy. On `main` the same file passes clean — that gap is the bug.
2. **Runtime still loads it.** Open an already-published guide of that shape in the sidebar: every block still renders and only that one step fails. This is the whole point of the approach, so it's the most valuable thing to confirm.
3. **Multistep is untouched.** Move the same step into a `multistep` block — validation passes. `JsonStepSchema` is deliberately unchanged.
4. **Snippets.** `build-snippets` errors on a guided `navigate` nested in a snippet body and writes no `index.json`; the online resolver only warns, so published snippets keep resolving.
5. **Runtime outcome of a bad step** — the one behaviour change: skippable → `skipped` and the run continues; non-skippable → step fails. (Skippable `popout` was previously `error`.)
6. **Tests:** `npx jest src/validation/guided-action-validator src/interactive-engine/action-handlers/guided-handler src/integrations/cross-tab/live-tab-executor --coverage=false`
7. **The ratchet bites:** add a bogus verb to `GUIDED_ACTION_TYPES` and `tsc` fails at the `satisfies` clause naming it; drop a verb from `GUIDED_VERBS` and a named executor test fails.

## Pipeline

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>
<summary>⚠️ **Review** - 2 issues (1 warning, 1 info)</summary>

- 🚨 `src/components/docs-panel/utils/prepare-guide-launch.ts:108` - Missed the fourth runtime guide loader, so an already-published guide with a guided navigate/popout step now fails to open at all — the exact regression the intent&#39;s key decision was designed to prevent.

The intent states the required behaviour: tightening would &#34;make any already-published guide with one bad guided step stop loading ENTIRELY rather than stalling on that one step&#34;, and the fix is &#34;downgraded to an advisory for the three runtime guide loaders which now pass allowUnsupportedGuidedAction: true&#34;. Three loaders were updated (json-parser.ts:127, content-fetcher.ts:233, content-fetcher/backend-guide.ts:82) but `prepareGuideLaunch` — a fourth runtime loader — still calls `validateGuide(guide, { allowDuplicateHeading: true })` with no `allowUnsupportedGuidedAction`.

This is not an inference from code shape: docs/design/CONCERN_DETAILS.md:69 is the contract anchor for this exact layering and enumerates it explicitly — &#34;the four runtime guide loaders (`json-parser.ts`, `content-fetcher.ts`, `content-fetcher/backend-guide.ts`, `prepare-guide-launch.ts`) all do [pass the tolerance flag] ... The flag therefore belongs on every path that renders a guide and on none that ships one; a new runtime loader added without it silently blanks such a guide.&#34;

Concrete failing sequence: a guide published before this gate existed contains `blocks[0] = { type: &#39;guided&#39;, steps: [{ action: &#39;navigate&#39;, reftarget: &#39;/explore&#39; }] }`. A reader clicks it on the home page or My learning -&gt; MyLearningTab.tsx:112 calls prepareGuideLaunch -&gt; Zod passes -&gt; validateGuidedActions returns one issue -&gt; validate-guide.ts:140 pushes a `unsupported_guided_action` error because the flag is absent -&gt; `isValid: false` -&gt; prepare-guide-launch.ts:110 returns `{ ok: false, errorCode: &#39;schema-invalid&#39; }` -&gt; MyLearningTab.tsx:133 publishes the alert &#34;Could not open the guide / Something went wrong while loading the guide.&#34; The guide never renders. Opening the same guide through the docs-panel fetch path still renders it (content-fetcher passes the flag), so the same guide is reachable from one surface and dead from another.

The module&#39;s own header comment also becomes false: it documents &#34;the same `validateGuide` gate `parseJsonGuide` applies, so nothing that renders today is rejected here&#34; — parseJsonGuide now passes a flag this call does not.

Earliest supported shared boundary: add `allowUnsupportedGuidedAction: true` to this call site (one line, mirroring the sibling loaders). Nothing deeper is needed — validate-guide.ts already owns the error/advisory split.
- ⚠️ `src/validation/guided-action-validator.ts:26` - The new author-time gate cannot reach snippet bodies, so the original bug stays fully reachable for any guided block authored inside a snippet, with no author-time signal.

`validateGuidedActions(guide: JsonGuide)` only accepts a whole guide, and the only gate for snippet bodies is `JsonSnippetSchema` (Zod-only) in src/cli/commands/build-snippets.ts:49 and src/snippet-engine/online-snippet-resolver.ts:57. `JsonSnippetSchema` allows any `JsonBlock` except `snippet-ref` (json-snippet.schema.ts:49), so a `guided` block with a `navigate` step is valid in a snippet body.

Concrete reachable sequence: a snippet body `{ id: &#39;x&#39;, title: &#39;X&#39;, description: &#39;d&#39;, blocks: [{ type: &#39;guided&#39;, content: &#39;c&#39;, steps: [{ action: &#39;navigate&#39;, reftarget: &#39;/explore&#39; }] }] }` passes `build-snippets` and publishes to the CDN. A guide references it via a `snippet-ref` block. At render time `prepareGuideLaunch` / `parseJsonGuide` validate the guide BEFORE `inlineSnippetRefsInGuideWithStatus` expands the ref (prepare-guide-launch.ts:105 then :118), so the guided navigate step is never seen by the gate at any point. The reader gets the failed step; no author ever sees an error.

Earliest supported shared boundary: change the walker to accept `JsonBlock[]` (it only reads `guide.blocks`) and call it from `build-snippets.ts` alongside the Zod parse, keeping `validateGuide` as a thin wrapper that passes `guide.blocks`. That is the same boundary the snippet publish gate already owns, and it needs no new traversal code.
- ⚠️ `docs/design/CONCERN_DETAILS.md:69` - The `guide-schema-and-contracts` contract anchor documents the `allowDuplicateHeading` error/advisory layering in detail, including the authoritative list of authoring gates and runtime loaders. This change adds a second option (`allowUnsupportedGuidedAction`) that follows the identical layering, but the anchor is unchanged, so the repo&#39;s record of that pattern now covers only one of its two instances.

AGENTS.md (&#34;Extending existing capabilities&#34;) requires stating whether a change follows, extends, or replaces the established contract and updating the anchor when an implementation establishes or replaces one. This is the concern&#39;s own pattern being extended, and the anchor&#39;s own loader enumeration is exactly the evidence that would have caught the missed `prepare-guide-launch.ts` call site above — which is why keeping it current is load-bearing rather than cosmetic.

Suggested edit: extend the #1764 sentence to say the same layering now carries a second flag, `allowUnsupportedGuidedAction`, over the guided-verb gate, and repeat the four-runtime-loader / authoring-gate split for it.
- ℹ️ `src/components/block-editor/forms/StepEditor.tsx:205` - The block editor still offers `navigate` and `popout` in the step action picker when editing a guided block, so an author can pick a verb the same editor&#39;s lint will then reject.

`ACTION_OPTIONS` is a module-level constant built from all of `INTERACTIVE_ACTIONS` and is used unfiltered at lines 765 (edit) and 1075 (add). `StepEditor` already receives `isGuided` (StepEditorProps:266, defaulted at :337) and branches on it a dozen times, so the narrowing information is in hand. Record mode compounds it: line 575 notes recorded steps go straight into a multistep/guided `steps` array, so recording a navigation into a guided block silently authors a now-invalid step.

This does not contradict the intent — the block editor does gate the guide via guide-lint.ts:189, and docs/developer/interactive-examples/json-guide-format.md accurately says the editor rejects them. It is a nice-to-have: filter `ACTION_OPTIONS` by `isGuided` against `GUIDED_ACTION_TYPES` so the invalid choice is never offered, rather than offered and then refused.
- ℹ️ `src/integrations/cross-tab/live-tab-executor.test.ts:373` - `flushExecutor` is a fixed 50 ms real-timer sleep, and both negative assertions that follow it can pass vacuously: if the executor has not yet processed the message when the sleep expires, `expect(guidedStepMock()).not.toHaveBeenCalled()` holds for the wrong reason, so the test would keep passing even if a future change let the verb reach the guided handler slightly later.

For the three verbs that survive the receive gate (`navigate`, `multistep`, `guided` — all present in KNOWN_TARGET_ACTIONS at cross-tab.types.ts:224), there IS a positive signal to synchronise on: runGuided returns false and runStepCommand posts `{ kind: &#39;step-complete&#39;, stepId, runId, ok: false }` (live-tab-executor.ts:307). Await that with `waitFor` and then assert the handler was never called; keep the sleep only for the verbs the envelope validator drops outright (`popout`, `sequence`, and the `noop` case), where no signal exists. The comment above the helper already identifies that distinction — the code just applies the weaker path to both.

🔧 Fix: extend guided verb gate to snippets and launch path
1 warning still open:

- ⚠️ `src/interactive-engine/action-handlers/guided-handler.ts:152` - The new undrivable-verb guard is placed above the `isSkippable` element-not-found branch, so a **skippable** guided step with `navigate`/`popout` changes outcome from `skipped` to `error` — a user-visible regression for already-published content, contrary to the intent&#39;s &#34;Runtime outcome deliberately PRESERVES today&#39;s observable behaviour&#34;.

Concrete trace. Guide block: `{ type: &#39;guided&#39;, steps: [ { action: &#39;navigate&#39;, reftarget: &#39;/explore&#39;, skippable: true }, { action: &#39;button&#39;, reftarget: &#39;Save&#39; } ] }`.

- Before: `runGuidedStep` fell through to `findTargetElementWithRetry(refTarget, &#39;navigate&#39;, timeout, retryInterval, /* skipRetryOnFailure */ true)`; `skipRetryOnFailure` throws on the first attempt (guided-handler.ts:413), and the catch at guided-handler.ts:178 returned `finishGuidedStep(&#39;skipped&#39;, 0)`. `interactive-guided.tsx:426` treats `skipped` like `completed`, so step 2 ran and the block could complete.
- After: the guard at line 152 returns `&#39;error&#39;` before that branch is ever reached. `interactive-guided.tsx:442` sets `failedStepIndex` and `&#34;Step 1 failed. Click \&#34;Retry\&#34; to try again.&#34;` and returns false. Retry re-enters the same guard, so the step can never pass, and unlike the `timeout` branch (interactive-guided.tsx:433) the error branch offers no Skip affordance. The reader is stuck on a guide that previously completed.

The intent does reject &#34;skipping the step&#34; as a general outcome, so this may well be the call you want — but the rejected alternative was about *silently* advancing an undriven step, whereas `skippable: true` is the author explicitly asking for that step to be skippable, and that is what the code already did. If you want to keep today&#39;s behaviour, move the guard below the `isSkippable` catch (or return `&#39;skipped&#39;` when `action.isSkippable` is true) and keep `&#39;error&#39;` for the non-skippable case, which is where the pre-change outcome genuinely was `&#39;error&#39;`. Either way the unsupported-verb `logger.warn` should stay.

Note the blast radius is narrow: only guided steps that carry `navigate`/`popout` *and* `skippable: true`. No in-repo guide or package JSON matches (I walked every tracked `*.json` with a `blocks` root), so this is CDN/backend-published content only.

🔧 Fix: honour isSkippable for undrivable guided verbs
2 infos still open:

- ℹ️ `src/interactive-engine/action-handlers/guided-handler.ts:160` - The stated intent says: &#34;Runtime outcome deliberately PRESERVES today&#39;s observable behaviour: an unsupported verb still resolves the step as &#39;error&#39;. The user explicitly rejected the alternatives of skipping the step or degrading it to a noop&#34;. The code at guided-handler.ts:160 now settles `action.isSkippable ? &#39;skipped&#39; : &#39;error&#39;`, so a skippable guided `navigate`/`popout` step resolves as `&#39;skipped&#39;`, not `&#39;error&#39;`.

This is not a defect and should not be reverted: it was explicitly authorized mid-pipeline (round 2), and it is what today&#39;s code actually did — a skippable `navigate` step hit `findTargetElementWithRetry(..., skipRetryOnFailure: true)` (guided-handler.ts:397), threw on the first attempt, and returned `finishGuidedStep(&#39;skipped&#39;)` at guided-handler.ts:178, after which interactive-guided.tsx:426 treated it like `completed` and the run continued. So the new behaviour satisfies the intent&#39;s governing principle (&#34;PRESERVES today&#39;s observable behaviour&#34;) while diverging from its literal `&#39;error&#39;` clause, which was written without the skippable case in mind. One genuinely new outcome is present and documented in the code comment: a skippable `popout` carries no `reftarget`, so it previously errored out of the `Non-noop action requires a refTarget` throw and now skips.

Asking rather than auto-resolving: the intent prose is authoritative acceptance criteria and it now contradicts the shipped behaviour. Reconcile the PR description / intent text with what merged (guided `navigate`/`popout`: `&#39;skipped&#39;` when the author marked the step skippable, `&#39;error&#39;` otherwise), so a later reader does not treat the code as drift from a recorded decision.
- ℹ️ `src/interactive-engine/action-handlers/guided-handler.ts:399` - `GuidedDomActionType` was introduced in this PR and applied to `createCompletionListener` (line 649) and `attachCompletionListener` (line 744), but four signatures in the same file still spell the identical union out by hand: `findTargetElementWithRetry` (line 399), `findTargetElement` (line 463), `highlightTarget` (line 565), and `getActionMessage` (line 631). All four take exactly `&#39;hover&#39; | &#39;button&#39; | &#39;highlight&#39; | &#39;formfill&#39;`.

Not a correctness bug — adding or removing a member of `GUIDED_DOM_ACTION_TYPES` would still fail to compile at these call sites, so the ratchet described in the intent&#39;s negative test #2 holds. It is a deduplication cleanup: replacing the four literal unions with `GuidedDomActionType` leaves one declaration of the drivable-verb set instead of five, which is the stated point of introducing the alias. (`navigation-manager.ts:32` carries the same literal union and could follow, but it is outside this change&#39;s scope.)

🔧 Fix: use GuidedDomActionType in remaining guided signatures

- ⚠️ `src/interactive-engine/action-handlers/guided-handler.ts:160` - The authoritative intent states: &#34;Runtime outcome deliberately PRESERVES today&#39;s observable behaviour: an unsupported verb still resolves the step as &#39;error&#39;. The user explicitly rejected the alternatives of skipping the step or degrading it to a noop&#34;. The shipped code at guided-handler.ts:160 settles `arbiter.settle(action.isSkippable ? &#39;skipped&#39; : &#39;error&#39;)`, so a skippable guided `navigate` or `popout` step resolves `&#39;skipped&#39;`, not `&#39;error&#39;`.

This is NOT a defect and must not be reverted — it was explicitly authorized by the user in round 2 of this run, and for `navigate` it is the faithful preservation of prior behaviour (a `navigate` refTarget is a URL path, so `findTargetElementWithRetry(..., skipRetryOnFailure: true)` threw on the first attempt and the `isSkippable` catch at guided-handler.ts:181 already returned `&#39;skipped&#39;`). A skippable `popout` is a genuine, authorized change from `&#39;error&#39;` to `&#39;skipped&#39;` (it carries no reftarget, so it previously hit the `Non-noop action ... requires a refTarget` throw at guided-handler.ts:166).

What was NOT done is the reconciliation the user directed in round 3 (&#34;Make the commit message and the PR body describe the runtime behaviour accurately instead of repeating the blanket &#39;no observable behaviour change&#39; claim&#34;). Commit 38026729&#39;s message still closes with &#34;No runtime behaviour change: an unsupported verb already resolved as a failed step via the caught throw, and still does&#34;, and the three fix-round commits (bc9a3366, 989b9cc0, 2b263125) have empty bodies, so no commit in the branch records the skippable outcome. The same message is also now numerically stale: it says &#34;the three runtime guide loaders pass `allowUnsupportedGuidedAction`&#34; while four do (`json-parser.ts:127`, `content-fetcher.ts:235`, `content-fetcher/backend-guide.ts:82`, `prepare-guide-launch.ts:108`) — `docs/design/CONCERN_DETAILS.md` already correctly says four.

Recommended: record the accurate three-line outcome table in the commit history (non-skippable undrivable verb -&gt; &#39;error&#39;, unchanged; skippable &#39;navigate&#39; -&gt; &#39;skipped&#39;, unchanged; skippable &#39;popout&#39; -&gt; &#39;skipped&#39;, CHANGED and deliberate) and fix the three/four count. Raising this as ask-user rather than resolving it because the authoritative intent text itself is the thing that is now wrong.
- ℹ️ `src/components/block-editor/utils/block-import.ts:90` - `parseAndValidateGuide` calls `validateGuideFromString(jsonString)` with no options, so the new `unsupported_guided_action` error is blocking there. Concrete reachable path: a guide already published to a CDN or backend repository carries `{ type: &#39;guided&#39;, steps: [{ action: &#39;navigate&#39;, reftarget: &#39;/explore&#39; }] }`. At runtime it now loads fine (the four loaders pass `allowUnsupportedGuidedAction: true`) but that step still fails for readers. The author opens the block editor to fix it and pastes or drops the JSON into `ImportGuideModal`; `validateGuide` returns `isValid: false` with `guide: null` (validate-guide.ts:151), so the import is refused and the guide never reaches the editor. The one surface designed to repair the guide is the one that will not open it.

This follows the established `allowDuplicateHeading` layering exactly — `docs/design/CONCERN_DETAILS.md` lists `block-import.ts` among the erroring authoring gates — so it is intent-conformant, and the error message does name the offending path and the remedy. The asymmetry worth a decision is that `duplicate_heading` is cosmetic (the guide still works), whereas `unsupported_guided_action` marks a guide that is actually broken for readers, which inverts who needs the editor most.

If that friction matters, the minimal fix at the right boundary is to pass `allowUnsupportedGuidedAction: true` in `parseAndValidateGuide` only. The editor would still refuse to ship it: `guide-lint.ts:189`, `block-export.ts:108`, and `github-pr.ts:117` all stay strict, and the lint diagnostic surfaces in the health panel (its path `blocks[N].steps[M].action` passes both `forPath` and `forPathDirect`, since `steps` is not in `NESTED_BLOCK_CONTAINERS`). Raising as ask-user because it changes an authoring-surface behaviour the author deliberately aligned with the existing contract.
</details>

<details>

✅ No issues found.
- <code>`npx jest --coverage=false src/validation/guided-action-validator.test.ts src/interactive-engine/action-handlers/guided-handler.test.ts src/validation/build-snippets.test.ts src/snippet-engine/online-snippet-resolver.test.ts src/components/docs-panel/utils/prepare-guide-launch.test.ts src/integrations/cross-tab/live-tab-executor.test.ts` — 6 suites / 137 tests pass</code>
- <code>Regression check: extracted the base commit (`git archive f618feb8`) to a temp tree, copied the two new test files in unchanged, and ran `npx jest src/validation/guided-action-validator.test.ts src/interactive-engine/action-handlers/guided-handler.test.ts` there — 11 tests fail and the validator suite cannot resolve its module; both suites (45 tests) pass on the target commit</code>
- <code>`npm run build:cli` then `node dist/cli/cli/index.js validate explore-logs.json` on a guide with a guided `navigate` + `popout` step — fails with the new per-step message, exit 1; the same command on a base-commit build of the CLI reports all guides valid, exit 0</code>
- <code>`node dist/cli/cli/index.js validate explore-logs-fixed.json` after moving the verbs out of the guided block — passes, exit 0</code>
- <code>`node dist/cli/cli/index.js build-snippets ./snippets` with a guided `navigate` nested in a `section` inside a snippet body — errors and writes no `index.json`; succeeds and writes the catalog after the verb is removed</code>
- <code>`node` against the compiled `dist/cli/validation/validate-guide.js`, calling `validateGuide` on the same guide with `{}` (authoring) vs `{ allowDuplicateHeading: true, allowUnsupportedGuidedAction: true }` (what the four runtime loaders pass) — errors with code `unsupported_guided_action` vs valid-with-advisories and all three blocks retained</code>
- <code>`npm run validate` (10 bundled guides) and `npm run validate:packages` (10 bundled packages) — all valid, so the new gate breaks no shipped content</code>
- <code>Manual browser verification: `npm run build:all` + `docker compose up -d --build`, then Playwright/Chrome against localhost:3000 — block editor `More actions -&gt; Import -&gt; Paste JSON` shows the new message for the guided `navigate`/`popout` guide (screenshot)</code>
- <code>Manual browser verification: injected the rejected guide as the reader-facing panel guide (`grafana-pathfinder-app-e2e-test-guide` + `pathfinder-auto-open-docs`, the e2e runner&#39;s own delivery path) — the guide loads and all blocks render (screenshot)</code>
- <code>Manual browser verification: clicked `Start guided interaction` on a guided block with skippable `navigate` + `popout` — the run skips both, reaches `STEP 3 OF 3`, and logs `logger.warn` (screenshot + console capture)</code>
- <code>Manual browser verification: clicked `Start guided interaction` on a non-skippable guided `navigate` — panel shows `Step 1 didn&#39;t complete` / `Try again`, i.e. the step resolves as failed (screenshot + console capture)</code>
- <code>Cleanup: `docker compose down -v --rmi local`, removed the `dist/` build output and temp build tree; `git status --porcelain` is clean</code>
</details>

<details>

- ℹ️ `docs/developer/interactive-examples/json-guide-format.md:1305` - The new &#34;Actions a guided step accepts&#34; table hand-copies the membership of GUIDED_ACTION_TYPES (src/types/interactive-actions.types.ts). This repository already has the drift-guard pattern for exactly this file — src/validation/json-guide-format-docs.test.ts reads block types out of JsonBlockSchema at runtime and fails until the reference documents a new variant — but it only covers the block-type summary, so adding or removing a guided verb leaves this table silently wrong. Closing it means extending that test to read GUIDED_ACTION_TYPES and assert a row per verb, which is a test change and outside this documentation pass. Reported as a follow-up rather than fixed.
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>


Spidy88 (commented):
Two open validation gaps I want on record from reviewing this branch:

**1. `src/components/block-editor/utils/block-import.ts:90`** — `parseAndValidateGuide` (used by the block editor's Import → Paste JSON) calls `validateGuideFromString` with no options, so it runs in strict/authoring mode. That's the right call for `duplicate_heading` (cosmetic), but `unsupported_guided_action` means the guide is *actually broken for readers*. Net effect: an author trying to open an already-published guide with a bad guided step, specifically to fix it, gets refused entirely (`isValid: false, guide: null`) — the one repair path is blocked. Worth either passing `allowUnsupportedGuidedAction: true` at this call site (the editor still can't re-publish it unfixed — `guide-lint.ts`, `block-export.ts`, and `github-pr.ts` all stay strict), or making the tradeoff explicit and noting the asymmetry in the `guide-schema-and-contracts` anchor in `CONCERN_DETAILS.md`.

**2. `src/components/block-editor/forms/StepEditor.tsx:765` and `:1075`** — `ACTION_OPTIONS` is still built unfiltered from all of `INTERACTIVE_ACTIONS` and used for both the guided-step add and edit pickers, so the editor still *offers* `navigate`/`popout` when editing a guided step, even though `guide-lint.ts` will reject that choice afterward. Not a bypass, but an offer-then-refuse authoring UX gap on the exact surface this PR is tightening. `StepEditor` already receives `isGuided` — filtering `ACTION_OPTIONS` against `GUIDED_ACTION_TYPES` when `isGuided` is true would keep the invalid choice from ever being offered.

loglapa:
Coordination with #1914 (PR 2 of #1525): the branches merge without Git conflicts, but removing `sequence` from the canonical vocabulary makes the new `UNDRIVABLE` table in `guided-handler.test.ts` fail TypeScript. Whichever PR lands second should remove `sequence` from that table; the retired value is now rejected at DOM ingress, covered in #1914. I verified the combined tree with this one-line adjustment: TypeScript and 182 focused tests pass. No changes to this branch or the cross-tab receive gate.

moxious:
@Spidy88 both gaps are now fixed on this branch (c5980b1f), so this PR **Fixes #1908** — the issue that had deferred exactly these two.

**1. `block-import.ts:90`** — took your first option and passed `allowUnsupportedGuidedAction: true`. Two things made it cheaper than it looked:

- `ImportGuideModal:478` already renders warnings, and the downgraded message already carries the path plus the remedy — `blocks[0].steps[1].action: guided steps cannot use "navigate" … Use a separate interactive block … or move this step into a multistep block`. So the author sees which step and what to change with no new UI.
- The same function backs JSON mode (`useJsonModeHandlers`), which is the other repair path, so one line fixed both.

Nothing downstream is relaxed: `guide-lint.ts:189` still raises it live on the step, and `block-export.ts` / `github-pr.ts` still refuse to ship it — so a tolerated guide can't leave the editor unfixed. `allowDuplicateHeading` is deliberately still *not* passed there, per your read that the cosmetic one is worth refusing at the door. (#1811 is that same trap, unresolved, for the other flag.)

**2. `StepEditor.tsx`** — `ACTION_OPTIONS` now narrows to `GUIDED_ACTION_TYPES` when `isGuided`, at both `:765` and `:1075`, derived from the tuple so a new guided verb needs no second edit. One wrinkle fix 1 creates: a pre-gate step's own verb is retained in the list, otherwise the step the author came to fix shows no action at all. `GuidedBlockForm:134` is the only `isGuided={true}` caller, so nothing else changed.

No code needed for #1908's record-mode criterion — `action-recorder.hook.ts` only emits `hover` / `formfill` / `highlight`, all guided-legal.

**Contract anchor** — `guide-schema-and-contracts` listed `block-import.ts` among the unconditionally strict authoring gates, so it now records the divergence: the split at that call site is repair-versus-ship rather than author-versus-runtime, and why.

**Tests** — `block-import.test.ts` pins that the guide imports with the step named in a warning and that an otherwise-malformed guide is still refused; `StepEditor.guided-actions.test.tsx` pins the guided option set against `GUIDED_ACTION_TYPES`, the pre-gate retention, and that a multistep parent keeps the full set. Also corrected `docs/sources/block-editor/_index.md`, which this PR had previously stated the old JSON-mode refusal.

Merged latest `main` and re-validated on the merged tree: typecheck clean, `npx eslint` clean on the changed files, 2348 tests green across `src/validation`, `src/components/block-editor`, `src/docs-retrieval`, `src/snippet-engine`.

@loglapa on your #1914 coordination note — still accurate, and unchanged by this commit: whichever of us lands second drops `sequence` from the `UNDRIVABLE` table in `guided-handler.test.ts`.

Spidy88 (approved):

PR #1914, opened 2026-09-16
refactor: remove the retired sequence action path

### LINKED ISSUE #1525: Guided blocks accept navigate/popout at validation, then throw at runtime

loglapa:
## Summary

Remove the retired DOM-scraping `sequence` executor and reject that verb at DOM ingress. This is PR 2 of the split agreed in #1525; section runs, multistep and guided actions retain their current execution paths.

## Why

JSON authoring and the cross-tab receive gate already exclude `sequence`, and interactive HTML parsing was deliberately retired. Repository-wide caller checks found no remaining production ingress to SequenceManager. Keeping the verb after deleting its executor would leave the canonical vocabulary accepting an action it cannot execute.

This follows the existing canonical-vocabulary and raw-then-validated DOM contract: remove the retired member, without introducing another action list or widening the wire gate.

## What to look at

- Delete SequenceManager, the hook-only dispatcher, sequence result plumbing, and their dedicated tests/mocks. Keep both the Faro and caller-facing `completionSuppressed` failure checks.
- Remove `sequence` from the canonical tuple and affected local types/switches. The cross-tab receive gate is unchanged.
- The DOM rejection test fails if `sequence` is restored to the tuple; the retained decoder, hook, section-runner, fullscreen and cross-tab tests pass.
- Preserve the hook options signature for current callers; `containerRef` no longer scopes a retired interpreter. Record stale documentation references in the maintenance backlog.

## Validation and coordination

All nine local checks passed again after merging current main (567 Jest suites, 9,353 tests). The production build passed before this test-only compatibility update. The decoder negative test fails when the retired tuple member is restored.

#1907 has merged and closed #1525’s guided-validation bug. This PR remains the separately agreed sequence cleanup (PR 2). Current main is merged here, including the agreed removal of `sequence` from guided-handler’s `UNDRIVABLE` table: it is now rejected at DOM ingress. The merge-order compatibility adjustment is complete.

## How to verify

In Grafana, exercise a section run, an automated multistep block, and a guided block. Verify that a missing fullscreen-handoff target does not persist step completion. These runtime walkthroughs remain manual review checks.

## Breaking changes

The internal `interactiveSequence`/SequenceManager exports and `sequence` action value are removed. Raw DOM input using that retired value now fails decoding. Supported JSON guide shapes and wire commands are unchanged.

## Reversibility

No stored-data migration. This assumes interactive HTML ingress stays retired; if it returns, the old interpreter is recoverable from Git history.

## Out of scope

Guided-action validation landed in #1907. Shared telemetry event vocabulary, legacy style/constants cleanup, and the #1763 architectural boundaries remain separate.

Refs #1525. Follows #1757.

🤖 Prepared with OpenAI Codex.


loglapa:
Updated onto current main after #1907 merged, and applied the agreed one-line adjustment: `sequence` is removed from guided-handler’s `UNDRIVABLE` table because the DOM decoder now rejects it. All nine local checks pass again (567 Jest suites / 9,353 tests).

#1525 is closed by the guided-validation fix; this PR remains its separately agreed sequence-removal follow-up. The description now reflects that and the completed merge-order adjustment.

moxious (approved):
LGTM, thank you

Verified independently during review: JsonInteractiveActionSchema never admitted the verb and ContentJsonSchema.safeParse is a hard runtime load gate, so no published guide can carry it; HTML interactive ingress is retired and pinned by post-removal assertions; KNOWN_TARGET_ACTIONS is byte-unchanged so the cross-tab wire gate neither widened nor narrowed; and the completionSuppressed full-screen-handoff guarantee survives intact on both the Faro span and the caller-facing return. The contract-evolution scan came back follows_contract against the interactive-engine anchor, with no anchor update required.

PR #1915, opened 2026-09-16
test(validation): drift-check the guided-action table against GUIDED_ACTION_TYPES

### LINKED ISSUE #1910: Guided-action table in json-guide-format.md is not drift-checked against GUIDED_ACTION_TYPES
## Summary

The "Actions a guided step accepts" table added to `docs/developer/interactive-examples/json-guide-format.md` hand-copies the membership of `GUIDED_ACTION_TYPES` (`src/types/interactive-actions.types.ts`). Adding or removing a guided verb leaves that table silently wrong.

This repository already has the drift-guard pattern for this exact file. `src/validation/json-guide-format-docs.test.ts` reads block types out of `JsonBlockSchema` at runtime and fails until the reference documents a new variant — but it only covers the block-type summary table, not the new guided-action table.

## User impact / flow change

Guide authors and authoring agents read `json-guide-format.md` as the reference for what a guided step may contain. If the guided vocabulary changes and the table does not, an author is told a supported verb is forbidden (so they avoid it needlessly) or that a forbidden verb is supported (so they hit the authoring gate they were trying to satisfy). Agents are the more sensitive consumer, since they treat the reference as authoritative and cannot notice the inconsistency the way a person might.

No runtime behaviour is affected — the gate in `validate-guide.ts` is the enforcement, and it derives from the tuple correctly. This is purely about the documentation staying true.

## Acceptance criteria

- [ ] `src/validation/json-guide-format-docs.test.ts` reads `GUIDED_ACTION_TYPES` at runtime and asserts the guided-action table has a row per member, marked as accepted for `guided`.
- [ ] The test also asserts the verbs absent from `GUIDED_ACTION_TYPES` but present in `INTERACTIVE_ACTION_TYPES` appear as rejected for `guided`, so the table cannot go stale in either direction.
- [ ] Adding a member to `GUIDED_ACTION_TYPES` fails the test until the table documents it, and the failure message names the missing verb.
- [ ] Negative-tested and restored: add a verb to the tuple, confirm the test fails naming it, revert.

## Scope and non-goals

In scope: extending the existing doc-drift test to cover the guided-action table.

Non-goals: the gate itself, the tuples, and the runtime behaviour — all of which are correct. Do not replace the table with generated text; the point is a guard that fails loudly, not codegen into a hand-written reference. Also not in scope: auditing every other hand-written table in this file, which would be its own sweep.

## Where it lives / grounding

- `docs/developer/interactive-examples/json-guide-format.md` — the "Actions a guided step accepts" table, and the `action` row in the step-structure table which also names the verbs.
- `src/validation/json-guide-format-docs.test.ts` — the existing drift guard to extend; it already resolves the doc path and parses the markdown tables.
- `src/types/interactive-actions.types.ts` — `GUIDED_ACTION_TYPES`, the population to derive from.
- `src/types/interactive.types.ts` — `INTERACTIVE_ACTION_TYPES`, for the rejected-verb half.

## Provenance

Raised by the `no-mistakes` document step on the PR that added the table (PR 3 of the #1525 split), classified informational. Deferred rather than fixed in that PR because it is a test change in a different file and the PR was already four review rounds deep; the house rule is one ratchet per PR plus only the fixes it forces.

This is a direct instance of the general rule that a tripwire's population should be derived structurally rather than hand-listed.

devtejasx:
## Summary

The doc drift test now also covers the "Actions a guided step accepts" table. It reads `GUIDED_ACTION_TYPES` at runtime and fails if a step action has no row, or if its `guided` cell doesn't match the tuple, naming the verb in the message.

Stacked on #1907, which adds the tuple and the table, so this targets that branch. GitHub will retarget it to `main` when #1907 merges.

## What to look at

- [`json-guide-format-docs.test.ts`](https://github.com/devtejasx/grafana-pathfinder-app/blob/1651c93f654f64ced746ccaf8f94f8aae83a350f/src/validation/json-guide-format-docs.test.ts): `summaryTableSection` became `tableSection(markdown, heading)` so both tables share the same section cutter. The new `describe.each` iterates `JsonInteractiveActionSchema.options`, the step `action` enum, rather than `INTERACTIVE_ACTION_TYPES`. The latter also contains `sequence`, `multistep` and `guided`, which are block types, not step verbs, and never appear in the table. Each verb is checked in both directions: ✅ when `GUIDED_ACTION_TYPES` contains it, ❌ otherwise.
- The `guided` column is found by header name, so reordering columns doesn't break the check.

## How to verify

Negative-tested and reverted, each producing exactly one failure naming the verb:

1. Add `'navigate'` to `GUIDED_ACTION_TYPES`: `navigate` fails.
2. Flip the `popout` row's `guided` cell to ✅: `popout` fails.
3. Delete the `hover` row: `hover` fails.

## Breaking changes

None, test only.

Fixes #1910


moxious:
Thanks for extending the doc-drift test to cover the guided-action table — this closes #1910 cleanly.

Your commits are untouched on your fork. Could you reopen this PR retargeted at `main` (or open a fresh PR from the same branch, `fix/guided-action-doc-drift-1910`, against `main`)? Once that's done we'll pick the review back up — the fix itself was verified as complete and ready to go.

PR #1916, opened 2026-09-16
fix(block-editor): let import open a guide to repair its guided verbs, and narrow the guided picker

### LINKED ISSUE #1908: Block editor refuses the pre-gate guided guide an author must open to repair it, and offers verbs its own lint rejects
## Summary

The guided-verb authoring gate added for #1525 (PR 3) leaves the block editor's authoring surface in two incoherent states. Both are consequences of the gate rather than defects in it, and both were deliberately deferred out of that PR to keep it one ratchet.

**1. An author cannot open a published-but-broken guide in order to repair it.** `parseAndValidateGuide` in `src/components/block-editor/utils/block-import.ts` calls `validateGuideFromString(jsonString)` with no options, so the new `unsupported_guided_action` error is blocking there. A guide already published to a CDN or backend repository can carry `{ type: 'guided', steps: [{ action: 'navigate', reftarget: '/explore' }] }`. At runtime it still loads — the four runtime guide loaders pass `allowUnsupportedGuidedAction: true` — but that one step fails for readers. The author opens the block editor to fix it, pastes or drops the JSON into `ImportGuideModal`, and the import is refused with `guide: null`. The repair tool rejects exactly the content it exists to repair.

This is contract-conformant: `CONCERN_DETAILS.md:69` lists `block-import.ts` among the authoring gates, and authoring gates are supposed to error. The catch-22 is nonetheless real.

**2. The step action picker offers verbs the editor's own lint rejects.** `ACTION_OPTIONS` in `src/components/block-editor/forms/StepEditor.tsx` is a module-level constant built from all of `INTERACTIVE_ACTIONS` and is used unfiltered at line 765 (editing a step) and line 1075 (adding one). The component already receives an `isGuided` prop (`StepEditorProps:266`, defaulted at `:337`) and branches on it about a dozen times, so the narrowing information is in hand. An author picks `navigate` from the dropdown for a guided block and then gets a lint error for the choice the dropdown just offered. Record mode compounds it: `StepEditor.tsx:575` notes recorded steps go straight into a `multistep`/`guided` `steps` array.

## User impact / flow change

For (1), an author who wants to fix a broken guided step is blocked at the door: the only in-product editor for the guide refuses to load it, with a schema error rather than guidance. The workaround is hand-editing JSON outside the editor, which is precisely what the block editor exists to avoid. This is the more serious of the two because it has no in-product path forward.

For (2), the author learns the rule by failing rather than by the UI not offering the option. No broken content can ship — `guide-lint.ts` blocks it — so this one is a papercut.

After the fix, an author can load a pre-gate guide, see the offending step flagged, and correct it; and the picker never offers a verb that will be rejected.

## Acceptance criteria

- [ ] `block-import.ts` can load a guide whose only fault is an unsupported guided verb, surfacing it as a visible, actionable diagnostic on the offending step rather than refusing the whole import.
- [ ] The import path still refuses genuinely malformed guides — this is not a blanket downgrade of `block-import.ts` to tolerant. Only the `unsupported_guided_action` code becomes non-blocking there, and the reason is recorded.
- [ ] `ImportGuideModal` shows the author which step is at fault and what to change it to.
- [ ] Saving or exporting from the editor still refuses the unsupported verb, so the gate is not defeated — the author must fix it before it can ship.
- [ ] With a `guided` parent, the step action picker offers only the members of `GUIDED_ACTION_TYPES`, derived from that tuple rather than a second hand-written list. With a `multistep` parent the full authorable set remains, including `navigate` and `popout`.
- [ ] Both picker use sites are covered (`:765` edit, `:1075` add), and record mode against a guided block cannot produce a step carrying `navigate` or `popout`.
- [ ] A test asserts the guided picker's option set equals `GUIDED_ACTION_TYPES`, so a verb added to the tuple appears without a second edit.
- [ ] The `guide-schema-and-contracts` anchor in `docs/design/CONCERN_DETAILS.md` records whichever way `block-import.ts` is resolved, since that anchor currently lists it as an unconditional authoring gate.

## Scope and non-goals

In scope: the block editor's handling of pre-gate guided content — the import path and the action picker.

Non-goals: the gate itself, which works and is contracted. `validateGuidedActions` errors for authoring gates and downgrades to an advisory for the four runtime guide loaders; do not weaken, duplicate, or bypass it to accommodate the editor. In particular, do not make the runtime loaders stricter or the authoring gates uniformly tolerant — the asymmetry is the contract.

Also a non-goal: the cross-tab `noop` receive-gate gap, which is a separate wire-contract issue filed alongside this one.

## Where it lives / grounding

- `src/components/block-editor/utils/block-import.ts` — `parseAndValidateGuide`, the unconditional `validateGuideFromString(jsonString)` call.
- `src/components/block-editor/ImportGuideModal.tsx` — where the refusal surfaces to the author (`:467` error display, `:513` disabled confirm).
- `src/components/block-editor/forms/StepEditor.tsx` — `ACTION_OPTIONS` (module level), use sites `:765` and `:1075`, `isGuided` at `StepEditorProps:266` / `:337`, record-mode note at `:575`.
- `src/types/interactive-actions.types.ts` — `GUIDED_ACTION_TYPES`, the tuple both fixes should derive from.
- `src/validation/guided-action-validator.ts` — the gate, and the `unsupported_guided_action` error code.
- `src/validation/validate-guide.ts` — `allowUnsupportedGuidedAction`, the existing escape hatch and the layering to match.
- `docs/design/CONCERN_DETAILS.md:69` — the `guide-schema-and-contracts` anchor listing authoring gates and runtime loaders.

## Provenance

Both found by the `no-mistakes` review pass on the PR implementing PR 3 of the #1525 split (the picker in the first review round, the import catch-22 in a later one), and deliberately deferred out of it: both are block-editor UI work in files that PR does not otherwise touch, and this repo prefers one ratchet per PR plus only the fixes it forces.

devtejasx:
## Summary

The block editor can now open a published guide whose only problem is a guided step using `navigate` or `popout`, so the author can fix it in the editor. The guide still can't be saved until those steps are fixed. The step action picker and record mode now only offer a guided block the verbs in `GUIDED_ACTION_TYPES`.

Stacked on #1907, which adds the gate, so this targets that branch.

## What to look at

- **Import (`block-import.ts`, `ImportGuideModal.tsx`)**: `parseAndValidateGuide` / `importGuideFromFile` take an `allowUnsupportedGuidedAction` option. With it, the strict result is kept unless every error has code `unsupported_guided_action`. Only then is the guide re-validated tolerantly and returned, with the offending steps (line numbers included) in `unsupportedGuidedActions`. The modal lists them under "Fix these steps before saving". Any other error, such as a missing title or a duplicate heading, still refuses the import. JSON mode calls `parseAndValidateGuide` without the option, so it stays strict.
- **Save (`useBackendSaveFlow.ts`)**: `orchestrateSave` refuses when `validateGuidedActions(guide)` reports anything, with the validator's message. The GitHub PR export already refuses through `validateGuide`. Copy and download are raw JSON dumps that have never validated, and I left them alone.
- **Picker and record mode (`StepEditor.tsx`)**: `stepActionOptions(isGuided)` filters `ACTION_OPTIONS` with `isGuidedActionType`, so the list comes from the tuple. Both use sites (edit and add) use it. `handleStepRecorded` drops a recorded step whose verb a guided block can't drive.
- **Contract anchor**: `guide-schema-and-contracts` in `CONCERN_DETAILS.md` records the import exception and why it doesn't defeat the gate.

## Why

Before this, `block-import.ts` refused the whole guide, so a guide already published with a `navigate` step in a guided block could only be fixed by hand-editing JSON. The asymmetry between runtime loaders (tolerant) and authoring gates (strict) is kept. The import is the one gate that admits the verb, and only for repair: saving is where it is enforced.

## How to verify

1. Import (file or paste) a guide with `{ "type": "guided", "content": "x", "steps": [{ "action": "navigate", "reftarget": "/explore" }] }`: the modal enables Import and names the step and line.
2. Save as draft without changing it: "Cannot save guide" with the guided-verb message.
3. Change the step's action: the picker offers highlight, button, form fill, hover and info only. Save now works.
4. Add the same guide with a missing `title`: import is still refused.

Tests added: `block-import.test.ts` (strict default, repair path, and two cases with an extra error), `useBackendSaveFlow.test.ts` (save refused), and `StepEditor.guided.test.tsx` (picker equals `GUIDED_ACTION_TYPES`, multistep keeps `navigate`/`popout`, record mode drops `navigate` only for guided). Five of those fail with the source changes reverted.

## Breaking changes

None. `parseAndValidateGuide` and `importGuideFromFile` gained an optional parameter.

Fixes #1908


moxious:
Thanks for the block-editor import/guided-picker fix.

Your commits are untouched on your fork. Could you reopen this PR retargeted at `main` (or open a fresh PR from the same branch, `fix/guided-editor-pregate-1908`, against `main`)? One thing to check while you're there: our last check against the old base showed this branch had merge conflicts — worth confirming it merges cleanly against current `main` before we pick review back up.

PR #1917, opened 2026-09-16
fix(cross-tab): accept noop at the step-command receive gate

### LINKED ISSUE #1909: A guided block containing a noop step cannot be relayed cross-tab: the message is dropped at the receive gate with no reply
## Summary

`noop` is a verb the guided handler drives — it is in `GUIDED_ACTION_TYPES` and `GuidedHandler.executeNoopStep` implements it — but it is absent from the cross-tab receive gate's `KNOWN_TARGET_ACTIONS` (`src/types/cross-tab.types.ts:224`, which lists only `button`, `highlight`, `formfill`, `navigate`, `hover`, `guided`, `multistep`).

Substeps are validated against that same set (`cross-tab.types.ts:280`, the `internalActions.every(...)` check added for T1 / #1068). So a `guided` block containing a `noop` step fails envelope validation as a whole message: the live tab drops it silently and posts no reply at all — not even a `step-complete` with `ok: false`.

The result is that a guided block with an informational `noop` step cannot be relayed cross-tab at all, while the same block works normally in a single tab.

## User impact / flow change

A learner running a two-tab (controller + live tab) session on a guide whose guided block contains a `noop` step sees the step command produce nothing. Because the message is dropped at the envelope validator rather than refused by a handler, there is no `step-complete` to report failure, so the controller gets no outcome for that step at all — it is a silent stall rather than a visible error.

`noop` steps are informational ("read this, then continue"), and the CLI actively steers authors away from using them as filler (`src/cli/utils/warnings.ts:42`), so the affected population is guided blocks that use a noop deliberately. Single-tab runs are unaffected.

## Acceptance criteria

- [ ] A guided block containing a `noop` substep relays successfully over cross-tab and the learner can complete it, or it is explicitly and visibly refused with a `step-complete` carrying `ok: false` — not silently dropped.
- [ ] Whichever behaviour is chosen, the controller always receives an outcome for the step; no step command produces zero reply.
- [ ] The existing test `drops a guided noop substep at the receive gate, so it never reaches the handler` (`src/integrations/cross-tab/live-tab-executor.test.ts`) is updated to pin the new intended behaviour rather than the current gap.
- [ ] The decision is recorded on the `cross-tab` contract anchor, since `KNOWN_TARGET_ACTIONS` is a wire contract.

## Scope and non-goals

The judgment call is whether `noop` belongs in `KNOWN_TARGET_ACTIONS`. That set is a **wire receive-gate**, not a local union: widening it changes what one tab accepts from another. Decision §5 on #1525 deliberately keeps it hand-written for exactly this reason, and notes that `cross-tab.types.ts:33-36` and `:188-192` carry a live contradiction about whether cross-version compatibility is a goal. **Settle that contradiction before widening the set.**

Non-goals: deriving `KNOWN_TARGET_ACTIONS` or `GUIDED_VERBS` from any action union. #1525 §5 rules that out, and #1531's lesson is that a projection from the wrong parent union is worse than the hand-maintained list it replaces. If `noop` is added, add it as a deliberate literal.

Also a non-goal: the guided authoring gate from the #1525 PR 3 work. That gate is about which verbs an author may put in a guided block; this issue is about which verbs survive the cross-tab wire. They are independent — `noop` is authorable and correct, it just cannot be relayed.

## Where it lives / grounding

- `src/types/cross-tab.types.ts:224` — `KNOWN_TARGET_ACTIONS`, the receive gate that omits `noop`.
- `src/types/cross-tab.types.ts:280` — the substep validation that applies the same set to `internalActions`.
- `src/types/cross-tab.types.ts:33-36` and `:188-192` — the unresolved cross-version-compatibility contradiction that gates any change here.
- `src/integrations/cross-tab/live-tab-executor.ts:44` — `GUIDED_VERBS`, which does include `noop`, so the two lists disagree.
- `src/types/interactive-actions.types.ts` — `GUIDED_ACTION_TYPES`, which includes `noop`.
- `src/interactive-engine/action-handlers/guided-handler.ts` — `executeNoopStep`, the handler that implements the verb.

## Provenance

Found while writing behavioural tests for the cross-tab guided verb gate in the PR implementing PR 3 of the #1525 split. Pinned there as current behaviour (with a comment explaining why) rather than fixed, because changing it means changing a wire contract, which #1525 §5 puts out of that PR's scope.

devtejasx:
## Summary

A guided block with a `noop` step now relays over the two-tab controller: the live tab runs it and posts `step-complete`. Before, `noop` was missing from `KNOWN_TARGET_ACTIONS`, so the whole `step-command` failed envelope validation and was dropped with no reply, and the controller stalled.

Stacked on #1907, whose pinned test this updates, so it targets that branch.

## What to look at

- **Receive gate (`cross-tab.types.ts`)**: `noop` added to `KNOWN_TARGET_ACTIONS` as a literal. It is not derived from any union, per #1525 §5. This widens a wire gate, so **it needs the `cross-tab-controller` security review**. The case for it being safe: `runAction` has an empty `case 'noop'`, and `GuidedHandler.executeNoopStep` only waits, so it adds no DOM-mutation or navigation reachability. Messages still need the same signature and pairing checks.
- **The contradiction the issue asked to settle first**: the `RemoteRequirementError` comment said the two tabs "may run different plugin versions", while the same-build note in the same file (and `CROSS_TAB_CONTROLLER.md`) says cross-version compatibility is not a goal. I kept the same-build assumption, since it is the documented one, and reworded the first comment so the restatement is about accidental drift. If cross-version *is* meant to be a goal, this needs a different answer: an older live tab would still drop `noop`.
- **Contract anchor**: new `cross-tab-controller` row in `CONCERN_DETAILS.md` (#1068 → #1909), recording the same-build assumption, the literal receive gate, and why `noop` is in it.
- **Test**: `drops a guided noop substep at the receive gate…` becomes `relays a guided noop substep to the guided handler and reports the step complete`. It fails with the gate change reverted. The "known limitation" bullet in `CROSS_TAB_CONTROLLER.md` is removed.

## How to verify

With `enableTwoTabController` on, pair a controller and a live tab and run a guide whose guided block has a `noop` step followed by a `button` step. The noop step completes in the controller and the block continues to the button step.

## Breaking changes

None. A live tab without this change keeps dropping `noop`, which is the same-build limitation already documented.

Fixes #1909


moxious:
Thanks for the cross-tab `noop` receive-gate fix — this closes #1909, including settling the same-build-vs-cross-version contradiction the issue flagged as a precondition.

Your commits are untouched on your fork. Could you reopen this PR retargeted at `main` (or open a fresh PR from the same branch, `fix/cross-tab-guided-noop-1909`, against `main`)? Once that's done we'll pick the review back up — CI was green and the fix was verified as complete. One small note carried over from review: the PR description's "adds no DOM-mutation surface" framing is slightly stronger than the actual behavior (the noop widening surfaces an existing, human-gated informational prompt rather than a true no-op) — worth a small wording tweak, not a blocker.

PR #1920, opened 2026-09-16
fix(completion): record one identity for a bundled guide regardless of launch shape

tomglenn:
## What

A bundled guide has two live launch shapes — `bundled:<id>` (My Learning) and `bundled:<id>/content.json` (the package resolver). The journey percentage writer derived the guide id with a bare `journeyBaseUrl.replace('bundled:', '')`, so a package-path launch recorded `<id>/content.json`, while the reset path stripped the suffix. Writer and reader disagreed, so a completion could be stored under one id and looked up under another.

## Changes

- Add `normalizeGuideId` in `completion-records/completion-identity.ts` (with a named `PACKAGE_CONTENT_SUFFIX`, mirroring `path-member-join.ts`) as the **single** owner of the `/content.json` suffix rule.
- Route both the journey writer (`docs-retrieval/learning-journey-helpers.ts`) and the reset path (`docs-panel/hooks/resetGuideProgress.ts`) through it, removing their separate inline strips. There is now exactly one place that knows the suffix.
- Flip the `completion-tracking.spec.ts` bundled-package-path case from `test.fail()` to a passing test and remove the now-stale divergence comment.
- Unit coverage: package-path and bare launches resolve to the same `(guideSource, guideId)`, and a record stored under the old suffixed id is still found.

## What to look at — the migration guarantee

The durable write shipped in **2.17.0** and records already exist under the suffixed id on Cloud. `invalidateEmittedCompletion` now clears **both** the normalized and the legacy suffixed key. There is no route that deletes a durable record, so this read-both behaviour must stay until those records age out and someone deliberately retires it — reading only the clean id would orphan real completions. This is the part most worth a careful review.

## Scope

The whole-journey `markMilestoneDone` source-axis divergence is a separate defect and is left untouched.

## Testing

`typecheck`, `lint`, and the `completion-identity` / `completion-recorder` unit suites pass on Node 24.21. The derivation was verified by temporarily disabling the strip and confirming the identity tests go red. The `completion-tracking` Playwright e2e requires a live stack and was not executed in this environment.


src/docs-retrieval/learning-journey-helpers.ts
```diff
@@ -510,7 +521,7 @@ function persistJourneyCompletionPercentage(journeyBaseUrl: string, percentage:
 
   // Update learning paths progress when a bundled guide reaches 100%
   if (percentage >= 100 && journeyBaseUrl.startsWith('bundled:')) {
-    const guideId = journeyBaseUrl.replace('bundled:', '');
+    const guideId = guideIdFromBundledJourneyBase(journeyBaseUrl);
```
Jayclifford345, line 524: [P1] Preserve the legacy guard when recording normalized identities

For a user with a persisted suffixed fallback identity, the normalized writer bypasses the existing exactly-once guard. Seeding guide:bundled:intro/content.json and recording intro emits a new completion without reset. record() checks only the canonical key; the new dual-key clearing runs only during reset. Manifest-backed identities that were already canonical are unaffected.

Recognize or migrate the legacy guard before recording the canonical identity, and test upgrade/reload followed by explicit reset and recompletion.

src/components/docs-panel/hooks/resetGuideProgress.ts
```diff
@@ -35,7 +36,9 @@ function fallbackGuideIdFromContentKey(contentKey: string): string {
   if (/^https?:\/\//.test(contentKey)) {
     return getMilestoneSlug(contentKey) || contentKey;
   }
-  return contentKey.replace(/^(bundled|backend-guide):/, '').replace(/\/content\.json$/, '');
+  // Suffix stripping goes through the single shared derivation so the reset path
+  // and the journey writer can never disagree about a bundled guide's identity.
+  return normalizeGuideId(contentKey.replace(/^(bundled|backend-guide):/, ''));
```
Jayclifford345, line 41: [P2] Repair the reset tests affected by the new helper and keys

CI has five failures in two suites: resetGuideProgress.test.ts omits normalizeGuideId from its explicit module mock (four TypeErrors), and learning-paths.hook.test.ts expects only the old invalidated-key set. The four reset failures also reproduce locally; the identity and recorder suites pass.

Include the real helper in the mock, update the dual-key assertions, and rerun both affected suites.

Jayclifford345 (changes requested):
Blockers:

1. [blocking] **legacy-completion-guard — Preserve the legacy guard when recording normalized identities** (high · completion-records)
   For a user with a persisted suffixed fallback identity, the normalized writer bypasses the existing exactly-once guard. Seeding guide:bundled:intro/content.json and recording intro emits a new completion without reset. record() checks only the canonical key; the new dual-key clearing runs only during reset. Manifest-backed identities that were already canonical are unaffected.
   Required: Recognize or migrate the legacy guard before recording the canonical identity, and test upgrade/reload followed by explicit reset and recompletion.

2. [blocking] **completion-reset-tests — Repair the reset tests affected by the new helper and keys** (medium · testing-and-verification)
   CI has five failures in two suites: resetGuideProgress.test.ts omits normalizeGuideId from its explicit module mock (four TypeErrors), and learning-paths.hook.test.ts expects only the old invalidated-key set. The four reset failures also reproduce locally; the identity and recorder suites pass.
   Required: Include the real helper in the mock, update the dual-key assertions, and rerun both affected suites.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1920
Summary: record one identity for a bundled guide regardless of launch shape
Verdict: Request Changes

tomglenn:
@Jayclifford345 thanks for the review — both blockers addressed in `c2acd0fd`.

**1. Preserve the legacy guard when recording normalized identities (P1).** The exactly-once guard now honours the migration on the *write* path, not just on reset. `record()` writes the canonical (normalized) key but **reads every spelling** via a new shared `bundledGuideIdReadVariants` — the normalized id plus the legacy `/content.json`-suffixed id 2.17.0 persisted. So a package-path completion already stored under the suffixed id is seen, and a reload-to-100% no longer mints a second durable record. `invalidateEmittedCompletion` uses the same helper, keeping read-both in one place. Added recorder coverage for exactly the upgrade/reload → reset → recompletion sequence you called out.

**2. Repair the reset tests (P2).** `resetGuideProgress.test.ts` now includes the real `normalizeGuideId` in its `completion-records` mock (fixes the four TypeErrors), and `learning-paths.hook.test.ts` asserts the dual-key invalidation set.

`npm run test:ci` (568 suites, 9424 passing), `typecheck` and `lint` are all clean on the pinned Node. Ready for another look.

Jayclifford345 (approved):
LGTM

PR #1925, opened 2026-09-16
refactor(completion): consolidate milestone progress onto interactiveCompletionStorage

Spidy88:
## Summary
- Consolidates the two local "how far along is this learner" mechanisms (mechanism 1 and 2 from the path-tracks-audit-x1 investigation) into one: the cover page and in-guide milestone toolbar now write and read milestone completion through the guide-ID-keyed `interactiveCompletionStorage`/`calculatePathRollup` model instead of the separate URL-keyed `milestoneCompletionStorage`.
- `markMilestoneDone` now writes directly into `interactiveCompletionStorage` under the milestone's sanitized URL (the same content-key scheme `calculatePathRollup`/`journeyMilestonePercentages` already read), so one store and one percent calculation serves the cover page, the toolbar, and My Learning cards.
- Adds a one-time, additive-only backfill: `journeyMilestonePercentages` reads any pre-existing `milestoneCompletionStorage` completion into the new store the first time a journey is read, so a learner's already-recorded progress isn't silently reset by the cutover. The write never lowers a value and is guarded against duplicate backfill writes across renders.
- The cover page's completed/locked calculation now goes through `journeyMilestonePercentages` (the same function backing the toolbar's segmented bar) instead of a separate async read, removing the progress-not-loaded-yet race the CTA/click-target previously guarded against.
- Fixes a follow-on regression found during review: locked milestones are now excluded from the journey-completion check.
- Corrects a stale claim in `docs/design/COMPLETION-MODEL.md` about which store writes milestone completions.
- Mechanism 3 (the backend `CompletionRecord` read path) is intentionally untouched — `recordGuideCompletion`/`recordJourneyCompletion` dual-writes are unchanged, and the durable read side stays deferred per the Path Tracks RFC scope.
- Avoids computing `journeyMilestonePercentages` twice per `LearningPathTableOfContents` render: it's called once now, and both `completedUrls` and `progress` (via the new `percentagesToProgress`) are derived from that single result.

## Why
An investigation (path-tracks-audit-x1) found three non-unified progress-tracking mechanisms in this codebase. The upcoming Path Tracks feature needs exactly one answer to "how far along is this learner," not a fourth. This PR consolidates the two live local mechanisms onto the guide-ID-keyed model, since `calculatePathRollup`/`resolvePathMemberPercentages` are already list-agnostic (`path.guides: string[]` + a `completedMemberIds` set) — the shape a future track's guide list needs to be evaluated against.

## How to verify
- `npm run test:ci` (full suite, including the completion-boundary and round-trip tests touched here)
- Manually check My Learning cards, the cover page, and the in-guide milestone toolbar's progress display for both a fresh guide/path and one with pre-existing `milestoneCompletionStorage`-only progress, confirming the backfill preserves visible progress.

## Breaking changes
None — this is a client-side storage/read-path consolidation. Existing progress recorded only in the old URL-keyed store is backfilled into the new store on first read rather than lost.

## Reversibility
Not a plain revert once this version has been used. Reverting before any milestone is completed under this version is safe: no writes have happened under the new scheme yet, and `main`'s `milestoneCompletionStorage`-only read paths still see everything a learner previously recorded.

Reverting **after** a learner completes a milestone under this version is not safe: `markMilestoneDone` here writes only to `interactiveCompletionStorage`, never to `milestoneCompletionStorage`. `main`'s cover page (`LearningPathTableOfContents.tsx`) and `markMilestoneDone`'s own whole-journey badge-award / `journey_completed` trigger consult only `milestoneCompletionStorage`, so a revert would make that interim completion disappear from the cover page again and would skip the path badge / durable journey-completion record for it. Old (pre-cutover) progress is unaffected either way, since it's still readable from `milestoneCompletionStorage` and is additionally backfilled into `interactiveCompletionStorage` on read.


moxious (changes requested):
Blockers:

1. [blocking] **pf-1925-milestone-storage-oneway-migration — markMilestoneDone stops writing milestoneCompletionStorage, so a revert after real usage would silently un-complete milestones on the cover page and skip the path badge / journey_completed trigger** (medium · reversibility-and-one-way-door · partially reversible)
   markMilestoneDone now writes milestone completion only into interactiveCompletionStorage; the old milestoneCompletionStorage.markCompleted write is deleted with no replacement. On main, the cover page (LearningPathTableOfContents.tsx) derives its checkmarks, CTA target, and locked state purely from milestoneCompletionStorage.getCompleted, and markMilestoneDone's own whole-journey badge-award / journey_completed trigger also reads only milestoneCompletionStorage — neither consults interactiveCompletionStorage. So a milestone completed while this PR is deployed lives only in interactiveCompletionStorage; if the PR is later reverted, main's cover page shows that milestone as incomplete again and the path badge / durable journey-completion record never fires for it. The PR's Reversibility section claims a revert loses no data, but that only holds for progress recorded before this PR shipped.
   Required: Scope the PR's reversibility claim to "safe to revert only before any milestone completion under this version," or add a narrow compatibility write (or reverse backfill) so a post-usage revert doesn't silently regress the cover page and skip badge/journey-completion for interim progress.

Suggestions & nits:

1. [suggestion] **pf-1925-percentages-getter-side-effect — journeyMilestonePercentages performs a storage write + event dispatch as a side effect of what call sites treat as a pure percentage read, and runs twice per LearningPathTableOfContents render** (low · correctness-and-reliability)
   After computing the resolved percentages, journeyMilestonePercentages loops unlocked milestones and calls backfillLegacyMilestoneCompletion, which fires interactiveCompletionStorage.set(...) and dispatchProgress(...) as a side effect of a function documented and named as a percentage getter. LearningPathTableOfContents.tsx then calls it directly for completedUrls, and again indirectly through journeyProgressFromMilestones for progress a few lines later, so the backfill scan and the underlying storage reads run twice per render. A module-level guard prevents a duplicate write today, so this isn't a live bug, but the write is easy to reintroduce a real one against later (a future caller that assumes purity, or a change that removes the guard), and the duplicate computation is pure waste.
   Suggested: Compute journeyMilestonePercentages once per render and derive both completedUrls and progress from that single result, and/or move the backfill write into a useEffect at the call sites so the storage write isn't reachable from a plain read function.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1925
Summary: consolidate milestone progress onto interactiveCompletionStorage
Verdict: Request Changes
Results: 1 blocker, 1 non-blocking finding, 0 follow-ups

Spidy88:
Addressed both findings from @moxious's review:

**1. [blocking] pf-1925-milestone-storage-oneway-migration** — took the smaller fix per the requested scope: no new backfill/compat-write machinery. Corrected the PR's **Reversibility** section to state plainly that reverting is only safe *before* any milestone is completed under this version — a post-usage revert would make `main`'s `milestoneCompletionStorage`-only read paths (cover page, badge-award/`journey_completed` trigger) miss any milestone completed while this version was deployed, since `markMilestoneDone` here writes only to `interactiveCompletionStorage`. Pre-cutover progress is unaffected either way (still readable from the legacy store, and additionally backfilled forward). No code change for this one — it's a documentation-accuracy fix.

**2. [suggestion] pf-1925-percentages-getter-side-effect** — fixed (4d34eb5e). Extracted `percentagesToProgress(percentages)` out of `journeyProgressFromMilestones`, and `LearningPathTableOfContents` now calls `journeyMilestonePercentages` once per render and derives both `completedUrls` and `progress` from that single result, instead of calling it a second time indirectly through `journeyProgressFromMilestones`. `journeyProgressFromMilestones` itself is unchanged for its other callers (`DocsPanelContentArea`, `getJourneyProgress`, etc.).

Full `test:ci` (574 suites, 9478 tests) and `typecheck` are clean on 4d34eb5e.

Spidy88:
Rebased onto \`main\` again (65cbfceb) — main moved another 8 commits while the above was in flight, including #1939 which fixed the exact whole-journey identity-divergence bug this branch's ported \`it.failing\` test was tracking (\`resolveJourneyCompletionIdentity\`). That test is now a real passing assertion instead of an expected failure, rewritten against the current \`seedMilestoneComplete\`/5-arg \`markMilestoneDone\` pattern. No other content changes from the previous push — same two fixes described above. \`typecheck\` and full \`test:ci\` (577 suites, 9535 tests) clean, and all 35 CI checks (including \`CI Gate\` and the e2e/compatibility matrix) are green on 65cbfceb.

moxious (changes requested):
Blockers:

1. [blocking] **pf-1925-backfill-lost-updates — Backfill legacy completions without losing sibling writes** (high · correctness-and-reliability)
   At `src/docs-retrieval/learning-journey-helpers.ts:166-173,214-219`, a single read launches one unawaited write per legacy-complete milestone and permanently guards each key. Each `set` independently reads and replaces the same JSON record (`src/lib/storage/bounded-record-storage.ts:86-94`; fresh JSON objects at `src/lib/user-storage.ts:279-287`). With legacy m1/m2 and an empty new store, both writes read {}, and only the last survives. Re-rendering cannot repair the guarded m1; completing m3 then fails the new whole-journey predicate at helper:850-853, skipping its badge and durable record despite complete legacy progress. The scheduling model reproduced this; current tests use an immediate Map mutation or assert one mocked write, so they cannot catch it.
   Required: Make migration an awaited serialized or single-record merge, coordinated with the completion write, and mark migration done only after persistence succeeds. Add a real-storage integration regression with two legacy-complete members, a final new completion, a repeated read, and assertions for every stored member plus badge/journey emission.

2. [blocking] **pf-1925-milestone-url-roundtrip — Preserve milestone identity across launch URL variants** (high · completion-records)
   At `src/docs-retrieval/learning-journey-helpers.ts:740-744,815-816,852`, completion is written under `currentUrl` but checked against the manifest URL. `sanitizeContentKey` does not reconcile URL spellings. For the already-supported launch `https://ex/m1/content.json` with milestone URL `https://ex/m1/` (the current boundary test itself uses this at lines 687-694,749-768), the guide record fires but the whole-journey check misses it. The cover also stays incomplete because the HTTP branch of `path-member-join.ts:175-177` reads the manifest URL verbatim, keeping the next milestone locked. Base normalized both sides with `getMilestoneSlug`; this change removes that round trip. The test checks guide emission and any numeric recommendation write, never the missing journey or cover completion.
   Required: Resolve the active milestone to a shared canonical member identity for writing and reading, preserving supported launch aliases and the existing safe-key rules; reuse that resolution in the journey completion predicate. Add real-storage round trips for permalink/trailing-slash/content.json/unstyled.html launches and assert both cover progress and the badge/journey record.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1925
Summary: consolidate milestone progress onto interactiveCompletionStorage
Verdict: Request Changes

Spidy88:
Addressed both round-2 blockers from @moxious's review, and verified with a mandatory local `/review` gate before this push:

**1. [blocking] pf-1925-backfill-lost-updates** — fixed. `interactiveCompletionStorage.set()` does its own read-modify-write of one shared JSON record; the legacy backfill fired one unawaited `set()` per legacy-complete milestone, so two in flight without an await between them could read the same pre-write record and lose one of them (a real hazard — reproduced directly in a new characterization test on `bounded-record-storage.ts`). Added `queueMilestoneCompletionWrite`, which serializes every write this module makes (the backfill and `markMilestoneDone`'s own write) onto one queue, and moved the backfill's "done" guard to fire only once the write actually persists (with a separate in-flight guard so a concurrent render doesn't re-queue a write already waiting its turn).

**2. [blocking] pf-1925-milestone-url-roundtrip** — fixed. `markMilestoneDone` wrote under the raw launch URL's content key while the whole-journey membership check and the cover page/toolbar read by the manifest's own milestone URL — a `.../content.json` launch and the manifest's trailing-slash URL sanitize to different keys. Added `resolveMilestoneContentKey`, which canonicalizes a launch URL to the manifest's own spelling for that milestone (matched by slug) before computing the content key, used on both the write and the membership-check read so they're provably aligned rather than independently correct.

Both fixes have real-storage (not mocked) regression tests, each verified to fail against the pre-fix code and pass against the fix (reverted-and-restored test runs, not just asserted).

An independent adversarial verification pass surfaced one legitimate but pre-existing, out-of-scope issue: two other call sites (`completion-store.ts`, `MarkCompleteFooter.tsx`) write to the same storage record without going through this new queue, so they can still race against each other or against this module — not introduced or worsened by this PR, but a real gap in the underlying primitive. Filed as #1954 for separate follow-up rather than expanding this PR's scope.

`typecheck`, `lint`, and full `test:ci` (578 suites, 9560 tests) are clean on the latest commit.

Spidy88:
Addressed all 3 issues from the new Cursor Bugbot round (against 3eb33649), each verified real before fixing:

**1. [High] Completion writes race and drop keys** — fixed at the root. The previous round's fix only serialized writes made from within `learning-journey-helpers.ts`; `completion-store.ts` and `MarkCompleteFooter.tsx` write to the same shared `interactiveCompletionStorage` record directly, unqueued, and could still race against it. Moved the serialization into `BoundedRecordStorage` itself: every mutating method (`set`/`clear`/`clearMany`/`cleanup`/`clearAll`) on a store instance now shares one internal queue, so **any** caller writing to that instance is serialized against every other caller, regardless of module. (`setInternal`'s quota-retry path calls the new internal `cleanupInternal()` directly rather than the now-serialized public `cleanup()`, to avoid a re-entrant deadlock.)

**2. [Medium] Journey check misses legacy completions** — fixed. `GuideReaderOverlay.tsx` renders no milestone toolbar and can call `markMilestoneDone` directly, so the only backfill trigger (cover page / toolbar) never runs for that surface. The whole-journey membership check now also falls back to `milestoneCompletionStorage` directly for any URL not yet in `interactiveCompletionStorage`, and queues the same backfill for consistency.

**3. [Medium] Missing milestones array can throw** — fixed. `recordGuideCompletionForSurface` only optional-chained `learningJourney`, not `milestones` itself; a present `learningJourney` with a missing `milestones` array threw, and `markMilestoneDone` never ran. Guarded both this site and a second one a few lines later with the same risk.

Each fix has a real-storage (or, for #3, a plain unit) regression test, verified against reverted code to fail before the fix and pass after. An independent adversarial verification pass re-confirmed all three by reverting each fix individually and re-running the tests, plus a broader regression sweep (learning-journey-helpers*, completion-store, mark-complete, user-storage, learning-paths.hook — 181 tests) to confirm the other unqueued callers named in finding 1 aren't broken by the new serialization.

`typecheck`, `lint`, and full `test:ci` (578 suites, 9563 tests) are clean on the latest commit.

Spidy88:
Ran a formal `/review` pass (via the Skill tool, Route → Observe → Verify, not just an ad-hoc check) on the round-3 diff before this push, per the standing local-gate rule. It confirmed all three Bugbot findings are correctly fixed, and surfaced one real, non-blocking gap it verified was newly introduced by the finding-3 fix: guarding `metadata.learningJourney.milestones` with `?? []` stopped the crash, but on that same malformed-input path it now computes a 0% mean over the empty fallback and unconditionally writes it to `journeyCompletionStorage` — silently overwriting whatever real percentage the recommendation card was reading for that journey. Before the guard, this same malformed shape threw before reaching that code at all, so nothing was overwritten; the fix traded a crash for a silent data loss on an admittedly narrow, type-violating input.

Fixed: the refresh now guards on `milestones` itself, not just `learningJourney`, so it's skipped entirely on malformed input rather than computing a spurious value. Extended the existing "malformed journey metadata" test to assert the journey-progress write is skipped, verified against reverted code to fail before this fix and pass after.

`typecheck`, `lint`, and full `test:ci` (578 suites, 9563 tests) are clean on the latest commit (342915bc).

moxious (approved):



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1925
Summary: consolidate milestone progress onto interactiveCompletionStorage
Verdict: Approve

PR #1937, opened 2026-09-17
fix(mark-complete): await the completion write before announcing progress

### LINKED ISSUE #1931: Mark complete announces progress before its own write has landed
The "Mark complete" button announces 100% progress before its own storage write has landed, so a listener woken by that announcement can read the old value.

## Location

`src/components/mark-complete/MarkCompleteFooter.tsx`, in the click handler:

```ts
void interactiveCompletionStorage.set(contentKey, 100);
dispatchProgress({ kind: 'guide', contentKey, percentage: 100, hasProgress: true });
```

The write is fire-and-forget — `void` with a `.catch` for logging — and the event fires on the very next line. Anything that recomputes a total from storage in response to that event races the write and can use the stale percentage.

## Impact

Bounded and intermittent: a rollup read on that event may be one guide short until something else recomputes it. Nothing is lost permanently, because the write does land.

## Fix shape

Await the write before dispatching, keeping the existing failure logging:

```ts
try {
  await interactiveCompletionStorage.set(contentKey, 100);
} catch (error) {
  logger.warn('Failed to persist guide completion percentage', { error });
}
dispatchProgress({ kind: 'guide', contentKey, percentage: 100, hasProgress: true });
```

The handler is already a `useCallback`, so it can be made async. Keep the existing behaviour that the completion write does not wait on the celebration animation.

## Also worth checking

`src/components/interactive-tutorial/interactive-section.tsx` dispatches a section event and then calls `sectionDoneStorage.set(...)`, which has the same shape. Confirm whether that write is synchronous before deciding if it needs the same treatment — do not change it blind.

## Test

A test that resolves the storage write on a deferred promise and asserts the event has not fired until the write settles.

hackov81:
## Summary

The Mark complete control announced 100% progress before its own `interactiveCompletionStorage` write had settled, so a listener that recomputed a rollup from storage on that event could read the previous percentage. The click handler now awaits the write (and logs a failure without swallowing the announcement) before `dispatchProgress`. Celebration and continue stay on the click path so the write does not wait on the animation.

## How to verify

1. Open a guide that is not yet complete and click Mark complete. The footer should still celebrate and show Completed immediately.
2. With a delayed storage write in mind: a rollup listener subscribed to the progress event should see 100% for that content key, not the pre-click value.
3. With reduced motion enabled, continue should still happen on the click itself, not after the write.

## Breaking changes

## Out of scope

`interactive-section.tsx` also dispatches then calls `sectionDoneStorage.set`, which is async. Left unchanged because the issue asked not to treat that path blindly; it is a different event shape (section, not guide percentage).

Fixes #1931


moxious:
@hackov81 thank you for the contribution! I'd be happy to take a look at this and merge it, but to proceed we need the community CLA signed, and to ensure that all commits are signed. Thank you.

Issue #1923, opened 2026-09-16
Pay down the orphan-module ratchet's grandfathered baseline (7 orphans + 7 test-only files)


moxious:
### Summary

Pay down the initial grandfathered baseline left by the orphan-module ratchet added for #1743 (file-level orphan detection over `buildModuleGraph()`). 7 orphaned files and 7 test-only-reachable files are allowlisted there today; the goal is to get both allowlists to zero (or to explicit permanent `by-design` entries where that is the right call).

### User impact / flow change

Internal only - no user-facing change.

### Acceptance criteria

- [ ] Every entry in `ALLOWED_ORPHANED_MODULES_ENTRIES` (`src/validation/architecture.test.ts`) is resolved: the file is deleted (dragging any doc reference with it), or a real consumer is repointed at its barrel so the file becomes reachable — then the entry is removed.
- [ ] Every entry in `ALLOWED_TEST_ONLY_REACHABLE_ENTRIES` is resolved: either reconnected to production code, deleted, or deliberately re-tagged `by-design` with a substantive reason (e.g. governance tooling that is legitimately never bundled) — then removed from the ratchet's growable side or moved to the by-design bucket, whichever the implementation lands on.
- [ ] Both allowlists are empty, or contain only reasoned `by-design` entries the team explicitly signed off on.
- [ ] `npm run test:governance` passes against the shrunk allowlists.

### Scope and non-goals

File-level only, same as the parent ratchet. Export/symbol-level dead code (unused exports, unused exported types — the "knip half" of #1743's original two-approach writeup) is explicitly out of scope here; file separately if that work is still wanted.

Two of the seven test-only entries (`validation/package-io.ts`, `validation/validate-package.ts`) are also reached from `src/cli/**`, which `buildModuleGraph()` structurally excludes (`EXCLUDED_TOP_LEVEL`) — they are not actually test-only in truth, just invisible to the scan from that side too. Worth keeping in mind when deciding their disposition; may need `by-design` rather than a fix.

### Where it lives / grounding

- `src/validation/import-graph.ts` — `findOrphanedModules()`, `APP_ENTRY_ROOTS`.
- `src/validation/architecture.test.ts` — `ALLOWED_ORPHANED_MODULES_ENTRIES`, `ALLOWED_TEST_ONLY_REACHABLE_ENTRIES`, describe block `Import graph: orphaned modules`.

**Orphaned (7):**

[...]

PR #1940, opened 2026-09-17
test(validation): add file-level orphan-module ratchet to the import graph

moxious:
## Intent

Implement grafana-pathfinder-app issue #1743 (file-level orphan-module detection, step 3 of the dead-code enforcement work started by #1742/#1744). The issue itself said it was not ready for an agent to one-shot without a human decision on approach/entrypoints/test-only-reachability/baseline handling — those decisions were already recorded 2026-09-05 by David Allen under theme T6 via the prep-issues bot comment (verdict=ready): build the check on this repo's own import-graph.ts/architecture.test.ts ratchet machinery (approach b), not knip (approach a, deferred to a separate future issue as export/symbol-level dead code is out of scope here). A file reached only from a test import (not from the app) is a third state (testOnlyReachable), not an orphan, because buildModuleGraph() excludes test files as both nodes and edge targets so that edge is structurally invisible otherwise. I confirmed these decisions still hold against current main and got the user's explicit sign-off on: (1) file-level-only scope, (2) testOnlyReachable as its own small ratchet allowlist (not just informational), (3) grandfather the whole current baseline in this PR and file one follow-up paydown issue rather than deleting anything now (matching the #1815/#1531 ratchet-shaped-PR convention already used in this repo). Implementation: added APP_ENTRY_ROOTS (['module.tsx'], the plugin entrypoint), getTestImportedNodes(), and findOrphanedModules() to src/validation/import-graph.ts; wired two new ratchet allowlists (ALLOWED_ORPHANED_MODULES_ENTRIES, ALLOWED_TEST_ONLY_REACHABLE_ENTRIES) and a new 'Import graph: orphaned modules' describe block into src/validation/architecture.test.ts following the exact existing pattern used by the cycle-detection and percentage-calculation ratchets (assertRatchet, accountability validation, progress-report line); added unit tests for findOrphanedModules in import-graph.test.ts using synthetic graphs (graphFromAdjacency) so the test-import edge collection is injectable rather than always hitting the real filesystem. I verified every one of the 7 orphaned + 7 test-only-reachable baseline entries by hand (grepped real import statements, not just filenames-in-comments) before writing its allowlist reason — all 7 orphans are genuine (a stale duplicate style file, a dead component, an undocumented constant, and four barrels whose sole consumers all deep-import around them) and all 7 test-only entries are each imported exclusively by their own sibling *.test.ts file (two of those, validation/package-io.ts and validation/validate-package.ts, are also reached from src/cli/** which the graph structurally excludes — noted in their allowlist reason and in the follow-up issue). Filed the paydown follow-up as https://github.com/grafana/grafana-pathfinder-app/issues/1923 and tracked every grandfathered entry to it. No production behavior changes anywhere — this is a pure governance/test-infrastructure addition, matching the T6 ratchet-shaped-PR convention: land the mechanism, fix nothing else, grandfather today's findings, follow up separately to zero.

## What Changed

- `src/validation/import-graph.ts` gains a file-level dead-code scan: `APP_ENTRY_ROOTS` (`module.tsx`), a BFS forward-reachability walk, and `findOrphanedModules()` returning two buckets — `orphaned` (no inbound edge at all) and `offGraphReachable` (imported, but only from a test file or from tooling the graph never walks). `collectSourceFiles()` is refactored onto a shared `collectFilesUnder()` helper, and new `OFF_GRAPH_IMPORTER_ROOTS` / `collectOffGraphImporterFiles()` / `getOffGraphImportedNodes()` recover the edges `buildModuleGraph()` structurally drops (`src/cli/`, `src/test-utils/`, repo-root `tests/`, and `*.test.ts` importers).
- `src/validation/architecture.test.ts` wires two new ratchet allowlists — `ALLOWED_ORPHANED_MODULES_ENTRIES` (7 entries: a superseded duplicate style file, a dead component, an unimported constants module, and four barrels whose consumers all deep-import around them) and `ALLOWED_OFF_GRAPH_REACHABLE_ENTRIES` (7 entries, each justified by its actual test-only or CLI-only importer) — plus an `Import graph: orphaned modules` describe block using the existing `assertRatchet` / accountability-validation / progress-report pattern. Every entry is grandfathered against tracking issue #1923; both lists only shrink.
- `src/validation/import-graph.test.ts` covers `findOrphanedModules` against synthetic graphs (clean scan, orphan, off-graph classification, transitive off-graph chaining, already-reached node, entry-root drift) and asserts `getOffGraphImportedNodes` / `collectOffGraphImporterFiles` resolve real test-file, `src/cli/`, and `tests/` importers. `AGENTS.md` and `.cursor/rules/systemPatterns.mdc` document orphan reachability as a third check orthogonal to tiers. No production code is modified.

## Risk Assessment

✅ Low: Pure test/governance infrastructure with no production behavior change: the two-pass reachability partition is logically sound under every state I traced, all 14 grandfathered baseline entries check out against real import statements, the single-webpack-entry assumption behind APP_ENTRY_ROOTS is verified, the new doc prose does not perturb the existing tier-doc parser, and the only findings are latent or informational.

## Testing

I drove the governance check the way a contributor hits it and then tried to break each boundary the intent names. Baseline reproduced the hand-verified 7 orphans / 7 off-graph-reachable footprint and the whole governance command stayed green. Adversarially, a new unimported production file is caught as an orphan with the delete/barrel advice; the identical file drops into the off-graph-reachable third state when its only importer is a sibling test, src/cli/, src/test-utils/, or the repo-root tests/ tree, and drags transitively-reached files into the same bucket rather than the orphan bucket; a file genuinely imported from module.tsx yields a clean scan, so the check does not false-positive on live app code. On the shrink side, deleting a grandfathered dead barrel makes the ratchet insist the allowlist entry be removed, and both new lists reject by-design tracking, a rubber-stamp reason, and non-issue tracking text — enforcing the accountability decision recorded earlier in this run. Real entrypoint drift (renaming src/module.tsx) fails loud with 677 listed violations; worth noting for the reviewer that on the real graph drift lands mostly in the off-graph bucket (19 orphaned / 669 off-graph) rather than "everything orphaned" as the synthetic unit test depicts, because test files import nearly the whole tree — either way it is a loud ratchet failure, never a silent pass. There is no rendered UI surface to screenshot: the change touches only src/validation/** and agent docs, and I proved that by building the plugin at target and with import-graph.ts reverted to base, getting 56 byte-identical JS bundles with none of the new symbols present in any chunk. Evidence is CLI transcripts plus the bundle comparison; the worktree is back at b6bf4cc0 with no build output or probe files left behind.

- Live validation: ✅ go - 12 of 12 scenarios driven live against the product

| Scenario | Result | Live | Evidence |
| --- | --- | --- | --- |
| A contributor runs the governance check on a clean tree and sees the grandfathered baseline of 7 orphans and 7 off-graph-reachable modules, all tests green | ✅ pass | live | `npm run test:governance -- --coverage=false` (203 tests, 6 suites) plus `[architecture-ratchet] orphans: orphaned=7 offGraphReachable=7` from the verbose architecture run; evidence/01-baseline-footpr… |
| A contributor adds a production file nothing imports; the check blocks it as a new orphan and tells them to delete it or repoint a barrel | ✅ pass | live | Created src/components/probe-dead-file.ts, ran the architecture suite: `New orphaned modules detected: - components/probe-dead-file.ts` with the full advice text, 1 failed test; evidence/02-dead-file-… |
| A production file whose only importer is its sibling *.test.ts reports as off-graph-reachable, not as an orphan | ✅ pass | live | Added probe-dead-file.test.ts; orphaned stayed 7 and the file appeared under `New off-graph-reachable modules detected`; evidence/03-test-only-importer.log |
| A production file whose only importer is src/cli tooling reports as off-graph-reachable, not as an orphan | ✅ pass | live | Added src/cli/probe-importer.ts as sole importer: orphaned=7 offGraphReachable=8, file listed in the off-graph bucket; evidence/04-cli-importer.log |
| A production file whose only importer is src/test-utils reports as off-graph-reachable, not as an orphan | ✅ pass | live | Added src/test-utils/probe-importer.ts as sole importer: orphaned=7 offGraphReachable=8, file listed in the off-graph bucket; evidence/13-test-utils-importer.log |
| A production file whose only importer is the repo-root Playwright tests/ tree reports as off-graph-reachable, not as an orphan | ✅ pass | live | Added tests/probe-importer.ts as sole importer: orphaned=7 offGraphReachable=8, file listed in the off-graph bucket; evidence/05-repo-root-tests-importer.log |
| A file reachable only transitively through an off-graph-imported file joins the off-graph bucket instead of being misreported as dead | ✅ pass | live | Added src/components/probe-chained.ts imported only by the probe: orphaned stayed 7, offGraphReachable=9 listing both probe-chained.ts and probe-dead-file.ts; evidence/06-transitive-off-graph.log |
| A new production file genuinely imported from the app entrypoint produces a clean scan, so live app code is never falsely flagged | ✅ pass | live | Imported the probe from src/module.tsx: `orphaned=7 offGraphReachable=7`, 31/31 architecture tests passed; evidence/07-app-reachable-clean.log |
| A contributor who actually deletes a grandfathered dead barrel is told to remove its allowlist entry, so the ratchet can only shrink | ✅ pass | live | Removed src/components/UserProfileBar/index.ts: `Stale entries in ALLOWED_ORPHANED_MODULES_ENTRIES (orphaned modules allowlist — violation was fixed, remove the entry): - components/UserProfileBar/ind… |
| Entrypoint drift (src/module.tsx renamed without updating APP_ENTRY_ROOTS) fails the check loudly instead of silently reporting a clean scan | ✅ pass | live | `mv src/module.tsx src/module-renamed.tsx` then ran the architecture suite: orphaned=19 offGraphReachable=669, both orphan ratchet tests failed with 677 individual violations listed; a direct findOrph… |
| Neither new allowlist can be silenced without accountability: by-design tracking, a rubber-stamp reason, and free-text tracking are all rejected | ✅ pass | live | Edited entries in both lists in turn; the accountability test reported by-design not allowed for ALLOWED_ORPHANED_MODULES and ALLOWED_OFF_GRAPH_REACHABLE, the too-short reason, and the invalid &#39;TODO l… |
| An end user of the plugin sees no behavior change: the shipped frontend bundle is identical with and without this change | ✅ pass | live | `npm run build` at target vs. with src/validation/import-graph.ts reverted to base e0649b83 produced 56 byte-identical JS bundles (shasum diff empty), and no new symbol (OFF_GRAPH_IMPORTER_ROOTS, APP_… |

<details>
<summary>Evidence: Live orphan-ratchet transcript (12 driven scenarios, consolidated)</summary>

```text
# Orphan-module ratchet — live transcript (issue #1743)

Surface driven: the developer-facing governance check (`npm run test:governance`,
`npx jest src/validation/architecture.test.ts`). Each block below is a real run
against the real repository import graph (688 production nodes).

## 1. Baseline at target commit b6bf4cc0

    [architecture-ratchet] orphans: orphaned=7 offGraphReachable=7
    [architecture-ratchet] vertical=2 lateral=7 barrel=3 cycles=2 percentageCalculations=6 orphanedModules=7 offGraphReachable=7
    Test Suites: 6 passed, 6 total
    Tests:       203 passed, 203 total

## 2. New dead file (nothing imports src/components/probe-dead-file.ts) -> blocked as an orphan

    New orphaned modules detected:
      - components/probe-dead-file.ts

    A production file under src/ is not reached by APP_ENTRY_ROOTS (module.tsx), and no test file and
    nothing under OFF_GRAPH_IMPORTER_ROOTS (src/cli/, src/test-utils/, the repo-root tests/ tree)
    imports it either. This is usually one of: a genuinely dead file (delete it, and drag any doc
    reference with it), or a barrel (index.ts) whose only consumers deep-import the internal files
    instead ...
    Tests:       1 failed, 30 passed, 31 total

## 3. Same file, sole importer is its sibling *.test.ts -> third state, not an orphan

    [architecture-ratchet] orphans: orphaned=7 offGraphReachable=8
    New off-graph-reachable modules detected:
      - components/probe-dead-file.ts

## 4. Sole importer is src/cli/ tooling -> off-graph-reachable

    [architecture-ratchet] orphans: orphaned=7 offGraphReachable=8
    New off-graph-reachable modules detected:
      - components/probe-dead-file.ts

## 5. Sole importer is src/test-utils/ -> off-graph-reachable

    [architecture-ratchet] orphans: orphaned=7 offGraphReachable=8
    New off-graph-reachable modules detected:
      - components/probe-dead-file.ts

## 6. Sole importer is the repo-root tests/ (Playwright) tree -> off-graph-reachable

    [architecture-ratchet] orphans: orphaned=7 offGraphReachable=8
    New off-graph-reachable modules detected:
      - components/probe-dead-file.ts

## 7. Transitive: probe-chained.ts reached only through the off-graph-imported probe

    [architecture-ratchet] orphans: orphaned=7 offGraphReachable=9
    New off-graph-reachable modules detected:
      - components/probe-chained.ts
      - components/probe-dead-file.ts

## 8. No false positive: same new file imported from src/module.tsx -> clean scan

    [architecture-ratchet] orphans: orphaned=7 offGraphReachable=7
    Tests:       31 passed, 31 total

## 9. Ratchet only shrinks: deleting the grandfathered dead barrel demands the entry go

    Stale entries in ALLOWED_ORPHANED_MODULES_ENTRIES (orphaned modules allowlist — violation was fixed, remove the entry):
      - components/UserProfileBar/index.ts

## 10. Entrypoint drift: `git mv src/module.tsx src/module-renamed.tsx` fails loud

    [architecture-ratchet] orphans: orphaned=19 offGraphReachable=669
    ● Import graph: orphaned modules › should not introduce new orphaned modules beyond the ratchet allowlist
    ● Import graph: orphaned modules › should not introduce new off-graph-reachable modules beyond the ratchet allowlist
    Tests:       4 failed, 27 passed, 31 total
    (677 individual violations listed — no silent clean scan)

## 11. Accountability: both new lists reject `by-design`, rubber-stamp reasons, and non-issue tracking

    - ALLOWED_ORPHANED_MODULES: lib/index.ts: 'tracking' must point to an issue; 'by-design' is not allowed here.
    - ALLOWED_OFF_GRAPH_REACHABLE: validation/test-helpers.ts: 'tracking' must point to an issue; 'by-design' is not allowed here.
    - ALLOWED_ORPHANED_MODULES: lib/index.ts: 'reason' is missing or too short — explain why the violation is tolerated.
    - ALLOWED_OFF_GRAPH_REACHABLE: validation/test-helpers.ts: 'tracking' must be 'by-design', an issue reference ('#1234'), or a GitHub issues URL, got 'TODO later'.

## 12. Shipped bundle unaffected

`npm run build` at target vs. with src/validation/import-graph.ts reverted to base e0649b83:

    IDENTICAL: all 56 emitted JS bundles byte-identical between base and target source

and no new symbol reaches any chunk:

    grep -rlE "OFF_GRAPH_IMPORTER_ROOTS|APP_ENTRY_ROOTS|offGraphReachable|findOrphanedModules" dist --include='*.js'
    NONE: no new symbol ships in any bundle chunk
```
</details>
<details>
<summary>Evidence: Baseline ratchet footprint at target commit</summary>

```text
[architecture-ratchet] orphans: orphaned=7 offGraphReachable=7
[architecture-ratchet] vertical=2 lateral=7 barrel=3 cycles=2 percentageCalculations=6 orphanedModules=7 offGraphReachable=7
Test Suites: 6 passed, 6 total
Tests: 203 passed, 203 total
```
</details>
<details>
<summary>Evidence: New dead file blocked as an orphan (adversarial)</summary>

```text
New orphaned modules detected:
- components/probe-dead-file.ts

A production file under src/ is not reached by APP_ENTRY_ROOTS (module.tsx), and no test file and nothing under OFF_GRAPH_IMPORTER_ROOTS (src/cli/, src/test-utils/, the repo-root tests/ tree) imports it either. This is usually one of: a genuinely dead file (delete it, and drag any doc reference with it), or a barrel (index.ts) whose only consumers deep-import the internal files instead ...
Tests: 1 failed, 30 passed, 31 total
```
</details>
<details>
<summary>Evidence: Same file reclassified to the off-graph-reachable third state (test / cli / test-utils / tests roots)</summary>

```text
[architecture-ratchet] orphans: orphaned=7 offGraphReachable=8
New off-graph-reachable modules detected:
- components/probe-dead-file.ts
```
</details>
<details>
<summary>Evidence: Entrypoint drift fails loud after renaming src/module.tsx</summary>

```text
[architecture-ratchet] orphans: orphaned=19 offGraphReachable=669
● Import graph: orphaned modules › should not introduce new orphaned modules beyond the ratchet allowlist
● Import graph: orphaned modules › should not introduce new off-graph-reachable modules beyond the ratchet allowlist
Tests: 4 failed, 27 passed, 31 total (677 individual violations listed)
```
</details>
<details>
<summary>Evidence: Accountability guards on both new allowlists</summary>

```text
- ALLOWED_ORPHANED_MODULES: lib/index.ts: 'tracking' must point to an issue; 'by-design' is not allowed here.
- ALLOWED_OFF_GRAPH_REACHABLE: validation/test-helpers.ts: 'tracking' must point to an issue; 'by-design' is not allowed here.
- ALLOWED_ORPHANED_MODULES: lib/index.ts: 'reason' is missing or too short — explain why the violation is tolerated.
- ALLOWED_OFF_GRAPH_REACHABLE: validation/test-helpers.ts: 'tracking' must be 'by-design', an issue reference ('#1234'), or a GitHub issues URL, got 'TODO later'.
```
</details>
<details>
<summary>Evidence: Shipped bundle unchanged (target vs base source)</summary>

```text
IDENTICAL: all 56 emitted JS bundles byte-identical between base and target source
grep -rlE "OFF_GRAPH_IMPORTER_ROOTS|APP_ENTRY_ROOTS|offGraphReachable|findOrphanedModules" dist --include='*.js'
NONE: no new symbol ships in any bundle chunk
```
</details>

## Pipeline

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>

<details>
<summary>⚠️ **Review** - 4 infos</summary>

- ⚠️ `src/validation/architecture.test.ts:737` - The `orphaned` bucket cannot see importers under `src/cli/**` or `src/test-utils/**`, because `getTestImportedNodes()` (import-graph.ts:629) is built on `getAllFileImports()` -&gt; `collectSourceFiles()`, which skips every `EXCLUDED_TOP_LEVEL` directory (import-graph.ts:49). Concrete sequence in intended usage: a contributor adds `src/lib/foo.ts` imported only by `src/cli/commands/bar.ts` (or only by a `src/test-utils/` helper). `foo.ts` is a graph node, is unreached from `module.tsx`, and has no test-file importer the scan can see, so it lands in `scan.orphaned` and CI fails with the advice at architecture.test.ts:737 — &#34;a genuinely dead file (delete it, and drag any doc reference with it)&#34; — about live code. This is a near-miss today: the author&#39;s own allowlist notes that `validation/package-io.ts` and `validation/validate-package.ts` are reached from `src/cli/**`, and they only escaped the orphan bucket because each happens to have a sibling test. Note the sibling test-only message (architecture.test.ts:752-757) does mention the CLI blind spot, while the orphan message does not. Smallest in-scope remedy: name the `src/cli/**` / `src/test-utils/**` blind spot in the orphan advice string so the first instruction is not &#34;delete it&#34;. The fuller remedy (collecting import edges from the excluded test/CLI roots) extends the change&#39;s stated scope, which is why this needs your decision rather than an auto-fix.
- ⚠️ `src/validation/architecture.test.ts:433` - `ALLOWED_TEST_ONLY_REACHABLE` is registered with `allowByDesign: false`, so `validateAllowedArchitectureEntries` (import-graph.ts:712) rejects any entry whose `tracking` is `by-design`. That contradicts two things this change ships alongside it: (1) its own failure advice at architecture.test.ts:754-757, which tells the contributor that if the test-only status &#34;is deliberate (e.g. governance/validation tooling that only ever runs under test or the CLI)&#34; they should add an allowlist entry — the only honest tracking value for a permanent case is `by-design`, which is refused, forcing a tracking issue that implies debt; and (2) follow-up #1923, whose acceptance criteria say each entry is resolved by being &#34;deliberately re-tagged `by-design`&#34; and that both lists should end up &#34;empty, or contain only reasoned `by-design` entries&#34;. Several baseline entries are structurally permanent, not debt: `validation/import-graph.ts` is the ratchet machinery itself, and `validation/test-helpers.ts`, `validation/cli-build-contract.ts`, `types/backend-api.schema.ts` are governance/test tooling that will never ship in the bundle, so #1923 cannot be closed for them by fixing anything. Remedy is a policy choice you own: either set `allowByDesign: true` for `ALLOWED_TEST_ONLY_REACHABLE` (matching the vertical/barrel/percentage lists) or reword the advice to state that a tracking issue is required even for permanent cases.
- ℹ️ `src/validation/architecture.test.ts:398` - The allowlist reason for `validation/import-graph.ts` says &#34;imported only by architecture.test.ts and import-graph.test.ts&#34;, but `src/validation/cli-build-contract.ts:5` (a production graph node, not a test file) also imports `./import-graph`. So `import-graph.ts` reaches the `testOnlyReachable` bucket both directly and transitively through another test-only node, and the recorded reason understates the dependency a paydown attempt under #1923 would have to move. Correct the reason string to name `validation/cli-build-contract.ts` as the production importer.
- ⚠️ `src/validation/import-graph.ts:652` - `findOrphanedModules` introduces two guards that are provably no-ops, and one unit test that does not cover what it claims. (a) `nodeSet` plus `roots.filter((root) =&gt; nodeSet.has(root))` (import-graph.ts:652-656): `reachedFromApp` is only ever consulted as `!reachedFromApp.has(node)` for nodes drawn from `graph.nodes`, and `bfsReachable` expands via `adjacency.get(current) ?? []`, which yields nothing for a non-node root — so seeding a non-node root changes no output. (b) The `if (unreached.length === 0) return ...` early return (import-graph.ts:658-660): with `unreached` empty, `testRoots` is `[]`, `bfsReachable(adjacency, [])` is empty, and both final `filter(...).sort()` calls already produce `[]`. Neither guard is required by the stated intent, and removing both is behavior-preserving. Consequence for the tests: `import-graph.test.ts:672` (&#34;ignores a root that is not a node in the graph&#34;) passes identically with the filter deleted, so it documents an incidental property rather than guarding the filter. Recommended remedy is removal of the two guards (and either dropping or re-pointing that test), not hardening them.
- ℹ️ `src/validation/architecture.test.ts:721` - This adds a third mechanical enforcement dimension to `architecture.test.ts` (orphan / test-only reachability from `APP_ENTRY_ROOTS`) alongside the tier model and the Node environment-reachability scan, but neither `.cursor/rules/systemPatterns.mdc` (which documents the other two in detail at lines 87 and 97) nor `AGENTS.md` mentions it. A contributor or agent hitting the new failure has no routing doc explaining the `orphaned` vs `testOnlyReachable` split or the `src/cli` blind spot. Adding that paragraph extends the change beyond its stated &#34;pure governance/test-infrastructure addition&#34; scope, so it is your call whether it lands here or in #1923.

🔧 Fix applied.
5 issues (2 warnings, 3 infos) still open:

- ⚠️ `src/validation/import-graph.test.ts:706` - The test named &#39;resolves imports made from a test file&#39; asserts getOffGraphImportedNodes() contains &#39;validation/package-io.ts&#39;, but no test file imports that module. Its only importers are src/cli/utils/file-loader.ts:9, src/cli/commands/{validate,build-graph,build-snippets,build-repository,build-stats}.ts, and src/validation/validate-package.ts (itself off-graph, and only a transitive edge which this function does not follow). So it exercises the tooling-importer branch, identical to the test above it. Concrete failure: delete the `testImporters` half of getOffGraphImportedNodes (import-graph.ts:665) and both tests in this describe block still pass green — the test-file branch has zero unit coverage despite being the branch the intent calls out as the reason testOnlyReachable exists. Fix: assert a node whose only off-graph importer is a test file, e.g. &#39;validation/unicode-format-characters.ts&#39; (sole importer src/validation/unicode-format-characters.test.ts:5) or &#39;validation/test-helpers.ts&#39; (sole importer src/validation/validate-guide.security.test.ts:2).
- ⚠️ `src/validation/architecture.test.ts:408` - The ALLOWED_TEST_ONLY_REACHABLE_ENTRIES reason for &#39;validation/package-io.ts&#39; reads &#39;Imported by its sibling package-io.test.ts and by src/cli/** tooling&#39;, but src/validation/package-io.test.ts does not exist (`find src -name &#39;package-io*&#39;` yields only src/validation/package-io.ts, src/cli/utils/package-io/, and src/cli/__tests__/package-io.test.ts — the last of which tests the CLI&#39;s own separate package-io module, not this one). The actual importers are six src/cli/** files plus src/validation/validate-package.ts. The classification as testOnlyReachable is still correct (the CLI edges are real), but the accountability reason is the mechanism validateAllowedArchitectureEntries exists to enforce, and this one states a false fact that misdirects the #1923 paydown: a reader will look for a test to retire when the only thing holding the entry open is CLI usage. Fix the reason string to &#39;Imported only by src/cli/** tooling (and by validation/validate-package.ts, itself test-only-reachable) — no test file and nothing in the app imports it.&#39;
- ℹ️ `src/validation/import-graph.test.ts:459` - `tooling.filter((file) =&gt; collectSourceFiles().includes(file))` calls collectSourceFiles() once per element. `tooling` currently holds 184 files (179 under src/cli, 5 under src/test-utils), so this performs 184 full recursive readdirSync walks of src/ (~200 directories each) plus a linear `.includes` over ~1180 paths each time — tens of thousands of redundant syscalls in one assertion. Hoist it: `const sourceFiles = new Set(collectSourceFiles());` before the filter, then `tooling.filter((file) =&gt; sourceFiles.has(file))`. Same applies more mildly in the getOffGraphImportedNodes describe (lines 700 and 706), which re-parses all 184 tooling files with the TypeScript parser twice — hoist one call into the describe body.
- ℹ️ `src/validation/architecture.test.ts:403` - Two more reason strings misstate their importers. (1) Line 403 enumerates &#39;validation/import-graph.ts&#39; as &#39;imported by architecture.test.ts, import-graph.test.ts, and one production node&#39; — it is also imported by src/validation/eslint-disable-justifications.test.ts:5, src/validation/cli-build-contract.test.ts:7, and src/validation/coda-capability-gate.contract.test.ts:5. The enumeration reads as exhaustive but is not. (2) Line 392 calls src/validation/backend-api-contract.test.ts the &#39;sibling&#39; of &#39;types/backend-api.schema.ts&#39;; they are in different directories, so &#39;sibling&#39; is wrong (unlike the other entries, where it is accurate). Both are cosmetic relative to finding 2 but undermine the same accountability record.
- ℹ️ `src/validation/import-graph.ts:665` - getOffGraphImportedNodes() collects off-graph edges from src/ test files and from EXCLUDED_TOP_LEVEL tooling dirs, but not from the repo-root tests/ tree, which does import src modules directly (e.g. tests/e2e-runner/utils/selector-resolver.ts:1 -&gt; src/lib/dom/selector-resolver-core, tests/e2e-runner/utils/guide-runner/static-analysis.ts:1 -&gt; src/constants/json-guide-classification, and 30+ more). I checked every such module and all are independently app-reachable today, so the baseline is correct and nothing is misclassified right now. The latent case: a src/ helper added for a Playwright spec only would land in `orphaned`, and the failure advice at architecture.test.ts:731 would tell the contributor &#39;no test file... imports it either... delete it&#39; about a live file. The advice text and the AGENTS.md / systemPatterns.mdc paragraphs also describe the covered set as &#39;a test file, or tooling under EXCLUDED_TOP_LEVEL&#39;, which omits this. The remedy (adding a repo-root tests/ importer sweep, or narrowing the advice wording to say the check does not see repo-root tests/) extends the change past its stated file-level-orphan scope, so it is the remedy rather than the defect that needs your authorization — it may equally belong in #1923.

🔧 Fix applied.

- ⚠️ `src/validation/architecture.test.ts:439` - A fix round flipped `ALLOWED_TEST_ONLY_REACHABLE` from `allowByDesign: false` (commit 3386d119) to `allowByDesign: true`, and added a doc-comment sentence sanctioning it (&#34;A permanently test-only entry may carry the `by-design` marker&#34;) plus a matching clause in the assertRatchet failure advice. This is an accountability escape hatch the intent does not ask for: the authorized decision was &#34;(3) grandfather the whole current baseline in this PR and file one follow-up paydown issue&#34;, and all 7 current entries point at #1923 — none uses `by-design`. Concretely, validateAllowedArchitectureEntries (import-graph.ts:736) now accepts `tracking: &#39;by-design&#39;` on this list, so a future contributor can permanently exempt a test-only-reachable file with no tracking issue, which is exactly what the sibling new list ALLOWED_ORPHANED_MODULES still forbids. The remedy is removing the flag (revert to `allowByDesign: false` and drop the two prose clauses), not hardening it — but because it exceeds the minimal fix the earlier findings asked for, it needs your call rather than an automatic revert. If a permanent category genuinely exists here, that belongs in the #1923 paydown once the plan decides which entries are debt and which are permanent.
- ℹ️ `src/validation/import-graph.test.ts:696` - Commit d0dbd720 removed both the `roots.filter((root) =&gt; nodeSet.has(root))` guard in findOrphanedModules and the unit test that pinned it (`&#39;ignores a root that is not a node in the graph&#39;`). Dropping the guard is correct — it was a genuine no-op, since bfsReachable seeding `reached` with a non-node root cannot affect `unreached`, which only iterates `graph.nodes`. But the deleted test was pinning behavior, not the guard, and that behavior still holds and still matters: it is the failure mode when APP_ENTRY_ROOTS drifts (e.g. `module.tsx` renamed), where the check must degrade to &#34;everything is orphaned&#34; (a loud ratchet failure) rather than throwing or silently reporting nothing. The test passes unchanged against the current implementation — `findOrphanedModules(graphFromAdjacency({ a: [] }), [&#39;nonexistent-root&#39;], [])` still returns `{ orphaned: [&#39;a&#39;], testOnlyReachable: [] }` — so restoring it is a pure coverage recovery with no code change.

🔧 Fix applied.
2 infos still open:

- ℹ️ `src/validation/import-graph.ts:110` - `OFF_GRAPH_IMPORTER_ROOTS` covers the `EXCLUDED_TOP_LEVEL` src dirs plus `tests`, but not the repo-root single-file Node context that the sibling `NODE_CONTEXT_ROOTS` (line 807) does cover: `playwright.config.ts`, nor `.config/webpack/*.ts`. I confirmed zero current impact — I grepped every tracked .ts/.tsx/.js/.mjs outside `src/` and `tests/` and none of them imports a src module (the only `src/` mentions in `scripts/cli-build-utils.js`, `scripts/sync-terms-and-conditions.js` and `.config/webpack/utils.ts` are `readFileSync`/glob strings, not imports). The latent case is the same one the earlier `tests/` finding described: if a future build/config file imports a src constant (e.g. injecting a version into webpack config), that src file lands in `orphaned` and the advice at architecture.test.ts:744 tells the contributor to delete a live file. Note also that the remedy is not a one-element append as-is: `collectOffGraphImporterFiles` (line 113) hands every root to `collectFilesUnder`, whose `walk()` calls `fs.readdirSync` unconditionally, so adding a *file* root such as `playwright.config.ts` throws ENOTDIR — `collectNodeContextFiles` guards this with `statSync(...).isDirectory()`. Adding roots (and the file/dir guard) extends the change&#39;s root set beyond its stated scope, so it is the remedy rather than the defect that needs authorization; it may equally belong in #1923.
- ℹ️ `src/validation/architecture.test.ts:407` - After the authorized broadening from test-file importers to all of `OFF_GRAPH_IMPORTER_ROOTS`, the `testOnlyReachable` label no longer describes its contents. The clearest witness is in the list itself: the `validation/package-io.ts` entry sits in `ALLOWED_TEST_ONLY_REACHABLE_ENTRIES` with the reason &#39;No test file imports it&#39; (line 409) — a self-contradicting record in the accountability artifact. The progress line at architecture.test.ts:838 likewise prints `testOnlyReachable=7` when one of those 7 is CLI-only-reachable. Nothing is computed wrongly; the classification is right and the failure advice spells out the real rule. The remedy is a rename (`offGraphOnlyReachable` / `ALLOWED_OFF_GRAPH_ONLY_REACHABLE_ENTRIES`, matching the already-renamed `getOffGraphImportedNodes`), which contradicts the name the intent explicitly fixed as the &#39;third state&#39;, so it is your call rather than a mechanical fix. A narrower alternative that keeps the intent&#39;s name is to reword just the `package-io.ts` reason so it does not open by denying the bucket it is in.

🔧 Fix applied.
4 infos still open:

- ℹ️ `src/validation/import-graph.ts:656` - `getOffGraphImportedNodes()` collects test-file importers via `isTestFile()` (line 70), which matches only `*.test.ts` / `*.spec.ts` / `test-utils/` / `__tests__/`. Its sibling in the same module, `isJestManagedTestFile()` (line 818), mirrors the repo&#39;s real jest `testMatch` and additionally matches `.jest.ts` — `.config/jest.config.js:21` really is `*.{spec,test,jest}.{js,jsx,ts,tsx}`. Concrete consequence of the new check: a contributor adds `src/foo/bar.jest.ts` (a real jest-run test). `isTestFile` says production, so `buildModuleGraph()` makes it a graph *node*, nothing reaches it from `module.tsx`, and nothing off-graph imports it — so the new ratchet reports it under `orphaned` and the advice at architecture.test.ts:742 tells the contributor to delete a live test file. Its imports are also counted as production edges, so anything it pulls in reads as app-reachable. No `.jest.*` file exists in the repo today (I checked `src/` and `tests/`), so this is latent, not live. Flagging as ask-user rather than auto-fix because the honest remedy is reconciling `isTestFile` with `isJestManagedTestFile`, and `isTestFile` is pre-existing shared machinery for the tier, cycle, and barrel checks — changing its semantics reaches well beyond this change&#39;s stated scope and plausibly belongs in #1923.
- ℹ️ `src/validation/architecture.test.ts:742` - Entrypoint drift now fails loud (the round-3 regression test at import-graph.test.ts:696 pins that, correctly), but the failure it produces is hard to act on: if an `APP_ENTRY_ROOTS` entry stops matching a node — a typo&#39;d second entry, or a nested plugin whose `module.tsx` webpack&#39;s `getEntries()` picks up but this constant does not — every other node lands in `orphaned`, and the advice tells the reader each one is &#39;usually ... a genuinely dead file (delete it, and drag any doc reference with it)&#39;. ~600 files with delete-it advice is loud but points at the wrong remedy. A precondition assert that each `APP_ENTRY_ROOTS` entry is present in `graph.nodes` would turn that into one precise message naming the stale root. Marked ask-user because the remedy adds a new check rather than correcting existing logic, and because the user already adjudicated this area in round 3 (choosing fail-loud over the silenced `roots.filter(nodeSet.has)` guard) — this does not contradict that choice, it only sharpens the message. Low value; fine to skip.
- ℹ️ `src/validation/import-graph.test.ts:716` - The three `getOffGraphImportedNodes` branch tests (lines 711, 716, 721) and `collectOffGraphImporterFiles`&#39;s nesting test (line 462) assert hardcoded real-repo facts: that `src/cli/commands/build-snippets.ts` imports `guided-action-validator`, that `tests/helpers/completion.helpers.ts` imports `completion-write-timing`, and that `src/cli/commands/validate.ts` and `tests/helpers/completion.helpers.ts` exist at those paths. I verified each is currently the sole off-graph importer for its assertion, so the tests do genuinely exercise the cli and tests/ branches that synthetic graphs cannot reach — that is real coverage the round-3/4 findings asked for. The cost is that an unrelated refactor (dropping one import, renaming one helper) breaks a test that is not about that code. Noting the tradeoff only: the robust alternative needs `getOffGraphImportedNodes` to expose importer-to-node attribution, which is a larger API change than this coverage warrants. No action recommended.
- ℹ️ `src/validation/import-graph.ts:654` - For the record: the --intent text describes the third state as `testOnlyReachable` / `getTestImportedNodes()` / `ALLOWED_TEST_ONLY_REACHABLE_ENTRIES`, defined as &#39;a file reached only from a test import (not from the app)&#39;. The shipped code is broader and differently named — `offGraphReachable` / `getOffGraphImportedNodes()` / `ALLOWED_OFF_GRAPH_REACHABLE_ENTRIES`, covering test files plus all tooling under `OFF_GRAPH_IMPORTER_ROOTS` (`src/cli/`, `src/test-utils/`, repo-root `tests/`). This is not an unauthorized deviation: the broadening was accepted in round 3 and the rename was explicitly instructed in round 4 (&#39;rename the testOnlyReachable field/bucket ... to something that accurately reflects it now covers test files AND src/cli/test-utils/tests tooling imports&#39;). It is also the better behavior — `src/cli/**` is excluded from the graph by the same structural mechanism as test files, so without it live CLI-only code such as `validation/package-io.ts` would read as dead. No source change needed; the only residual is that the PR body should describe the off-graph bucket rather than the test-only one, and a later pipeline step owns that.
</details>

<details>

✅ No issues found.
- Live validation: ✅ go - 12 of 12 scenarios driven live against the product

| Scenario | Result | Live | Evidence |
| --- | --- | --- | --- |
| A contributor runs the governance check on a clean tree and sees the grandfathered baseline of 7 orphans and 7 off-graph-reachable modules, all tests green | ✅ pass | live | `npm run test:governance -- --coverage=false` (203 tests, 6 suites) plus `[architecture-ratchet] orphans: orphaned=7 offGraphReachable=7` from the verbose architecture run; evidence/01-baseline-footpr… |
| A contributor adds a production file nothing imports; the check blocks it as a new orphan and tells them to delete it or repoint a barrel | ✅ pass | live | Created src/components/probe-dead-file.ts, ran the architecture suite: `New orphaned modules detected: - components/probe-dead-file.ts` with the full advice text, 1 failed test; evidence/02-dead-file-… |
| A production file whose only importer is its sibling *.test.ts reports as off-graph-reachable, not as an orphan | ✅ pass | live | Added probe-dead-file.test.ts; orphaned stayed 7 and the file appeared under `New off-graph-reachable modules detected`; evidence/03-test-only-importer.log |
| A production file whose only importer is src/cli tooling reports as off-graph-reachable, not as an orphan | ✅ pass | live | Added src/cli/probe-importer.ts as sole importer: orphaned=7 offGraphReachable=8, file listed in the off-graph bucket; evidence/04-cli-importer.log |
| A production file whose only importer is src/test-utils reports as off-graph-reachable, not as an orphan | ✅ pass | live | Added src/test-utils/probe-importer.ts as sole importer: orphaned=7 offGraphReachable=8, file listed in the off-graph bucket; evidence/13-test-utils-importer.log |
| A production file whose only importer is the repo-root Playwright tests/ tree reports as off-graph-reachable, not as an orphan | ✅ pass | live | Added tests/probe-importer.ts as sole importer: orphaned=7 offGraphReachable=8, file listed in the off-graph bucket; evidence/05-repo-root-tests-importer.log |
| A file reachable only transitively through an off-graph-imported file joins the off-graph bucket instead of being misreported as dead | ✅ pass | live | Added src/components/probe-chained.ts imported only by the probe: orphaned stayed 7, offGraphReachable=9 listing both probe-chained.ts and probe-dead-file.ts; evidence/06-transitive-off-graph.log |
| A new production file genuinely imported from the app entrypoint produces a clean scan, so live app code is never falsely flagged | ✅ pass | live | Imported the probe from src/module.tsx: `orphaned=7 offGraphReachable=7`, 31/31 architecture tests passed; evidence/07-app-reachable-clean.log |
| A contributor who actually deletes a grandfathered dead barrel is told to remove its allowlist entry, so the ratchet can only shrink | ✅ pass | live | Removed src/components/UserProfileBar/index.ts: `Stale entries in ALLOWED_ORPHANED_MODULES_ENTRIES (orphaned modules allowlist — violation was fixed, remove the entry): - components/UserProfileBar/ind… |
| Entrypoint drift (src/module.tsx renamed without updating APP_ENTRY_ROOTS) fails the check loudly instead of silently reporting a clean scan | ✅ pass | live | `mv src/module.tsx src/module-renamed.tsx` then ran the architecture suite: orphaned=19 offGraphReachable=669, both orphan ratchet tests failed with 677 individual violations listed; a direct findOrph… |
| Neither new allowlist can be silenced without accountability: by-design tracking, a rubber-stamp reason, and free-text tracking are all rejected | ✅ pass | live | Edited entries in both lists in turn; the accountability test reported by-design not allowed for ALLOWED_ORPHANED_MODULES and ALLOWED_OFF_GRAPH_REACHABLE, the too-short reason, and the invalid &#39;TODO l… |
| An end user of the plugin sees no behavior change: the shipped frontend bundle is identical with and without this change | ✅ pass | live | `npm run build` at target vs. with src/validation/import-graph.ts reverted to base e0649b83 produced 56 byte-identical JS bundles (shasum diff empty), and no new symbol (OFF_GRAPH_IMPORTER_ROOTS, APP_… |

- <code>`npx jest src/validation/architecture.test.ts src/validation/import-graph.test.ts --coverage=false` (baseline, 174 tests)</code>
- <code>`npm run test:governance -- --coverage=false` (203 tests, 6 suites)</code>
- <code>`npx jest src/validation/architecture.test.ts --coverage=false --verbose` capturing the `[architecture-ratchet] orphans: orphaned=7 offGraphReachable=7` footprint line</code>
- <code>Manual: added an unimported `src/components/probe-dead-file.ts` and re-ran the orphan ratchet (expect orphan failure)</code>
- <code>Manual: added `src/components/probe-dead-file.test.ts` as sole importer and re-ran (expect off-graph-reachable reclassification)</code>
- <code>Manual: added `src/cli/probe-importer.ts` as sole importer and re-ran</code>
- <code>Manual: added `src/test-utils/probe-importer.ts` as sole importer and re-ran</code>
- <code>Manual: added `tests/probe-importer.ts` (repo-root Playwright tree) as sole importer and re-ran</code>
- <code>Manual: added `src/components/probe-chained.ts` imported only by the probe and re-ran (transitive off-graph reachability)</code>
- <code>Manual: imported the probe from `src/module.tsx` and re-ran (expect clean scan, no false positive)</code>
- <code>Manual: removed the grandfathered `src/components/UserProfileBar/index.ts` barrel and re-ran (expect stale-entry failure)</code>
- <code>Manual: `mv src/module.tsx src/module-renamed.tsx` and re-ran (entrypoint drift, expect loud failure)</code>
- <code>Manual: temporary probe test calling `findOrphanedModules(buildModuleGraph(), [&#39;module-renamed.tsx&#39;])` over the real graph</code>
- <code>Manual: set `tracking: &#39;by-design&#39;`, then a 4-char `reason`, then `tracking: &#39;TODO later&#39;` on entries in both new allowlists and re-ran the accountability test</code>
- <code>`npm run build` at target, then with `src/validation/import-graph.ts` reverted to base e0649b83, comparing `shasum` of all 56 emitted dist JS bundles</code>
- `grep -rlE "OFF_GRAPH_IMPORTER_ROOTS|APP_ENTRY_ROOTS|offGraphReachable|findOrphanedModules" dist --include='*.js'`
</details>

<details>

- ℹ️ `docs/developer/constants/README.md:16` - Two of the seven grandfathered orphans are still documented as live architecture: docs/developer/components/docs-panel/README.md:133 has a section for `MinimizedSidebarIcon.tsx`, and docs/developer/constants/README.md:16 plus docs/developer/src/README.md:32,75 describe `constants/selectors.ts` and its exported selector groups. The new ratchet records both as having no importer anywhere in src/, and the allowlist reasons at architecture.test.ts:363 and :368 already name the READMEs as the only remaining references. I deliberately did not act: this change grandfathers rather than deletes, the files still exist, and AGENTS.md&#39;s own advice is to drag the doc reference along with the file deletion — so removing these sections now would document the repo as it will be rather than as it is. Proposed follow-up: fold the doc-section removal into the #1923 paydown, so each file deletion lands with its README section in the same commit.
</details>

<details>

✅ No issues found.
</details>

<details>

✅ No issues found.
</details>


tomglenn (dismissed):
## PR #1940 Review: Orphan-Module Ratchet Detection

### Summary

This PR adds file-level orphan-module detection to `architecture.test.ts`, walking forward from `module.tsx` (the plugin's webpack entry) to identify unreachable production files. It splits unreached files into two buckets: genuinely orphaned vs. test/CLI-only ("off-graph reachable"). The implementation is well-designed and integrates cleanly with the existing ratchet machinery, but **one correctness defect blocks approval**.

---

### 🚫 Blocking Issue

**Type-only imports incorrectly prevent orphan detection**

`extractRelativeImports()` includes type-only imports in the module graph, causing files imported **only** via `import type { X } from './foo'` to incorrectly appear reachable.

**Location:** `src/validation/import-graph.ts:388-392`

```typescript
export function extractRelativeImports(content: string, aliasContext?: AliasResolutionContext): string[] {
  return extractImportRecords(content, aliasContext)
    .filter((record) => record.relative)  // ❌ Missing: && !record.typeOnly
    .map((record) => record.specifier);
}
```

**Evidence:** The `typeOnly` flag is correctly detected in `extractImportRecords()` (lines 285-308) and is used by the Node-env reachability scanner (line 899-942), but it's discarded when building the orphan module graph. The PR description states "type-only imports exempt," but this exemption applies only to Node-env checks, not orphan detection.

**Fix:** Add `&& !record.typeOnly` to the filter, or create a dedicated `extractValueImports()` function for orphan graph building.

**Impact:** False negatives—files with only type-only importers will be missed by the orphan ratchet.

---

### ✅ What Works Well

**Correctness (Graph Traversal)**
- `bfsReachable()` correctly terminates with cycles via the `reached` Set guard (line 621-632)
- Two-pass BFS properly partitions unreached nodes: app-unreached + off-graph-reached vs. fully orphaned
- Filtering already-reached nodes before the second BFS is intentional and correct (confirmed by test at line 694-696)

**Entry Point Configuration**
- `APP_ENTRY_ROOTS = ['module.tsx']` is correct—verified against webpack config (`.config/webpack/utils.ts:48-68`)
- Documentation accurately describes this as the plugin's sole webpack entry

**Off-Graph Detection**
- `getOffGraphImportedNodes()` comprehensively collects: test files, `src/cli/`, `src/test-utils/`, and `tests/` at repo root
- `OFF_GRAPH_IMPORTER_ROOTS` correctly maps `EXCLUDED_TOP_LEVEL` + `tests/`

**Baseline Entries**
- All 7 orphaned modules are reasonable: 4 are barrel files consumers deep-import around, 2 are dead code, 1 is docs-only
- All 7 off-graph-reachable modules are legitimately test/CLI-only
- All entries reference #1923 for paydown tracking

**Security**
- Path traversal mitigated by `relative.startsWith('..')` checks (lines 397, 468)
- Test infrastructure properly isolated—cannot leak to production bundle
- No ReDoS patterns in regexes

**Documentation**
- `AGENTS.md` and `.cursor/rules/systemPatterns.mdc` updates accurately describe the new ratchet semantics

**Test Coverage**
- Unit tests with synthetic graphs cover all bucket transitions (`import-graph.test.ts:665-706`)
- Integration tests verify real off-graph imports from test files, CLI, and Playwright tree

---

### 💡 Suggestions (Non-Blocking)

1. **Add a test for type-only orphan detection** once the fix is in:
   ```typescript
   it('marks a file imported only via type-only imports as orphaned', () => { ... });
   ```

2. **Consider caching off-graph imports** like `getAllFileImports()` does for src files. Minor perf consideration, not urgent.

3. **Document the computed-import limitation**: The scanner won't catch `import('./' + name)` or `require.context`. Worth a comment noting this assumption.

---

### Files Changed

| File | Changes |
|------|---------|
| `src/validation/import-graph.ts` | +140 lines: `APP_ENTRY_ROOTS`, `bfsReachable()`, `getOffGraphImportedNodes()`, `findOrphanedModules()` |
| `src/validation/architecture.test.ts` | +150 lines: Orphan ratchet tests + baseline allowlists |
| `src/validation/import-graph.test.ts` | +70 lines: Unit tests for orphan detection |
| `AGENTS.md`, `.cursor/rules/systemPatterns.mdc` | Documentation of orphan reachability semantics |

---

### Verdict: Request Changes

The architecture is sound and the implementation is otherwise correct, but the type-only import defect creates false negatives in the orphan ratchet. Fix `extractRelativeImports()` to filter type-only edges, add a corresponding test, and this is ready to merge.

Jayclifford345 (approved):

Issue #1934, opened 2026-09-17
Two remaining places derive a completion identity independently


tomglenn:
Two remaining places derive a completion identity independently rather than through the shared derivation. Each is a candidate for the same divergence that has already produced four fixed instances: the writer stores a record under one key and the reader looks under another, so completed work reads as not done.

Neither is confirmed broken today, and neither should be changed by guessing at the right default — guessing at an identity is what produced the earlier instances.

## Site 1 — whole-journey completion picks its own fallback source

`src/docs-retrieval/learning-journey-helpers.ts`, inside `markMilestoneDone`'s whole-journey branch: the identity for the `journey_completed` trigger calls `resolveCompletionIdentity` directly with `fallbackSource: 'bundled'`.

A journey whose manifest has an id but no repository is therefore keyed under `bundled` rather than the schema default.

**Why it is not fixed:** journeys have no established correct comparison writer the way guides do (`recordBundledGuideCompletion` / `recordStandaloneGuideCompletion`), so the right default cannot be settled without a real case of a journey manifest with no repository to check against.

**Not silently broken today:** `resetPath`'s guard lift for a whole path already tries all three known sources. Only a narrower reset, which does not exist yet, would expose it.

## Site 2 — a trailing `/content.json` is stripped on one side only

`src/components/docs-panel/hooks/resetGuideProgress.ts`: `fallbackGuideIdFromContentKey` strips a trailing `/content.json` from a bundled content key.

The writer does not. `persistJourneyCompletionPercentage`, inside `setJourneyCompletionPercentage`, derives the guide id as `journeyBaseUrl.replace('bundled:', '')` with no such stripping.

`docs/design/COMPLETION-MODEL.md` decision 9 already records `bundled:<id>` and `bundled:<id>/content.json` as two live launch shapes for the same guide, so a reset landing on the package form for a guide with no manifest can derive a different id than the writer used.

**Why it is not fixed:** how reachable the package form actually is at reset time, as opposed to at initial load, is unconfirmed.

## The durable fix for the class

Stop letting any caller supply a fallback at all: one derivation per content kind, with the defaults inside it. That is already done for milestones and for standalone guides. Journeys are the remaining kind, and doing that is what closes this class rather than patching a fifth instance.

Both open questions above need answering before either site changes, which is why this is labelled as needing a decision.

moxious:
## 🧭 Implementation readiness

[...]

PR #1941, opened 2026-09-17
Fix path reset clearing sibling paths with shared URL prefix

### LINKED ISSUE #1928: Resetting a learning path clears progress for a sibling path whose URL shares its prefix
Resetting a learning path recovers the stored keys it needs to clear by matching a URL prefix. Nothing marks where one identifier ends, so a path whose URL is the start of another path's URL collects and clears the sibling's records too.

## Mechanism

`resetPath` in `src/learning-paths/learning-paths.hook.ts` strips trailing slashes from the path URL and then sweeps three stores with that bare prefix:

```ts
const normalizedUrl = path.url.replace(/\/+$/, '');
const milestoneKeys = Object.keys(completions).filter((key) => key.startsWith(normalizedUrl));
const journeyKeys = [path.url, ...Object.keys(journeyCompletions).filter((k) => k.startsWith(normalizedUrl))];
...
await guideCompletionMarkStorage.clearAllWithPrefix(normalizedUrl);
```

A `startsWith` test has no terminator, so resetting `.../loki` also matches every key under `.../loki-advanced` — interactive progress, journey completion, and the reader's own "mark complete" flags.

## Impact

Silent progress loss for content the reader did not reset. There is no error and nothing to undo.

## Is it reachable today?

Not with the content published in `interactive-tutorials`: of the 73 path packages there, no id is a prefix of another followed by `-`. The nearest collisions (`rca-demo`, `rca-demo-ops`, `rca-demo-v2`) are standalone guides, not paths.

URL-based journeys served from `grafana.com/docs/learning-journeys/` cannot be enumerated from this repo, so that set is unchecked and is the first thing to confirm.

Either way nothing enforces the naming that currently avoids the collision, so a new path can introduce the fault at any time.

## What to look at

The same collision was already fixed for guide progress keys by using a length-prefixed key format. Extending that format here would give one scheme covering both key spaces rather than inventing a second answer. The keys live in `interactiveCompletionStorage`, `journeyCompletionStorage` and `guideCompletionMarkStorage`.

A regression test should reset a path whose URL is a prefix of a sibling's and assert the sibling's records survive.

tomglenn:
# Fix path reset clearing sibling paths with shared URL prefix

## Summary

Closes #1928.

When a user resets progress on a learning path, the app incorrectly clears progress for sibling paths whose URLs share a prefix. For example, resetting `alerting` also clears `alerting-advanced` because the code uses naive `startsWith()` matching with no path-boundary check. This is a data-loss bug—cleared progress cannot be recovered.

This PR adds boundary-aware prefix matching so a key matches only if it equals the prefix exactly OR begins with the prefix followed by `/`. This is the immediate, targeted fix; deeper canonical-identity re-keying is tracked as follow-up in #1934.

### New utility function (`src/lib/storage/key-utils.ts`)
- Added `normalizeForPrefixMatch()` helper to strip trailing slashes
- Added `matchesPrefixOrChild(key, prefix)` function that returns `true` only when:
  - `normalizedKey === normalizedPrefix`, or
  - `normalizedKey.startsWith(normalizedPrefix + '/')`
- Empty prefix matches all keys (preserves intentional bulk-clear behavior)
- Includes comment noting that `?` and `#` delimiters are not handled as path boundaries

### Learning paths hook (`src/learning-paths/learning-paths.hook.ts`)
- Line 539: Changed `milestoneKeys` filter from `key.startsWith(normalizedUrl)` to `matchesPrefixOrChild(key, normalizedUrl)`
- Line 540: Changed `journeyKeys` filter from `k.startsWith(normalizedUrl)` to `matchesPrefixOrChild(k, normalizedUrl)`

### User storage (`src/lib/user-storage.ts`)
- Line 1500: Changed `clearAllWithPrefix` filter from `parsed.contentKey.startsWith(contentKeyPrefix)` to `matchesPrefixOrChild(parsed.contentKey, contentKeyPrefix)`

### Unit tests (`src/lib/storage/key-utils.test.ts`)
- Added 12 test cases for `matchesPrefixOrChild` covering:
  - Exact matches with and without trailing slashes
  - Direct and deeply nested child paths
  - Sibling paths with shared prefix substrings (`/alerting` vs `/alerting-advanced`, `/drilldown` vs `/drilldown-logs`)
  - Full URL keys with domains
  - Bundled content keys

### Regression tests (`src/learning-paths/reset-path-completion.test.ts`)
- Added test suite "resetPath — sibling path isolation (regression for #1928)" with 4 tests:
  - Verifies resetting `/alerting` does NOT clear `/alerting-advanced` across all three storage namespaces
  - Verifies legitimate child paths like `/alerting/milestone-1/` ARE still cleared
  - Verifies keys stored with and without trailing slashes are handled consistently
  - Covers interactive completions, journey completions, and guide completion marks

## Testing

Ran locally:
- `npm run typecheck` — passed (0 errors)
- `npm run lint` — passed
- `npm test -- src/lib/storage/key-utils.test.ts` — 22 tests passed
- `npm test -- src/learning-paths/reset-path-completion.test.ts` — 14 tests passed

## Risks / Unresolved

- **Query strings and hash fragments**: The fix treats only `/` as a path boundary. If keys ever contain `?` or `#` as child delimiters, they would not match. This is documented in a code comment; no evidence that such keys exist in practice.
- **Root path edge case**: A prefix of `/` normalizes to empty string, which matches everything. Verified that `path.url` is always a full HTTPS URL, never just `/`, so this is not reachable.
- **Deeper re-keying**: This fix addresses the symptom (prefix collision) but not the root cause (keys derived from volatile URLs). Canonical-identity keying is follow-up work tracked in #1934.


moxious (changes requested):
Blockers:

1. [blocking] **B1 — An undeclared whole-tree package-lock.json regeneration rides into this bugfix, moves 282 package versions, and silently changes shipped terminal behaviour** (high · build-and-ci)
   `package.json` has no diff, yet `package-lock.json` accounts for 4033 of the PR's 4269 changed lines: 282 package versions move, 18 packages are added and 75 removed. It is not a formatting artifact and it is avoidable — replaying `npm install --package-lock-only` on the base `package.json` and base lockfile leaves the file byte-identical under both npm 11.18.0 and the declared `npm@12.0.2`. Commit 7895882f fuses the whole regeneration into one commit titled "Fix lint errors", and the PR body's Changes section names neither `package-lock.json` nor the two source files the bump forced. The lockfile is the artifact contract: CI runs `npm ci` at seven call sites keyed on `hashFiles('package-lock.json')`, and neither `js-yaml` nor `@grafana/coda-client` is a webpack external, so both ship — the PR's own bundle-size check reports the total bundle up 42.47 KB across 130 files with every chunk hash rotated. Two consequences nobody reviewed: `@grafana-coda/sandbox-client` goes 1.2.2 → 1.11.0 just 1h38m after 1.11.0 was published, against the 14-day `minimumReleaseAge` that `renovate.json`'s last rule names that exact package and alias for with a written justification for withholding the `@grafana` waiver; and that bump re-codes the terminal's client-side idle and channel-down timers from `terminal_disconnected` to `stream_stalled`, so an idle or dropped terminal loses Pathfinder's bespoke "Connect again to start a new session" sentence and shows the raw SDK string instead. `CONTRIBUTING.md` asks for one concern per PR and roughly 500 lines. Two things cut the other way and belong in the decision: the refresh is what clears the base lockfile's advisories (`npm audit --omit=dev` goes from 26 findings, including one high `@tiptap/core` prototype-pollution and ReDoS advisory, to 0), and it is cleanly separable — with the base install, head source passes `tsc --noEmit`, `eslint` over the whole repo, and all 577 jest suites and 9552 tests.
   Required: Revert `package-lock.json` on this branch so the PR carries only the storage fix, and raise the refresh as its own dependency PR naming the advisories it closes and reviewing the coda-client 1.2.2 → 1.11.0 behaviour delta, including the `stream_stalled` re-coding and the stale claim at `useTerminalLive.hook.ts:18` that the drop is covered by the idle timer's `terminal_disconnected`. Dropping the lockfile also removes the need for the `coda-api.ts` edit, since `switch-exhaustiveness-check` only fires against 1.11.0's widened union. If the refresh must stay, add a Dependencies section to the PR body stating the bumps, the reason for regenerating an unchanged `package.json`, the audit delta, and an explicit waiver of the Renovate age gate on the coda client.

Follow-ups:

1. [follow_up] **F1 — A second key-boundary scheme lands on a judgment call issue #1928 explicitly held open, with no declared relationship to the recorded one** (medium · cross-cutting-architecture)
   `src/lib/storage/progress-keys.ts` already records a boundary contract for this bug class — `{prefix}{len(contentKey)}:{contentKey}:{sectionId}`, whose header says "The character count in front of the content key is what makes it safe" and cites the very collision the new test asserts, `bundled:welcome-to-grafana` versus `bundled:welcome-to-grafana-cloud`. Issue #1928's body asks for "one scheme covering both key spaces rather than inventing a second answer", and its readiness comment records the choice between extending that format and adding a local delimiter check as open judgment call 1, verdict "Needs a decision on (1) first"; the `agent/needs-decision` label is still on the issue with no unlabel event. No decision of record settles it: #1928 has no decision comment, `COMPLETION-MODEL.md`'s nine decisions cover no key boundary, and `CONCERN_DETAILS.md` records no anchor for storage-key boundaries. Verification added a fair qualification the PR should state rather than hide: the two schemes are not drop-in substitutes here, because two of the three sweeps are bounded-record maps with no composite key to length-delimit and the third already recovers `contentKey` via `parseVersionedStorageKey`, so the delimiter answer is a defensible choice — just an undeclared and unrecorded one.
   Follow-up: Say in the PR body whether this follows, extends, or replaces the `progress-keys.ts` boundary contract and why the delimiter answer was chosen, then record that decision on #1928 and drop its `agent/needs-decision` label. File the durable re-keying question against #1934's successor if it is still wanted.

2. [follow_up] **F2 — The unforced loadAll predicate rewrite inverts Grot guide document selection for a falsy welcome or screens** (medium · build-and-ci)
   `convertYamlToBlock`'s document predicate moved from truthiness, `doc.welcome || doc.screens`, to `welcome !== undefined || screens !== undefined`. Reproduced on both js-yaml 5.2.2 and 5.4.2, so it is the rewrite and not the API swap: for a two-document stream whose first document declares `welcome: null`, `welcome: ''`, `welcome: false`, `screens: false` or `screens: 0`, base selects the real guide and head selects the first document, which then throws `Missing "welcome" field in YAML` naming the wrong document. Verification found the change not reachable for the real corpus — the authoritative Grot guide format carries `meta`, `header`, `welcome` and `screens` together in one file's front matter, so the guide is always the first document and every realistic shape selects identically under both predicates — and the failure is author-time, caught into an inline Alert beside the paste textarea with nothing persisted. It is also unforced: `@typescript-eslint/no-deprecated` is severity 1 with no `--max-warnings` anywhere, and the deprecated iterator overload still works at runtime in 5.4.2, so the commit title "Fix lint errors" does not apply to this half. Alongside it, `(doc): doc is any` is a widening pseudo-guard that turns off checking for every later field read in the function, and it leaves the `typeof parsed !== 'object'` half of the next guard unreachable.
   Follow-up: Restore truthiness semantics under the array form, assert a real shape such as `doc is Record<string, unknown>` instead of `any`, drop the two `as any` casts, and add a test with a first document carrying `welcome: null` that asserts the real guide is still selected. If the `!== undefined` semantics are actually wanted, make that its own change with the error message naming the document chosen.

Suggestions & nits:

1. [suggestion] **S1 — The one unintended behavioural delta, query and fragment keys no longer matching, is documented as a caveat with no test and no caller handling it** (medium · testing-and-verification)
   `matchesPrefixOrChild`'s own doc comment names the gap: "Query strings (`?`) and hash fragments (`#`) are not handled as delimiters—if keys can contain these, callers must handle them separately." The reset call site is exactly such a caller and does not handle them, and neither new test file contains a single `?` or `#` case across their 26 and 3 added cases. Verification established that no record is written under such a key today, but only incidentally: safety rests on `getContentUrls` stripping `?` and `#` when it rebuilds the content ladder URL and on the cover-page footer gate, not on anything the matcher or the call site asserts.
   Suggested: Add characterization cases for `matchesPrefixOrChild('/alerting?x=1', '/alerting')` and `('/alerting#top', '/alerting')` asserting whatever behaviour you decide, so the decision is pinned by a test rather than inherited from two unrelated modules.

2. [suggestion] **S2 — The PR body calls the empty-prefix branch unreachable, but it is load-bearing for the global reset** (low · reversibility-and-one-way-door)
   `clearAllWithPrefix(contentKeyPrefix = '')` defaults to the empty string, which routes straight into the helper's `if (normalizedPrefix === '') { return true; }`, and `MyLearningTab.tsx:356` calls it with no argument inside the "Reset all learning progress?" confirm. The PR body's premise is right — `path.url` is always a full HTTPS URL from `paths.json` or `paths-cloud.json`, and App Platform paths carry no `url` — but its conclusion is not: the branch is reached from the other caller, and a later cleanup that trusts the PR body and deletes it would make the global reset silently stop clearing guide completion marks.
   Suggested: Correct the PR body to say the empty-prefix branch is required by `clearAllWithPrefix()`'s default argument. No code change needed.

3. [suggestion] **S3 — The regression test's path fixture is a population that cannot occur, which overstates what it verifies** (low · testing-and-verification)
   The new suite seeds a `learning-journeys/alerting` path with an `alerting-advanced` sibling, but every real `path.url` is one of six `/docs/learning-paths/` entries in `paths-cloud.json` and none is a prefix of another, so a curated-path-versus-curated-path collision cannot arise today. The reachable form of #1928 is a stray docs content key that happens to extend a path slug, which no test covers. The test does still pin the fix — under the base implementation the sibling-preserved assertions fail — it just proves the helper's semantics rather than a live collision.
   Suggested: Use a real `paths-cloud.json` url in at least one case with a seeded `.../drilldown-logs-extra/...` content key, and note in the PR body that the curated-path collision is prophylactic.

4. [suggestion] **S4 — A prerelease, geotiff 3.1.0-beta.0, and six new transitives enter the production closure as a side effect of the refresh** (low · security)
   `ol` moves 10.7.0 → 10.10.0, which pins `geotiff ^3.1.0-beta.0`, making geotiff the one newly prerelease production entry and pulling in `zarrita`, `@zarrita/storage`, `numcodecs`, `fflate`, `unzipit` and `reference-spec-reader`. Shipped-artifact exposure is nil because the whole subtree hangs off webpack-externalized `@grafana/*` packages, and `npm audit --omit=dev` at head reports 0 findings, but `npm ci --omit=dev` now installs unreleased code and six packages join the closure that CI, SBOM and any future audit gate must cover, none of it chosen.
   Suggested: Reverting `package-lock.json` per B1 keeps geotiff at 2.1.3 and the six transitives out. If the refresh is kept, note the beta in the PR body so a later SBOM reviewer sees it was accepted knowingly and inherited from `ol@10.10.0`.

5. [suggestion] **S5 — Three codes the SDK documents as needing distinct handling are swept to the generic fallback, untested** (low · coda-terminal)
   `too_many_sessions`, `host_identity_unverified` and `recovery_exhausted` are enumerated to `return fallback`. The SDK says the five-terminal cap never clears by waiting and the only remedy is closing one; that host-key failure must read as a security failure stating no shell was opened and must never offer a reconnect; and that an exhausted recovery should offer to start again rather than report the sandbox broken. Verification confirmed this is not a regression — base already reached `default: return fallback` for all of them, and which code arrives is the deployed backend's choice — so it is a standing guidance gap rather than something this PR broke. But enumerating a code as a case label asserts its handling was considered, and 19 labels were added in one sweep with no test for any of them.
   Suggested: Either give at least those three their own sentences with a table-driven test per code, or file the gap as its own issue rather than leaving 19 labels that imply a decision nobody made.

6. [suggestion] **S6 — The new case arm duplicates the fallback group beside it and carries a comment restating the line below** (low · coda-terminal)
   Lines 145-165 are 19 cases returning `fallback`, immediately followed by a pre-existing run of 18 cases returning `fallback` — two adjacent groups with identical bodies. The added `// These error codes fall through to the fallback message` sits directly above `return fallback`, which is the QC8 restates-the-code shape; the neighbouring kept comments each carry a non-obvious reason instead. `assertExhaustive` cannot apply here, because `CodaErrorCode` ends in `| (string & {})` so `code` in `default` is `string` rather than `never`, and the enumerate-plus-default shape the PR follows is the local convention.
   Suggested: If the lockfile stays and this hunk is kept, fold the 19 labels into the existing fallback group and delete the comment.

7. [nit] **N1 — An unrelated rewrite downgrades two eslint-disable justifications into duplicated prose** (low · correctness-and-reliability)
   Both sites carried one compliant line whose `--` justification held the full explanation. Head splits each into two lines of free-standing prose above a shortened `-- revision re-identifies callback`, so the explanation is stated twice and the machine-checked justification is now the weaker copy. Nothing forced it: Prettier does not reflow comments, there is no `max-len` rule, and `prettier --check` passes on the base form.
   Suggested: Revert both comment blocks to their base form.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1941
Summary: Fix path reset clearing sibling paths with shared URL prefix
Verdict: Request Changes
Results: 1 blocker, 7 non-blocking findings, 2 follow-ups

moxious (approved):

Follow-ups:

1. [follow_up] **base-tiptap-core-advisory — Track the existing high-severity Tiptap dependency advisory separately** (high · security)
   package-lock.json:5498-5500 pins @tiptap/core 3.29.1; base and head lockfile blobs are identical. npm audit --omit=dev --ignore-scripts --json reports 1 high and 25 moderate production vulnerabilities, with high GHSA-j95f-988m-3j2f affecting @tiptap/core >=3.7.0 <3.30.5; fixAvailable is true. src/components/block-editor/forms/MarkdownBlockForm.tsx:23-26 imports Tiptap React, StarterKit and Markdown; .config/bundler/externals.ts does not externalize Tiptap. The existing production dependency remains within the reported ReDoS advisory range. This is a base dependency issue, not a regression introduced by this reset fix; exploitability was not reproduced.
   Follow-up: Track remediation in a separate dependency change through Grafana security reporting or existing private vulnerability tracking, with the advisory and installed version. Do not restore the broad lockfile refresh in this reset fix.

2. [follow_up] **description-omits-yaml-refactor — Describe the retained YAML importer change in the final PR scope** (medium · cross-cutting-architecture)
   src/components/block-editor/forms/GrotGuideBlockForm.tsx:24-41 changes the loadAll overload, document narrowing, and welcome typing against base. src/components/block-editor/forms/GrotGuideBlockForm.test.tsx:88-129 adds an importer regression test. The current PR description lists only storage and path-reset changes; it contains no YAML, loadAll, or GrotGuideBlockForm mention. CONTRIBUTING.md:27-38 requires focused changes, and :48 requires a description of what changed and why. The final review record does not describe all source changes being accepted. The restored truthiness is correct, but the remaining importer refactor is separate from path reset and needs an explicit rationale or removal.
   Follow-up: Before merge, add a short description of the retained YAML importer refactor, its compatibility intent, and its new test; alternatively remove those two unrelated file changes and handle them separately.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1941
Summary: Fix path reset clearing sibling paths with shared URL prefix
Verdict: Approve with Minor
Results: 0 blockers, 0 non-blocking findings, 2 follow-ups

tomglenn:
Handled already by #1950

PR #1942, opened 2026-09-17
Fix guided tour progress bar never reaching 100%

### LINKED ISSUE #1935: Guided tour progress bar can never reach 100%, and reads 0% for a single-step block
The progress bar in a guided tour's overlay can never show all steps complete. It tops out one step short, and a guided block with a single step reads 0% for its whole life.

## Mechanism

Each step paints its own comment box. `guided-handler.ts` builds the step's progress snapshot when the box is created:

```ts
const stepInfo: CommentBoxStepInfo = {
  current: stepIndex,
  total: totalSteps,
  completedSteps: [...this.completedSteps], // Copy to avoid mutations
  progress: 'performed',
};
```

The snapshot is taken before the step it belongs to has completed, so the box for step N shows N-1 steps done. When step N settles, its box is destroyed and the next step's box is created with an updated snapshot — which is how the bar advances at all.

On the final step there is no next box. The tour ends and nothing ever paints with every step complete, so the bar stops at `(total - 1) / total`. With one step there is no second paint at all, so it shows 0% throughout.

Reordering the progress update above the cleanup does not fix the symptom: no paint happens between the two.

## Impact

Cosmetic, and the bar is truthful at every paint it actually makes — it is never showing a wrong number, only never showing the last one. But a tour that visibly stops short of finished undercuts the reader's sense that they completed it.

## Why this is not a small fix

Closing it needs a repaint after the final step settles, which means the comment box outlives the tour. That raises product questions that should be settled before coding:

1. How long does the box linger after the last step? A fixed dwell, or until the reader dismisses it?
2. Do the cancel and timeout outcomes repaint too? A cancelled tour showing a full bar would be wrong; showing a stale one is the current behaviour.
3. Does the final box keep its controls (skip, cancel) once there is nothing left to act on?

## What to look at

`showCommentBoxForStep` in `src/interactive-engine/action-handlers/guided-handler.ts` and the cleanup paths that destroy the box on `completed`, `cancelled` and skip. The existing limitation is recorded at the contract anchor for this behaviour.

tomglenn:
# Fix guided tour progress bar never reaching 100%

## Summary

Fixes #1935: The guided tour progress bar could never display 100% and single-step tours showed 0% for their entire duration.

The root cause was in `finishGuidedStep()`, which destroyed the comment box (via `cleanupListeners(true)`) before pushing the completed step index onto `completedSteps`. Since progress is calculated as `completedSteps.length / total * 100`, the bar topped out at `(total-1)/total`. For a single-step tour, this meant 0% for its whole life.

### `src/interactive-engine/action-handlers/guided-handler.ts`
- Reordered operations in `finishGuidedStep()`: now pushes to `completedSteps` **before** calling `cleanupListeners(true)`
- Modified `createGuidedStepArbiter()` to accept `stepIndex` and `totalSteps` parameters
- Added final-step completion logic: on successful completion of the last step, updates progress bar to 100% via DOM manipulation and holds it visible for ~600ms before cleanup
- Added `updateProgressBarTo100()` helper method for direct DOM manipulation
- The delay is cancellable—if aborted during the delay, cleanup proceeds immediately

### `src/constants/interactive-config.ts`
- Added `progressBarCompletionDelayMs: 600` to the guided config section

### `src/interactive-engine/comment-box.contract.test.ts`
- Updated existing test comments to reflect that 100% is now reachable
- Added test: "reaches 100% on final step before cleanup"
- Added test: "single-step tour shows 100% on completion"

### `src/interactive-engine/action-handlers/guided-handler.test.ts`
- Added "Progress bar completion delay behavior" test suite with 8 tests:
  - Final step sets progress bar to 100% before cleanup
  - Cleanup delays ~600ms on final step
  - Cancel during delay resolves immediately
  - Single-step tour shows 100%
  - Non-final steps have no delay
  - Skipped final step shows 100% briefly
  - Error results skip the 100% display
  - Missing progress bar handled gracefully

### `docs/design/CONCERN_DETAILS.md`
- Updated the `interactive-engine` contract to document the new behavior, delay mechanism, and single-step semantics

## Design decision: single-step tours

**Reviewers should confirm:** For a one-step guided tour, the implementation defaults to showing 100% once the action is performed. This treats completing the only step as completing the tour.

**Alternatives considered:**
- Show 50% while in progress (0% → 50% → cleanup feels jarring)
- Hide the progress bar entirely for `total === 1`

The current approach uses the same code path as multi-step final steps, keeping the logic unified. The single-step handling is isolated and easy to adjust if product feedback differs.

## Testing

- `npx tsc --noEmit` — no type errors
- `npx eslint --cache .` — no lint errors
- `npm test -- --testPathPattern="comment-box.contract"` — 41/41 passed
- `npm test -- --testPathPattern="guided-handler"` — 26/26 passed
- `npm test -- --testPathPattern="navigation-manager"` — 32/32 passed

## Risks / Unresolved

- **Noop steps**: These use `showNoopCommentBox()` which creates step checkboxes, not a progress bar. The 100% repaint only applies to regular guided steps with highlighted targets. This is documented behavior, not a regression.
- **Already-satisfied targetState**: If the target state is satisfied before the comment box renders, the progress bar may not exist when `updateProgressBarTo100()` is called. This edge case is handled gracefully (the method returns false and skips the delay).


moxious (changes requested):
Blockers:

1. [blocking] **contract-branching-second-progress-bar-writer — Second, formula-bypassing writer of the comment box progress bar** (high · interactive-engine)
   `updateProgressBarTo100` (guided-handler.ts:1188-1196) writes `style.width = '100%'` straight to the first `.interactive-comment-progress-bar` in the document, with no `stepInfo`, no `progress` discriminator and no call to `progressBarPercent`. `navigation-manager.ts` has zero changed lines, so that element now has two production writers. The anchor sentence naming `CommentBoxStepInfo.progress` the sole owner of the bar's formula was left byte-identical, and the base row's own forecast of the correct fix — a repaint after the final settle rather than a change to the formula, restated in #1841's body — was deleted rather than satisfied. The selector is also owner-blind: BubbleTour.tsx:73 drives the same box with `progress: 'position'`, which the owner computes differently; both skeptics confirmed that is a latent property of the selector rather than a collision reachable today.
   Required: Prefer the shape the anchor and #1841 already name: one more repaint through `highlightWithComment` with a `stepInfo` whose `completedSteps` includes the final step. If the direct write is kept deliberately, scope the query to the comment box this step created and amend the sole-owner sentence to declare the exception, so the anchor and the code agree.

2. [blocking] **contract-anchor-mechanism-false — Rewritten contract anchor states a mechanism the code does not implement** (high · cross-cutting-architecture)
   The new anchor clause says `finishGuidedStep` pushing onto `completedSteps` before destroying the box is what makes the final step's completion reflected in the progress calculation. No calculation re-runs. The width is written once per box at navigation-manager.ts:964-965 from the snapshot taken at guided-handler.ts:615 when that box is built; the push and `cleanupListeners(true)` are adjacent synchronous statements; the only `completedSteps` readers (:367, :615) run on the next step's box under base and head ordering alike. Two skeptics searched position tracking, drift detection and the ResizeObserver and found no later write to the bar's width. The real mechanism is the direct DOM write inside `settle`, which runs strictly before `finishGuidedStep` is entered. Issue #1935's body states the same conclusion. The PR also deletes a sentence that is still true, that the contract test's 100% case guards the formula and not a reachable paint.
   Required: Rewrite the clause to name the real mechanism — `updateProgressBarTo100` inside `createGuidedStepArbiter().settle` plus the 600ms dwell from `INTERACTIVE_CONFIG.guided.progressBarCompletionDelayMs`, with `progressBarPercent` still topping out at (total-1)/total for the formula path — and restore the sentence explaining what the contract test's 100% case actually guards.

3. [blocking] **post-settle-click-window-rearms-target — 600ms hold leaves the capture click listener armed, re-firing element.click() on a completed final step** (medium · interactive-engine)
   `complete` is now async and reaches `cleanup` only after `await arbiter.settle(...)`, but `isResolved` is set exclusively inside `cleanup` (guided-handler.ts:866-881), so it stays false for the whole hold. The capture-phase document listener is removed only by `cleanupListeners` at :267, which `runGuidedStep` reaches after the arbiter promise resolves. The reachable sequence, as narrowed by two independent skeptics: the first click lands on the target, then a second click within 600ms lands outside the target but inside the 16px ring, and the `isWithinBounds` branch calls `element.click()` again on the application control. The reader's Cancel is simultaneously swallowed, since `settle('cancelled')` is a no-op on a settled arbiter. At base `complete` was synchronous, so the window was one statement wide. The 100ms connectivity interval closes it early when the click detaches the target, and the claim that the overlay's own buttons sit inside the ring was refuted — the box's near edge is 20px out against a ring ending at 16px.
   Required: Flip the `waitForClick` guard synchronously before awaiting: set `isResolved` (or a separate `completing` flag that `handleClick` checks) at the top of `complete()`, and let `cleanup` tolerate being called with the guard already set — or remove the document click listener before awaiting the delay.

4. [blocking] **panel-bar-diverges-from-overlay-bar — Overlay bar forced to 100% while the paired panel bar stays at (n-1)/n** (medium · cross-cutting-architecture)
   The anchor sentence that the overlay bar and `.interactive-guided-progress-fill` measure the same quantity and must not be changed independently is retained byte-identical, but only the overlay changed. `interactive-guided.tsx:972-975` paints `(currentStepIndex / internalActions.length) * 100`; `setCurrentStepIndex(i)` is set before the await and never advanced past the last index, and `setCurrentStepStatus('completed')` runs only after the await, so for the whole hold the panel renders the waiting-instruction arm at (n-1)/n — 0% for a single-step block — beside a 100% overlay. Both skeptics checked the render gating specifically: `deriveGuidedUiState` returns EXECUTING while `isExecuting`, which is cleared only in the `finally` after the loop, so the panel bar is on screen throughout. `interactive-guided.tsx` is not touched by this PR, and the same position formula also stands in interactive-multi-step.tsx and interactive-section.tsx.
   Required: Make the joint decision the anchor requires: either advance the panel fill for the same window from the same terminal signal, or amend the must-not-be-changed-independently sentence to state why the two surfaces are now allowed to disagree during teardown. Do not leave that sentence unamended.

5. [blocking] **needs-decision-answered-implicitly — Three product questions #1935 reserves for a human are decided silently in the diff** (medium · cross-cutting-architecture)
   #1935 is open with `needs-design` and `agent/needs-decision`; its only comment is the prep-issues v3 marker recording verdict needs-decision, and neither the issue, the PR body nor any PR comment records a human decision. Its body names three questions it says should be settled before coding: how long the box lingers (fixed dwell or explicit dismiss), whether cancel and timeout outcomes repaint, and whether the final box keeps its skip and cancel controls. The diff answers all three — a fixed 600ms dwell at interactive-config.ts:147, repaint gated to completed/skipped at guided-handler.ts:133-134, and controls left untouched, which is what makes the click re-entry above reachable — while the source itself concedes the question is open with `// Reviewers should confirm this is the desired behavior.` at guided-handler.ts:127-132. The issue also states, correctly, that reordering the progress update above the cleanup does not fix the symptom.
   Required: Get the three questions answered on #1935 or in the PR body and record the outcome in the interactive-engine anchor, then remove the "Reviewers should confirm" comment from source. If the answers are not available yet, narrow the change so it does not implicitly settle them.

Follow-ups:

1. [follow_up] **completedsteps-reorder-inert-and-miscommented — The finishGuidedStep reorder is observationally inert and its comment asserts an effect it does not have** (medium · correctness-and-reliability)
   `finishGuidedStep` is fully synchronous, and every abort handler reachable from its `abort()` early-returns on an already-settled arbiter, so no observer can see the intermediate state; the only `completedSteps` readers run when the next step's box is built. The reorder therefore changes no painted bar, step dot, or noop step-list item, yet it carries the comment "Push to completedSteps FIRST before cleanup so the progress bar can reflect the completion" and is named in the PR body as the root-cause fix. A skeptic walked all six `completedSteps` assertions in the test file and confirmed none would fail if the reorder were reverted.
   Follow-up: Drop the reorder and its comment — the behaviour stands on `updateProgressBarTo100` alone — or keep it and delete the comment plus the root-cause claim in the PR body, so no comment asserts an effect the code does not have.

2. [follow_up] **contract-test-additions-are-duplicates — New comment-box contract cases pass unchanged on base and replace an accurate comment** (medium · testing-and-verification)
   `renderBar` drives `navigationManager.highlightWithComment` with a hand-built `CommentBoxStepInfo` and never touches `GuidedHandler`, and the formula it reaches is byte-identical at base, so both new cases would pass at base. They duplicate the n/n assertion in the pre-existing test immediately above, and one concedes in its own comment that it verifies what that test already verified. The accurate comment they displaced — that the 100% case guards the formula, not a state a live guided tour can paint — was the one place telling a reader what this file's scope is.
   Follow-up: Drop the two duplicate cases and keep the real coverage in `guided-handler.test.ts`, then restore the scope-limiting comment on the surviving 100% test, adjusted to say the live 100% paint comes from `GuidedHandler`'s post-settle write and is covered there.

3. [follow_up] **progress-bar-selector-coupling-unpinned — Nothing pins the progress-bar class coupling, so a rename degrades silently** (medium · testing-and-verification)
   The new describe block replaces `highlightWithComment` wholesale and hand-builds its own `div.interactive-comment-progress-bar`, so `NavigationManager.createCommentBox` never runs in any of the eight new tests; the file that does run it never involves `GuidedHandler`. The class literal has no shared constant across its producer, its new consumer and the stylesheet, and `updateProgressBarTo100` returns false on a miss with no logger call, so the caller skips both the write and the hold with no signal. Rename the bar in `createCommentBox` and the feature disappears while both suites stay green — the one test covering the miss path asserts only that nothing crashes.
   Follow-up: Export the class name as a shared constant used by both `createCommentBox` and `updateProgressBarTo100`, or add one test that drives `executeGuidedStep` against a real NavigationManager-rendered comment box and asserts the bar reaches 100%. The constant is the cheaper of the two.

Suggestions & nits:

1. [suggestion] **qc8-design-decision-comment-block — QC8: a six-line reviewer-addressed comment block and four restating comments ship in source** (medium · correctness-and-reliability)
   guided-handler.ts:127-132 carries "DESIGN DECISION", "Reviewers should confirm this is the desired behavior." and two rejected alternatives; :137, :143, :148 and :258 each restate the line beneath them, and :258 restates it incorrectly; :1182-1187 is a five-line JSDoc on a six-line private helper whose sibling `cleanupListeners` carries none. The same unresolved framing is duplicated into a contract-test comment. AGENTS.md keeps only counterintuitive code, hidden invariants, linked external-bug workarounds and correctness warnings, and says to restructure rather than write a comment that will not fit on one short line.
   Suggested: Move the single-step decision and its alternatives into the PR body and the anchor, delete the four restating comments and the helper JSDoc, and if the single-step semantics need a note reduce it to one short line stating the settled invariant rather than the open question.

2. [suggestion] **hundred-percent-paint-misses-other-final-step-successes — Four other final-step success paths still leave the bar at (n-1)/n, including a final noop step** (low · interactive-engine)
   The skippable element-not-found path returns `finishGuidedStep('skipped', ...)` and bypasses `settle` entirely; the already-satisfied `targetState` path and the non-drivable-action path both settle before any box exists, so the query finds nothing; and `showNoopCommentBox` builds a step list rather than a progress bar, so a final noop step can never paint 100%. Only a driven click and the final-step skip button actually fill the bar. A guide ending in a noop "you are done" step — a common authoring shape — still shows (n-1)/n while the anchor now claims the bar can reach 100%.
   Suggested: Decide and record the intended scope: paint once from a single place after the terminal result is known so every successful final-step path is covered, or state in the anchor that only the driven-click and skip paths fill the bar and that a final noop step has no progress bar at all.

3. [suggestion] **final-step-persistence-window-widened — For the default completeEarly false shape, completion persists about 600ms later than the final action** (low · correctness-and-reliability)
   `beforeSettle?.()` still runs before the delay, so the anchor's persistence-during-capture invariant survives when `completeEarly` is set. But `completeBeforeActionEffect` is passed only in that case, and `interactive-step.tsx`, `interactive-multi-step.tsx` and the JSON parser all default `completeEarly` to false, so in the common shape `completeStep()` runs only after the loop and the post-action window before `persistCompletion` widens from roughly 500ms to roughly 1100ms. A final step whose click navigates or unmounts the panel has correspondingly longer to lose the record; `cancel()` on unmount aborts the delay, so the realistic exposure is small.
   Suggested: No code change is required if this is accepted, but state the trade-off in the PR body and in the anchor: the 600ms sits before the result reaches `InteractiveGuided`, so it precedes persistence in the default shape.

4. [suggestion] **cancel-during-delay-test-does-not-discriminate — The cancel-during-delay test cannot fail for the reason it names** (low · testing-and-verification)
   It asserts only that the result is 'completed' and that `clearAllHighlights` was called, but `cancel()` itself calls `cleanupListeners(true)` which calls `clearAllHighlights()`, so that assertion is satisfied by the test's own action. Because a false return from `updateProgressBarTo100` skips the delay entirely, the same test passes with the new code stubbed to a no-op and on the pre-PR code where there is no delay to cancel. The abort listener is covered only indirectly: removing it would hang the awaited promise and the test would time out rather than assert.
   Suggested: Assert the discriminating facts: that the width was set to 100% before the cancel, and that the completion promise settles without advancing 600ms — for example by capturing resolution with a flag after `advanceTimersByTimeAsync(0)` and checking `jest.getTimerCount()` shows the delay timer gone.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1942
Summary: Fix guided tour progress bar never reaching 100%
Verdict: Request Changes
Results: 5 blockers, 4 non-blocking findings, 3 follow-ups

moxious (changes requested):
Blockers:

1. [blocking] **post-settle-click-window-rearms-target — Skipping the final step leaves the skipped control activatable during the hold** (high · interactive-engine)
   `src/interactive-engine/action-handlers/guided-handler.ts:732-736,877-925` and `guided-handler.test.ts:738-800`. After a reader skips the final highlighted action, a subsequent click outside its target but inside the 16px highlight padding still invokes element.click() on that skipped Grafana control. The new completing flag is set only by complete(), so the skip winner leaves it false throughout the hold. Base resolved the arbiter immediately and removed the listener before another user event; head creates a 600ms activation window.
   Required: Guard the click handler against any terminal arbiter result, or disarm the action listener when any outcome wins, before awaiting the visual hold. Add a fake-timer regression that skips the last step, clicks inside the padding during the hold, and asserts zero target activations while the result remains skipped.

2. [blocking] **panel-bar-diverges-from-overlay-bar — The new completion hold shows conflicting progress for the same tour** (medium · cross-cutting-architecture)
   `src/interactive-engine/action-handlers/guided-handler.ts:132-155`, `src/components/interactive-tutorial/interactive-guided.tsx:407-426,944-979`, and `docs/design/CONCERN_DETAILS.md:63`. For a one-step tour, the new overlay shows 100% for 600ms while the visible sidebar still shows 0% and asks the reader to perform the same action. For n steps the sidebar remains (n-1)/n. Both bars agreed before this change. The revised anchor explicitly permits this new disagreement, which resolves the old documentation mismatch but preserves the user-visible defect and adds deferred synchronization work.
   Required: Drive the panel and overlay completion display from the same terminal success signal for this hold, and cover a single-step tour with both surfaces visible. Keep the contract aligned with the resulting behavior instead of deferring the paired display fix.

Follow-ups:

1. [follow_up] **completedsteps-reorder-inert-and-miscommented — The PR summary still attributes the fix to an inert reorder** (medium · correctness-and-reliability)
   PR body Summary; `src/interactive-engine/action-handlers/guided-handler.ts:132-159,254-267,608-612` and `src/interactive-engine/navigation-manager.ts:966`. The code comment and contract now name the real mechanism, but the PR summary still gives reviewers the wrong causal explanation. Moving the push above cleanup cannot repaint the already-created bar; the actual behavior change is the direct terminal paint and extended overlay lifetime.
   Follow-up: Correct the PR Summary in this revision to describe the missing final repaint and the terminal paint/600ms hold as the fix. Describe the reorder as incidental or remove it; no code change is needed to correct the explanation.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1942
Summary: Fix guided tour progress bar never reaching 100%
Verdict: Request Changes
Results: 2 blockers, 0 non-blocking findings, 1 follow-up

PR #1948, opened 2026-09-18
fix(docs-retrieval): fall back to milestone url when slug is empty

### LINKED ISSUE #1932: Dead fallback when deriving a milestone id from its URL
A fallback that looks like it protects against a missing milestone id can never run, because the function it guards never returns a nullish value.

## Location

`src/docs-retrieval/learning-journey-helpers.ts`:

```ts
const members: PathMember[] = unlocked.map((m) => ({ id: getMilestoneSlug(m.url) ?? m.url, url: m.url }));
```

`getMilestoneSlug` in `src/lib/learning-journey-url.ts` is typed `string` and ends with an empty-string fallback:

```ts
export function getMilestoneSlug(url: string): string {
  ...
  return withoutContentFile.replace(/\/+$/, '').split('/').pop() || '';
}
```

It returns `''`, never `null` or `undefined`, so the `?? m.url` arm is unreachable.

## Impact

Small and bounded. A milestone whose URL yields no slug gets an id of `''` rather than falling back to its URL. It still scores from its persisted record, so nothing visible breaks. The real cost is that the code reads as if a fallback protects this and it does not, which is how someone later builds on a guarantee that was never there.

## Fix shape

Pick one and make the intent explicit:

- Drop the dead arm, if an empty id is genuinely acceptable here.
- Or make the fallback real: `getMilestoneSlug(m.url) || m.url`.

Note other call sites in the same file already treat an empty slug as the falsy case (`.filter((slug): slug is string => Boolean(slug))`), which suggests the second option matches the intent.

kingmakeruix:
## Summary

Milestones whose URL yields no slug were keyed under an empty id, so stored completion never matched them and their progress read 0%. `getMilestoneSlug` returns `''` rather than a nullish value, which made the `?? m.url` fallback unreachable. This change makes the fallback real with `||`, matching how other call sites in the same file already treat an empty slug as falsy.

## How to verify

1. In `src/docs-retrieval/learning-journey-helpers.ts` ([helpers, line 150](https://github.com/kingmakeruix/grafana-pathfinder-app/blob/afa1d0a2c3ed41c823563549312836b72b9ffc96/src/docs-retrieval/learning-journey-helpers.ts)), confirm the member id reads `getMilestoneSlug(m.url) || m.url`.
2. Run the new regression test (`falls back to the milestone url when the url yields no slug` in [learning-journey-helpers.test.ts](https://github.com/kingmakeruix/grafana-pathfinder-app/blob/afa1d0a2c3ed41c823563549312836b72b9ffc96/src/docs-retrieval/learning-journey-helpers.test.ts)): with the fix it reports 100% for a completion record stored under the raw URL; reverting the one-character change drops it to 0%.

## Breaking changes

Fixes #1932

Written with AI assistance (Muse Spark) under human direction and review.


moxious:
Thanks for the fix to fall back to the milestone URL when the slug is empty — that's a useful robustness improvement for the docs-retrieval path.

One thing to clear before we review: our Contributor License Agreement isn't signed yet. You can sign it via the CLA check on this PR, or see our [contributing guide](https://github.com/grafana/grafana-pathfinder-app/blob/611d123c9fc4b23a0aac1caf71e7cdd550636edc/CONTRIBUTING.md) for details.

Issue #1653, opened 2026-08-19
completion-records: namespace-wide read-cache invalidation lets one user's writes hold the shared index cold


tomglenn:
The completion-records read index is cached per **namespace**, but it is invalidated by **per-user** events. This issue is about whether that keying is what we want before the write path carries real traffic. It is a design question, not a defect.

## Where this lives

`invalidateCompletionIndex(namespace)` arrives with #1433 (the durable completion-record write proxy — approved, not yet merged). Nothing described here is on `main` today: `main` has the read proxy only, so there is no invalidation and no write rate limit yet.

The read proxy caches one collated index per namespace and serves it for `completionCacheTTL = 5m`. The write proxy calls `invalidateCompletionIndex` after every successful create, so a newly written completion surfaces promptly instead of up to five minutes later. That is the right thing to want.

The tension is that the cache is one shared entry per stack while the trigger is one user finishing one guide. **One person's write invalidates the index for every reader in the namespace.**

## Why it is worth deciding rather than leaving

The write path permits `completionWriteRateRefillPerSec = 1.0` sustained per user (burst 20). A single active user therefore invalidates the shared index far faster than the five-minute TTL would ever expire it, and every other reader in that namespace pays an upstream LIST on their next read.

Two things follow:

- **Under load the cache stops behaving like a cache.** On a busy stack the steady state is a cold index, and the TTL that was sized to protect the upstream stops doing so.
- **The cost lands on people unrelated to whoever triggered it.** It scales with the number of writers and is paid by the number of readers.

Neither is a correctness bug. Both are consequences of a namespace-keyed cache being invalidated by per-user events.

## Options

- **Per-user index.** Invalidation becomes precise, at the cost of N entries per namespace and N upstream LISTs to warm them. The upstream LIST is already namespace-scoped and paginated, so this trades one shared cost for many small ones.
- **Debounce or coalesce invalidation.** Bound how often a namespace can be invalidated (say at most once per M seconds) so a burst of writes costs one refresh rather than one per write. Cheapest change; caps the amplification without touching the keying.
- **Merge the written record into the cached index** instead of invalidating, since the writer already holds the fact it just committed. Keeps the cache warm and makes the post-write read correct with no upstream round trip. Most work, best steady state.

## Already resolved — do not re-open

[...]

PR #1950, opened 2026-09-18
Fix #1928: Prevent sibling path progress wipe on reset

### LINKED ISSUE #1928: Resetting a learning path clears progress for a sibling path whose URL shares its prefix

tomglenn:
## Summary

Fixes #1928 - Resetting a learning path was clearing progress for sibling paths whose URLs share its prefix.

## Root Cause

The `resetPath` function used `startsWith()` to discover completion keys in storage:

```typescript
// VULNERABLE
const milestoneKeys = Object.keys(completions).filter((key) => key.startsWith(normalizedUrl));
const journeyKeys = [path.url, ...Object.keys(journeyCompletions).filter((k) => k.startsWith(normalizedUrl))];
```

This caused **prefix collision**: resetting `https://grafana.com/docs/learning-journeys/alerting` also matched and cleared `https://grafana.com/docs/learning-journeys/alerting-advanced`.

### Vulnerable Code Locations

1. `src/learning-paths/learning-paths.hook.ts:538-539` - milestoneKeys and journeyKeys filters
2. `src/lib/user-storage.ts:1500` - `guideCompletionMarkStorage.clearAllWithPrefix()`

### Affected Storage Namespaces

- `interactiveCompletionStorage` - stored as flat JSON objects with raw URLs as keys
- `journeyCompletionStorage` - stored as flat JSON objects with raw URLs as keys  
- `guideCompletionMarkStorage` - versioned keys, but `clearAllWithPrefix` used raw `startsWith` on parsed contentKey

Other storages were already safe due to exact-match or canonicalized URL comparison.

## Solution

Introduced a **delimiter-bounded prefix matcher** that prevents sibling collisions:

```typescript
export function isKeyUnderPrefix(key: string, prefix: string): boolean {
  if (prefix === '') return true;  // Empty prefix = clear all
  if (key === prefix) return true;  // Exact match
  const boundary = prefix.endsWith('/') ? prefix : prefix + '/';
  return key.startsWith(boundary);  // Must be child in hierarchy
}
```

### Why This Works

- `/alerting` (prefix) matches `/alerting` (exact) ✅
- `/alerting` (prefix) matches `/alerting/milestone-1` (child) ✅
- `/alerting` (prefix) does NOT match `/alerting-advanced` (sibling) ❌

### 1. Added Helper Function

**File:** `src/lib/storage/key-utils.ts`
- Added `isKeyUnderPrefix(key, prefix)` with comprehensive documentation
- Handles empty prefix (clear-all semantics)
- Handles prefix already ending with `/`
- Uses `/` as hierarchy delimiter

### 2. Fixed Vulnerable Call Sites

**File:** `src/learning-paths/learning-paths.hook.ts` (lines 538-539)
- Replaced `startsWith(normalizedUrl)` with `isKeyUnderPrefix(key, normalizedUrl)`
- Applied to both `milestoneKeys` and `journeyKeys` filters

**File:** `src/lib/user-storage.ts` (line 1500)
- Replaced `parsed.contentKey.startsWith(contentKeyPrefix)` with `isKeyUnderPrefix(parsed.contentKey, contentKeyPrefix)`
- Fixes `guideCompletionMarkStorage.clearAllWithPrefix()`

### 3. Comprehensive Test Coverage

**File:** `src/lib/storage/key-utils.test.ts` (+58 lines)
- Empty prefix matches everything (clear-all)
- Exact match
- Child paths (slash-delimited hierarchy)
- **Sibling paths that share string prefix (the bug case)**
- Prefix already ending with slash
- Key shorter than prefix
- Key with trailing slash
- Bundled scheme keys (`bundled:welcome` vs `bundled:welcome-cloud`)
- Full URL keys

**File:** `src/learning-paths/reset-path-completion.test.ts` (+45 lines)
- **Regression test for #1928**: Proves resetting `/alerting` does NOT clear `/alerting-advanced`
- Verifies target path and all milestones are cleared
- Verifies sibling path and its milestones remain untouched
- Tests both `interactiveCompletionStorage`, `journeyCompletionStorage`, and `guideCompletionMarkStorage`

## Verification

✅ **All tests pass** (18 tests in key-utils.test.ts, 12 tests in reset-path-completion.test.ts)  
✅ **TypeScript compilation clean** (`npx tsc --noEmit`)  
✅ **Linting/formatting clean** (ran automatically on commit)

## Migration Notes

✅ **No migration required** - existing storage keys unchanged  
✅ **Fully backward compatible** - only changes key matching logic  
✅ **No breaking changes** - fix is purely additive

## Known Limitations & Follow-Ups

The following questions from the investigation plan are **intentionally deferred** for separate discussion:

### 1. Query/Hash URL Edge Cases (Q1)

**Issue:** If milestone keys can contain query strings or anchors (e.g., `https://example.com/path?v=1`), the `/` delimiter won't correctly identify children.

**Current assumption:** Tests and production code use path-style URLs without query/hash. No explicit validation forbids them.

**Options:**
- A) Document that URL-based path keys must not contain query/hash (current implicit assumption)
- B) Add URL-aware normalization that strips query/hash before comparison
- C) Add validation that rejects paths with query/hash URLs

**Recommendation:** Option A is sufficient unless there's evidence of query/hash URLs in production data.

### 2. Related Issues #1653 and #1586 (Q2)

**Finding:** No references to these issues found in codebase. Issues #1863 and #1864 (collision-safe storage) were found and appear to be the related work.

**Action needed:** Verify whether #1653 and #1586 are duplicates, separate issues, or already resolved.

### 3. Architecture Test or Lint Rule (Q3)

**Issue:** The vulnerable `startsWith` pattern could recur if future code uses raw prefix matching on these stores.

**Options:**
- Add an architecture test that fails if `startsWith` is used on completion store keys without the helper
- Add a code comment warning against raw `startsWith`
- Nothing (rely on code review)

**Recommendation:** Consider adding a comment or lint rule as future-proofing.

## Testing Instructions

1. Create two learning paths: `/alerting` and `/alerting-advanced`
2. Complete some milestones in both paths
3. Reset `/alerting` path
4. Verify `/alerting-advanced` progress remains intact
5. Verify `/alerting` progress is fully cleared

---

**Severity:** HIGH (data loss)  
**Type:** Bug fix  
**Breaking:** No


moxious (commented):

Follow-ups:

1. [follow_up] **reset-prefix-docs-stale — Every doc statement of the reset sweep still describes an unbounded prefix match** (low · cross-cutting-architecture)
   The delimiter boundary is now the only thing preventing #1928, and the issue itself records that nothing else enforces it — yet no prose says so. `docs/developer/learning-paths/README.md:183` still reads "discovers interactive and journey completion keys by normalized URL prefix", line 257 says marks are "cleared by resetPath (by content-key prefix on the URL branch)", and the JSDoc on the destructive bulk-clear at `src/lib/user-storage.ts:1490` still promises "every mark whose content key starts with contentKeyPrefix" while line 1500 now rejects keys that do start with it. A future author adding a fourth sweep to resetPath would reintroduce the bug while matching the documented contract.
   Follow-up: State the boundary in all three places: discovery matches the path URL itself and its slash-delimited children, never a sibling that merely shares the text prefix. For `clearAllWithPrefix`, also state the trailing-slash precondition that its one caller satisfies today by stripping.

Suggestions & nits:

1. [suggestion] **pr-description-file-policy — PR_DESCRIPTION.md is committed at the repo root against the file-creation policy** (low · cross-cutting-architecture)
   The branch adds a tracked 90-line `PR_DESCRIPTION.md` at the repo root, duplicating the PR body. AGENTS.md's file-creation policy says not to create summary `.md` files and to report summaries in chat instead, and nothing in `npm run check` guards the root doc set, so CI will not catch it. Merged, it becomes permanent repository content with no owner and no update path, carrying undated verification claims and open questions about #1653/#1586.
   Suggested: Delete PR_DESCRIPTION.md from the branch; its content already lives in the GitHub PR body.

2. [nit] **prefix-boundary-query-fragment — Query and fragment are not boundaries, so one cover-key shape stops being cleared** (low · correctness-and-reliability)
   The fix narrows matching, and one shape falls out the other side. Because `normalizedUrl` strips the trailing slash, a content key of the form `https://.../alerting?utm=x` or `https://.../alerting#top` no longer matches, where the base `startsWith` cleared it. That shape is persistable: `src/global-state/content-key.ts:26-28` sanitizes the active tab URL with only `replace(/\.\./g,'')` and a 200-char slice, stripping neither `?` nor `#`. The effect is the mirror image of #1928 — a cover that still reads complete after the reader asked to start over, with no error. The commoner `.../alerting/?utm=x` and `.../alerting/#top` forms still clear, and no writer was found that demonstrably emits the no-slash shape, which is why this is small.
   Suggested: Treat `?` and `#` as boundary characters alongside `/` in `isKeyUnderPrefix` (or strip them from `key` before the test), and pin both shapes in `src/lib/storage/key-utils.test.ts` so the semantics are chosen rather than left as a documented gap.

3. [nit] **key-utils-module-scope-doc — key-utils module docblock misdescribes its own exports after the addition** (low · correctness-and-reliability)
   `src/lib/storage/key-utils.ts:4` scopes the module to "Two prefix-sweep operations" and lines 12-14 promise "These helpers operate on a Storage instance directly". `isKeyUnderPrefix` is the third export, takes no Storage, and touches no storage — it is a pure predicate over content keys, which is the kind of thing `src/lib/storage-keys.ts` already owns. That docblock is the only statement of what belongs in the module, so leaving it stale stops it steering the next author. The new JSDoc also carries a four-bullet "Matching rules" list that restates the four lines of the body.
   Suggested: Reword the module docblock to cover both storage sweeps and key-shape predicates, or home `isKeyUnderPrefix` in `src/lib/storage-keys.ts` beside the other key-shape functions. Drop the "Matching rules" bullets and keep the sibling-collision rationale and the query/fragment caveat, which are the load-bearing parts.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1950
Summary: Fix #1928: Prevent sibling path progress wipe on reset
Verdict: Approve with Minor
Results: 0 blockers, 3 non-blocking findings, 1 follow-up

tomglenn:
Thanks @moxious — folded in the follow-ups from your review (pushed as `5ef792d3`):

1. **reset-prefix-docs-stale** — documented the delimiter boundary in all three spots: `README.md` URL-path discovery para + the `guideCompletionMarkStorage` bullet, and the `clearAllWithPrefix` JSDoc (including the trailing-slash precondition its one caller satisfies by stripping via `normalizedUrl`). Each now says matching is the path URL and its `/`,`?`,`#`-delimited children, never a text-prefix sibling.
2. **prefix-boundary-query-fragment** — `isKeyUnderPrefix` now treats `?` and `#` as boundaries alongside `/`, so the no-trailing-slash `.../alerting?utm=x` and `.../alerting#top` shapes are cleared again while `/alerting-advanced` stays spared. Both shapes (and the sibling-not-rescued case) are pinned in `key-utils.test.ts`.
3. **key-utils-module-scope-doc** — reworded the module docblock to cover the key-shape predicate alongside the two storage sweeps, and dropped the redundant "Matching rules" bullets (kept the #1928 rationale + query/fragment note).
4. **pr-description-file-policy** — deleted `PR_DESCRIPTION.md`; its content already lives in this PR body.

typecheck + lint clean; `key-utils` suite 20/20.

moxious (approved):

PR #1952, opened 2026-09-18
test(validation): guard guided-action documentation against drift

### LINKED ISSUE #1910: Guided-action table in json-guide-format.md is not drift-checked against GUIDED_ACTION_TYPES

loglapa:
## Summary

The guided-action reference table could drift from the accepted action vocabulary without failing CI. Extend the existing documentation guard to require one row per action and the correct marker in the `guided` column.

Accepted actions come from `GUIDED_ACTION_TYPES`; rejected actions come from `JsonInteractiveActionSchema.options` minus that tuple, following the clarification in #1910 and excluding container verbs. This follows the existing guided-action contract; only the documentation test changes.

## How to verify

1. Temporarily add an undocumented verb to `GUIDED_ACTION_TYPES` and run the focused `json-guide-format-docs.test.ts` suite; expect a failure naming the missing action and documentation file.
2. Change the table’s `navigate` guided marker from ❌ to ✅ and rerun; expect a failure naming the action and expected marker.
3. Restore the temporary edits and rerun the focused suite.

## Breaking changes

None. Fully reversible.

Closes #1910


moxious (approved):

PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1952
Summary: guard guided-action documentation against drift
Verdict: Approve

PR #1953, opened 2026-09-18
Fix interactive steps inside conditional blocks earning no progress cred

### LINKED ISSUE #1930: Interactive steps inside a conditional block earn no progress credit
Real interactive work inside a conditional block earns no progress credit, and the conditional itself can never be earned either. A guide with interactive steps inside a conditional branch reports a percentage that cannot rise past the position before the conditional, silently.

## Mechanism

Two rules in `src/lib/guide-stats/` combine badly.

`block-index.ts` treats `conditional` as an opaque parent:

```
 * - `multistep`, `guided`, `conditional`, and `snippet-ref` count as exactly
 *   ONE block each, with the traversal never entering them.
```

So blocks inside `whenTrue` / `whenFalse` have no position in the index, and evidence from them credits nothing.

`completion-affordance.ts` then omits `conditional` from `COMPLETION_AFFORDANCE_BLOCK_TYPES`. `multistep`, `guided`, `quiz`, `terminal`, `code-block` and `challenge` are all present; `conditional` is the one opaque parent that emits no completion evidence of its own.

The result is the exact failure that file's own header warns about: a counted block that emits no evidence. The conditional adds one to the denominator that can never be earned, and everything inside it is invisible to the numerator.

## Impact

A reader does real work — runs steps, completes a terminal exercise — and the percentage does not move. It is an accuracy defect, not a completion blocker: a guide completes through the "Mark complete" flag rather than by reaching 100%, so the guide still completes.

Bounded to guides that place interactive content inside conditional blocks.

## What to look at

Two candidate shapes, and the choice needs some thought:

1. Let `conditional` emit completion evidence when its taken branch completes. Smallest change, keeps traversal opaque, but credits one position for a branch of any size.
2. Let the index descend into the taken branch. More accurate, but the denominator then depends on which branch the reader takes, and the denominator is currently a pure function of the guide, computed the same way by the CLI, the frontend and any other consumer.

Option 2 changes a property the counting rule relies on, so option 1 is likely the right first step. Whichever is chosen, `completion-affordance.parity.test.ts` guards the step-type registry and will need to stay in step.

tomglenn:
# Fix interactive steps inside conditional blocks earning no progress credit

Closes #1930

## Summary

Interactive steps inside conditional blocks (`whenTrue`/`whenFalse`) execute correctly and persist their completion to storage, but contribute zero progress toward guide completion. This happens because the completion index treats `conditional` as opaque—it counts as 1 block without descending into branches—so branch-child step IDs have no position mapping. Evidence lookup returns 0.

This PR implements Option D from the investigation: alias branch-child completion evidence to the parent conditional's single indexed position. Branch-internal work now earns credit while the static denominator is preserved.

### `src/lib/guide-stats/block-index.ts`
- Added `branchChildPositions: ReadonlyMap<string, number>` to `GuideBlockIndex` interface
- Added `collectBranchChildStepIds()` to traverse `whenTrue`/`whenFalse` branches and collect step IDs
- Added `collectBranchChildrenRecursive()` helper for transparent containers (sections, assistants) inside branches
- Branch-child step IDs map to the conditional's counted position
- Handles nested conditionals (inner branch children map to outer conditional)
- Respects snippet-ref boundaries (stops collecting after ref inside branch; skips entirely if conditional's path is shifted)

### `src/lib/guide-stats/progress.ts`
- Modified `evidencedPosition()` to check `branchChildPositions` first in the lookup chain
- Lookup order: `branchChildPositions` → `positionsByStepId` → `positionsById` → 0

### `src/lib/guide-stats/block-index.test.ts`
Added 15 tests for `branchChildPositions`: basic mapping, empty branches, nested conditionals, sections inside branches, snippet-ref handling, non-completable content, one-sided conditionals.

### `src/lib/guide-stats/progress.test.ts`
Added 7 tests for branch child evidence: position credit, lookup precedence, 100% reachability, nested structures, unknown signals.

## Testing

- `npm run typecheck` — passes
- `npm run lint` — passes
- `npm run test:ci -- src/lib/guide-stats/` — 174 tests pass

## Behavior preserved

- Conditional still counts as ONE block in the denominator
- `conditional` remains in `OPAQUE_PARENT_BLOCK_TYPES`
- 100% is always reachable
- Progress is monotonic

## Risks / Unresolved

None identified. The change is additive and follows the existing `containerEndPositions` pattern for position aliasing.


moxious (approved):

Follow-ups:

1. [follow_up] **guide-stats-version-not-bumped — Numerator rule change ships without a GUIDE_STATS_VERSION bump** (low · completion-records)
   The completion-records contract anchor requires bumping GUIDE_STATS_VERSION on any counting-rule change ("without a bump, an old-rule stamp and a new-rule stamp are byte-indistinguishable"), and this PR changes the numerator side of the rule (which step ids credit which position) via the new branchChildPositions lookup, but GUIDE_STATS_VERSION is still 1. Checked impact today: GuideStatsSummary's stamped fields (blockCount, sectionCount, completableBlockCount, finalCompletablePosition) are unaffected by this diff, and CompletionFact.completionPercent is only recorded at terminal (~100%) completion, so no currently-persisted, version-tagged value is actually misinterpreted yet.
   Follow-up: Bump GUIDE_STATS_VERSION, or add a short note recording that this contract's bump requirement is denominator-scoped and this numerator-only change is exempt, so the next person changing this file has the same evidence trail.

Suggestions & nits:

1. [suggestion] **branch-credit-first-child-wins — Any one branch-child action grants full credit for the conditional, not completing the whole taken branch** (medium · completion-records)
   collectBranchChildStepIds() maps every step id inside whenTrue/whenFalse to the same conditional position, so the first branch-child "do-it" earns full credit for that slot, not the last one. Issue #1930's own triage comment explicitly flagged this exact question ("FIRST vs EVERY" evidence-emitting block) as an open judgment call needing a decision, and the issue has no follow-up comment recording that a maintainer decided "first" before this PR implemented it. It is a real, deliberate, tested choice, not a bug, but it changes what a completed guide's percentage means for the guides named in the issue (e.g. k8s-connect-first-cluster, 3 interactive steps in one branch): a reader who does 1 of 3 sees the same credit as one who does all 3.
   Suggested: Get explicit sign-off from whoever triaged #1930 (or note it in the PR/issue) that first-branch-child-completes-the-slot is the accepted semantics, closing the loop the issue's own triage asked for.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1953
Summary: credit branch-child completion evidence to the parent conditional
Verdict: Approve with Minor

PR #1955, opened 2026-09-18
fix(interactive-tutorial): clean up abandoned terminal provisioning

### LINKED ISSUE #1896: handleStart's post-openTerminal continuation has no liveness guard — unmount mid-provision can register a phantom session holder and silently break another block's Cancel/Skip
### Context

PR https://github.com/grafana/grafana-pathfinder-app/pull/1768 ("Wire requirements/objectives/skippable through ChallengeBlock") carries a session-holder refcounting fix in `ChallengeBlock` (module-level `activeSessionHolders` / `activeConnectingInstances` maps, `claimSession`/`releaseSession`, `hasOtherSessionHolders`/`hasOtherConnectingInstances`) that closes a worse, already-verified bug: without it, one ChallengeBlock's Cancel/Skip could disconnect a Coda sandbox session that a *different*, still-live ChallengeBlock was legitimately using — killing another learner's active session outright.

That fix is a real improvement and should not block on this issue. But round-14 review of #1768 (finding F17) identified a narrower defect it introduces, filed here separately per that review's recommendation.

### Root cause

In `src/components/interactive-tutorial/challenge-block.tsx`, `handleStart`'s only guard on the continuation after `await terminalCtx.openTerminal(vmOpts)` is `cancelRequestedRef.current`, which is set **only** by `handleCancel`. There is no mounted/liveness guard.

```ts
const nextSessionId = await terminalCtx.openTerminal(vmOpts);
if (cancelRequestedRef.current) {
  ...
  return;
}
if (nextSessionId) {
  claimSession(nextSessionId);   // <-- registers instanceId into activeSessionHolders
  ...
  runSetup(nextSessionId);       // <-- re-claims + runs real exec calls against the sandbox
}
```

If the `ChallengeBlock` instance **unmounts** while `openTerminal()` is still pending (e.g. the learner navigates to a different guide/step before a slow VM provision resolves, without clicking Cancel first), the unmount cleanup effect only runs `releaseSession()`:

```ts
useEffect(() => {
  return () => {
    releaseSession();
  };
}, [releaseSession]);
```

At the moment of unmount, `heldSessionIdRef.current` is still `null` (nothing has been claimed yet — `claimSession` hasn't run), so `releaseSession()` has nothing to undo for this in-flight request. It does *not* stop the pending `handleStart` continuation — React does not cancel in-flight promises on unmount.

When `openTerminal()` eventually resolves (the VM finishes provisioning), the dead closure's continuation still runs: `cancelRequestedRef.current` is `false` (Cancel was never clicked), so it unconditionally calls `claimSession(nextSessionId)` — which inserts the unmounted instance's `instanceId` into the module-level `activeSessionHolders` map for that session — and then `runSetup(nextSessionId)`, which actually executes setup commands against the sandbox. Nothing will ever call `unregisterSessionHolder` for that `(sessionId, instanceId)` pair again, because the component's single unmount cleanup already fired before this registration happened.

`hasOtherSessionHolders(sessionId, instanceId)` / `hasOtherConnectingInstances(instanceId)` — the functions this round's fix added specifically so `handleCancel` can tell whether another live block still needs the shared session — then see this phantom holder as a legitimate second owner. A later, *actually live* ChallengeBlock sharing that same Coda session can have its own legitimate Cancel/Skip silently skip the `terminalCtx.disconnect()` call it should make (`isOwner && !hasOther` becomes false), leaving that sandbox VM running rather than being torn down when the learner intended to end it.

No test in `challenge-block.test.tsx` exercises unmount while `openTerminal()` is still pending — the existing `unmount()` calls in the suite (e.g. the counter-reset test) unmount only after the component has already settled into a stable state, never mid-await.

### Reproduction scenario

1. Learner A opens a guide, reaches a `ChallengeBlock` in `coda` mode, clicks **Start challenge**. `openTerminal()` is dispatched and is slow (VM provisioning takes, say, 10-20s).
2. Before it resolves, Learner A navigates away from that step/guide (without clicking Cancel) — the `ChallengeBlock` unmounts.
3. `openTerminal()` eventually resolves with a session id. The dead component's `handleStart` continuation runs anyway, calls `claimSession(nextSessionId)` (registering a phantom, never-released holder) and `runSetup(nextSessionId)`.
4. Learner A (or a different learner sharing the same underlying terminal session, per the shared-session model this fix targets) later opens a *different* `ChallengeBlock` that reuses the same live Coda session, then clicks **Cancel** or **Skip** intending to tear it down.
5. `handleCancel` calls `hasOtherSessionHolders(liveSessionId, instanceId)`, which returns `true` because of the phantom holder from step 3 — so `terminalCtx.disconnect()` is skipped. The Cancel/Skip action silently no-ops on the backend disconnect, leaving the Coda sandbox VM running under a control the UI told the learner had ended.

### Required fix

Per the round-14 review recommendation:
- Track liveness — a mounted ref (e.g. `isMountedRef`), or an `AbortController` wired through `openTerminal` — and check it in `handleStart`'s post-await continuation *before* calling `claimSession`/`runSetup`. If not live, do not register a holder and do not run setup.
- On unmount, if this instance was still awaiting `openTerminal()` (i.e. `handleStart` was in flight and had not yet claimed a session), run the same `hasOtherSessionHolders`/`hasOtherConnectingInstances`-gated disconnect logic that `handleCancel` performs today — not just the local bookkeeping `releaseSession()` currently does — so an instance that dies mid-provision still relinquishes (and, if it turns out to be the sole owner, disconnects) the session it was about to claim.

### Required test coverage

Add to `src/components/interactive-tutorial/challenge-block.test.tsx`:
- Unmount a `ChallengeBlock` while its `openTerminal()` promise is still pending, then resolve it — assert `activeSessionHolders`/`activeConnectingInstances` end up with no holder registered for the dead instance (exported test helper `resetSessionHoldersForTest` already exists for setup/teardown; a similar accessor may be needed to assert on the maps, or assert indirectly via a second block's Cancel behaving correctly afterward).
- A second, live `ChallengeBlock` sharing the same session must still successfully `disconnect()` on Cancel/Skip after a sibling block unmounted mid-provision — i.e. the phantom-holder regression this issue describes must not reproduce.

### Severity / priority

**P2 — bounded resource waste and a real cross-block correctness bug, not an emergency.**

Verified against the Coda backend (`grafana/grafana-coda`, `server/src/vms.js` + `pool-manager.js`): every VM is granted a hard `expires_at` (default 30 min, env-configurable via `VM_LIFETIME_MINUTES`/`MAX_VM_LIFETIME_MINUTES`), swept by a cron reaper (`destroyExpiredVMs`) that runs every ~1 minute regardless of frontend behavior, with an independent guest-side self-shutdown backstop (~40 min from boot) in case the reaper itself is down. So this bug cannot leak a sandbox VM indefinitely — the worst case is roughly **30-40 minutes of wasted/billed compute per occurrence**, plus a silently-broken Cancel/Skip for whoever legitimately owns that shared session during that same bounded window.

This should be fixed in a follow-up soon after #1768 merges (it's a real, easily-triggered gap — "navigate away before Start finishes" is an ordinary learner action, not an edge case), but it should not block #1768, which fixes a strictly worse bug (killing a different learner's *live* session outright, with no bound at all on user-visible impact).

### Reference

Introduced by the session-holder refcounting fix in https://github.com/grafana/grafana-pathfinder-app/pull/1768. Relevant code: `src/components/interactive-tutorial/challenge-block.tsx` (`handleStart`, `handleCancel`, `claimSession`/`releaseSession`, module-level `activeSessionHolders`/`activeConnectingInstances`).

Karna1611:
## What does this PR do?

Fixes a race condition in `ChallengeBlock` where unmounting during an in-flight terminal provision could leave a phantom session holder behind.

`handleStart()` previously checked only `cancelRequestedRef` after `await terminalCtx.openTerminal(...)`. Unmounting a `ChallengeBlock` does not cancel the pending promise, so the continuation could still call `claimSession()` after the component had been unmounted. The unmount cleanup had already run and therefore had nothing to release, leaving a session holder that could never be removed.

A later, live `ChallengeBlock` sharing the same session could then have its legitimate Cancel/Skip disconnect suppressed because `hasOtherSessionHolders()` saw the stale holder.

The fix:

- Adds an `isMountedRef` liveness guard, following the existing pattern in `step-checker.hook.ts`.
- Re-arms the ref inside the mount effect so React StrictMode's effect lifecycle cannot leave it permanently `false`.
- Checks `isMountedRef` alongside `cancelRequestedRef` after `openTerminal()` resolves.
- Routes an abandoned, unmounted provision through the existing ownership/disconnect logic used by Cancel.
- Preserves the existing ownership rules: `disconnect()` only occurs when this instance is the sole provisioner and there are no other session holders or connecting instances.
- Skips `resetToIdle()` on the unmounted path because the component is no longer mounted.

### Regression test

Adds coverage for the case where a `ChallengeBlock` is unmounted while `openTerminal()` is still pending and the promise resolves afterward.

The test verifies:

1. The abandoned provision disconnects after resolving instead of leaving a phantom session holder.
2. A second, live `ChallengeBlock` can still perform its own independent disconnect, proving the abandoned instance no longer interferes with shared-session ownership.

The focused regression test and the 29 existing Cancel/connecting/shared-session tests in `challenge-block.test.tsx` pass.

### Validation

- TypeScript: pass
- ESLint: pass
- Focused regression test: pass
- Existing `challenge-block.test.tsx` tests: 29 pass
- Full Jest suite: 568/577 suites passed

The 9 failing suites were traced to unrelated issues in the review environment:
- Windows path-separator handling in several validation/CLI tests.
- A WebCrypto/pairing-signature failure cascade, consistent with the environment using Node v22.14.0 while the repository declares Node >=24.

None of the failing suites exercise `challenge-block.tsx` or the terminal session-ownership logic changed by this PR.

`npm run check` has not been run to completion, so its result is not being claimed here.

## Linked issue(s)

Fixes #1896

## Checklist

- [ ] `npm run check` passes locally — the pre-merge gate. `npm run check -- --list` prints what it runs.


moxious:
Thanks for cleaning up the abandoned terminal provisioning logic in the interactive tutorial — this closes a real resource-leak path.

One thing to clear before we review: GitHub can't verify the signature on your commit. This repository requires signed commits — see the [contributor guide](https://github.com/grafana/grafana-pathfinder-app/blob/598e9ebc073fed24a50c356d08dcb7432c8c1e8b/CONTRIBUTING.md) for how to set up commit signing, then amend and re-push so verification passes.

Karna1611:
Thanks for the feedback! I’ve set up SSH commit signing on my GitHub account and amended/re-pushed the commit. GitHub is now showing the commit as Verified.

The CLA is also signed and passing. Thanks again for pointing me in the right direction — happy for you to pick it up for review.

moxious (approved):
**Approve.** This fixes a race in `ChallengeBlock` where unmounting during an in-flight terminal provision could leave a phantom session holder behind: the added unmount liveness guard after `openTerminal()` resolves prevents a session claim and UI reset by an already-unmounted component. The change is scoped to the component and its regression test, and the safety gate — base-differenced scanner, CLA, and commit signatures — is clean. No blocking issues.

PR #1957, opened 2026-09-18
feat(learning-paths): show assigned paths on My Paths

dgehring-grafana:
**Draft — not ready to merge.** This is an integration spike for Path Assignments. The Assignment kind is not registered upstream yet, so nothing on a real stack can service this route. Dev-only scaffolding is intentional and will come out once a real backend exists. RFC gaps that still belong to Pathfinder are listed under **Not done yet**.

## Summary

My Paths and the recommendations panel can now show when a learning path is assigned to the signed-in user — with assigned/due/overdue chrome on the existing path card, not a second card type. The plugin backend adds `GET /assignments/my`, an identity-gated App Platform read proxy that returns the caller's slice (or a capability envelope when the stack cannot serve the kind).

This is the first Pathfinder surface for the assignments RFC. It is a draft because the live join against completions, the completion-time status write, the registered Assignment kind, and the removal of local-dev fixtures are still ahead.

## What to look at

- **Plugin backend (`GET /assignments/my`)** — [assignments.go](https://github.com/grafana/grafana-pathfinder-app/blob/b462f45d1583b6d52582c40c79166a20ceecdd49/pkg/plugin/assignments.go) is the load-bearing handler: identity from the verified ID-token `sub` (same helper as completion records), namespace from plugin context, namespace-wide LIST then filter to the caller. [assignments_dev.go](https://github.com/grafana/grafana-pathfinder-app/blob/b462f45d1583b6d52582c40c79166a20ceecdd49/pkg/plugin/assignments_dev.go) is compiled only under the `pathfinderdev` build tag and serves [demo/assignments-fixture.json](https://github.com/grafana/grafana-pathfinder-app/blob/b462f45d1583b6d52582c40c79166a20ceecdd49/demo/assignments-fixture.json). That fixture, the extra catalogue rows, contract goldens, and local-dev docs look like leftover test harness — they are the only way to exercise the UI until a real Assignment API exists, and they are meant to disappear with that work.
- **Client and resolution** — [assignments-client.ts](https://github.com/grafana/grafana-pathfinder-app/blob/b462f45d1583b6d52582c40c79166a20ceecdd49/src/lib/assignments-client.ts) mirrors the custom-guide catalogue client's capability-gated fetch (no `isBackendApiAvailable()` pre-check, so the dev fixture stays reachable). [useMyAssignments.ts](https://github.com/grafana/grafana-pathfinder-app/blob/b462f45d1583b6d52582c40c79166a20ceecdd49/src/learning-paths/useMyAssignments.ts) merges overlapping windows into one card, keeps disjoint cycles separate, and ORs local path completion onto wire `satisfied` until the live completion join lands.
- **My Paths card** — [LearningPathCard.tsx](https://github.com/grafana/grafana-pathfinder-app/blob/b462f45d1583b6d52582c40c79166a20ceecdd49/src/components/LearningPaths/LearningPathCard.tsx) takes assignment badges and due-state styling on the existing card. Estimated time was removed in that refactor because assigned/due badges do not fit the previous chrome. The `x/y guides` marker is still there; it may also go — that is intentional and still under debate, not an accidental leftover.
- **Recommendations panel** — [context-panel.tsx](https://github.com/grafana/grafana-pathfinder-app/blob/b462f45d1583b6d52582c40c79166a20ceecdd49/src/components/docs-panel/context-panel.tsx) badges suggested learning journeys that match an outstanding assignment, so assignment state shows up in the home the user already uses.

## Why

The Pathfinder team is integrating assignments into the platform per the assignments RFC. These changes exist to support and display that effort: a learner should see obligations on paths they already know, not on a parallel object.

A separate assignment card was considered so My Paths would not get denser. That was dropped to follow Nick's design — assignments will live in a home the user is already familiar with, and an assignment is an implication of a path, so the display should change where the path already lives.

## Not done yet

Gaps against [PATH_ASSIGNMENTS.md](https://github.com/grafana/pathfinder-rfcs/blob/main/rfc/PATH_ASSIGNMENTS.md). Pathfinder owns the first two; they are why this stays a draft. Create-time cache writes and the daily sweep stay with the runtime provisioner — §7.2 is explicit that the plugin backend cannot host that job (no scheduler, caller identity only, proxy-only shape).

- **Live join on `GET /assignments/my`.** §6.12 and §7.4 require the route to LIST assignments and completions for the namespace, join them in memory on (subject, assignment target), and return each row with criteria evaluated. Evaluation is the source of truth for the learner view; the cached `status.satisfied` boolean is for consumers that cannot join (the admin Infinity dashboard) and must not be what My Paths reads. Today `unevaluatedSatisfaction` always returns false. The UI then ORs local path completion (`isPathCompleted`) onto that stub. That is a browser stand-in, not the RFC join against `CompletionRecord`. For a path or track target, "evaluated" also means Path Tracks' own derived per-guide check, not a second ad-hoc rollup.
- **Completion-time `status` PATCH.** §6.12 writes the `status.satisfied` cache at three moments. Moment 2 is Pathfinder's: the same request as the completion write, as the caller, PATCHing `assignments/status` so dashboards are not stuck until the next provisioner sweep. No service identity and no queue. This draft has no such write. Updating the CR is also not how the plugin marks the card complete for the learner — that is the live join above. Forging `satisfied: true` as a standalone user action was allowed in §6.1 only because it is equivalent to forging a completion; the designed write is "completion lands → evaluate → cache."
- **Local satisfaction ignores `acceptCompletionsFrom`.** §6.9 says completions before that bound do not count. The UI OR treats any local path completion as satisfaction, so a prior completion can mark a later-cycle assignment done. The live join has to test one completion event against every configured criterion together — never the general-purpose completion aggregate (`collateByUser`), which can report a satisfaction no single completion ever achieved (§6.12).
- **Caller filter happens after the drain budget.** §7.4 says the proxy drains a namespace LIST under an aggregate cap. This handler caps the unfiltered LIST, then filters to the verified caller. When the namespace exceeds 50k records, Kubernetes page order — not the caller — decides which obligations survive, and the 200 still looks complete.

## How to verify

Nothing remote can actually service assignments yet. Local verification uses the `pathfinderdev` fixture documented in [LOCAL_DEV.md](https://github.com/grafana/grafana-pathfinder-app/blob/b462f45d1583b6d52582c40c79166a20ceecdd49/docs/developer/LOCAL_DEV.md).

1. Build a tagged backend (`go build -tags pathfinderdev …`) and restart Grafana so it picks up the new binary (`docker compose restart grafana` if the stack was already running).
2. `curl` `GET /api/plugins/grafana-pathfinder-app/resources/assignments/my` and confirm `capability.available: true` with the fixture rows (including overdue, due today, satisfied 2025 vs outstanding 2026 on `observability-basics`, and no withdrawn record).
3. Open My Paths: assigned cards should show Assigned / due / overdue chrome on the existing path card. Estimated time should be gone. `x/y guides` should still be present.
4. Expand an assigned card and confirm assigned-by and due detail. Edit `demo/assignments-fixture.json`, refresh, and confirm the route re-reads the file with no rebuild.
5. Open the recommendations / docs panel and confirm a suggested path that matches a fixture title (for example github-visualize on OSS dashboards) shows Assigned / due badges.
6. Confirm an untagged (`go build` without `pathfinderdev`) binary does not serve the fixture — `capability.available` should be false on the local stack.

## Breaking changes

None for API consumers. This adds `GET /assignments/my` and is otherwise additive.

User-visible chrome: My Paths cards no longer show estimated time. That is an intentional layout trade for assignment badges and may still change (including whether `x/y guides` stays).

## Reversibility

Revert removes the route, the client, and the card chrome. There is no new persisted storage format. The new Faro event `pathfinder_assignments_unavailable` stops firing on revert; it does not reshape existing warehouse events. The `pathfinderdev` fixture is build-tagged and is not in shipped artifacts.

## Out of scope

- Registering the Assignment kind and pointing this proxy at a real aggregator — next; until then every hosted stack is `upstream-404` / capability-false.
- Removing the pathfinderdev fixture, extra catalogue rows, and local-dev docs — they stay while this is a draft.
- Whether the `x/y guides` marker remains on the card.


moxious:
Quick note on scope: this isn't a line-by-line code review — given the draft status and the fake data, I looked at this at the architecture level instead: what shape is being built, and does it match where we want to land.

## Backend model is the critical path

The proxy, identity, and capability-envelope plumbing here already follows the completion-records pattern closely and reads as solid. The thing that actually unblocks everything else, though, is shaping and hardening the backend model — the `Assignment` kind, the spec/status split, the target reference shape — so the data flowing through this UI can become real instead of hand-typed fixture JSON. Everything downstream (satisfaction, the frontend resolution logic, the badges) is only as good as that model, and right now it's the one piece that's still notional. I'd treat firming that up as the next milestone, ahead of any UI polish.

## `useMyAssignments.ts`: decompose now, not later

This hook is already doing five different jobs in one file — fetch, window-based record clustering, satisfaction fallback, overdue/date math, and catalogue resolution — and every one of those is going to keep growing as real assignment data (and the real join) lands. I'd rather we pay the decomposition cost now, while it's cheap and everything's fake, than after it's accreted into something nobody wants to touch:

- Pull the already-pure functions (`assignmentWindow`, `collapseCluster`, `mergeDuplicates`, `daysUntilDue`, `isOverdue`, `resolveAssignments`) out into their own module, following the repo's `*-core.ts` convention — small, pure, independently testable units rather than closures inside a hook.
- Give the window-merge logic ("do these two records collapse into one card") its own named home. It's a real business rule with room to grow special cases as actual assignment rules show up, and it deserves its own tests, not hook-adjacent ones.
- Name the satisfaction stand-in (`assignment.satisfied || isPathCompleted(...)`) as an explicit, single-purpose function tied to the tracking issue for the real join, so ripping it out later is one call site, not an archaeology dig.

None of this changes behavior — it's purely about making sure the module can keep growing without needing a rewrite in three months.

## RFC inconsistencies — flagging as choices, not bugs

These aren't things I think are wrong; they're spots where the code has made a call (reasonably, for a mocked draft) that the RFC either left open or answers differently. Worth either ratifying deliberately or revising before this leaves draft:

1. **Client-side cache vs. live evaluation.** `assignments-client.ts` caches the wire response for 30s per namespace. RFC §7.4 is explicit that this route must evaluate live so a learner sees their own completion reflected immediately. Harmless today since `satisfied` is stubbed `false`, but once the real join ships this reintroduces exactly the staleness §7.4 rules out.
2. **Target schema shape.** `{pathId, trackId?}` as flat sibling fields answers §12.2, which review explicitly still owns ("this RFC does not commit to field names"). Probably the right answer, but worth surfacing as a proposal rather than letting it become the answer by default.
3. **Satisfaction granularity for track-scoped assignments.** The local stand-in checks whole-path completion (`isPathCompleted(path.id)`) regardless of whether the assignment names a `trackId`. §7.1 wants the specific target's own derived check. Worth adding to the "not done yet" list — it's a sharper gap than "satisfaction isn't real yet."
4. **Silent-drop of unresolvable targets.** Assignments whose `pathId` doesn't resolve against the current catalogue disappear with no signal. This answers §12.13, also explicitly open. Worth at least a log/telemetry line so a real assignment vanishing (renamed/unpublished path) is discoverable rather than silently swallowed.
5. **Aggregate LIST budget.** The 50k record cap is carried over from the completion-records pattern as-is. §7.4 asks for this to be sized "across two kinds rather than one" once the live join holds both LISTs in memory together. Not a problem yet since there's no join, but worth a note so it gets revisited rather than assumed adequate.

Spidy88 (commented):
Looks good to me on a first pass. Seems to be intentionally missing completion record joining, so that simplifies a lot of whats reviewed to the UI elements, which look good. Besides what moxious pointed out, I'd take a design pass: reconsider the default due-badge color — a neutral tone would remove the ambiguity with "completed" without touching the well-differentiated urgent states.

PR #1960, opened 2026-09-21
fix(block-editor): stop trapping authors in JSON mode over a pre-existing duplicate heading

### LINKED ISSUE #1811: Block editor: JSON-mode exit trap for guides with a duplicate-heading error

chiliec:
## What does this PR do?

Fixes the JSON-mode dead end from #1811. Opening an already-published guide whose `blocks[0]` heading duplicates the title, then switching to JSON mode, left the author stuck: `handleExitJsonMode` refused to switch back (strict `duplicate_heading` error), and `handleJsonUndo` / `restoreJsonMode` re-validated at the same strict default, so undo and reload reproduced the trap instead of escaping it.

The fix derives leniency from the `originalJson` snapshot that `handleEnterJsonMode` already captures: if the snapshot itself carries `duplicate_heading`, the JSON pane validates with `allowDuplicateHeading: true`; otherwise it stays strict. So:

- an author editing a pre-existing guide with the heading can leave JSON mode, undo, and reload without being blocked on a fault they didn't introduce;
- pasting a guide that newly introduces the duplicate heading into a clean session is still blocked, matching the "block editor should block on this error" decision from #1764;
- `block-import.ts` other call sites (`ImportGuideModal`, `importGuideFromFile`) are unchanged and remain strict — `parseAndValidateGuide` only gained an optional options parameter.

No new state is cached on `JsonModeState`, so `useBlockPersistence`'s restore allowlist needs no change (per the trap noted in the issue thread).

## Linked issue(s)

Closes #1811

## Tests

Three new cases in `useJsonModeHandlers.test.ts` (exit / undo / restore for a guide that already carried the heading). They fail on `main` (3 failed, 7 passed) and pass with this change. The existing "blocks leaving JSON mode for a guide whose blocks[0] duplicates the title" case (clean session, heading pasted in) still passes.

Ran locally:

- `npx jest src/components/block-editor` — 64 suites, 886 tests passed
- `npx jest src/validation` — passed (incl. `test:governance` import-graph checks, 206 tests)
- `npx tsc --noEmit` — clean
- `npx eslint` + `prettier --check` on the three touched files — clean

Not run locally: `mage` (Go) lint/test, `test:scripts`, `test:review-contract`.

## Checklist

- [x] `npm run check` passes locally — the pre-merge gate (TS steps run; Go steps not run, no `mage` on this machine).


chiliec:
Signed

moxious (approved):
Thanks for tracking this down and fixing it, @chiliec — a pre-existing duplicate heading that traps an author in JSON mode with no way out is exactly the kind of dead end that makes fixing guide content painful, and scoping the fix to only pre-existing duplicates (rather than loosening the check generally) was the right call.

Suggestions & nits:

1. [nit] **1960-setstate-in-updater — Sibling setState called from inside the setJsonModeState updater** (low · correctness-and-reliability)
   handleJsonChange and handleJsonUndo now call setIsJsonValid/setJsonValidationErrors from inside the setJsonModeState(prev => ...) updater callback (useJsonModeHandlers.ts:164-182). React updater functions are expected to be pure and may run more than once; today the calls are idempotent (same prev.originalJson each time) so nothing actually breaks, but the pattern would misbehave if that ever stopped being true.
   Suggested: Read originalJson from state outside the updater (or via a ref) and call validateJsonAgainstSnapshot / setIsJsonValid / setJsonValidationErrors after setJsonModeState, keeping the updater a pure state transform.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1960
Summary: stop trapping authors in JSON mode over a pre-existing duplicate heading
Verdict: Approve with Minor

The nit above is optional and won't block merge — no changes needed from you. Thanks again for the contribution!

PR #1966, opened 2026-09-21
fix(app-platform): proxy settings and custom guide reads through OBO

Jayclifford345:
## Summary

Anonymous Play users inherit Viewer permissions and advertise `pathfindersettings:get`, but direct reads of `pathfindersettings/default` return 403 and prevent Pathfinder startup. The existing backend custom-guide catalogue succeeds anonymously. Custom-guide content reads hit the same service-token rejection. Route settings reads, guide content, and guide publication probes through the existing caller-scoped OBO proxy pattern.

## What to look at

- Adds `GET /api/plugins/grafana-pathfinder-app/resources/pathfinder-settings`. It verifies the caller's identity, derives the namespace from trusted plugin context, and mints an OBO token for that caller. The upstream target is fixed to the `default` settings object; responses retain the spec and resource version, with bounded reads, no redirects, and no shared cache.
- Extends the existing backend proxy contract to singleton settings reads, documented in `docs/design/BACKEND_PROXY_PATTERN.md`. Unlike the optional catalogue, settings errors remain errors. Only explicit upstream 404/405/501 responses allow absent-store fallback; a missing proxy route cannot trigger legacy settings writes.
- Adds `GET /api/plugins/grafana-pathfinder-app/resources/custom-guide?name=<resource name>` for full guide resources. Both guide content loading and the package resolver publication probe use it. The backend fixes the kind and namespace and rejects unsafe resource names. Draft share-link behavior and the resolver publication gate are preserved.
- Frontend reads use the proxies. Administrative writes retain the direct App Platform API and optimistic concurrency checks.

## How to verify

The dev deployment already in progress contains settings commit `cd3b2f77` only. Test settings against that deployment. Guide changes in `1e0c669f` use the same reader but require a subsequent deployment for live verification; they are not yet live-tested. Deploy both backend and frontend; a Graft-only override cannot exercise the new routes.

1. On a dev stack with anonymous Viewer access and an existing `default` settings object, open Pathfinder without signing in. Confirm the proxy returns the persisted spec and `metadata.resourceVersion`, and the sidebar starts with the tenant overrides.
2. Repeat as a signed-in Viewer. Confirm the returned settings match the anonymous read where both identities are authorized.
3. As an admin, change a setting and reload anonymously. Confirm the change appears. Check that concurrent admin edits still conflict rather than overwrite each other.
4. Simulate an upstream 403 or outage, and separately a missing plugin proxy route. Confirm the read fails and an administrative save does not write legacy `jsonData`.
5. On a stack with no settings object, confirm the explicit upstream-absence response retains the existing fallback behavior.

6. After deploying the guide commit, open a published custom guide anonymously from the catalogue and via its direct link. Confirm content loads and publication probes use `/resources/custom-guide`. Check that draft share links still load while published-only discovery excludes drafts.

## Deployment and scope

Roll out matching frontend and backend artifacts together. No data migration or new role grants are included. General startup resilience when settings fail is a separate follow-up; this PR repairs the read transport.


leslie-heinzen (approved):

Suggestions & nits:

1. [suggestion] **shared-item-owner — Keep shared item-read transport with the shared App Platform client** (medium · go-backend)
   The generic item-read handler and `getItem` transport live in the settings-specific file, while the documented shared App Platform client owns transport for resource kinds.
   Suggested: Move generic item transport and the generic GET handler to the shared client area; keep settings-only error classification in the settings route.

2. [suggestion] **method-order — Check the method before validating custom guide names** (low · go-backend)
   Malformed `POST /custom-guide` requests return 400 without the route's 405/Allow contract because name validation runs before the shared method guard.
   Suggested: Check the method before validating the query name, and add coverage for malformed POST requests.

3. [suggestion] **proxy-scope-doc — Update the proxy pattern scope for the new item-read routes** (low · cross-cutting-architecture)
   The document's opening scope still names only completion records and the custom-guide catalogue, although this PR adds settings and custom-guide item-read proxies.
   Suggested: Update the opening scope or explicitly mark item reads as a separate documented extension.



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1966
Summary: proxy settings and custom guide reads through OBO
Verdict: Approve with Minor
Results: 0 blockers, 3 non-blocking findings, 0 follow-ups

PR #1969, opened 2026-09-22
feat(guides): let admins edit public guides as private copies

Jayclifford345:
## Summary

Admins can select **Edit as private guide** from an open public JSON guide’s menu to customize an independent copy in the visual block editor. The copy gets a fresh identity, expands shared snippets into editable blocks, and uses the existing private-guide Save → Publish flow. The public original remains available and unchanged.

Learning journeys and paths are excluded, including cover pages and milestones.

## What to look at

- **Guide menu and eligibility:** organization and server admins only; reject journey/path context and metadata, private sources, previews, and HTML-derived documentation. Bundled guides opened by short ID now carry the same native-JSON marker as package URLs.
- **Editor handoff:** prepare and validate the copy before changing the draft; confirm replacement, then clear prior backend and recording state. This extends the existing authoring workflow while preserving its storage format and save/publish contract.
- **Content portability:** resolve relative media and document links against the original source while keeping Grafana routes and local anchors intact. Markdown containing rewritten URLs is normalized to sanitized HTML within the same editable block; external assets remain hosted at their original URLs.

## Why

Companies need to adapt recommended guides to their own workflows without replacing public content. A standalone snapshot lets admins customize blocks without future upstream guide or snippet updates changing their private version.

## How to verify

1. As an admin, open a standalone public JSON guide and choose **More options → Edit as private guide**. Confirm the visual editor opens with the original blocks, a fresh ID, and “(copy)” in the title. Check that Editors and Viewers do not see the action.
2. Keep an existing editor draft, including unapplied JSON changes. Cancel replacement and confirm the draft is intact; then confirm replacement and check that the copy has no previous publication or recording state.
3. Edit, save, and publish the copy. Confirm it creates a separate private resource and the original guide still opens unchanged. Check shared snippets and relative media links.
4. Move the editor through sidebar, full-screen, and floating views, then reload. Confirm the edited copy survives without replaying the import.
5. Open learning journeys, paths, milestones, private guides, and HTML documentation. Confirm the copy action is absent.

Browser coverage exercises the real UI in local Grafana 12.3 with mocked App Platform settings and private-guide endpoints. Live App Platform persistence has not been validated.

## Breaking changes

None. No backend API, guide schema, storage migration, or recommender changes.

## Reversibility

Reverting removes the entry point; saved copies remain ordinary private guides. Confirming draft replacement discards the prior local working draft, but does not change saved private resources or the public original.

Generated with Codex.


moxious (changes requested):
**Request changes.** The private-copy handoff is coherent and CI is green, but two defects break the advertised copy behavior. Please fix both and add the covering tests.

**1. Challenge-brief links are not made portable (medium)**
`preserve-guide-urls.ts:56` rewrites Markdown URLs only for `content`, `question`, `prompt`, `body`, `text`, and `hint`. A challenge `brief` is also Markdown (`json-guide.schema.ts:699`), and the copy preparation runs this rewriter before the private guide is saved (`private-guide-copy.ts:28`). A public challenge guide whose brief holds a relative image or document link therefore produces a private copy that keeps the link relative, so it resolves from the private-guide location instead of the original source and breaks after save/publish.
_Required:_ include `brief` in the Markdown URL rewrite, and add a test with a relative image or document link in a challenge brief.

**2. Same-guide refresh cancels or hides the copy operation (medium)**
`usePrivateGuideCopy.ts:17` resets the lifecycle whenever the `tab` object reference changes, and line 25 shows pending state only when `operation.tab === tab`. The refresh path replaces the tab with a new object while keeping its ID (`docs-panel.tsx:457`), and the active-tab refresh calls that path (`docs-panel.tsx:1120`). Refreshing the current guide mid-preparation marks the operation inactive; refreshing while the replacement confirmation is showing makes it disappear — the user cannot finish the copy.
_Required:_ key the lifecycle and pending comparisons to a stable identity such as `tab.id`, and add same-ID refresh coverage for both preparation and confirmation.

PR #1973, opened 2026-09-22
feat: sticky guide progress bar

tomglenn:
## Summary

Adds a sticky progress bar to the guide panel so readers always have a persistent, at-a-glance sense of how far through a guide they are. It shows a single standard progress bar with the overall completion percentage, pinned to the top of the panel while the content scrolls.

## What changed

- New `GuideProgressBar` component (`src/components/guide-progress/`).
- Displays overall guide completion as a standard progress bar plus an "X% complete" label.
- Reads live progress from the existing completion store (`peekGuidePercentage` / `subscribeProgress`) via `useSyncExternalStore`, so it stays in sync with the rest of the UI without introducing a second source of truth. The "mark complete" state is already folded into that percentage.
- Only shown when the current content has a guide index.
- Rendered in both guide surfaces: `DocsPanelContentArea`, and pinned in a sticky toolbar in `FloatingPanelContent`.
- Accessible: exposed as `role="progressbar"` with `aria-valuenow` / `aria-valuemin` / `aria-valuemax` and a descriptive label.

## Testing

- Unit tests for the new component (`GuideProgressBar.test.tsx`), covering the rendered percentage and ARIA values, and live updates from the store.
- Updated `FloatingPanelContent` and `MarkCompleteFooter` tests.
- Type checking passes.


moxious (changes requested):
**Request changes.** Both sticky-progress surfaces have reachable lifecycle races. Please fix both and add regression tests.

**1. Progress bar subscribes to the previous guide (high)**
`GuideProgressBar.tsx:40` resolves the content key during render and binds `useSyncExternalStore` to it (lines 42–48). The renderer documents that content-key producers publish in layout effects, so resolving during render yields the _previous_ milestone's key (`content-renderer.tsx:764`) — which is why the existing footer resolves the same key in an effect (`MarkCompleteFooter.tsx:101`). After a reader advances, the sticky bar can stay subscribed to the prior guide and show its percentage, including 100%.
_Required:_ resolve/retain the key after the layout-effect publishers update it, then subscribe to the retained key, and add a milestone-navigation regression test.

**2. Floating and full-screen bars can stay absent (high)**
`FloatingPanelContent.tsx:126` decides whether to mount the bar by synchronously reading `getGuideIndex(getContentKey())`, but the index is only published by a later passive effect (`content-renderer.tsx:786`). `active-guide-index.ts:55` notifies only on eviction; `publishGuideIndex` mutates the map without notifying consumers (lines 79–84). On a cold floating or full-screen open, the first render sees no index and omits the bar, and nothing re-renders when the index publishes — so the bar stays missing until an unrelated render.
_Required:_ make visibility reactive to index publication (or derive it from reactive guide content), and add a cold-open regression test.

PR #1977, opened 2026-09-22
fix: keep learning sidebar available when settings fail

Jayclifford345:
A failed settings read previously replaced the learning sidebar with an error. Keep browsing available using defaults, with diagnostics through the existing console/telemetry warning and no user-facing warning banner or retry button.

This mitigates the sidebar failure in incident #5878. It does not repair the underlying credential or Viewer authorization failures.

Defaults are never published as authoritative settings. Settings-dependent startup and saves still require a successful read. The sidebar remains mounted during later refreshes and adopts recovered settings without losing open content. This extends the tenant-settings contract from #1691; the contract anchor is updated here.

Validation: 55 focused tests passed across sidebar fallback/content preservation, config resolution, and settings-save protection. Targeted ESLint and formatting passed. Full pre-merge gate and live validation of this patch have not been run.

Verify by returning 403/502 from the settings read: browsing should remain available with no warning UI, and a console warning should be recorded. Recovering settings should preserve open content. Saves must remain blocked while the authoritative settings read fails.


Spidy88 (approved):
LGTM

Issue #1970, opened 2026-09-22
Bump GUIDE_STATS_VERSION (or document denominator-scoped exemption) for branchChildPositions numerator change


tomglenn:
Follow-up from review of #1953.

## Context
The completion-records contract anchor requires bumping `GUIDE_STATS_VERSION` on any counting-rule change — *"without a bump, an old-rule stamp and a new-rule stamp are byte-indistinguishable."*

#1953 changes the **numerator** side of the rule (which step ids credit which position) via the new `branchChildPositions` lookup, but `GUIDE_STATS_VERSION` is still `1`.

## Impact checked today (why this is a follow-up, not a blocker)
- `GuideStatsSummary`'s stamped fields (`blockCount`, `sectionCount`, `completableBlockCount`, `finalCompletablePosition`) are unaffected by this diff.
- `CompletionFact.completionPercent` is only recorded at terminal (~100%) completion, so no currently-persisted, version-tagged value is actually misinterpreted yet.

## Ask
Either:
- Bump `GUIDE_STATS_VERSION`, **or**
- Add a short note recording that this contract's bump requirement is denominator-scoped and this numerator-only change is exempt.

Either way, the next person changing this file gets the same evidence trail.

_Filed from PR #1953 review (approve-with-minor); relates to #1930._

PR #1978, opened 2026-09-22
chore(guide-stats): bump GUIDE_STATS_VERSION for the branchChildPositions numerator change

dvd233:
## What

Bumps `GUIDE_STATS_VERSION` from 1 to 2, following the bump option in #1970.

## Why

The contract anchor in `summary.ts` requires a bump on any counting-rule change: *"without a bump, an old-rule stamp and a new-rule stamp are byte-indistinguishable."* #1953 changed the numerator side of the rule (which step ids credit which position via `branchChildPositions`), so stamps written before it must not read as the same rule.

As the issue's impact check notes, no currently-persisted value is misinterpreted (stamped summary fields and denominators are unaffected, and `CompletionFact.completionPercent` is only recorded at terminal completion), which is why this is a follow-up rather than a blocker. I also verified that the `version` field is stamp-only — nothing in `src/`, `pkg/`, or `tests/` compares against it at read time — so the bump is behaviorally safe.

## What changed

- `src/lib/guide-stats/summary.ts`: `GUIDE_STATS_VERSION = 2`, with a provenance comment recording the numerator-change trigger (#1953) and the impact boundary (#1970) so the next person editing the rule gets the same evidence trail.
- `src/cli/__tests__/build-stats.test.ts`, `src/components/block-editor/hooks/useBackendGuides.test.ts`: the three pinned output literals move to `2`, keeping the bump a deliberate tripwire. Input fixtures simulating older stamps stay at `version: 1` (and now also exercise old-stamp replacement).

## Verification

- `src/lib/guide-stats`: 6 suites / 174 tests pass.
- Affected areas (`build-stats`, `useBackendGuides`, `derive-manifest`): 297/297 pass.
- Full Jest run: 579 suites / 9,619 tests pass — the 13 failing suites / 8 failing tests are identical on unmodified `main` (environment-bound e2e-runner/CLI specs), so this change adds no new failures.
- `tsc --noEmit`, ESLint, and Prettier on changed files: clean.


tomglenn (dismissed):

tomglenn (changes requested):
As per Cursor Bugbots review, bundled manifests need to account for this change in version or CI validate guides step fails.

PR #1979, opened 2026-09-22
feat(e2e): support codeblock insertion steps

leslie-heinzen:
## Summary

The e2e runner now executes codeblock Insert steps instead of omitting them. It requires explicit completion and reports insertion failures, so later steps cannot pass merely because the codeblock disappeared.

## What to look at

- Runner execution: drivers own root lookup and requirements checks. Existing plain, multistep, and guided behavior stays unchanged.
- Codeblock driver: uses Insert only, waits for completion, and synchronizes optional Skip. Show me and Copy are not insertion substitutes.
- Product contract: exposes insertion errors, requirement explanations, and skippability. Blocked optional steps retain clickable Skip controls.

## Why

An actual guide failed because its mandatory codeblock was omitted and the next step required its completion. Clicking the product's Insert control exercises the authored behavior without duplicating editor insertion logic in the runner.

This extends the supported step kinds while preserving the existing tracked-root and guide-result ownership contracts.

## How to verify

1. Load this plugin build and run `alerting-query-language-build-query` from interactive-tutorials against the alert-rule editor.
2. Provide its named Prometheus data source and explicitly select Code mode before the run. Confirm Insert places the authored PromQL in Monaco and the following step becomes executable.
3. Repeat with Builder mode active. Confirm Insert fails and the following mandatory step is not executed.
4. Test a blocked optional codeblock. Confirm Skip remains clickable and updates product completion before execution continues.
5. Run the standalone codeblock browser fixtures to check insertion errors, disabled Insert, and root detachment.

## Validation notes

The actual guide recorded four passing runner steps after explicit Code-mode setup. Trace inspection independently confirmed the exact authored query in real Monaco.

This is not backend query-result validation: QuickPizza metrics were absent, and no query-execution request appeared in the trace. Local completion-record persistence also returned HTTP 404.

The full local typecheck remains blocked by 41 confirmed baseline errors.

## Breaking changes

Codeblocks previously reported as unsupported are now executed and can fail a guide. Their DOM state now exposes insertion errors, with additional requirement and skippability attributes. No guide-schema or report-format changes are required.

## Reversibility

Reverting restores the previous unsupported-codeblock behavior. No storage migration is introduced; ordinary progress recorded during test runs remains until reset.

## Out of scope

- Manual-input automation and automatic Code-mode setup.
- Editor-content equality assertions in the driver.
- Backend query-result validation.
- Dynamic discovery and broader partial-coverage semantics.

🤖 Generated with pi


moxious (approved):
**Approve.** The e2e runner now executes codeblock Insert steps instead of skipping them, requires explicit completion, and surfaces insertion failures so later steps cannot pass merely because the codeblock disappeared. Drivers own root lookup and requirements handling; existing plain, multistep, and guided behavior is unchanged. New component, driver, discovery, runner, and browser coverage matches the changed behavior, and the documented public DOM contract keeps a compatibility fallback for older plugin builds. Full review found no blockers or material issues.

PR #1986, opened 2026-09-23
fix(kiosk): preserve panel preferences and diagnose catalog fallbacks

Jayclifford345:
## Summary

Kiosk tile launches now select the sidebar transiently, preserving the user's saved floating/fullscreen preference. An unconfigured kiosk immediately serves bundled cards without a catalog request or failure warning. Catalog recovery identifies the catalog actually shown and emits bounded diagnostics and a typed outcome event.

Follow-up to merged #1971. This PR accounts for all six findings in the [final review](https://github.com/grafana/grafana-pathfinder-app/pull/1971#pullrequestreview-5291560088), the three earlier inline findings, the secret-scan report, and all six investigated-and-dropped findings.

## What to look at

- Launch surfaces: instance tiles stop synthesizing `panelMode=sidebar` and call `setModeTransient('sidebar')` before navigation. Explicit author-supplied deep links retain their existing behavior. This follows the panel-persistence contract: automatic surface selection never overwrites the saved preference.
- Catalog recovery: override → configured → generic → bundled remains available when a default URL is configured. Without a configured default, fallback goes directly to bundled cards. URL validation, credential omission, redirect rejection, cancellation, and URL deduplication remain in place.
- Observability: `pathfinder_kiosk_catalog_loaded` records served tier and degradation. Failure logs carry only tier and bounded reason; rejected-rule logs name the field without catalog values. `KioskDemoStarted` gains `launch_mode`, and the telemetry reference explains the existing mode-dependent meaning of `target_instance`.

### Final review: all six addressed

| Finding | Disposition |
| --- | --- |
| `kiosk-instance-launch-overwrites-panel-mode-preference` | Fixed with transient surface selection. Component regression checks the real panel manager and persisted floating preference; browser regression also checks storage after the guide opens. |
| `kiosk-unconfigured-default-now-fetches-cdn` | Fixed: no configured URL means bundled cards, no catalog fetch, no degradation warning. Administrator reference updated. |
| `kiosk-catalog-fallback-ladder-unobservable` | Fixed: typed served-tier outcome, bounded failure reasons, and invalid-rule field logging. Cancelled loads do not report success/degradation. |
| `kiosk-fallback-warning-misstates-which-catalog-failed` | Fixed: warnings distinguish requested/configured failure and configured/generic/bundled recovery. |
| `kiosk-demo-started-lacks-launch-mode` | Addressed: `instance`/`presentation` payload field, assertions for both modes, and documented `target_instance` semantics and historical-event limitations. |
| `kiosk-rule-targeturl-comment-stale` | Addressed: removed the misleading unconditional comment; administrator documentation already describes presentation-only targets. |

### Earlier findings: no additional product fix needed

| Finding | Disposition and evidence |
| --- | --- |
| [Kiosk exit desyncs Grafana router](https://github.com/grafana/grafana-pathfinder-app/pull/1971#discussion_r4070664026) | Already fixed in #1971. `clearKioskLaunchParams` uses `locationService.replace`; navigation regressions still pass. |
| [Package cards omit journey type](https://github.com/grafana/grafana-pathfinder-app/pull/1971#discussion_r4070664038) | No catalog type change: package resolution derives rendering from the manifest (`path`/`journey` versus standalone `guide`). Current mapping/resolution tests pass; existing browser coverage checks milestone controls. Forcing all three cards to journey would misclassify the standalone guide. |
| [HTTP kiosk links skip SPA navigation](https://github.com/grafana/grafana-pathfinder-app/pull/1971#discussion_r4070864079) | Already fixed in #1971. Same-origin HTTP(S) navigation uses separate protocol/origin/subpath checks; the stricter catalog policy remains intact. Navigation regressions pass. |
| [Secret-scan example URL](https://github.com/grafana/grafana-pathfinder-app/pull/1971#issuecomment-5775422554) | Already addressed: the rejection test constructs dummy username/password fields rather than embedding a credential-shaped URI literal. No real secret was involved; the test remains intact. |

### Investigated-and-dropped findings: retained decisions

These are the six items in the reviewer's [investigation record](https://github.com/grafana/grafana-pathfinder-app/pull/1971#issuecomment-5795658750), not additional unresolved blockers.

| Investigation | Decision |
| --- | --- |
| Banner sanitizer permits CSS URLs/images and target without forced rel | No change: pre-existing sanitizer behavior, with catalog authors restricted by the newer trust validation. This follow-up introduces no new HTML sink or author origin. |
| Current origin is always trusted for catalog overrides | Preserve the documented same-origin capability; no demonstrated weaker attacker primitive in the original investigation. |
| URL kiosk launch bypasses `enableKioskMode` | Preserve the intentional launch contract: the setting controls sidebar/presentation access, while URL launch is separately supported under Pathfinder's overall gate. |
| Global document click listener | No change: bounded human-rate work; only validated kiosk links are intercepted. Listener cleanup and navigation tests remain intact. |
| Invalid `targetUrl` drops entire catalog entries | Preserve fail-closed validation rather than silently retargeting presentation guides. The actionable diagnostic gap is fixed by rejected-field logging above. |
| Unconditional `locationService.replace` on clearing kiosk parameters | No change: replaces the current history entry and preserves route state; no demonstrated user-visible defect. Existing no-reopen regression remains. |

The review also recorded clean checks for tiers/orphans, portal theme context, theme observer behavior, surface-event deduplication, external-store snapshot stability, focus trapping, launch-parameter stripping, subpath redirects, and sentence case. Those are not unresolved findings. Its suggestion to add dedicated kiosk concern routing is review-tooling maintenance outside this runtime fix.

## How to verify

1. Save floating or fullscreen as the panel preference. Open a URL kiosk and launch a tile. Confirm the guide opens in the sidebar; reload and confirm the saved preference was not rewritten.
2. Leave the configured catalog empty and open `?pathfinderKiosk=1`. Confirm bundled cards appear without a catalog network request or warning. Guide content still needs its normal network access.
3. Fail an override while the configured catalog works, then fail the configured catalog while the generic catalog works, then fail every remote candidate. Check each warning names the actual fallback.
4. Include an invalid rule and inspect the warning log for its field name. Check failures log bounded reasons and successful loading records the served tier; cancelling a load should not emit an outcome.
5. Compare instance and presentation tile events: `launch_mode` distinguishes them and `target_instance` retains its existing mode-specific meaning.

The complete kiosk browser suite passed locally on Grafana Enterprise 13.2.2 (13 tests including authentication), using the exact application artifact from the original CI run. The preference regression now establishes and asserts floating mode after startup recovery, immediately before launching the tile: an empty floating panel can legitimately recover to sidebar during cold startup, before the action under test.

## Validation results

The full local pre-merge gate passed all nine steps: typecheck, ESLint, formatting, documentation sync, Go lint and tests, frontend coverage (588 suites; 9,752 tests passed), review-contract tests, and shell-script tests. The kiosk Playwright suite also passed on Grafana Enterprise 13.2.2. The CI rerun for the test-fixture correction is pending.

## Breaking changes

No settings, storage, or catalog-schema migration. The unconfigured default is bundled-first; deployments wanting the remote generic catalog can configure its URL explicitly. Telemetry additions are additive.

## Reversibility

Reverting restores the old runtime behavior. This fix cannot reconstruct preferences already overwritten by #1971, and reverting does not remove telemetry already emitted. No storage migration is performed.

Refs #1971.


moxious (approved):

Follow-ups:

1. [follow_up] **kiosk-compound-failure-message-drops-override — The fallback warning still collapses distinct failures: a failed override plus a failed configured default reads identically to a configured-only failure** (low · correctness-and-reliability)
   The served-tier half of the prior warning blocker is fully fixed, but the failure-attribution half is not. src/components/kiosk/kiosk-rules.ts:143-145 derives the failure clause from failed.has('configured') alone and never consults failed.has('override'), even though reject() at :133-136 put 'override' in that same set and logged it. Reproduced at this head: a failed override plus a failed configured default with the generic catalog succeeding emits "The configured kiosk could not be loaded. Showing the generic learning kiosk.", byte-identical to the no-override-at-all case; the PR pins both at kiosk-rules.test.ts:96-105 and :149-154. The served-tier clause is not a substitute tell, because the attempted-URL dedup at :167-169 skips the configured tier without adding it to failed, so loadKioskData(defaultUrl, defaultUrl) reaches tier 'generic' with only 'override' in failed and says "The requested kiosk could not be loaded. Showing the generic learning kiosk." - that dedup order is itself pinned at :79-80. The bundled row already collapses two cases the same way. Base named the override here and got the served tier wrong, so this PR trades one collapse for another rather than removing it. The banner is the only surface a booth operator reads, and on the compound row it blames the tier they do not control while omitting their own kioskRulesUrl, routing them to escalate instead of checking their URL. Note for the record: the disposition table in the PR body lists this prior finding as fully fixed, which overstates this one row - the served-tier half is fixed, the attribution half is not.
   Follow-up: Filed for follow-up as https://github.com/grafana/grafana-pathfinder-app/issues/1987, not blocking this merge. Derive the failure clause from the whole failed set and add a third compound string naming both tiers, then pin the compound row separately in kiosk-rules.test.ts. Do not simply invert the precedence to name the override first: because the dedup path can reach the generic tier with only the override in failed, inverting relocates the collapse and would make the bundled row a three-way collision.

2. [follow_up] **kiosk-partial-catalog-truncation-reports-undegraded — A catalog that served but lost rules to validation reports degraded false, leaving the likeliest drift visible only in logs** (low · analytics-and-telemetry)
   finish() computes recordKioskCatalogLoaded(tier, failed.size > 0) at kiosk-rules.ts:138-140, and failed only gains a tier in reject() at :133-136, which never runs for per-rule rejections. So a catalog that still parses but whose rules stopped validating under the newer checks serves with degraded false. That is the most likely real drift on a customer-published catalog, and a degraded-true pathfinder_kiosk_catalog_loaded is the signal an on-call person would alert on. It is instead readable only by inspecting the per-rule logger.warn at :104-112, which is also unaggregated - one Faro log item per rejected rule per kiosk open, with no length cap on data.rules.
   Follow-up: Filed for follow-up as https://github.com/grafana/grafana-pathfinder-app/issues/1988, not blocking this merge. Count rejected rules in fetchKioskData and let that count either set the degraded flag or land as a bounded rejected_rules attribute on recordKioskCatalogLoaded, aggregating the per-rule warnings into one log per tier at the same call site. The corresponding TELEMETRY.md sentence does not currently distinguish tier-fallback degradation from partial-catalog degradation.

3. [follow_up] **kiosk-no-test-for-failed-override-on-unconfigured-install — The one catalog combination whose served tier this PR changed has no test** (low · testing-and-verification)
   An unconfigured install with a trusted override that fetches and fails is the only reachable combination in the ladder with no test, and it is also the only one whose served tier differs between base and this head: base reached the generic CDN catalog, this head falls straight to bundled because the generic candidate is nulled when defaultUrl is empty. Both nearby tests dodge it - kiosk-rules.test.ts:87-94 supplies no override at all, and :156-162 uses an untrusted host that validateKioskOverride rejects before any fetch, which that test asserts. The behaviour change itself was examined and accepted this round, because the bundled copy is currently deep-equal to what the CDN serves and the emitted warning is accurate, but nothing in the suite pins it.
   Follow-up: Filed for follow-up as https://github.com/grafana/grafana-pathfinder-app/issues/1989, not blocking this merge. Add one case calling loadKioskData('', 'https://interactive-learning.grafana.net/my-kiosk.json') with mockFetch rejecting once, asserting that the fetch did happen and that bundled is the intended served tier.

4. [follow_up] **kiosk-admin-doc-overstates-no-fetch-on-unconfigured-install — Rewritten administrators-reference sentence claims no catalog fetch on an unconfigured install, which is untrue when a selection is supplied** (low · cross-cutting-architecture)
   docs/sources/administrators-reference/_index.md:186, rewritten by this PR, says "If no default is configured, it shows the bundled catalog immediately without fetching a catalog." With defaultUrl empty and a trusted kioskRulesUrl present, the override candidate is non-null and is fetched before the loop falls through to bundled, so neither "immediately" nor "without fetching a catalog" holds. validateKioskOverride trusts interactive-learning.grafana.net and the current origin without any configured default, so this is the documented booth workflow on the same page. Both independent verification passes flagged the same imprecision.
   Follow-up: Filed for follow-up as https://github.com/grafana/grafana-pathfinder-app/issues/1989, not blocking this merge. Scope the clause to the no-selection case, for example "If no default is configured and no selection is supplied, it shows the bundled catalog immediately without fetching a catalog."



PR Review: https://github.com/grafana/grafana-pathfinder-app/pull/1986
Summary: preserve panel preferences and diagnose catalog fallbacks
Verdict: Approve with Minor
Results: 0 blockers, 0 non-blocking findings, 4 follow-ups

